PR #479, opened 2018-11-08
Unify to a single Azure Pipelines definition

natemcmaster:
Changes:

* Only publish artifacts for Windows release builds
* Don't run Debug configs on CI (they're not very useful)
* Remove dependency on aspnet/BuildTools for CI templates


eng/templates/default-build.yml
```diff
@@ -0,0 +1,143 @@
...
+  demands: []
+  beforeBuild: []
+  afterBuild: []
+  codeSign: false
+  variables: {}
+  dependsOn: ''
+  # buildSteps: [] - don't define an empty object default because there is no way in template expression yet to check "if isEmpty(parameters.buildSteps)"
+  # jobName: '' - use agentOs by default.
+  # jobDisplayName: '' - use agentOs by default.
+  # matrix: {} - don't define an empty object default because there is no way in template expression yet to check "if isEmpty(parameters.matrix)"
```
muratg: Do you expect to use these in the future (hence leaving them commented)?

eng/templates/default-build.yml
```diff
@@ -0,0 +1,143 @@
...
+  poolName: ''
+  buildArgs: ''
+  configuration: 'Release'
+  demands: []
+  beforeBuild: []
+  afterBuild: []
+  codeSign: false
+  variables: {}
+  dependsOn: ''
+  # buildSteps: [] - don't define an empty object default because there is no way in template expression yet to check "if isEmpty(parameters.buildSteps)"
```
ryanbrandenburg: Are these 4 commented out on purpose?

eng/templates/default-build.yml
natemcmaster: Possibly. It depends on the outcome of Helix work.

eng/templates/default-build.yml
natemcmaster: Commented out to show usage, but must be commented out be cause if not it would define this as empty object default and there is no way in template expression yet to check "if isEmpty(parameters.buildSteps)"

natemcmaster:
:up: :date: The template had a lot of unused templates because I copied it from aspnet/BuildTools. I've removed the dead code.

PR #488, opened 2018-11-12
Change async void tests to async Task

javiercn:
Switches tests from async void to async Task as this might be causing build hangs.


Tratcher:
How did these run for 2 years?

PR #494, opened 2018-11-14
Drop messages in azure loggers if queue is full

pakrym:
Blocking thread pool threads causes thread pool starvation.
This also prevents waiting for the interval when there are more messages in the queue.

Fixes: https://github.com/aspnet/Logging/issues/939
Fixes: https://github.com/aspnet/Logging/issues/933


pakrym:
@muratg I'd like to port this to 2.1 and 2.2.

muratg:
@DamianEdwards for barcheck. 

I am not sure if this would 2.1.x bar, but I think it would meet 2.2.2. 

@pakrym please prep a 2.2 PR and fill out the shiproom template.

PR #499, opened 2018-11-14
Add solution cache for analyzer DiagnosticProject

pakrym:
Cache a ton of repeated IO.

Takes logging analyzer tests from 30 sec to 5 sec.


src/TestingUtils/Microsoft.AspNetCore.Analyzer.Testing/src/DiagnosticProject.cs
```diff
@@ -23,24 +24,35 @@ public class DiagnosticProject
         /// </summary>
         public static string TestProjectName = "TestProject";
 
+        private static readonly Dictionary<Assembly, Solution> _solutionCache = new Dictionary<Assembly, Solution>();
```
pranavkm: Does this need to be `Concurrent`? xunit runs tests across classes in parallel?

pakrym: That's why I lock around access.

PR #510, opened 2018-11-15
Add Compaction logging

JunTaoLuo:
Reopening https://github.com/aspnet/Caching/pull/441 in the right repo.

Adding logging to test out my theory on https://github.com/aspnet/Caching/issues/429. 

Also fixing a test for https://github.com/aspnet/AspNetCore-Internal/issues/1374.

### DIFF
--- a/src/Caching/Memory/src/MemoryCache.cs
+++ b/src/Caching/Memory/src/MemoryCache.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Caching.Memory
@@ -73,6 +74,8 @@ namespace Microsoft.Extensions.Caching.Memory
 
         // internal for testing
         internal long Size { get => Interlocked.Read(ref _cacheSize); }
+
+        internal ILogger Logger { private get; set; }
 
         private ICollection<KeyValuePair<object, CacheEntry>> EntriesCollection => _entries;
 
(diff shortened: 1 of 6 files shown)


src/Caching/Memory/test/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -553,7 +553,7 @@ public void AddAndReplaceEntries_AreThreadSafe()
                     var entrySize = random.Next(0, 5);
                     cache.Set(random.Next(0, 10), entrySize, new MemoryCacheEntryOptions { Size = entrySize });
                 }
-            }, cts.Token);
```
JunTaoLuo: We don't need to pass the token since we check via cts.IsCancellationRequested inside the tasks. Alternatively we can catch the TaskCancelledException from WaitAll

src/Caching/Memory/src/Microsoft.Extensions.Caching.Memory.csproj
```diff
@@ -11,6 +11,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.Extensions.Caching.Abstractions" />
     <Reference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
+    <Reference Include="Microsoft.Extensions.Logging" />
```
Tratcher: Not Logging.Abstractions?

src/Caching/Memory/src/MemoryCache.cs
```diff
@@ -74,6 +75,8 @@ public int Count
         // internal for testing
         internal long Size { get => Interlocked.Read(ref _cacheSize); }
 
+        internal ILogger Logger { private get; set; }
```
Tratcher: We should at least look at adding additional constructors to inject logging.

poke:
Just out of curiosity: Why is the memory cache injecting a `LoggerFactory` and not directly an `ILogger<MemoryCache>`? Just to be able to use the `NullLoggerFactory` as a default?

PR #530, opened 2018-11-16
Add service validation at provider build time

pakrym:
Fixes: https://github.com/aspnet/AspNetCore/issues/2348

### DIFF
--- a/src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
+++ b/src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
@@ -107,6 +107,24 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <inheritdoc />
         public Func<IServiceProvider, object> ImplementationFactory { get; }
 
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            var lifetime = $"{nameof(ServiceType)}: {ServiceType} {nameof(Lifetime)}: {Lifetime} ";
+
+            if (ImplementationType != null)
+            {
+                return lifetime + $"{nameof(ImplementationType)}: {ImplementationType}";
+            }
+
+            if (ImplementationFactory != null)
+            {
+                return lifetime + $"{nameof(ImplementationFactory)}: {ImplementationFactory}";
+            }
+
+            return lifetime + $"{ nameof(ImplementationInstance)}: { ImplementationInstance}";
+        }
+
         internal Type GetImplementationType()
         {
             if (ImplementationType != null)
--- a/src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
+++ b/src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
@@ -119,10 +119,10 @@ namespace Microsoft.Extensions.DependencyInjection
 
             if (ImplementationFactory != null)
             {
-                return lifetime + $"{nameof(ImplementationFactory)}: {ImplementationFactory}";
+                return lifetime + $"{nameof(ImplementationFactory)}: {ImplementationFactory.Method}";
             }
 
-            return lifetime + $"{ nameof(ImplementationInstance)}: { ImplementationInstance}";
+            return lifetime + $"{nameof(ImplementationInstance)}: {ImplementationInstance}";
         }
 
         internal Type GetImplementationType()
--- a/src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
+++ b/src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
@@ -7,8 +7,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.CompilerServices;
-using System.Threading.Tasks;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
@@ -26,7 +24,7 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         {
             _stackGuard = new StackGuard();
             _descriptors = descriptors.ToList();
-            Populate(descriptors);
+            Populate(_descriptors);
         }
 
         private void Populate(IEnumerable<ServiceDescriptor> descriptors)
@@ -78,6 +76,18 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 #else
             return _callSiteCache.GetOrAdd(serviceType, type => CreateCallSite(type, callSiteChain));
 #endif
+        }
+
+        internal ServiceCallSite GetCallSite(ServiceDescriptor serviceDescriptor, CallSiteChain callSiteChain)
+        {
+            if (_descriptorLookup.TryGetValue(serviceDescriptor.ServiceType, out var descriptor))
+            {
+                return TryCreateExact(serviceDescriptor, serviceDescriptor.ServiceType, callSiteChain, descriptor.GetSlot(serviceDescriptor));
+            }
+
+            // This should never happen
+            Debug.Assert(false);
+            return null;
         }
 
         private ServiceCallSite CreateCallSite(Type serviceType, CallSiteChain callSiteChain)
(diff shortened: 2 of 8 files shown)


src/DependencyInjection/DI/test/ServiceProviderValidationTests.cs
```diff
@@ -109,6 +109,30 @@ public void GetService_DoesNotThrow_WhenScopeFactoryIsInjectedIntoSingleton()
             Assert.NotNull(result);
         }
 
+        [Fact]
+        public void BuildServiceProvider_ThrowsForUnresolvableServices()
```
davidfowl: Can you add some more interesting cases? IEnumerable, open generic etc.

pakrym: Open generics?

davidfowl: So you loop over the service collection and call GetService. What happens when you encounter IFoo<>, Foo<> for example. What ends up happening?

pakrym: Whoops.

pakrym: Done

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -80,6 +78,18 @@ internal ServiceCallSite GetCallSite(Type serviceType, CallSiteChain callSiteCha
...
 
+        internal ServiceCallSite GetCallSite(ServiceDescriptor serviceDescriptor, CallSiteChain callSiteChain)
+        {
+            if (_descriptorLookup.TryGetValue(serviceDescriptor.ServiceType, out var descriptor))
+            {
+                return TryCreateExact(serviceDescriptor, serviceDescriptor.ServiceType, callSiteChain, descriptor.GetSlot(serviceDescriptor));
+            }
+
+            // This should never happen
+            Debug.Assert(false);
```
davidfowl: Debug.Fail?

src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
```diff
@@ -107,6 +107,24 @@ private ServiceDescriptor(Type serviceType, ServiceLifetime lifetime)
...
+            {
+                return lifetime + $"{nameof(ImplementationType)}: {ImplementationType}";
+            }
+
+            if (ImplementationFactory != null)
+            {
+                return lifetime + $"{nameof(ImplementationFactory)}: {ImplementationFactory}";
+            }
+
+            return lifetime + $"{ nameof(ImplementationInstance)}: { ImplementationInstance}";
```
halter73: Nit: There,'s a space in "{ ImplementationInstance}"

src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
```diff
@@ -107,6 +107,24 @@ private ServiceDescriptor(Type serviceType, ServiceLifetime lifetime)
...
+            var lifetime = $"{nameof(ServiceType)}: {ServiceType} {nameof(Lifetime)}: {Lifetime} ";
+
+            if (ImplementationType != null)
+            {
+                return lifetime + $"{nameof(ImplementationType)}: {ImplementationType}";
+            }
+
+            if (ImplementationFactory != null)
+            {
+                return lifetime + $"{nameof(ImplementationFactory)}: {ImplementationFactory}";
```
halter73: Is `Func<IServiceProvider, object>.ToString()` ever useful/interesting?

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -26,7 +24,7 @@ public CallSiteFactory(IEnumerable<ServiceDescriptor> descriptors)
         {
             _stackGuard = new StackGuard();
             _descriptors = descriptors.ToList();
-            Populate(descriptors);
+            Populate(_descriptors);
```
halter73: It's kind of weird that descriptors is even a parameter since it's already in a field.

src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
poke: At the beginning too: `$"{ nameof(`

I also think that the variable name `lifetime` is really odd for what it contains.

src/DependencyInjection/DI/test/ServiceProviderValidationTests.cs
```diff
@@ -109,6 +110,61 @@ public void GetService_DoesNotThrow_WhenScopeFactoryIsInjectedIntoSingleton()
...
+        [Fact]
+        public void BuildServiceProvider_ValidateOnBuild_ValidatesAllDescriptors()
+        {
+            // Arrange
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddTransient<IBaz, BazRecursive>();
+            serviceCollection.AddTransient<IBaz, Baz>();
+
+            // Act + Assert
+            var aggregateException = Assert.Throws<AggregateException>(() => serviceCollection.BuildServiceProvider(new ServiceProviderOptions() { ValidateOnBuild = true }));
```
halter73: With this setup, you could actually resolve IBaz if you tried.

src/DependencyInjection/DI/src/ServiceProviderOptions.cs
```diff
@@ -18,6 +18,11 @@ public class ServiceProviderOptions
         /// </summary>
         public bool ValidateScopes { get; set; }
 
+        /// <summary>
+        /// <c>true</c> to perform check verifying that all services can be created during <code>BuildServiceProvider</code> call; otherwise <c>false</c>.
```
halter73: Might be worth mentioning that ValidateOnBuild defaults to false. We could do the same to ValidateScopes for consistency.

src/DependencyInjection/DI/test/ServiceProviderValidationTests.cs
poke: But not `IEnumerable<IBaz>` so throwing still sounds right.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -415,6 +425,25 @@ public int Count
...
+                if (_items != null)
+                {
+                    var index = _items.IndexOf(descriptor);
+                    if (index != -1)
+                    {
+                        return index + 1;
+                    }
+                }
+
+                throw new InvalidOperationException("Requested service descriptor doesn't exist.");
```
halter73: Is this expected to be possible? Or should it be another Debug.Fail?

src/DependencyInjection/DI/test/ServiceProviderValidationTests.cs
pakrym: Technically IEnumerable can be resolved from this setup too. We just don't support it because our recursive checks are based on service type and not descriptors.

src/DependencyInjection/DI/src/ServiceProviderOptions.cs
halter73: Maybe this should also mention that this doesn't verify open generic services.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
pakrym: Not expected but should fail in release too.

src/DependencyInjection/DI/test/ServiceProviderValidationTests.cs
halter73: Exactly. This setup would only be broken if a service implementing IBaz tried to resolve `IEnumerable<IBaz>`.

src/DependencyInjection/DI.Abstractions/src/ServiceDescriptor.cs
pakrym: I changed it to `ImplementationFactory.Method`

PR #531, opened 2018-11-16
Handle nullable enum default values

pakrym:
Fixes https://github.com/aspnet/AspNetCore/issues/3165

### DIFF
--- a/src/Shared/src/ParameterDefaultValue/ParameterDefaultValue.cs
+++ b/src/Shared/src/ParameterDefaultValue/ParameterDefaultValue.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Extensions.Internal
 {
     internal class ParameterDefaultValue
     {
-        private static Type _nullable = typeof(Nullable<>);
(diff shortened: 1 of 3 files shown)


pakrym:
Test failure in MemoryCache seems completely unrelated:
```
Microsoft.Extensions.Caching.Memory.Tests.Microsoft.Extensions.Caching.Memory.CapacityTests.DoNotAddIfSizeOverflows

Assert.True() Failure
Expected: True
Actual:   False
```

src/Shared/src/ParameterDefaultValue/ParameterDefaultValue.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Extensions.Internal
 {
     internal class ParameterDefaultValue
     {
+        private static Type _nullable = typeof(Nullable<>);
```
pakrym:
```suggestion
        private static readonly Type _nullable = typeof(Nullable<>);
```

PR #533, opened 2018-11-16
Update netcore to 3.0.0-preview-27114-01

natemcmaster:
Update netcore dependency to 3.0.0-preview-27114-01


natemcmaster:
This is going backwards, nevermind.

PR #534, opened 2018-11-16
Add DependencyInjection event source

pakrym:
With one event that dumps call site trees that are being created


src/DependencyInjection/DI/src/CallSiteFormatter.cs
```diff
@@ -0,0 +1,142 @@
...
+        protected override object VisitServiceScopeFactory(ServiceScopeFactoryCallSite serviceScopeFactoryCallSite, CallSiteFormatterContext argument)
+        {
+            argument.ContinueLine("<ScopeFactory>");
+
+            return null;
+        }
+
+        protected override object VisitIEnumerable(IEnumerableCallSite enumerableCallSite, CallSiteFormatterContext argument)
+        {
+            argument.ContinueLine("IEnumerable<{0}> (size {1})", enumerableCallSite.ItemType, enumerableCallSite.ServiceCallSites.Length);
```
davidfowl: nit: Remove size, just put the number

src/DependencyInjection/DI/src/CallSiteFormatter.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                VisitCallSite(parameter, newContext);
+            }
+
+            return null;
+        }
+
+        protected override object VisitConstant(ConstantCallSite constantCallSite, CallSiteFormatterContext argument)
+        {
+            argument.ContinueLine("{0}", constantCallSite.DefaultValue ?? "<null>");
```
anurse: Do we expect this value to be formattable? Most of the time these are user-provided instances of services, right? 

Perhaps just `<Constant: [TypeName]>` would be more appropriate.

src/DependencyInjection/DI/src/CallSiteFormatter.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                VisitCallSite(item, newContext);
+            }
+
+            return null;
+        }
+
+        protected override object VisitFactory(FactoryCallSite factoryCallSite, CallSiteFormatterContext argument)
+        {
+            argument.ContinueLine("Factory ({0})", factoryCallSite.Factory.Method);
```
anurse: The other "special token" values use `<>`, I think that might be appropriate here: `<Factory: [MethodName]>`

src/DependencyInjection/DI/src/CallSiteFormatter.cs
```diff
@@ -0,0 +1,142 @@
...
+
+            return null;
+        }
+
+        protected override object VisitDisposeCache(ServiceCallSite callSite, CallSiteFormatterContext argument)
+        {
+            argument.StartLine("DisposeCache ");
```
anurse: Are these decorators applied to call-sites?

anurse: Might be clearer if these were something like `[DisposeCache] `

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.DependencyInjection.ServiceLookup;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    [EventSource(Name = "Microsoft-Extensions-DependencyInjection")]
+    internal sealed class DependencyInjectionEventSource : EventSource
+    {
+        public static readonly DependencyInjectionEventSource Instance = new DependencyInjectionEventSource();
```
anurse: nit: The idiomatic name of this is `Log`... not sure why, but it just is :)

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource?view=netframework-4.7.2#examples

Not sure what other ASP.NET Core sources use though.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
```diff
@@ -0,0 +1,43 @@
...
+        [Event(1, Level = EventLevel.Verbose)]
+        private void CallSiteBuilt(string callSite)
+        {
+            WriteEvent(1,  callSite);
+        }
+
+        [NonEvent]
+        public void CallSiteBuilt(ServiceCallSite callSite)
+        {
+            if (IsEnabled())
```
anurse: Pass in the level here, to protect us when/if we add non-Verbose events.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        // NOTE
+        // - The 'Start' and 'Stop' suffixes on the following event names have special meaning in EventSource. They
+        //   enable creating 'activities'.
+        //   For more information, take a look at the following blog post:
+        //   https://blogs.msdn.microsoft.com/vancem/2015/09/14/exploring-eventsource-activity-correlation-and-causation-features/
+        // - A stop event's event id must be next one after its start event.
+        // - Avoid renaming methods or parameters marked with EventAttribute. EventSource uses these to form the event object.
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
anurse: Any reason for no-inlining here?

src/DependencyInjection/DI/test/DependencyInjectionEventSourceTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        [Fact]
+        public void EmitsCallSiteBuiltEvent()
+        {
+            // Arrange
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton<FakeDisposeCallback>();
+            serviceCollection.AddTransient<IFakeOuterService, FakeDisposableCallbackOuterService>();
+            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddScoped<IFakeMultipleService, FakeDisposableCallbackInnerService>();
```
anurse: Could you put a couple other types of call-sites in? A Factory and a Constant in particular would be nice.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
```diff
@@ -0,0 +1,43 @@
...
+        [Event(1, Level = EventLevel.Verbose)]
+        private void CallSiteBuilt(string callSite)
```
anurse: Is there anything we should promote to top-level metadata here? The service type name for example?

It's easier to query/filter events by payload values, so things that we think people will want to query should be direct parameters.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: Copied from Kestrel.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: `IsEnabled` doesn't take level.

src/DependencyInjection/DI/src/CallSiteFormatter.cs
pakrym: `object.ToString()` defaults to type name anyway.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: Added serviceType.

src/DependencyInjection/DI/test/DependencyInjectionEventSourceTests.cs
pakrym: Done

pakrym:
I changed the format to JSON. It's not as human readable but better for bulk analysis.

I'll switch to a real JSON writer when we get appropriate BCL version.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: Removed.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: Renamed.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
anurse: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource.isenabled?view=netframework-4.7.2#System_Diagnostics_Tracing_EventSource_IsEnabled_System_Diagnostics_Tracing_EventLevel_System_Diagnostics_Tracing_EventKeywords_ ?

anurse:
> I'll switch to a real JSON writer when we get appropriate BCL version.

File a bug to track this? Otherwise we'll just forget it :)

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: Whoops.

PR #536, opened 2018-11-18
Adding support for constrained open generics

jbogard:
Original discussion: https://github.com/aspnet/DependencyInjection/issues/471
Original pull request: https://github.com/aspnet/DependencyInjection/pull/635


jbogard:
As a last-chance check, it uses first-chance exceptions. We shouldn't hit that point however because we check all the type constraints explicitly. @pakrym @davidfowl

pakrym:
How does the error look when resolving a single instance of open-generic type that doesn't satisfy the constraint?

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +359,103 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
 
+        private static bool IsCompatibleWithGenericParameterConstraints(Type genericTypeDefinition, Type[] parameters)
+        {
+            var genericArgumentDefinitions = genericTypeDefinition.GetTypeInfo().GenericTypeParameters;
+            for (var i = 0; i < genericArgumentDefinitions.Length; ++i)
+            {
+                var argumentDefinitionTypeInfo = genericArgumentDefinitions[i].GetTypeInfo();
+                var parameter = parameters[i];
+                var parameterTypeInfo = parameter.GetTypeInfo();
+                if (argumentDefinitionTypeInfo.GetGenericParameterConstraints()
```
pakrym: Avoid linq on the hot path.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +359,103 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+                   .Select(constraint => SubstituteGenericParameterConstraint(parameters, constraint))
+                   .Any(constraint => !ParameterCompatibleWithTypeConstraint(parameter, constraint)))
+                {
+                    return false;
+                }
+                var specialConstraints = argumentDefinitionTypeInfo.GenericParameterAttributes;
+                if ((specialConstraints & GenericParameterAttributes.DefaultConstructorConstraint)
+                   == GenericParameterAttributes.DefaultConstructorConstraint)
```
pakrym: Join lines.

pakrym:
This needs perf tests, a lot of new code on the hot path for open generics.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
jbogard: Fixed!

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
jbogard: Fixed! And others with multi-line conditionals.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +358,109 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+        private static Type SubstituteGenericParameterConstraint(Type[] parameters, Type constraint)
+        {
+            if (!constraint.IsGenericParameter) return constraint;
+            return parameters[constraint.GenericParameterPosition];
+        }
+
+        private static bool ParameterCompatibleWithTypeConstraint(Type parameter, Type constraint)
+        {
+            bool hasAnyCompatibleType = false;
+            foreach (var p in new[] {parameter, parameter.BaseType}.Concat(parameter.GetTypeInfo().ImplementedInterfaces))
```
pakrym: We avoid LINQ because of the number of allocations it produces. This method allocates way too much for what it does.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +358,109 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+                        var constraintArgument = constraintArguments[i];
+                        if (!constraintArgument.IsGenericParameter && !constraintArgument.IsAssignableFrom(genericArguments[i]))
+                        {
+                            return false;
+                        }
+                    }
+                    Type genericType;
+                    try
+                    {
+                        genericType = typeDefinition.MakeGenericType(genericArguments);
```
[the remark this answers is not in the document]
jbogard: Yes now, `CreateCallSite_ReturnsNull_IfComplexClosedTypeDoesNotSatisfySelfReferencingConstraint` test

src/DependencyInjection/DI/test/ServiceLookup/Types/TypeWithInterfaceConstraint.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System.Collections;
+using Microsoft.Extensions.DependencyInjection.Specification.Fakes;
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    public class TypeWithInterfaceConstraint<T> : IFakeOpenGenericService<T>
+        where T : IEnumerable
```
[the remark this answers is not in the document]
jbogard: Added some more complex cases - abstract class constraints, deep hierarchies, self-referencing constraints

jbogard:
@pakrym I've added positive tests for each of the negative tests

pakrym:
For perf test, it has to be time to the first service or something directly targeting CallSiteFactory. 

Let's do a scenario where we register multiple open generic implementations of the same service and resolve them as `IEnumerable`.

```
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>
IFakeGeneric<> -> FakeGeneric<>

GetService<IEnumerable<IFakeGeneric<Something>>>()
```

pakrym:
We need to move more (all?) tests to spec tests and make sure they pass with other supported containers too.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +358,119 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+                    if (!ParameterCompatibleWithTypeConstraint(parameter, substituteGenericParameterConstraint))
+                    {
+                        return false;
+                    }
+                }
+
+                var specialConstraints = argumentDefinitionTypeInfo.GenericParameterAttributes;
+                if ((specialConstraints & GenericParameterAttributes.DefaultConstructorConstraint) == GenericParameterAttributes.DefaultConstructorConstraint)
+                {
+                    if (!parameterTypeInfo.IsValueType && parameterTypeInfo.DeclaredConstructors.Where(c => c.IsPublic).All(c => c.GetParameters().Length != 0))
```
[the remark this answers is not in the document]
jbogard: Done!

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +358,119 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
             return parameterCallSites;
         }
 
+        private static bool IsCompatibleWithGenericParameterConstraints(Type genericTypeDefinition, Type[] parameters)
+        {
+            var genericArgumentDefinitions = genericTypeDefinition.GetTypeInfo().GenericTypeParameters;
+            for (var i = 0; i < genericArgumentDefinitions.Length; ++i)
```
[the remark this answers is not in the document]
jbogard: Done. I blame Resharper on that one.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +358,119 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+        }
+
+        private static bool IsGenericTypeDefinedBy(Type type, Type openGeneric)
+        {
+            return !type.ContainsGenericParameters && type.IsGenericType && type.GetGenericTypeDefinition() == openGeneric;
+        }
+
+        private static Type SubstituteGenericParameterConstraint(Type[] parameters, Type constraint)
+        {
+            return !constraint.IsGenericParameter ? constraint : parameters[constraint.GenericParameterPosition];
```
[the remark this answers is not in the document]
jbogard: Done.

jbogard:
For the spec tests, I'll copy them over to target the service collection instead of `CallSiteFactory`

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +356,132 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+                    }
+                }
+
+                var specialConstraints = argumentDefinitionTypeInfo.GenericParameterAttributes;
+                if ((specialConstraints & GenericParameterAttributes.DefaultConstructorConstraint) == GenericParameterAttributes.DefaultConstructorConstraint)
+                {
+                    var hasPublicNoArgCtor = false;
+                    foreach (var c in parameterTypeInfo.DeclaredConstructors)
+                    {
+                        if (c.IsPublic)
```
pakrym: Merge ifs into one.

jbogard: Done.

jbogard:
I added more relevant spec tests - but the `class` and `struct` ones didn't make sense as they're not all supported. I don't think it hurts to leave the more specific checks in the `CallSiteFactory` and tests. I think all that's left is perf tests.

pakrym:
> but the class and struct ones didn't make sense as they're not all supported

What do other containers do in these cases?

jbogard:
I take that back - it’s just one of the fake ones.

jbogard:
For perf tests, I'm looking at testing scenarios for:

 - No DI
 - No constraints
 - `new()` constraint (one of the `GenericParameterAttributes` cases)
 - Simple type-based constraint (interface)
 - Complex self-referencing edge case constraint that triggers the exception flow

pakrym:
> Not supported from what I can see - most have some sort of `class` method constraint in the API. So I could either skip those constraint checks, remove the tests, or leave as-is. The public API doesn't support struct-based dependency registration in any case. I wouldn't put any spec tests in, though.

You don't have to inject struct instance to test that constraint is being applied. For example if you remove ctor from https://github.com/aspnet/Extensions/pull/536/files#diff-d2cca72e65acaa575dd3c9657a9605dbR8 it should be supported by all providers.

Perf cases look good.

jbogard:
Ah. Duh. Will update.

jbogard:
Here's the benchmark results locally:

https://gist.github.com/jbogard/a2be5b903c44c8b2b3c2ef0d1e8c1e0c

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
```diff
@@ -0,0 +1,346 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+
+namespace Microsoft.Extensions.DependencyInjection.Performance
+{
+    public class ConstrainedGenericBenchmark
```
pakrym: This not exactly what I'd like to be benchmarked. In this benchmark, we are exercising resolution path but all the changes are in the call site creation code path.

What I would like to see is a benchmark that uses CallSiteFactory directly and measures how long does it take to build a call site with a generic constraint before and after this change.

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
```diff
@@ -545,6 +545,160 @@ public void OpenGenericServicesCanBeResolved()
             Assert.Same(singletonService, genericService.Value);
         }
 
+        [Fact]
```
pakrym: What happens if a singular service is resolved and the generic argument doesn't satisfy its constraint?

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
jbogard: So basically like the `CallSiteFactoryTest`'s? I can do that. What's the before/after look like to you? Literally executing the same benchmarks on before/after all my changes? All of the one with constraints will just throw exceptions, is that what you're looking for?

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
jbogard: I wasn't sure about adding that test since there's not an equivalent spec test against resolving a missing service. I can add one, though.

jbogard: The behavior is to return `null`, fyi.

pakrym: Good point, lets add it to `ServiceProviderContainerTests` instead.

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
jbogard: I'll change the NoDI one to be NoConstraints, too.

jbogard: And there's just one branch that goes into the constraints stuff, so I can just remove that branch and run the benchmarks for the "before" picture.

pakrym: > So basically like the CallSiteFactoryTest

Yes

There are two goals here:

1. Make sure we didn't regress perf for existing scenarios
2. Make sure we have reasonable perf for new ones

Running benchmarks with constraints that are satisfied before and after the change would show us if it made things slower in existing cases and we would have something to compare the performance of new scenarios too.

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
pakrym: Hm, am I correct in assuming that it would throw an exception before this change? How hard is it to keep throwing in singular resolution cases?

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
jbogard: Here's an example benchmark:

```c#
[GlobalSetup(Target = nameof(NoConstraints))]
public void SetupNoConstraints()
{
    _noConstraintsCsf = new CallSiteFactory(new[]
    {
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(A<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(B<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(C<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(D<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(E<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(F<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(G<>)),
        new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(H<>)),
    });
}

[Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
public void NoConstraints()
{
    for (int i = 0; i < OperationsPerInvoke; i++)
    {
        var _ = _noConstraintsCsf.GetCallSite(typeof(IEnumerable<IFakeGeneric<M>>), new CallSiteChain());
    }
}
```

This about right?

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
pakrym: It doesn't have to be part of the spec, failure cases generally aren't. I'm just trying to figure out what would be a better diagnostics experience for customers and if we should change it in this case.

pakrym: When you call `GetRequiredService` and container returns `null` user would get `No service for type '{0}' has been registered.` which would be quite misleading in case of not satisfied constraints.

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
pakrym: You can reuse the same `_noConstraintsCsf` field. And I think you don't need multiple implementation types for this case, registering the same type multiple times would do. Iteration time for this benchmark would be fairly substantial so you can drop `OperationsPerInvoke`

Another thing to consider is that `GetCallSite` caches result in a dictionary so only first iteration would measure what we want while all others would just measure dictionary retrieval. You would have to move `new CallSiteFactory` into benchmark itself to run code in question every iteration.

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
jbogard: Ah, I see. I had added a failure case, but I'll remove it. I honestly don't know what customers would expect - I generally don't try to resolve single constrained services. It's almost always part of an enumerable, for a chain of responsibility pattern or the like.

Personally, I'd prefer null, as an exception would mean I would need to add some nullo type just to allow a resolution.

jbogard: This might be a difficult one to signal, too, since you're really looking for the case that open generic service descriptors were found that matched the open generic type, but not the type parameter constraints. Partial matches don't really fit the API. It's either `null` or some implementation of `ServiceCallSite`. Even if the message is technically correct, I agree it's not too helpful in this scenario.

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
jbogard: Something like this then:

```c#
[Benchmark(OperationsPerInvoke = OperationsPerInvoke)]
public void NoConstraints()
{
    for (int i = 0; i < OperationsPerInvoke; i++)
    {
        var csf = new CallSiteFactory(new[]
        {
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
            new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>)),
        });
        var _ = csf.GetCallSite(typeof(IEnumerable<IFakeGeneric<M>>), new CallSiteChain());
    }
}
```

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
pakrym: It's not very difficult I think. Adding a flag to `TryCreateOpenGeneric` to say if it should skip or throw on non-satisfiable constraints and setting it to skip in TryCreateIenumerable.

jbogard: Ah I see it now!

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
pakrym: Remove OperationsPerInvoke and yeah let's put ServiceDescriptor array into a static

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
jbogard: I think I'm missing something here - `GetRequiredService` just calls `GetService`, then checks for `null`. How would `CallSiteFactory` know if it should throw or not in the case of a single service? As a consumer, I'd only want to see the specific exception on `GetRequiredService` but I don't see the connection between that method all the way down to `CallSiteFactory`, where the constraint check happens.

pakrym: Ah, I meant throw exception all the time both for `GetService` and `GetRequiredService` like we currently do. It's not super important scenario anyway so let's not change the behaviour.

jbogard: Sounds good! I kept the test in `CallSiteFactory` to confirm returning null, and removed the test from the specs.

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
jbogard: Before:

https://gist.github.com/jbogard/daa09b6a9741e34806986b516fc6f147

After:

https://gist.github.com/jbogard/355f85195a4008bdf61f61284d579325

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
pakrym: Sorry if I wasn't clear enough, I'd like us to continue throwing in singular resolution case as we did before.

jbogard: Just to make sure I have this correct:

- `GetService<INonexistentService>()` - returns `null`
- `GetRequiredService<INonexistentService>()` - throws
- `GetService<IFakeGenericService<INonmatchingConstraint>>()` - throws
- `GetRequiredService<IFakeGenericService<INonmatchingConstraint>>()` - throws

For the third case, it needs to check all potential matches before throwing. I can't just let the existing try-catch throw - that was the original problem. It needs to exhaust all options, and skip non-matching ones.

I don't want the existing exception to be thrown, either. It's `ArgumentException` from trying to close a generic type, which is different than "There were services registered, but none whose constraints matched".

This seems like an inconsistency between "regular" services returning null for `GetService` but that one throwing for no services matching the constraint.

In any case, if we throw, it looks like we'd want a more specific exception than what currently gets thrown by the BCL. And I'd only want the more specific exception thrown when concrete types *were* found, but *none* matched.

If no concrete types were found, I'd expect the existing behavior to stand - return null, then let `GetService`/`GetRequiredService` behavior determine exception throwing.

src/DependencyInjection/DI.Specification.Tests/src/Fakes/ClassInheritingAbstractClass.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.DependencyInjection.Specification.Fakes
+{
+    public class ClassInheritingAbstractClass : AbstractClass
```
pakrym: I don't see ClassAlsoInheritingAbstractClass  and ClassInheritingAbstractClass used anywhere.

src/DependencyInjection/DI.Specification.Tests/src/Fakes/ClassInheritingAbstractClass.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        
+    }
+
+    public class ClassAlsoInheritingAbstractClass : AbstractClass
+    {
+        
+    }
+
+    public class ClassInheritingClassInheritingAbstractClass : AbstractClass
```
pakrym: This class is not what is says it is.

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
```diff
@@ -0,0 +1,223 @@
...
+            var csf = new CallSiteFactory(_noConstraintSds);
+            var _ = csf.GetCallSite(typeof(IEnumerable<IFakeGeneric<M>>), new CallSiteChain());
+        }
+
+        [GlobalSetup(Target = nameof(NewConstraint))]
+        public void SetupNewConstraint()
+        {
+            _newConstraintSds = new[]
+            {
+                new ServiceDescriptor(typeof(IFakeGeneric<>), typeof(NoConstraintsOpenGenericService<>), ServiceLifetime.Transient),
```
pakrym: Make all of the implementation types be the same. Having only one of them being different downplays perf difference.

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
pakrym: This is how I currently see it:

# Single registration - single resolution

Registration:
```
IService<> -> Constrained<>
```

Resolution:
```
IService<Matching> - Success
IService<NonMatching> - Exception
```

# Multiple registrations - single resolution

Registration:
```
IService<> -> NonConstrained<>
IService<> -> Constrained<>
```

Resolution:
```
IService<Matching> - Success
IService<NonMatching> - Exception
```

# Single registration - multiple resolution

Registration:
```
IService<> -> Constrained<>
```

Resolution:
```
IEnumerable<IService<Matching>> - Success 1 results
IEnumerable<IService<NonMatching>> - Success 0 results
```

# Multiple registrations - multiple resolution

Registration:
```
IService<> -> NonConstrained<>
IService<> -> Constrained<>
```

Resolution:
```
IEnumerable<IService<Matching>> - Success 2 results
IEnumerable<IService<NonMatching>> - Success 1 result
```


Same rules apply when IService is dependency of other service being resolved.

jbogard: To summarize:

- Enumerables return empty
- Singular throw

For singular, do you want to expose the `ArgumentException` directly or bubble up something else, like `InvalidOperationException`? Keep the `ArgumentException` as the inner?

pakrym: Having a nice exception is better if it doesn't overcomplicate things too much.

jbogard: And I think this does make sense now - it's more of an edge case to have singular constrained service resolution, and then you get into weird precedence rules if you try to emulate how the compiler chooses the "right" one (like with generic methods and overloads)

src/DependencyInjection/DI.Specification.Tests/src/Fakes/ClassInheritingAbstractClass.cs
jbogard: Fixed.

src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs
jbogard: Yes, it does after running the benchmark after. Fixed, will update benchmarks after the exception change.

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
jbogard: Added an exception in the singular case, with its own localizable message.

jbogard:
Updated benchmarks.

### Before

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17134
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical cores and 4 physical cores
.NET Core SDK=2.1.500
  [Host]     : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT
  Job-SVFTTC : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 2.1  
RunStrategy=Throughput  

```
|                                Method |     Mean |     Error |   StdDev |   Median |     Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|-------------------------------------- |---------:|----------:|---------:|---------:|---------:|-------:|---------:|-------:|----------:|
|                         NoConstraints | 17.19 us | 1.0293 us | 2.903 us | 15.99 us | 58,180.4 |   1.00 |     0.00 | 0.0610 |    4840 B |
|                         NewConstraint | 16.12 us | 0.4414 us | 1.287 us | 16.02 us | 62,041.8 |   0.96 |     0.16 | 0.0610 |    4840 B |
|              NotMatchingNewConstraint |       NA |        NA |       NA |       NA |       NA |      ? |        ? |    N/A |       N/A |
|                  SimpleTypeConstraint | 16.42 us | 0.5701 us | 1.636 us | 15.99 us | 60,883.1 |   0.98 |     0.17 | 0.0610 |    4840 B |
|            ComplexInterfaceConstraint | 17.74 us | 1.1190 us | 3.211 us | 17.01 us | 56,366.2 |   1.06 |     0.24 | 0.0610 |    4840 B |
| NotMatchingComplexInterfaceConstraint |       NA |        NA |       NA |       NA |       NA |      ? |        ? |    N/A |       N/A |

Benchmarks with issues:
  ConstrainedGenericBenchmark.NotMatchingNewConstraint: Job-SVFTTC(Runtime=Core, Server=True, Toolchain=.NET Core 2.1, RunStrategy=Throughput)
  ConstrainedGenericBenchmark.NotMatchingComplexInterfaceConstraint: Job-SVFTTC(Runtime=Core, Server=True, Toolchain=.NET Core 2.1, RunStrategy=Throughput)


### After:

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17134
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical cores and 4 physical cores
.NET Core SDK=2.1.500
  [Host]     : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT
  Job-IKWAWI : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 2.1  
RunStrategy=Throughput  

```
|                                Method |      Mean |     Error |    StdDev |    Median |      Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|-------------------------------------- |----------:|----------:|----------:|----------:|----------:|-------:|---------:|-------:|----------:|
|                         NoConstraints | 20.247 us | 0.6795 us | 1.9388 us | 20.035 us |  49,389.1 |   1.00 |     0.00 | 0.0610 |   5.23 KB |
|                         NewConstraint | 22.327 us | 0.5545 us | 1.6087 us | 21.924 us |  44,788.0 |   1.11 |     0.13 | 0.0916 |   5.73 KB |
|              NotMatchingNewConstraint |  9.399 us | 0.2328 us | 0.6718 us |  9.154 us | 106,399.9 |   0.47 |     0.05 | 0.0458 |   2.81 KB |
|                  SimpleTypeConstraint | 22.062 us | 0.4401 us | 1.0544 us | 22.004 us |  45,327.1 |   1.10 |     0.11 | 0.0610 |   5.48 KB |
|            ComplexInterfaceConstraint | 36.113 us | 0.8208 us | 2.3812 us | 35.409 us |  27,690.8 |   1.80 |     0.20 | 0.1221 |   7.73 KB |
| NotMatchingComplexInterfaceConstraint | 62.102 us | 1.5588 us | 4.3711 us | 60.704 us |  16,102.4 |   3.09 |     0.35 | 0.1831 |  12.44 KB |
```

File: src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs (hunk 10 above)
Comment (jbogard) [reply src/DependencyInjection/DI/perf/ConstrainedGenericBenchmark.cs]: Fixed the benchmarks to have static fields

**Comment by pakrym** [comment]
We regressed `ComplexInterfaceConstraint` > 2x, can you take a look if there is anything that could be done there?

File: src/DependencyInjection/DI/test/ServiceLookup/CallSiteFactoryTest.cs [hunk 15]
```diff
@@ -121,9 +121,8 @@ public void CreateCallSite_ReturnsNull_IfClosedTypeDoesNotSatisfyStructGenericCo
             var callSiteFactory = GetCallSiteFactory(descriptor);
             // Act
             var nonMatchingType = typeof(IFakeOpenGenericService<object>);
-            var nonMatchingCallSite = callSiteFactory(nonMatchingType);
             // Assert
-            Assert.Null(nonMatchingCallSite);
+            Assert.Throws<InvalidOperationException>(() => callSiteFactory(nonMatchingType));
```
pakrym: Assert messages.

jbogard: Done.

jbogard:
I'll look into the `ComplexInterfaceConstraint` case.

jbogard:
Ha! Look what I found https://github.com/dotnet/BenchmarkDotNet/issues/387

jbogard:
I'm looking in to the internals to see if there are some easy wins, short-circuiting logic and the like. I've also opened an issue in CoreCLR:

https://github.com/dotnet/coreclr/issues/21302

To add a `TryMakeGenericType` method. I don't think it would be *too* terrible to add - there's a very explicit method to check constraints in the C++ layer.

jbogard:
Looking at the perf numbers, the slowdown primarily comes from constructing the generic type to validate:

image

Then:

image

Removing that from the hot path would mean reimplementing the rest of generic constraint checking (starting here https://github.com/dotnet/coreclr/blob/master/src/vm/typedesc.cpp#L2012).

I could go down that path, but I think it would be buggy and error prone since the data structures are a bit different than the public Type/TypeInfo API. I think the better choice would be to pursue `TryMakeGenericType` as an option so that the same checks do not get performed twice.

pakrym:
Sorry for a long response, got buried in issues for the last couple days.

I agree with you that we shouldn't reimplement entirety of generic constraint checking in DI. I think current solution is a good balance  of performance vs complexity untill we can get `TryMakeGenericType`.

jbogard:
Great! Any other feedback items?

pakrym:
Give me a day or two to go through PR again. Looks good for now.

pakrym:
cc @halter73 if you have time

seesharper:
@jbogard Is this in any way relevant to what you are doing here ?
https://github.com/aspnet/AspNetCore/issues/4531

jbogard:
@seesharper I'm not sure, I can add a test and find out. Do you have a minimal repro that doesn't take a dependency on LightInject?

I was intentionally leaving variance out of the issue as I want to address explicitly generic constraints. I hit variance way less often than generic constraints.

pakrym:
@seesharper commented on aspnet/AspNetCore#4531, seems like it's a side-effect of small stacks in InProcess mode.

seesharper:
@jbogard See my comment https://github.com/aspnet/AspNetCore/issues/4531#issuecomment-445482302 here. I've created a repro using LightInject with all the sources included. You should be able to simply use the same example with what you have here

pakrym:
@seesharper I don't think the failure we observed is related to a feature added here. This is about *excluding* services that are not matched by generic constraint instead of throwing. 

This change doesn't cause any additional services to be included in existing scenarios.

seesharper:
@pakrym @jbogard 
I agree, probably not related. I wasn't sure if this PR included support for variance and was just making sure that it did not cause trouble with AspNetCore 2.2 👍

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +361,129 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
             return parameterCallSites;
         }
 
+        private static bool IsCompatibleWithGenericParameterConstraints(Type genericTypeDefinition, Type[] parameters)
+        {
+            var genericArgumentDefinitions = genericTypeDefinition.GetTypeInfo().GenericTypeParameters;
+            for (var i = 0; i < genericArgumentDefinitions.Length; i++)
+            {
+                var argumentDefinitionTypeInfo = genericArgumentDefinitions[i].GetTypeInfo();
+                var parameter = parameters[i];
```
halter73: This turns a helpful ArgumentException:

```
System.ArgumentException : The number of generic arguments provided doesn't equal the arity of the generic type definition.
Parameter name: instantiation
Stack Trace:
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateOpenGeneric(ServiceDescriptor descriptor, Type serviceType, CallSiteChain callSiteChain, Int32 slot, Boolean throwOnConstraintViolation) in C:\dev\aspnet\Extensions\src\DependencyInjection\DI\src\ServiceLookup\CallSiteFactory.cs:line 216
...
```

into a relatively unhelpful IndexOutOfRangeException:

```
System.IndexOutOfRangeException : Index was outside the bounds of the array.
Stack Trace:
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.IsCompatibleWithGenericParameterConstraints(Type genericTypeDefinition, Type[] parameters) in C:\dev\aspnet\Extensions\src\DependencyInjection\DI\src\ServiceLookup\CallSiteFactory.cs:line 370
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateOpenGeneric(ServiceDescriptor descriptor, Type serviceType, CallSiteChain callSiteChain, Int32 slot, Boolean throwOnConstraintViolation) in C:\dev\aspnet\Extensions\src\DependencyInjection\DI\src\ServiceLookup\CallSiteFactory.cs:line 212
```

This change also made me realize that reordering generic parameters between the implementation type and service type is completely broken. The following implementation type could never implement `IFakeOpenGenericService2<,>` with our DI system:

```
public class FakeOpenGenericService2<T, U> : IFakeOpenGenericService2<U, T>
```

This change makes the situation slightly worse though because before you'd at least get a TypeLoadException if you tried to resolve `IEnumerable<IFakeOpenGenericService2<Foo, Bar>>` with `FakeOpenGenericService2<,>` registered as an `IFakeOpenGenericService2<,>`. After this change, FakeOpenGenericService2 would silently be omitted from the resolved IEnumerable.

pakrym: Agree with the first part. The second part is sort of the goal of this change.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +361,129 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+            if (genericArguments.Length > 0 && constraint.IsGenericType)
+            {
+                var typeDefinition = constraint.GetGenericTypeDefinition();
+                if (typeDefinition.GetTypeInfo().GenericTypeParameters.Length == genericArguments.Length)
+                {
+                    var constraintArguments = constraint.GetTypeInfo().GenericTypeArguments;
+                    for (var i = 0; i < constraintArguments.Length; i++)
+                    {
+                        var constraintArgument = constraintArguments[i];
+                        if (!constraintArgument.IsGenericParameter && !constraintArgument.IsAssignableFrom(genericArguments[i]))
```
halter73: Should this be reversed if the constraint is a contravariant?

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -358,6 +361,129 @@ private ServiceCallSite TryCreateOpenGeneric(ServiceDescriptor descriptor, Type
...
+            {
+                return true;
+            }
+
+            if (ParameterEqualsConstraint(parameter, constraint))
+            {
+                return true;
+            }
+
+            if (parameter.BaseType != null && ParameterEqualsConstraint(parameter.BaseType, constraint))
```
halter73: It seems odd to me that you would check the parent type, but not the grandparent type, great-grandparent type, etc...

seesharper: @jbogard @halter73 
Chiming in here as I have some experience with this from LightInject. Actually this was one of the things that was really hard to get right as it would practically mimic the behaviour of the compiler. I looked at several other implementations  and if I remember correctly, they all do a `MakeGenericType` in the end to make sure. 

For what its worth, here is what I've got for LightInject 

https://github.com/seesharper/LightInject/blob/70cfc68718a57bf02959bf7a09fe59f0d372cae5/src/LightInject/LightInject.cs#L6808

jbogard: @halter73 that was my first go (see https://github.com/aspnet/DependencyInjection/pull/472/files#diff-599c17314f9888a15039a1f175fb6c35) - to only rely on MakeGenericType and not try to reproduce any of the generic type loading arguments. Honestly, I'd still rather go that route, as it's the most correct option. However, it makes exceptions part of the hot path and becomes a pretty big perf hit for constraint misses.

jbogard: Though I'm a still bit unsure of the overall performance hit, since hit's only the first resolution that the exception hits for constraint misses, and from then on out the resolution plan is cached.

jbogard: And to me the long-term solution is to remove all of this in favor of my API proposal https://github.com/dotnet/corefx/issues/33769 - `TryMakeGenericType`, since I've just tried to reproduce the easy cases with this approach. These checks, if successful, still get retried with a call to `MakeGenericType`.

pakrym: I agree with @halter73. Taking perf hit for an uncommon case to make sure things stay correct is acceptable.

jbogard: Alright, I'll push the simple, but comprehensive, but perf hit case. I prefer that as well - it will make things easy when (if?) `TryMakeGenericType` happens.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
```diff
@@ -208,14 +205,26 @@ private ServiceCallSite TryCreateExact(ServiceDescriptor descriptor, Type servic
...
+                    closedType = descriptor.ImplementationType.MakeGenericType(serviceType.GenericTypeArguments);
+                }
+                catch (ArgumentException ex)
+                {
+                    if (throwOnConstraintViolation)
+                    {
+                        throw new InvalidOperationException(Resources.FormatGenericConstraintViolation(serviceType, descriptor.ImplementationType), ex);
+                    }
+
+                    return null;
```
halter73: I think this could end up swallowing exceptions that aren't constraint violations but instead just bad registrations. Mismatched arity between the service and implementation type would be an example of this.

Maybe we should make this feature opt-in. What do you think @pakrym?

jbogard: The other way I look at this is - if the type can't be closed today, it's automatic exception and there is zero mitigation as a consumer other than switching containers, so my overall goal was just to stop blowing up if a type couldn't close. It results in edge cases where it can't be closed for other reasons, but that's all I have, an exception type that covers a main failure case and a few much less frequent ones.

pakrym:
I ran this PR against tests in https://github.com/aspnet/Extensions/pull/629 and it fails a bunch of tests in Grace, Autofac, Stashbox and StructureMap

jbogard:
It's the spec tests I added themselves - even reverting my change to `CallSiteFactory` led to the same test failures with my change in. I need to tweak my tests so that they're able to be run against those containers.

jbogard:
@pakrym well that made things quite a bit better. It looks like the spec project you added highlights some edge case bugs in a couple of containers:

Autofac
 - PublicNoArgCtorConstrainedOpenGenericServicesCanBeResolved

Stashbox
 - PublicNoArgCtorConstrainedOpenGenericServicesCanBeResolved
 - SelfReferencingConstrainedOpenGenericServicesCanBeResolved
 - ClassConstrainedOpenGenericServicesCanBeResolved

I checked my library and I'm able to get all the features passing with those checks that I did.

jbogard:
Should I just skip those then? I can open issues with those respective projects.

ENikS:
I don’t think you should skip these by default.
Perhaps file an issue with each of these containers and give them a chance to fix it first?

jbogard:
Created an issue for Autofac's DI extensions:

https://github.com/autofac/Autofac.Extensions.DependencyInjection/issues/40

jbogard:
Created an issue for Stashbox's DI extensions:

https://github.com/z4kn4fein/stashbox-extensions-dependencyinjection/issues/3

jbogard:
Autofac seems to support that behavior of the failed test in its base DI container, but the test fails when I add the extensions. Stashbox is a little more complex.

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
```diff
@@ -545,6 +545,160 @@ public void OpenGenericServicesCanBeResolved()
...
+            // Assert
+            Assert.Equal(2, allServices.Count);
+            Assert.Same(enumerableVal, allServices[0].Value);
+            Assert.Same(enumerableVal, allServices[1].Value);
+            Assert.Equal(1, constrainedServices.Count);
+            Assert.Same(singletonService, constrainedServices[0].Value);
+        }
+
+        [Fact]
+        public void PublicNoArgCtorConstrainedOpenGenericServicesCanBeResolved()
```
tillig: I think this test has an invalid premise, so I'm hoping to get clarification on the intent. I think it may have an impact on how the spec for constraints should work in a single instance case.

Here's how I'm reading it:

- Register an open generic with no constraints. This should work for any `T` in the generic.
- Register an open generic with a constraint of `new()`. This will only work for things with a public constructor.
- Resolve all the services using a class that can be constructed. You should get two.
- Resolve all the services using a class that doesn't satisfy the constraint. The constraint violation will be ignored but the others will be resolved.

A challenge I'm having is that this impacts the way a single instance resolution would work.

```csharp
var collection = new TestServiceCollection();
collection.AddTransient(typeof(IFakeOpenGenericService<>), typeof(ClassWithNoConstraints<>));
collection.AddTransient(typeof(IFakeOpenGenericService<>), typeof(ClassWithNewConstraint<>));
var provider = CreateServiceProvider(collection);
// What do you get here? Exception? Fallback to ClassWithNoConstraint<T>? null?
var singleService = provider.GetService<IFakeOpenGenericService<ClassWithPrivateCtor>>();
```

From autofac/Autofac.Extensions.DependencyInjection#40 I gather the intent is...

- In the single instance, you'd get an exception. The last thing you registered - the override - doesn't satisfy the constraint, so resolution fails.
- In the collection, you get a sort of "filtered" collection. Only things that satisfy the constraint are resolved.

From a use case perspective, let's say I have a bunch of message handlers. I register them all and resolve the collection of them. Sometimes I get them all and sometimes I don't. How do I know why? How do I troubleshoot that? 

I'm not sure I agree with the filtered collection concept. It may pass on the base container, but I'm not sure it should. I'd expect any constraint violation to notify me. Perhaps I'm alone in that?

pakrym:
Considering that some containers do not support this feature and it's not required for ASP.NET Core framework to function I'm thinking about not adding it to the spec and keeping it only in MEDI.

Let me think about the way to do it correctly without ASP.NET Core accidentally taking a dependency on the new features.

seesharper:
@pakrym random idea. Run all the relevant aspnetcore tests with each adapter.?

jbogard:
@pakrym that makes sense - really the only important behavior are constraints for types, not the `new()`, `class`, `struct` ones. Having type-based constraints is effectively giving us the ability to do pattern matching via dependency injection.

Would you be OK moving the `GenericParameterAttributes`-based tests out of the spec, but keeping the type-based ones? That's by far the most important behavior.

ENikS:
I would second @tillig point. Pattern matching should be done in post processing using LINQ or something similar.

pakrym:
@jbogard I don't think anything should go into spec tests as it not required for ASP.NET Core itself to work (which is the point of spec tests). 

I somewhat agree with @tillig's  concern about diagnosability. It is very unfortunate that this feature targets all of the `IEnumerable<>` resolutions while only 1-2 of them expect this behaviour while others just get worse diagnostics.  

I also wonder if this matching logic should be ofloaded to DI at all and if there are more scenarios that this solves that IMessageHandler<TMessage>

jbogard:
@ENikS this feature is not possible with pattern matching with LINQ. If this was F# that has pattern matching, partial application, and currying, I wouldn't need this feature.

But this is OO-land, so I have to work within dependency injection and types.

Having used this feature for maybe 10 years or more in other containers, it's a fairly natural thing to do. I have some infrastructure code that does something "give me all possible implementations of IFoo<T>" where have many different plugged versions of T.

Ultimately, there is no such thing as the "whole collection" when dealing with open generics. You have to close the interface type first, then determine what services match the closed type. Only when you have the closed type can you determine what the "whole collection" is.

In practice, what this looks like is I have many different concrete types that close the service, perhaps in the hundreds. If I want to do cross-cutting concerns for *some* of the generic parameter types, I have those types implement some interface, and the cross-cutting concern then uses that interface as a generic constraint.

A concrete example:

```
public interface IValidator<T> {
    bool IsValid(T instance);
}

public class FooValidator : IValidator<Foo> { ... }
public class BarValidator : IValidator<Bar> { ... }
public class RequiresApprovalValidator<T> : IValidator<T>
    where T : IRequiresApproval { ... }
```

When we go to perform an operation against a known type T, I'll ask the container, "find all validators that match type T". This could be zero or many.

So this is not about a singular collection of dependencies, because the resolved list can contain closed implementations, or open implementations, or both. Specifically in the validation scenario, we do this *all the time*.

The way around it today is to define closed type:

```c#
public class RequiresApprovalValidator : IValidator<IRequiresApproval> { ... }
```

However, since we've lost the type parameter, our concrete type can't take any dependencies that vary on T (say, `IRepository<T>` or `ILogger<T>`).

With our applications, we reserve concretions for the concrete type, and open generics for cross-cutting types.

tillig:
I guess the way I've seen this solved is that the `IValidator<T>` would have a `bool CanValidate(T obj)` where it would do the check rather than having it bound to a collection via DI.

```csharp
provider.GetServices<IValidator<Foo>>().Where(v => v.CanValidate(something));
```

Alternatively, it wouldn't be a generic type constraint.

```csharp
public class RequiresApprovalValidator<T> : IValidator<T>
{
  public void Validate(T obj)
  {
    if(!(T is IRequiresApproval))
    {
      return;
    }
  }
}
```

The validator only works on some things, not others, so it's not really a generic validator anymore - It might implement the interface, but it's not treated the same in practice.

Doing the pattern matching in the collection resolution feels like an overloaded usage of the DI framework. Like, I sometimes see folks using Autofac that love to attach non-DI-related application startup logic to the "container build" event - it's convenient, but if the app startup logic causes the container to fail to build or something, that's because you're trying to get the container to act outside its realm of responsibility.

I don't mind if folks want to use containers that do that pattern matching, but it shouldn't be in the spec, nor should ASP.NET Core add dependencies on that type of behavior. As it stands, the assumption on collection ordering got added and isn't so great; it'd be better if the spec is as loose as possible to allow the most compatibility.

pakrym:
@jbogard ASP.NET Core DI spec is not intended to be the lowest common denominator of all DI container features. It only contains a feature set required to run ASP.NET Core using it. The reason is that if a new container decides it implement the spec it shouldn't be too large or too restrictive.

We also never did a feature in MEDI that wasn't part of the spec (except validation/diagnostics) so this one is a new territory for us and might require more design in general.

As for "all the implementing containers support this feature, so let's add it to the core container", I don't think it's true either, MEDI doesn't have a lot of more basic and popular features of other .NET DI containers (Property injection/Func<T> etc.) and it's not a goal to get all of them. So I would still look a this from "is it a valuable feature to have" perspective.

jbogard:
@pakrym in terms of should the matching logic apply - no, not all, so some of my extensions to MS.Ext.DI do the scanning, matching, and registration myself. I find all the concretions and interfaces and provide explicit registration (see https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection/blob/master/src/MediatR.Extensions.Microsoft.DependencyInjection/ServiceCollectionExtensions.cs#L113 as an example).

It's only the open generic case I'm interested in fixing - I don't want to find all the potentially closed types of an open generic and try to close each and register.

This would be these tests:

 - `ConstrainedOpenGenericServicesCanBeResolved`
 - `ConstrainedOpenGenericServicesReturnsEmptyWithNoMatches`
 - `InterfaceConstrainedOpenGenericServicesCanBeResolved`
 - `AbstractClassConstrainedOpenGenericServicesCanBeResolved`

I'd remove:

 - `PublicNoArgCtorConstrainedOpenGenericServicesCanBeResolved`
 - `ClassConstrainedOpenGenericServicesCanBeResolved`
 - `StructConstrainedOpenGenericServicesCanBeResolved`
 - `SelfReferencingConstrainedOpenGenericServicesCanBeResolved`

That last one because it's a weird edge case that I don't think comes up in practice, and not all containers support it anyway.

@tillig I could do that, but that's what a constraint does. I'm using constraints to constrain results, which seems in line with the definition. I don't have to compromise on my design, adding methods and matches and LINQ etc.

jbogard:
Or we can remove all the spec tests - I thought it was using it as a means to validate that containers support this feature, not that ASP.NET Core can use it safely. I don't see ASP.NET Core using this (although something like Orleans, NServiceBus, and other message-oriented frameworks would).

jbogard:
So it seems like our options here are:

1. Don't add the feature. If you want this feature, you must use a 3rd party container
2. Add the feature, but don't add it to the spec
3. Add the feature, only add the minimal intended behavior to the spec
4. Add the feature, add all the possible behaviors to the spec

The argument for 1 is "our direct intended consumer doesn't need/use this feature". That makes sense, but I do think MEDI will grow in its usage outside of ASP.NET Core, such as Orleans already adopting it. I understand though its goal is not to replace the P&P container or CommonServiceLocator.

If a container supports this behavior today, but doesn't agree they should/do (like @tillig), then that would seem it would lean to option 2.

The argument for 3 and 4 is much weaker, since ASP.NET Core doesn't, and I don't think ever will, use this feature. It's much more likely message-oriented libraries will, like my own, that center so much around open generics and message types.

pakrym:
@jbogard, to give you some update, I'm looking into design to enable \# 2 (started a spec https://gist.github.com/pakrym/3a026eb03386467050da1ddadda7d4e5)

src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
jbogard: Don't think of it as "resolving a collection". I'm not registering a static collection. I'm registering a bunch of individual handlers, then I come back later and ask "find all the handlers that match this specification".

jbogard:
Ping @pakrym :)

pakrym:
We are not able to take this PR until we can figure out a way to add features to MEDI without requiring in the specification. I have some ideas but unfortunately, it doesn't look like we would have resources to implement them in 3.0.

KeithBarrows:
I am very interested in the outcome of this.  I just spent 3 days trying to get a work around to work when several code paths end up using the same definitions.  Had to resort to some trickery to get the desired behavior.

AlonAm:
This is an essential feature. Having two DI containers in the same app to solve this problem is not a good option.

vyacheslav-mikhaylov:
It would be great to have this feature available in 3.0

ffMathy:
@jbogard friendly ping: Looks like this has conflicted.

z4kn4fein:
Hi, I just wanted to leave a note here, that I fixed those ignored test cases in the latest version of Stashbox, so they should pass now.

jbogard:
@ffMathy Fixed!

muratg:
cc @anurse

halter73:
> This is an essential feature. Having two DI containers in the same app to solve this problem is not a good option.

We wouldn't recommend two DI containers. We would recommend replacing the MEDI container with a container that supports constrained open generics.

It would be nice to have this feature in MEDI so that customers who want the feature can use it in their own apps without needing to switch from the default container. The risk is that 3rd-party libraries and ASP.NET Core itself might start relying on the behavior and break customers using non-MEDI containers that don't support constrained open generics. If we came up with a good way of ensuring that doesn't happen in an automated way, I'd feel more comfortable about this change.

jbogard:
Isn’t the DI Council the proposed solution for new features (if you can call adding a try-catch a feature)? The spec tests automate, sure, but maintaining those tests are fairly tedious as I’ve found in MediatR for documenting container features.

glucaci:
Any news on this, or at least a Milestone?

JoasE:
@KeithBarrows Care to share your workaround?

jbogard:
Any thoughts on this for the next release?

jbogard:
Sounds like the approach of try...catch will be the one per https://github.com/dotnet/runtime/issues/28033#issuecomment-595380158

src/DependencyInjection/DI/src/ServiceLookup/CallSiteFactory.cs
davidfowl: That seems like an argument for making it optional

davidfowl: That seems like an argument for making it optional

PR #538, opened 2018-11-18
Refactor out GetChildrenImplementation into extension method on IConf…

Euphoric:
Moving my PR from previous repo ( https://github.com/aspnet/Configuration/pull/899 )

When creating a decorator around `IConfigurationRoot`, I found out that I'm unable to use `ConfigurationSection` as it requires specific implementation of `ConfigurationRoot` instead of abstraction `IConfigurationRoot`.
To make this change possible, it was necessary to change internal method `ConfigurationRoot.GetChildrenImplementation` into static method on `IConfigurationRoot` instead of instance method. I also went ahead and turned it into extension method.

This change doesn't require any new tests, as it is refactoring of internal structure.

I will leave it to maintainers to decide if this is breaking change or not.
I don't see this a breaking change, as it goes from specific class to it's interface. But there might be some weirdness, that I might not know off.

### DIFF
--- a/src/Configuration/Config/src/ConfigurationRoot.cs
+++ b/src/Configuration/Config/src/ConfigurationRoot.cs
@@ -81,16 +81,7 @@ namespace Microsoft.Extensions.Configuration
         /// Gets the immediate children sub-sections.
         /// </summary>
         /// <returns></returns>
-        public IEnumerable<IConfigurationSection> GetChildren() => GetChildrenImplementation(null);
-
-        internal IEnumerable<IConfigurationSection> GetChildrenImplementation(string path)
-        {
-            return _providers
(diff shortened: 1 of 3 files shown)


Euphoric:
@poke As I explained above. I tried to make a decorator around `IConfigurationRoot`, but I was unable to use existing `ConfigurationSection`, as it is tightly coupled with `ConfigurationRoot`. It can be worked-around by making local implementation of `ConfigurationSection `(like https://github.com/Euphoric/Standard.Configuration.Alias/blob/master/Standard.Configuration.Alias/ConfigurationSection.cs), but I find that unelegant, as it is basically copy/paste of original.

ajcvickers:
Closing in favor of https://github.com/aspnet/Extensions/pull/605, which still contains your commit.

PR #546, opened 2018-11-20
Add event for expression tree generation and chunk call sites

pakrym:
1. Add event that logs size of expression trees that are generated
2. Modify call site built event to avoid printing duplicated nodes - it would output  `{ "ref": "TypeName" }` instead
3. Modify call site built event to chunk output in 10k chunks.

### DIFF
--- a/src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
+++ b/src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics.Tracing;
+using System.Linq.Expressions;
 using System.Runtime.CompilerServices;
 using Microsoft.Extensions.DependencyInjection.ServiceLookup;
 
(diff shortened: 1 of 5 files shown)


pakrym:
Also, remove JSON formatting, it's not human readable when chunked anyway.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
```diff
@@ -13,7 +14,9 @@ internal sealed class DependencyInjectionEventSource : EventSource
     {
         public static readonly DependencyInjectionEventSource Log = new DependencyInjectionEventSource();
 
-        private DependencyInjectionEventSource()
+        private int MaxChunkSize = 10 * 1024;
```
davidfowl: Add a comment about this

src/DependencyInjection/DI/test/Microsoft.Extensions.DependencyInjection.Tests.csproj
```diff
@@ -6,6 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Reference Include="Newtonsoft.Json" />
```
davidfowl: ?

pakrym: I use it in tests to make sure JSON is valid and reformat it before asserting.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
pakrym: Done

PR #552, opened 2018-11-21
Make AddHostedService register singletons

Tratcher:
https://github.com/aspnet/Extensions/issues/553

For the host it doesn't matter if IHostedServices are transient or singletons, they're only resolved once. However we've had several reports of users that also want to inject their hosted service into other components in order to interact with it.

Also adding an sln.

### DIFF
--- a/src/Hosting/Abstractions/src/ServiceCollectionHostedServiceExtensions.cs
+++ b/src/Hosting/Abstractions/src/ServiceCollectionHostedServiceExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.Extensions.Hosting;
@@ -15,6 +15,6 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>The original <see cref="IServiceCollection"/>.</returns>
         public static IServiceCollection AddHostedService<THostedService>(this IServiceCollection services)
             where THostedService : class, IHostedService
-            => services.AddTransient<IHostedService, THostedService>();
+            => services.AddSingleton<IHostedService, THostedService>();
     }
 }
--- a/src/Hosting/Hosting/test/HostTests.cs
+++ b/src/Hosting/Hosting/test/HostTests.cs
@@ -160,6 +160,27 @@ namespace Microsoft.Extensions.Hosting
         }
 
         [Fact]
+        public async Task HostedServiceRegisteredAsSingletons()
+        {
+            using (var host = CreateBuilder()
+                   .ConfigureServices((hostContext, services) =>
+                   {
+                       services.AddHostedService<TestHostedService>();
+                   })
+                   .Build())
+            {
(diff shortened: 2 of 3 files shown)


dotnetchris:
Yes, hosted services should be viewed as 'servers' and a very common need of a server is to be able to expose status information about its health and state. With these currently registered as transient, it's impossible to access this information without using global variables to marshal state.

src/Hosting/Abstractions/src/ServiceCollectionHostedServiceExtensions.cs
```diff
@@ -15,6 +15,6 @@ public static class ServiceCollectionHostedServiceExtensions
         /// <returns>The original <see cref="IServiceCollection"/>.</returns>
         public static IServiceCollection AddHostedService<THostedService>(this IServiceCollection services)
             where THostedService : class, IHostedService
-            => services.AddTransient<IHostedService, THostedService>();
+            => services.AddSingleton<IHostedService, THostedService>();
```
pakrym: Add a test.

davidfowl:
@dotnerchris you mean you want to use the hosted service as a global variable instead of another object....

Tratcher:
@sguryev right. If it were Transient then the host and app would get difference instances and StartCalled would be false. Scoped wouldn't resolve at all, there is no scope established at this point.

dotnetchris:
> @dotnetchris you mean you want to use the hosted service as a global variable instead of another object....

If you're making the pedantic argument that singleton and global are synonymous, sure.

davidfowl:
> If you're making the pedantic argument that singleton and global are synonymous, sure.

No, I'm saying that you just want to use 1 global variable instead of 2. That's what this change enables.

dotnetchris:
> > If you're making the pedantic argument that singleton and global are synonymous, sure.
> 
> No, I'm saying that you just want to use 1 global variable instead of 2. That's what this change enables.

In well designed adherence to encapsulation / open-closed principle. As opposed to side effecting global state in violation of _every_ design principle.

Tratcher:
@billpoole-mi no change for IDisposable services. Note the host is single use, you can't call Start and Stop multiple times. All disposable services are disposed when you dispose the host, not when you call Stop.

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -159,6 +159,27 @@ public async Task HostedServiceStartNotCalledIfHostNotStarted()
...
+        public async Task HostedServiceRegisteredAsSingletons()
+        {
+            using (var host = CreateBuilder()
+                   .ConfigureServices((hostContext, services) =>
+                   {
+                       services.AddHostedService<TestHostedService>();
+                   })
+                   .Build())
+            {
+                var svc = (TestHostedService)host.Services.GetRequiredService<IHostedService>();
```
AgentFire, line 172: Call `GetService` at elast twice to prove the instance is indeed the same?

dotnetchris: +1 and ensure reference equality of the 2nd result

davidfowl: Send a pull request?

sguryev: I have asked about it here: https://github.com/dotnet/extensions/pull/552#commitcomment-31388021
Answers are below.

PR #556, opened 2018-11-21
Replace Console.WriteLine with Logging in generic host

Tratcher:
https://github.com/aspnet/Hosting/issues/1584
@glennc

### DIFF
--- a/src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
+++ b/src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
@@ -1,9 +1,11 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Hosting.Internal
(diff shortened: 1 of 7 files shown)


src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
```diff
@@ -28,15 +34,17 @@ public ConsoleLifetime(IOptions<ConsoleLifetimeOptions> options, IHostingEnviron
 
         private IApplicationLifetime ApplicationLifetime { get; }
 
+        private  ILogger Logger { get; }
```
pakrym: nit space

src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
```diff
@@ -16,10 +18,14 @@ public class ConsoleLifetime : IHostLifetime, IDisposable
...
 
         public ConsoleLifetime(IOptions<ConsoleLifetimeOptions> options, IHostingEnvironment environment, IApplicationLifetime applicationLifetime)
+            : this(options, environment, applicationLifetime, NullLoggerFactory.Instance) { }
+
+        public ConsoleLifetime(IOptions<ConsoleLifetimeOptions> options, IHostingEnvironment environment, IApplicationLifetime applicationLifetime, ILoggerFactory loggerFactory)
         {
             Options = options?.Value ?? throw new ArgumentNullException(nameof(options));
             Environment = environment ?? throw new ArgumentNullException(nameof(environment));
             ApplicationLifetime = applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime));
+            Logger = loggerFactory.CreateLogger("Microsoft.Extensions.Hosting.Host");
```
pakrym: `CreateLogger<Host>`?

Tratcher: Microsoft.Extensions.Hosting.Host doesn't actually exist, it's Microsoft.Extensions.Hosting.Internal.Host. I didn't want Internal in the log scope.

pakrym: What about https://github.com/aspnet/Extensions/blob/master/src/Hosting/Hosting/src/Internal/Host.cs#L23 then?

Tratcher: Yeah, we should re-think that. We've changed that kind of thing elsewhere. @glennc

Tratcher: Offline note: making the pubternal types properly internal and flattening the namespace would indirectly deal with this. We've taken that approach in other projects like MVC.

PR #567, opened 2018-11-27
[WIP] Initial helix target which runs tests on windows queue

HaoK:
Initial PR for discussion, now that things mostly work on windows (88/92) passing:
UserSecrets has one test that can't find soln root and might just need to be reworked
Logging.Analyzer.Tests is very busted and needs further investigation why

`build.cmd /t:Helix /p:HelixAccessToken=<PAT>` 

Example run results: https://mc.dot.net/#/user/HaoK/pr~2Faspnet~2Fextensions/ci/20181127.124408

Overview:
- Single helix job
- One work item per test project + framework permutation (i.e. Primitives.Test.csproj / net461)
- different runtest.cmd for net461 (uses xunit-console in payload) / netcoreapp (downloads sdk for dotnet vstest)

TODOs:
- Only targets windows queue (need to add logic to the test command to run on mac/linux)
- Hardcodes sdk version to 2.1.500 for the helix jobs (which appears to be already installed for some of the work items, machines getting reused?)
- Fix reporting (unit test summary got lost recently)


Directory.Build.targets
```diff
@@ -44,4 +44,62 @@
...
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(PublishAbsoluteDir)" />
+
+    <ItemGroup>
+      <TestPayload Include="$(PublishAbsoluteDir.TrimEnd('\').TrimEnd('\'))">
+        <TestAssembly>$(TargetFileName)</TestAssembly>
+        <TestName>$(MSBuildProjectName)/$(TargetFramework)</TestName>
+        <!-- <Command>runtests.cmd $(TargetFileName) $(_Microsoft_NETCore_App_Version)</Command> -->
+        <Command>runtests.cmd $(TargetFileName) 2.1.500</Command>
```
natemcmaster: I think you can use `$(NETCoreSdkVersion)` to get the .NET Core SDK version.

Directory.Build.targets
```diff
@@ -44,4 +44,62 @@
...
+    <Copy SourceFiles="@(XunitConsoleRunnerFiles)" DestinationFolder="$(PublishDir)" />
+  </Target>
+
+  <Target Name="_CreateTestPayloadInner" DependsOnTargets="Publish;CollectXunitConsoleRunner" Condition="'$(IsTestProject)' == 'true'" Returns="@(TestPayload)">
+
+    <ConvertToAbsolutePath Paths="$(PublishDir)">
+      <Output TaskParameter="AbsolutePaths" PropertyName="PublishAbsoluteDir" />
+    </ConvertToAbsolutePath>
+
+    <ItemGroup>
```
natemcmaster:
```suggestion
    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'" >
```
You can condition ItemGroups too. It makes it a little more read-able when there are several items with the same, or similar, condition

test/helix.proj
```diff
@@ -0,0 +1,46 @@
...
+    <MSBuild Projects="@(ProjectsToTest)"
+              Targets="CreateTestPayload"
+              BuildInParallel="true">
+      <Output TaskParameter="TargetOutputs" ItemName="TestPayload" />
+    </MSBuild>
+    <ItemGroup>
+      <HelixWorkItem Include="%(TestPayload.TestName)">
+        <Command>%(TestPayload.Command)</Command>
+        <PayloadDirectory>%(TestPayload.Identity)</PayloadDirectory>
+        <Timeout>00:30:00</Timeout>
```
natemcmaster:
```suggestion
        <Timeout>%(TestPayload.TestTimeout)</Timeout>
```

Directory.Build.targets
```diff
@@ -44,4 +44,62 @@
...
+
+    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(PublishAbsoluteDir)" />
+
+    <ItemGroup>
+      <TestPayload Include="$(PublishAbsoluteDir.TrimEnd('\').TrimEnd('\'))">
+        <TestAssembly>$(TargetFileName)</TestAssembly>
+        <TestName>$(MSBuildProjectName)/$(TargetFramework)</TestName>
+        <!-- <Command>runtests.cmd $(TargetFileName) $(_Microsoft_NETCore_App_Version)</Command> -->
+        <Command>runtests.cmd $(TargetFileName) 2.1.500</Command>
+        <TestPlatform>%(DistributedTestPlatform.Identity)</TestPlatform>
```
natemcmaster:
```suggestion
        <TestPlatform>%(DistributedTestPlatform.Identity)</TestPlatform>
        <TestTimeout>$(DistributedTestTimeout)</TestTimeout>
```

Also, in Directory.Build.props, add a property `<DistributedTestTimeout>00:30:00</DistributedTestTimeout>`

test/helix.proj
```diff
@@ -0,0 +1,46 @@
...
+      </HelixWorkItem>
+    </ItemGroup>
+  </Target>
+
+  <PropertyGroup>
+    <HelixSource>pr/aspnet/extensions</HelixSource>
+    <HelixType>ci</HelixType>
+    <!-- TODO: use package version -->
+    <HelixBuild>$([System.DateTime]::Now.ToString(yyyyMMdd-hhmmss))</HelixBuild>
```
natemcmaster:
```suggestion
    <HelixBuild Condition=" '$(CI)' == 'true' ">$(BUILD_BUILDNUMBER)</HelixBuild>
    <HelixBuild Condition=" '$(CI)' != 'true' ">$([System.DateTime]::Now.ToString(yyyyMMdd-hhmmss))</HelixBuild>
```

test/vstest/runtests.cmd
```diff
@@ -0,0 +1,7 @@
+set target=%1
+set version=%2
+powershell.exe -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "./dotnet-install.ps1 -Version %version% -Channel 2.0 -NoCdn -InstallDir %HELIX_CORRELATION_PAYLOAD%"
```
natemcmaster:
```suggestion
powershell.exe -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "Invoke-WebRequest https://get.dot.net/v1/dotnet-install.ps1 -o ./dotnet-install.ps1"
powershell.exe -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "./dotnet-install.ps1 -Version %version% -Channel 2.0 -NoCdn -InstallDir %HELIX_CORRELATION_PAYLOAD%"
```
An alternative to having a copy of dotnet-install.ps1 in the payload (although its size is negligible).

natemcmaster:
This is a great start! I'm glad we're getting closer to some thing that works.

My primary concern so far is about test time. You mentioned that this currently takes about 20 minutes just to upload test payloads. Extension current gets a complete build+test pass in 9 minutes by running tests on the CI agent. I'm not sure Helix is going to make sense for us if it takes longer to upload test projects than it does to run locally.

test/vstest/runtests.cmd
```diff
@@ -0,0 +1,7 @@
...
+set PATH=%HELIX_CORRELATION_PAYLOAD%;%PATH%
```
natemcmaster:
```suggestion
set PATH=%HELIX_CORRELATION_PAYLOAD%;%PATH%
set DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
```
This should save 30-60 seconds per test run.

HaoK:
Updated
- single payload per target framework: artifacts/Helix/[net461 | netcoreapp20 | netcoreapp21]
- added mac/redhat/unbutu queues

Run results: https://mc.dot.net/#/user/HaoK/pr~2Faspnet~2Fextensions/ci/haok.11.28.17

Configuration test failures, but pretty good starting point for passing tests

HaoK:
11:33 time to send the jobs (60s for windows job (extra net461), then ~40s for the other 3)

Directory.Build.targets
HaoK: Actually now I remember why I hardcoded this, the netcoreapp2.1 tests fail with the build sdk looking for 2.1.6:

```
The specified framework 'Microsoft.NETCore.App', version '2.1.6' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\dotnetbuild\work\0374a9cf-3982-43d1-9cc8-33ea1a569f82\Payload\sdk\
  - Installing .NET Core prerequisites might help resolve this problem:
      http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
  - The .NET Core framework and SDK can be installed from:
      https://aka.ms/dotnet-download
  - The following versions are installed:
      2.1.5 at [C:\dotnetbuild\work\0374a9cf-3982-43d1-9cc8-33ea1a569f82\Payload\sdk\shared\Microsoft.NETCore.App]
```

natemcmaster: Oh. Try updating KoreBuild (run `build.cmd -u /t:Cow`). I updated it to 2.1.500 a couple weeks ago: https://github.com/aspnet/BuildTools/pull/875

natemcmaster:
> send the jobs on wired connex (60s for windows job (extra net461), then ~40s for the other 3)

Wow, that's a huge improvement over 20 minutes to upload. Nice work :)

Can we make this PR run Helix instead of running tests locally?

test/vstest/runtests.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --version $2 --channel 2.0 --no-cdn --install-dir $HELIX_CORRELATION_PAYLOAD/sdk
+$HELIX_CORRELATION_PAYLOAD/sdk/dotnet vstest $1 --logger:trx
```
natemcmaster:
```suggestion
export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
$HELIX_CORRELATION_PAYLOAD/sdk/dotnet vstest $1 --logger:trx
```

test/vstest/runtests.cmd
```diff
@@ -0,0 +1,7 @@
+set target=%1
+set version=%2
```
natemcmaster: I recommend we add another variable here since we frequently consume new runtimes before SDKs.

```suggestion
set sdk_version=%2
set runtime_version=%3
```

Then, add a second call to dotnet-install.ps1 to install the runtime:

    ....   dotnet-install.ps1 -Runtime dotnet -Version %runtime_version% ....

test/vstest/runtests.cmd
```diff
@@ -0,0 +1,7 @@
...
+powershell.exe -NoProfile -ExecutionPolicy unrestricted -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; &([scriptblock]::Create((Invoke-WebRequest -useb 'https://dot.net/v1/dotnet-install.ps1'))) -Version %version% -Channel 2.0 -NoCdn -InstallDir %HELIX_CORRELATION_PAYLOAD%\sdk"
```
natemcmaster:
```suggestion
powershell.exe -NoProfile -ExecutionPolicy unrestricted -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; &([scriptblock]::Create((Invoke-WebRequest -useb 'https://dot.net/v1/dotnet-install.ps1'))) -Version %version% -InstallDir %HELIX_CORRELATION_PAYLOAD%\sdk"
```

test/vstest/runtests.cmd
```diff
@@ -0,0 +1,7 @@
+set target=%1
```
natemcmaster: Starting to think we should have `runtests.cmd` and `runtests.ps1`, and make the cmd a dummy script that starts powershell.

test/vstest/runtests.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --version $2 --channel 2.0 --no-cdn --install-dir $HELIX_CORRELATION_PAYLOAD/sdk
+export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
```
natemcmaster: Just thought of a few more environment variables we should probably set in Linux **and** Windows.
```sh
# Ensures every invocation of dotnet apps uses the same dotnet.exe
export DOTNET_ROOT="$HELIX_CORRELATION_PAYLOAD/sdk"

# Ensure dotnet comes first on PATH
export PATH="$DOTNET_ROOT:$PATH"

# Prevent fallback to global .NET locations. This ensures our tests use the shared frameworks we specify and don't rollforward to something else that might be installed on the machine
export DOTNET_MULTILEVEL_LOOKUP=0

# Avoid contaminating userprofiles
export DOTNET_CLI_HOME="$HELIX_CORRELATION_PAYLOAD/home"
```

HaoK:
@natemcmaster do I need to set some kind of OutputPath in the helix.proj, when I try to build with release configuration locally or on CI, I get this error:

```
D:\a\1\s\.dotnet\x64\sdk\2.1.500-preview-009398\Microsoft.Common.CurrentVersion.targets(780,5): error : The OutputPath property is not set for project 'helix.proj'.  Please check to make sure that you have specified a valid combination of Configuration and Platform for this project.  Configuration='Release'  Platform=''.  You may be seeing this message because you are trying to build a project without a solution file, and have specified a non-default Configuration or Platform that doesn't exist for this project. [D:\a\1\s\test\helix.proj]
```

natemcmaster:
@HaoK That seems like something Microsoft.DotNet.Helix.Sdk should fix. You can workaround by setting the property `<SkipInvalidConfigurations>true</SkipInvalidConfigurations>` in helix.proj

HaoK:
Helix runs now triggered from CI, redhat is completely failing now though hrm, https://mc.dot.net/#/user/dotnet-github-anon-kaonashi-bot/pr~2Faspnet~2Fextensions/ci/haok.11.29.2

Will fixup the metadata now

natemcmaster:
In servicing branches, the version of Microsoft.Extensions.Configuration.dll may not be the same across all test projects. This is because we attempt to maximize the compatibility of packages within a major.minor train. (e.g. in 2.2.x, projects still reference Microsoft.Extensions.Configuration 2.2.0).

I think the MethodAccessExceptions can be explain by changes to InternalsVisibleTo made as a part of mondo-repo refactoring.

HaoK:
Any ideas for things I can try to fix these?  This is a complication of having a single payload for all tests unfortunately...I'm guessing this won't be the last case where files are slightly different and it causes issues...

Should I have a second class of tests that continue to have their own dedicated payloads?

HaoK:
Or maybe can we just fix the InternalsVisible to for now so the tests pass?

HaoK:
Meaning we build our own manifest for the payload and have the workers download the files for the payload instead of using built in payload directory other than for shipping a payload manifest?

Basically just have a list of urls for the worker to assemble the payload?

natemcmaster:
What you described would be the ideal, but something I don't think we can implement on our own. If Helix did this out-of-the box, that would be awesome.

As a workaround, we could do something like this:

```
test-payload.zip
   manifest.txt
   files/
     5AA877025867541745C4828D24A13A4DB94580C1DC2D9F7B2CE50A0F7898DBA5
     758A85F5B28EEB661CCFBF1292A66587C334C0FDF0A01CD843F8438DF3FCC22A
```

manifest.txt maps files to their destination. Ideally, the file is something simple that bash/powershell can easily parse so copying files to the right location is easy. Something like this might work:
```
5AA877025867541745C4828D24A13A4DB94580C1DC2D9F7B2CE50A0F7898DBA5: test1/Microsoft.Extensions.Configuration.dll
758A85F5B28EEB661CCFBF1292A66587C334C0FDF0A01CD843F8438DF3FCC22A: test2/Microsoft.Extensions.Configuration.dll
758A85F5B28EEB661CCFBF1292A66587C334C0FDF0A01CD843F8438DF3FCC22A: test3/Microsoft.Extensions.Configuration.dll
758A85F5B28EEB661CCFBF1292A66587C334C0FDF0A01CD843F8438DF3FCC22A: test4/Microsoft.Extensions.Configuration.dll
```

natemcmaster:
Let's hold off on this. Just talked to the Helix team, and they thing the file hashing/de-dupping thing would be generally useful and not too hard to create, so they're going to look into adding this as a first-class thing in the Helix SDK.

natemcmaster:
cref https://github.com/dotnet/arcade/issues/1483 and https://github.com/dotnet/arcade/issues/1484

HaoK:
Ok restored back to the old payload per work item, will look into the test failures now:

https://mc.dot.net/#/user/dotnet-github-anon-kaonashi-bot/pr~2Faspnet~2Fextensions/ci/20181210.14

All queues
UserSecrets(2.0/2.1)
Logging.Analzyers(2.0/2.1)

Only Osx
Localization(2.0/2.1 only osx)
Sources(2.0/2.1 only osx)

Only ubuntu
Http(2.1)

HaoK:
I messed up this PR, replaced by https://github.com/aspnet/Extensions/pull/628

PR #570, opened 2018-11-27
Split scope resolution logic into seperate dynamic methods

pakrym:
Goal is to decrease method size in cases of large service trees with a lot of scoped services.

This change is based on https://github.com/aspnet/Extensions/pull/546.
I'll rebase once it goes in.

Fixes https://github.com/aspnet/AspNetCore/issues/3054 and https://github.com/aspnet/AspNetCore/issues/2737

Perf doesn't seem to be impacted.


pakrym:
Some perf numbers:

Before:

```

      Method |        Mode |        Mean |      Error |      StdDev |         Op/s | Allocated |
------------ |------------ |------------:|-----------:|------------:|-------------:|----------:|
      Scoped |     Dynamic |    78.37 ns |  1.9784 ns |   2.0317 ns | 12,760,581.9 |       0 B |
 LargeScoped |     Dynamic | 3,368.88 ns | 64.6920 ns |  60.5129 ns |    296,834.2 | 9200000 B |
      Scoped | Expressions |    79.04 ns |  0.2384 ns |   0.1991 ns | 12,651,093.5 |       0 B |
 LargeScoped | Expressions | 3,369.41 ns | 63.6107 ns |  59.5015 ns |    296,787.7 | 9200000 B |
      Scoped |      ILEmit |    78.58 ns |  1.1371 ns |   0.9495 ns | 12,725,924.2 |       0 B |
 LargeScoped |      ILEmit | 4,648.14 ns | 91.3279 ns | 105.1734 ns |    215,139.7 | 9200000 B |
      Scoped |     Runtime |   105.83 ns |  1.5913 ns |   1.4106 ns |  9,448,873.0 |       0 B |
 LargeScoped |     Runtime | 1,970.86 ns | 39.0219 ns |  57.1978 ns |    507,393.2 | 9200000 B |
```

After:

```

      Method |        Mode |        Mean |      Error |     StdDev |         Op/s | Allocated |
------------ |------------ |------------:|-----------:|-----------:|-------------:|----------:|
      Scoped |     Dynamic |    76.04 ns |  0.7249 ns |  0.6781 ns | 13,150,404.5 |       0 B |
 LargeScoped |     Dynamic | 1,059.56 ns | 19.9302 ns | 18.6427 ns |    943,787.1 | 9200000 B |
      Scoped | Expressions |    79.50 ns |  1.7264 ns |  2.0552 ns | 12,579,365.1 |       0 B |
 LargeScoped | Expressions | 1,031.06 ns |  3.1943 ns |  2.9879 ns |    969,879.3 | 9200000 B |
      Scoped |      ILEmit |    79.00 ns |  0.5764 ns |  0.5109 ns | 12,658,282.6 |       0 B |
 LargeScoped |      ILEmit | 1,019.38 ns |  3.2009 ns |  2.6729 ns |    980,989.3 | 9200000 B |
      Scoped |     Runtime |   100.78 ns |  0.5367 ns |  0.4758 ns |  9,922,939.0 |       0 B |
 LargeScoped |     Runtime | 1,919.32 ns | 25.6269 ns | 23.9714 ns |    521,018.4 | 9200000 B |
```

Simple case didn't regress while large tree case improved 4x just because of reduced method size.

src/DependencyInjection/DI/src/DependencyInjectionEventSource.cs
```diff
@@ -17,7 +17,8 @@ internal sealed class DependencyInjectionEventSource : EventSource
         // Event source doesn't support large payloads so we chunk formatted call site tree
         private int MaxChunkSize = 10 * 1024;
 
-        private DependencyInjectionEventSource()
+
+        private DependencyInjectionEventSource():base(EventSourceSettings.EtwSelfDescribingEventFormat)
```
halter73: nit: spaces around `:`

pakrym:
```suggestion
        private DependencyInjectionEventSource() : base(EventSourceSettings.EtwSelfDescribingEventFormat)
```

src/DependencyInjection/DI/src/ServiceLookup/Expressions/ExpressionResolverBuilder.cs
```diff
@@ -205,29 +186,31 @@ private static Expression Convert(Expression expression, Type type)
...
+        protected override Expression VisitScopeCache(ServiceCallSite callSite, object context)
         {
-            return BuildScopedExpression(callSite, context, VisitCallSiteMain(callSite, context));
+            var lambda = Build(callSite);
+            return Expression.Invoke(Expression.Constant(lambda), ScopeParameter);
         }
 
         // Move off the main stack
-        private Expression BuildScopedExpression(ServiceCallSite callSite, CallSiteExpressionBuilderContext context, Expression service)
+        private Expression BuildScopedExpression(ServiceCallSite callSite, object context)
```
halter73: Nit: context is unnecessary here.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
```diff
@@ -276,42 +270,7 @@ private void AddCacheKey(ILEmitResolverBuilderContext argument, ServiceCacheKey
...
-                callingConvention: CallingConventions.Standard,
-                returnType: typeof(object),
-                parameterTypes: new [] {typeof(ILEmitResolverBuilderRuntimeContext), typeof(ServiceProviderEngineScope) },
-                owner: GetType(),
-                skipVisibility: true);
-
-            var info = ILEmitCallSiteAnalyzer.Instance.CollectGenerationInfo(callSite);
-            var runtimeContext = GenerateMethodBody(callSite, dynamicMethod.GetILGenerator(info.Size), info);
-
-#if SAVE_ASSEMBLY
```
halter73: Do we have a test for this? Apparently this should have been SAVE_ASSEMBLIES even before this change.

src/DependencyInjection/DI/src/ServiceLookup/Expressions/ExpressionResolverBuilder.cs
```diff
@@ -41,102 +40,84 @@ internal class ExpressionResolverBuilder : CallSiteVisitor<CallSiteExpressionBui
...
-                // If root call site is singleton we can return Func calling
-                // _runtimeResolver.Resolve directly and avoid Expression generation
-                if (TryResolveSingletonValue(callSite, out var value))
-                {
-                    return scope => value;
-                }
-
-                return scope => _runtimeResolver.Resolve(callSite, scope);
+#if NETCOREAPP
+                return _scopeResolverCache.GetOrAdd(callSite.Cache.Key, (key, cs) => BuildNoCache(cs), callSite);
```
halter73: Is it really worth it to conditionally use this overload? `this` is captured by closure either way. You could cache the closure though.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
```diff
@@ -16,8 +17,6 @@ internal sealed class ILEmitResolverBuilder : CallSiteVisitor<ILEmitResolverBuil
         private static readonly MethodInfo ResolvedServicesGetter = typeof(ServiceProviderEngineScope).GetProperty(
```
halter73: Not part of this change, but it is a little weird to have TResult be Expression since it's never used. Probably makes sense to change it to object like TArgument in ExpressionResolverBuilder.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
```diff
@@ -49,25 +56,71 @@ private class ILEmitResolverBuilderRuntimeContext
             _runtimeResolver = runtimeResolver;
             _serviceScopeFactory = serviceScopeFactory;
             _rootScope = rootScope;
+            _scopeResolverCache = new ConcurrentDictionary<ServiceCacheKey, GeneratedMethod>();
         }
 
         public Func<ServiceProviderEngineScope, object> Build(ServiceCallSite callSite)
         {
-            if (callSite.Cache.Location == CallSiteResultCacheLocation.Root)
```
halter73: Might be worth an assert that `callSite.Cache.Location != CallSiteResultCacheLocation.Root` if only to make it obvious what's going on in the future.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
```diff
@@ -49,25 +56,71 @@ private class ILEmitResolverBuilderRuntimeContext
...
-                {
-                    return scope => value;
-                }
+            return BuildType(callSite).Lambda;
+        }
 
-                return scope => _runtimeResolver.Resolve(callSite, scope);
+        private GeneratedMethod BuildType(ServiceCallSite callSite)
+        {
+            // Only scope methods are cached
```
halter73: I'm probably missing something obvious, but why not cache GeneratedMethods for transient services?

halter73: I think I remember now. It's because the methods that don't need to check the ResolvedServices dictionary are already small enough. Still, if there's no perf impact of splitting scoped service instantiation into individual methods, it might be worth trying for non-scoped services.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
pakrym: This is a feature I only use when debugging code IL code generation it's always excluded from shipping builds.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
pakrym: My current thinking is that there is no perf impact because scope resolution methods themselves are comparably slow. There is also not much benefit in caching transient resolvers because most of them contain a single method call anyway.

src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
pakrym: Good point.

src/DependencyInjection/DI/src/ServiceLookup/Expressions/ExpressionResolverBuilder.cs
pakrym: You are right. https://github.com/dotnet/csharplang/issues/1277 would've helped.

PR #576, opened 2018-11-28
Reenable IL Emit DI backend

pakrym:
https://github.com/aspnet/Extensions/pull/570 is fixing the bug we had with code generation.

It has the same runtime performance but is faster and uses less memory to generate methods.

Runtime flag is not available yet.


davidfowl:
cc @ajcvickers We're going to enable this again. 

@pakrym I'm not sure we should be putting this in preview1. We need to verify that EF works.

pakrym:
I can merge it after preview1 goes out. We didn't have problems with EF and IL Emit before. I'll do a quick test with the music store.

ajcvickers:
@pakrym @davidfowl Anything that even might be risky should not go in while we're trying to get preview1 builds stabilized, so we should hold of on this.

/cc @muratg @natemcmaster @Eilon

natemcmaster:
@pakrym you're safe to checkin preview 2 work to this repo. I've only locked the master branch of aspnet/AspNetCore for preview 1 work.

PR #586, opened 2018-12-04
Use Arcade

ryanbrandenburg:
First pass at updating this repo to use dotnet/Arcade.


build/repo.targets
```diff
@@ -8,16 +8,6 @@
...
-    <MSBuild Projects="$(MSBuildThisFileDirectory)..\eng\ProdConPublish.csproj"
-             Targets="Restore"
-             Properties="$(BuildProperties);_Dummy=Restore" />
-
-    <MSBuild Projects="$(MSBuildThisFileDirectory)..\eng\ProdConPublish.csproj"
-             Targets="Publish"
-             Properties="$(BuildProperties)" />
-  </Target>
-
   <Target Name="GenerateProjectList" DependsOnTargets="ResolveProjects">
```
natemcmaster: We'll need to find a new place to put this. I don't know what the equivalent of repo.targets is in KoreBuild, but importing via Directory.Build.props is not it. The important thing is that we need to invoke `build.cmd /t:GenerateProjectList`, or something like it.

eng/Dependencies.props
```diff
@@ -57,6 +57,12 @@
     <LatestPackageReference Include="Microsoft.Build.Utilities.Core" Version="15.8.166" />
   </ItemGroup>
 
+  <!-- These versions mustn't be renamed because they over-ride the implicite package references from Arcade. -->
```
natemcmaster:
```suggestion
  <!-- These versions must not be renamed because they override the implicit package references from Arcade. -->
```

Nit: spelling

eng/ProdConPublish.csproj
```diff
@@ -11,14 +11,13 @@
     <!-- Detect branch and commit from the build environment -->
     <RepositoryBranch Condition=" '$(RepositoryBranch)' == '' ">$(BUILD_SOURCEBRANCH)</RepositoryBranch>
```
natemcmaster: This entire file can be deleted. Good bye prodcon!

eng/Versions.props
```diff
@@ -1,12 +1,26 @@
-﻿<Project>
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <MajorVersion>3</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <PreReleaseLabel>preview</PreReleaseLabel>
+    <PreReleaseVersionLabel>$(PreReleaseLabel)</PreReleaseVersionLabel>
```
natemcmaster: Is this an Arcade thing? Can we just rename PreReleaseLabel to PreReleaseVersionLabel? I think Arcade implements a lot of the funky MSBuild version match in this file and we may not need to repeat it.

eng/Versions.props
```diff
@@ -1,12 +1,26 @@
...
     <OfficialBuildId Condition="'$(OfficialBuildId)' == ''">$(BUILD_BUILDNUMBER)</OfficialBuildId>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <RestoreSources>
```
natemcmaster: We should remove this if we are going to have a non-empty NuGet.config.

eng/build.ps1
```diff
@@ -0,0 +1,52 @@
...
+        "Restore" { $defaultargs = "-restore" }
+        "Test" { $defaultargs = "-test -warnaserror:0" }
+        "InitTools" { $defaultargs = "-restore -warnaserror:0" }
+        default { $arguments += " /p:$argument=$($PSBoundParameters[$argument])" }
+    }
+}
+
+$arguments = "$defaultargs $arguments"
+
+Invoke-Expression "& `"$PSScriptRoot/common/build.ps1`" $arguments"
```
natemcmaster: Was this something we copied from somewhere? Because Invoke-Expression is one of the cardinal powershell sins. We should fix this.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/Microsoft.AspNetCore.Testing.csproj
```diff
@@ -12,6 +12,10 @@
     <UseLatestPackageReferences>true</UseLatestPackageReferences>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.AspNetCore.Testing.Tests" />
```
natemcmaster: 👍

src/ObjectPool/perf/Properties/AssemblyInfo.cs
```diff
@@ -1,4 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-[assembly: BenchmarkDotNet.Attributes.AspNetCoreBenchmark]
```
natemcmaster: I think we still need this attribute.

src/Logging/Logging.Testing/test/Properties/AssemlyInfo.cs
```diff
@@ -1,4 +0,0 @@
-﻿using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Testing;
-
-[assembly: LogLevel(LogLevel.Trace)]
```
natemcmaster: Still important. Let's undo this.

src/Logging/Logging.EventSource/test/AssemblyInfo.cs
```diff
@@ -1,10 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using Xunit;
-
-// There can only be one EventSource per AppDomain, and when an event is raised through that EventSource,
-// all existing listeners that enabled that EventSource will receive the event.
-// This makes running EventSourceLogger tests in parallel difficult. We mark this assembly
-// with CollectionBehavior.CollectionPerAssembly to ensure that all tests in this assembly are executed serially.
-[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]
```
natemcmaster: Please keep.

src/HttpClientFactory/Polly/test/Microsoft.Extensions.Http.Polly.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
+    <StrongNameKeyId>MicrosoftAspNet</StrongNameKeyId>
```
natemcmaster: Isn't this set by Directory.Build.props?

src/FileProviders/Manifest.MSBuildTask/test/Microsoft.Extensions.FileProviders.Embedded.Manifest.Task.Test.csproj
```diff
@@ -9,6 +9,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="xunit" />
```
natemcmaster: Why was this necessary? Is it because this project doesn't end in "Tests"? If so, let's rename the test project.

src/DependencyInjection/DI/perf/AssemblyInfo.cs
```diff
@@ -1 +0,0 @@
-[assembly: BenchmarkDotNet.Attributes.AspNetCoreBenchmark]
```
natemcmaster: I think we still need this.

src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
```diff
@@ -12,6 +12,7 @@
     -->
     <EnableApiCheck>false</EnableApiCheck>
     <IsTestProject>false</IsTestProject>
+    <IsUnitTestProject>false</IsUnitTestProject>
```
natemcmaster: Is IsUnitTestProject an Arcade thing?  I wonder if we can replace usages of IsTestProject. They seem redundant.

global.json
```diff
@@ -0,0 +1,11 @@
+{
+  "tools": {
+    "dotnet": "3.0.100-preview-009750"
+  },
+  "sdk": {
+    "version": "3.0.100-preview-009750"
```
natemcmaster: Why is this specified twice? Arcade requirement?

eng/tools/Directory.Build.props
```diff
@@ -1,3 +1,2 @@
 ﻿<Project>
```
natemcmaster: Can you add a comment that says "Intentionally empty to isolate tools projects from the result of the repo."

Directory.Build.props
```diff
@@ -43,15 +57,10 @@
   <PropertyGroup>
     <DeveloperBuildTestTfms>netcoreapp3.0</DeveloperBuildTestTfms>
     <StandardTestTfms>$(DeveloperBuildTestTfms)</StandardTestTfms>
-    <StandardTestTfms Condition=" '$(DeveloperBuild)' != 'true' AND '$(OS)' == 'Windows_NT' ">$(StandardTestTfms);net461</StandardTestTfms>
```
natemcmaster: Keep this. We need to continue running tests on .NET Framework.

Directory.Build.props
```diff
@@ -43,15 +57,10 @@
...
   </PropertyGroup>
 
+  <Import Project="build\repo.props" />
   <Import Project="eng\Dependencies.props" />
   <Import Project="eng\PatchConfig.props" />
   <Import Project="eng\ProjectReferences.props" />
-  <Import Project="eng\Sources.props" />
-
-  <ItemGroup>
-    <PackageReference Include="Internal.AspNetCore.Sdk" PrivateAssets="All" Version="$(InternalAspNetCoreSdkPackageVersion)" />
```
natemcmaster: I'm assuming we'll need to add a bunch of properties as a result of removing this SDK. For example, this bunch: https://github.com/aspnet/BuildTools/blob/master/src/Internal.AspNetCore.Sdk/build/Common.props#L12-L29.


We need to be careful not to regress important things, like metadata, package icons, etc.

eng/build.ps1
ryanbrandenburg: This is me ape-ing dotnet/corefx since they were the only ones I knew of using arcade at the time. I'll do https://github.com/aspnet/AspNetCore-Tooling/blob/master/build.cmd instead.

NuGet.config
```diff
@@ -2,6 +2,11 @@
 <configuration>
   <packageSources>
     <clear />
-    <!-- Restore sources should be defined in build/sources.props. -->
+    <add key="aspnetcore dev" value="https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json" />
+    <add key="aspnetcore tools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
+    <add key="arcade" value="https://dotnetfeed.blob.core.windows.net/dotnet-tools-internal/index.json" />
+    <add key="myget dnc" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
+    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
rynowak: I ended up needing to put all of the restore sources that I care about in `<RestoreSources>` for various reasons. The nuget.config in my repo just has the minimal stuff that arcade needs

azure-pipelines.yml
```diff
@@ -1,41 +1,161 @@
...
+
+  # Variables for internal Official builds
+  ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), contains(variables['Build.DefinitionName'], 'official')) }}:
+    _HelixType: build/product
+    _HelixSource: official/aspnet/AspNetCore-Tooling/$(Build.SourceBranch)
+    _SignType: test
+    _SignArgs: /p:DotNetSignType=$(_SignType) /p:TeamName=$(_TeamName)
+    _OfficialBuildIdArgs: /p:OfficialBuildId=$(Build.BuildNumber)
+    _PublishArgs: '/p:DotNetSymbolServerTokenMsdl=$(microsoft-symbol-server-pat) /p:DotNetSymbolServerTokenSymWeb=$(symweb-symbol-server-pat)'
+
```
rynowak: I've revised a bunch of stuff in the AspNetCore-Tooling repo since you took a copy, mostly to enable asset publishing. You might want to resolve that now, and just set BAR publishing to false

build/repo.targets
rynowak: I think in general the Arcade version of this is `Build.proj` - that's restored in a package. 

What I did for similar functionality was to put it in `Tools.props` because that's pulled in by `Build.proj`. However, I don't super understand what the purpose is of these things.

eng/Versions.props
rynowak: I ended up needing both, but that might be because of how testing for MSBuild is done in Razor.

global.json
```diff
@@ -0,0 +1,11 @@
+{
+  "tools": {
+    "dotnet": "3.0.100-preview-009750"
+  },
```
rynowak: What does this do? I just needed `"sdk"`

src/Caching/Memory/src/Microsoft.Extensions.Caching.Memory.csproj
```diff
@@ -8,6 +8,10 @@
     <PackageTags>cache;memorycache</PackageTags>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.Extensions.Caching.Memory.Tests" />
+  </ItemGroup>
```
rynowak: This is nice, we should start using this in Razour.

src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
rynowak: Yes - this is an arcade thing. They map tests into 3 categories (unit test, integration test, perf test). I'm not sure what's different about them.

src/FileProviders/Manifest.MSBuildTask/test/Microsoft.Extensions.FileProviders.Embedded.Manifest.Task.Test.csproj
rynowak: Yeah once a project is treated as a unit test project, arcade will nag you to remove this.

NuGet.config
natemcmaster: I don't think we'll need the same thing for this repo. NuGet.config is preferrable, but if there are scenarios which require restore-config-in-MSBuild, I'm okay with `<RestoreSources>` too. Let's just try to avoid keeping two lists of feeds.

global.json
ryanbrandenburg: Uhh, [that's not true](https://github.com/aspnet/AspNetCore-Tooling/blob/master/global.json#L3)?

global.json
```diff
@@ -0,0 +1,8 @@
...
+  "msbuild-sdks": {
```
rynowak: Make sure you **do** have SDK here. Our build just broke when they made some changes to the agents (because we missed specifying it here).

rynowak: https://github.com/aspnet/AspNetCore-Tooling/pull/17

src/Caching/Memory/src/Microsoft.Extensions.Caching.Memory.csproj
ryanbrandenburg: Yeah, I greatly prefer this to adding AssemblyInfo files which have to manually copy the key.

ryanbrandenburg:
@natemcmaster I worked around the skip problems [here](https://github.com/aspnet/Extensions/pull/586/commits/a80aff3cd3c87362aac15dc75521a541b78c3ace), but I'm not a big fan or the fix since it's pretty hacky. Thoughts or suggestions?

ryanbrandenburg:
@bricelam is looking into the binding issues.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/ConditionalTheoryDiscoverer.cs
```diff
@@ -14,6 +14,25 @@ public ConditionalTheoryDiscoverer(IMessageSink diagnosticMessageSink)
         {
         }
 
+        private sealed class OptionsWithPreEnumerationEnabled : ITestFrameworkDiscoveryOptions
```
natemcmaster: Why is this workaround required? Is there a bug in xunit or Arcade?

ryanbrandenburg:
@tmat did the heavy lifting here and said:

> Xunit has an option called preEnumerateTheories:
> “Gets a flag that determines whether theories are pre-enumerated. If enabled, then the discovery system will return a test case for each row of test data; they are disabled, then the discovery system will return a single > test case for the theory.”
> The console runner forces this option to false (setting it to true in xunit.runner.json does not have any effect), while vstest runner sets it to true.
> As written, the tests that validate that ConditionalTheoryDiscoverer works implicitly assume this option is true – that is that “each theory produces a test case for each row of test data”.

as well as pasting the snippet you commented on.

I wasn't able to come up with a better solution while keeping the current behavior of skipping the appropriate TheoryData, but then again I don't have much understanding of XUnit discovery mechanics. I can get spun up on it if this is a no-go, but I figured I'd ask what you think before spending that time.

tmat:
BTW, you can continue using `dotnet vstest` to run tests if you find the xunit console test runner insufficient. 

If you go this route I'd suggest copying `/eng/common/build.ps1` to `/eng/build.ps1`, updating it to execute tests using `dotnet vstest` when `-test` is passed and not pass `/t:Test` to `MSBuild` function. The same for `build.sh`. Roslyn also uses a different runner that optimizes test scheduling when running tests on desktop CLR:  https://github.com/dotnet/roslyn/blob/master/eng/build.ps1#L403-L405

eng/Versions.props
```diff
@@ -0,0 +1,55 @@
...
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MajorVersion>3</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>0</PatchVersion>
+    <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
+    <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/aspnet/AspNetCore/2.0.0/LICENSE.txt</PackageLicenseUrl>
+    <Serviceable Condition="'$(Configuration)' == 'Release'">true</Serviceable>
+    <UsingToolNetFrameworkReferenceAssemblies Condition="'$(OS)' != 'Windows_NT'">true</UsingToolNetFrameworkReferenceAssemblies>
```
ryanbrandenburg: @natemcmaster I'm not sure making this conditional on non-windows is the right move, but it's the only way I could figure out to not have binding re-direct issues.

natemcmaster: I'd rather have this always set to `true`, regardless of host OS. It means we don't have to force all contributors to acquiring a new targeting packs when we update netfx in our tests.

tmat: It's odd that this would change binding redirects. You need binding redirects regardless of where reference assemblies come from. Setting this to true means the ref assemblies are found in a nuget package, instead of Program Files.

ryanbrandenburg: @tmat this is the `Cannot find reference assembly 'System.ComponentModel.Composition.dll' file for package System.ComponentModel.Composition` thing that I brought up over email. I'm open to the above being the wrong solution but without this I get the missing reference assembly stuff, so I need an alternative solution or a suggestion on how to produce one.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/ConditionalTheoryDiscoverer.cs
ryanbrandenburg: I'm certainly not an XUnit expert but it seems to me like a bug in XUnit. There doesn't seem to be any way to skip a particular DataRow when `PreEnumerateTheories` is false. Unless this is "by design", in which-case our options would seem to be:

1. Do this workaround (or a similar one) to force enumeration in this scenario anyway.
2. Abandon this scenario all together and not have SkippableDataRows
3. Call it a known issue an move on.

I figure I'll file an issue against XUnit, a tracking issue on us and see where it gets us? Are we fine with leaving this in until we hear back from them?

natemcmaster: Yes, I'm fine leaving this as is for now.

ryanbrandenburg: https://github.com/xunit/xunit/issues/1871

ryanbrandenburg: https://github.com/aspnet/Extensions/issues/747

ryanbrandenburg:
The success is a lie. There's a problem in Arcade wherein when you have a failing test with warnaserrors=false the test failures just show as warnings. https://github.com/dotnet/arcade/issues/1589

eng/Versions.props
ryanbrandenburg: After much debate and concern, we eventually landed on essentially the fix I have above for this kind of problem.

ryanbrandenburg:
:up::date: @natemcmaster we're worked through the two blockers, If I could get your re-review I'll do one last pass of checking the nupkg metadata and we should be good to go.

ryanbrandenburg:
Talked to Nate, things that still remain:
1. Publishing packages.
2. Producing Source packages.
3. Verifying that all the packages we used to produce are still produced.
4. Update arcade version.

ryanbrandenburg:
:up::date: Source packages are now produced and we're publishing artifacts. I'm going to test our Manifest publishing from the official build if @natemcmaster doesn't have an objection.

Directory.Build.props
```diff
@@ -3,24 +3,59 @@
     Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), AspNetCoreSettings.props))\AspNetCoreSettings.props"
     Condition=" '$(CI)' != 'true' AND '$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), AspNetCoreSettings.props))' != '' " />
 
-  <Import Project="version.props" />
+  <PropertyGroup>
+    <UsingToolXliff>false</UsingToolXliff>
+  </PropertyGroup>
+
+  <!--
+    We don't follow their conventions for project naming.
```
natemcmaster:
```suggestion
    Update variables to match this repo's conventions for project naming.
```

Directory.Build.props
```diff
@@ -3,24 +3,59 @@
...
+
+  <PropertyGroup Condition="'$(IsUnitTestProject)' == 'false'">
+    <IsPackable>true</IsPackable>
+    <IsShipping>true</IsShipping>
+  </PropertyGroup>
+
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
   <PropertyGroup>
+    <Authors Condition="'$(Authors)' == ''">Microsoft</Authors>
```
natemcmaster: These duplicate what is already defined by Arcade:  https://github.com/dotnet/arcade/blob/c9f5d6ad24d3078f004168fbd4c476d489a87510/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.props#L12. You can remove the following:

* Authors
* Copyright
* Serviceable
* NeutralLanguage 
* PackageRequireLicenseAcceptance
* SuppressNETCoreSdkPreviewMessage

.gitignore
```diff
@@ -36,3 +36,6 @@ node_modules/
 .testPublish/
 BenchmarkDotNet.Artifacts/
 *.binlog
+.dotnet/
+**/src/xlf/
```
natemcmaster: Is this one still necessary?

Directory.Build.props
```diff
@@ -3,24 +3,59 @@
...
+    <Company Condition="'$(Company)' == ''">Microsoft Corporation.</Company>
+    <Copyright Condition="'$(Copyright)' == ''">© Microsoft Corporation. All rights reserved.</Copyright>
+    <IncludeSymbols Condition="'$(IncludeSymbols)' == ''">true</IncludeSymbols>
+    <NeutralLanguage Condition="'$(NeutralLanguage)' == ''">en-US</NeutralLanguage>
+    <PackageLicenseUrl Condition="'$(PackageLicenseUrl)' == ''">https://raw.githubusercontent.com/aspnet/AspNetCore/2.0.0/LICENSE.txt</PackageLicenseUrl>
```
natemcmaster: You can remove these conditionals. Directory.Build.props is evaluated before the content of projects. If anyone need to override this for a particular project, they can re-assign the value in the csproj file.
```suggestion
    <PackageLicenseUrl>https://raw.githubusercontent.com/aspnet/AspNetCore/2.0.0/LICENSE.txt</PackageLicenseUrl>
```

Directory.Build.targets
```diff
@@ -4,14 +4,14 @@
     <PackageOutputPath Condition="'$(IsProductComponent)' == 'true' ">$(ProductPackageOutputPath)</PackageOutputPath>
     <PackageOutputPath Condition="'$(IsProductComponent)' != 'true' ">$(InternalPackageOutputPath)</PackageOutputPath>
 
-    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp3.0' ">$(MicrosoftNETCoreAppPackageVersion)</RuntimeFrameworkVersion>
```
natemcmaster: Can you restore this line? We still want to ensure our tests run on the version of NETCoreApp that we specify, not the default picked by the SDK.

src/Shared/src/CommandLineUtils/Microsoft.Extensions.CommandLineUtils.Sources.projitems
```diff
@@ -1,14 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
natemcmaster: Can you also either restore this or update https://github.com/aspnet/Extensions/blob/rybrande/Arcade/eng/tools/BaselineGenerator/BaselineGenerator.csproj#L13? I'm assuming that tool doesn't build anymore as a result of deleting this file.

src/WebEncoders/src/Microsoft.Extensions.WebEncoders.csproj
```diff
@@ -7,6 +7,8 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore</PackageTags>
+    <IsPackable>true</IsPackable>
```
natemcmaster: I'd like to see if we can avoid having to define these two properties in every .csproj file. We already have convention-based settings in Directory.Build.props, and we also have the data to infer these values in most cases.

src/Configuration/Config.AzureKeyVault/src/Properties/AssemblyInfo.cs
```diff
@@ -1,8 +1,6 @@
...
-[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
 
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
natemcmaster: Does the `InternalsVisibleTo` property work for `DynamicProxyGenAssembly2`?

Directory.Build.props
```diff
@@ -29,15 +64,11 @@
       Packages which are intended to ship to NuGet.org must opt-in by setting this to true in their project file.
     -->
     <IsProductComponent>false</IsProductComponent>
```
natemcmaster: IsProductComponent is the KoreBuild version of IsShipping. Could we keep `IsProductComponent` as is and just put this code into Directory.Build.targets?
```xml
<IsShipping>$(IsProductComponent)</IsShipping>
```

Directory.Build.targets
```diff
@@ -4,14 +4,14 @@
     <PackageOutputPath Condition="'$(IsProductComponent)' == 'true' ">$(ProductPackageOutputPath)</PackageOutputPath>
```
natemcmaster: Does this still work? Or is Arcade overriding us?

azure-pipelines.yml
```diff
@@ -26,8 +26,8 @@ jobs:
     publishToBar: true
     artifacts:
       publish: true
-      name: artifacts
-      path: artifacts/
+      name: packages
+      path: artifacts/packages/Release
```
natemcmaster: This is going to break our release pipeline. 😬 let's chat about this.

build.cmd
```diff
@@ -1,2 +1,3 @@
-@ECHO OFF
-PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0build.ps1' %*; exit $LASTEXITCODE"
+@echo off
+powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0eng\common\Build.ps1""" -warnaserror $False -build -restore -pack %*"
```
natemcmaster: Can we open an item to finish resolving warnings and set `warnaserror` to true?

build/repo.targets
```diff
@@ -1,43 +0,0 @@
...
-    <MSBuild Projects="$(MSBuildThisFileDirectory)..\eng\tools.proj"
-             Targets="Restore"
-             Properties="$(BuildProperties);RestorePackagesPath=$(RepositoryRoot).packages\" />
-
-    <MSBuild Projects="$(MSBuildThisFileDirectory)..\eng\publish.proj"
-             Targets="Publish"
-             Properties="$(BuildProperties);RestorePackagesPath=$(RepositoryRoot).packages\" />
-  </Target>
-
-  <Target Name="GenerateProjectList" DependsOnTargets="ResolveProjects">
```
natemcmaster: We need this target still. It's used to generate the eng/ProjectReferences.props file. Is there an Arcade equivalent of repo targets?

build/repo.targets
```diff
@@ -1,43 +0,0 @@
-<Project InitialTargets="TagCiBuilds">
-  <Target Name="TagCiBuilds" Condition=" '$(CI)' == 'true' AND '$(BUILD_REASON)' != 'PullRequest' ">
```
natemcmaster: We still need these tags to be generated. These are used to control our release pipeline.

eng/Versions.props
```diff
@@ -1,6 +1,61 @@
...
+    <SystemRuntimeInteropServicesRuntimeInformationPackageVersion>4.3.0</SystemRuntimeInteropServicesRuntimeInformationPackageVersion>
+    <SystemSecurityCryptographyCngPackageVersion>4.6.0-preview.18604.2</SystemSecurityCryptographyCngPackageVersion>
+    <SystemSecurityCryptographyXmlPackageVersion>4.6.0-preview.18604.2</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemTextEncodingsWebPackageVersion>4.6.0-preview.18604.2</SystemTextEncodingsWebPackageVersion>
+    <SystemThreadingTasksExtensionsPackageVersion>4.6.0-preview.18604.2</SystemThreadingTasksExtensionsPackageVersion>
+    <SystemValueTuplePackageVersion>4.5.0</SystemValueTuplePackageVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <DotNetRuntime Include="$(MicrosoftNETCoreAppPackageVersion)" />
+    <DotNetRuntime Include="$(MicrosoftNETCoreApp20PackageVersion)" />
```
natemcmaster:
```suggestion
```
I don't think these repo needs .NET Core 2.0.0 anymore.

eng/Versions.props
```diff
@@ -1,6 +1,61 @@
...
+    <ExperimentalVersionPrefix>0.1.$(PatchVersion)</ExperimentalVersionPrefix>
+    <!-- The version in files -->
+    <ExperimentalPackageVersion>$(ExperimentalVersionPrefix)</ExperimentalPackageVersion>
+    <ExperimentalPackageVersion Condition=" '$(IncludePreReleaseLabelInPackageVersion)' == 'true' ">$(VersionPrefix)-$(VersionSuffix)</ExperimentalPackageVersion>
+    <ExperimentalVersionSuffix>$(PreReleaseVersionLabel)</ExperimentalVersionSuffix>
+    <!-- This is used for error checking to ensure generated code and baselines are up to date when we increment the patch. -->
+    <PreviousExtensionsReleaseVersion Condition=" '$(PatchVersion)' != '0' ">$(MajorVersion).$(MinorVersion).$([MSBuild]::Subtract($(PatchVersion), 1))</PreviousExtensionsReleaseVersion>
+  </PropertyGroup>
+  <PropertyGroup Label="Package Versions">
+    <MicrosoftNetCoreApp20PackageVersion>2.0.0</MicrosoftNetCoreApp20PackageVersion>
```
natemcmaster:
```suggestion
```

eng/Versions.props
```diff
@@ -1,6 +1,61 @@
...
+    <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/aspnet/AspNetCore/2.0.0/LICENSE.txt</PackageLicenseUrl>
+    <Serviceable Condition="'$(Configuration)' == 'Release'">true</Serviceable>
+    <UsingToolNetFrameworkReferenceAssemblies Condition="'$(OS)' != 'Windows_NT'">true</UsingToolNetFrameworkReferenceAssemblies>
+  </PropertyGroup>
+  <PropertyGroup>
+    <ExperimentalVersionPrefix>0.1.$(PatchVersion)</ExperimentalVersionPrefix>
+    <!-- The version in files -->
+    <ExperimentalPackageVersion>$(ExperimentalVersionPrefix)</ExperimentalPackageVersion>
+    <ExperimentalPackageVersion Condition=" '$(IncludePreReleaseLabelInPackageVersion)' == 'true' ">$(VersionPrefix)-$(VersionSuffix)</ExperimentalPackageVersion>
```
natemcmaster: Where is IncludePreReleaseLabelInPackageVersion defined?

eng/Versions.props
```diff
@@ -1,6 +1,61 @@
...
+    <SystemSecurityCryptographyXmlPackageVersion>4.6.0-preview.18604.2</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemTextEncodingsWebPackageVersion>4.6.0-preview.18604.2</SystemTextEncodingsWebPackageVersion>
+    <SystemThreadingTasksExtensionsPackageVersion>4.6.0-preview.18604.2</SystemThreadingTasksExtensionsPackageVersion>
+    <SystemValueTuplePackageVersion>4.5.0</SystemValueTuplePackageVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <DotNetRuntime Include="$(MicrosoftNETCoreAppPackageVersion)" />
+    <DotNetRuntime Include="$(MicrosoftNETCoreApp20PackageVersion)" />
+  </ItemGroup>
+  <!-- Run 'dotnet msbuild version.props' to test changes to this file. -->
```
natemcmaster: I don't this this works anymore. Remove?

eng/Versions.props
```diff
@@ -1,6 +1,61 @@
...
+  <Target Name="InspectVersionNumbers">
+    <Message Importance="High" Text="InformationalVersion    = '$(InformationalVersion)'" />
+    <Message Importance="High" Text="PackageVersion          = '$(PackageVersion)'" />
+  </Target>
+  <!--Package names-->
```
natemcmaster: I think this next property block is unused. https://github.com/dotnet/arcade/issues/1643

azure-pipelines.yml
```diff
@@ -26,8 +26,8 @@ jobs:
     publishToBar: true
```
natemcmaster: Can you open an issue to follow-up on converting this file to Arcade, too? In theory, our azure-pipelines.yml file is supposed to use templates from `eng/common/`.

NuGet.config
```diff
@@ -2,11 +2,11 @@
 <configuration>
   <packageSources>
     <clear />
-    <!--
-      Restore sources should be defined in build/sources.props.
-      The only allowed feed here is MSBuild SDK packages which is required to work around
-      https://github.com/Microsoft/msbuild/issues/2914
-    -->
+    <add key="aspnetcore dev" value="https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json" />
```
natemcmaster: You don't have to address this now, but let's open an item to prune our myget feeds. At this point, I think we shouldn't need any.

Extensions.sln
```diff
@@ -29,14 +29,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Shared", "Shared", "{36617B
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.Sources.Tests", "src\Shared\test\Microsoft.Extensions.Sources.Tests.csproj", "{33E57F0C-3CDB-4EB8-B6BE-6A0875B42EDE}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "testassets", "testassets", "{2C7E2059-8542-4BC8-A170-39F6B9EDE548}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ThrowingLibrary", "src\Shared\testassets\ThrowingLibrary\ThrowingLibrary.csproj", "{727D7A4D-49ED-4FBF-B713-1091191C78AD}"
-EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tools", "tools", "{7A597F60-480C-4BFF-A940-26094BB0CA62}"
-EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BaselineGenerator", "eng\tools\BaselineGenerator\BaselineGenerator.csproj", "{532F85C9-F589-40E7-814F-F0463A927788}"
```
natemcmaster: We still use this project. Can we keep it in the sln?

src/Primitives/perf/Properties/AssemblyInfo.cs
```diff
@@ -1,4 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-[assembly: BenchmarkDotNet.Attributes.AspNetCoreBenchmark]
```
natemcmaster: Presumably this attribute is still required, right?

version.props
```diff
@@ -1,65 +0,0 @@
...
-
-  <PropertyGroup>
-    <_BuildNumberSuffix Condition=" '$(_BuildNumberSuffix)' == '' ">0</_BuildNumberSuffix>
-    <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix>$(PreReleaseLabel).$(_BuildNumberSuffix)</VersionSuffix>
-
-    <ExperimentalVersionPrefix>0.1.$(PatchVersion)</ExperimentalVersionPrefix>
-
-    <!-- Servicing builds have different characteristics for the way dependencies, baselines, and versions are handled. -->
-    <IsServicingBuild Condition=" '$(PreReleaseLabel)' == 'servicing' ">true</IsServicingBuild>
```
natemcmaster: We need this property still. It's used by reference resolution in servicing builds. Can you keep this code? It can live with the other version variables.

version.props
```diff
@@ -1,65 +0,0 @@
...
-    <!-- Run the build with /p:IsFinalBuild=true to produce the product with 'final' branding and versioning -->
-    <IsFinalBuild Condition=" '$(IsFinalBuild)' == '' ">false</IsFinalBuild>
```
natemcmaster: Is there an Aracde equivalent to this? If not, we'll need to keep this code.

Directory.Build.targets
ryanbrandenburg: Arcade does override us. I'm removing our attempts to do it the old way in a new commit. If we want to try to keep it the old way it seems like we'd need Arcade to stop overwriting these?

src/Configuration/Config.AzureKeyVault/src/Properties/AssemblyInfo.cs
ryanbrandenburg: It does not 😢

Directory.Build.targets
natemcmaster: Let's keep our layout until we can figure out what to do with our release pipeline. Our release pipeline very much depends on the layout of the artifacts folder.

You can override SDK by assigning PackageOutputPath after the Import to Arcade's Sdk.targets file.

build/repo.targets
ryanbrandenburg: There's no repo.targets equivalent, but we can just do "dotnet msbuild repo.targets /t:GenerateProjectList". I made some tweeks and tested it out locally and it works (although it's not as "clean" as having the target already be there.)

build.cmd
ryanbrandenburg: https://github.com/aspnet/Extensions/issues/852

eng/Versions.props
ryanbrandenburg: it is now defined just above this.

version.props
ryanbrandenburg: It's now in Versions.props

version.props
ryanbrandenburg: Not as far as I can find. I added it back to `eng/Versions.props`

ryanbrandenburg:
🆙📅

Directory.Build.props
```diff
@@ -3,41 +3,55 @@
...
 
   <PropertyGroup>
+    <IncludeSymbols>true</IncludeSymbols>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/aspnet/AspNetCore/2.0.0/LICENSE.txt</PackageLicenseUrl>
+    <PackageIconUrl>https://go.microsoft.com/fwlink/?LinkID=288859</PackageIconUrl>
+    <PackageProjectUrl>https://asp.net</PackageProjectUrl>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <LangVersion>7.2</LangVersion>
+    <!-- Instructs the compiler to use SHA256 instead of SHA1 when adding file hashes to PDBs. -->
+    <ChecksumAlgorithm>SHA256</ChecksumAlgorithm>
```
natemcmaster: You can remove this too, now. It's a project default already.

src/Caching/samples/Directory.Build.props
```diff
@@ -1,8 +1,8 @@
 <Project>
-
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
 
   <PropertyGroup>
-    <PackageId>Microsoft.AspNetCore.BenchmarkRunner.Sources</PackageId>
+    <IsPackable>false</IsPackable>
```
natemcmaster: I think most of the places where you have added IsPackable should already be covered by the conventions expressed in the repo root Directory.Build.props. Are these broken by Arcade?

E.g. samples, perf projects, and test projects should already be set to false

natemcmaster:
I believe this is still going to break our release pipeline. I'll stop by your desk and show you what I mean, but let's try to resolve that first.

src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
```diff
@@ -11,7 +11,11 @@
       https://github.com/dotnet/corefx/issues/12338 & https://github.com/dotnet/corefx/issues/11797
     -->
     <EnableApiCheck>false</EnableApiCheck>
+    <IsUnitTestProject>false</IsUnitTestProject>
     <IsTestProject>false</IsTestProject>
+    <TestRunnerName></TestRunnerName>
```
natemcmaster: What's this about?

ryanbrandenburg:
1. YML needs to publish packages to artifacts.
2. Double check that directory structure is kept.
3. Create tags from eng/repo.targets.

src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
ryanbrandenburg: Context [here](https://github.com/dotnet/arcade/issues/1464). This is the work-around I landed on to get us unblocked. Tomas had some other suggestions but I wanted to keep moving forward so I've stuck with what works. I filed [this issue](https://github.com/aspnet/Extensions/issues/901) to track it.

HaoK:
@ryanbrandenburg how do you want to coordinate this with https://github.com/aspnet/Extensions/pull/628  do you want me to rebase onto your PR?  Or should I try to get mine in first since its more self contained maybe?

ryanbrandenburg:
@HaoK you do you. I'm going to be out of the office for a bit, so unless someone volunteers to babysit any fallout from this PR it will likely have to wait to merge until I come back. In that case you should just go ahead and merge, it doesn't look like there will be too much conflict anyway.

src/Caching/samples/Directory.Build.props
ryanbrandenburg: It seems like they are broken by Arcade. It's not clear to me in what order files are imported in Arcade, but I couldn't seem to make generalized IsPackable rules work. So I figured I' do the thing that works for now and filed https://github.com/aspnet/Extensions/issues/915 to re-examine this later.

ryanbrandenburg:
:up::date: but I know @natemcmaster is busy and this isn't an emergency so no rush.

ryanbrandenburg:
This is blocked by the failures of JSInterop. I'm going to need either @javiercn or @SteveSandersonMS's help to fix these tests before we can merge.

PR #590, opened 2018-12-05
Fix errors in VS about ProjectSystemFault: Replace all usages of `<Reference>` with `<Dependency>`

natemcmaster:
Still haven't figured out what leads to https://developercommunity.visualstudio.com/content/problem/377676/argumentexception-thrown-by-referenceshostbridge-a.html (cref https://github.com/dotnet/project-system/issues/4237), but it seems the root cause is that we are attempting to use the `<Reference>` item group name.

This PR replaces all usages of `<Reference>` with `<Dependency>` to workaround this.

### DIFF
--- a/Extensions.sln
+++ b/Extensions.sln
@@ -245,8 +245,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{1E66
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GenericHostSample", "src\Hosting\samples\GenericHostSample\GenericHostSample.csproj", "{97AE13C4-B72C-4E30-8835-FF1603A9E224}"
 EndProject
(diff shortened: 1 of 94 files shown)


natemcmaster:
Command line builds work fine with either item group name. I attempted to figure out what was wrong in VS. Design time builds in VS are running our custom `<Reference>` resolution correctly, but CPS was still blowing up with an ArgumentException in ReferencesHostBridge. I don't have the VS debugging skills to figure out what is wrong, so I'm giving up on using `<Reference>`. Just a hunch - my guess is there is something else in VS that still reads the project statically to find `<Reference>` items, and this leads to duplicate keys in some internal VS dictionary because we convert `<Reference>` to `<ProjectRefenence>` or `<PackageReference>` based on which is the correct thing to do given versions/servicing/patching etc.

cc @Pilchie @davkean

natemcmaster:
Closing in favor of a less-intrusive workaround: https://github.com/aspnet/Extensions/pull/616.

PR #591, opened 2018-12-06
Update Dependencies.props

sebastienros:
Updating core-setup and corefx packages with the latest build results.

Please triple check as this is the first time I am trying this.

### DIFF
--- a/eng/Dependencies.props
+++ b/eng/Dependencies.props
@@ -1,4 +1,4 @@
-﻿<Project>
+<Project>
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


sebastienros:
Don't tell me what's wrong right now, I want to understand by myself like a grown-up.

sebastienros:
It looks like it's due to the fast that prefixes went from `preview1` to `preview`.

For instance: 

`System.Memory 4.6.0-preview1-26717-04 -> System.Runtime.CompilerServices.Unsafe (>= 4.6.0-preview1-26717-04)`

The `System.Memory` package is the correct latest version available. But the version specified for `System.Runtime.CompilerServices.Unsafe` in other references which is `4.6.0-preview.18606.1` is found to be lower that the `>= 4.6.0-preview1-26717-04` dependency that `System.Memory` has on it. 

This result in a `package downgrade` build error.

AFAIK it can only be fixed by referencing explicitly `System.Runtime.CompilerServices.Unsafe` in the projects that have a conflict.

sebastienros:
I looked at why it worked before this PR. `System.Memory` was on `4.5.2-servicing-27114-05` which is newer than `4.6.0-preview1-26717-04`, and actually has a lower upper bound which made it work with `preview` prefixes. Oh boy!!! I could be tempted to use some words that I think it would be bad for my job ;)

sebastienros:
Should I file an issue somewhere for the `4.6` package so that the `System.Memory` packages on myget get unlisted? Or at least such that https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt actually points to the correct version?

eng/Dependencies.props
```diff
@@ -7,29 +7,28 @@
...
-    <SystemSecurityCryptographyCngPackageVersion>4.6.0-preview.18571.3</SystemSecurityCryptographyCngPackageVersion>
-    <SystemSecurityCryptographyXmlPackageVersion>4.6.0-preview.18571.3</SystemSecurityCryptographyXmlPackageVersion>
-    <SystemTextEncodingsWebPackageVersion>4.6.0-preview.18571.3</SystemTextEncodingsWebPackageVersion>
-    <SystemThreadingTasksExtensionsPackageVersion>4.6.0-preview.18571.3</SystemThreadingTasksExtensionsPackageVersion>
-    <SystemValueTuplePackageVersion>4.5.0</SystemValueTuplePackageVersion>
+    <SystemSecurityCryptographyCngPackageVersion>4.6.0-preview.18606.1</SystemSecurityCryptographyCngPackageVersion>
+    <SystemSecurityCryptographyXmlPackageVersion>4.6.0-preview.18606.1</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemTextEncodingsWebPackageVersion>4.6.0-preview.18606.1</SystemTextEncodingsWebPackageVersion>
+    <SystemThreadingTasksExtensionsPackageVersion>4.6.0-preview.18606.1</SystemThreadingTasksExtensionsPackageVersion>
+    <SystemValueTuplePackageVersion>4.6.0-preview1-26829-04</SystemValueTuplePackageVersion>
```
natemcmaster:
```suggestion
    <SystemValueTuplePackageVersion>4.5.0</SystemValueTuplePackageVersion>
```

We need to keep this one at 4.5.0. There won't be a 4.6.0 release. This is one of the packages that corefx no longer produces from their master branch. "4.6.0-preview1-26829-04" is an old, dead-end version.

sebastienros: Is it the same reasoning for System.Memory?

PR #592, opened 2018-12-06
[2.2.x] Port Drop messages in azure loggers if queue is full

pakrym:
Port of https://github.com/aspnet/Extensions/pull/494 to 2.2

cc @muratg 

### Description:
Today, when Azure AppServices Log Stream logger queue becomes full we start blocking in calls to `logger.Log` that often happen on thread pool threads causing thread pool starvation. Making things worse the process of writing messages uses `HttpClient` and queues thread pool items itself making starvation almost non-recoverable.

### Customer Impact

If customers are using Azure AppServices logger and get a spike of logging messages on multiple ThreadPool threads their app might effectively "hang" and stop serving requests.

### Regression?

No

### Risk:
Low. Fix is limited to the case where the queue if full and the logic is just to drop messages.

### DIFF
--- a/eng/PatchConfig.props
+++ b/eng/PatchConfig.props
@@ -6,6 +6,7 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(VersionPrefix)' == '2.2.2' ">
     <PackagesInPatch>
+      Microsoft.Extensions.Logging.AzureAppServices
     </PackagesInPatch>
   </PropertyGroup>
 
--- a/src/Logging/Logging.AzureAppServices/src/Internal/BatchingLoggerProvider.cs
+++ b/src/Logging/Logging.AzureAppServices/src/Internal/BatchingLoggerProvider.cs
@@ -18,6 +18,8 @@ namespace Microsoft.Extensions.Logging.AzureAppServices.Internal
         private readonly int? _batchSize;
         private readonly IDisposable _optionsChangeToken;
 
+        private int _messagesDropped;
+
         private BlockingCollection<LogMessage> _messageQueue;
         private Task _outputTask;
         private CancellationTokenSource _cancellationTokenSource;
@@ -78,6 +80,16 @@ namespace Microsoft.Extensions.Logging.AzureAppServices.Internal
                     limit--;
                 }
 
+                var messagesDropped = Interlocked.Exchange(ref _messagesDropped, 0);
+                if (messagesDropped != 0)
+                {
+                    _currentBatch.Add(new LogMessage()
+                    {
+                        Message = $"{messagesDropped} message(s) dropped because of queue size limit. Increase the queue size or decrease logging verbosity to avoid this.{Environment.NewLine}",
+                        Timestamp = DateTimeOffset.Now
+                    });
+                }
+
                 if (_currentBatch.Count > 0)
                 {
                     try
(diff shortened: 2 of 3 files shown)


vivmishra:
Approved for 2.2.2

src/Logging/Logging.AzureAppServices/src/Internal/BatchingLoggerProvider.cs
```diff
@@ -91,8 +103,10 @@ private async Task ProcessLogQueue(object state)
 
                     _currentBatch.Clear();
                 }
-
-                await IntervalAsync(_interval, _cancellationTokenSource.Token);
+                else
+                {
+                    await IntervalAsync(_interval, _cancellationTokenSource.Token);
```
halter73: I'm not sure moving this into an else is worth it in a patch. It doesn't seem necessary for dropping messages, and it has the potential to make ProcessLogQueue more CPU hungry. Unless we have some benchmark or stress scenario that shows a major benefit.

pakrym: It allows dropping fewer messages by unloading them faster. Without doing this and having no backpressure a lot of messages might be lost even if app is not overloaded

natemcmaster:
Please wait until https://github.com/aspnet/AspNetCore-Internal/issues/1610 is resolved to merge changes.

pakrym:
:up::date:

@halter73 - I've reverted changes in question
@dougbu - I've added PackagesInPatch

src/Logging/Logging.AzureAppServices/src/Internal/BatchingLoggerProvider.cs
```diff
@@ -78,6 +80,16 @@ private async Task ProcessLogQueue(object state)
                     limit--;
                 }
 
+                var messagesDropped = Interlocked.Exchange(ref _messagesDropped, 0);
+                if (messagesDropped != 0)
+                {
+                    _currentBatch.Add(new LogMessage()
+                    {
+                        Message = $"{messagesDropped} message(s) dropped because of queue size limit. Increase the queue size or decrease logging verbosity to avoid this.{Environment.NewLine}",
```
halter73: What do you think about referencing BatchingLoggerOptions in this message or including a URL for the docs?

src/Logging/Logging.AzureAppServices/src/Internal/BatchingLoggerProvider.cs
```diff
@@ -107,7 +118,10 @@ internal void AddMessage(DateTimeOffset timestamp, string message)
             {
                 try
                 {
-                    _messageQueue.Add(new LogMessage { Message = message, Timestamp = timestamp }, _cancellationTokenSource.Token);
+                    if (!_messageQueue.TryAdd(new LogMessage { Message = message, Timestamp = timestamp }, millisecondsTimeout: 0, cancellationToken: _cancellationTokenSource.Token))
```
halter73: Is the `millisecondsTimeout: 0` needed for this change?

pakrym: https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1.tryadd?view=netframework-4.7.2#System_Collections_Concurrent_BlockingCollection_1_TryAdd__0_System_Int32_System_Threading_CancellationToken_

It's not an optional parameter in this overload.

eng/PatchConfig.props
```diff
@@ -6,6 +6,7 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(VersionPrefix)' == '2.2.2' ">
     <PackagesInPatch>
+      Microsoft.Extensions.Logging.AzureAppServices
```
dougbu: 👍

PR #599, opened 2018-12-07
Update Dependencies.props

sebastienros:


sebastienros:
@natemcmaster starting fresh, and keeping the same core-setup and corefx dependencies all along this time, even if new builds happen along.

PR #600, opened 2018-12-07
Added tests that can be implemented for any provider to ensure consistency

ajcvickers:
Partly to allow me to get a better understanding of the code, partly so we can have a better idea that all the providers work in a consistent manner.

Part of https://github.com/aspnet/Configuration/issues/559

I will file issues for things found.


src/Configuration/Config/test/Microsoft.Extensions.Configuration.Tests.csproj
```diff
@@ -1,9 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
+    <TargetFrameworks>$(StandardTestTfms);net461</TargetFrameworks>
```
HaoK: Just making sure this net461 is intentional

ajcvickers: Yes, because the AzureKeyVault tests target it, and they now depend on this for the base infrastructure.

PR #606, opened 2018-12-09
[2.1+] Add robustness to Config FileProvider tests

ajcvickers:
Specifically:
* Don't let current thread continue until file system change is visible
* Retry when waiting for notifications

Fixes: https://github.com/aspnet/AspNetCore-Internal/issues/1319

### DIFF
--- a/src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
+++ b/src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
@@ -5,9 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


ajcvickers:
@natemcmaster Can we merge test changes for 2.1 and 2.2 here?

natemcmaster:
I believe test-only fixes to 2.x branches is okay, but we usually check with @Eilon just be sure.

ajcvickers:
@natemcmaster I actually was just asking if this was the right place for 2.1 fixes.

natemcmaster:
Yes, this is the right place for 2.1 fixes.

ajcvickers:
@HaoK Finally have these passing on 2.1--I did the work initially on 3.0, but figured we should backport. Can you review? Also see https://github.com/aspnet/Extensions/issues/617

Issue #571, opened 2018-11-28
BachLoggingProvider: There's no need to use async methods inside a dedicated thread 


BBGONE:
If there will be a real asynchronous operation, then its continuation will be on a thread pool thread. After that it won't return to the dedicated thread and can potentially use one threadpool thread for the rest of application lifetime.

```
_outputTask = Task.Factory.StartNew<Task>(
    ProcessLogQueue,
    null,
    TaskCreationOptions.LongRunning);
```

https://github.com/aspnet/Extensions/blob/master/src/Logging/Logging.AzureAppServices/src/BatchingLoggerProvider.cs

This certainly  will have continuation on a threadpool thread:

```
protected virtual Task IntervalAsync(TimeSpan interval, CancellationToken cancellationToken)
{
            return Task.Delay(interval, cancellationToken);
}
```

### Describe the solution you'd like
Remove async methods. It's not a problem to block a dedicated thread. It is used only by only one task.

### Describe alternatives you've considered
Or start the task without TaskCreationOptions.LongRunning (but that was probably was meant and desirable to run the logger on dedicated thread - so it is not a good option)

### Additional context
https://github.com/andrewlock/NetEscapades.Extensions.Logging/issues/8

pakrym:
`TaskCreationOptions.LongRunning` doesn't start a dedicated thread, it would still run on thread. Using sync methods would just mean that we are blocking a thread waiting on an async code that still would use thread pool to complete.

BBGONE:
Do you mean that now, CoreFx has a bug and uses a threadpool thread to execute a longrunning task on it? 
In the full Net Framework it used a dedecated thread. I will try to find the code for the default TaskScheduler for CoreFx to check it.

pakrym:
Yep, you are right. https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/Threading/Tasks/ThreadPoolTaskScheduler.cs#L42 . Let's change it to `Task.Run` then.

pakrym:
Want to send a PR?

BBGONE:
Yes, it will work.  But in my own implementation, i will use a dedicated thread. It is more robust and reliable in this case.

BBGONE:
What is PR?

pakrym:
A pool request.

BBGONE:
No. I dont use it. I found it only because of trying to create my own LogProvider. I used it as example.
I did some optimisations
[...]

PR #611, opened 2018-12-09
Use Task.Run instead of Task.Factory.StartNew

andrewlock:
Using `TaskCreationOptions.LongRunning` creates a dedicated thread, but it's released on the first await. Using `Task.Run` uses the thread pool instead.

Addresses #571

### DIFF
--- a/src/Logging/Logging.AzureAppServices/src/BatchingLoggerProvider.cs
+++ b/src/Logging/Logging.AzureAppServices/src/BatchingLoggerProvider.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Extensions.Logging.AzureAppServices
 
         internal abstract Task WriteMessagesAsync(IEnumerable<LogMessage> messages, CancellationToken token);
 
-        private async Task ProcessLogQueue(object state)
+        private async Task ProcessLogQueue()
         {
             while (!_cancellationTokenSource.IsCancellationRequested)
             {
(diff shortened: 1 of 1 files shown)


src/Logging/Logging.AzureAppServices/src/BatchingLoggerProvider.cs
```diff
@@ -143,10 +143,7 @@ private void Start()
                 new BlockingCollection<LogMessage>(new ConcurrentQueue<LogMessage>(), _queueSize.Value);
 
             _cancellationTokenSource = new CancellationTokenSource();
-            _outputTask = Task.Factory.StartNew<Task>(
-                ProcessLogQueue,
-                null,
-                TaskCreationOptions.LongRunning);
+            _outputTask = Task.Run(ProcessLogQueue);
```
pakrym: Do you need to `Task.Run`?  Calling `ProcessLogQueue` should be enough.

andrewlock: As I understand it, Calling `ProcessLogQueue()` directly would start running `ProcessLogQueue` inline, on the same thread, instead of using a `ThreadPool` thread. It would only return and complete the constructor after the first `await` is hit (either `WriteMessagesAsync` if that completes asynchronously, or `IntervalAsync`).

While that's _probably_ fine, it seems undesirable to be doing real work here rather than just kicking off the background work, but I'm happy to defer to you on that 🙂

pakrym:
Thank you, @andrewlock!

PR #616, opened 2018-12-10
Workaround errors in Visual Studio project system

natemcmaster:
Alternative to #590 which doesn't require editing every .csproj, but has the same effect. This should solve the issue with project system error prompts which were popping up dozens of times when VS launched. Tested in VS 2017 15.9.

FYI @ajcvickers


natemcmaster:
Hmm, github <-> Azure Pipelines is not playing nice today.

/AzurePipelines run Extensions-ci

PR #628, opened 2018-12-11
[WIP] Use helix on CI

HaoK:
Replaces https://github.com/aspnet/Extensions/pull/567 (which I busted with a bad force push)

PR validation contains a new helix job which can be used to view MC tests results (in bottom build.cmd output for link to mc.dot.net)

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -76,7 +76,7 @@
 
   <Target Name="_CreateTestPayloadInner" 
     DependsOnTargets="Publish;CollectXunitConsoleRunner" 
-    Condition="'$(IsTestProject)' == 'true' AND ($(HelixTargetQueue).StartsWith('Windows') OR '$(TargetFrameworkIdentifier)' != '.NETFramework')" 
+    Condition="'$(IsTestProject)' == 'true' AND ($(HelixTargetQueue.StartsWith('Windows')) OR '$(TargetFrameworkIdentifier)' != '.NETFramework')" 
     Returns="@(TestPayload)">
 
     <ConvertToAbsolutePath Paths="$(PublishDir)">
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -76,7 +76,7 @@
 
   <Target Name="_CreateTestPayloadInner" 
     DependsOnTargets="Publish;CollectXunitConsoleRunner" 
-    Condition="'$(IsTestProject)' == 'true' AND ('$(HelixTargetQueue)' == 'Windows.10.Amd64.Open' OR '$(TargetFrameworkIdentifier)' != '.NETFramework')" 
+    Condition="'$(IsTestProject)' == 'true' AND ($(HelixTargetQueue).StartsWith('Windows') OR '$(TargetFrameworkIdentifier)' != '.NETFramework')" 
     Returns="@(TestPayload)">
 
     <ConvertToAbsolutePath Paths="$(PublishDir)">
@@ -84,17 +84,17 @@
     </ConvertToAbsolutePath>
 
     <!-- Windows NetCore -->
-    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework' AND '$(HelixTargetQueue)' == 'Windows.10.Amd64.Open'" >
+    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework' AND $(HelixTargetQueue.StartsWith('Windows'))" >
         <_CopyItems Include="$(MSBuildThisFileDirectory)test\vstest\runtests.cmd" />
     </ItemGroup>
 
     <!-- Windows NetFramework -->
-    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' AND '$(HelixTargetQueue)' == 'Windows.10.Amd64.Open'" >
+    <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' AND $(HelixTargetQueue.StartsWith('Windows'))" >
         <_CopyItems Include="$(MSBuildThisFileDirectory)test\xunit\runtests.cmd" />
     </ItemGroup>
 
     <!-- NonWindows -->
-    <ItemGroup Condition="'$(HelixTargetQueue)' != 'Windows.10.Amd64.Open'"  >
+    <ItemGroup Condition="!$(HelixTargetQueue.StartsWith('Windows'))"  >
         <_CopyItems Include="$(MSBuildThisFileDirectory)test\vstest\runtests.sh" />
     </ItemGroup>
 
@@ -105,8 +105,8 @@
         <TestAssembly>$(TargetFileName)</TestAssembly>
         <TestName>$(MSBuildProjectName)-$(TargetFramework)</TestName>
         <!-- Look into how to be able to use $(NETCoreSdkVersion) -->
-        <Command Condition="'$(HelixTargetQueue)' == 'Windows.10.Amd64.Open'">runtests.cmd $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
-        <Command Condition="'$(HelixTargetQueue)' != 'Windows.10.Amd64.Open'">./runtests.sh $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
+        <Command Condition="$(HelixTargetQueue.StartsWith('Windows'))">runtests.cmd $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
+        <Command Condition="!$(HelixTargetQueue.StartsWith('Windows'))">./runtests.sh $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
         <TestTimeout>$(DistributedTestTimeout)</TestTimeout>
       </TestPayload>
     </ItemGroup>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -103,7 +103,7 @@
     <ItemGroup>
       <TestPayload Include="$(PublishAbsoluteDir.TrimEnd('\').TrimEnd('\'))">
         <TestAssembly>$(TargetFileName)</TestAssembly>
-        <TestName>$(MSBuildProjectName)/$(TargetFramework)</TestName>
+        <TestName>$(MSBuildProjectName)-$(TargetFramework)</TestName>
         <!-- Look into how to be able to use $(NETCoreSdkVersion) -->
         <Command Condition="'$(HelixTargetQueue)' == 'Windows.10.Amd64.Open'">runtests.cmd $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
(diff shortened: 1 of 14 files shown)


HaoK:
Preserving unaddressed comments from @natemcmaster on old pr:

- Starting to think we should have runtests.cmd and runtests.ps1, and make the cmd a dummy script that starts powershell.

- We'll need to use different queue names for internal builds. We should find a different way to check if a queue is targeting windows that doesn't depend on the exact queue name "Windows.10.Amd64.Open"

- Use $(NETCoreSdkVersion) to get the .NET Core SDK version.

HaoK:
@natemcmaster how about using `$(HelixTargetQueue.StartsWith('Windows')` to determine windows queue or not?  That doesn't seem too unreasonable

Directory.Build.targets
```diff
@@ -44,4 +44,71 @@
...
+
+    <!-- NonWindows -->
+    <ItemGroup Condition="!$(HelixTargetQueue.StartsWith('Windows'))"  >
+        <_CopyItems Include="$(MSBuildThisFileDirectory)test\vstest\runtests.sh" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(PublishAbsoluteDir)" />
+
+    <ItemGroup>
+      <TestPayload Include="$(PublishAbsoluteDir.TrimEnd('\').TrimEnd('\'))">
```
natemcmaster:
```suggestion
      <TestPayload Include="$(PublishAbsoluteDir.TrimEnd('\').TrimEnd('/'))">
```

azure-pipelines.yml
```diff
@@ -39,3 +39,14 @@ jobs:
...
     configuration: Release
+
+- job: Helix
+  pool:
+    name: Hosted VS2017
+    vmImage: vs2017-win2016
+  steps:
+  - checkout: self
+    clean: true
+  - script: .\build.cmd -ci /t:Helix /p:Configuration=Debug
```
natemcmaster:
```suggestion
  - script: .\build.cmd -ci /t:Helix
```

We usually don't run the Debug config for tests. Let's try doing this for Release configs instead.

build/repo.targets
```diff
@@ -44,4 +45,17 @@
 
     <WriteLinesToFile File="$(ProjectListFile)" Lines="$(ProjectListContent)" Overwrite="true" />
   </Target>
+
+<!-- will move into korebuild -->
+  <PropertyGroup>
+    <DefaultHelixTestTimeout>00:30:00</DefaultHelixTestTimeout>
```
natemcmaster: This isn't used anywhere, so I believe you can remove it.

test/helix.proj
```diff
@@ -0,0 +1,48 @@
...
+      </HelixWorkItem>
+    </ItemGroup>
+  </Target>
+
+  <PropertyGroup>
+    <!-- TODO: Follow up with helix SDK -->
+    <SkipInvalidConfigurations>true</SkipInvalidConfigurations>
+    <HelixSource>pr/aspnet/extensions</HelixSource>
+    <HelixType>ci</HelixType>
+    <HelixBuild>private-$(USERNAME)</HelixBuild>
```
natemcmaster: Add a fallback to `$(USER)` when USERNAME is empty, e.g. macOS/Linux.

test/helix.proj
```diff
@@ -0,0 +1,48 @@
...
+    <HelixBuild Condition=" '$(CI)' == 'true' ">$(BUILD_BUILDNUMBER)</HelixBuild>
+    <IsExternal>true</IsExternal>
```
natemcmaster: Should this be false when building official builds internally?

Directory.Build.targets
```diff
@@ -44,4 +44,71 @@
...
+        <TestAssembly>$(TargetFileName)</TestAssembly>
+        <TestName>$(MSBuildProjectName)-$(TargetFramework)</TestName>
+        <Command Condition="$(HelixTargetQueue.StartsWith('Windows'))">runtests.cmd $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
+        <Command Condition="!$(HelixTargetQueue.StartsWith('Windows'))">./runtests.sh $(TargetFileName) $(NETCoreSdkVersion) $(RuntimeFrameworkVersion)</Command>
+        <TestTimeout>$(DistributedTestTimeout)</TestTimeout>
```
natemcmaster: DistributedTestTimeout isn't set anywhere. Was this mean to align with DefaultHelixTestTimeout? If so, the timeout property needs to be in Directory.Build.props, not build/repo.targets.

test/helix.proj
```diff
@@ -0,0 +1,48 @@
...
+    <Creator>aspnet-extensions</Creator>
+    <WaitForWorkItemCompletion Condition=" '$(CI)' == 'true' ">true</WaitForWorkItemCompletion>
+
+    <HelixTargetQueues>Windows.10.Amd64.Open;OSX.1012.Amd64.Open;Ubuntu.1604.Amd64.Open</HelixTargetQueues>
+    <!-- <HelixTargetQueues>Windows.10.Amd64.Open;OSX.1012.Amd64.Open;Ubuntu.1604.Amd64.Open;Ubuntu.1810.Amd64.Open;Centos.7.Amd64.Open;Debian.8.Amd64.Open;Debian.9.Amd64.Open;Fedora.27.Amd64.Open;Fedora.28.Amd64.Open;Redhat.7.Amd64.Open;Redhat.6.Amd64.Open;</HelixTargetQueues> -->
```
natemcmaster: Let's try running the full suite of queues and see how it goes.

Directory.Build.targets
HaoK: This is supposed to have been fixed, I will try removing this trim to see if their fix works

test/helix.proj
HaoK: I'm not sure, I think we have to target different queues if we want this flag to be false (the jobs started failing when using the bot token for the current target queues, probably because we are using the .Open queues?)

natemcmaster: Ok. Can you add this detail in a comment for now? I don't think we have a reason to use anything but .Open queues for tests.

Directory.Build.targets
HaoK: I think so, but do you think we really need different timeouts per test project?  I would imagine something standard like 30min or 1 hour should be plenty, do we really want any tests projects that take longer than that to run?

HaoK:
Results from more target queues is here: https://mc.dot.net/#/user/aspnet-extensions/pr~2Faspnet~2Fextensions/ci/20190107.16

Redhat appears busto on dotnet-install.
There are two unbuntu failures with one being a 'killed' dotnet-install hang?  The caching test I believe might just be flaky since its passed before

HaoK:
Nice the helix fix works, test failures now fail the build (I tried it on the EF repo: https://dnceng.visualstudio.com/public/_build/results?buildId=67638)

I'll redisable the localization tests, and update the sdk here

HaoK:
@natemcmaster Kicking off a helix run locally is just `build.cmd /t:Helix` no PAT needed or anything anymore so its pretty straight forward, should I just add it to one of the internal wikis?  I think linking to https://github.com/dotnet/arcade/blob/master/Documentation/AzureDevOps/SendingJobsToHelix.md is probably a good enough starting point for more info about helix

natemcmaster:
Can you add to this doc? https://github.com/aspnet/Extensions/blob/master/docs/build-from-source.md

Directory.Build.targets
natemcmaster: if we don't specify anything, what is the default? I'm okay if we don't add this knob until we need it.

natemcmaster:
It looks like all RedHat 6 tests failure with:

> Couldn't find a valid ICU package installed on the system. Set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support.

Since tests pass on the RedHat 7 image, which has ICU installed, I'm OK if we just remove RedHat 6.

natemcmaster:
> There are two unbuntu failures with one being a 'killed' dotnet-install hang? 

dotnet-install.sh is notoriously flaky on Linux. Let's try this:

1. Add a retry around the dotnet-install.sh call
2. Remove the `--no-cdn` flag.

Directory.Build.targets
HaoK: Default if we don't specify anything looks like only 5 minutes, https://github.com/dotnet/arcade/blob/bd3e6b007d9a4b5d8926343e79e538dc2a52d7e5/src/Microsoft.DotNet.Helix/JobSender/WorkItemDefinition.cs#L21

That seems kinda short but we could try it if we want?

HaoK:
Added a section to the doc and removed redhat 6

natemcmaster:
Can you merge in the latest release/2.1 changes, which dropped netcoreapp2.0? I want to see how long tests take on Helix.

HaoK:
@natemcmaster hard to say for sure what average numbers would be long term, but after rebasing total time is down to 25 min for a helix run, was 36 min last night, so that squares with 30% drop...

HaoK:
Everything looks to be green or skipped with the exception of a few flaky caching tests, I'll kick off a few more runs to see what the bad boys are and disable them, so far I've seen:

AbsoluteExpirationExpiresInBackground (Debian)
DoNotAddIfSizeOverflows (Windows, Debian)
AddingReplacementWhenTotalSizeExceedsCapacityDoesNotUpdateRemovesOldEntryAndTriggersCompaction(Ubuntu, Debian)

natemcmaster:
25 min doesn't seem too bad. I imagine this will get faster if Helix implements de-duping on test payloads, right?

I expect using Helix is going to make flaky tests even more obvious in PRs. By running on 10 different OSes, you increase the probability of tests failing on PRs.

HaoK:
Yeah possibly, I'm not sure how much time is spent on the CI uploading the payloads, its definitely still a sizeable chunk though as sometimes many of the early queues are done completely before the tests for the later ones are even finished uploading yet, so if we wanted to, we could speed things up by splitting the job into two that published in parallel to half the queues each, or I could even try adding a ci job per helix queue as an experiment to see if that is the fastest overall run time?

HaoK:
Like ideally we could 
1. Build/publish everything
2. publish to each helix queue in parallel using the publish dirs (vs today's publish to each helix queue sequentially)

Although just running this would add whatever the build time to the front of each which maybe isn't too bad I'll try it just to see
```
./build.cmd /t:Helix /p:HelixQueue=WIndows.Amd64.Open
./build.cmd /t:Helix /p:HelixQueue=OSX10.Open
...etc
```

HaoK:
FYI second run with failing tests disabled only took 16 minutes

natemcmaster:
16-25 minutes is still not bad for running a few thousand test on 10 operating systems. I wish their was more visibility into the status of Helix tests. Combing thru the build log to find the link to Mission Control is annoying.

HaoK:
Totally agree, and then clicking thru the MC links is another 3-4 slowww clicks to actually get to the actual failing queues.  I thought they were eventually going to move to the azure pipeline reporting tabs right?

natemcmaster:
They demoed some prototypes back in November. I don't know what the timeline is for making it a real thing.

natemcmaster:
Closing. This was a really useful experiment in using Helix for the first time. We discussed as a team, and don't plan to run tests on Helix for this repo. The distribute testing and platform coverage isn't really necessary and doesn't warrant the overhead at this time.

PR #629, opened 2018-12-11
Add spec tests for external container implementations

pakrym:
So when there is a request to add a new feature to spec tests we can easily verify against external container implementations.

This list is a combination between README.md from this project and @jbogard's MediatR samples (https://github.com/jbogard/MediatR/tree/master/samples)

Some implementations do not pass all the tests so I've added a way to skip test methods per container. It's not ideal, but works.

I'll try to update versions every time new spec test is added but feel free to send PRs if you have a new release.

### DIFF
--- a/src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
+++ b/src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
@@ -10,7 +10,7 @@
     <Reference Include="Newtonsoft.Json" />
 
     <Reference Include="Microsoft.Extensions.DependencyInjection" />
-    <ProjectReference Include="..\..\DI.Specification.Tests\src\Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj" />
+    <Reference Include="Microsoft.Extensions.DependencyInjection.Specification.Tests" />
 
     <Reference Include="Autofac.Extensions.DependencyInjection" />
     <Reference Include="DryIoc.Microsoft.DependencyInjection" />
--- a/src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
+++ b/src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
@@ -9,11 +9,17 @@
   <ItemGroup>
     <Reference Include="Newtonsoft.Json" />
 
+    <ProjectReference Include="..\..\DI\src\Microsoft.Extensions.DependencyInjection.csproj" />
     <ProjectReference Include="..\..\DI.Specification.Tests\src\Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj" />
 
     <Reference Include="Autofac.Extensions.DependencyInjection" />
     <Reference Include="DryIoc.Microsoft.DependencyInjection" />
-
+    <Reference Include="Lamar.Microsoft.DependencyInjection" />
+    <Reference Include="LightInject.Microsoft.DependencyInjection" />
+    <Reference Include="StructureMap.Microsoft.DependencyInjection" />
+    <Reference Include="Grace.DependencyInjection.Extensions" />
+    <Reference Include="Stashbox.Extensions.Dependencyinjection" />
+    <Reference Include="Unity.Microsoft.DependencyInjection" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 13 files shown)


natemcmaster:
cc @Eilon for OSS approval.

src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
```diff
@@ -0,0 +1,29 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
+    <RootNamespace>Microsoft.Extensions.DependencyInjection</RootNamespace>
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="Newtonsoft.Json" />
+
+    <ProjectReference Include="..\..\DI\src\Microsoft.Extensions.DependencyInjection.csproj" />
```
natemcmaster:
```suggestion
    <Reference Include="Microsoft.Extensions.DependencyInjection" />
```

natemcmaster: Use `<Reference>` whenever possible.

src/DependencyInjection/DI.External.Tests/test/SkipableDependencyInjectionSpecificationTests.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.Extensions.DependencyInjection.Specification
+{
+    public abstract class SkipableDependencyInjectionSpecificationTests: DependencyInjectionSpecificationTests
```
tillig: Typo - "skippable" has two p

tillig:
How will container owners get notified if a new spec test doesn't pass? We run these for Autofac, too, and I'm excited to see the double-check here, but if something gets discovered I'd love to hear about it as it happens. Maybe some feed to subscribe to or something in the future?

pakrym:
@tillig we, in general, won't add any new feature specification test that breaks existing implementations until their authors acknowledge and agree to support the feature.

For spec tests that uncover bugs, we are going to merge and notify authors.

I also was thinking about repeating cc'ing everyone in https://github.com/aspnet/Extensions/pull/629#issuecomment-446401943 for all spec changes that go in.

src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
+    <RootNamespace>Microsoft.Extensions.DependencyInjection</RootNamespace>
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="Newtonsoft.Json" />
+
+    <Reference Include="Microsoft.Extensions.DependencyInjection" />
+    <ProjectReference Include="..\..\DI.Specification.Tests\src\Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj" />
```
natemcmaster:
```suggestion
    <Reference Include="Microsoft.Extensions.DependencyInjection.Specification.Tests" />
```

One more

pakrym:
What exactly are you worried about?

jbogard:
When I was looking at container feature support for my library, I had to write these kinds of tests myself.

It's not fun, containers did not always update in a timely basis, and just trying to understand how to fix registration problems either put the onus on the container owners, or myself.

I don't have a great alternative though - if I want to add a feature to the spec, like the constraints thing, how do I know what the bar is? If we have a list of containers that are to be targeted for support, like a blessed list of supported containers, then at least we can hone in on some consensus. At the very least an easy way for container authors to be able to verify against the spec would be very helpful. I don't thin that's easy today.

ENikS:
I do not think feature support should be a problem. I would add new features and tests for them freely and publish resulting matrix per container. It could be a good starting point for developers to evaluate container implementations. Breaking existing code is bad but new stuff should be OK. Just because one of the containers does not implement feature it should not be omitted from library.

src/DependencyInjection/DI.External.Tests/test/Microsoft.Extensions.DependencyInjection.ExternalContainers.Tests.csproj
pakrym: Oh, so we do it for tests projects too?

natemcmaster: You're right about using ProjectRef to other test projects, but Microsoft.Extensions.DependencyInjection.Specification.Tests is actually a class library of tests.

pakrym:
Got OSS approval from @Eilon.

PR #630, opened 2018-12-11
Fix running published analyzer tests

pakrym:
We were resolving only first assembly from compilation libraries instead of all of them.
Fixes: https://github.com/aspnet/Extensions/issues/625


natemcmaster:
I'm not sure. @pakrym would know best if this would apply to 2.1 as well.

pakrym:
@natemcmaster it does apply. I'll retarget

Issue #612, opened 2018-12-10
Add overload of UseServiceProviderFactory that takes a delegate


davidfowl:
If you want to configure the DI factory based on the hosting environment the current methods makes it pretty impossible to do so. Unlike the other methods that take a delegate, these don't compose very well and the last delegate wins. 

The WebHostBuilder implementation is currently sorta hacky and likely broken in subtle ways because of this:

https://github.com/aspnet/AspNetCore/blob/926034ac4eed763cf7baf4acaeccba3272738d46/src/Hosting/Hosting/src/GenericHost/GenericWebHostBuilder.cs#L197-L212

API proposal (breaking API change)

```C#
public interface IHostBuilder
{
    IHostBuilder UseServiceProviderFactory<TContainerBuilder>(Func<HostBuilderContext, IServiceProviderFactory<TContainerBuilder>> factory);
}
```

```c#
var host = new HostBuilder()
                      .UseServiceProviderFactory((context) =>
                      {
                             var options = new ServiceProviderOptions() { ValidateScopes = context.HostingEnvironment.IsDevelopment() };
                             return new DefaultServiceProviderFactory(options);
                      });
```

davidfowl:
cc @Tratcher

davidfowl:
When you add the API fix the web host as well.

Issue #597, opened 2018-12-07
Create opinionated meta package for extensions


davidfowl:
We'd like to create a meta package that contains a `Host.CreateDefaultBuilder` with a set of opinionated defaults. Thinking about this more I think we should keep the good name for the meta package **Microsoft.Extensions.Hosting** and make a new **Microsoft.Extensions.Hosting.Core** that contains what Hosting is today. 

cc @glennc @rynowak @Tratcher

rynowak:
metapackage of shared framework :trollface:

Tratcher:
I'm not convinced we actually need a new assembly/package in this case. It was required for WebHost because there was no other package that could pull in Hosting, servers, middleware, etc. In the generic host scenario it's only pulling in lower level components. E.g. these are the dependencies that might need to be added, none of which are onerous:
Microsoft.Extensions.Configuration.EnvironmentVariables
Microsoft.Extensions.Configuration.CommandLine
Microsoft.Extensions.Configuration.Json
Microsoft.Extensions.Configuration.UserSecrets
Microsoft.Extensions.Logging.Console
Microsoft.Extensions.Logging.Debug
Microsoft.Extensions.Logging.EventSource

Including Console and CommandLine makes sense as we already have a built in ConsoleLifetime. UserSecrets seems like the only very optional feature.

davidfowl:
We'll need user secrets to be in by default as well so unless anyone has any objections. These packages will be added to the default Microsoft.Extensions.Hosting.

PR #637, opened 2018-12-12
Add Host.CreateDefaultBuilder and some dependencies

Tratcher:
#597 Host.CreateDefaultBuilder
#612 UseServiceProviderFactory overloads (breaking interface addition) 👍 👎 ?
https://github.com/aspnet/Hosting/issues/1418 UseDefaultServiceProvider
~~Ensure ApplicationName has a default value.~~
Moved the Internal.Host tests to the internal folder due to name conflicts.

### DIFF
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -1,0 +1,68 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Reflection;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public static class Host
+    {
+        public static IHostBuilder CreateDefaultBuilder() =>
+            CreateDefaultBuilder(args: null);
+
+        public static IHostBuilder CreateDefaultBuilder(string[] args)
+        {
+            var builder = new HostBuilder();
+
+            builder.UseContentRoot(Directory.GetCurrentDirectory());
+            builder.ConfigureHostConfiguration(config =>
+            {
+                config.AddEnvironmentVariables(); // TODO: Prefix?
+                if (args != null)
+                {
+                    config.AddCommandLine(args);
+                }
+            });
+
+            builder.ConfigureAppConfiguration((hostingContext, config) =>
+            {
+                var env = hostingContext.HostingEnvironment;
+
+                config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
+                      .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: true);
+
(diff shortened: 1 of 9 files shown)


JunTaoLuo:
tests failed

davidfowl:
I think the breaking change is fine here as it will only break implementors. We made a similar change to the header collection in 2.0 adding content length. 

If the bar is higher then we can add a new interface and do ilight up using an extension method.

Tratcher:
Yeah, I have no expectation this one will break any consumers.

src/Hosting/Hosting/src/Host.cs
```diff
@@ -0,0 +1,68 @@
...
+    public static class Host
+    {
+        public static IHostBuilder CreateDefaultBuilder() =>
+            CreateDefaultBuilder(args: null);
+
+        public static IHostBuilder CreateDefaultBuilder(string[] args)
+        {
+            var builder = new HostBuilder();
+
+            builder.UseContentRoot(Directory.GetCurrentDirectory());
```
davidfowl: Do we need the check we have in the other logic for the env var override?

src/Hosting/Hosting/src/Internal/ServiceFactoryAdapter.cs
```diff
@@ -9,20 +9,42 @@ namespace Microsoft.Extensions.Hosting.Internal
...
+        private readonly Func<HostBuilderContext> _contextResolver;
+        private Func<HostBuilderContext, IServiceProviderFactory<TContainerBuilder>> _factoryResolver;
 
         public ServiceFactoryAdapter(IServiceProviderFactory<TContainerBuilder> serviceProviderFactory)
         {
-            _serviceProviderFactory = serviceProviderFactory ?? throw new System.ArgumentNullException(nameof(serviceProviderFactory));
+            _serviceProviderFactory = serviceProviderFactory ?? throw new ArgumentNullException(nameof(serviceProviderFactory));
+        }
+
+        public ServiceFactoryAdapter(Func<HostBuilderContext> contextResolver, Func<HostBuilderContext, IServiceProviderFactory<TContainerBuilder>> factoryResolver)
```
davidfowl: Why a func for the HostBuilderContext? Why not just pass it directly?

src/Hosting/Hosting/src/Host.cs
Tratcher: No, Host builder doesn't include env var config by default. We add it afterwards and it overrides.

src/Hosting/Hosting/src/Internal/ServiceFactoryAdapter.cs
Tratcher: Because HostBuilderContext isn't created until you call Build and this gets called first.

src/Hosting/Hosting/src/Host.cs
davidfowl: OK just make sure the scrnaeio fixed is indeed fixed here as well.

src/Hosting/Hosting/src/Internal/ServiceFactoryAdapter.cs
davidfowl: I see.

PR #640, opened 2018-12-13
Add LoggerFactory.Create method

pakrym:
Adds the `LoggerFactory.Create` that simplifies LoggerFactory creation in cases where dependency injection is not used in the application.

So the pattern becomes:

```C#
var loggerFactory = LoggerFactory.Create(builder => 
    builder.AddEventSourceLogger()
           .AddConsole()
           .AddFilter(...));
```

Fixes: https://github.com/aspnet/Extensions/issues/615

### DIFF
--- a/src/Logging/Logging/src/LoggerFactory.cs
+++ b/src/Logging/Logging/src/LoggerFactory.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Extensions.Logging
         /// </summary>
         public static ILoggerFactory Create(Action<ILoggingBuilder> configure)
         {
-            IServiceCollection serviceCollection = new ServiceCollection();
+            var serviceCollection = new ServiceCollection();
             serviceCollection.AddLogging(configure);
             var serviceProvider = serviceCollection.BuildServiceProvider();
             var loggerFactory = serviceProvider.GetService<ILoggerFactory>();
@@ -222,7 +222,7 @@ namespace Microsoft.Extensions.Logging
 
         private class DisposingLoggerFactory: ILoggerFactory
         {
-            private ILoggerFactory _loggerFactory;
+            private readonly ILoggerFactory _loggerFactory;
 
             private readonly ServiceProvider _serviceProvider;
 
@@ -234,7 +234,6 @@ namespace Microsoft.Extensions.Logging
 
             public void Dispose()
             {
-                _loggerFactory.Dispose();
                 _serviceProvider.Dispose();
             }
 
--- a/src/Logging/Logging/src/LoggerFactory.cs
+++ b/src/Logging/Logging/src/LoggerFactory.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Logging
@@ -41,6 +42,18 @@ namespace Microsoft.Extensions.Logging
 
             _changeTokenRegistration = filterOption.OnChange(RefreshFilters);
             RefreshFilters(filterOption.CurrentValue);
+        }
+
+        /// <summary>
+        /// Creates new instance of <see cref="ILoggerFactory"/> configured using provided <paramref name="configure"/> delegate.
+        /// </summary>
+        public static ILoggerFactory Create(Action<ILoggingBuilder> configure)
+        {
+            IServiceCollection serviceCollection = new ServiceCollection();
+            serviceCollection.AddLogging(configure);
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+            var loggerFactory = serviceProvider.GetService<ILoggerFactory>();
+            return new DisposingLoggerFactory(loggerFactory, serviceProvider);
         }
 
         private void RefreshFilters(LoggerFilterOptions filterOptions)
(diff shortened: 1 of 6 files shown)


src/Logging/Logging/src/LoggerFactory.cs
```diff
@@ -43,6 +44,18 @@ public LoggerFactory(IEnumerable<ILoggerProvider> providers, IOptionsMonitor<Log
             RefreshFilters(filterOption.CurrentValue);
         }
 
+        /// <summary>
+        /// Creates new instance of <see cref="ILoggerFactory"/> configured using provided <paramref name="configure"/> delegate.
+        /// </summary>
+        public static ILoggerFactory Create(Action<ILoggingBuilder> configure)
+        {
+            IServiceCollection serviceCollection = new ServiceCollection();
```
davidfowl: nit: var

src/Logging/Logging/src/LoggerFactory.cs
```diff
@@ -206,5 +219,34 @@ private struct ProviderRegistration
...
+
+            public DisposingLoggerFactory(ILoggerFactory loggerFactory, ServiceProvider serviceProvider)
+            {
+                _loggerFactory = loggerFactory;
+                _serviceProvider = serviceProvider;
+            }
+
+            public void Dispose()
+            {
+                _loggerFactory.Dispose();
```
davidfowl: Just dispose the container.

src/Logging/Logging/src/Microsoft.Extensions.Logging.csproj
```diff
@@ -10,7 +10,7 @@
     <Compile Include="../../shared/*.cs" />
 
     <Reference Include="Microsoft.Extensions.Configuration.Binder" />
-    <Reference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
+    <Reference Include="Microsoft.Extensions.DependencyInjection" />
```
davidfowl: This is slightly unfortunate. But not the end of the world if I guess.

src/Logging/Logging/src/LoggerFactory.cs
```diff
@@ -206,5 +219,34 @@ private struct ProviderRegistration
             public ILoggerProvider Provider;
             public bool ShouldDispose;
         }
+
+        private class DisposingLoggerFactory: ILoggerFactory
+        {
+            private ILoggerFactory _loggerFactory;
```
davidfowl: readonly

src/Logging/Logging/src/LoggerFactory.cs
```diff
@@ -206,5 +219,34 @@ private struct ProviderRegistration
             public ILoggerProvider Provider;
             public bool ShouldDispose;
         }
+
+        private class DisposingLoggerFactory: ILoggerFactory
```
davidfowl: The fact that we need to hold onto the container means all of the sample code is sorta busted

pakrym: Busted how?

davidfowl: Maybe you’re saying we don’t need to dispose the service provider?

davidfowl: Why does this wrapper exist to begin with?

pakrym: To dispose ILoggerProviders created by DI

davidfowl: Doesn’t disposing the factory do that?

davidfowl: Ah, I see it wouldn’t. That makes the current usage without this API leaky.

pakrym: If you don't dispose provider - yes.

davidfowl: I’m taking about examples that you would use today (even in our samples)

pakrym: Pushed a sample fix.

src/Logging/Logging/src/Microsoft.Extensions.Logging.csproj
ryanbrandenburg: It seems that this change caused [these failures](https://dnceng.visualstudio.com/public/_build/results?buildId=57771&view=logs&jobId=f0f0520d-467b-5d1d-a95d-4e3a2581fe8a&taskId=859aa418-c131-516f-eebe-b28b785cce53&lineStart=39&lineEnd=42&colStart=1&colEnd=234) in my PR build. It actually [showed up in your PR check too](https://dnceng.visualstudio.com/public/_build/results?buildId=57675&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&taskId=3d94a178-cdfb-5c90-a971-5deeccbbe043&lineStart=33&lineEnd=34&colStart=1&colEnd=222), but as a warning instead of an error.

@natemcmaster 1. Is the proper solution here (assuming it is an approved breaking change) to [edit this line](https://github.com/aspnet/Extensions/blob/82ad9d6edda284321a10e633cd2863852acf34fe/eng/Baseline.Designer.props#L386)? 2. Should this kind of change be a warning or an error?

natemcmaster: @ryanbrandenburg does your PR invoke build with `--warnaserror`?

ryanbrandenburg: Yes, that's what [Arcade defaults to](https://github.com/dotnet/arcade/blob/d7c9d455f3de1edd96618a1e5ab6a668ad138e02/eng/common/tools.ps1#L27). We can certainly override that anywhere (or everywhere), but I thought I'd ask since this seems like the kind of thing we might want to fail a build (even if it's just so we can document that this was an intended break).

PR #798, opened 2018-12-18
Bring consistency to handling of nulls in configuration sources

ajcvickers:
Fixes #685

In the current code base, the following is true:
* Where null values can be represented in the config (e.g. JSON) they are put into the configuration like any other value. Specifically, no provider skips null values are treats them the same as missing values.
* Some providers convert the null values to empty strings; some don't.

This change changes those providers (notably JSON) that convert nulls to strings to stop doing this. This is means that where nulls can be represented in the configuration source, they are correctly represented in the resultant configuration. Also added TryGet methods so that, for those who care, it is possible to determine the difference at the top level between a null value and a not-set value. (Note this was always possible, but required traversing the configuration structure and looking at configuration sections/values directly.)

This is technically a breaking change since in some cases a configuration value that was returned as an empty string will now be returned as null. However, this seems like a good change to make because it both matches common customer expectations about how it should work, and also allows better handling in our binder and other binders that people may have.


src/Configuration/Config.Xml/test/ConfigurationProviderXmlTest.cs
```diff
@@ -20,6 +20,8 @@ public override void Combine_after_other_provider()
             // Disabled test due to XML handling of empty section.
         }
 
+        protected override bool CanRepresentNulls => false;
```
HaoK: So the fact that not all providers can represent nulls make me wary of making a change like this...

HaoK:
So summarizing the main difference after this PR: its no longer true that `Get(string) = null` means no configuration provider defined the key, since some providers allow representing null.  Instead users would need to use the new `bool TryGet()` method instead of relying on `Get == null`  [The existing Exists method would still work the same].

I didn't completely grok the implications from our initial discussions this week. Prior to this change, I believe configuration always tried to normalize null (to be empty string).  My main concern is that there's some danger that we might end up in a halfway state if we allow nulls now.  (If consistency is the main concern, I think it would be safer to go the other direction, normalize nulls to empty string consistently everywhere)

At the least maybe this should be gated behind a setting specified in the ConfigurationBuilder.  I don't think its clear that this would always be the preferred behavior for everyone, I can see how some people would want this behavior, but seems like it would be safer to be opt-in when building configuration...?

`configuration.AddJson().Build(allowNullValues: true) // default is false like today`

ajcvickers:
@HaoK In general, `Get(string) = null` doesn't mean that the key is not defined since some providers (notably InMemory) put nulls in config, and config for any key in any provider can be set to null. If you don't care about the difference, then you can still do this.

The InMemory provider does not normalize, and setting a value to null is possible for any provider. If we make all providers normalize the other way, then it's still a breaking change, and unless we make further breaking changes (for example, to how setting values works) then it will still be possible to put nulls in the config. 

Another thing to note is that nulls in the config behave differently to no value in the config. An actual value from a later provider in the config will never be used if there is a null before it, because the TryGet internally returns true and we stop. Basically I think we have to either accept that nulls are part of the configuration story (which happens to be what people want anyway) or we need to go back any make it like it was apparently intended to be (at least as far as Diego's comments suggest) where nulls really mean the same as no key set. But that would be a much bigger breaking change.

With regard to the flag, the only thing it would control (that is the only functional change in product code in this PR) is to continue converting nulls to empty strings in JSON. But if that's the behavior you want, then why not just put empty strings in your JSON instead of nulls?

HaoK:
Okay I know we just added Remove in the other PR, but I think I'm coming around to Diego's POV now.

What if we just instead add a compat flag to ConfigurationBuilder to ignoreNullValues: true, and get rid of Remove/TryGet and inside the main TryGet for ConfigurationRoot ignore nulls consistently when this flag is set so nulls would be consistently be nothing, so it would also be how to delete keys as well.

We can still fix the Json behavior that's converting json null => "", perhaps in another custom flag in JsonSource (convertNullToEmptyString: true)?

This seems like a reasonable simple way to consistently enforce that nulls will be ignored by Get, they could still enumerate the children directly if they wanted to find keys with null values as well.

ajcvickers:
I didn't actually send out the Remove PR. :-) At least, not intentionally. Only the DebugView part.

The main reason is so that nulls can have useful semantics, such as "this has been explicitly set to null here, even if some other previous provider has a value for it." And also, the information can be used when binding to options, such that a missing key can mean don't touch the property in the options, because nothing in config set it, while a null value means set the options property to null. I'm going to suggest we do this in our binder, but even if you shoot that down, doing it here still allows people to do it in their binding code.

ajcvickers:
This is what happens when I start to read code. I start being a pain. :-)

HaoK:
I guess I'm less concerned about options because at that point you can just manipulate the options class directly and null out whatever properties you want after binding has occurred.  

So the new scenario that allowing nulls in config would enable is the ability for one provider to effectively delete a key that another previous provider provided?

That still seems better done as a post bind/configure step where you just set options.SomeValue = null explicitly in code, as opposed to playing games in config.

src/Configuration/Config.Abstractions/src/IConfiguration.cs
```diff
@@ -18,6 +18,14 @@ public interface IConfiguration
         /// <returns>The configuration value.</returns>
         string this[string key] { get; set; }
 
+        /// <summary>
+        /// Tries to get a configuration value for the specified key.
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="value">The value.</param>
+        /// <returns><c>True</c> if a value for the specified key was found, otherwise <c>false</c>.</returns>
+        bool TryGet(string key, out string value);
```
davidfowl: How breaking do you think this will be?

ajcvickers: @davidfowl The interface break? Minimal. I don't think Hao is pushing back on this. The breaking change push back is that previously any null value in the JSON is converted to an empty string, and so shows up in config as an empty string, while after the change it will show up in config as null, like it was in the JSON.

ajcvickers:
Closing this, as we could not come to agreement on the break, and I now have no time available to work on it.

dorki:
@ajcvickers 
Hi, any chance we can rethink this?

krispenner:
I would also like to see this revisited, either of the proposed changes would be better than converting null to an empty string as to me that makes little sense. If I wanted an empty string I would just set it to an empty string.

1. Treat null as if the property was never in the config, allowing other providers further up to be applied.

1. Treat null as a set value of null on the property, preventing other providers further up from being applied.

PR #855, opened 2018-12-21
Clean up event handlers and other subscriptions

davidfowl:
- Also removed the closure when logging
- Captured _applicationStartedRegistration though it shouldn't strictly be necessary since the CTS *should* be disposed when the host is disposed

This change was inspired by looking at memory dumps and seeing this:

image

Maybe worth doing in the WebHost as well (but we're moving away from that)

### DIFF
--- a/src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
+++ b/src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
@@ -84,8 +84,8 @@ namespace Microsoft.Extensions.Hosting.Internal
         {
             _shutdownBlock.Set();
 
-            AppDomain.CurrentDomain.ProcessExit += OnProcessExit;
-            Console.CancelKeyPress += OnCancelKeyPress;
+            AppDomain.CurrentDomain.ProcessExit -= OnProcessExit;
+            Console.CancelKeyPress -= OnCancelKeyPress;
 
             _applicationStartedRegistration.Dispose();
         }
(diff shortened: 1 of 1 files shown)


src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
```diff
@@ -72,6 +83,11 @@ public Task StopAsync(CancellationToken cancellationToken)
         public void Dispose()
         {
             _shutdownBlock.Set();
+
+            AppDomain.CurrentDomain.ProcessExit += OnProcessExit;
```
JunTaoLuo: Why are these added again?

davidfowl: Wow fail, should be -, copy pasta

JunTaoLuo: That would make a bit more sense.

davidfowl: 🆙 📅

JunTaoLuo:
No issue filed?

davidfowl:
@JunTaoLuo I remember one from a while back but I can't find it. Funnily enough, I'm trying to reproduce the problem and found another more severe leak..

davidfowl:
After:

image

davidfowl:
Turns out it was holding onto more state that I originally recognized.

PR #867, opened 2018-12-24
Fix aspnet/Extensions#639

rynowak:
The issue is that the hosted service would not be registered if other
hosted services are present. The fix is to use TryAddEnumble so that we
allow multipled `IHostedService`s but not multiple instances of
the same type.

### DIFF
--- a/src/HealthChecks/HealthChecks/src/DependencyInjection/HealthCheckServiceCollectionExtensions.cs
+++ b/src/HealthChecks/HealthChecks/src/DependencyInjection/HealthCheckServiceCollectionExtensions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Extensions.DependencyInjection
         public static IHealthChecksBuilder AddHealthChecks(this IServiceCollection services)
         {
             services.TryAddSingleton<HealthCheckService, DefaultHealthCheckService>();
-            services.TryAddSingleton<IHostedService, HealthCheckPublisherHostedService>();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, HealthCheckPublisherHostedService>());
             return new HealthChecksBuilder(services);
         }
     }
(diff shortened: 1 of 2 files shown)


src/HealthChecks/HealthChecks/src/DependencyInjection/HealthCheckServiceCollectionExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static class HealthCheckServiceCollectionExtensions
         public static IHealthChecksBuilder AddHealthChecks(this IServiceCollection services)
         {
             services.TryAddSingleton<HealthCheckService, DefaultHealthCheckService>();
-            services.TryAddSingleton<IHostedService, HealthCheckPublisherHostedService>();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, HealthCheckPublisherHostedService>());
```
davidfowl: I wonder if AddHostedService should do this cc @Tratcher

Tratcher: You think people are going to try adding the same hosted service multiple times? It hasn't come up yet.

davidfowl: I guess the question is more about if people will use it to build libraries. If you wanted to use AddHostedService to add your hosted service and make your call idempotent, you have to revert to using this manual call.

rynowak: > I guess the question is more about if people will use it to build libraries.

Infinity times this. Library authors have to know about tryaddenumerable, or they have to throw when called twice.

PR #868, opened 2018-12-24
Fix memory leak with OptionsMonitor

### LINKED ISSUE #860: OptionsMonitor does not clean up ChangeToken.OnChange registrations
Leads to leakages when the change token isn't disposed across usages:

```C#
using System;
using System.Threading.Tasks;
using Microsoft.Extensions.Options;
using Microsoft.Extensions.Primitives;
using System.Threading;

namespace WebApplication45
{
    public class Program
    {
        private static CancellationTokenSource s_CancellationTokenSource = new CancellationTokenSource();

        public static async Task Main(string[] args)
        {
            Console.WriteLine("STARTING");
            Console.ReadLine();
            CreateOptions();
            Console.WriteLine("DONE");
            Console.ReadLine();
        }

        private static void CreateOptions()
        {
            for (int i = 0; i < 1000; i++)
            {
                services.AddOptions();
                services.AddSingleton<IOptionsChangeTokenSource<MyOptions>>(new MyOptionsChangeToken(s_CancellationTokenSource.Token));
                using (var sp = services.BuildServiceProvider())
                {
                    var monitor = sp.GetRequiredService<IOptionsMonitor<MyOptions>>();
                    using (monitor.OnChange(o => { }))
                    {

                    }
                }
            }
        }


        public class MyOptions
        {

        }

        public class MyOptionsChangeToken : IOptionsChangeTokenSource<MyOptions>
        {
            private readonly CancellationChangeToken _cancellationChangeToken;
            public MyOptionsChangeToken(CancellationToken cancellationToken)
            {
                _cancellationChangeToken = new CancellationChangeToken(cancellationToken);
            }

            public string Name => null;

            public IChangeToken GetChangeToken() => _cancellationChangeToken;
        }
    }
}
```

This is what you see after a GC:

image

There are all rooted by the static cancellation token.

davidfowl:
- Made OptionsMonitor IDisposable and remove all registrations on disposal.

Fixes #860

### DIFF
--- a/src/Options/Options/src/OptionsMonitor.cs
+++ b/src/Options/Options/src/OptionsMonitor.cs
@@ -11,11 +11,12 @@ namespace Microsoft.Extensions.Options
     /// Implementation of IOptionsMonitor.
     /// </summary>
     /// <typeparam name="TOptions"></typeparam>
-    public class OptionsMonitor<TOptions> : IOptionsMonitor<TOptions> where TOptions : class, new()
+    public class OptionsMonitor<TOptions> : IOptionsMonitor<TOptions>, IDisposable where TOptions : class, new()
     {
         private readonly IOptionsMonitorCache<TOptions> _cache;
         private readonly IOptionsFactory<TOptions> _factory;
         private readonly IEnumerable<IOptionsChangeTokenSource<TOptions>> _sources;
+        private readonly List<IDisposable> _registrations = new List<IDisposable>();
         internal event Action<TOptions, string> _onChange;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


HaoK:
We probably have a similar leak in configuration as well here https://github.com/aspnet/Extensions/blob/d7f8e253d414ce6053ad59b6f974621d5620c0da/src/Configuration/Config.FileExtensions/src/FileConfigurationProvider.cs#L32

HaoK:
We should probably make ConfigurationRoot disposable and all the providers so the root can dispose all the providers?

davidfowl:
Yep, see https://github.com/aspnet/Extensions/issues/861.

PR #869, opened 2018-12-24
Made changes to ChangeToken.OnChange

### LINKED ISSUE #558: ChangeToken.OnChange can not be detached
### Describe the bug
I am attempting to monitor changes to configuration but I noticed that I once I am monitoring configuration changes using `ChangeToken.OnChange`, I cannot stop monitoring.

### To Reproduce
Consider the following sample...

```c#
class Program
{
	static void Main() {
		var configurationRoot = new ConfigurationBuilder().Build();
		var disposable = ChangeToken.OnChange(configurationRoot.GetReloadToken, () => Console.WriteLine("Configuration changed"));

		configurationRoot.Reload();

		Console.ReadKey();
		
		configurationRoot.Reload();

		disposable.Dispose();
		Console.ReadKey();
		
		configurationRoot.Reload();
	}
}
```

Notice that running this will always print `Configuration changed` even after the `IDisposable` has been disposed of. 

### Expected behavior
I would expect that once the `IDisposable` returned by `ChangeToken.OnChange` is disposed, then it is no longer receiving change notifications. I believe that the issue is due to the fact that `ChangeToken.OnChange` does not [re-capture](https://github.com/aspnet/Extensions/blob/master/src/Primitives/src/ChangeToken.cs#L45) the `IDisposable` returned by `RegisterChangeCallback` when attempting to re-register the change notification. So the `IDisposable` that is returned is only valid while the current token is alive. In the case of `ConfigurationRoot`, this token is [recreated](https://github.com/aspnet/Configuration/blob/master/src/Config/ConfigurationRoot.cs#L128) after every change notification.

As an aside, shouldn't the implementation of `ConfigurationReloadToken` be more similar to the following to ensure that `ExecutionContext` is suppressed when registering the change callback, which already occurs in [CancellationChangeToken](https://github.com/aspnet/Extensions/blob/master/src/Primitives/src/CancellationChangeToken.cs#L32).

```c#
public class ConfigurationReloadToken : CancellationChangeToken
{
	private readonly CancellationTokenSource _cts;

	public ConfigurationReloadToken() : this(new CancellationTokenSource())
	{
	}

	private ConfigurationReloadToken(CancellationTokenSource cts) : base(cts.Token)
	{
		_cts = cts;
	}

	public void OnReload() => _cts.Cancel();
}
```

davidfowl:
- Introduced ChangeTokenRegistration to reduce allocations and managed the the registration to the underlying change token.
- Support dispose working if the token fires more than once

cc @stephentoub to check if my lock free code is BS 😄 

Fixes #558

### DIFF
--- a/src/Primitives/src/ChangeToken.cs
+++ b/src/Primitives/src/ChangeToken.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
+using System.Threading;
 
 namespace Microsoft.Extensions.Primitives
 {
@@ -27,26 +29,7 @@ namespace Microsoft.Extensions.Primitives
                 throw new ArgumentNullException(nameof(changeTokenConsumer));
             }
 
-            Action<object> callback = null;
-            callback = (s) =>
-            {
-                // The order here is important. We need to take the token and then apply our changes BEFORE
-                // registering. This prevents us from possible having two change updates to process concurrently.
-                //
-                // If the token changes after we take the token, then we'll process the update immediately upon
-                // registering the callback.
-                var t = changeTokenProducer();
-                try
-                {
-                    changeTokenConsumer();
-                }
-                finally // We always want to ensure the callback is registered
-                {
-                    t.RegisterChangeCallback(callback, null);
-                }
-            };
-
-            return changeTokenProducer().RegisterChangeCallback(callback, null);
+            return new ChangeTokenRegistration<object>(changeTokenProducer, state => ((Action)state)(), changeTokenConsumer);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Primitives/src/ChangeToken.cs
```diff
@@ -67,26 +50,91 @@ public static IDisposable OnChange<TState>(Func<IChangeToken> changeTokenProduce
...
-            return changeTokenProducer().RegisterChangeCallback(callback, state);
+            private void SetDisposable(IDisposable disposable)
+            {
+                // We don't want to transition from _disposedSentinel => anything since it's terminal
+                // but we want to allow going from previously assigned disposable, to another
+                // disposable. The assumption here is that we don't have to dispose the registration if an IChangeToken
+                // fires.
+                var current = Volatile.Read(ref _disposable);
+
+                var previous = Interlocked.CompareExchange(ref _disposable, disposable, current);
```
rynowak: Don't you also need to check for `current == _disposedSentinel` in between these two lines? it seems like you'd happily transition from `_disposedSentinel` -> something else.

benaadams: Not sure what `previous` and `current` are achieving here?

The `CompareExchange` would prevent something between the two; but doesn't add anything over the single statement:
```csharp
var previous = Interlocked.Exchange(ref _disposable, disposable);
```
So I assume its trying to do something else?

davidfowl: Yes, you're right.

davidfowl: I want to assign _disposable to disposable only if current is not _disposedSentinel

benaadams: e.g. is should the whole method be:
```csharp
var previous = Interlocked.CompareExchange(ref _disposable, disposable, null);

if (previous == _disposedSentinel)
{
    // The subscription was disposed so we dispose immediately and return
    disposable.Dispose();
}
else if (previous != null)
{
    Debug.Fail("Somebody else set the _disposable");
}
```

davidfowl: That’s not what I want because SetDisposable can be called multiple times without calling Dispose which means you can transition from non null disposable to disposable. That’s why I used current instead of null.

src/Primitives/src/ChangeToken.cs
```diff
@@ -67,26 +50,99 @@ public static IDisposable OnChange<TState>(Func<IChangeToken> changeTokenProduce
...
 
-            return changeTokenProducer().RegisterChangeCallback(callback, state);
+                if (previous == _disposedSentinel)
+                {
+                    // The subscription was disposed so we dispose immediately and return
+                    disposable.Dispose();
+                }
+                else if (previous == current)
+                {
+                    // We successfuly assigned the _disposable field to disposable
```
stephentoub: You don't need to dispose of the previous one if it's non-null?

davidfowl: I was assuming once he callback is triggered there’s no need to dispose the previous token (like a CancellationTokenRegistration). Maybe it still make sense to always dispose it.

davidfowl: This seems to break things, I'll revert then investigate

rynowak:
> Support dispose working if the token fires more than once

Is this just fixing a bug? or for safety in the case of a bad implementation? I guess anyone could implement `IChangeToken`, but there's really only one correct way to do it, so it's a kinda silly interface 😆

src/Primitives/src/ChangeToken.cs
```diff
@@ -27,26 +29,7 @@ public static IDisposable OnChange(Func<IChangeToken> changeTokenProducer, Actio
...
-                    changeTokenConsumer();
-                }
-                finally // We always want to ensure the callback is registered
-                {
-                    t.RegisterChangeCallback(callback, null);
-                }
-            };
-
-            return changeTokenProducer().RegisterChangeCallback(callback, null);
+            return new ChangeTokenRegistration<Action>(changeTokenProducer, callback => callback(), changeTokenConsumer);
```
rynowak: this is real spicy

src/Primitives/src/ChangeToken.cs
```diff
@@ -67,26 +50,91 @@ public static IDisposable OnChange<TState>(Func<IChangeToken> changeTokenProduce
...
+                    RegisterChangeTokenCallback(token);
                 }
-            };
+            }
+
+            private void RegisterChangeTokenCallback(IChangeToken token)
+            {
+                var registraton = token.RegisterChangeCallback(s => ((ChangeTokenRegistration<TState>)s).OnChangeTokenFired(), this);
+
+                SetDisposable(registraton);
```
rynowak: Are you concerned at all about the case of a bad `IChangeToken` that doesn't implicitly dispose registrations after it fires? I guess there's not much we could do to mitigate that, just thinking about it since we're only going dispose the 'current' registration.

src/Primitives/src/ChangeToken.cs
rynowak: I don't think you've resolved this... the cex here could result in swapping `_disposedSentinel` out for a `disposable`.

src/Primitives/src/ChangeToken.cs
davidfowl: Yes, but that's another change, my previous attempt to dispose the previous registrations broke configuration so there's something else going on there, that needs to be looked into.

davidfowl:
> Is this just fixing a bug? or for safety in the case of a bad implementation? I guess anyone could implement IChangeToken, but there's really only one correct way to do it, so it's a kinda silly interface 😆

The current logic is just completely broken so the primary thing being fixed is that you can call dispose and it will stop listening to change token callbacks instead of just the first one.

davidfowl:
Ping!

PR #870, opened 2018-12-26
Reduce event source logger allocations

pakrym:
https://github.com/aspnet/Extensions/issues/864

1. Use EventData
1. No-op more
1. Split event id and event name so we don't ToString it all the time

### DIFF
--- a/src/Logging/Logging.EventSource/src/EventSourceLogger.cs
+++ b/src/Logging/Logging.EventSource/src/EventSourceLogger.cs
@@ -191,14 +191,11 @@ namespace Microsoft.Extensions.Logging.EventSource
         {
             if (state is IReadOnlyList<KeyValuePair<string, object>> keyValuePairs)
             {
-                var arguments = new List<KeyValuePair<string, string>>();
-                for (int i = 0; i < keyValuePairs.Count; i++)
+                var arguments = new KeyValuePair<string, string>[keyValuePairs.Count];
+                for (var i = 0; i < keyValuePairs.Count; i++)
                 {
-                    var keyValue = keyValuePairs[i];
-                    if (keyValue.Key != null)
-                    {
-                        arguments.Add(new KeyValuePair<string, string>(keyValue.Key, keyValue.Value?.ToString()));
-                    }
+                    var keyValuePair = keyValuePairs[i];
+                    arguments[i] = new KeyValuePair<string, string>(keyValuePair.Key, keyValuePair.Value?.ToString());
                 }
                 return arguments;
             }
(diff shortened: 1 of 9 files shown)


src/Logging/Logging.EventSource/src/EventSourceLogger.cs
```diff
@@ -190,19 +197,20 @@ private ExceptionInfo GetExceptionInfo(Exception exception)
         /// </summary>
         private IEnumerable<KeyValuePair<string, string>> GetProperties(object state)
         {
-            var arguments = new List<KeyValuePair<string, string>>();
-            var asKeyValues = state as IEnumerable<KeyValuePair<string, object>>;
-            if (asKeyValues != null)
+            if (state is IEnumerable<KeyValuePair<string, object>> keyValuePairs)
             {
```
davidfowl: Is there no way to write the object directly as EventData before turning it into a list of string, string?

pakrym: I suspect there is but it's not easy. I didn't touch methods that take IEnumerable for the first pass.

davidfowl: Ok we need to do that though. I’m ok looking into that next. I think it’s worrhwhile to do.

davidfowl:
Do a before and after for allocations? I’d like to see what this saves. Maybe we should also add a micro benchmark for it.

pakrym:
Before:

```
            Method | HasSubscribers |  Json |        Mean |        Error |       StdDev |        Op/s |  Gen 0 | Allocated |
------------------ |--------------- |------ |------------:|-------------:|-------------:|------------:|-------:|----------:|
 EventSourceLogger |          False | False |    667.5 ns |    12.749 ns |    12.521 ns | 1,498,047.7 | 0.0029 |     752 B |
 EventSourceLogger |          False |  True |    646.7 ns |     3.370 ns |     2.631 ns | 1,546,216.1 | 0.0029 |     752 B |
 EventSourceLogger |           True | False | 74,842.2 ns | 1,475.193 ns | 2,697.473 ns |    13,361.5 |      - |    3464 B |
 EventSourceLogger |           True |  True |  3,512.6 ns |    69.683 ns |    71.560 ns |   284,687.0 | 0.0191 |    4384 B |
```

After:

```

            Method | HasSubscribers |  Json |        Mean |        Error |       StdDev |        Op/s |  Gen 0 | Allocated |
------------------ |--------------- |------ |------------:|-------------:|-------------:|------------:|-------:|----------:|
 EventSourceLogger |          False | False |    199.4 ns |     4.010 ns |     8.886 ns | 5,014,173.1 |      - |      24 B |
 EventSourceLogger |          False |  True |    195.9 ns |     4.114 ns |     5.900 ns | 5,105,384.5 |      - |      24 B |
 EventSourceLogger |           True | False | 71,330.0 ns | 1,405.246 ns | 2,707.432 ns |    14,019.3 |      - |    2640 B |
 EventSourceLogger |           True |  True |  4,879.7 ns |    80.149 ns |    74.971 ns |   204,928.6 | 0.0153 |    3976 B |
```

image

image

PR #878, opened 2018-12-28
Obsolete InplaceStringBuilder

pakrym:
Fixes: https://github.com/aspnet/Extensions/issues/877
Reaction PR: https://github.com/aspnet/AspNetCore/pull/6163


pakrym:
Talked offline @davidfowl is fine with obsoletion.

PR #879, opened 2018-12-28
Validate that implemenation types are assignable to service types

pakrym:
We didn't have an explicit check like this so event with `ValidateOnBuild` nothing would get thrown and resolution fails with type cast exception.

Fixes: https://github.com/aspnet/Extensions/issues/696

### DIFF
--- a/src/DependencyInjection/DI/src/ServiceLookup/ConstantCallSite.cs
+++ b/src/DependencyInjection/DI/src/ServiceLookup/ConstantCallSite.cs
@@ -11,6 +11,11 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 
         public ConstantCallSite(Type serviceType, object defaultValue): base(ResultCache.None)
         {
+            if (defaultValue != null && !serviceType.IsInstanceOfType(defaultValue))
+            {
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/DI/src/ServiceLookup/ConstantCallSite.cs
```diff
@@ -11,6 +11,11 @@ internal class ConstantCallSite : ServiceCallSite
 
         public ConstantCallSite(Type serviceType, object defaultValue): base(ResultCache.None)
         {
+            if (defaultValue != null && !serviceType.IsInstanceOfType(defaultValue))
```
halter73: It might be worth validating that the service type is reference/nullable type when the constant is null. This doesn't feel too different in principle to the validation being added.

davidfowl:
Why not make it so you can't construct bad descriptors?

pakrym:
It would be a breaking change without opt-out.

src/DependencyInjection/DI/src/ServiceLookup/ConstantCallSite.cs
pakrym: We don't support value type services. So I'm not sure it's worth it.

davidfowl:
Isn’t it broken anyways?

pakrym:
No, service might have never been resolved or overridden by another descriptor.

PR #900, opened 2019-01-03
Fix inaccessible commit from removed NuGet/SubModuleIntegration branch

dagood:
Merges e6fac80 into `master` without taking any changes.

This makes the commit itself accessible for NuGet.Client git submodule initialization without making any changes to the files in this repo.

Addresses https://github.com/dotnet/source-build/issues/939 and NuGet CI failures.


natemcmaster:
We added a tag instead. If that doesn't work, I'll reconsider doing this.

PR #902, opened 2019-01-05
Add shorthand for getting connection string "DefaultConnection"

weitzhandler:
Since there is an extension method [`GetConnectionString`](https://github.com/aspnet/Extensions/blob/master/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs#L34L37) that avoids having to use the `"ConnectionStrings"` hard-coded string, why not add another ex. method that would avoid using hard-coded `"DefaultConnection"` too?

### DIFF
--- a/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
+++ b/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Extensions.Configuration
         /// <returns></returns>
         public static string GetDefaultConnectionString(this IConfiguration configuration)
         {
-            return configuration?.GetConnectionString("DefaultConnection");
+            return configuration.GetConnectionString("DefaultConnection");
         }
 
         /// <summary>
--- a/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
+++ b/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Extensions.Configuration
         /// <returns></returns>
         public static string GetDefaultConnectionString(this IConfiguration configuration)
         {
-            return configuration.GetConnectionString("DefaultConnection");
+            return configuration?.GetConnectionString("DefaultConnection");
         }
 
         /// <summary>
--- a/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
+++ b/src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
@@ -37,6 +37,16 @@ namespace Microsoft.Extensions.Configuration
         }
 
         /// <summary>
+        /// Shorthand for GetSection("ConnectionStrings")["DefaultConnection"].
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Configuration/Config.Abstractions/src/ConfigurationExtensions.cs
```diff
@@ -36,6 +36,16 @@ public static string GetConnectionString(this IConfiguration configuration, stri
...
         }
 
+        /// <summary>
+        /// Shorthand for GetSection("ConnectionStrings")["DefaultConnection"].
+        /// </summary>
+        /// <param name="configuration">The configuration.</param>
+        /// <returns></returns>
+        public static string GetDefaultConnectionString(this IConfiguration configuration)
+        {
+            return configuration.GetConnectionString("DefaultConnection");
```
gfoidl:
```suggestion
            return configuration?.GetConnectionString("DefaultConnection");
```

To prevent a NRE (when not called as extension method). Maybe add a test for that too?

weitzhandler:
```suggestion
            return configuration.GetConnectionString("DefaultConnection");
```

Actually, I'm not sure this is required, because even `configuration` will be `null`, the compiler will still link the ex. method properly, and `GetConnectionString` will in turn return `null`.
I'll add a new commit containing a test for this case.

gfoidl: You are right! Smart compiler :smile:

weitzhandler: It's the one thing that might be a bit confusing about ex. methods, they feel like instance methods when they're essentially static ones.

weitzhandler:
@gfoidl why is it failing?

gfoidl:
It's failing because of
```
Microsoft.Extensions.Configuration.UserSecrets.MsBuildTargetTest.GeneratesAssemblyAttributeFile(projectExt: ".fsproj", sourceExt: ".fs")

Assert.Equal() Failure
Expected: 0
Actual:   1

at Microsoft.Extensions.Configuration.UserSecrets.MsBuildTargetTest.AssertDotNet(String args) in /_/src/Configuration/Config.UserSecrets/test/MsBuildTargetTest.cs:line 142
   at Microsoft.Extensions.Configuration.UserSecrets.MsBuildTargetTest.GeneratesAssemblyAttributeFile(String projectExt, String sourceExt) in /_/src/Configuration/Config.UserSecrets/test/MsBuildTargetTest.cs:line 102
```
(from https://dnceng.visualstudio.com/public/_build/results?buildId=68143)

Seems to be unrelated to your change, but I don't know how to restart a build. Someone from the team will do this.

gfoidl:
> Can you ping someone?

They will come here from alone, as they have to review / approve the PR anyway.

ajcvickers:
@weitzhandler We have discussed this several times in the past and always concluded that we will not add any more sugar methods like this, since they assume a well-known and static configuration layout which can be and does change in a number of situations. It's debatable whether `GetConnectionString` was a good thing for the same reasons.

I would suggest that you add the extension method to your own code if is a convention you are using.

PR #905, opened 2019-01-07
Add attribute to allow LoggedTest to collect dump on failure

JunTaoLuo:
Adding support to collect dumps upon test failure. The code was originally used in https://github.com/aspnet/AspNetCore/pull/6177 but now I'm moving it to LoggedTest so it's available to all LoggedTests. Note that the implementation currently only works on Windows.

The original motivation is to use this feature to help figure out what's causing the threadpool starvation for https://github.com/aspnet/AspNetCore-Internal/issues/1370.


eng/Dependencies.props
```diff
@@ -62,6 +62,7 @@
     <LatestPackageReference Include="Microsoft.CodeAnalysis.Common" Version="2.8.0" />
     <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.8.0" />
     <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.8.0" />
+    <LatestPackageReference Include="Microsoft.Diagnostics.Runtime" Version="1.0.2" />
```
davidfowl: You can remove this

src/Logging/Logging.Testing/src/CollectDumpAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    /// <summary>
+    /// Capture the memory dump upon test failure.
+    /// </summary>
+    /// <remarks>
+    /// This currently only works in Windows environments
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
```
JunTaoLuo: Currently setting this to method only since only few tests should need this. Could expandable to class/assembly if we want.

src/Logging/Logging.Testing/src/Microsoft.Extensions.Logging.Testing.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <Reference Include="Microsoft.AspNetCore.Testing" />
+    <Reference Include="Microsoft.Diagnostics.Runtime" />
```
davidfowl: Remove this as well

src/Logging/Logging.Testing/src/DumpCollector/DumpCollector.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static partial class DumpCollector
+    {
+        public static void Collect(Process process, string fileName)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                Windows.Collect(process, fileName);
+            }
+            // No implementations yet for macOS and Linux
```
JunTaoLuo: Warrants a follow up issue.

src/Logging/Logging.Testing/src/DumpCollector/DumpCollector.WIndows.cs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
Tratcher: WIndows?

BrennanConroy:
Should this just be on by default?

Tratcher:
@BrennanConroy no. There's such a thing as too much information.

PR #916, opened 2019-01-09
Use runtime feature detection when selecting DI mode

pakrym:
I've added another value for `ServiceProviderMode` so we can test `Dynamic` mode explicitly no matter which runtime flags are set.

Fixes: https://github.com/aspnet/Extensions/issues/422

### DIFF
--- a/src/DependencyInjection/DI/src/ServiceProvider.cs
+++ b/src/DependencyInjection/DI/src/ServiceProvider.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using Microsoft.Extensions.DependencyInjection.ServiceLookup;
 
 namespace Microsoft.Extensions.DependencyInjection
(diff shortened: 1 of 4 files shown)


davidfowl:
I think we should expose this type as an option. It's super useful 😄

src/DependencyInjection/DI/src/ServiceProvider.cs
```diff
@@ -27,6 +28,21 @@ internal ServiceProvider(IEnumerable<ServiceDescriptor> serviceDescriptors, Serv
 
             switch (options.Mode)
             {
+                case ServiceProviderMode.Default:
+#if !NETCOREAPP
+                    _engine = new DynamicServiceProviderEngine(serviceDescriptors, callback);
+#else
+                    if (RuntimeFeature.IsSupported("IsDynamicCodeCompiled"))
```
davidfowl: Isn't this a constant?

pakrym: Not in `netcoreapp2.0`, I can change to cross-compile to `netcoreapp3.0` but we'll lose IL_EMIT on downlevel netcoreapps

pakrym:
> I think we should expose this type as an option. It's super useful 😄

I don't see a reason for customers to ever change this.

davidfowl:
@jkotas Will coreRT support this soon? It would be good to try it out.

jkotas:
The live CoreRT versions have correct implementation of these APIs already:

https://github.com/dotnet/corert/blob/39071c55955b66ef1c7ce78ba2042ea6ff5f570e/src/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeFeature.CoreRT.cs#L9

Issue #861, opened 2018-12-21
ConfigurationProviders leak memory


davidfowl:
```C#
using System;
using System.Threading.Tasks;
using Microsoft.Extensions.Configuration;

namespace WebApplication45
{
    public class Program
    {
        public static async Task Main(string[] args)
        {
            Console.WriteLine("STARTING");
            Console.ReadLine();
            CreateConfig();
            Console.WriteLine("DONE");
            Console.ReadLine();
        }

        private static void CreateConfig()
        {
            for (int i = 0; i < 1000; i++)
            {
                var config = new ConfigurationBuilder()
                               .SetBasePath(Environment.CurrentDirectory)
                               .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
                               .Build();
            }
        }
    }
}
```

This roots tons of objects because of https://github.com/aspnet/Extensions/issues/786 they can't be cleaned up.

![image](https://user-images.githubusercontent.com/95136/50367962-d3216780-0538-11e9-8e15-e53e489b7455.png)

poke:
Just wondering if this would be the way we would solve this:

- Make `ConfigurationRoot` disposable
- Dispose all disposable providers when disposing the `ConfigurationRoot`
- In `WebHost`, dispose of the `ConfigurationRoot` when the web host gets disposed

If that’s correct, I would be interested to pick this up.

Since the configuration is also used in the `WebHostBuilder` I assume that we would need to add a way to dispose of the configuration root there as well (making the builder disposable)?

davidfowl:
> Make ConfigurationRoot disposable

Yes.

> Dispose all disposable providers when disposing the ConfigurationRoot

Yes

> In WebHost, dispose of the ConfigurationRoot when the web host gets disposed

This should happen for free since the ConfigurationRoot  is in the container.

poke:
@davidfowl
> This should happen for free since the ConfigurationRoot is in the container.

But doesn’t aspnet/DependencyInjection#462 prevent the configuration root from getting disposed since it is being registered [as an instance](https://github.com/aspnet/AspNetCore/blob/v2.2.1/src/Hosting/Hosting/src/WebHostBuilder.cs#L282)?

davidfowl:
Ah, yes you're right! We need to do the same for the hosting environment as well.

poke:
[...]

PR #928, opened 2019-01-10
Dispose configuration providers

poke:
This is an attempt to fix the Extensions.Configuration side of #861. The next step would be to actually dispose the configuration root inside the WebHost. I’ll attempt a PR for that later.

With this, I’ve also fixed the leak that @HaoK mentioned in https://github.com/aspnet/Extensions/pull/868#issuecomment-449690016.

cc @davidfowl

### DIFF
--- a/src/Configuration/Config/src/ConfigurationRoot.cs
+++ b/src/Configuration/Config/src/ConfigurationRoot.cs
@@ -12,9 +12,10 @@ namespace Microsoft.Extensions.Configuration
     /// <summary>
     /// The root node for a configuration.
     /// </summary>
-    public class ConfigurationRoot : IConfigurationRoot
+    public class ConfigurationRoot : IConfigurationRoot, IDisposable
     {
         private readonly IList<IConfigurationProvider> _providers;
+        private readonly IList<IDisposable> _changeTokenRegistrations;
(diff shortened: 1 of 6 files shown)


src/Configuration/Config/src/ConfigurationRoot.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.Configuration
     /// <summary>
     /// The root node for a configuration.
     /// </summary>
-    public class ConfigurationRoot : IConfigurationRoot
+    public class ConfigurationRoot : IConfigurationRoot, IDisposable
```
davidfowl: While you're here, also dispose the ChangeToken subscriptions.

poke: Good idea, I didn’t even see that before.

HaoK: That's prolly because he just added them recently :)

poke:
Squashed, rebased (just in case) and adjusted a variable name for consistency..

ajcvickers:
@davidfowl Okay to merge now?

davidfowl:
LGTM

PR #931, opened 2019-01-10
Update branding to 2.2.2

dougbu:


dougbu:
:up::date:

dougbu:
@natemcmaster please double-check in case I missed something the CI won't catch

eng/Baseline.xml
```diff
@@ -1,4 +1,4 @@
-<Baseline Version="2.2.0">
+<Baseline Version="2.2.2">
```
natemcmaster:
```suggestion
<Baseline Version="2.2.1">
```

natemcmaster: This should be the most recent released version, not the currently building version.

dougbu:
:ok: build is working now. Did I leave anything out @natemcmaster?

natemcmaster:
One more thing. Can you update dependencies.props? You can remove `-servicing-27114-05` from the 4 System.* packages (those were temporary while we waited for the January patch), and update MicrosoftNETCoreAppPackageVersion to 2.2.1. https://github.com/aspnet/Extensions/blob/6e758d41c17b5c8b76237b405593d684c4e6f69d/eng/Dependencies.props#L16-L25

dougbu:
@natemcmaster CI can't find a `4.5.1` version of the System.Security.Cryptography.Cng package. Should we build against the 7 month old `4.5.0` version or the very new `4.6.0-preview.18571.3`?

natemcmaster:
Try reverting to 4.5.0. It looks like they didn't end up needing the patch that was in the 4.5.1-servicing build.

PR #944, opened 2019-01-14
Loads secrets in parallel

pakrym:
Fixes: https://github.com/aspnet/Extensions/issues/910

### DIFF
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -52,23 +52,24 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
             var data = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
             var secrets = await _client.GetSecretsAsync(_vault).ConfigureAwait(false);
+            var tasks = new List<Task<SecretBundle>>(secrets.Count());
             do
             {
-                var tasks = new List<Task<KeyValuePair<string, string>>>();
+                tasks.Clear();
+
                 foreach (var secretItem in secrets)
                 {
-                    if (!_manager.Load(secretItem) || secretItem.Attributes?.Enabled != true)
+                    if (_manager.Load(secretItem) && secretItem.Attributes?.Enabled == true)
                     {
-                        continue;
+                        tasks.Add(_client.GetSecretAsync(secretItem.Id));
                     }
-                    tasks.Add(LoadSecretValue(secretItem));
                 }
 
-                await Task.WhenAll(tasks);
+                await Task.WhenAll(tasks).ConfigureAwait(false);
 
                 foreach (var task in tasks)
                 {
-                    data.Add(task.Result.Key, task.Result.Value);
+                    data.Add(_manager.GetKey(task.Result), task.Result.Value);
                 }
 
                 secrets = secrets.NextPageLink != null ?
@@ -77,13 +78,6 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
             } while (secrets != null);
 
             Data = data;
-        }
-
-        private async Task<KeyValuePair<string, string>> LoadSecretValue(SecretItem secretItem)
-        {
-            var value = await _client.GetSecretAsync(secretItem.Id).ConfigureAwait(false);
-            var key = _manager.GetKey(value);
-            return new KeyValuePair<string, string>(key, value.Value);
         }
     }
 }
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -1,10 +1,12 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


src/Configuration/Config.AzureKeyVault/test/AzureKeyVaultConfigurationTest.cs
```diff
@@ -170,6 +171,50 @@ public void ReplaceDoubleMinusInKeyName()
...
+
+            client.Setup(c => c.GetSecretAsync(It.IsAny<string>()))
+                .Returns(async (string id) => {
+                    var shortId = id.Substring(id.LastIndexOf('/') + 1);
+                    if (Interlocked.Decrement(ref expectedCount) == 0)
+                    {
+                        tcs.SetResult(null);
+                    }
+
+                    await tcs.Task;
```
BrennanConroy: nit: use the `WithTimeout` helper functions here

pakrym:
This is technically a breaking change. We now require thread safety from customer provided `IKeyVaultSecretManager` implementations where it wasn't required before.

On the other hand it's inlikely that `IKeyVaultSecretManager` implementations would contain any state non-thread safe state considering it's purpose is filtering and transforming secret names into configuration key names.

``` C#
/// <summary>
    /// The <see cref="IKeyVaultSecretManager"/> instance used to control secret loading.
    /// </summary>
    public interface IKeyVaultSecretManager
    {
        /// <summary>
        /// Checks if <see cref="SecretItem"/> value should be retrieved.
        /// </summary>
        /// <param name="secret">The <see cref="SecretItem"/> instance.</param>
        /// <returns><code>true</code> is secrets value should be loaded, otherwise <code>false</code>.</returns>
        bool Load(SecretItem secret);

        /// <summary>
        /// Maps secret to a configuration key.
        /// </summary>
        /// <param name="secret">The <see cref="SecretBundle"/> instance.</param>
        /// <returns>Configuration key name to store secret value.</returns>
        string GetKey(SecretBundle secret);
    }

```

/cc @muratg @Eilon

Eilon:
Out of curiosity, with this change, what is the time delta in the scenario of ~79 secrets being loaded?

Eilon:
As far as breaking changes go, I agree with the assessment. Can we update the doc comments for the affected types/methods to indicate the thread safety requirements?

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -52,16 +54,21 @@ private async Task LoadAsync()
             var secrets = await _client.GetSecretsAsync(_vault).ConfigureAwait(false);
             do
             {
+                var tasks = new List<Task<KeyValuePair<string, string>>>();
```
gfoidl: Can the List be initialized with the count of `secrets`?
Though in this case it's really minor...

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -52,16 +54,21 @@ private async Task LoadAsync()
...
                 foreach (var secretItem in secrets)
                 {
-                    if (!_manager.Load(secretItem) || (secretItem.Attributes?.Enabled != true))
+                    if (!_manager.Load(secretItem) || secretItem.Attributes?.Enabled != true)
```
gfoidl: I think the "positive case" is more readable:

```c#
foreach (var secretItem in secrets)
{
    if (_manager.Load(secretItem) && secretItem.Attributes?.Enabled ?? false)
    {
        tasks.Add(LoadSecretValue(secretIItem);
    }
}
```

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -52,16 +54,21 @@ private async Task LoadAsync()
...
                     {
                         continue;
                     }
+                    tasks.Add(LoadSecretValue(secretItem));
+                }
+
+                await Task.WhenAll(tasks);
```
gfoidl: `ConfigureAwait`?

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -52,16 +54,21 @@ private async Task LoadAsync()
...
 
-                    var value = await _client.GetSecretAsync(secretItem.Id).ConfigureAwait(false);
-                    var key = _manager.GetKey(value);
-                    data.Add(key, value.Value);
+                foreach (var task in tasks)
+                {
+                    data.Add(task.Result.Key, task.Result.Value);
```
gfoidl:
```suggestion
                    var (key, value) = task.Result;
                    data.Add(key, value);
```
?

Offtopic: it's a bit strange to "deconsruct" a `KeyValuePair<K, V>` if you want to add it to a dictionary :wink:

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -71,5 +78,12 @@ private async Task LoadAsync()
 
             Data = data;
         }
+
+        private async Task<KeyValuePair<string, string>> LoadSecretValue(SecretItem secretItem)
```
gfoidl: `LoadSecretValueAsync` as name?

pakrym:
Not a breaking change anymore. All `IKeyVaultSecretManager` interactions are done on the same thread now.

pakrym:
@Eilon our connection to KeyVault is pretty good so on my machine with 100 secrets load time went from 11 to 7 sec. I would imagine much better results on machines with slower access to keyvault.

PR #951, opened 2019-01-16
Use span APIs to compute StringSegment hashes when available

pranavkm:
Fixes https://github.com/aspnet/Extensions/issues/457

### DIFF
--- a/src/Primitives/src/Microsoft.Extensions.Primitives.csproj
+++ b/src/Primitives/src/Microsoft.Extensions.Primitives.csproj
@@ -1,11 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Primitives shared by framework extensions. Commonly used types include:
 Microsoft.Extensions.Primitives.IChangeToken
 Microsoft.Extensions.Primitives.StringValues
 Microsoft.Extensions.Primitives.StringSegment</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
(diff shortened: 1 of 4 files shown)


src/Primitives/src/Microsoft.Extensions.Primitives.csproj
```diff
@@ -1,11 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Primitives shared by framework extensions. Commonly used types include:
 Microsoft.Extensions.Primitives.IChangeToken
 Microsoft.Extensions.Primitives.StringValues
 Microsoft.Extensions.Primitives.StringSegment</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
pranavkm: cc @natemcmaster

natemcmaster: Is this the final TFM, or will this be reverted to netstandard2.1 before we RTM?

cc @davidfowl

src/Primitives/src/StringSegmentComparer.cs
```diff
@@ -34,8 +35,17 @@ public bool Equals(StringSegment x, StringSegment y)
...
         public int GetHashCode(StringSegment obj)
         {
-            // TODO: PERF; Note that .NET Core strings use randomized hash codes for security reasons.
+#if NETCOREAPP
+            return string.GetHashCode(obj.AsSpan(), Comparison);
+#else
+            if (!obj.HasValue)
+            {
+                return 0;
+            }
```
rynowak: Is this just being done for perf for the degenerate case?

pranavkm: The current code throws if you have a `StringSegment` instance without a backing buffer (e.g. `var x = default(StringSegment);`). I used existing test data for `StringSegment` which cover this case.

PR #963, opened 2019-01-17
Add async analyzers projects and simple async analyzer

pakrym:
Let's bootstrap async analyzer work.

This is just a POC for a simple data-driven analyzer.


src/Analyzers/Async/src/AsyncAnalysisData.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Immutable;
+using System.IO;
+using System.Reflection;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Extensions.Logging.Analyzers
+{
+    public class AsyncAnalysisData
+    {
+        private readonly Dictionary<string, bool> _methods;
```
rynowak: Should this just be a set?

pakrym: I'm going to add more failure kinds in the future, can be a set in this PR but would have to eventually become a dictionary.

src/Analyzers/Async/src/AsyncAnalyzer.csv
```diff
@@ -0,0 +1,3 @@
+System.Threading.Tasks.Task.Wait
+System.Threading.Tasks.Task<TResult>.Result
```
rynowak: Note that this is a pretty printed name. Most of the time type names you get from Roslyn in are the metadata format rather than looking like code a user would write. It might be more of a gotcha to try and avoid using metadata format which looks like `System.Threading.Tasks.Task`1.Result`

src/Analyzers/Async/src/Descriptors.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Extensions.Logging.Analyzers
+{
+    internal class Descriptors
+    {
+        public static DiagnosticDescriptor MEL0001SynchronouslyBlockingMethod = new DiagnosticDescriptor(
```
rynowak: Should the field name's prefix be the same as the diagnostic id?

src/Analyzers/Async/src/Descriptors.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Extensions.Logging.Analyzers
```
rynowak: ????

src/Analyzers/Async/src/Descriptors.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    internal class Descriptors
+    {
+        public static DiagnosticDescriptor MEL0001SynchronouslyBlockingMethod = new DiagnosticDescriptor(
+            "ASYNC0001", "Synchronously blocking method should not be used in async context",
+            "Synchronously blocking method should not be used in async context", "Usage", DiagnosticSeverity.Info, true);
```
rynowak: nit: put these on their own line diffing is possible.

src/Analyzers/Async/src/AsyncAnalyzer.csv
pakrym: Hm, I'm using ToDisplayString to get these did you have problems with it returning metadata formatted values?

src/Analyzers/Async/src/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+                    }
+
+                    break;
+
+                default:
+                    return;
+            }
+
+            var originalType = symbolInfo.Symbol.ContainingType.ConstructedFrom.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
+            if (store.Contains(originalType, symbolInfo.Symbol.Name))
```
rynowak: If you want, you can probably use `ToDisplayString()` on the method itself, but I assume this is done to avoid problems with closed generics not matching. I get it now, but I think a comment + example would be good to have.

src/Analyzers/Async/test/LoggingDiagnosticRunner.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Analyzer.Testing;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.Extensions.Logging.Analyzer
+{
+    public class LoggingDiagnosticRunner : DiagnosticAnalyzerRunner
```
rynowak: ???

pakrym:
@rynowak thank you for review.

src/Analyzers/Async/src/Descriptors.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.Dotnet.Analyzers.Async
+{
+    internal class Descriptors
+    {
+        public static DiagnosticDescriptor ASYNC0001SynchronouslyBlockingMethod = new DiagnosticDescriptor(
+            id: "ASYNC0001",
+            title: "Synchronously blocking method should not be used in async context",
+            messageFormat: "Synchronously blocking method should not be used in async context",
+            category: "Usage",
+            defaultSeverity: DiagnosticSeverity.Info,
```
pranavkm: Warning?

src/Analyzers/Async/src/AsyncAnalysisData.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.Dotnet.Analyzers.Async
+{
+    public class AsyncAnalysisData
+    {
+        private readonly HashSet<string> _methods;
+
+        public AsyncAnalysisData()
+        {
+            _methods = new HashSet<string>();
```
pranavkm: `StringComparer.Ordinal`

global.json
```diff
@@ -1,13 +1,8 @@
 {
```
pranavkm: Undo

src/Analyzers/Async/src/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: File name is incorrect

src/Analyzers/Async/src/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+            });
+        }
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+
+            context.RegisterCompilationStartAction(analysisContext => {
```
pranavkm: Formatting (brace on next line)

src/Analyzers/Async/src/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+        private void AnalyzeInvocation(SyntaxNodeAnalysisContext syntaxContext, AsyncAnalysisData store)
+        {
+            // Skip non-async methods
+            if (!(syntaxContext.ContainingSymbol is IMethodSymbol methodSymbol) || !methodSymbol.IsAsync)
+            {
+                return;
+            }
+
+            SymbolInfo symbolInfo;
+            switch (syntaxContext.Node.Kind())
```
pranavkm: You could just switch on the node no?

```C#
switch (syntaxContext.Node)
{
   case InvocationExpressionSyntax invocation: ...
}
```

src/Analyzers/Async/src/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                case SyntaxKind.InvocationExpression:
+                    var invocation = (InvocationExpressionSyntax)syntaxContext.Node;
+
+                    symbolInfo = ModelExtensions.GetSymbolInfo(syntaxContext.SemanticModel, invocation, syntaxContext.CancellationToken);
```
pranavkm: `semanticModel.GetSymbolInfo(...`

src/Analyzers/Async/src/Microsoft.Dotnet.Analyzers.Async.csproj
```diff
@@ -0,0 +1,32 @@
...
+
+  <ItemGroup>
+    <EmbeddedResource Include="AsyncAnalyzer.csv" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.CodeAnalysis.Common" PrivateAssets="All" />
+    <Reference Include="Microsoft.CodeAnalysis.CSharp" PrivateAssets="All" />
+    <Reference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" PrivateAssets="All" />
+    <PackageReference Update="NETStandard.Library" PrivateAssets="All" />
```
pranavkm: Why is this needed?

src/Analyzers/Async/src/Microsoft.Dotnet.Analyzers.Async.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <VerifyVersion>false</VerifyVersion>
+    <TargetFramework>netstandard1.3</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <EnableApiCheck>false</EnableApiCheck>
+    <BuildOutputTargetFolder>analyzers/dotnet/cs/</BuildOutputTargetFolder>
+    <UseLatestPackageReferences>true</UseLatestPackageReferences>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Remove="AsyncAnalyzer.csv" />
```
pranavkm: Remove?

src/Analyzers/Async/src/Microsoft.Dotnet.Analyzers.Async.csproj
pakrym: I remember it being required at one point.

natemcmaster: Still required. I recommend keeping this.

The reason is that this assembly compiles for netstandard1.3, but it's not really a .NET Standard 1.3 project. Without adding PrivateAssets='all' to suppress this dependency, nuget generates an incorrect nuspec.

An alternative which is easier to grok is
```xml
    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
```

PR #977, opened 2019-01-19
ObjectPool supports Dispose

gfoidl:
Added `DisposableObjectPool`, that disposes items which are not returned to the pool, and all items in the pool when `Dispose` is called.
After a call to `Dispose` a call to `Get` will throw a `ObjectDisposedException`, whilst `Return` will just dispose the given object.

`DefaultObjectPoolProvider` will choose on `T` which object pool to create. If `T` implements `IDisposable` the (new) `DisposableObjectPool` will be created, else the `DefaultObjectPool`.

For scenarious without DI a factory method `ObjectPool.Create<T>` got introduced.

Addresses https://github.com/aspnet/Extensions/issues/973


src/ObjectPool/src/DefaultObjectPool.cs
```diff
@@ -10,10 +10,10 @@ namespace Microsoft.Extensions.ObjectPool
 {
     public class DefaultObjectPool<T> : ObjectPool<T> where T : class
     {
-        private readonly ObjectWrapper[] _items;
+        protected readonly ObjectWrapper[] _items;
```
rynowak: hmmm, I'd really like to avoid this showing up in a public API surface. Move to a top level struct and mark internal?

src/ObjectPool/src/DefaultObjectPool.cs
```diff
@@ -71,29 +71,43 @@ private T GetViaScan()
...
+        private bool ReturnViaScan(T obj)
         {
             var items = _items;
-            for (var i = 0; i < items.Length && Interlocked.CompareExchange(ref items[i].Element, obj, null) != null; ++i)
+            for (var i = 0; i < items.Length; i++)
             {
+                if (Interlocked.CompareExchange(ref items[i].Element, obj, null) == null)
+                {
+                    return true;
+                }
```
rynowak: I assume the structure of this had to change because you need to return true now instead of exiting the loop? IDK from your earlier optimization work if putting this test and loop condition together was a significant improvement. 

You could also do something like

```
var returned = false;
for (var i = 0; i < (items.Length && !(returned = Interlocked.CompareExchange(ref items[i].Element, obj, null) != null)); i++)
{
}

return returned;
```

Yeah, it's "too clever" 😉  but it might be worthwhile.

rynowak: This might end up generating the same code anyway.

src/ObjectPool/src/DefaultObjectPoolProvider.cs
```diff
@@ -11,6 +11,11 @@ public class DefaultObjectPoolProvider : ObjectPoolProvider
 
         public override ObjectPool<T> Create<T>(IPooledObjectPolicy<T> policy)
         {
+            if (typeof(IDisposable).IsAssignableFrom(typeof(T)))
```
rynowak: This should also have argument validation for `policy`. I know it will just throw later, but for maintenance reasons we usually like to validate arguments at public entry points.

src/ObjectPool/src/DisposableObjectPool.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                ThrowObjectDisposedException();
+            }
+
+            return base.Get();
+
+            void ThrowObjectDisposedException()
+            {
+                throw new ObjectDisposedException(this.GetType().Name);
+            }
```
rynowak: Do this give the same code quality benefits as static throw helper? Curious because it captures `this`.

src/ObjectPool/src/DisposableObjectPool.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        public override void Return(T obj)
+        {
```
rynowak: Should this have a null-check?

src/ObjectPool/src/DisposableObjectPool.cs
```diff
@@ -0,0 +1,70 @@
...
+            // When the pool is disposed or the obj is not returned to the pool, dispose it
+            if (_isDisposed || !ReturnCore(obj))
+            {
+                DisposeItem(obj);
+            }
```
rynowak: This is so simple and clean.

src/ObjectPool/src/DisposableObjectPool.cs
```diff
@@ -0,0 +1,70 @@
...
+            _firstItem = null;
+
+            ObjectWrapper[] items = _items;
+            for (var i = 0; i < items.Length; i++)
+            {
+                DisposeItem(items[i].Element);
+                items[i].Element = null;
+            }
+
+            _isDisposed = true;
```
rynowak: Is there a race condition here? Someone could *return* an item on another thread after L50 has run on a thread that's disposing the pool.

src/ObjectPool/src/DefaultObjectPool.cs
rynowak: That would accomplish roughly the same thing so that's fine. I just want to make sure that implementation details like don't leak out. It's probably appropriate to make `DisposableObjectPool<T>` internal as well.

src/ObjectPool/src/DefaultObjectPool.cs
gfoidl: Your cleverness produces a bit better code, but it still regresses by about 7%. So I'll move this part to the `DisposableObjectPool` and leave the `DefaultObjectPool` with the fast `Return`-code.

src/ObjectPool/src/DefaultObjectPool.cs
gfoidl: Done -- will be updated with the next push.

src/ObjectPool/src/DefaultObjectPoolProvider.cs
gfoidl: Done -- will be updated with the next push.

src/ObjectPool/src/DisposableObjectPool.cs
gfoidl: Yeah, same benefit. The key here is that the compiler generated static method will just call `GetType`, so no `this` is needed, as the type can be also be obtained from the static method.

I'll remove the explicit `this`, as it is not needed and doesn't follow the coding conventions here.

<details>
  <summary>JIT produces</summary>

```asm
G_M36605_IG01:
       55                   push     rbp
       488BEC               mov      rbp, rsp
       90                   nop      

G_M36605_IG02:
       807F2900             cmp      byte  ptr [rdi+41], 0
       750E                 jne      SHORT G_M36605_IG05

G_M36605_IG03:
       48B8B80C359A1B7F0000 mov      rax, 0x7F1B9A350CB8

G_M36605_IG04:
       5D                   pop      rbp
       48FFE0               rex.jmp  rax

G_M36605_IG05:
       E87AFAFFFF           call     DisposableObjectPool`1:<Get>g__ThrowObjectDisposedException|3_0():this
       CC                   int3 
```
Even with tail-call for `base.Get` -- good JIT :smile:
</details>

src/ObjectPool/src/DisposableObjectPool.cs
gfoidl: I don't see any need for the null-check.

If `obj != null` everything is fine.
if `obj == null` the return-code will handle it, for the disposed-case there is `item is IDisposable`, so if null the branch won't be taken.

src/ObjectPool/src/DisposableObjectPool.cs
gfoidl: The dispose-method is not threadsafe. So yes there is a race condition.
Will move the `_isDisposed` to top of method.

src/ObjectPool/src/DefaultObjectPool.cs
gfoidl: From the JIT-dumps I saw that the "scan-methods" aren't inlined due `runtime dictionary lookup` (which comes from generics). When manually inlining them, perf rises a bit.

src/ObjectPool/test/ThreadingTest.cs
```diff
@@ -13,10 +13,22 @@ public class ThreadingTest
...
-        public void RunThreadingTest()
+        public void DefaultObjectPool_RunThreadingTest()
         {
-            _cts = new CancellationTokenSource();
             _pool = new DefaultObjectPool<Item>(new DefaultPooledObjectPolicy<Item>(), 10);
+            RunThreadingTest();
+        }
+
+        [Fact]
+        public void DisposableObjectPool_RunThreadingTest()
```
gfoidl: `DisposableObjectPool` handles the `Return` by itself, so I think this should be tested.

src/ObjectPool/test/DisposableObjectPoolTest.cs
```diff
@@ -0,0 +1,145 @@
...
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Microsoft.Extensions.ObjectPool
+{
+    public class DisposableObjectPoolTest
+    {
+        [Fact]
+        public void DisposableObjectPoolWithDefaultPolicy_GetAnd_ReturnObject_SameInstance()
```
gfoidl: Copied some tests from the `DefaultObjectPool` as the returning is handled by the `DisposableObjectPool`, so this should be tested.

Before the last commit it was just delegating to its base class, so no explicit tests were needed.

PR #984, opened 2019-01-22
Add TryGetService and IsRegistered

khellang:
Ref. https://github.com/aspnet/AspNetCore/pull/6915#discussion_r249700661.

I found traces of this method [way back to 2014](https://github.com/aspnet/DependencyInjection/issues/75), but couldn't see it anywhere.

Some places these methods could be useful:

### IsRegistered

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs#L184-L192

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Http/Routing/src/Builder/RoutingBuilderExtensions.cs#L64-L70

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs#L88-L96

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Http/Routing/src/RouteBuilder.cs#L26-L32

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs#L45-L49

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/SignalR/server/SignalR/src/SignalRAppBuilderExtensions.cs#L23-L28

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/SignalR/server/Core/src/SignalRConnectionBuilderExtensions.cs#L24-L29

### TryGetService

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Middleware/NodeServices/src/Configuration/NodeServicesOptions.cs#L38-L49

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Middleware/NodeServices/src/Configuration/NodeServicesOptions.cs#L51-L55

https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs#L122-L132

These were just found based on a quick GitHub search...

### DIFF
--- a/src/DependencyInjection/DI.Abstractions/src/ServiceProviderServiceExtensions.cs
+++ b/src/DependencyInjection/DI.Abstractions/src/ServiceProviderServiceExtensions.cs
@@ -118,6 +118,28 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Determines whether a service of type <typeparamref name="T"/> has been registered with the <see cref="IServiceProvider"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of service to check.</typeparam>
+        /// <param name="provider">The <see cref="IServiceProvider"/> to check.</param>
+        /// <returns><c>true</c> if the service has been registered, <c>false</c> otherwise.</returns>
+        public static bool IsRegistered<T>(this IServiceProvider provider)
+        {
+            return provider.IsRegistered(typeof(T));
+        }
+
+        /// <summary>
+        /// Determines whether a service of type <paramref name="serviceType"/> has been registered with the <see cref="IServiceProvider"/>.
+        /// </summary>
+        /// <param name="provider">The <see cref="IServiceProvider"/> to check.</param>
+        /// <param name="serviceType">An object that specifies the type of service to check.</param>
+        /// <returns><c>true</c> if the service has been registered, <c>false</c> otherwise.</returns>
+        public static bool IsRegistered(this IServiceProvider provider, Type serviceType)
+        {
+            return provider.TryGetService(serviceType, out _);
+        }
+
+        /// <summary>
         /// Tries to get the service of type <typeparamref name="T"/> from the <see cref="IServiceProvider"/>.
         /// </summary>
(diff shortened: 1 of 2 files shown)


pakrym:
Not sure what benefit this adds, it's just another way to write a null check.

khellang:
You could say that about all Try* methods. It's a pretty established concept. I think it reads and conveys intent better.

pakrym:
No, we usually add Try methods to avoid throwing an exception in a well-known failure case.

khellang:
But why does the pattern exist instead of just returning null? For disambiguation?

pakrym:
It depends on the context, often null is a valid value (like with Dictionary<>) so you have to do something else for the error case so we get exception and a `TryGet` method. Also returning null doesn't give the customer a lot of information about the failure.

khellang:
> Also returning null doesn't give the customer a lot of information about the failure.

Neither does returning `false` 😉

Anyway, it's up to you if wanna take it or not. BTW, I did a quick GitHub search for [`TryGetService IServiceProvider`](https://github.com/search?utf8=✓&q=trygetservice+iserviceprovider&type=Code) and there's a ton of these methods around, which I think suggest that I'm not crazy (or alone) in wanting these methods.

pakrym:
I'm sorry, we don't think this brings enough additional value over null check to justify adding.

Issue #851, opened 2018-12-20
ConfigurePrimaryHttpMessageHandler<T>() not loading handler in ASP.NET Core 2.2


bkaidbb:
### Describe the bug
Registering an HttpMessageHandler by generic type parameter is causing that handler to not be loaded.  

### To Reproduce
1. Using version '2.2' of package 'Microsoft.AspNetCore.App', and 'Microsoft.Extensions.Http'
2. Run this code at startup:
```
services.AddTransient<MyClientHttpMessageHandler>();
services
    .AddHttpClient<IMyClient, MyClient>()
    .ConfigurePrimaryHttpMessageHandler<MyClientHttpMessageHandler>();
```
3. Make an http client 
4. Error is that MyClientHttpMessageHandler won't be loaded or invoked.

### Expected behavior
The handler is properly invoked when the client is used.

### Workaround
```
services
    .AddHttpClient<IMyClient, MyClient>()
    .ConfigurePrimaryHttpMessageHandler(s => s.GetRequiredService<MyClientHttpMessageHandler>())
```

### Additional context
This was the only issue I ran into upgrading a large project from netcoreapp2.1 to netcoreapp2.2. 

.NET Core SDK (reflecting any global.json):
 Version:   2.2.101
 Commit:    236713b0b7

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.101\

Host (useful for support):
  Version: 2.2.0
  Commit:  1249f08fed

satano:
We hit this issue as well. But we found the other side of its behaviour. Our message handler is used in default (not named) `HttpClient`, where it should not be used.

The problem is, that `ConfigurePrimaryHttpMessageHandler<THandler>()` is not using builder name as the other two `ConfigurePrimaryHttpMessageHandler()` methods. So effectively, it registers this handler not for the named client, but for default one.

- [ConfigurePrimaryHttpMessageHandler(builder, configureHandler)](https://github.com/aspnet/Extensions/blob/master/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs#L228) - using `builder.Name`
[...]

PR #988, opened 2019-01-22
ConfigurePrimaryHttpMessageHandler<T> uses builder name

satano:
`HttpClientBuilderExtensions.ConfigurePrimaryHttpMessageHandler<THandler>()` uses builder name, so handler is configured for correct `HttpClient`.

Addresses #851


rynowak:
Merged. Thanks!

PR #993, opened 2019-01-23
Enable strict mode in roslyn compiler

pakrym:
https://github.com/aspnet/AspNetCore/issues/3645


pakrym:
FYI @muratg I think this is safe to merge.

PR #1005, opened 2019-01-27
Add AsyncDisposable support

pakrym:
Fixes: https://github.com/aspnet/Extensions/issues/426

Implement IAsyncDisposable on ServiceProvider and scopes.

`.Dispose` throws if there is a service that only implements `IAsyncDisposable`

### DIFF
--- a/src/DependencyInjection/DI/src/Resources.resx
+++ b/src/DependencyInjection/DI/src/Resources.resx
@@ -156,4 +156,7 @@
   <data name="ImplementationTypeCantBeConvertedToServiceType" xml:space="preserve">
     <value>Implementation type '{0}' can't be converted to service type '{1}'</value>
   </data>
+  <data name="AsyncDisposableServiceDispose" xml:space="preserve">
+    <value>'{0}' type only implements IAsyncDisposable. Use DisposeAsync to dispose the container.</value>
+  </data>
 </root>
--- a/src/DependencyInjection/DI/src/Microsoft.Extensions.DependencyInjection.csproj
+++ b/src/DependencyInjection/DI/src/Microsoft.Extensions.DependencyInjection.csproj
@@ -2,12 +2,13 @@
 
   <PropertyGroup>
     <Description>Default implementation of dependency injection for Microsoft.Extensions.DependencyInjection.</Description>
-    <TargetFrameworks>netcoreapp2.0;net461;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.0;net461;netstandard2.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>dependencyinjection;di</PackageTags>
 
     <ILEmitBackend Condition="$(TargetFramework) != 'netstandard2.0'">True</ILEmitBackend>
     <DefineConstants Condition="'$(ILEmitBackend)' == 'True'">$(DefineConstants);IL_EMIT</DefineConstants>
+    <DefineConstants Condition="'$(TargetFramework)' == 'netcoreapp3.0'">$(DefineConstants);DISPOSE_ASYNC</DefineConstants>
 
     <!-- Debug IL generation -->
     <ILEmitBackendSaveAssemblies>False</ILEmitBackendSaveAssemblies>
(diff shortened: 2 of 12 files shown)


src/DependencyInjection/DI/src/Resources.resx
```diff
@@ -156,4 +156,7 @@
   <data name="ImplementationTypeCantBeConvertedToServiceType" xml:space="preserve">
     <value>Implementation type '{0}' can't be converted to service type '{1}'</value>
   </data>
+  <data name="AsyncDisposableServiceDispose" xml:space="preserve">
+    <value>'{0}' type only implements IAsyncDisposable. Use DisposeAsync to dispose the container.</value>
```
davidfowl: I'm not 100% sure this is reasonable. There might be cases where something is added to the container out of somebody's control and it'll cause hosting to just crash and burn when the container gets disposed.

pakrym: They can always DisposeAsync().Wait() to workaround

davidfowl: Hosting is the one calling Dispose on the container.

davidfowl: Maybe make it an option so people can work around it in cases where they don't own the container (like Azure Functions)? By default we can throw but you can turn it off to get the sync over async behavior back.

pakrym: Both us and Azure Functions are going to move to DisposeAsync() in time for 3.0 and I doubt that there are a lot of types that implement only IAsyncDisposable

pakrym: I can add a flag if you insist.

davidfowl: Our templates are calling Dispose not DisposeAsync, it's likely this will be the case until we ship the RTM of .NET Core 3.0. How do we mitigate this in the mean time?

pakrym: Why RTM? I'll make a change to hosting after merging this PR. IAsyncDisposable is netcoreapp3.0 only feature so backcompat is not a concern here.

davidfowl: Because the templates need a higher c# version to use async main, in the templates. Today we call Run. This will also be breaking if somebody updates from 2.0 to 3.0 and components start adding IAsyncDisposable things. Their templates will have the old call to Run and nothing will tell people they need to change that to be an async main with RunAsync

pakrym: I don't understand, `.Run` already calls `RunAsync` and blocks on it so templates would work just fine.

https://github.com/aspnet/Hosting/blob/master/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L65

davidfowl: Ah true, so I guess this error would never happen unless you explicitly called Dispose on the host.

davidfowl: Maybe TestHost scenarios have the potential to break?

pakrym: Nope, WebHost.Dispose() blocks StopAsync and `provider.DisposeAsync` calls would get moved there 
https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/Hosting/src/Internal/WebHost.cs#L350

davidfowl: You shouldn't call Dispose in StopAsync. Thats not right. Stopping needs to happen before disposing, not during.

pakrym: You a right, Dispose will call DisposeAsync and block.

pakrym:
Seems like we decided that we have a reasonable migration story for customers with current implementation. Merging.

davidfowl:
What’s the migration story?

pakrym:
WebHost and TestHost continue working as before, customers would have to change their calls to Dispose to DisposeAsync if they use async disposable only types.

davidfowl:
Get this into ASP.NET Core and EF as soon as possible.

Issue #999, opened 2019-01-24
HttpClientFactory managed HttpClient deadlocks in single-threaded SynchronizationContext


akunzai:
### Describe the bug
my legacy app is still running on ASP.NET,
after internal library implementation adopting the HttpClientFactory managed HttpClient,
async call with await in HttpClient never returns. 
this issue is very similar to #563 

### To Reproduce

```sh
dotnet new console -o Console1
cd Console1
dotnet add package Microsoft.Extensions.DependencyInjection --version 2.2.0
dotnet add package Microsoft.Extensions.Http --version 2.2.0
```

 Program.cs
```csharp
using System;
using System.Collections.Generic;
using System.Net.Http;
using System.Threading;
using Microsoft.Extensions.DependencyInjection;

namespace Console1
{
    public class Program
    {
        static void Main(string[] args)
        {
            var services = new ServiceCollection();
            services.AddHttpClient();
            var provider = services.BuildServiceProvider();
            var factory = provider.GetRequiredService<IHttpClientFactory>();
            var client1 = new HttpClient();
            SingleThreadedSynchronizationContext.Run(() =>
            {
                Console.Write("default HttpClient in single-threaded SynchronizationContext ... ");
                client1.GetAsync("http://example.com").GetAwaiter().GetResult();
                Console.WriteLine($"Done");
            });
            var client2 = factory.CreateClient();
            SingleThreadedSynchronizationContext.Run(() =>
            {
                Console.Write("managed HttpClient in single-threaded SynchronizationContext ... ");
                client2.GetAsync("http://example.com").GetAwaiter().GetResult();
                Console.WriteLine($"Done");
            });
        }
    }

    public class SingleThreadedSynchronizationContext : SynchronizationContext
    {
        private readonly Queue<(SendOrPostCallback Callback, object State)> _queue = new Queue<(SendOrPostCallback Callback, object State)>();

        public override void Post(SendOrPostCallback d, object state)
        {
            _queue.Enqueue((d, state));
        }

        public static void Run(Action action)
        {
            var previous = Current;
            var context = new SingleThreadedSynchronizationContext();
            SetSynchronizationContext(context);
            try
            {
                action();
                while (context._queue.TryDequeue(out var item))
                {
                    item.Callback(item.State);
                }
            }
            finally
            {
                SetSynchronizationContext(previous);
            }
        }
    }
}
```

Run this code

```sh
$ dotnet run
default HttpClient in single-threaded SynchronizationContext ... Done
[...]

PR #1006, opened 2019-01-28
Fix deadlocks in single-threaded SynchronizationContext

akunzai:
Add `ConfigureAwait(false)` for Logging(Scope)HttpMessageHandler
to prevent deadlocks in single-threaded SynchronizationContext

Addresses #999

### DIFF
--- a/src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
+++ b/src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
@@ -1020,7 +1020,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
             public TransientService Service { get; }
         }
-        
+
         private class SingleThreadedSynchronizationContext : SynchronizationContext
         {
             private readonly Queue<(SendOrPostCallback Callback, object State)> _queue = new Queue<(SendOrPostCallback Callback, object State)>();
@@ -1030,20 +1030,15 @@ namespace Microsoft.Extensions.DependencyInjection
                 _queue.Enqueue((d, state));
             }
 
-            public override void Send(SendOrPostCallback d, object state)
-            {
-                base.Send(d, state);
-            }
-
             public static void Run(Action action)
             {
-                var previous = SynchronizationContext.Current;
+                var previous = Current;
                 var context = new SingleThreadedSynchronizationContext();
-                SynchronizationContext.SetSynchronizationContext(context);
+                SetSynchronizationContext(context);
                 try
                 {
                     action();
-                    while(context._queue.Count > 0)
+                    while (context._queue.Count > 0)
                     {
                         var item = context._queue.Dequeue();
                         item.Callback(item.State);
@@ -1051,7 +1046,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 }
                 finally
                 {
-                    SynchronizationContext.SetSynchronizationContext(previous);
+                    SetSynchronizationContext(previous);
                 }
             }
         }
--- a/src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
+++ b/src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
@@ -778,39 +778,44 @@ namespace Microsoft.Extensions.DependencyInjection
         public void AddHttpClient_GetAwaiterAndResult_InSingleThreadedSynchronizationContext_ShouldNotHangs()
         {
             // Arrange
-            var serviceCollection = new ServiceCollection();
-            serviceCollection.AddHttpClient("example.com")
-                .ConfigurePrimaryHttpMessageHandler(() =>
-                {
-                    var mockHandler = new Mock<HttpMessageHandler>();
-                    mockHandler
-                    .Protected()
-                    .Setup<Task<HttpResponseMessage>>(
-                        "SendAsync",
-                        ItExpr.IsAny<HttpRequestMessage>(),
-                        ItExpr.IsAny<CancellationToken>()
-                    )
-                    .Returns(async () =>
+            using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10)))
+            {
+                var token = cts.Token;
+                token.Register(() => throw new OperationCanceledException(token));
+                var serviceCollection = new ServiceCollection();
+                serviceCollection.AddHttpClient("example.com")
+                    .ConfigurePrimaryHttpMessageHandler(() =>
                     {
-                        await Task.Delay(1).ConfigureAwait(false);
-                        return new HttpResponseMessage(HttpStatusCode.OK);
+                        var mockHandler = new Mock<HttpMessageHandler>();
+                        mockHandler
+                        .Protected()
+                        .Setup<Task<HttpResponseMessage>>(
+                            "SendAsync",
+                            ItExpr.IsAny<HttpRequestMessage>(),
+                            ItExpr.IsAny<CancellationToken>()
+                        )
+                        .Returns(async () =>
+                        {
+                            await Task.Delay(1).ConfigureAwait(false);
+                            return new HttpResponseMessage(HttpStatusCode.OK);
+                        });
+                        return mockHandler.Object;
                     });
-                    return mockHandler.Object;
+
+                var services = serviceCollection.BuildServiceProvider();
+                var factory = services.GetRequiredService<IHttpClientFactory>();
+                var client = factory.CreateClient("example.com");
+                var hangs = true;
+                SingleThreadedSynchronizationContext.Run(() =>
+                {
+                    // Act
+                    client.GetAsync("http://example.com", token).GetAwaiter().GetResult();
+                    hangs = false;
                 });
 
-            var services = serviceCollection.BuildServiceProvider();
-            var factory = services.GetRequiredService<IHttpClientFactory>();
-            var client = factory.CreateClient("example.com");
-            var hangs = true;
-            SingleThreadedSynchronizationContext.Run(() =>
-            {
-                // Act
-                client.GetAsync("http://example.com").GetAwaiter().GetResult();
-                hangs = false;
-            });
-
-            // Assert
-            Assert.False(hangs);
+                // Assert
+                Assert.False(hangs);
+            }
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -996,5 +1020,149 @@ public TypedClientWithTransientService(HttpClient httpClient, TransientService s
 
             public TransientService Service { get; }
         }
+
+        // https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/SingleThreadedSynchronizationContext.cs
+        private class SingleThreadedSynchronizationContext : SynchronizationContext, IDisposable
```
davidfowl: Just use blocking collection, this is a bit overengineered for a unit test.

akunzai: because the simple one `SingleThreadedSynchronizationContext.cs` cannot re-produce same deadlocks behavior.

```csharp
using System;
using System.Collections.Concurrent;
using System.Net.Http;
using System.Threading;

namespace Console1
{
    public class Program
    {
        static void Main(string[] args)
        {
            services.AddHttpClient();
            var provider = services.BuildServiceProvider();
            var factory = provider.GetRequiredService<IHttpClientFactory>();
            var client1 = new HttpClient();
            SingleThreadedSynchronizationContext.Run(() =>
            {
                Console.Write("default HttpClient in single-threaded SynchronizationContext ... ");
                client1.GetAsync("http://example.com").GetAwaiter().GetResult();
                Console.WriteLine($"Done");
            });
            var client2 = factory.CreateClient();
            SingleThreadedSynchronizationContext.Run(() =>
            {
                Console.Write("managed HttpClient in single-threaded SynchronizationContext ... ");
                client2.GetAsync("http://example.com").GetAwaiter().GetResult();
                Console.WriteLine($"Done");
            });
        }
    }

    public class SingleThreadedSynchronizationContext : SynchronizationContext
    {
        private readonly BlockingCollection<(SendOrPostCallback Callback, object State)> _queue = new BlockingCollection<(SendOrPostCallback Callback, object State)>();
        private readonly Thread _thread;

        public SingleThreadedSynchronizationContext()
        {
            _thread = new Thread(ProcessQueue);
            _thread.Start();
        }

        private void ProcessQueue(object obj)
        {
            foreach (var item in _queue.GetConsumingEnumerable())
            {
                item.Callback(item.State);
            }
        }

        public override void Post(SendOrPostCallback d, object state)
        {
            _queue.Add((d, state));
        }

        public static void Run(Action action)
        {
            var previous = SynchronizationContext.Current;
            var context = new SingleThreadedSynchronizationContext();
            SynchronizationContext.SetSynchronizationContext(context);
            try
            {
                action();
            }
            finally
            {
                SynchronizationContext.SetSynchronizationContext(previous);
            }
        }
    }
}
```

Run this code

```sh
default HttpClient in single-threaded SynchronizationContext ... Done
managed HttpClient in single-threaded SynchronizationContext ... Done
```

It will hangs on Console exiting.

davidfowl: Thats because the thread needs to be a background thread. I'll test it out. It should be easy to reproduce

davidfowl: Odd, when I run it, it never hits the sync context.

davidfowl: Ah, wait my bad that was the first one.

src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -996,5 +1020,149 @@ public TypedClientWithTransientService(HttpClient httpClient, TransientService s
...
+                        throw new InvalidOperationException("Shutting down SingleThreadedTestSynchronizationContext with work still in the queue.");
+                }
+            }
+
+            private static InvalidOperationException CreateInvalidWhenShutDownException()
+            {
+                return new InvalidOperationException("This SingleThreadedTestSynchronizationContext has been shut down.");
+            }
+
+            public void Run()
```
davidfowl: Who calls this?

akunzai: The entire `SingleThreadedSynchronizationContext.cs` is copy from https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/SingleThreadedSynchronizationContext.cs

davidfowl: This sync context isn't working. It isn't single threaded, it just doesn't run the continuations. If that's the test then you could just have a noop sync context and prove that nothing ever gets scheduled on it.

davidfowl:
```C#
public class SingleThreadedSynchronizationContext : SynchronizationContext
{
	private readonly Queue<(SendOrPostCallback Callback, object State)> _queue = new Queue<(SendOrPostCallback Callback, object State)>();

	public override void Post(SendOrPostCallback d, object state)
	{
		_queue.Enqueue((d, state));
	}

	public override void Send(SendOrPostCallback d, object state)
	{
		base.Send(d, state);
	}

	public static void Run(Action action)
	{
		var previous = SynchronizationContext.Current;
		var context = new SingleThreadedSynchronizationContext();
		SynchronizationContext.SetSynchronizationContext(context);
		try
		{
			action();
			while (context._queue.TryDequeue(out var item))
			{
				item.Callback(item.State);
			}
		}
		finally
		{
			SynchronizationContext.SetSynchronizationContext(previous);
		}
	}
}
```

akunzai: Hi @davidfowl , It works, thanks for your help.

akunzai:
The master branch seems broken?

davidfowl:
@akunzai Nope, that's something that's decided at a higher level. We'll backport the fix if it meets the patch bar (just FYI I think it does meet the bar but there's a process). It's always fixed in the "next release" first.

src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -771,6 +774,45 @@ public async Task AddHttpClient_MessageHandler_Scope_TransientDependency()
             }
         }
 
+        [Fact]
+        public void AddHttpClient_GetAwaiterAndResult_InSingleThreadedSynchronizationContext_ShouldNotHangs()
```
davidfowl: I think we need a way for this test to avoid the hang if a bug is introduced. We need a timeout somewhere.

src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -771,6 +774,50 @@ public async Task AddHttpClient_MessageHandler_Scope_TransientDependency()
...
+        {
+            // Arrange
+            using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10)))
+            {
+                var token = cts.Token;
+                token.Register(() => throw new OperationCanceledException(token));
```
davidfowl: Why do you need this?

src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -771,6 +774,50 @@ public async Task AddHttpClient_MessageHandler_Scope_TransientDependency()
...
+                        mockHandler
+                        .Protected()
+                        .Setup<Task<HttpResponseMessage>>(
+                            "SendAsync",
+                            ItExpr.IsAny<HttpRequestMessage>(),
+                            ItExpr.IsAny<CancellationToken>()
+                        )
+                        .Returns(async () =>
+                        {
+                            await Task.Delay(1).ConfigureAwait(false);
```
davidfowl: Is this important?

src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
```diff
@@ -996,5 +1043,35 @@ public TypedClientWithTransientService(HttpClient httpClient, TransientService s
 
             public TransientService Service { get; }
         }
+
+        private class SingleThreadedSynchronizationContext : SynchronizationContext
+        {
+            private readonly Queue<(SendOrPostCallback Callback, object State)> _queue = new Queue<(SendOrPostCallback Callback, object State)>();
```
rynowak: Isn't it somewhat problematic to use `Queue<>` here because it's not thread-safe? (use `ConcurrentQueue<>`) I'd prefer to fix this even if it's not hit in these tests because someone else might move this code or copy from it.

davidfowl: Sure, this sync context impl is kinda of a lie anyways. The test may not be testing the right thing TBH, the "fix" is basically that it doesn't queue any work to the sync context outside of top level code.

rynowak: Sure. I'm going to merge this PR and then fix it. I'm worried about someone copying this and creating flaky tests.

rynowak:
Thanks @akunzai

src/HttpClientFactory/Http/test/DependencyInjection/HttpClientFactoryServiceCollectionExtensionsTest.cs
slang25: It looks like this is really a "deadlock detection" sync context, a bit [like this](https://github.com/ramondeklein/deadlockdetection/blob/master/DeadlockDetection/DeadlockDetectionSynchronizationContext.cs). I agree it doesn't feel right for a unit test.

PR #1010, opened 2019-01-29
Add DisposeAsync support to generic host

pakrym:
aspnet/Extensions#426

### DIFF
--- a/src/Hosting/Hosting/src/Internal/Host.cs
+++ b/src/Hosting/Hosting/src/Internal/Host.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -13,6 +13,9 @@ using Microsoft.Extensions.Options;
 namespace Microsoft.Extensions.Hosting.Internal
 {
     internal class Host : IHost
+#if DISPOSE_ASYNC
+      , IAsyncDisposable
+#endif
     {
         private readonly ILogger<Host> _logger;
         private readonly IHostLifetime _hostLifetime;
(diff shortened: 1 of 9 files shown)


src/Hosting/Hosting/test/Internal/HostTests.AsyncDisposable.cs
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NETCOREAPP
```
Tratcher: netcoreapp3.0? Or use the same DISPOSE_ASYNC

pakrym:
> What about the places where Host.Dispose is currently called?

What about them?

src/Hosting/Hosting/test/Internal/HostTests.AsyncDisposable.cs
pakrym: No need to be specific to 3.0, every netcoreapp past 3.0 would work.

davidfowl:
What about RunAsync?

pakrym:
Done

src/Hosting/Hosting/src/Internal/Host.cs
```diff
@@ -98,9 +101,29 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
...
 
+#if DISPOSE_ASYNC
+        public void Dispose()
+        {
+            DisposeAsync().GetAwaiter().GetResult();
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            switch (Services)
```
Tratcher: does switch work on null?

src/Hosting/Hosting/test/Internal/HostTests.AsyncDisposable.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.Extensions.Hosting.Internal
+{
+    public partial class HostTests
+    {
+        [Fact]
+        public async Task HostCallsDisposeAsyncOnServiceProvider()
```
Tratcher: No test that directly calls DisposeAsync?

src/Hosting/Hosting/src/Internal/Host.cs
pakrym: Yes

src/Hosting/Hosting/test/Internal/HostTests.AsyncDisposable.cs
pakrym: Added

PR #1016, opened 2019-01-29
Reduce build times

ryanbrandenburg:
Knocks ~14 minutes off non-PR builds by not uploading files we don't care about.


ryanbrandenburg:
I was going to do that, but you had been talking about preserving the shape of the artifact we publish and I didn't see a way to do that while being more specific about the files we publish. Are we cool with breaking the shape now or did I miss a parameter to accomplish this on the task?

natemcmaster:
We had long talk (aka fight) with Arcade while you were OOF and we decided to concede to their conventions. I think we can update the release pipeline to use their new shape.

PR #1021, opened 2019-01-30
Workaround for dotnet/cli#10528 - fix NU1605 error when Maestro updates corefx dependencies

natemcmaster:
Unblocks https://github.com/aspnet/Extensions/pull/1020

We are getting package downgrade errors due to this implicit package reference: https://github.com/dotnet/sdk/blob/2eb6c546931b5bcb92cd3128b93932a980553ea1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#L67-L73.
This works around the problem by updating the implicit version variable.

cc @nguerrera @dsplaisted

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -63,5 +63,9 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>044fa99d22e576fcfaf63f09ee9bba821855dddb</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.NETCore.Platforms" Version="3.0.0-preview.19073.11">
+      <Uri>https://github.com/dotnet/corefx</Uri>
+      <Sha>0000</Sha>
+    </Dependency>
   </ToolsetDependencies>
 </Dependencies>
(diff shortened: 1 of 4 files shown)


eng/Version.Details.xml
```diff
@@ -63,5 +63,9 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>044fa99d22e576fcfaf63f09ee9bba821855dddb</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.NETCore.Platforms" Version="3.0.0-preview.19073.11">
```
dougbu: Will we just have to remember to update this and https://github.com/aspnet/Extensions/pull/1021/files#diff-8b8f08ffbf7b863fb3700c1718eeb4cbR19 simultaneously? Seems perilous&hellip;

eng/Version.Details.xml
```diff
@@ -63,5 +63,9 @@
...
+      <Uri>https://github.com/dotnet/corefx</Uri>
+      <Sha>0000</Sha>
```
dougbu: Don't we need the correct SHA for the chosen version?

natemcmaster: It will be updated by the bot on the next run. Commit sha isn't actually used by anything (yet).

eng/Version.Details.xml
natemcmaster: The bot should take care of updating both files.

PR #1023, opened 2019-01-30
Kill processes after completing build and before publishing artifacts

dougbu:
- aspnet/AspNetCore-Internal#1735


dougbu:
Oops, meant to update my comments: It's artifact publication that caused aspnet/AspNetCore-Internal#1735. Doesn't look like a sign check is enabled in this repo.

dougbu:
> Are we confident that no process will be holding a lock on binlong in a failure case?

Thinking more about this, I suspect the `Copy-Item` attempt could also fail when the job is cancelled. The  aspnet/AspNetCore-Internal#1735 failure occured because a later build step attempted to upload the binary log before it killed the `msbuild` process. Should change the `condition` from `failure()` to a check that `TASK_STATUS` is `Falure' or `SucceededWithIssues`. Another option would be to kill all `msbuild` processes before attempting the copy.

> Uh oh, &hellip;

The other way 'round here would be to add a different step to `PublishBuildArtifacts@1` which kills `msbuild` processes. Something similar to a step we have in most build configurations on http://aspnetci/ **Thoughts?**

dougbu:
:up::date: to change the way we avoid locked files completely. Bit more like our Teamcity infrastructure.

And, no, I couldn't find a built-in task that does anything similar.

natemcmaster:
> line 1: eng/scripts/KillProcesses.sh: Permission denied

I think you need to `chmod +x`.

eng/scripts/KillProcesses.ps1
```diff
@@ -0,0 +1,14 @@
...
+taskkill /T /F /IM testhost.exe
+taskkill /T /F /IM iisexpress.exe
+taskkill /T /F /IM iisexpresstray.exe
+taskkill /T /F /IM w3wp.exe
+taskkill /T /F /IM msbuild.exe
+taskkill /T /F /IM vbcscompiler.exe
+taskkill /T /F /IM git.exe
+taskkill /T /F /IM vctip.exe
+taskkill /T /F /IM chrome.exe
+iisreset /restart
```
natemcmaster:
```suggestion
iisreset /restart
exit /b 0
```

dougbu:
New failure is completely different:
```
2019-01-31T05:10:16.3603515Z F:\vsagent\64\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19072.5\tools\Publish.proj(108,5): error : Package already exists: Microsoft.DotNet.Analyzers.Async 3.0.0-preview.
2019-01-31T05:10:16.3604684Z F:\vsagent\64\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19072.5\tools\Publish.proj(108,5): error : Item 'F:\vsagent\64\s\artifacts\packages\Release\NonShipping\Microsoft.Extensions.Logging.Analyzers.3.0.0-preview.symbols.nupkg' already exists at 'assets/symbols/Microsoft.Extensions.Logging.Analyzers.3.0.0-preview.symbols.nupkg'
2019-01-31T05:10:16.3606023Z F:\vsagent\64\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19072.5\tools\Publish.proj(108,5): error : Item 'F:\vsagent\64\s\artifacts\packages\Release\NonShipping\Microsoft.DotNet.Analyzers.Async.3.0.0-preview.symbols.nupkg' already exists at 'assets/symbols/Microsoft.DotNet.Analyzers.Async.3.0.0-preview.symbols.nupkg'
2019-01-31T05:10:16.3607331Z F:\vsagent\64\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19072.5\tools\Publish.proj(108,5): error : A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
2019-01-31T05:10:16.3608465Z F:\vsagent\64\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19072.5\tools\Publish.proj(108,5): error :    at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
...
```

May be two separate issues: (a) packages are not picking up build number suffix and (b) publish target was down or inaccessible.

eng/scripts/KillProcesses.ps1
dougbu: Tried and reverted this suggestion. (Would have worked if this were a `cmd` file.)

dougbu:
@natemcmaster while #1043 remains open and tests continue to fail or hang (worse than flaky at the moment), what are your thoughts on retargeting this to `release/2.1`?

natemcmaster:
> Should I back-port this feature to the release/2.1 and release/2.2 branches?

Yeah, probably worth doing.

PR #1028, opened 2019-01-30
Switch Configuration.Json to use System.Json.Text

HaoK:
Things compile and run, but half the tests are failing, first main difference is that ' is not allowed anymore, which we were using heavily in our test json, i.e.

```C#
            // Causes the parser to throw, needed to replace with ' => "
            var json = @"{
                'ip': [
                    '1.2.3.4',
                    '7.8.9.10',
                    '11.12.13.14'
                ]
            }";
```

Still investigating the test failures, but figured it was worth getting the initial PR open

FYI @ajcvickers @ahsonkhan


eng/Dependencies.props
```diff
@@ -41,6 +41,7 @@
     <LatestPackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.6" />
     <LatestPackageReference Include="Microsoft.Azure.KeyVault" Version="2.3.2" />
     <LatestPackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
+    <LatestPackageReference Include="Microsoft.Bcl.Json.Sources" Version="$(MicrosoftBclJsonSourcesPacakgeVersion)" />
```
natemcmaster: Can you put this in the section above, with the rest of the System.* packages?

eng/Versions.props
```diff
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <MicrosoftNETCoreAppPackageVersion>3.0.0-preview-27324-5</MicrosoftNETCoreAppPackageVersion>
     <MicrosoftWin32RegistryPackageVersion>4.6.0-preview.19073.11</MicrosoftWin32RegistryPackageVersion>
+    <MicrosoftBclJsonSourcesPacakgeVersion>4.6.0-preview.19073.11</MicrosoftBclJsonSourcesPacakgeVersion>
```
natemcmaster: To add this to dependency flow, you also need to add an entry to eng/Versions.Details.xml. Without an entry in Versions.Details.xml, automation won't update this value.

src/Configuration/Config.Json/src/Microsoft.Extensions.Configuration.Json.csproj
```diff
@@ -5,6 +5,10 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <PackageTags>$(PackageTags);json</PackageTags>
     <IsShipping>true</IsShipping>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <IsPackable>true</IsPackable>
```
natemcmaster:
```suggestion
```
You can remove this. It should already true.

src/Configuration/Config.Json/src/Microsoft.Extensions.Configuration.Json.csproj
```diff
@@ -5,6 +5,10 @@
...
+    <NoWarn>$(NoWarn);3021</NoWarn>
+    <LangVersion>latest</LangVersion>
```
natemcmaster:
```suggestion
```
You can remove this too. The repo-wide default was update to 7.3 a few commits ago.

Tratcher:
So far so good.

eng/Versions.props
natemcmaster: By the way, I've written step by step instructions what what to do now when adding new dependencies. See https://github.com/aspnet/Extensions/blob/master/docs/ReferenceResolution.md#example-adding-a-new-dependency

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
```diff
@@ -19,89 +17,70 @@ internal class JsonConfigurationFileParser
...
+                case JsonValueType.String:
+                case JsonValueType.True:
+                case JsonValueType.False:
+                case JsonValueType.Null:
+                    var key = _currentPath;
+                    if (_data.ContainsKey(key))
+                    {
+                        throw new FormatException(Resources.FormatError_KeyIsDuplicated(key));
+                    }
+                    _data[key] = value.ToString(CultureInfo.InvariantCulture);
```
HaoK: FYI @ajcvickers there is no overload that takes culture info in the new stack currently

HaoK:
@ajcvickers tests now are all passing, the trailing comma was a big issue in the new common provider tests you added, smallest change was to run it thru JSON.NET to clean it up.  Thoughts?

HaoK:
Added NewtonsoftJson config package that is a copy of the old code, basically just has a new Newtonsoft prefix.

`configbuilder.AddNewtonsoftJsonFile()`

Tratcher:

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
```diff
@@ -19,89 +17,70 @@ internal class JsonConfigurationFileParser
...
         private IDictionary<string, string> ParseStream(Stream input)
         {
             _data.Clear();
-            _reader = new JsonTextReader(new StreamReader(input));
-            _reader.DateParseHandling = DateParseHandling.None;
-
-            var jsonConfig = JObject.Load(_reader);
 
-            VisitJObject(jsonConfig);
+            using (var reader = new StreamReader(input))
```
ahsonkhan: Can you pass the stream directly to the `JsonDocument.Parse` method instead?

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
```diff
@@ -19,89 +17,70 @@ internal class JsonConfigurationFileParser
...
-
-        private void VisitToken(JToken token)
-        {
-            switch (token.Type)
-            {
-                case JTokenType.Object:
-                    VisitJObject(token.Value<JObject>());
+            switch (value.Type) {
+                case JsonValueType.Object:
+                    VisitElement(value);
```
ahsonkhan: Are we generally OK with recursive calls like this?

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
```diff
@@ -19,89 +17,70 @@ internal class JsonConfigurationFileParser
...
-            _reader.DateParseHandling = DateParseHandling.None;
-
-            var jsonConfig = JObject.Load(_reader);
 
-            VisitJObject(jsonConfig);
+            using (var reader = new StreamReader(input))
+            using (JsonDocument doc = JsonDocument.Parse(reader.ReadToEnd(),
+                new JsonReaderOptions { CommentHandling = JsonCommentHandling.Skip }))
+            {
+                VisitElement(doc.RootElement);
```
ahsonkhan: What if the root element is not of type object (for example array or single value)? Is it fine that we would just throw on enumerate object since such an input is unexpected?

eng/Versions.props
```diff
@@ -12,6 +12,7 @@
     <!-- Packages from dotnet/core-setup -->
     <MicrosoftNETCoreAppPackageVersion>3.0.0-preview-27404-9</MicrosoftNETCoreAppPackageVersion>
     <!-- Packages from dotnet/corefx -->
+    <MicrosoftBclJsonSourcesPacakgeVersion>4.6.0-preview.19105.1</MicrosoftBclJsonSourcesPacakgeVersion>
```
natemcmaster:
```suggestion
    <MicrosoftBclJsonSourcesPackageVersion>4.6.0-preview.19105.1</MicrosoftBclJsonSourcesPackageVersion>
```

This variable has a typo everywhere it's used. The bot won't be able to update automatically with this typo.

FYI - this seems to be a common problem, so I'm adding a test for this in https://github.com/aspnet/Extensions/pull/1089

HaoK: Yeah I copied this from the SignalR pr that had a typo I think, so that's why its been common :)

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
HaoK: Yeah we would have thrown an exception before as well, so this would just be a different exception

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
HaoK: The flow of the code isn't really changed from before which was recursive, config should only be loaded once on startup for an app typically, or whenever the file changes, which should be pretty infrequent.

HaoK:
@natemcmaster what's the proper way to mark that the removed reference is intentional?

```

D:\a\1\s\eng\targets\ResolveReferences.targets(135,5): error PKG0001: Package references changed since the last release. This could be a breaking change. References removed: [D:\a\1\s\src\Configuration\Config.Json\src\Microsoft.Extensions.Configuration.Json.csproj]
D:\a\1\s\eng\targets\ResolveReferences.targets(135,5): error PKG0001:  - Newtonsoft.Json [D:\a\1\s\src\Configuration\Config.Json\src\Microsoft.Extensions.Configuration.Json.csproj]
    0 Warning(s)
    1 Error(s)
```

HaoK:
Should I just add NoWarn PKG0001?

natemcmaster:
Yeah, `NoWarn` PKG0001 is fine for now. I'll clean it up later with a more granular solution for these kinds of changes.

HaoK:
Looks like code check is failing due to diffs in generated files, I'm not sure what I need to update for the generated code to match:

```
-- a/eng/ProjectReferences.props
+++ b/eng/ProjectReferences.props
@@ -20,7 +20,7 @@
     <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.Ini" ProjectPath="$(RepoRoot)src\Configuration\Config.Ini\src\Microsoft.Extensions.Configuration.Ini.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config.Ini\ref\Microsoft.Extensions.Configuration.Ini.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.Json" ProjectPath="$(RepoRoot)src\Configuration\Config.Json\src\Microsoft.Extensions.Configuration.Json.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config.Json\ref\Microsoft.Extensions.Configuration.Json.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.KeyPerFile" ProjectPath="$(RepoRoot)src\Configuration\Config.KeyPerFile\src\Microsoft.Extensions.Configuration.KeyPerFile.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config.KeyPerFile\ref\Microsoft.Extensions.Configuration.KeyPerFile.csproj" />
-    <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.NewtonsoftJson" ProjectPath="$(RepoRoot)src\Configuration\Config.NewtonsoftJson\src\Microsoft.Extensions.Configuration.NewtonsoftJson.csproj" />
+    <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.NewtonsoftJson" ProjectPath="$(RepoRoot)src\Configuration\Config.NewtonsoftJson\src\Microsoft.Extensions.Configuration.NewtonsoftJson.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config.NewtonsoftJson\ref\Microsoft.Extensions.Configuration.NewtonsoftJson.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.UserSecrets" ProjectPath="$(RepoRoot)src\Configuration\Config.UserSecrets\src\Microsoft.Extensions.Configuration.UserSecrets.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config.UserSecrets\ref\Microsoft.Extensions.Configuration.UserSecrets.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration.Xml" ProjectPath="$(RepoRoot)src\Configuration\Config.Xml\src\Microsoft.Extensions.Configuration.Xml.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config.Xml\ref\Microsoft.Extensions.Configuration.Xml.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Configuration" ProjectPath="$(RepoRoot)src\Configuration\Config\src\Microsoft.Extensions.Configuration.csproj" RefProjectPath="$(RepoRoot)src\Configuration\Config\ref\Microsoft.Extensions.Configuration.csproj" />
diff --git a/src/Configuration/Config.Json/ref/Microsoft.Extensions.Configuration.Json.csproj b/src/Configuration/Config.Json/ref/Microsoft.Extensions.Configuration.Json.csproj
index edf64e38b..e01786407 100644
--- a/src/Configuration/Config.Json/ref/Microsoft.Extensions.Configuration.Json.csproj
+++ b/src/Configuration/Config.Json/ref/Microsoft.Extensions.Configuration.Json.csproj
@@ -1,13 +1,18 @@
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
   </PropertyGroup>
     <Compile Include="Microsoft.Extensions.Configuration.Json.netstandard2.0.cs" />
     <Reference Include="Microsoft.Extensions.Configuration"  />
     <Reference Include="Microsoft.Extensions.Configuration.FileExtensions"  />
-    <Reference Include="Newtonsoft.Json"  />
+    <Reference Include="System.Runtime.CompilerServices.Unsafe"  />
+  </ItemGroup>
+<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
+    <Compile Include="Microsoft.Extensions.Configuration.Json.netcoreapp3.0.cs" />
+    <Reference Include="Microsoft.Extensions.Configuration"  />
+    <Reference Include="Microsoft.Extensions.Configuration.FileExtensions"  />
   </ItemGroup>
 </Project>
```

Tratcher:
@pakrym ref assembly docs?

pakrym:
@Tratcher Working on it. 

@HaoK run `.\eng\scripts\GenerateReferenceAssemblies.ps1`

src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
HaoK: This causes some tests failures as it looks like the parser doesn't handle things like the byte order marker well, it results in parse errors with 0xEF

ahsonkhan: cc @bartonjs, maybe this is something we should document.

bartonjs: JsonDocument.Parse takes UTF-8.  Per The Unicode Consortium, there is no UTF-8 "BOM".  So that's actually an illegal start.

Tratcher: > JsonDocument.Parse takes UTF-8. Per The Unicode Consortium, there is no UTF-8 "BOM". So that's actually an illegal start.

@bartonjs that's not accurate. There is a BOM `0xEF,0xBB,0xBF`, it's just not required. The parser should allow for it.
https://www.unicode.org/versions/Unicode5.0.0/ch02.pdf page 36
"Use of a BOM is neither required nor recommended for UTF-8, but may
be encountered in contexts where UTF-8 data is converted from other encoding forms that
use a BOM or where the BOM is used as a UTF-8 signature"

bartonjs: Hm, clearly I've misremembered.  The current FAQ entry is pretty clear on "yes":

http://unicode.org/faq/utf_bom.html#bom5

> A: Yes, UTF-8 can contain a BOM.  ...

When I would have sworn that it said "this makes no sense, don't do it" (paraphrased).

@ahsonkhan Perhaps that means the right answer is that Utf8JsonReader should skip `0xEF, 0xBB, 0xBF` if that's the start of the data stream, since JsonDocument doesn't really have any understanding of what's going on (treat it as a special case of whitespace?)

ahsonkhan: According to the JSON grammar (http://json.org/), only white space can be inserted between tokens which can be ignored and the BOM doesn't count as white space. However, the JSON RFC let's the parser decide what to do in the presence of a BOM. Do we **really** need our reader to ignore the BOM? If so, I am find with adding that logic (likely an extra if check in `ReadFirstToken`).

https://tools.ietf.org/html/rfc8259
> Implementations MUST NOT add a byte order mark (U+FEFF) to the
   beginning of a networked-transmitted JSON text.  In the interests of
   interoperability, implementations that parse JSON texts MAY ignore
   the presence of a byte order mark rather than treating it as an
   error.

ahsonkhan: Discussing this further @GrabYourPitchforks, the layer that supports reading from a stream (or pipe) should strip out the BOM before passing the "text" data to the underlying reader. In this case, that should be JsonDocument The same would apply to the Deserializer. We shouldn't add the BOM stripping logic at the reader layer since we accept `ReadOnlySpan<byte>` which is meant to represent just the text data (without any metadata around it), and works as a stand-in for "utf8 string". Stream/etc. can contain this metadata.

cc @steveharter

HaoK:
@ahsonkhan i'm seeing some weird errors with respect to the exception strings from the json sources package, have you seen anything like this before?  

```
  at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object arg0)
   at System.SR.Format(String resourceFormat, Object p1) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/SR.cs:line 63
   at System.Text.Json.ThrowHelper.GetResourceString(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, String characters) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/ThrowHelper.cs:line 311
   at System.Text.Json.ThrowHelper.GetJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/ThrowHelper.cs:line 218
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/ThrowHelper.cs:line 212
   at System.Text.Json.Utf8JsonReader.HasMoreData() in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 377
   at System.Text.Json.Utf8JsonReader.SkipAllComments(Byte& marker) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 1482
   at System.Text.Json.Utf8JsonReader.ConsumeNextTokenUntilAfterAllCommentsAreSkipped(Byte marker) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 1550
   at System.Text.Json.Utf8JsonReader.ConsumeNextToken(Byte marker) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 1229
   at System.Text.Json.Utf8JsonReader.ConsumeNextTokenOrRollback(Byte marker) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 1192
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment() in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 303
   at System.Text.Json.Utf8JsonReader.Read() in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/Utf8JsonReader.cs:line 200
   at System.Text.Json.JsonDocument.Parse(ReadOnlySpan`1 utf8JsonSpan, Utf8JsonReader reader, MetadataDb& database, StackRowStack& stack) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/JsonDocument.cs:line 430
   at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 utf8Json, JsonReaderOptions readerOptions, Byte[] extraRentedBytes) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/JsonDocument.Parse.cs:line 166
   at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 json, JsonReaderOptions readerOptions) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/JsonDocument.Parse.cs:line 129
   at System.Text.Json.JsonDocument.Parse(String json, JsonReaderOptions readerOptions) in /_/.packages/microsoft.bcl.json.sources/4.6.0-preview.19109.6/contentFiles/cs/netstandard2.0/Microsoft.Bcl.Json.Sources/JsonDocument.Parse.cs:line 142
   at Microsoft.Extensions.Configuration.Json.JsonConfigurationFileParser.ParseStream(Stream input) in /_/src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs:line 28
   at Microsoft.Extensions.
```

BrennanConroy:
Is there anything blocking this PR?

HaoK:
Doesn't look like the resources are working properly on net472 (they result in null refs if you try to look at any of the error messages)  So we can skip/ignore those tests for now if we want to unblock the PR

ahsonkhan:
> Doesn't look like the resources are working properly on net472 (they result in null refs if you try to look at any of the error messages) So we can skip/ignore those tests for now if we want to unblock the PR

As a short term workaround, you need to add `<IsSharedSource>false</IsSharedSource>` to your source package reference. Thanks @ericstj for helping debug this!

This change was interfering with the source package consumption since its explicitly blocking inclusion of the build directory (which contains the resx file).
https://github.com/aspnet/Extensions/pull/1106

We should consider including targets/props from the /build/ directory as well.
https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files

cc @natemcmaster

natemcmaster:
Instead of setting IsSharedSource=false, can you update [this line](https://github.com/aspnet/Extensions/blob/master/eng/targets/ResolveReferences.targets#L92) to

```xml
<IncludeAssets Condition="'%(IsSharedSource)' == 'true'">ContentFiles;Build</IncludeAssets>
```

Microsoft.Bcl.Json.Sources is a shared source package, so setting IsSharedSource=false is not correct.

HaoK:
@ajcvickers everything is finally green, do you want to do a final review + signoff and we can merge and see if there are any other issues up the stack (aspnetcore)

src/Configuration/Config.NewtonsoftJson/ref/Microsoft.Extensions.Configuration.NewtonsoftJson.netstandard2.0.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Configuration
+{
+    public static partial class NewtonsoftJsonConfigurationExtensions
+    {
+        public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddNewtonsoftJsonFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, Microsoft.Extensions.FileProviders.IFileProvider provider, string path, bool optional, bool reloadOnChange) { throw null; }
```
ajcvickers: Is "NewtonsoftJson" how we are going to refer to this everywhere we need it? (i.e. in other packages.)

HaoK: I just went with what MVC did: https://github.com/aspnet/AspNetCore/tree/master/src/Mvc/Mvc.NewtonsoftJson

ajcvickers: Cool. Just wanted us to be consistent.

davidfowl:
@HaoK can you file an issue to use the reader instead of the DOM?

HaoK:
Sure https://github.com/aspnet/Extensions/issues/1195

PR #1029, opened 2019-01-31
Track long running tests for hosting

ryanbrandenburg:
There seems to be a hosting test that hangs, so lets make it tattle on itself.

### DIFF
--- a/src/Hosting/Hosting/test/Microsoft.Extensions.Hosting.Tests.csproj
+++ b/src/Hosting/Hosting/test/Microsoft.Extensions.Hosting.Tests.csproj
@@ -20,6 +20,7 @@
     <None Update="appSettings.json">
       <CopyToOutputDirectory>Never</CopyToOutputDirectory>
     </None>
+    <None Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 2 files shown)


Tratcher:
That doesn't seem to have worked
```
xUnit.net Console Runner v2.4.0 (64-bit Desktop .NET 4.5.2, runtime: 4.0.30319.42000)
  Discovering: Microsoft.Extensions.Hosting.Tests
  Discovered:  Microsoft.Extensions.Hosting.Tests
  Starting:    Microsoft.Extensions.Hosting.Tests
info: Microsoft.Extensions.Hosting.HostTests[0]
      Request starting
```

ryanbrandenburg:
Closing this since it didn't produce anything useful. @Tratcher will address whatever the root cause is in aspnet/AspNetCore-Internal#1743.

Tratcher:
@ryanbrandenburg it looks like the MSBuild verbosity needs to be at "normal" or higher for this to show up. Any idea where we control that?

ryanbrandenburg:
You can try setting it [like this](https://github.com/aspnet/Extensions/blob/master/eng/common/build.ps1#L5) but it doesn't seem that that variable is used in the later code, so it seems like it wouldn't work unless I missed something. Special note: Don't edit files in eng/common/*. Those need to remain direct copies of the files in Arcade, if we need to make a change to those files we should make it in Arcade and let it propagate to us.

PR #1037, opened 2019-01-31
Remove implicit references for non-test projects

### LINKED ISSUE #1036: DI.Spec.Tests have a package dependency on Microsoft.NET.Test.Sdk
The dependencies on Microsoft.Extensions.DependencyInjection.Specification.Tests changed as a result of converting to Arcade. We now have a package dependency on Microsoft.NET.Test.Sdk. We should remove this reference because it is not meant to end up in nuspec. Consumers of this package are expected to have Microsoft.NET.Test.Sdk in their project file.

```
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.Extensions.DependencyInjection.Abstractions" version="3.0.0-preview.19079.8" exclude="Build,Analyzers" />
        <dependency id="Microsoft.NET.Test.Sdk" version="15.9.0" exclude="Build,Analyzers" />
        <dependency id="xunit.assert" version="2.4.0" exclude="Build,Analyzers" />
        <dependency id="xunit.extensibility.core" version="2.4.0" exclude="Build,Analyzers" />
      </group>
```

natemcmaster:
Fix #1036

Workaround for https://github.com/dotnet/arcade/issues/1910 which adds implicit references to Spec.Tests which should not be used.

### DIFF
--- a/src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
+++ b/src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
@@ -7,8 +7,8 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>dependencyinjection;di</PackageTags>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <!-- This is actually a library of test projects that we ship to consumer, not a test project. -->
     <IsUnitTestProject>false</IsUnitTestProject>
-    <TestRunnerName></TestRunnerName>
     <IsShipping>true</IsShipping>
   </PropertyGroup>
 
@@ -17,6 +17,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!--
+      This intentionally does not reference 'xunit', 'xunit.core', or any runner packages.
+      XUnit recommends only using these packages which "extend" xunit. This allows consumers
+      to decide which type of xunit runner they want to use to run these tests.
+     -->
     <Reference Include="xunit.assert" />
     <Reference Include="xunit.extensibility.core" />
   </ItemGroup>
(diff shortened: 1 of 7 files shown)


src/DependencyInjection/DI.Specification.Tests/src/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
```diff
@@ -17,6 +17,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!--
+      This intentionally does not reference 'xunit', 'xunit.core', or any runner packages.
+      XUnit recommends only using xunit.extensibility.* an xunit.assert for packages which "extend" xunit.
```
dougbu: s/an/and/

src/TestingUtils/Microsoft.AspNetCore.Testing/src/Microsoft.AspNetCore.Testing.csproj
```diff
@@ -19,6 +20,15 @@
   <ItemGroup>
     <Reference Include="Microsoft.Win32.Registry" />
     <Reference Include="System.ValueTuple" />
+
+    <!--
+      This intentionally does not reference 'xunit', 'xunit.core', or any runner packages.
+      XUnit recommends only using xunit.extensibility.* an xunit.assert for packages which "extend" xunit.
```
dougbu: s/an/and/

src/TestingUtils/Microsoft.AspNetCore.Testing/src/Microsoft.AspNetCore.Testing.csproj
```diff
@@ -6,6 +6,7 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore</PackageTags>
+    <IsUnitTestProject>false</IsUnitTestProject>
```
dougbu: Suggest adding "This is actually a library of test projects &hellip;" comment here too

natemcmaster:
Apparently azure-pipelines is having issues today.

PR #1038, opened 2019-01-31
Make VS2019 a prerequisite, and add scripts to set required env vars

natemcmaster:
Changes:

* Update Extensions.sln to require VS 2019.
* Add two scripts (inspired by Python's virtualenv) which set required environment variables so you can use `dotnet` and launch VS Code on command line. I did this because Arcade puts dotnet core in "$repoRoot/.dotnet" by default now.

### DIFF
--- a/Extensions.sln
+++ b/Extensions.sln
@@ -2,7 +2,7 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.28508.60
-MinimumVisualStudioVersion = 15.0.26124.0
+MinimumVisualStudioVersion = 16.0.0.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "DiagnosticAdapter", "DiagnosticAdapter", "{13B8928D-1FE1-4B9D-8B2C-E3248612052B}"
 EndProject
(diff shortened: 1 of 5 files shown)


natemcmaster:
AFAIK the only thing Visual Studio and `dotnet.exe` reads from global.json is SDK version. The other data in that file is used by other tooling and scripts, like MSBuild and Arcade.

https://github.com/aspnet/Extensions/blob/8535a13651cde6d28ee7059c7d8e5ecb5f384f42/global.json#L2-L4

poke:
@natemcmaster Hey, I was wondering if the VS 2019 requirement is a hard requirement, or if there was a way to get this into a buildable state with VS 2017 too. I tried lowering the requirement in the sln and was able to load the solution properly but then I got an error that the _“current .NET SDK does not support targeting .NET Standard 2.0.”_ I did activate the environment for VS, so the 3.0 SDK should be there?

The reason I’m asking is because I don’t really want to waste another 5+ gigabytes on multiple machines for a VS that I’m not going to use otherwise. And I have been able to use 2017 for both Extensions and AspNetCore just fine before the move to Arcade. I just called `build.ps1` back then and launched VS from the console to inherit the environment variables the build script set up.

natemcmaster:
> I was wondering if the VS 2019 requirement is a hard requirement,

Yes, it is. We are also updating AspNetCore to require VS2019. (see https://github.com/aspnet/AspNetCore/pull/7005)

PR #1040, opened 2019-01-31
Fixed OptionsTest for different cultures

alefranz:
The test was failing for en-GB, as on the parsing side the invariant culture is used


HaoK:
We took a similar PR from someone else yesterday, But thanks for submitting this!  https://github.com/aspnet/Extensions/pull/1204

PR #1042, opened 2019-01-31
Update BaselineGenerator TargetFramework

alefranz:
As there is a `global.json` in the root that force the prelease version of .NET Core 3.0


natemcmaster:
@alefranz we must be on the same page. I also updated this project yesterday in https://github.com/aspnet/Extensions/pull/1044. I'm closing this now because this change is present in master now.

Thanks,
Nate

PR #1044, opened 2019-02-01
Add code check CI tests, add docs about ReferenceResolution, and other cleanups

natemcmaster:
Changes:
* Add a step which checks that generated code is up to date
* Copy over the documentation on how to work with `<reference>` 
* Fixup a broken reference in the .sln
* Fix the casing on Microsoft.Extensions.ValueStopwatch.Sources
* Remove some unused references and variables


eng/scripts/CodeCheck.ps1
```diff
@@ -0,0 +1,91 @@
...
+        $slnDir = Split-Path -Parent $_
+        $sln = $_
+        & dotnet sln $_ list `
+            | ? { $_ -ne 'Project(s)' -and $_ -ne '----------' } `
+            | % {
+                $proj = Join-Path $slnDir $_
+                if (-not (Test-Path $proj)) {
+                    LogError "Missing project. Solution references a project which does not exist: $proj. [$sln] "
+                }
+            }
```
dougbu: Suggest being more consistent with indentation in the two `foreach` blocks. (This is a bit harder to read that in needs to be.) The first `Foreach-Object` block has no indentation but the second has. It's fine if either the close curlies aren't laid out perfectly (as they aren't for the outer line continuation).

FYI I'm going to put `dotnet sln $_ list ...` into a function in my PowerShell profile 😃

src/Analyzers/Async/src/Microsoft.DotNet.Analyzers.Async.csproj
```diff
@@ -3,10 +3,11 @@
   <PropertyGroup>
     <Description>Analyzer package for async best practices.</Description>
     <VersionPrefix>$(ExperimentalVersionPrefix)</VersionPrefix>
-    <VerifyVersion>false</VerifyVersion>
     <TargetFramework>netstandard1.3</TargetFramework>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <BuildOutputTargetFolder>analyzers/dotnet/cs/</BuildOutputTargetFolder>
+      <!-- Analyzers cannot be referenced as a ProjectReference. -->
```
dougbu: You've got a `<ProjectReference />` from `test/Microsoft.DotNet.Analyzers.Async.Tests.csproj` to this project. Did you mean `as a Reference`?

src/Configuration/Config/test/Microsoft.Extensions.Configuration.Tests.csproj
```diff
@@ -6,11 +6,11 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Config.Binder\src\Microsoft.Extensions.Configuration.Binder.csproj" />
+    <ProjectReference Include="$(RepoRoot)src\Configuration\testassets\Test.Common\Microsoft.Extensions.Configuration.Test.Common.csproj" />
```
dougbu: Suggest being consistent -- `..\..\testasssets\`

src/TestingUtils/Internal.AspNetCore.Analyzers/src/Internal.AspNetCore.Analyzers.csproj
```diff
@@ -6,6 +6,10 @@
     <PackageTags>$(PackageTags);analyzers</PackageTags>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <BuildOutputTargetFolder>analyzers/dotnet/cs/</BuildOutputTargetFolder>
+    <!-- Analyzers cannot be referenced as a ProjectReference. -->
```
dougbu: Again, suggest "as a Reference"

eng/scripts/CodeCheck.ps1
natemcmaster: I too which the indentation was better. The auto-formatter in VSCode does this, so I'll manually undo it for now.

src/Analyzers/Async/src/Microsoft.DotNet.Analyzers.Async.csproj
natemcmaster: This is deliberate. Analyzers, like MSBuild tasks, are "special package" types. Although they carry assemblies, they don't end up in the `lib/` folder of a NuGet package. As a result, PackageRef and ProjectRef are not interchangeable, so I've made it an explicit ProjectRef.

natemcmaster: I'll update the comment to clarify.

dougbu: I understand the context but saying "PackageReference" here confuses the issue. My thought was just to say "as a Reference".

natemcmaster: I updated the comment. Let me know if the update is clearer.

PR #1048, opened 2019-02-01
Publish Microsoft.Interop.Js and Mono.WebAssembly.Interop

pranavkm:


eng/targets/Npm.Common.targets
```diff
@@ -0,0 +1,75 @@
+<Project DefaultTargets="Build" InitialTargets="_CheckForInvalidConfiguration">
+
+  <PropertyGroup>
+    <PackageVersion Condition=" '$(VersionSuffix)' != '' ">$(VersionPrefix)-$(VersionSuffix)</PackageVersion>
```
pranavkm: Identical to https://github.com/aspnet/AspNetCore/blob/master/eng/targets/Npm.Common.targets except the first 2 lines where the `PackageVersion` is calculated.

src/JSInterop/Microsoft.JSInterop.JS/src/tslint.json
```diff
@@ -0,0 +1,14 @@
+{
```
pranavkm: Copied this from SignalR's tslint. Filed https://github.com/aspnet/AspNetCore/issues/7197 to enable this because there's  a bunch of warnings that show up when turned on.

eng/targets/Npm.Common.targets
natemcmaster: This is going to use SemVer2 by default. Is that okay for NPM packages?

Btw, here is how Arcade computes this stuff https://github.com/dotnet/arcade/blob/9104c21492e1c6ed98105dd87388bca8f1f1abaa/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.props and https://github.com/dotnet/arcade/blob/c0451509373be4842e44d97fd6b79bf51064e2bc/src/Microsoft.DotNet.Build.Tasks.Packaging/src/build/Packaging.targets

pranavkm: Does look like semver v2 works with npm, although it does constraint the range specifier: https://github.com/npm/node-semver#prerelease-tags

PR #1049, opened 2019-02-01
Avoid clearing ResolvedServices

pakrym:
The race that was happening is the following:

1. IL compilation is kicked off for the service on the background thread
2. The container is disposed clearing singleton cache
3. As optimization IL compilation tries to resolve direct reference to a singleton service that it knows should already be resolved and cached
4. Because the cache is cleared new object is created triggering the Assert.

The solution is to keep service cache around and let GC clean it.

Fixes: https://github.com/aspnet/Extensions/issues/1043

### DIFF
--- a/src/DependencyInjection/DI/src/ServiceLookup/ServiceProviderEngineScope.cs
+++ b/src/DependencyInjection/DI/src/ServiceLookup/ServiceProviderEngineScope.cs
@@ -128,8 +128,9 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
                 _disposables = null;
 
                 // Not clearing ResolvedServices here because there might be a compilation running in background
-                // trying to get a cached singleton service instance and if it won't find it it will try to create a new one tripping the Debug.Assert in CaptureDisposable
-                // and leaking a Disposable object
+                // trying to get a cached singleton service instance and if it won't find 
+                // it it will try to create a new one tripping the Debug.Assert in CaptureDisposable
+                // and leaking a Disposable object in Release mode
             }
 
             return toDispose;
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/DI/src/ServiceLookup/ServiceProviderEngineScope.cs
```diff
@@ -127,7 +127,9 @@ private List<object> BeginDispose()
                 toDispose = _disposables;
                 _disposables = null;
 
-                ResolvedServices.Clear();
+                // Not clearing ResolvedServices here because there might be a compilation running in background
+                // trying to get a cached singleton service instance and if it won't find it it will try to create a new one tripping the Debug.Assert in CaptureDisposable
```
Tratcher: nit: line wrapping

src/DependencyInjection/DI/src/ServiceLookup/ServiceProviderEngineScope.cs
```diff
@@ -127,7 +127,10 @@ private List<object> BeginDispose()
                 toDispose = _disposables;
                 _disposables = null;
 
-                ResolvedServices.Clear();
+                // Not clearing ResolvedServices here because there might be a compilation running in background
+                // trying to get a cached singleton service instance and if it won't find 
+                // it it will try to create a new one tripping the Debug.Assert in CaptureDisposable
+                // and leaking a Disposable object in Release mode
```
halter73: What should happen in CaptureDisposable if the container is already disposed? Shouldn't we throw an ODE that can be observed by the top-level GetService() call?

pakrym: It shouldn't get called after dispose. We have ODE checks on all customer accessed entry points. This one is internal that that's why it was able to slip past those checks.

PR #1052, opened 2019-02-01
Produce reference assemblies and include them in the nuget package

natemcmaster:
This enables the C# compiler feature which produces reference assemblies automatically from the implementation code. Impact of this change:

* It adds a `ref/` folder to every .nupkg produce by this repo.
* Inner dev-loop should be faster. In theory, when you change code that doesn't touch public API, Visual Studio and command line builds are smart enough to avoid unnecessary work.
* The generated reference assemblies these packages will end up in the targeting pack for Microsoft.AspNetCore.App.

Note: this does not complete reference assembly work. 

TODO:
* Figure out how to produce reference assemblies with consistent assembly versions in aspnetcore while still consuming higher assembly version implementations (important for patching scenarios)
* Build reference assemblies from generated source (replacement for API check)
* Add tests for this somehow. I manually verified the packages, but we don't have NuGetPackageVerifier anymore, so we don't have a good way to prevent regressions in packaging structure.


natemcmaster:
Discussed this further in person. This is probably not the direction we want to go. There are a handful of requirements we had to tackle, and it's unclear if there is any benefit to putting refs in packages.

We do want to try using ProduceReferenceAssemblies for the sake of inner-dev loop perf though, but let's add that separately after we investigate generated reference assemblies from source.

Issue #771, opened 2018-05-15
Provide a reloadOnChange setting to AddAzureKeyVault


traffk-viet:
We have an ASP.NET core app that pulls secrets from KeyVault on startup. These secrets may change behind the scene and we would like to be able to get these updates.

I believe the current way of doing this is to have a thread run on a schedule and call Configuration.Reload(); But I think this is a fairly common use case and a better way is to have something similar to AddJsonFile and have a reloadOnChange parameter.

PR #1054, opened 2019-02-01
Provide a reloadOnChange setting to AddAzureKeyVault

HobbyProjects:
 - Added a polling mechanism to AzureKeyVaultConfigurationProvider to implement the reloadOnChange behavior 
- Added a unit test to cover the feature

Addresses #771  (in this specific format)


global.json
```diff
@@ -1,13 +1,13 @@
 {
   "sdk": {
     "version": "3.0.100-preview-009750"
-  },
+  },
```
omusavi: This file only has whitespace changes, should remove from PR

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -78,6 +108,43 @@ private async Task LoadAsync()
...
+                long? currentSecretLastUpdateTick = secret.Attributes.Updated.Value.Ticks;
+                long? loadedSecretLastUpdateTick = _loadedSecrets[secret.Identifier.Name].Updated.Value.Ticks;
+
+                if(currentSecretLastUpdateTick != null && loadedSecretLastUpdateTick != null)
+                {
+                    bool isKeyUpdateTimeDifferent = (loadedSecretLastUpdateTick != currentSecretLastUpdateTick);
+
+                    if (isKeyUpdateTimeDifferent)
+                    {
+                        return true;
```
omusavi: can't you just return isKeyUpdateTimeDifferent here? If its true then it will return true, otherwise it would return false anyways

src/Configuration/Config.AzureKeyVault/test/AzureKeyVaultConfigurationTest.cs
```diff
@@ -149,6 +149,47 @@ public void SupportsReload()
             Assert.Equal("Value2", provider.Get("Secret1"));
         }
 
+        [Fact]
+        public void SupportsReloadOnChange() {
```
omusavi: `{` on next line

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
HobbyProjects: We're trying to break out of the loop and return if any of the secrets have changed.

omusavi: Ah you're right, makes sense

global.json
HobbyProjects: Sounds good. Thanks for catching.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationExtensions.cs
```diff
@@ -164,12 +176,16 @@ private static async Task<string> GetTokenFromClientCertificate(string authority
...
+        /// <param name="reloadPollDelay">Number of milliseconds to wait inbetween each attempt at polling the Azure KeyVault for changes. Default is 10000 ms.</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
         public static IConfigurationBuilder AddAzureKeyVault(
             this IConfigurationBuilder configurationBuilder,
             string vault,
             KeyVaultClient client,
-            IKeyVaultSecretManager manager)
+            IKeyVaultSecretManager manager,
+            bool reloadOnChange = false,
+            int reloadPollDelay = 10000)
```
pakrym: Take TimeSpan instead of int. I don't think we need two arguments here. `TimeSpan.Zero` should be enough to signal that no reload is required.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -15,44 +16,72 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
...
-            _vault = vault;
-            _manager = manager;
+            _pollingTask = null;
+            _cancellationToken = new CancellationTokenSource();
+            _reloadOnChange = reloadOnChange;
+            _reloadPollDelay = reloadPollDelay;
+            _loadedSecrets = new Dictionary<string, SecretAttributes>();
+        }
+
+        ~AzureKeyVaultConfigurationProvider()
```
pakrym: Implement `IDisposable` instead of finalizer.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -78,6 +108,43 @@ private async Task LoadAsync()
...
+
+            if (_reloadOnChange && _pollingTask == null)
+            {
+                _pollingTask = PollForSecretChangesAsync(_reloadPollDelay, _cancellationToken.Token);
+            }
+        }
+
+        private async Task<bool> ShouldReloadAsync()
+        {
+            var secrets = await _client.GetSecretsAsync(_vault).ConfigureAwait(false);
```
pakrym: This doesn't seem to account for pagination. GetSecretsAsync returns a single page of secrets you have to call GetSecretsNextAsync to get more.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -15,44 +16,72 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
...
+        }
+
+        ~AzureKeyVaultConfigurationProvider()
+        {
+            _cancellationToken.Cancel();
         }
 
         public override void Load() => LoadAsync().ConfigureAwait(false).GetAwaiter().GetResult();
 
+        private async Task PollForSecretChangesAsync(int reloadPollDelay, CancellationToken ct)
```
pakrym: This method would reload all keys twice. Once to check if reload is required another to actually reload them. I think reload logic can be merged into LoadAsync method that would call              OnReload(); only if new keys/versions of keys are detected.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -78,6 +108,43 @@ private async Task LoadAsync()
...
+            }
+
+            foreach (var secret in secrets)
+            {
+                if (!_loadedSecrets.ContainsKey(secret.Identifier.Name))
+                {
+                    return true;
+                }
+
+                long? currentSecretLastUpdateTick = secret.Attributes.Updated.Value.Ticks;
```
pakrym: Is there a reason to use `Updated` vs `Version`?

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -78,6 +108,43 @@ private async Task LoadAsync()
...
+
+        private async Task<bool> ShouldReloadAsync()
+        {
+            var secrets = await _client.GetSecretsAsync(_vault).ConfigureAwait(false);
+            if (secrets.Count() != _loadedSecrets.Count())
+            {
+                return true;
+            }
+
+            foreach (var secret in secrets)
```
pakrym: This doesn't account for new keys being added.

src/Configuration/Config.AzureKeyVault/test/AzureKeyVaultConfigurationTest.cs
```diff
@@ -149,6 +149,48 @@ public void SupportsReload()
             Assert.Equal("Value2", provider.Get("Secret1"));
         }
 
+        [Fact]
+        public void SupportsReloadOnChange()
```
pakrym: This feature needs way more tests.  Some examples:

1. Token doesn't fire when versions didn't change
2. Reload is multi page aware
3. Reload works when new key is added
4. Reload works if key is removed

`Thread.Sleep` should be avoided because it's not guaranteed that background task would finish in time making tests flaky. Prefer using the reload token with timeout.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
HobbyProjects: We found the version field was null when the secrets were updated. The "Updated" field however, reflected the update time.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationExtensions.cs
HobbyProjects: Thank you. Do we need to update any documents anywhere to specify what TimeSpan.Zero represents? Or updating the method docs would be sufficient?

Tratcher: Nullable TimeSpan would be preferable to having zero as a special value.

Tratcher:

ryancammer:
We're looking forward to this change. Is there any reason this hasn't been reviewed and merged yet? Anything I can help with? Thanks @pakrym!

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -7,77 +7,165 @@
...
+                allSecrets.AddRange(secretPage.ToList());
+                secretPage = secretPage.NextPageLink != null ?
+                    await _client.GetSecretsNextAsync(secretPage.NextPageLink).ConfigureAwait(false) :
+                     null;
+            } while (secretPage != null);
 
-                foreach (var secretItem in secrets)
+            var tasks = new List<Task<SecretBundle>>();
+            var newLoadedSecrets = new Dictionary<string, LoadedSecret>();
+            var oldLoadedSecrets = Interlocked.Exchange(ref _loadedSecrets, null);
```
Tratcher: Why Interlocked? It looks like _loadedSecrets is only used in this method and this method shouldn't be called concurrently (no?).

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -7,77 +7,165 @@
...
+                    oldLoadedSecrets.TryGetValue(secretId, out var existingSecret) &&
+                    existingSecret.IsUpToDate(secret.Attributes.Updated))
                 {
-                    data.Add(_manager.GetKey(task.Result), task.Result.Value);
+                    oldLoadedSecrets.Remove(secretId);
+                    newLoadedSecrets.Add(secretId, existingSecret);
                 }
+                else
+                {
+                    Console.WriteLine("Loading "+ secretId);
```
Tratcher: Remove

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
pakrym: In case someone calls `Reload` on configuration while automatic reload is in progress.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -7,77 +7,165 @@
...
+                await WaitForReload();
+                await LoadAsync();
+            }
+        }
 
-            var secrets = await _client.GetSecretsAsync(_vault).ConfigureAwait(false);
-            var tasks = new List<Task<SecretBundle>>(secrets.Count());
+        protected virtual async Task WaitForReload()
+        {
+            await Task.Delay(_reloadInterval.Value.Milliseconds, _cancellationToken.Token);
```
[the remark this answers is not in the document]
pakrym: Yes, but it would only surface up to the background task.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -7,77 +7,165 @@
...
         public override void Load() => LoadAsync().ConfigureAwait(false).GetAwaiter().GetResult();
 
-        private async Task LoadAsync()
+        private async Task PollForSecretChangesAsync()
         {
-            var data = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            while (!_cancellationToken.IsCancellationRequested)
+            {
+                await WaitForReload();
+                await LoadAsync();
```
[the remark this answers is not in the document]
pakrym: Good point

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
Tratcher: oldLoadedSecrets being null will give you inconsistent results on the overlapping refreshes. What if we guarded this at a higher level? If someone calls reload while an automatic refresh is in progress, let them await the one that's already happening and use those results.

pakrym: > oldLoadedSecrets being null will give you inconsistent results on the overlapping refreshes

It is eventually consistent. Both reloads would retrieve the correct set of secrets.

> What if we guarded this at a higher level

I'm not sure if this option is simpler.

src/Configuration/Config.AzureKeyVault/ref/Microsoft.Extensions.Configuration.AzureKeyVault.netstandard2.0.cs
```diff
@@ -16,6 +17,17 @@ public static partial class AzureKeyVaultConfigurationExtensions
 }
 namespace Microsoft.Extensions.Configuration.AzureKeyVault
 {
+    public partial class AzureKeyVaultConfigurationOptions
+    {
+        public AzureKeyVaultConfigurationOptions() { }
```
Tratcher: Do you need the no-paramater constructor? I'd think it would cause people to not discover the others.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationExtensions.cs
```diff
@@ -188,12 +153,42 @@ private static async Task<string> GetTokenFromClientCertificate(string authority
                 throw new ArgumentNullException(nameof(manager));
```
Tratcher: Why didn't this overload get consolidated?

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -7,77 +7,171 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.KeyVault;
 using Microsoft.Azure.KeyVault.Models;
+using System.Threading;
```
Tratcher: Sort

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationExtensions.cs
```diff
@@ -150,11 +116,10 @@ private static async Task<string> GetTokenFromClientCertificate(string authority
             string vault,
             IKeyVaultSecretManager manager)
         {
-            var azureServiceTokenProvider = new AzureServiceTokenProvider();
-            var authenticationCallback = new KeyVaultClient.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback);
-            var keyVaultClient = new KeyVaultClient(authenticationCallback);
-            
-            return AddAzureKeyVault(configurationBuilder, vault, keyVaultClient, manager);
```
Tratcher: You used to check if manager == null on this code path.

src/Configuration/Config.AzureKeyVault/ref/Microsoft.Extensions.Configuration.AzureKeyVault.netstandard2.0.cs
pakrym: I'd like to leave the ability to set everything manually without going through any custom logic

ryancammer:
Question: is the reloadOnChange functionality slated for the .NET Core 3.0 release, or will there be a 2.2.x release in the interim?

pakrym:
It would only go into 3.0 release.

ryancammer:
Gotcha. TY.

PR #1056, opened 2019-02-02
Remove CompactOnMemoryPressure

JunTaoLuo:
Addresses https://github.com/aspnet/Extensions/issues/1027.

@natemcmaster @ryanbrandenburg We used to have to update a breakingchanges.netcore.json file but I don't see these files anymore. Have we started to use the Arcade tool yet? I don't see any progress on https://github.com/aspnet/BuildTools/issues/747.


natemcmaster:
This repo no longer uses KoreBuild and API check, so there is no JSON file for you to update. And no, we haven't made much progress on the replacement in Arcade repos.

PR #1059, opened 2019-02-04
Kill processes after completing build and before publishing artifacts

dougbu:
- aspnet/AspNetCore-Internal#1735
- cherry pick e2bb8ac1a0 from `master`
  - change `default-build.yml` instead of `azure-pipelines.yml`


eng/scripts/KillProcesses.ps1
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env pwsh -c
```
natemcmaster: This could be removed. I wouldn't expect this script to run on Linux or MacOs.

PR #1060, opened 2019-02-04
WIP: Add components check step to builds

dougbu:
- aspnet/AspNetCore-Internal#1254


dougbu:
We've decided to start with aspnet/AspNetCore. Also need to change (remove) a few settings because the new task failed though the CI builds succeeded.

PR #1064, opened 2019-02-04
Collect dumps on failing memory cache capacity tests

JunTaoLuo:
Adding dumps to help debug https://github.com/aspnet/AspNetCore-Internal/issues/1741. Since all branches are equally likely to see the failure, I'm going to add this to all of 2.1, 2.2 and master.


natemcmaster:
Removing myself as a reviewer. I don't have any context on this one, but trust @Tratcher's judgment.

JunTaoLuo:
Actually I can only do this on master. `CollectDump` was only added in 3.0

PR #1065, opened 2019-02-04
Update HealthCheckPublisherOptions.cs

glennc:
This seems wrong.

Presumably fixes: https://github.com/aspnet/Extensions/issues/1041


guardrex:
For the docs, will this be patched for 2.2, or will it exclusively roll out for 3.0+?

glennc:
Issues get fixed in 3.0 by default and then there is a process for choosing them for back-porting. I don't know if it'll get moved back or not yet.

Issue #779, opened 2018-07-26
Improve binding error message


yfakariya:
`ConfigurationBinder` classes binding logic only says that there is a specific error in somewhere -- it does not report any information to specify what value was wrong.

It is very convenience that exceptions thrown by `ConfigurationBinder` contains path information where binding error was occurred.

Note: Personally, I tried to implement path reporting with `IConfigratuonSection.Path` property, and it worked fine.

HaoK:
Fixed by https://github.com/aspnet/Extensions/pull/1819

PR #1077, opened 2019-02-05
Include the path to the configuration value in conversion error messages

mvestergaard:
It is currently very difficult to figure out which value causes errors when using configuration binding.

This change will include the path to the value in error messages.

Addresses #779

### DIFF
--- a/src/Configuration/Config.Binder/test/ConfigurationBinderTests.cs
+++ b/src/Configuration/Config.Binder/test/ConfigurationBinderTests.cs
@@ -387,13 +387,13 @@ namespace Microsoft.Extensions.Configuration.Binder.Test
             Assert.NotNull(exception.InnerException);
             Assert.NotNull(getException.InnerException);
             Assert.Equal(
-                Resources.FormatError_FailedBinding(IncorrectValue, type),
+                Resources.FormatError_FailedBinding(IncorrectValue, "Value", type),
                 exception.Message);
             Assert.Equal(
-                Resources.FormatError_FailedBinding(IncorrectValue, type),
+                Resources.FormatError_FailedBinding(IncorrectValue, "Value", type),
                 getException.Message);
             Assert.Equal(
-                Resources.FormatError_FailedBinding(IncorrectValue, type),
+                Resources.FormatError_FailedBinding(IncorrectValue, "Value", type),
                 getValueException.Message);
         }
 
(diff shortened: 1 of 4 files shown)


mvestergaard:
I'm not sure why CI is failing. It doesn't seem related to my changes.

src/Configuration/Config.Binder/test/ConfigurationBinderTests.cs
```diff
@@ -387,13 +387,13 @@ public void ConsistentExceptionOnFailedBinding(Type type)
             Assert.NotNull(exception.InnerException);
             Assert.NotNull(getException.InnerException);
             Assert.Equal(
-                Resources.FormatError_FailedBinding(IncorrectValue, type),
+                Resources.FormatError_FailedBinding(IncorrectValue, "Value", type),
```
HaoK: Can you add a test that demonstrates what this change actually would look like with a path?

mvestergaard: Doesn't it already demonstrate it with this test?
Using an in memory collection as the configuration provider, it's kind of forced to be a flat key/value map, so adding `A:More:Elaborate:Path` doesn't really prove anything. It's basically the same thing.
But if you disagree, I can add it.

mvestergaard: Or are you looking for an example of binding to a deeper object graph?

HaoK: Yes, a deeper path test that demonstrates how this would be useful

mvestergaard: Test added

Tratcher:

mvestergaard:
Done

Eilon:
@HaoK / @ajcvickers - is this ready to merge?

ajcvickers:
@HaoK Can you review this an merge if you think it makes sense. (Note I don't think the path can contain sensitive data, so including it should be fine. The value itself should not be in the exception message, since it could be sensitive.)

HaoK:
Had to rebase, merged as part of https://github.com/aspnet/Extensions/pull/1819  Thanks @mvestergaard !

Issue #809, opened 2018-06-26
Add support for running Generic Host based applications as a Windows Service


feco93:
Currently RunAsService depends on IWebHost.  It should be depend on IHost interface, because it would be great if anyone could write windows service without the HTTP infrastructure.

There is an SO post, which about this feature request: 
https://stackoverflow.com/questions/50848141/netcore-2-1-generic-host-as-a-service

Tratcher:
Try to build in all the workarounds from https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-2.2
- current directory
- only register the service lifetime if in a service, otherwise console
- AddEventLog?

Tratcher:
@davidfowl AddEventLog turns out to be very useful, but it's also awkward to include by default because you can only configure it in the AddEventLog call. How about we add that to the docs instead?

The ContentRootPath workaroud isn't required, we already set it in generic host.
`ContentRootPath = ResolveContentRootPath(_hostConfiguration[HostDefaults.ContentRootKey], AppContext.BaseDirectory),`
We don't want to set CurrentDirectory, right?

Detecting if we're in a service turns out to be quite difficult...

davidfowl:
We can consider turning this on by default in CreateDefaultBuilder if running on windows and for errors only (it's useful for ASP.NET Core applications as well).

davidfowl:
> The ContentRootPath workaroud isn't required, we already set it in generic host.
ContentRootPath = ResolveContentRootPath(_hostConfiguration[HostDefaults.ContentRootKey], AppContext.BaseDirectory),
We don't want to set CurrentDirectory, right?

No, I don't think we want to do that (even though we're doing it in the IIS case). We can doc it though.

stevejgordon:
@davidfowl @Tratcher - Just FYI, my blog post on achieving this (sort of) with the code from Microsoft.AspNetCore.Hosting.WindowsServices is one of the highest viewed posts monthly at the moment, so it would suggest quite a high customer interest in this feature. Not that I use it personally as I'm all about the Linux containers; I'm happy to update the information or add a new post once something baked in is available in the previews.

PR #1083, opened 2019-02-06
Add Hosting.WindowsService package

Tratcher:
#809 This takes the sample we had for windows service integration and makes it an official package.
I added some opt in/out config switches and a little bit of auto detection. It turns out there's not a solid mechanic for identifying you're running in a service.

@glennc


src/Hosting/WindowsService/src/ServiceBaseLifetime.cs
```diff
@@ -1,41 +1,42 @@
...
         public Task WaitForStartAsync(CancellationToken cancellationToken)
         {
             cancellationToken.Register(() => _delayStart.TrySetCanceled());
-            ApplicationLifetime.ApplicationStopping.Register(Stop);
+            ApplicationLifetime.ApplicationStarted.Register(() =>
+            {
+                Logger.LogInformation("Application started. Hosting environment: {envName}; Content root path: {contentRoot}",
+                    Environment.EnvironmentName, Environment.ContentRootPath);
+            });
+            ApplicationLifetime.ApplicationStopping.Register(() =>
```
davidfowl: Remind me why we need this and why handling Stop isn't enough?

eng/ProjectReferences.props
```diff
@@ -30,6 +30,7 @@
     <ProjectReferenceProvider Include="Microsoft.Extensions.Diagnostics.HealthChecks" ProjectPath="$(RepoRoot)src\HealthChecks\HealthChecks\src\Microsoft.Extensions.Diagnostics.HealthChecks.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Hosting.Abstractions" ProjectPath="$(RepoRoot)src\Hosting\Abstractions\src\Microsoft.Extensions.Hosting.Abstractions.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Hosting" ProjectPath="$(RepoRoot)src\Hosting\Hosting\src\Microsoft.Extensions.Hosting.csproj" />
+    <ProjectReferenceProvider Include="Microsoft.Extensions.Hosting.WindowsService" ProjectPath="$(RepoRoot)src\Hosting\WindowsService\src\Microsoft.Extensions.Hosting.WindowsService.csproj" />
```
davidfowl: Microsoft.Extensions.Hosting.WindowsServices? Does it need an s? @glennc ?

src/Hosting/WindowsService/src/ServiceBaseLifetime.cs
Tratcher: So you can initiate shutdown from within the app.

davidfowl: What case do you need this?

https://github.com/aspnet/Extensions/blob/b560068287b890594046de034db2c28c278f19e9/src/Hosting/Abstractions/src/HostingAbstractionsHostExtensions.cs#L97-L106

Is it only useful when you call Start on the host?

davidfowl:
I think you need to skip tests on other platforms

src/Hosting/WindowsService/src/Microsoft.Extensions.Hosting.WindowsService.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>.NET Core hosting and startup infrastructures for Windows Services.</Description>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
natemcmaster:
```suggestion
    <TargetFramework>netstandard2.0</TargetFramework>
```

src/Hosting/WindowsService/src/Microsoft.Extensions.Hosting.WindowsService.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>.NET Core hosting and startup infrastructures for Windows Services.</Description>
```
natemcmaster: This isn't exclusively for .NET Core, right? You could use it on .NET Framework too.

JunTaoLuo:
```
2019-02-07T17:36:26.5989175Z /home/vsts/work/1/s/.dotnet/sdk/3.0.100-preview-010184/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(294,5): error NETSDK1074: The application host executable will not be customized because adding resources requires that the build be performed on Windows (excluding Nano Server). [/home/vsts/work/1/s/src/Hosting/samples/GenericHostSample/GenericHostSample.csproj]
```
?

xqrzd:
Excited to see this released as an official package. A while ago I grabbed ServiceBaseLifetime and used [this](https://gist.github.com/xqrzd/a799487f201d91ad6d1a0c67159e234b) to determine if the app is running as a service. It checks that the parent process is `services`, and that the parent is running in session 0.

```csharp
bool runAsService = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
    ParentProcessUtilities.IsChildOfServices()
```

davidfowl:
@xqrzd Great! Thanks for this! @Tratcher this looks like a decent heuristic to use.

src/Hosting/WindowsService/src/ServiceBaseLifetime.cs
Tratcher: That one covers CTL+C, service stop, applicationLifetime.StopApplication(), etc..

davidfowl: I'm confused. What do you mean? What case does this cover that that one does not? The reason I'm asking is because I wouldn't expect every lifetime to do this (it's not part of the contract which means there's s design flaw somewhere)

Tratcher: Stopping can be called from a lot of places, some of which are circular. The various lifetimes also have things they want to do when the app is stopping, such as log, so they listen to the same event as they may or may not have triggered it.

Here are a few flows:
- Console lifetime, CTL+C or ProcessExit -> ApplicationLifetime.StopApplication() -> unblock WaitForShutdownAsync -> host.StopAsync
- ServiceBaseLifetime service Stop -> ApplicationLifetime.StopApplication() -> unblock WaitForShutdownAsync -> host.StopAsync
- Application code -> ApplicationLifetime.StopApplication() -> unblock WaitForShutdownAsync -> host.StopAsync -> ServiceBaseLifetime.StopAsync -> ServiceBase.Stop
 - Full control Program.Main - Doesn't use WaitForShutdown, calls Host.StopAsync directly -> ServiceBaseLifetime.StopAsync -> ServiceBase.Stop

Ok, so ServiceBaseLifetime might not need ApplicationStopping to call Stop, but it still wants it for logging just like ConsoleLifetime because ApplicationStopping fires at the start of the shutdown process.

src/Hosting/WindowsService/src/Microsoft.Extensions.Hosting.WindowsService.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <Description>.NET hosting infrastructure for Windows Services.</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>hosting</PackageTags>
+    <IsShipping>true</IsShipping>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.Extensions.Hosting.Abstractions" />
+    <Reference Include="System.ServiceProcess.ServiceController" />
```
JunTaoLuo: What happens on non-windows systems? When I try to run the tests I see a PlatformNotSupportedException when the ServiceBase constructor is called.

Tratcher: The tests are marked with OSSkipCondition's. Any idea why those wouldn't be working?

src/Hosting/WindowsService/src/ServiceBaseLifetime.cs
```diff
@@ -1,41 +1,41 @@
...
     public class ServiceBaseLifetime : ServiceBase, IHostLifetime
     {
         private TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>();
 
-        public ServiceBaseLifetime(IApplicationLifetime applicationLifetime)
+        public ServiceBaseLifetime(IHostingEnvironment environment, IApplicationLifetime applicationLifetime, ILoggerFactory loggerFactory)
         {
+            Environment = environment ?? throw new ArgumentNullException(nameof(environment));
             ApplicationLifetime = applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime));
+            Logger = loggerFactory.CreateLogger("Microsoft.Hosting.Lifetime");
```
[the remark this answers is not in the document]
davidfowl: No.

src/Hosting/WindowsService/test/UseServiceBaseLifetimeTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            using (host)
+            {
+                var lifetime = host.Services.GetRequiredService<IHostLifetime>();
+                Assert.IsType<ConsoleLifetime>(lifetime);
+            }
+        }
+
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX, "Only available on Windows")]
```
JunTaoLuo: Remove the string. It's being interpreted as a version string which causes the condition to fail.

JunTaoLuo:
```suggestion
        // Only available on Windows
        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX,)]
```

src/Hosting/WindowsService/test/UseServiceBaseLifetimeTests.cs
```diff
@@ -0,0 +1,105 @@
...
```
JunTaoLuo:
```suggestion
        // Only available on Windows
        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
```

src/Hosting/WindowsService/test/UseServiceBaseLifetimeTests.cs
Tratcher: *head->desk* Thank you. This is why we always use named parameters.

Tratcher:
@xqrzd @davidfowl note some of those APIs are unsupported / obsolete, but it looks like CreateToolhelp32Snapshot might work.

Tratcher:
The new detection heuristic works well (using a different API than the one suggested).

When testing I realized that CreateDefaultBuilder is overriding ContentRoot to CurrentDirectory so we're going to have to overwrite it back. Unfortunately this precludes having a config based opt-in/out switch since ContentRoot is best set before config is built. The new heuristic might be good enough to avoid the config switch. We'll see.

Tratcher:
@pakrym ?
```
...
warning: LF will be replaced by CRLF in src/WebEncoders/ref/Microsoft.Extensions.WebEncoders.csproj.
The file will have its original line endings in your working directory.
error: Generated code is not up to date.
```

Tratcher:
```
home/vsts/work/1/s/.dotnet/sdk/3.0.100-preview-010184/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(294,5): error NETSDK1074: The application host executable will not be customized because adding resources requires that the build be performed on Windows (excluding Nano Server). [/home/vsts/work/1/s/src/Hosting/samples/GenericHostSample/GenericHostSample.csproj]
```
That's a new one...

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
```diff
@@ -26,34 +26,32 @@ public static class ServiceBaseLifetimeHostBuilderExtensions
...
+            {
+                return false;
+            }
+
+            var parrent = WindowsServices.Internal.Win32.GetParrentProcess();
+            if (parrent == null)
+            {
+                return false;
+            }
+            return parrent.SessionId == 0 && "services".Equals(parrent.ProcessName, StringComparison.OrdinalIgnoreCase);
```
davidfowl: "services.".Equals looks super weird. What about just using the static `string.Equals`.

src/Hosting/WindowsServices/src/ServiceBaseLifetime.cs
```diff
@@ -1,41 +1,41 @@
...
     public class ServiceBaseLifetime : ServiceBase, IHostLifetime
     {
         private TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>();
 
-        public ServiceBaseLifetime(IApplicationLifetime applicationLifetime)
+        public ServiceBaseLifetime(IHostingEnvironment environment, IApplicationLifetime applicationLifetime, ILoggerFactory loggerFactory)
         {
+            Environment = environment ?? throw new ArgumentNullException(nameof(environment));
             ApplicationLifetime = applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime));
+            Logger = loggerFactory.CreateLogger("Microsoft.Hosting.Lifetime");
```
JunTaoLuo: Curious why we went the LoggerFactory route instead of something like https://github.com/aspnet/AspNetCore/blob/3fd8a97af2e8cf2885a8d5836cb5d4f199458d22/src/Hosting/Hosting/src/Internal/ApplicationLifetime.cs#L21

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private static bool IsWindowsService()
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return false;
+            }
+
+            var parrent = WindowsServices.Internal.Win32.GetParrentProcess();
```
JunTaoLuo: nit: parent

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+                return hostBuilder.ConfigureServices((hostContext, services) =>
+                {
+                    services.AddSingleton<IHostLifetime, ServiceBaseLifetime>();
+                });
+            }
+
+            return hostBuilder;
+        }
+
+        private static bool IsWindowsService()
```
pakrym: Make sure this works on Win7/Win2012

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
pakrym: Shouldn't this also check for 'svchost'?

src/Hosting/WindowsServices/ref/Microsoft.Extensions.Hosting.WindowsServices.csproj
```diff
@@ -0,0 +1,12 @@
+<!-- This file is automatically generated. -->
```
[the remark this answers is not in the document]
Tratcher: eng/scripts/GenerateReferenceAssemblies.ps1

src/Hosting/WindowsServices/src/ServiceBaseLifetime.cs
Tratcher: In this case we wanted to pick a specific provider name that's common across several components.

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
Tratcher: Yes, it works fine on win7

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
Tratcher: The parent of svchost appears to be services.exe

Tratcher:
Ready for final review

src/Hosting/WindowsServices/src/ServiceBaseLifetimeHostBuilderExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                return false;
+            }
+
+            var parent = WindowsServices.Internal.Win32.GetParentProcess();
+            if (parent == null)
+            {
+                return false;
+            }
+            return parent.SessionId == 0 && string.Equals("services", parent.ProcessName, StringComparison.OrdinalIgnoreCase);
```
JunTaoLuo: nit: new line between brace and return

PR #1084, opened 2019-02-07
Use the DOTNET_ env var prefix

Tratcher:
#4150
 @glennc

### DIFF
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Extensions.Hosting
             builder.UseContentRoot(Directory.GetCurrentDirectory());
             builder.ConfigureHostConfiguration(config =>
             {
-                config.AddEnvironmentVariables(); // TODO: Prefix?
+                config.AddEnvironmentVariables("DOTNET_");
                 if (args != null)
                 {
                     config.AddCommandLine(args);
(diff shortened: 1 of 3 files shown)


davidfowl:
Needs a test

src/Hosting/Hosting/src/Host.cs
```diff
@@ -20,7 +20,7 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
             builder.UseContentRoot(Directory.GetCurrentDirectory());
             builder.ConfigureHostConfiguration(config =>
             {
-                config.AddEnvironmentVariables(); // TODO: Prefix?
+                config.AddEnvironmentVariables("DOTNET_");
```
JunTaoLuo: nit: use a named variable for the prefix like we do everywhere else: https://github.com/aspnet/AspNetCore/blob/c458fe6ebe24ab94f098a453825e71207721b9b9/src/Hosting/Hosting/src/WebHostBuilder.cs#L44

Tratcher:
@davidfowl reliably testing env vars needs out-of-proc functional testing. We don't have that infrastructure in Extensions and it isn't worth building for this.

davidfowl:
@Tratcher we need more than 0 coverage. I think it would be reasonable to write those tests in ASP.NET Core hosting since we do that the infrastructure on that side.

Tratcher:
Or part of the template I'm working on.

PR #1091, opened 2019-02-08
Fix ignored exception parameter on Degraded method

marciomyst:
- The 'exception' parameter was ignored and a hardcoded null value was used on HealthCheckResult.Degraded


rynowak:
Thanks!

0xced:
Will there be a new release of the [Microsoft.AspNetCore.Diagnostics.HealthChecks](https://www.nuget.org/packages/Microsoft.AspNetCore.Diagnostics.HealthChecks/) package or is it now part of ASP.NET Core 3?

PR #1092, opened 2019-02-08
Update package branding to 3.0.0-preview3

natemcmaster:
Per latest 3.0 release plan, we'll start including the preview number in the package branding.

cc @leecow

### DIFF
--- a/version.props
+++ b/version.props
@@ -4,7 +4,7 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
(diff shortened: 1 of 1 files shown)


version.props
```diff
@@ -4,7 +4,7 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>preview3</PreReleaseVersionLabel>
```
dougbu: Another option is to make this `preview4` immediately and back up to `preview3` in the release branch. Up to you.

natemcmaster: We don't have a release branch yet for preview3, and I think we're going to hold off on making it for a couple weeks.

natemcmaster:
Before merging, I'm going to wait until we've figured out which channel these packages should be assigned to.

cc @mmitche

Issue #526, opened 2018-10-12
Add a header propagation middleware and message handler


rynowak:
We should consider providing an in-the-box solution for propagating headers from the current request to an outgoing http request.

This would consist of a middleware that can be configured to *stamp* certain headers on a feature of the current request. Then a message handler uses `IHttpContextAccessor` to grab the feature and apply the headers to outgoing requests. 

Maybe similar to:
https://github.com/stevejgordon/CorrelationId/tree/dev/src/CorrelationId from @stevejgordon or 
(sample @davidfowl made)

Also consider using `Activity`

alefranz:
I'd like to help moving this forward. Would be creating a PR the best approach?
I plan to use @davidfowl gist with minor tweaks and a small addition.

Thanks

Symbianx:
For people who step in this issue looking for a way to easily propagate the ISTIO tracing headers like I did, we created a library just for that, you can find it here: https://github.com/nosinovacao/istio-tracing-aspnetcore .

@alefranz it also may be a good starting point for your PR.

alefranz:
@Symbianx I was actually waiting for a reply to submit a PR but I had already hacked some code. I've now submitted it, IT would be great if you can have a look and let me know if this cover your needs. I will try to have a look at that library tomorrow.

acds:
I’ve been following this as a micro service developer, and really appreciate this is being looked at. 

My main use case is in terms of tracing and tracking but do also see value in propagating other headsets including and authentication tokens for end-to-end authentication between micros services.

It would also be useful to hook any header propagation middleware such that one can leverage the hook to push headers into properties in ones logging context in a central place (like a SeriLog enricher)

antoinebj:
Is #1099 really viable with multithreading?

I'm currently using v2.2 with the need to propagate a header. However, I'm stuck with the old ASP.NET Web API 2, so instead of IHttpContextAccessor, I have to make do with HttpContext.Current. Obviously I ran into trouble the moment the HttpClient was being used from another thread (therefore with no associated HttpContext), because the delegating handler's SendAsync() was also running on that thread.
So I was curious to look at the implementation done here to see if I could get a solution that I could backport.

I read the IHttpContextAccessor is not thread safe, I'm assuming it has the same limitations whereby you cannot get the HttpContext if the code is not running on the request's thread.
If that is the case, I don't see where you avoid the same problem that I ran into.

[...]

PR #1099, opened 2019-02-11
HttpClientFactory Header Propagation MessageHandler

alefranz:
This PR introduces a `MessageHandler` for the `HttpClientFactory` to forward headers received in the current MVC request.
Addresses #526 

I believe it is a common use case which deserves to be included in the `Extensions` namespace, so I'd like to help to land this feature. Its main use case is probably to track distributed transaction which requires the ability to pass through a transaction identifier as well as generating a new one when not present. 
The code is loosely based on this [gist](https://gist.github.com/davidfowl/c34633f1ddc519f030a1c0c5abe8e867) from @davidfowl with some extra features and a different behaviour as, by default, this PR doesn't add the header if it's already present in the `HttpRequestMessage`.

Features:
- Forward header from current request MVC context to outgoing `HttpClient` requests, allowing to use a different header name. The header is forwarded only if present and all values are used.
- Always forward the header to the `HttpRequestMessage` (`AlwaysAdd=true`) or only when the header is not already present (default).
- Define a default value(s) to be added to the `HttpRequestMessage` if the incoming request does not contain the header
- Define a default value(s) generator to be used to add the header value to the `HttpRequestMessage` if the incoming request does not contain the header. The generator can rely on the `HttpRequestMessage` or `HttpContext` to decide the value. _(note: in my mind the use case is to generate a new GUID/identifier for a transaction chain if it's not in the incoming request, I'm not sure if having access to the request and context add any value)_

_The [tests](https://github.com/alefranz/Extensions/blob/8d75c26b5167adc6730152836d631ea5202689ba/src/HttpClientFactory/HeaderPropagation/test/HeaderPropagation.Tests/HeaderPropagationTest.cs) should clarify the behaviour if needed._

It definitely needs some polishing on the definition of the configuration as well as probably removing some of the different behaviours, however I believe opening a PR is the best way to move forward the discussion and shape the behaviour.
Probably the discussion on the requirements should stay on the [issue](https://github.com/aspnet/Extensions/issues/526) to keep the focus of the discussion here on the implementation.

Please Note:
- I didn't have explicit feedback to open the PR following my [comment](https://github.com/aspnet/Extensions/issues/526#issuecomment-459548541) on the issue, however the issue if marked as `up-for-grabs`
- Part of this PR is based on this [gist](https://gist.github.com/davidfowl/c34633f1ddc519f030a1c0c5abe8e867) from @davidfowl , I'm not sure what is the correct way to give attribution.

I really appreciate you taking the time to review this PR. 

Thanks,
 Alessio


mkArtakMSFT:
@rynowak, should this change go into the `AspNetCore` repo insetad?

rynowak:
Thanks @alefranz - I think that this is a good start and is on the right track. Thanks for taking this on! It looks valuable.

@natemcmaster is right though - we can't put this in this repo in its current form because we can't add a `Microsoft.AspNetCore.*` reference. Could I have you close this and create a new PR on the aspnet/AspNetCore repo? 

@Tratcher - thoughts on where this should live in that repo? It's not really middleware, but it's at that level of abstraction.

Tratcher:
@rynowak It's middleware all right, just middleware in a different pipeline. Are you prepared to ship this as a standalone package? It doesn't belong as part of any of our existing packages.

alefranz:
Thank you all for taking the time to look into this.

@natemcmaster apologies, I still get a bit confused on the scope of the different repositories

I thought a bit about this and, if there is no suitable home for this package, a different approach could be to remove the AspNetCore dependency introducing a level of abstraction, e.g.:
```csharp
public interface IContextValuesAccessor
{
    IDictionary<string, StringValues> ContextValues { get; }
}
```
Which will make this no longer explicitly about forwarding headers, but about including headers based  of values in a generic context, so not necessary AspNetCore nor in Http request.

Of course, there would still be a need of a package on the AspNetCore ecosystem to make the consumption straightforward, otherwise you would have to register the `HttpContextAccessor` as well as implementing this interface (and registering it) to forward the values from the context, like
```csharp 
private class AspNetCoreContextAccessor : IContextValuesAccessor
{
    private readonly IHttpContextAccessor _accessor;

    public AspNetCoreContextAccessor(IHttpContextAccessor accessor)
    {
        _accessor = accessor;
    }

    public IDictionary<string, StringValues> ContextValues => _accessor.HttpContext.Request.Headers;
}
```

rynowak:
@alefranz - thanks for your patience on this - I'm hopefully going to be in the office tomorrow and I'll talk to some folks to get an agreement on where this code needs to live.

I also considered the idea of splitting this feature into two parts, but I think ultimately we will need something that ties `HttpContext` and a message handler together. I think it's simpler now to think of the header-propagation as one component until we find a need to break it up.

src/HttpClientFactory/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+            _contextAccessor = contextAccessor;
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (_contextAccessor.HttpContext != null)
+            {
+                foreach (var header in _options.Headers)
+                {
+                    if (!_contextAccessor.HttpContext.Request.Headers.TryGetValue(header.InputName, out var values)
```
Tratcher: @davidfowl we've run into trouble trying to use IHttpContextAccessor in DelegatingHandlers before haven't we? People would run their HttpClient requests in parallel and cause unsafe mulithreaded access to HttpContext.  When ApplicationInsights dealt with a similar problem they concluded the best way to handle it was to collect fields they needed up front and pass them through in a dedicated data structure.

What if instead of a DelegatingHandler we provided a set of request composition helpers on top of HttpClient and HttpRequestMessage?

The biggest downside I can see would be the caller needing to be HttpContext aware, which may or may not work for various libraries.

alefranz: Hi Tratcher,
yes the HttpContext is not thread-safe and it's a really valid point, not wure why I didn't considered sending requests in parallel :(
I should add a `HeaderPropagationInitializer` similar to the `TelemetryInitializer` that get the required values from the context and store it in it's own data structure. However this will mean that it needs to be aware of all the headers forwarded by all the `HttpClient`s as we will want to lock on the context only once per MVC request. I'll have to think about it.

About the composition helpers, I'm not sure how that will fit with the `HttpClientFactory`. Can you elaborate a bit more on what you have in mind?

Btw here is the issue you are referring to:
https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/373
I remembered he mentioned it on last year talk at NDC :)

Tratcher: The composition helpers wouldn't be affiliated with HttpClientFactory, they'd be APIs for directly creating or operating on HttpRequestMessages before calling into HttpClient. They may have extensions on HttpClient to facilitate.

rynowak: Yeah, I think the right thing to do here is to make a middleware that initializes state, and a standalone `DelegatingHandler` that reads that state to modify the outgoing request.

I don't think we need to tie this to the http client factory. In general library code doesn't need to integrate with it directly because wiring up a `DelegatingHandler` is easy.

rynowak:
@alefranz - can you please resend this to `aspnet/AspNetCore` and put it in here: https://github.com/aspnet/AspNetCore/tree/master/src/Middleware

We can discuss this and review in more detail there.

alefranz:
Moved this to https://github.com/aspnet/AspNetCore/pull/7921 as suggested by @rynowak 
Thanks everyone who for the help!

Issue #966, opened 2019-01-17
Deprecate and replace IHostingEnvronment & IApplicationLifetime


Tratcher:
The name conflicts with AspNetCore are painful, provide new ones with different names.

davidfowl:
Here's the new proposal:

We're going to deprecate a couple of interfaces:
- Microsoft.AspNetCore.Hosting.IHostingEnvironment
- Microsoft.AspNetCore.Hosting.IApplicationLifetime
- Microsoft.Extensions.Hosting.IHostingEnvironment
- Microsoft.Extensions.Hosting.IApplicationLifetime

We'll replace it with the following:

```C#
namespace Microsoft.Extensions.Hosting
{
    public interface IHostEnvironment
    {
        string EnvironmentName { get; set; }
        string ApplicationName { get; set; }
        string ContentRootPath { get; set; }
        IFileProvider ContentRootFileProvider { get; set; }
    }
    public interface IHostApplicationLifetime
    {
        CancellationToken ApplicationStarted { get; }
        CancellationToken ApplicationStopping { get; }
        CancellationToken ApplicationStopped { get; }

        void StopApplication();
    }

    public static class HostEnvironmentExtensions
    {
        public static bool IsDevelopment(this IHostEnvironment environment);
        public static bool IsStaging(this IHostEnvironment environment);
        public static bool IsProduction(this IHostEnvironment environment);
    }
}
```

```C#
namespace Microsoft.AspNetCore.Hosting
{
    public interface IWebHostEnvironment : IHostEnvironment
    {
        string WebRootPath { get; set; }
        IFileProvider WebRootFileProvider { get; set; }
    }
}
```

Tratcher:
Note these warnings were temporarily suppressed in AspNetCore while updating dependencies.
https://github.com/aspnet/AspNetCore/pull/7640/commits/694f6c92e1d46994d300e5b863d7e331f7732ad6

Tratcher:
https://github.com/aspnet/AspNetCore/issues/7749

PR #1100, opened 2019-02-12
Deprecate and replace IHostingEnvronment & IApplicationLifetime

Tratcher:
#966
- IHostingEnvronment -> IHostEnvronment 
- IApplicationLifetime -> IAppLifetime

@davidfowl since the old types still exist the namespace conflict remains. It might be better to remove the old types.

@glennc

### DIFF
--- a/src/Hosting/Abstractions/ref/Directory.Build.props
+++ b/src/Hosting/Abstractions/ref/Directory.Build.props
@@ -1,0 +1,7 @@
+<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\..\..\..\, Directory.Build.props))\Directory.Build.props" />
+
+  <PropertyGroup>
+    <NoWarn>CS0618</NoWarn>
+  </PropertyGroup>
+</Project>
--- a/src/Hosting/Abstractions/ref/Directory.Build.props
+++ b/src/Hosting/Abstractions/ref/Directory.Build.props
@@ -1,7 +1,0 @@
-<Project>
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\..\..\..\, Directory.Build.props))\Directory.Build.props" />
-
-  <PropertyGroup>
-    <NoWarn>CS0618</NoWarn>
-  </PropertyGroup>
-</Project>
--- a/src/Hosting/Abstractions/src/IAppLifetime.cs
+++ b/src/Hosting/Abstractions/src/IAppLifetime.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Threading;
 
 namespace Microsoft.Extensions.Hosting
@@ -9,8 +8,7 @@ namespace Microsoft.Extensions.Hosting
     /// <summary>
     /// Allows consumers to perform cleanup during a graceful shutdown.
     /// </summary>
-    [Obsolete("Use IAppLifetime instead.", error: false)]
-    public interface IApplicationLifetime
+    public interface IAppLifetime
     {
         /// <summary>
         /// Triggered when the application host has fully started and is about to wait
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -56,6 +56,7 @@
   <Import Project="eng\Dependencies.props" />
   <Import Project="eng\PatchConfig.props" />
   <Import Project="eng\ProjectReferences.props" />
+  <Import Project="eng\targets\Csharp.Refs.props" Condition="'$(IsReferenceAssemblyProject)' == 'true'" />
 
-  <Import Project="eng\targets\Npm.Common.props"  Condition="'$(MSBuildProjectExtension)' == '.npmproj'" />
+  <Import Project="eng\targets\Npm.Common.props" Condition="'$(MSBuildProjectExtension)' == '.npmproj'" />
 </Project>
--- a/eng/targets/Csharp.Refs.props
+++ b/eng/targets/Csharp.Refs.props
@@ -1,0 +1,6 @@
+<Project>
+  <PropertyGroup>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <NoWarn>$(NoWarn);CS0618</NoWarn>
+  </PropertyGroup>
+</Project>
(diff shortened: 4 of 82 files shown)


src/Hosting/Abstractions/src/IAppLifetime.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Allows consumers to perform cleanup during a graceful shutdown.
+    /// </summary>
+    public interface IAppLifetime
```
davidfowl: Call it `IHostingApplicationLifetime` I hate the abbreviated App and it's anti the .NET naming guidelines

JunTaoLuo:
😕 I would also be confused about why there are two types that look similar. Is this a case where we need to share code between Extensions and AspNetCore better?

src/Hosting/Abstractions/src/IAppLifetime.cs
Tratcher: IHostingApplicationLifetime's not great either. @glennc any ideas?

Tratcher:
@JunTaoLuo We we'd started with generic host then web host could have shared more types. As is we're breaking the thing most people haven't used yet rather than the one everyone is already using.

src/Hosting/Abstractions/src/IAppLifetime.cs
glennc: Why is this interface duplicated again? It's identical right?

Tratcher: So the name no longer conflicts with the one in Microsoft.AspNetCore.Hosting.

davidfowl:
This really is a soft breaking change as the old types still exist and work.

src/Hosting/Abstractions/src/IAppLifetime.cs
Tratcher: How about ``IHostApplicationLifetime` so it at least mirrors the other new one and isn't quite as long?

davidfowl: YUP.

Tratcher:
@davidfowl @glennc how about replacing IHostingEnvironment in the templates as well?
https://github.com/aspnet/AspNetCore/blob/1c6651a1ae54371de34358df1622a1fa645ef0cf/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/Startup.cs#L21

davidfowl:
@Tratcher I thought about that but it might be wrong because it doesn't have the WebRoot or WebRootFileProvider.

Tratcher:
Fair point, I'll leave that one.

src/Hosting/Abstractions/src/IAppLifetime.cs
glennc: HostApplicationLifetime is what I expected but I don't remember why this has to be renamed and not moved. Do we not now have two identical things in two namespaces?

davidfowl:
cc @jdom @ReubenBond

ReubenBond:
Thanks for the heads up, @davidfowl. Orleans is not using `IHostingEnvironment` or `IApplicationLifetime` so this should not affect us.

PinpointTownes:
I'm confused by this change too. So `IHostEnvironment` will replace `IHostingEnvironment` (and will be the new generic host interface) whereas `IHostingEnvironment` will be ASP.NET Core-specific and expose things like the web root file provider?

If so, wouldn't renaming ASP's `IHostingEnvironment` to `IWebHostingEnvironment` be way less confusing?

davidfowl:
Yes but too breaking, that’s why we chose this approach. Nobody is using the generic host APIs as yet but everyone that has used ASP.NET Core is using IHostingEnvironment. It’s in the default template.

Tratcher:
@PinpointTownes we talked it over today and updated the proposal: https://github.com/aspnet/Extensions/issues/966#issuecomment-463752915

PinpointTownes:
> Nobody is using the generic host APIs as yet but everyone that has used ASP.NET Core is using IHostingEnvironment.

It likely doesn't count, but I've used it a lot in `IHostedService`/`BackgroundService` implementations 😅 

> @PinpointTownes we talked it over today and updated the proposal: #966 (comment)

Nice, it's less confusing this way. That said, wouldn't marking ASP.NET Core's `IHostingEnvironment`/`IHostApplicationLifetime` as obsolete and removing them in 4.0 be simpler than introducing a third type?

The result will be roughly the same in both cases, anyway, since `Microsoft.AspNetCore.Hosting.IHostingEnvironment` will be marked as obsolete. Migrating to `Microsoft.Extensions.Hosting.IHostingEnvironment` or `Microsoft.Extensions.Hosting.IHostingEnvironment` doesn't seem that different 😄

Tratcher:
@PinpointTownes marking one obsolete isn't sufficient, you can't use either type. Having AspNetCore.Hosting.IHostingEnvironment and Extensions.Hosting.IHostingEnvironment causes compiler errors in your app if you include both namespaces (which is becoming more common).

PinpointTownes:
Makes sense. It's just sad you have to introduce a third type to mitigate this situation, that exists since 2.2 (or 2.1?) 😄

Tratcher:
Ready for review.

I also fixed startvs.cmd so it's not hardcoded to to the root sln.

davidfowl:
What is codecheck?

Tratcher:
Codecheck verifies things like the reference assemblies are up to date. I need to regen them with these changes.

src/Hosting/Abstractions/ref/Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\..\..\..\, Directory.Build.props))\Directory.Build.props" />
+
+  <PropertyGroup>
+    <NoWarn>CS0618</NoWarn>
```
pakrym: You don't need this, right?

hoborm:
> 
> 
> #966
> 
>     * IHostingEnvronment -> IHostEnvronment
> 
>     * IApplicationLifetime -> IAppLifetime
> 
> 
> @davidfowl since the old types still exist the namespace conflict remains. It might be better to remove the old types.
> 
> @glennc

There are some typos in this thread missing the "I" in IHostEnv**I**ronment. Might be important when people want to search for the new interfaces in the repo and not seeing this issue.

Tratcher:
Posted https://github.com/aspnet/AspNetCore/issues/7749 for discoverability.

Typos corrected.

PR #1101, opened 2019-02-12
Produce a transport package with symbols

natemcmaster:
Many of the assemblies from this repo are used in the AspNetCore shared framework. These symbols are needed so the crossgen tool can re-generate symbols.

We can't use regular `.symbols.nupkg` files because these are not actually available on a restore feed as a package.

cref https://github.com/aspnet/AspNetCore/issues/7073

### DIFF
--- a/src/Internal/Symbols/pkg/Microsoft.Internal.Extensions.Symbols.csproj
+++ b/src/Internal/Symbols/pkg/Microsoft.Internal.Extensions.Symbols.csproj
@@ -5,13 +5,15 @@
     <IsPackable>true</IsPackable>
     <IsShipping>false</IsShipping>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <PackageDescription>For internal-use only. This package contains symbols for Microsoft.Extensions projects.</PackageDescription>
 
-    <!-- Subject to change: see https://github.com/dotnet/designs/pull/50 -->
-    <PackageType>TargetingPack</PackageType>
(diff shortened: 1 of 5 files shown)


src/Internal/Symbols/pkg/Microsoft.Internal.Extensions.Symbols.csproj
```diff
@@ -0,0 +1,64 @@
...
+    <!--
+      This package is not meant to be used as a regular PackageReference. This is mean to transport
+      bits to the AspNetCore repo which builds a targeting pack with these assembies.
+    -->
+    <PackageType>Transport</PackageType>
+    <SymbolsPackagePath>data/symbols/$(TargetFramework)/</SymbolsPackagePath>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+
+    <!-- The project representing the shared framework doesn't produce a .NET assembly or symbols. -->
+    <DebugType>none</DebugType>
```
pakrym: Amount of properties duplicated between project files make me sad. Move to refs.props?

pakrym: Targets too.

natemcmaster:
Getting further clarification from the people who understand crossgen better. This may not be necessary...

natemcmaster:
Ok, got more information from the symbols devs. This is not needed. Assuming symbols are uploaded correctly from this repo, we don't need to transport them to the AspNetCore build for further use.

PR #1102, opened 2019-02-13
Make activate.sh executable

JunTaoLuo:
Chmod +x so this script is executable


dougbu:
Actually, no. This change should **not** be merged. The file must be sourced or it won't do anything useful.

natemcmaster:
What Doug said...
https://github.com/aspnet/Extensions/blob/8856183718db0aa615acee2c03b439a65ef63a2f/activate.sh#L2-L3

JunTaoLuo:
Ah I see

PR #1114, opened 2019-02-14
Removed unnecessary space.

jzeferino:
- Removed unnecessary space.


pakrym:
It isn't unnecessary. We put spaces after commas.

PR #1117, opened 2019-02-14
Make the generic IStringLocalizer interface covariant

PinpointTownes:
Fixes https://github.com/aspnet/Extensions/issues/975.

@Eilon I haven't been able to execute the `restore.cmd` script (most likely the same space/special chars bug that has plagued the ASP.NET tooling for years 😅) and it looks like there's no test project for `Microsoft.Extensions.Localization.Abstractions` so it will be one of those #yolo PRs:

image

### DIFF
--- a/src/Localization/Localization/test/StringLocalizerOfTTest.cs
+++ b/src/Localization/Localization/test/StringLocalizerOfTTest.cs
@@ -1,0 +1,159 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Extensions.Localization
+{
+    public class StringLocalizerOfTTest
+    {
+        [Fact]
+        public void Constructor_ThrowsAnExceptionForNullFactory()
+        {
+            // Arrange, act and assert
+            var exception = Assert.Throws<ArgumentNullException>(
+                () => new StringLocalizer<object>(factory: null));
+
+            Assert.Equal("factory", exception.ParamName);
+        }
+
+        [Fact]
+        public void Constructor_ResolvesLocalizerFromFactory()
+        {
+            // Arrange
+            var factory = new Mock<IStringLocalizerFactory>();
+
+            // Act
+            _ = new StringLocalizer<object>(factory.Object);
+
+            // Assert
+            factory.Verify(mock => mock.Create(typeof(object)), Times.Once());
+        }
+
+        [Fact]
+        public void WithCulture_InvokesWithCultureFromInnerLocalizer()
+        {
+            // Arrange
+            var factory = new Mock<IStringLocalizerFactory>();
+            var innerLocalizer = new Mock<IStringLocalizer>();
+            factory.Setup(mock => mock.Create(typeof(object)))
+                .Returns(innerLocalizer.Object);
+
+            var localizer = new StringLocalizer<object>(factory.Object);
+
+            // Act
+#pragma warning disable CS0618
+            localizer.WithCulture(CultureInfo.GetCultureInfo("fr-FR"));
+#pragma warning restore CS0618
+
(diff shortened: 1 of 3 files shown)


Eilon:
@mkArtakMSFT - can you assign someone to review?

@PinpointTownes thanks for the PR!

PinpointTownes:
Nice, I had not realized you had started using ref files in the ASP.NET repos 😅

Eilon:
Ah yes it's brand new! 😄

mkArtakMSFT:
@ryanbrandenburg, can you please review this? Thanks!

mkArtakMSFT:
Thanks for your effort, @PinpointTownes.
Can you please also add a test, which would validate the scenario you're making a change for. Otherwise if someone goes and removes the `out` keyword from the interface now, we may easily introduce a breaking change without noticing it in the future.

PinpointTownes:
I'd love to, @mkArtakMSFT, but as I said:
  - There's currently no test project for `Microsoft.Extensions.Localization.Abstractions` (which means things like the generic `StringLocalizer<T>` are not unit tested ATM).
  - The tooling doesn't work properly on my machine.

I'll try to did deeper to make the tooling work, but without it, it's not really practical for me to add new test projects 😅

mkArtakMSFT:
@PinpointTownes that makes sense. @ryanbrandenburg should we create a new test project for `Microsoft.Extensions.Localization.Abstractions` or should tests go to some other project for some concrete classes and implicitly test those type of changes?

Eilon:
If we want to avoid a billion projects, adding it to an existing test project might be best (or renaming an existing specific unit test project to be more generic sounding, and then add this new test to it).

hishamco:
Or we can add a new sample that using logging `IStringLocalizer<ILogger>` and add a functional test will be more valuable

PinpointTownes:
I added a bunch of unit tests for the generic `StringLocalizer` in the existing `Microsoft.Extensions.Localization.Tests` project.

src/Localization/Localization/test/StringLocalizerOfTTest.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            // Assert
+            innerLocalizer.Verify(mock => mock.GetAllStrings(true), Times.Once());
+        }
+
+        [Fact]
+        public void StringLocalizer_CanBeCastToBaseType()
+        {
+            // Arrange and act
+            IStringLocalizer<BaseType> localizer = new StringLocalizer<DerivedType>(Mock.Of<IStringLocalizerFactory>());
```
PinpointTownes: Definitely not the most exciting test I've ever written 🤣

src/Localization/Localization/test/StringLocalizerOfTTest.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            // Act
+            _ = new StringLocalizer<object>(factory.Object);
+
+            // Assert
+            factory.Verify(mock => mock.Create(typeof(object)), Times.Once());
+        }
+
+        [Fact]
+        public void WithCulture_InvokesWithCultureFromInnerLocalizer()
```
hishamco: I'm not sure if we need such test while the `WithCulture` is obsolete

src/Localization/Localization/test/StringLocalizerOfTTest.cs
```diff
@@ -0,0 +1,159 @@
...
+            // Arrange
+            var factory = new Mock<IStringLocalizerFactory>();
+            var innerLocalizer = new Mock<IStringLocalizer>();
+            factory.Setup(mock => mock.Create(typeof(object)))
+                .Returns(innerLocalizer.Object);
+
+            var localizer = new StringLocalizer<object>(factory.Object);
+
+            // Act
+#pragma warning disable CS0618
```
hishamco: I hate such warning :)

hishamco: I hate such warning :)

PR #1118, opened 2019-02-15
Make JSRuntime.Current non-public

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/6828

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
@@ -11,13 +11,9 @@ namespace Microsoft.JSInterop
     /// </summary>
     public static class JSRuntime
     {
-        private static AsyncLocal<IJSRuntime> _currentJSRuntime
-            = new AsyncLocal<IJSRuntime>();
+        private static readonly AsyncLocal<IJSRuntime> _currentJSRuntime = new AsyncLocal<IJSRuntime>();
 
-        /// <summary>
-        /// Gets the current <see cref="IJSRuntime"/>, if any.
-        /// </summary>
-        public static IJSRuntime Current => _currentJSRuntime.Value;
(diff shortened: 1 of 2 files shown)


src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
```diff
@@ -11,13 +11,9 @@ namespace Microsoft.JSInterop
...
     {
-        private static AsyncLocal<IJSRuntime> _currentJSRuntime
-            = new AsyncLocal<IJSRuntime>();
+        private static readonly AsyncLocal<IJSRuntime> _currentJSRuntime = new AsyncLocal<IJSRuntime>();
 
-        /// <summary>
-        /// Gets the current <see cref="IJSRuntime"/>, if any.
-        /// </summary>
-        public static IJSRuntime Current => _currentJSRuntime.Value;
+        internal static IJSRuntime Current => _currentJSRuntime.Value;
```
rynowak: It says remove. But from looking at it, I can see that it is not removed.

SteveSandersonMS: Indeed. Could the entire `JSRuntime.cs` file be removed?

pranavkm: Not in any trivial way. It's used from a static context by `DotnetNetDispatcher` to notify of completion:

* https://github.com/aspnet/Extensions/blob/master/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs#L99
* https://github.com/aspnet/Extensions/blob/master/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs#L192-L193

In addition, there's also some pretty gnarly coupling of the runtime \ SimpleJSON's serialization and object tracking. I was hoping not to mess around with the code here if it's meant to be cleaned up soon.

pranavkm:
Bump

PR #1128, opened 2019-02-17
Update ObjectPool reference assemblies

rynowak:


rynowak:
This change resolves a build break. I merged a PR that predated the reference assembly change, and thus didn't include that in its PR validation.

Issue #1009, opened 2019-01-29
Run healthchecks in parallel


NatMarchand:
Hi there,
I would like to suggest running healthchecks in parallel.
We currently have multiple checks on dependencies (4-5) which can take around 100-200ms and in the end the whole healtcheck request can take up to a second or more which can cause timeouts upstream.
If it's just parallelizing the checks I'll be glad to make the PR. If there's a gotcha that I missed, please tell me in order to take it in account.
Thanks a lot !

rynowak:
Hi @NatMarchand we would gladly accept a PR for this. 

You can find some similar logic inside [here](https://github.com/aspnet/Extensions/blob/master/src/HealthChecks/HealthChecks/src/HealthCheckPublisherHostedService.cs#L150) to what you will need. 

The one obvious gotcha here is that you will need to wait for all of the tasks to complete to create the results dictionary since dictionary isn't thread-safe.

@glennc - do you think it's worth adding an option or should we make this the default? This is something we considered from the beginning but I don't want to break people even if they are doing something we don't recommend.

rynowak:
Done! Thanks @NatMarchand

PR #1129, opened 2019-02-18
Feature/parallel healthchecks

NatMarchand:
I've updated DefaultHealthCheckService to run tests in parallel.

Solves:
#1009


rynowak:
Thanks! I will take a look at this in detail in a bit.

NatMarchand:
I've also another fork where I've implemented a timeout per check. Would you be also interested in a second PR ?
It would solve the issues raised there : https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/73

rynowak:
> I've also another fork where I've implemented a timeout per check. Would you be also interested in a second PR ?

Can you link to it? I want to look at what you did first

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -39,76 +39,81 @@ internal class DefaultHealthCheckService : HealthCheckService
             Func<HealthCheckRegistration, bool> predicate,
             CancellationToken cancellationToken = default)
         {
-            var registrations = _options.Value.Registrations;
-
-            using (var scope = _scopeFactory.CreateScope())
+            async Task<(string registrationName, HealthReportEntry result)> RunCheckAsync(IServiceScope scope, HealthCheckRegistration registration)
```
rynowak: Could this be a regular private method instead? I want to prevent this from capturing things.

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -39,76 +39,81 @@ internal class DefaultHealthCheckService : HealthCheckService
...
+            }
+
+            var totalTime = ValueStopwatch.StartNew();
+            Log.HealthCheckProcessingBegin(_logger);
 
-                var totalElapsedTime = totalTime.GetElapsedTime();
-                var report = new HealthReport(entries, totalElapsedTime);
-                Log.HealthCheckProcessingEnd(_logger, report.Status, totalElapsedTime);
-                return report;
+            (string registrationName, HealthReportEntry result)[] results;
```
rynowak: Now that this runs heath checks in parallel, we need to deal with exceptions that are thrown synchronously. An exception thrown synchronously from L60 could fail the whole batch.

One technique that I've used to solve this before is to put a `Task.Yield()` at the beginning of `RunCheckAsync` to force actual async. 

I'd suggest

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -39,76 +39,81 @@ internal class DefaultHealthCheckService : HealthCheckService
...
+            using (var scope = _scopeFactory.CreateScope())
+            {
+                results = await Task.WhenAll(registrations.Select(r => RunCheckAsync(scope, r)));
```
rynowak: I'd like to avoid the capturing lambda and `.Select` allocation if possible - just use a `foreach`?

This also should have a `.ConfigureAwait(false)`.

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -39,76 +39,81 @@ internal class DefaultHealthCheckService : HealthCheckService
...
+                            data: null);
+
+                        Log.HealthCheckError(_logger, registration, ex, duration);
                     }
+
+                    return (registration.Name, entry);
                 }
+            }
+
+            IEnumerable<HealthCheckRegistration> registrations = _options.Value.Registrations;
```
rynowak: `var` everywhere 😆

NatMarchand:
Thanks for your feedbacks ! I've taken in account most of the points. However, I have few things :
- `IEnumerable<HealthCheckRegistration>` was on purpose to be able to chain the where without recreating a collection with ToList or ToArray.
- I'm unable to make a check that throws synchronously. I've tried this but don't see any difference in behavior (with yield and no yield) :
```csharp
            // Arrange
            var service = CreateHealthChecksService(b =>
            {
                b.AddCheck("failing1", () => throw new Exception("failing1"));
                b.AddAsyncCheck("failing2", () => throw new Exception("failing2"));
#pragma warning disable
                b.AddAsyncCheck("failing3", async () =>
                {
                    throw new Exception("failing3");
                });
#pragma warning restore
            });
```
- I'm not totally sure on how to test parallelism, I've got a proposal but feel free to tell me if it's not good.
- The link to the other branch for the timeout :  https://github.com/NatMarchand/Extensions/commit/f64bac9ddfa99141477346f27d6cb82ba60b1e9c
Please note that this timeout is based on a CancellationToken and checks are free to opt in and use this token. Not all the checks will be able to use it (especially those using only sync APIs).

rynowak:
> * I'm unable to make a check that throws synchronously. I've tried this but don't see any difference in behavior (with yield and no yield) :

Yeah, you're right about this. The reason it's not an issue is because of the `await` at https://github.com/aspnet/Extensions/pull/1129/files#diff-d3c1e8bb845ae264c0de922c803fd8baR96 - we can ignore this case 👍 

-------------

For the async test, I'd suggest using task completion sources - and write some health checks that await those - instead of `Task.Delay`. [example here](https://github.com/aspnet/Extensions/blob/master/src/HealthChecks/HealthChecks/test/HealthCheckPublisherHostedServiceTest.cs#L225)

The overall flow I would expect is like this:

1. Create 2 `TaskCompletionSource<>` instances - these are the *input* tasks that health checks will await.
2. Create 2 more `TaskCompletionSource<>` instances - these are the *output* tasks that health checks will use to notify the test that they are running.
3. Register two async checks (each one with logic like the following:
```
b.AddAsyncCheck("test1", () =>
{
    output1.SetResult(null); // notify the test that I have started
    await input1.Task; // wait for test to unblock me
});
```
4. test calls `CheckHealthAsync` but does not await the task
5. test calls `await Task.WhenAll(output1, output2).TimeoutAfter(TimeSpan.FromSeconds(10));` - this waits until both checks have started.
6. test sets a result in `input1` and `input2` to unblock the checks to complete
7. `await` the task returned by `CheckHealthAsync`.

This might seem a little complicated at first, but it's a pretty robust strategy that I've used in a few places.  It doesn't rely on timing and it doesn't have any deadlocks.

rynowak:
> * The link to the other branch for the timeout :  [NatMarchand@f64bac9](https://github.com/NatMarchand/Extensions/commit/f64bac9ddfa99141477346f27d6cb82ba60b1e9c)
  Please note that this timeout is based on a CancellationToken and checks are free to opt in and use this token. Not all the checks will be able to use it (especially those using only sync APIs).

I think your design for this is awesome! This is basically how I would do it a few minor tweaks. It would be awesome to have you send that as a PR as well (separate PR 😁)

NatMarchand:
I've updated the test.

I also have a small question. What about AsyncLocal<T> fields with parallel tasks ? My concern : aspnetcore use Activity for diagnostics. ApplicationInsights uses it for correlating dependency calls to the request by fetching the current activity (which is AsyncLocal<T>). Do you think that running in parallel might cause the correlation to fail ? Shall I start a subactivity parented by the request Activity ?

rynowak:
> I also have a small question. What about AsyncLocal fields with parallel tasks ?

AsyncLocal is fine. Each level of async makes a copy of the current state, so the current activity is preserved.

rynowak:
Thanks! This will be in preview 4 of ASP.NET Core 3.0.0 - the next release is preview 3 but we've been done accepting changes for preview 3 that aren't essential for a few days.

PR #1133, opened 2019-02-19
Obsolete ResourceManagerWithCultureStringLocalizer and WithCulture

ryanbrandenburg:
Fixes https://github.com/aspnet/AspNetCore/issues/3324.


src/Localization/Abstractions/src/IStringLocalizer.cs
```diff
@@ -40,6 +41,7 @@ public interface IStringLocalizer
         /// </summary>
         /// <param name="culture">The <see cref="CultureInfo"/> to use.</param>
         /// <returns>A culture-specific <see cref="IStringLocalizer"/>.</returns>
+        [Obsolete("Set the CurrentCulture and CurrentUICulture instead.")]
```
mkArtakMSFT: I think this is too strict. Let's go with: "This method is obsolete. Use `CurrentCulture` and `CurrentUICulture` instead". And please use this patter for the rest of the messages.

PR #1137, opened 2019-02-20
Fail when activate is executed instead of 'dot sourced'

natemcmaster:
Helps users avoid a common mistake which is otherwise not obvious.

image


JunTaoLuo:
Is there something similar for activate.sh? Also, make sure this goes back to AspNetCore as well.

natemcmaster:
Updated activate.sh to have the same behavior

PR #1144, opened 2019-02-21
Implement Timeout for healthchecks

NatMarchand:
This PR enables checks to have an individual timeout.
It's based on a CancellationTokenSource that provides a CancellationToken to notify check. Meanwhile the CancellationTokenSource is also enrolled with the CancellationToken received in parameter (for connection aborted or timeout of publisher).

As agreed here : https://github.com/aspnet/Extensions/pull/1129#issuecomment-465855605


src/HealthChecks/Abstractions/src/HealthCheckRegistration.cs
```diff
@@ -35,7 +35,8 @@ public sealed class HealthCheckRegistration
         /// is <c>null</c>, then <see cref="HealthStatus.Unhealthy"/> will be reported.
         /// </param>
         /// <param name="tags">A list of tags that can be used for filtering health checks.</param>
-        public HealthCheckRegistration(string name, IHealthCheck instance, HealthStatus? failureStatus, IEnumerable<string> tags)
+        /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
+        public HealthCheckRegistration(string name, IHealthCheck instance, HealthStatus? failureStatus, IEnumerable<string> tags, TimeSpan? timeout = null)
```
rynowak: So, this is a breaking change, we can't do this 😆 It is a binary breaking change to add a new parameter with a default value. So even though people's code would *compile* it will break existing compiled code.

rynowak: So you'll need to add another overload of each of these.

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -72,7 +72,13 @@ internal class DefaultHealthCheckService : HealthCheckService
...
                         try
                         {
-                            var result = await healthCheck.CheckHealthAsync(context, cancellationToken);
+                            HealthCheckResult result;
+                            using (var timeoutCancellationTokenSource = new CancellationTokenSource(registration.Timeout))
+                            using (cancellationToken.Register(() => timeoutCancellationTokenSource.Cancel()))
+                            {
+                                result = await healthCheck.CheckHealthAsync(context, timeoutCancellationTokenSource.Token);
+                            }
+
```
rynowak: Can we find a way to make the allocation of `timeoutCancellationTokenSource` skip when the timeout is infinite? I suspect an infinite timeout is going to be the common case.

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -72,7 +72,13 @@ internal class DefaultHealthCheckService : HealthCheckService
                         HealthReportEntry entry;
                         try
                         {
-                            var result = await healthCheck.CheckHealthAsync(context, cancellationToken);
+                            HealthCheckResult result;
+                            using (var timeoutCancellationTokenSource = new CancellationTokenSource(registration.Timeout))
+                            using (cancellationToken.Register(() => timeoutCancellationTokenSource.Cancel()))
```
rynowak: You can use https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.createlinkedtokensource?view=netframework-4.7.2 to combine two tokens.

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -86,6 +92,19 @@ internal class DefaultHealthCheckService : HealthCheckService
...
+                            entry = new HealthReportEntry(
+                                status: HealthStatus.Unhealthy,
+                                description: "A timeout occured while running check.",
+                                duration: duration,
+                                exception: ex,
+                                data: null);
+
+                            Log.HealthCheckError(_logger, registration, ex, duration);
+                        }
+
```
rynowak: nit: update the comment at L108 to say `Allow cancellation to propagate if it's not a timeout.`

src/HealthChecks/HealthChecks/src/DependencyInjection/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -23,13 +23,15 @@ public static class HealthChecksBuilderAddCheckExtensions
...
+        /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
         /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
         public static IHealthChecksBuilder AddCheck(
             this IHealthChecksBuilder builder,
             string name,
             IHealthCheck instance,
             HealthStatus? failureStatus = null,
-            IEnumerable<string> tags = null)
+            IEnumerable<string> tags = null,
+            TimeSpan? timeout = null)
```
rynowak: This is a breaking change also. Here's a short doc on adding overloads to a method with default parameters. https://github.com/dotnet/roslyn/blob/master/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md

src/HealthChecks/HealthChecks/src/DependencyInjection/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -187,5 +191,44 @@ public static class HealthChecksBuilderAddCheckExtensions
...
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            return builder.Add(new HealthCheckRegistration(name, s => ActivatorUtilities.CreateInstance<T>(s, args), failureStatus, tags, timeout));
+        }
```
rynowak: This should be fine.

src/HealthChecks/HealthChecks/src/DependencyInjection/HealthChecksBuilderAddCheckExtensions.cs
```diff
@@ -60,18 +62,20 @@ public static class HealthChecksBuilderAddCheckExtensions
...
         /// </param>
         /// <param name="tags">A list of tags that can be used to filter health checks.</param>
+        /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
         /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
         /// <remarks>
         /// This method will use <see cref="ActivatorUtilities.GetServiceOrCreateInstance{T}(IServiceProvider)"/> to create the health check
-        /// instance when needed. If a service of type <typeparamref name="T"/> is registred in the dependency injection container
-        /// with any liftime it will be used. Otherwise an instance of type <typeparamref name="T"/> will be constructed with
+        /// instance when needed. If a service of type <typeparamref name="T"/> is registered in the dependency injection container
+        /// with any lifetime it will be used. Otherwise an instance of type <typeparamref name="T"/> will be constructed with
```
rynowak: Thanks for fixing all of my embarrasing typos

src/HealthChecks/HealthChecks/test/DefaultHealthCheckServiceTest.cs
```diff
@@ -375,6 +375,32 @@ public async Task CheckHealthAsync_CheckCanDependOnSingletonService()
...
+            // Assert
+            Assert.Collection(
+                results.Entries,
+                actual =>
+                {
+                    Assert.Equal("timeout", actual.Key);
+                    Assert.Equal(HealthStatus.Unhealthy, actual.Value.Status);
+                });
+        }
+
```
rynowak: nice 👍

src/HealthChecks/Abstractions/src/HealthCheckRegistration.cs
```diff
@@ -79,10 +88,16 @@ public HealthCheckRegistration(string name, IHealthCheck instance, HealthStatus?
                 throw new ArgumentNullException(nameof(factory));
             }
 
+            if (timeout.HasValue && timeout.Value <= TimeSpan.Zero && timeout.Value != System.Threading.Timeout.InfiniteTimeSpan)
```
rynowak: Can you add a using for `System.Threading`?

src/HealthChecks/Abstractions/src/HealthCheckRegistration.cs
NatMarchand: Is running assemblies compiled against 2.x with a runtime 3.x supported ? I thought dotnetcore rollforward rules applied (cf. https://docs.microsoft.com/en-us/dotnet/core/versions/selection#framework-dependent-apps-roll-forward )

NatMarchand: Ok forget my previous message, I totally missed that it can be executed in other contexts than aspnetcore 😅

src/HealthChecks/Abstractions/src/HealthCheckRegistration.cs
NatMarchand: Nope, name collision with the property :/

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
NatMarchand: Thanks for the tip! Didn't know this method, very useful !

src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
NatMarchand: Done

NatMarchand:
Rebased to take in account other changes and updated according to the review comments !

src/HealthChecks/HealthChecks/src/DependencyInjection/HealthChecksBuilderAddCheckExtensions.cs
NatMarchand: Backcompat method added

src/HealthChecks/Abstractions/src/HealthCheckRegistration.cs
rynowak: Yes, and libraries are a thing too. Most libraries are pretty naive about what version of our stuff the target. So they need to be compatible with a range of things.

We really intend for an area like health checks to be super stable.

NatMarchand:
My pleasure, thanks for you feedback ! 😄

PR #1154, opened 2019-02-22
Update core-setup version and workaround dotnet/sdk#2976

natemcmaster:
More dependency flow updates due to re-spins in dotnet/core-setup

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,7 +9,6 @@ trigger:
     include:
     - master
     - release/*
-    - internal/release/*
 
 pr:
   autoCancel: false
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -9,7 +9,6 @@ trigger:
     include:
     - master
     - release/*
-    - internal/release/*
```
dougbu: Instead make these conditional on the queue e.g. something like
``` yaml
${{ if ne(variables['System.TeamProject'], 'internal') }}:
  - release/*
${{ if eq(variables['System.TeamProject'], 'internal') }}:
  - internal/release/*
```

natemcmaster: We should not build internal/release/* at all right now. Arcade publishing is not designed to handle this and we do not want to accidentally push internal-only fixes publicly. I'm sending mail about this and will CC you

natemcmaster:
I'm going to use admin rights to merge this so we can kick of upstream updates. @dougbu let's discuss the branch thing separately. It's important for now that we do not build internal/release/*, so I'm leaving that change in.

PR #1156, opened 2019-02-22
[JSInterop] Skip test until we move to the new .NET Core JSON implementation.

javiercn:


src/JSInterop/Microsoft.JSInterop/src/Json/SimpleJson/SimpleJson.cs
```diff
@@ -2184,12 +2214,18 @@ public bool Remove(KeyValuePair<TKey, TValue> item)
...
+                        return _dictionary.GetEnumerator();
+                    }
                 }
 
                 System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
                 {
-                    return _dictionary.GetEnumerator();
+                    lock (_lock)
+                    {
+                        return _dictionary.GetEnumerator();
```
pakrym: This won't help, enumerator is being used outside the lock.

javiercn: Any cheap way of fixing this? I can create an enumerator wrapper and have it release the lock on Dispose but anything else seems overkill.

I'm not concerned too much about this as the callstacks that we have show another method is the one causing issues, but if there is something trivial (15 min) I'll do it.

rynowak: You can have it copy the underlying data. 

Is it feasible to eliminate this from the code? - ie does anything call this? If this is internal to simplejson can the calling code take the lock

src/JSInterop/Microsoft.JSInterop/src/Json/SimpleJson/SimpleJson.cs
```diff
@@ -2117,12 +2117,21 @@ public void Add(TKey key, TValue value)
...
                 }
 
                 public ICollection<TKey> Keys
                 {
-                    get { return _dictionary.Keys; }
+                    get
+                    {
+                        lock(_lock)
+                        {
+                            return _dictionary.Keys;
```
rynowak: This is also broken.

src/JSInterop/Microsoft.JSInterop/src/Json/SimpleJson/SimpleJson.cs
```diff
@@ -2132,18 +2141,33 @@ public bool Remove(TKey key)
...
                 }
 
                 public ICollection<TValue> Values
                 {
-                    get { return _dictionary.Values; }
+                    get
+                    {
+                        lock(_lock)
+                        {
+                            return _dictionary.Values;
```
rynowak: broken

PR #1159, opened 2019-02-22
Change publish feed to workaround 'Waiting to obtain an exclusive lock on the feed.'

natemcmaster:
2 hours is no longer enough. We're getting consistent build timeouts. Changing to a unique feed instead. Each upstream partner will need to add this to restore sources.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,7 +35,6 @@ jobs:
     enablePublishBuildAssets: true
     enableTelemetry: true
     helixRepo: aspnet/Extensions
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -35,7 +35,6 @@ jobs:
     enablePublishBuildAssets: true
     enableTelemetry: true
     helixRepo: aspnet/Extensions
-    timeoutInMinutes: 120 # increase timeout since BAR publishing might wait a long time
```
ryanbrandenburg: This brings us back to the default of 1 hour correct?

natemcmaster: yes

Issue #1189, opened 2019-02-28
InvalidOperationException thrown when binding operation underway during configuration reload


martincostello:
### Describe the bug

If `IConfigurationRoot.Reload()` is called and the application is serving requests that require an options object to be bound from configuration on other threads, sometimes an `InvalidOperationException` is thrown causing HTTP 500 responses while the typed config is being bound. An example stack trace is below.

#### Stack trace
<details>

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.SparseArrayBuilder`1.ReserveOrAdd(IEnumerable`1 items)
   at System.Linq.Enumerable.Concat2Iterator`1.ToArray()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.GetEnumerator()+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.SparseArrayBuilder`1.ReserveOrAdd(IEnumerable`1 items)
   at System.Linq.Enumerable.Concat2Iterator`1.ToArray()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.GetEnumerator()+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.SparseArrayBuilder`1.ReserveOrAdd(IEnumerable`1 items)
   at System.Linq.Enumerable.Concat2Iterator`1.ToArray()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.GetEnumerator()+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.SparseArrayBuilder`1.ReserveOrAdd(IEnumerable`1 items)
   at System.Linq.Enumerable.Concat2Iterator`1.ToArray()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
[...]

PR #1202, opened 2019-03-02
Atomically swap config data

martincostello:
Swap the config providers' data atomically instead of changing the property.

This is so that any enumeration of the dictionary running during a (re)load operation does not throw an `InvalidOperationException` due to the collection being modified.

Addresses #1189

### DIFF
--- a/src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
+++ b/src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
@@ -944,9 +944,7 @@ IniKey1=IniValue2");
             Assert.True(token.HasChanged);
         }
 
-        [ConditionalFact]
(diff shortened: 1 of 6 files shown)


src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
```diff
@@ -944,6 +944,50 @@ public async Task TouchingFileWillReloadForUserSecrets()
             Assert.True(token.HasChanged);
         }
 
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux, SkipReason = "File watching is flaky on non windows.")]
+        [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "File watching is flaky on non windows.")]
```
Tratcher: You're not doing file watching in this test, remove these.

Tratcher:
Thanks

Issue #1203, opened 2019-03-02
Test failure: Options test fails on en-GB localised machine


martincostello:
### Describe the bug

On a computer where an en-GB locale is used the `Configure_GetsNullableOptionsFromConfiguration(...)` tests fail.

```
Microsoft.Extensions.Options.Tests.OptionsTest.Configure_GetsNullableOptionsFromConfiguration(configValues: [[MyNullableBool, false], [MyNullableInt, -1], [MyNullableDateTime, 31/12/1995]], expectedValues: [[MyNullableBool, False], [MyNullableInt, -1], [MyNullableDateTime, 31/12/1995 00:00:00]])
System.InvalidOperationException : Failed to convert '31/12/1995' to type 'System.DateTime'.\r\n---- System.FormatException : 31/12/1995 is not a valid value for DateTime.\r\n-------- System.FormatException : String '31/12/1995' was not recognized as a valid DateTime.
   at Microsoft.Extensions.Configuration.ConfigurationBinder.BindInstance(Type type, Object instance, IConfiguration config, BinderOptions options) in C:\Coding\aspnet\Extensions\src\Configuration\Config.Binder\src\ConfigurationBinder.cs:line 303
   at Microsoft.Extensions.Configuration.ConfigurationBinder.BindProperty(PropertyInfo property, Object instance, IConfiguration config, BinderOptions options) in C:\Coding\aspnet\Extensions\src\Configuration\Config.Binder\src\ConfigurationBinder.cs:line 209
   at Microsoft.Extensions.Configuration.ConfigurationBinder.BindNonScalar(IConfiguration configuration, Object instance, BinderOptions options) in C:\Coding\aspnet\Extensions\src\Configuration\Config.Binder\src\ConfigurationBinder.cs:line 184
   at Microsoft.Extensions.Configuration.ConfigurationBinder.BindInstance(Type type, Object instance, IConfiguration config, BinderOptions options) in C:\Coding\aspnet\Extensions\src\Configuration\Config.Binder\src\ConfigurationBinder.cs:line 342
   at Microsoft.Extensions.Configuration.ConfigurationBinder.Bind(IConfiguration configuration, Object instance, Action`1 configureOptions) in C:\Coding\aspnet\Extensions\src\Configuration\Config.Binder\src\ConfigurationBinder.cs:line 119
[...]

PR #1204, opened 2019-03-02
Use invariant culture

martincostello:
Use InvariantCulture with ToString() operation on DateTime in test.

Addresses #1203.


HaoK:
Thanks for fixing this @martincostello

Issue #1205, opened 2019-03-02
Test failure: HttpClientSlimTest tests fail


martincostello:
### Describe the bug

Running the `HttpClientSlimTest` tests on my laptop, 4 of them fail with:

```
Test Name:	Microsoft.AspNetCore.Testing.HttpClientSlimTest.GetStringAsyncHttp
Test FullName:	Microsoft.AspNetCore.Testing.HttpClientSlimTest.GetStringAsyncHttp
Test Source:	C:\Coding\aspnet\Extensions\src\TestingUtils\Microsoft.AspNetCore.Testing\test\HttpClientSlimTest.cs : line 18
Test Outcome:	Failed
Test Duration:	0:00:00.039

Result StackTrace:	
at Microsoft.AspNetCore.Testing.HttpClientSlimTest.StartHost(String& address, Int32 statusCode, Func`2 handler) in C:\Coding\aspnet\Extensions\src\TestingUtils\Microsoft.AspNetCore.Testing\test\HttpClientSlimTest.cs:line 94
   at Microsoft.AspNetCore.Testing.HttpClientSlimTest.GetStringAsyncHttp() in C:\Coding\aspnet\Extensions\src\TestingUtils\Microsoft.AspNetCore.Testing\test\HttpClientSlimTest.cs:line 20
--- End of stack trace from previous location where exception was thrown ---
Result Message:	
IsListening
Expected: True
Actual:   False
```

The exception being thrown is an `HttpListenerException` with the message `The network location cannot be reached. For information about network troubleshooting, see Windows Help.`. The tests try to start a listener 10 times, and then fail the assertion as there's no listener active.

### To Reproduce

Specific reason these tests fail is unknown, but appears to be related to the use of `127.0.0.1`.

### Expected behavior

Tests pass.

martincostello:
See #1206.

Eilon:
@Tratcher - any idea on this? Does PR #1206 look good?

PR #1206, opened 2019-03-02
Use localhost for HttpListener

martincostello:
Use localhost instead of 127.0.0.1 to fix tests.

Addresses #1205


Tratcher:
Thanks

PR #1209, opened 2019-03-02
Use TryGetValue() for dictionary

martincostello:
Use `TryGetValue()` when accessing the mapping dictionary, rather than `ContainsKey()` + `[]`.


Tratcher:
Thanks

PR #1210, opened 2019-03-02
Use Array.Empty<T>()

martincostello:
Use `Array.Empty<string>()` instead of allocating a new array if either of the strings for comparison are null.

### DIFF
--- a/src/Configuration/Config/src/ConfigurationKeyComparer.cs
+++ b/src/Configuration/Config/src/ConfigurationKeyComparer.cs
@@ -26,8 +26,8 @@ namespace Microsoft.Extensions.Configuration
         /// <returns></returns>
         public int Compare(string x, string y)
         {
-            var xParts = x?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
-            var yParts = y?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
+            var xParts = x?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
+            var yParts = y?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
 
             // Compare each part until we get two parts that are not equal
             for (int i = 0; i < Math.Min(xParts.Length, yParts.Length); i++)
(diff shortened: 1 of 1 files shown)


src/Configuration/Config/src/ConfigurationKeyComparer.cs
```diff
@@ -41,7 +41,7 @@ public int Compare(string x, string y)
                 var xIsInt = x != null && int.TryParse(x, out value1);
```
gfoidl:
```c#
var xIsInt = int.TryParse(x, out int value1);
var yIsInt = int.TryParse(y, out int value2);
```

The null-check is done in `TryParse`, so here it is redundant. Variable initialized by the `out` directly.

src/Configuration/Config/src/ConfigurationKeyComparer.cs
```diff
@@ -26,8 +26,8 @@ public class ConfigurationKeyComparer : IComparer<string>
         /// <returns></returns>
         public int Compare(string x, string y)
         {
-            var xParts = x?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
-            var yParts = y?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
+            var xParts = x?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
+            var yParts = y?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
 
             // Compare each part until we get two parts that are not equal
             for (int i = 0; i < Math.Min(xParts.Length, yParts.Length); i++)
```
gfoidl:
```suggestion
            var n = Math.Min(xParts.Length, yParts.Length);
            for (var i = 0; i < n; i++)
```
Otherwise `n` is evaluated at each iteration.

src/Configuration/Config/src/ConfigurationKeyComparer.cs
```diff
@@ -26,8 +26,8 @@ public class ConfigurationKeyComparer : IComparer<string>
         /// <returns></returns>
         public int Compare(string x, string y)
         {
-            var xParts = x?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
-            var yParts = y?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
+            var xParts = x?.Split(_keyDelimiterArray, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
```
gfoidl: Is it a common case that either `x` or `y` is `null`?

If so, it may be better to hold `Array.Empty<string>` in a field, as the call to `Empty` won't be inlined (last time I checked this was true).

If not, nevermind this comment.

Tratcher:
Thanks

PR #1215, opened 2019-03-04
Fix aspnet/AspNetCore-Internal#1869

rynowak:
I think the issue here is that the HttpClient goes out of scope too
early and the cleanup pass is able to dispose it before we intend for it
to be disposed.

GC.Keepalive will preserve the liveness until the point we're ready to
null it out.

### DIFF
--- a/src/HttpClientFactory/Http/test/DefaultHttpClientFactoryTest.cs
+++ b/src/HttpClientFactory/Http/test/DefaultHttpClientFactoryTest.cs
@@ -482,7 +482,11 @@ namespace Microsoft.Extensions.Http
                 kvp = default;
             }
 
-            // Act - 1
+            // Let's verify the the ActiveHandlerTrackingEntry is gone. This would be prevent
+            // the handler from being disposed if it was still rooted.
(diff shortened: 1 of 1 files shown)


src/HttpClientFactory/Http/test/DefaultHttpClientFactoryTest.cs
```diff
@@ -482,7 +482,11 @@ public async Task Factory_CleanupCycle_DisposesLiveHandler()
                 kvp = default;
             }
 
-            // Act - 1
+            // Let's verify the the ActiveHandlerTrackingEntry is gone. This would be prevent
+            // the handler from being disposed if it was still rooted.
+            Assert.Empty(factory.ActiveEntryState);
+
+            // Act - 1 - Run a cleanup cycle, this will not dispose the handler, because the client is still live.
```
rynowak: The error is coming from L493 and basically says that L490 was able to dispose the message handler. It's not intended that L490 dispose the message handler, it's supposed to survive this cycle - which implies that `client1` been GCed already.

PR #1222, opened 2019-03-06
add FlakyAttribute to mark flaky tests

anurse:
part of aspnet/AspNetCore#8237

The attribute does nothing but apply traits right now, we still need to change our build to respect those traits. As a result, this PR will **fail** right now since it will run the "Flaky on AzDO" test :).

Despite being a draft, the `FlakyAttribute` and friends code is **ready for review**.

Examples on how to run/skip

* Run all **non-flaky** tests on AzDO: `dotnet test --filter "Flaky:AzDO!=true"`
* Run all **non-flaky** tests on Helix Queue "X": `dotnet test --filter "Flaky:Helix:all!=true&Flaky:Helix:X!=true`
* Run all **flaky** tests on AzDO: `dotnet test --filter "Flaky:AzDO=true"`

TODO:
* [x] Add build goop to make this all work in Extensions (which uses Arcade)
* [x] Add support for selecting AzDO jobs to skip on

After this PR:
* Add build goop for aspnet/AspNetCore

### DIFF
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -22,7 +22,6 @@
       <!-- Replace the previous runs -->
       <TestToRun Remove="@(TestToRun)" />
 
-
       <!-- TODO: Replace with below when https://github.com/dotnet/arcade/issues/2182 is in -->
       <!-- <TestToRun Include="@(NonFlakyRuns);@(FlakyRuns)" /> -->
       <TestToRun Include="@(NonFlakyRuns)" />
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -1,8 +1,8 @@
 <Project>
   <Target Name="_GenerateTestPasses" DependsOnTargets="$(_GetTestsToRunTarget)">
     <PropertyGroup>
-      <_FlakyRunAdditionalArgs>-trait "Flaky:AzDO=true"</_FlakyRunAdditionalArgs>
-      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:AzDO=true"</_NonFlakyRunAdditionalArgs>
+      <_FlakyRunAdditionalArgs>-trait "Flaky:AzDO:all=true"</_FlakyRunAdditionalArgs>
+      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:AzDO:all=true"</_NonFlakyRunAdditionalArgs>
     </PropertyGroup>
     <ItemGroup>
       <!-- Take the TestToRun values and update them to only run non-flaky tests -->
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -22,10 +22,10 @@
       <!-- Replace the previous runs -->
       <TestToRun Remove="@(TestToRun)" />
 
-      <!-- <TestToRun Include="@(NonFlakyRuns)" /> -->
 
-      <!-- TODO: Uncomment when [arcade issue to enable ignoring exit code] is in -->
-      <TestToRun Include="@(NonFlakyRuns);@(FlakyRuns)" />
+      <!-- TODO: Replace with below when https://github.com/dotnet/arcade/issues/2182 is in -->
+      <!-- <TestToRun Include="@(NonFlakyRuns);@(FlakyRuns)" /> -->
+      <TestToRun Include="@(NonFlakyRuns)" />
     </ItemGroup>
   </Target>
 
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -1,8 +1,8 @@
 <Project>
   <Target Name="_GenerateTestPasses" DependsOnTargets="$(_GetTestsToRunTarget)">
     <PropertyGroup>
-      <_FlakyRunAdditionalArgs>-trait "Flaky:AzP:OS:all=true" -trait "Flazy:AzP:OS:$(AGENT_OS)=true"</_FlakyRunAdditionalArgs>
-      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:AzP:OS:all=true" -trait "Flazy:AzP:OS:$(AGENT_OS)=true"</_NonFlakyRunAdditionalArgs>
+      <_FlakyRunAdditionalArgs>-trait "Flaky:All=true" -trait "Flaky:AzP:All=true" -trait "Flaky:AzP:OS:$(AGENT_OS)=true"</_FlakyRunAdditionalArgs>
+      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:All=true" -notrait "Flaky:AzP:All=true" -notrait "Flaky:AzP:OS:$(AGENT_OS)=true"</_NonFlakyRunAdditionalArgs>
     </PropertyGroup>
     <ItemGroup>
       <!-- Take the TestToRun values and update them to only run non-flaky tests -->
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -1,8 +1,8 @@
 <Project>
   <Target Name="_GenerateTestPasses" DependsOnTargets="$(_GetTestsToRunTarget)">
     <PropertyGroup>
-      <_FlakyRunAdditionalArgs>-trait "Flaky:AzP:all=true"</_FlakyRunAdditionalArgs>
-      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:AzP:all=true"</_NonFlakyRunAdditionalArgs>
+      <_FlakyRunAdditionalArgs>-trait "Flaky:AzP:OS:all=true" -trait "Flazy:AzP:OS:$(AGENT_OS)=true"</_FlakyRunAdditionalArgs>
+      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:AzP:OS:all=true"< -trait "Flazy:AzP:OS:$(AGENT_OS)=true"</_NonFlakyRunAdditionalArgs>
     </PropertyGroup>
     <ItemGroup>
       <!-- Take the TestToRun values and update them to only run non-flaky tests -->
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -1,0 +1,34 @@
+<Project>
+  <Target Name="_GenerateTestPasses" DependsOnTargets="$(_GetTestsToRunTarget)">
+    <PropertyGroup>
+      <_FlakyRunAdditionalArgs>-trait "Flaky:AzDO=true"</_FlakyRunAdditionalArgs>
+      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:AzDO=true"</_NonFlakyRunAdditionalArgs>
+    </PropertyGroup>
+    <ItemGroup>
+      <!-- Take the TestToRun values and update them to only run non-flaky tests -->
+      <NonFlakyRuns Include="@(TestToRun)">
+        <TestRunnerAdditionalArguments>%(TestRunnerAdditionalArguments) $(_NonFlakyRunAdditionalArgs)</TestRunnerAdditionalArguments>
+      </NonFlakyRuns>
+
+      <!-- Add new runs that do run flaky tests but ignore the exit code -->
+      <FlakyRuns Include="@(TestToRun)">
+        <TestRunnerAdditionalArguments>%(TestRunnerAdditionalArguments) $(_FlakyRunAdditionalArgs)</TestRunnerAdditionalArguments>
+        <TestRunnerIgnoreExitCode>true</TestRunnerIgnoreExitCode>
+        <ResultsHtmlPath>$([System.IO.Path]::ChangeExtension(%(ResultsHtmlPath), '.flaky.html'))</ResultsHtmlPath>
+        <ResultsStdOutPath>$([System.IO.Path]::ChangeExtension(%(ResultsStdOutPath), '.flaky.log'))</ResultsStdOutPath>
+        <ResultsXmlPath>$([System.IO.Path]::ChangeExtension(%(ResultsXmlPath), '.flaky.xml'))</ResultsXmlPath>
+      </FlakyRuns>
+
+      <!-- Replace the previous runs -->
+      <TestToRun Remove="@(TestToRun)" />
+
+      <!-- <TestToRun Include="@(NonFlakyRuns)" /> -->
+
+      <!-- TODO: Uncomment when [arcade issue to enable ignoring exit code] is in -->
+      <TestToRun Include="@(NonFlakyRuns);@(FlakyRuns)" />
+    </ItemGroup>
+  </Target>
+
+  <!-- Replace Arcade's Test target with a custom one -->
+  <Target Name="Test" DependsOnTargets="_GenerateTestPasses;RunTests" Condition="'$(IsUnitTestProject)' == 'true' or '$(IsPerformanceTestProject)' == 'true'" />
+</Project>
--- a/eng/FlakyTests.targets
+++ b/eng/FlakyTests.targets
@@ -2,7 +2,7 @@
   <Target Name="_GenerateTestPasses" DependsOnTargets="$(_GetTestsToRunTarget)">
     <PropertyGroup>
       <_FlakyRunAdditionalArgs>-trait "Flaky:AzP:OS:all=true" -trait "Flazy:AzP:OS:$(AGENT_OS)=true"</_FlakyRunAdditionalArgs>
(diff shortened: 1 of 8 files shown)


src/TestingUtils/Microsoft.AspNetCore.Testing/src/HelixQueues.cs
```diff
@@ -0,0 +1,28 @@
...
+        public const string Redhat7 = "Redhat.7." + HelixSuffix + ";";
+        public const string Debian9 = "Debian.9." + HelixSuffix + ";";
+        public const string Debian8 = "Debian.8." + HelixSuffix + ";";
+        public const string Centos7 = "Centos.7." + HelixSuffix + ";";
+        public const string Ubuntu1604 = "Ubuntu.1604." + HelixSuffix + ";";
+        public const string Ubuntu1810 = "Ubuntu.1810." + HelixSuffix + ";";
+        public const string macOS1012 = "OSX.1012." + HelixSuffix + ";";
+        public const string Windows10 = "Windows.10.Amd64.ClientRS4.VS2017.Open;"; // Doesn't have the default suffix!
+
+        private const string HelixSuffix = "Amd64.Open";
```
HaoK: Unfortunately we did/will have Arm64 queues too, they just were exploding so they aren't ready for frontline action yet.  Maybe we need a Flaky Queue too :)

anurse: I don't think we need a flaky queue, with this logic we can have our test scripts run both flaky **and** non-flaky tests, but only fail the build on non-flaky failures.

We can definitely update this as we have new queue explosions :). I can put `Amd64` in the field names for now to ensure they don't break when we add `Arm64` though.

src/TestingUtils/Microsoft.AspNetCore.Testing/test/FlakyAttributeTest.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                throw new Exception("Flaky on Helix!");
+            }
+        }
+
+        [Fact]
+        [Flaky("http://example.com", OnAzDO = false, OnHelixQueues = HelixQueues.macOS1012Amd64 + HelixQueues.Fedora28Amd64)]
+        public void FlakyInSpecificHelixQueue()
+        {
+            var queueName = Environment.GetEnvironmentVariable("HELIX");
```
anurse: @HaoK I noticed aspnet/AspNetCore#8231 is adding this env var, can you double-check that my logic here seems sound? The idea is to actually make this test properly fail on the specified Helix queues so we can leave it in as validation that flaky tests are indeed treated properly, so I want this test to actually be correctly detecting the queue it's running in :)

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+        public string GitHubIssueUrl { get; }
+
+        /// <summary>
+        /// Gets or sets a list of helix queues on which this test is flaky. Defaults to <see cref="HelixQueues.All"/> indicating it is flaky on all Helix queues. See
+        /// <see cref="HelixQueues"/> for a list of valid values.
+        /// </summary>
+        public string OnHelixQueues
+        {
+            get => string.Join(";", _queues);
+            set => _queues = new List<string>(value.Split(';'));
```
Eilon: Do you need to pass in `StringSplitOptions.RemoveEmptyEntries`?

src/TestingUtils/Microsoft.AspNetCore.Testing/test/FlakyAttributeTest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Microsoft.AspNetCore.Testing.Tests
+{
+    public class FlakyAttributeTest
+    {
+        [Fact]
+        [Flaky("http://example.com")]
+        public void AlwaysFlaky()
+        {
+            throw new Exception("Flakey!");
```
Eilon: Flaky

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
anurse: Probably, though it's not broken without it I believe. It would just generate the trait `Flaky:Helix:=true` which we never care about :).

src/TestingUtils/Microsoft.AspNetCore.Testing/test/FlakyAttributeTest.cs
anurse: Yep, also I don't want this test to fail for dev builds so I'll have it detect either AzDO or Helix and only fail there.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        /// <summary>
+        /// Gets or sets a boolean indicating if this test is flaky on Azure DevOps Pipelines. Defaults to <see langword="true" />.
+        /// </summary>
+        public bool OnAzDO { get; set; } = true;
```
pakrym: Does this need AzDO build definition name (OS)?

src/TestingUtils/Microsoft.AspNetCore.Testing/src/HelixQueues.cs
HaoK: Yeah what I mean is its possible we will want to run on new queues (i.e. arm testing) where the queues themselves have a high failure rate, today we just removed the queue entirely, but it would be nice to be able to run all tests on a flaky queue (just in the don't fail the build bucket).  We would have to mark every test with `[Flaky(OnHelixQueues="imsoflaky")]` I think without the concept.  Not something we need now, but something to keep in mind for the future

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
anurse: Perhaps

pakrym:
This needs to be merged with https://github.com/aspnet/AspNetCore/pull/8231

anurse:
> This needs to be merged with [aspnet/AspNetCore#8231](https://github.com/aspnet/AspNetCore/pull/8231)

Yes, but it's slightly tricky since `SkipOnHelixAttribute` is a shared-source file in aspnet/AspNetCore (not sure why it wasn't just added to this project since it depends on `ITestCondition` from it)

src/TestingUtils/Microsoft.AspNetCore.Testing/test/FlakyAttributeTest.cs
HaoK: Might be time to have a HelixUtils to consolidate all the helix stuff in one central file, or you could use SkipOnHelixAttribute.GetTargetHelixQueue() when that PR is merged.

Logic looks fine. One other pivot we should probably consider, is adding the concept of Platform as well, so there's an easy way to mark `Flaky(OnlyPlatforms = "OSX;Linux")` so we don't have to enumerate every queue

So maybe we need:
```
   public class HelixQueue {
       public string Id;
       public string Platform
   }
   public static class HelixPlatforms {
       public string Windows = "Windows";
       public string OSX = "OSX";
       public string Linux = "Linux";
   }
```

pakrym:
> Yes, but it's slightly tricky since SkipOnHelixAttribute is a shared-source file in aspnet/AspNetCore (not sure why it wasn't just added to this project since it depends on ITestCondition from it)

I wonder if SkipOnHelixAttribute even needs an ability to specify queues. Is there a case where we want to skip non-flaky tests on a single helix queue?

HaoK:
Yeah skipping tests on specific platform/queues seems like it would always be useful, but maybe we already have that with OSSkip. It might just be a short/medium term thing where some tests don't work on some helix queues that aren't the same set as OSSkip

src/TestingUtils/Microsoft.AspNetCore.Testing/test/FlakyAttributeTest.cs
```diff
@@ -0,0 +1,59 @@
...
+                var failingQueues = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { HelixQueues.macOS1012Amd64, HelixQueues.Fedora28Amd64 };
+                if (failingQueues.Contains(queueName))
+                {
+                    throw new Exception("Flaky on Helix!");
+                }
+            }
+        }
+
+        [Fact]
+        [Flaky("http://example.com", OnHelixQueues = HelixQueues.None)]
```
jkotalik: Can we use these on a few existing tests we know are flaky? Start putting the badge of shame up.

anurse: Not yet. We'll need build goop to actually exclude them

anurse: So you could use it when this is in, but it won't actually do anything.

anurse: To make it more fun, it looks like Extensions and AspNetCore will need slightly different build goop, but it looks doable.

HaoK: The good news is that helix isn't used on extensions... so you won't be getting many flaky helix tests on there hopefully :)

eng/FlakyTests.targets
```diff
@@ -0,0 +1,34 @@
...
+        <ResultsXmlPath>$([System.IO.Path]::ChangeExtension(%(ResultsXmlPath), '.flaky.xml'))</ResultsXmlPath>
+      </FlakyRuns>
+
+      <!-- Replace the previous runs -->
+      <TestToRun Remove="@(TestToRun)" />
+
+
+      <!-- TODO: Replace with below when https://github.com/dotnet/arcade/issues/2182 is in -->
+      <!-- <TestToRun Include="@(NonFlakyRuns);@(FlakyRuns)" /> -->
+      <TestToRun Include="@(NonFlakyRuns)" />
```
anurse: Note that we **do not run flaky tests at all** right now because of https://github.com/dotnet/arcade/issues/2182

anurse:
Our AzDO jobs are not named as consistently as our Helix queues. Perhaps we should do that before going too far down the path of selective AzDO flakiness? For example, this is what the helper static class looks like for AzDO (equiv of `HelixQueues`):

```csharp
    public static class AzP
    {
        // Pipeline names end in ';' because it makes it easier to concat these into a list using a constant expression:
        //  AzurePipelines.macOS + AzurePipelines.Windows

        public const string All = "all;";
        public const string None = "none;";

        public static class Extensions
        {
            public const string CodeCheck = "Code check Job;";
            public const string WindowsDebug = "Windows debug;";
            public const string WindowsRelease = "Windows release;";
            public const string Windows = WindowsDebug + WindowsRelease;
            public const string UbuntuDebug = "Ubuntu 16.04 debug;";
            public const string UbuntuRelease = "Ubuntu 16.04 release;";
            public const string Ubuntu = UbuntuDebug + UbuntuRelease;
            public const string macOSDebug = "OSX debug";
            public const string macOSRelease = "OSX release";
            public const string macOS = macOSDebug + macOSRelease;
        }

        public static class AspNetCore
        {
            public const string CodeCheck = "Code check Job;";
            public const string Windows = "Test: Windows Server 2016 x64 Job;";
            public const string macOS = "Test: macOS 10.13 Job;";
            public const string Ubuntu = "Test: Ubuntu 16.04 x64 Job;";
        }
    }
```

The name we get from AzDO appears to be [`Agent.JobName`](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#agent-variables) which is the display name (ugly). We could change our YAML to define a specific environment variable with a consistent set of names across both Extensions and AspNetCore (and EF?) to help out here. My thinking is that we leave AzDO as a hammer for now (on for all or off for all) and file an issue to track improvement here. Thoughts?

src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
```diff
@@ -91,6 +91,14 @@ public virtual void Initialize(MethodInfo methodInfo, object[] testMethodArgumen
 
         public virtual void Dispose()
         {
+            if(_testLog == null)
+            {
+                // It seems like sometimes the MSBuild goop that adds the test framework can end up in a bad state and not actually add it
+                // Not sure yet why that happens but the exception isn't clear so I'm adding this error so we can detect it better.
+                // -anurse
+                throw new InvalidOperationException("LoggedTest base class was used but nothing initialized it! The test framework may not be enabled. Try cleaning your 'obj' directory.");
```
anurse: Hit this while testing... not sure why, but want to make it a little easier to catch and possibly diagnose if it happens again. It doesn't seem to happen at all in AzP because we use a clean environment. The ".cache" file for my assembly attributes was broken somehow which meant the MSBuild goop never actually generated new Assembly Attributes.

pakrym:
> My thinking is that we leave AzDO as a hammer for now (on for all or off for all) and file an issue to track improvement here. Thoughts?

How is Helix more important then AzDO here? Just because we don't have that many queues on AzDO?

> Our AzDO jobs are not named as consistently as our Helix queues

You can base AzDo skip on the fact that we are running there + `Agent.OS` for OS.

HaoK:
At least for AspNetCore the current plan is to try and move all testing to helix, so eventually we won't have any tests running on azdo

src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
HaoK: Ah right, I think we are missing an issue tracking reenabling this on helix, If you can isolate/fix the root cause, we could take this out https://github.com/aspnet/AspNetCore/blob/master/eng/targets/Helix.props#L18

anurse:
> You can base AzDo skip on the fact that we are running there + `Agent.OS` for OS.

Let's do that. I'll update.

src/TestingUtils/Microsoft.AspNetCore.Testing/test/FlakyAttributeTest.cs
anurse: We could just do that by building combo-`const`s like so:

```csharp
        // Common Groupings
        public const string Ubuntu = Ubuntu1604Amd64 + Ubuntu1810Amd64;
        public const string Centos = Centos7Amd64;
        public const string Debian = Debian8Amd64 + Debian9Amd64;
        public const string Redhat = Redhat7Amd64;
        public const string Fedora = Fedora27Amd64 + Fedora28Amd64;

        public const string Windows = Windows10Amd64;
        public const string macOS = macOS1012Amd64;
        public const string Linux = Ubuntu + Centos + Debian + Redhat + Fedora;
```

We could also add other pivots to the flaky traits. I already changed it a bit so that the trait we insert is `Flaky:Helix:Queue:[queuename]`, we could have other facets in that sequence and as long as the script running the tests injects all the necessary filters, it should work.

anurse:
Ok, I think this is ready to go. I'd like another review sign-off (since all my changes) before merging. The attribute works a little differently now, instead of separate properties, there's just one `params` constructor argument that takes a list of "filters". Filters are strings like `AzP:All`, `AzP:OS:Darwin` or `Helix:Queue:Abc` which come from two helper static classes. These strings are simply prepended with `Flaky:` and added as xUnit traits with a value of `true`.

The build script specifies all the traits it wants to select/exclude, for example:

```xml
      <_FlakyRunAdditionalArgs>-trait "Flaky:All=true" -trait "Flaky:AzP:All=true" -trait "Flaky:AzP:OS:$(AGENT_OS)=true"</_FlakyRunAdditionalArgs>
      <_NonFlakyRunAdditionalArgs>-notrait "Flaky:All=true" -notrait "Flaky:AzP:All=true" -notrait "Flaky:AzP:OS:$(AGENT_OS)=true"</_NonFlakyRunAdditionalArgs>
```

If no "filters" are provided, the test is considered flaky everywhere:

```csharp
        [Flaky("http://example.com")]
        public void AlwaysFlaky() {...}
```

If any "filters" are provided, **all** relevant filters must be provided:

```csharp
        [Flaky("http://example.com", HelixQueues.All)]
        public void FlakyInHelixOnly() {...}

        [Flaky("http://example.com", HelixQueues.All, AzurePipelines.Windows)]
        public void FlakyInHelixAndAzPWindowsOnly() {...}
```

anurse:
Whoops, didn't finish updating the test 😺

PR #1226, opened 2019-03-07
hacky first pass at using AzP's MSBuild logger

anurse:
Just want to see if it's possible for now ;)


pakrym:
When we figure this out we might want to contribute the change to arcade.

rynowak:
@tmat - know anything about this?

tmat:
@rainersigwald Was looking into a logger.

anurse:
> When we figure this out we might want to contribute the change to arcade.

Yeah, this is indeed the plan.

> @rainersigwald Was looking into a logger.

Cool, the logger looks relatively straightforward to use (once you figure out the magic incantations). The trick is getting the DLL since it only seems to be available in this ZIP file on the magic blob share.

anurse:
We really need to start getting this data flowing, since a lot of our build tracking automation uses the AzP log output, so I'm going to press forward but would love to sync-up and unify our work in to Arcade.

azure-pipelines.yml
```diff
@@ -86,10 +86,7 @@ jobs:
       steps:
       - checkout: self
         clean: true
-      - script: eng\common\cibuild.cmd
-          -configuration $(_BuildConfig)
-          -prepareMachine
-          $(_BuildArgs)
+      - powershell: eng\azpbuild.ps1 -configuration $(_BuildConfig) $(_BuildArgs)
```
rainersigwald: It's not at all clear to me why `$(_BuildArgs)` isn't getting expanded by Pipelines and is instead passed all the way down to powershell. Maybe try the long form with explicit `arguments`?

```suggestion
      - task: PowerShell@2
        inputs:
          filePath: eng\azpbuild.ps1
          arguments: -configuration $(_BuildConfig) $(_BuildArgs)
```

anurse: I think it's because there is a combination in the `matrix` that doesn't define that variable. Just added something that might resolve that.

rainersigwald: Nice! Now you're just up against the "pipelines doesn't plumb the errors through to GitHub Checks" problem, looks like--the summary page looks nice.

pakrym:
@anurse 👍 looks great.

anurse:
Except it isn't working ;)

pakrym:
image

anurse:
Well that wasn't there earlier... I was sure I was looking at the right results too. Cool

eng/common/build.sh
```diff
@@ -137,13 +137,16 @@ while [[ $# > 0 ]]; do
...
       ;;
-    /m:*)
+    -m:*|/m:*)
       properties="$properties $1"
       ;;
-    /bl:*)
+    -bl:*|/bl:*)
+      properties="$properties $1"
+      ;;
+    -dl:*|/dl:*)
```
anurse: This change will need to be made in Arcade as well so we don't get out of sync. I'll put that together.

anurse: https://github.com/dotnet/arcade/pull/2191

eng/azpbuild.ps1
```diff
@@ -0,0 +1,40 @@
...
+if (!(Test-Path $tmpDir)) {
+    mkdir $tmpDir | Out-Null
+}
+
+# Until we figure out a better way to get this...
+# This the same URL used by the Azure tasks: https://github.com/Microsoft/azure-pipelines-tasks/blob/da57c7efbf858a628786c869d7d14b887d1d90d7/Tasks/Common/MSBuildHelpers/make.json#L5
+# And it's HTTPS, so I think we can trust it.
+$loggerUrl = "https://vstsagenttools.blob.core.windows.net/tools/msbuildlogger/3/msbuildlogger.zip"
+Invoke-WebRequest $loggerUrl -OutFile $loggersZip
+Expand-Archive $loggersZip -DestinationPath $tmpDir -Force
```
anurse: Before removing the "hacky" from the title and undrafting this, I want to see how people feel about this. (@Eilon @pakrym @dougbu @ryanbrandenburg specifically)

We need to get `Microsoft.TeamFoundation.DistributedTask.MSBuild.Logger.dll` on the machine. The URL we're using is what is used when AzP builds their own tasks, but I believe in that case the DLL lives in the Task "bundle" that gets deployed to the agent rather than being downloaded on the fly. There are a few issues with how we're doing this:

1. We don't control this blob share. The file could move or be replaced. It is owned by AzP and accessed via HTTPS, but they could just decide to move it.
1. Since we're downloading a remote file on build, there's a very small possibility for new flakiness here if there's a network disruption. Since we also need NuGet which is on Azure Blobs, I'm not too worried about this.

I do think we should follow up with AzDO about making this logger more readily available so we don't have to do this hack.

pakrym: I say let's continue using this blob just don't fail the build if a download fails.

anurse: That should be feasible and would reduce the flakiness. We could also embed a hash and do a hash check.

Eilon: @pakrym 's idea seems to be a reasonable strategy for now.

pakrym: > We could also embed a hash and do a hash check.

Why?

rainersigwald: One possible other option: include a dummy `MSBuild` task invocation in the build YAML, causing Pipelines to make the task (and its copy of the DLL) available on the machine. Then find and use it from there (no idea how to do this).

This might also work only for Windows.

On balance, I prefer your approach.

@chcosta do you have a Pipelines contact we can run this by?

chcosta: @vtbassmat, thoughts?

chcosta: @vtbassmatt thoughts?

eng/common/build.sh
ryanbrandenburg: It seems like you already know and are just trying to stay unblocked, but I'll say it for posterity: Don't edit this folder by hand: https://github.com/aspnet/Extensions/blob/master/eng/common/README.md. Instead make a PR to arcade then get the new version when it's merged. Unblocking development temporarily as @anurse has done here by editing local files is fine if you have a PR out and are waiting for it to be merged, but DO NOT merge it that way.

src/Configuration/Config.Abstractions/src/ConfigurationPath.cs
```diff
@@ -23,6 +23,7 @@ public static class ConfigurationPath
         /// <returns>The combined path.</returns>
         public static string Combine(params string[] pathSegments)
         {
+            Do you even code bro ?
```
ryanbrandenburg: Wat?

anurse: Need a build error to verify that it's working. I'll remove it before merging.

eng/azpbuild.ps1
anurse: To summarize: We use a custom PowerShell script to launch MSBuild but still want all the `##vso` logging goop that comes from the logger the official MSBuild task uses. However, we need to get that logger on the box somehow in order to use it. Right now, I'm just getting it from the URL listed in https://github.com/Microsoft/azure-pipelines-tasks/blob/da57c7efbf858a628786c869d7d14b887d1d90d7/Tasks/Common/MSBuildHelpers/make.json#L5 during the build script itself but that seems hacky

eng/azpbuild.ps1
```diff
@@ -0,0 +1,40 @@
+param(
+    [string]$Configuration = "Release",
+    [Parameter(ValueFromRemainingArguments = $true)][String[]]$OtherArgs
+)
+
+$repoRoot = Split-Path $PSScriptRoot -Parent
```
dougbu: Move this script down into `eng/scripts/` and get the parent's parent here

eng/azpbuild.sh
```diff
@@ -0,0 +1,42 @@
...
+while [[ -h "$source" ]]; do
+  scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
+  source="$(readlink "$source")"
+  # if $source was a relative symlink, we need to resolve it relative to the path where the
+  # symlink file was located
+  [[ $source != /* ]] && source="$scriptroot/$source"
+done
+scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
+
+repoRoot="$( cd "$scriptroot/.." && pwd )"
```
dougbu: Also move this one into `eng/scripts/`

eng/azpbuild.sh
```diff
@@ -0,0 +1,42 @@
...
+loggerRsp="$tmpDir/logger.rsp"
+loggersZip="$tmpDir/msbuildlogger.zip"
+loggerAssemblyName="Microsoft.TeamFoundation.DistributedTask.MSBuild.Logger.dll"
+loggerAssembly="$tmpDir/$loggerAssemblyName"
+
+# Until we figure out a better way to get this...
+# This the same URL used by the Azure tasks: https://github.com/Microsoft/azure-pipelines-tasks/blob/da57c7efbf858a628786c869d7d14b887d1d90d7/Tasks/Common/MSBuildHelpers/make.json#L5
+# And it's HTTPS, so I think we can trust it.
+echo "Downloading Azure Pipelines MSBuild logger..."
+curl -sSL https://vstsagenttools.blob.core.windows.net/tools/msbuildlogger/3/msbuildlogger.zip "-o$loggersZip" >/dev/null
```
dougbu: There should be a space between `-o` and the output filename.

The other option is to `cd` to the `$tmpDir` and use `-O`. That works because the filename in the URL matches what you want on disk.

eng/azpbuild.sh
```diff
@@ -0,0 +1,42 @@
...
+echo "Extracting logger..."
+unzip -p "$loggersZip" "$loggerAssemblyName" > "$loggerAssembly"
```
dougbu: If you're already in the `$tmpDir`, no reason for the redirection here. `unzip "$loggersZip" "$loggerAssemblyName"` will do the **Right Thing:tm:** and be more efficient. Add `-b` to be sure the file is treated as binary.

src/Configuration/Config.Abstractions/src/ConfigurationPath.cs
dougbu: Yes, please remove 😺

eng/azpbuild.ps1
vtbassmatt: Grabbing the blob from the network is the better option. Hitting an HTTPS endpoint is the same from everywhere; different agent builds could have subtly different behavior about where they lay out the task cache and so on.

The non-hacky answer is to resolve the question of making the logger available as a NuGet package. I'll ping that thread again.

eng/azpbuild.sh
anurse: I'd rather use `-o` so we fix the file name and know what it is for later.

I definitely had `-o [filename]` at first and `curl` errored. I'll try it again though.

anurse: The space is certainly not _required_ though it certainly is my stylistic preference.

src/Configuration/Config.Abstractions/src/ConfigurationPath.cs
anurse:
```suggestion
```

eng/azpbuild.sh
anurse: Well, I'm *not* in `$tmpDir`, quite intentionally.

anurse:
Removed @Tratcher as reviewer. You're welcome to look but were only added because I chose Configuration as my victim for testing that a build failure is reported properly :)

anurse:
I don't understand why sometimes it's working and sometimes it isn't... It seems random. Investigating further.

anurse:
Hrm, "Windows release" and "OSX debug" both ran without unexpected errors but aren't reporting the failure properly...

anurse:
And now it's completely different ones that aren't working. I'm just going to fix the build error and merge this I think...

PR #1227, opened 2019-03-07
Use Utf8JsonWriter in EventSourceLogger

pakrym:
Fixes: https://github.com/aspnet/Extensions/issues/1013

### DIFF
--- a/src/Logging/Logging.EventSource/src/EventSourceLogger.cs
+++ b/src/Logging/Logging.EventSource/src/EventSourceLogger.cs
@@ -5,8 +5,10 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.Tracing;
 using System.IO;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading;
-using Newtonsoft.Json;
 
 namespace Microsoft.Extensions.Logging.EventSource
 {
@@ -205,19 +207,19 @@ namespace Microsoft.Extensions.Logging.EventSource
 
         private string ToJson(IReadOnlyList<KeyValuePair<string, string>> keyValues)
         {
-            var sw = new StringWriter();
-            var writer = new JsonTextWriter(sw);
-            writer.DateFormatString = "O"; // ISO 8601
+            var arrayBufferWriter = new ArrayBufferWriter<byte>();
+            var writer = new Utf8JsonWriter(arrayBufferWriter);
 
             writer.WriteStartObject();
-            for (int i = 0; i < keyValues.Count; i++)
+            foreach (var keyValue in keyValues)
             {
-                var keyValue = keyValues[i];
-                writer.WritePropertyName(keyValue.Key, true);
-                writer.WriteValue(keyValue.Value);
+                writer.WriteString(keyValue.Key, keyValue.Value, true);
             }
             writer.WriteEndObject();
-            return sw.ToString();
+
+            writer.Flush(true);
+
+            return Encoding.UTF8.GetString(arrayBufferWriter.WrittenMemory.ToArray());
         }
     }
 }
--- a/src/Logging/Logging.EventSource/src/Microsoft.Extensions.Logging.EventSource.csproj
+++ b/src/Logging/Logging.EventSource/src/Microsoft.Extensions.Logging.EventSource.csproj
@@ -6,13 +6,17 @@
     <PackageTags>$(PackageTags);EventSource;ETW</PackageTags>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsShipping>true</IsShipping>
+    <NoWarn>$(NoWarn);CS3021;PKG0001</NoWarn>
   </PropertyGroup>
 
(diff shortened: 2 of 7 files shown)


src/Logging/Logging.EventSource/test/EventSourceLoggerTest.cs
```diff
@@ -692,7 +692,7 @@ private static class EventTypes
 
             { "E5JS", (e) => VerifySingleEvent(e, "Logger2", EventTypes.MessageJson, 5, null, LogLevel.Critical,
                 @"""ArgumentsJson"":{""stringParam"":""bar"",""int1Param"":""23"",""int2Param"":""45""",
-                @"""ExceptionJson"":{""TypeName"":""System.Exception"",""Message"":""oops"",""HResult"":""-2146233088"",""VerboseMessage"":""System.Exception: oops ---> System.Exception: inner oops") },
+                @"""ExceptionJson"":{""TypeName"":""System.Exception"",""Message"":""oops"",""HResult"":""-2146233088"",""VerboseMessage"":""System.Exception: oops ---\u003e System.Exception: inner oops") },
```
anurse: ?

I mean, it's the same, right? `\u003e` should be `>`

pakrym: Yep, Utf8JsonWriter just decides to escape it for some reason. The end result is equivalent so I didn't spend too much time digging.

anurse: Fair enough :)

anurse: Looks like https://github.com/dotnet/corefx/blob/3eacad602f7505f9f199dcd1992429506dd98330/src/System.Text.Json/src/System/Text/Json/Writer/JsonWriterHelper.Escaping.cs#L15-L17

```csharp
        // Only allow ASCII characters between ' ' (0x20) and '~' (0x7E), inclusively,
        // but exclude characters that need to be escaped as hex: '"', '\'', '&', '+', '<', '>', '`'
        // and exclude characters that need to be escaped by adding a backslash: '\n', '\r', '\t', '\\', '/', '\b', '\f'
```

It's surprising to me that `<` needs to be escaped in JSON. @ahsonkhan, any particular reason for this? Mostly curious but it does seem like a user would be surprised to have it escaped in a string...

ahsonkhan: The JSON escaping here is following a strict mode to be SDL compliant. This is following the semantics of the default Javascript escaper which also escapes all non-ascii characters (and some ascii as well, including those that show up in html). This is primarily for defense in depth since JSON payloads could be embedded within html.

We currently do not have a way to customize the escaping behavior, but I imagine a writer option to plug in a custom escaper. Something that would work with https://github.com/dotnet/corefx/issues/34830

Some related issues:
https://github.com/dotnet/corefx/issues/34958
https://github.com/dotnet/corefx/issues/35385

cc @GrabYourPitchforks

anurse: Sounds fine to me. I mostly just wanted to confirm that this was an intentional choice because it was a little surprising. Thanks!

src/Logging/Logging.EventSource/src/EventSourceLogger.cs
```diff
@@ -205,19 +207,19 @@ private ExceptionInfo GetExceptionInfo(Exception exception)
 
         private string ToJson(IReadOnlyList<KeyValuePair<string, string>> keyValues)
         {
-            var sw = new StringWriter();
-            var writer = new JsonTextWriter(sw);
-            writer.DateFormatString = "O"; // ISO 8601
+            var arrayBufferWriter = new ArrayBufferWriter<byte>();
```
ahsonkhan: Hmm... using an internal type, which wasn't meant to be used here. But hey, source package :)

src/Logging/Logging.EventSource/src/Microsoft.Extensions.Logging.EventSource.csproj
```diff
@@ -6,13 +6,17 @@
     <PackageTags>$(PackageTags);EventSource;ETW</PackageTags>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsShipping>true</IsShipping>
+    <NoWarn>$(NoWarn);CS3021</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="../../shared/*.cs" />
 
+    <Reference Include="Microsoft.Bcl.Json.Sources">
```
ahsonkhan: Does this library have to support netstandard2.0?

anurse: Yes, it was `netstandard2.0` in `2.2` and I don't think we have plans to change that. Most of `Microsoft.Extensions.*` is designed to remain `netstandard2.0`

PR #1248, opened 2019-03-13
clean up flaky filter constants

anurse:
Before people start using them, I'd like to clean up some things in the "Flaky Filters" constants.

* I had a `HelixQueues` type with queue names but they were formatted for use in `[Flaky]` and thus wouldn't work when just comparing the queue name against the `HELIX` env var.
* I wanted to group all the flaky filters into one type.

So I:
1. Merged `AzurePipelines` and `HelixQueues` classes into a single `FlakyOn` class (and set of nested classes) that provide the flaky filter syntax
1. Brought back a `HelixQueues` class that just has the plain queue names.
1. I also changed the attribute to require at least one filter be provided so that users have to explicitly write `FlakyOn.All` or some other filter.


src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
```diff
@@ -64,12 +64,29 @@ public sealed class FlakyAttribute : Attribute, ITraitAttribute
...
+            if(string.IsNullOrEmpty(firstFilter))
+            {
+                throw new ArgumentNullException(nameof(firstFilter));
+            }
+
             GitHubIssueUrl = gitHubIssueUrl;
-            Filters = new List<string>(filters);
+            var filters = new List<string>();
+            filters.Add(firstFilter);
+            filters.AddRange(additionalFilters);
```
anurse: This funky logic allows us to have a `params` but require at least one item be provided.

anurse:
Failure was "a failure to provision the agent" for Ubuntu. Retrying.

anurse:
The CI passed, it just seems to have not updated GitHub...

src/TestingUtils/Microsoft.AspNetCore.Testing/src/FlakyOn.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            public const string Fedora28Amd64 = QueuePrefix + HelixQueues.Fedora28Amd64;
+            public const string Fedora27Amd64 = QueuePrefix + HelixQueues.Fedora27Amd64;
+            public const string Redhat7Amd64 = QueuePrefix + HelixQueues.Redhat7Amd64;
+            public const string Debian9Amd64 = QueuePrefix + HelixQueues.Debian9Amd64;
+            public const string Debian8Amd64 = QueuePrefix + HelixQueues.Debian8Amd64;
+            public const string Centos7Amd64 = QueuePrefix + HelixQueues.Centos7Amd64;
+            public const string Ubuntu1604Amd64 = QueuePrefix + HelixQueues.Ubuntu1604Amd64;
+            public const string Ubuntu1810Amd64 = QueuePrefix + HelixQueues.Ubuntu1810Amd64;
+            public const string macOS1012Amd64 = QueuePrefix + HelixQueues.macOS1012Amd64;
```
HaoK: nit: macOS gets the special lowercase treatment? :)

anurse: I mean... it's the name of the OS (macOS). I definitely spent a non-zero amount of time fretting over that ;)

anurse:
> Will this require a reaction PR in AspNetCore?

No.

Issue #1255, opened 2019-03-14
Caching - Null value semantics


brandondahler:
### Describe the bug
When storing a null value in caches, `TryGetValue<TItem>(object key, out TItem value)` always returns `false` even when`TryGetValue(object key, out object value)` returns `true`.

### To Reproduce
``` C#
    class Program
    {
        static int Main(string[] args)
        {
            var memoryCache = new MemoryCache(Options.Create<MemoryCacheOptions>(new MemoryCacheOptions()));

            memoryCache.Set<object>("a", null);

			
            Console.WriteLine(memoryCache.TryGetValue("a", out _) == memoryCache.TryGetValue<object>("a", out _));

            return 0;
        }
    }
```

#### Expected behavior
Program prints `True`

#### Actual Result
Program prints `Flase`

### Recommended Solution
I believe TryGetValue<TItem> should be updated to always return `true` if the resulting value is null.  Pull request to follow.

### Workaround
Issue can be worked around using a wrapper object; however, it appears the intent of CacheEntry is to allow null value to be properly cached -- no other code deals with the null semantics in this way.

### Additional context
Appears to be a side-effect of the addition of https://github.com/aspnet/Caching/pull/395 .  `null is TItem` will always return `false` while the previous code `(TItem) null` would always successfully cast null to reference types.

I don't personally think this would be a breaking change, but could be considered a breaking change since it would be altering the way null semantics are handled for this particular function.

Tratcher:
That check is here:
https://github.com/aspnet/Extensions/blob/4b314dd3ce296dc1f951b91623ed070bccdaa918/src/Caching/Abstractions/src/MemoryCacheExtensions.cs#L27-L31

`is TItem` returns false for null values.

PR #1260, opened 2019-03-15
Update caching's null value semantics for TryGetValue<TItem>

brandondahler:
 - TryGetValue<TItem> will return true and assign the default value to result when a null object value is stored for a given key.
 - Tests added to validate resulting semantics of TryGetValue<TItem>

Addresses #1255 .

#### Notes
Test `TryGetValue_WillCreateDefaultValueAndSucceed_WhenValueNullForValueType` exercises a potentially less-than-desirable semantic that results from this -- storing a null object (of any type) and using TryGetValue<TItem> for any value type will return true and result in the default value for that value type.  This technically falls in line with the fact that TryGetValue<TItem> will return true and result in a null object of any type even if the originally stored type isn't the same as the requested type.

Changes were made while working for employer, Fifty Seven Pounds.  CLA should already be signed but can be re-signed if needed.

### DIFF
--- a/src/Caching/Abstractions/src/MemoryCacheExtensions.cs
+++ b/src/Caching/Abstractions/src/MemoryCacheExtensions.cs
@@ -24,6 +24,12 @@ namespace Microsoft.Extensions.Caching.Memory
         {
             if (cache.TryGetValue(key, out object result))
             {
+                if (result == null)
+                {
+                    value = default;
+                    return true;
+                }
+
                 if (result is TItem item)
                 {
                     value = item;
--- a/src/Caching/Memory/test/MemoryCacheSetAndRemoveTests.cs
+++ b/src/Caching/Memory/test/MemoryCacheSetAndRemoveTests.cs
@@ -213,6 +213,32 @@ namespace Microsoft.Extensions.Caching.Memory
             Assert.False(cache.TryGetValue(key, out string obj));
         }
 
+
+        [Fact]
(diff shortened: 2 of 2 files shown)


Tratcher:
I have to ask, what value do you get out of cashing a null value?

brandondahler:
In my use-case it denotes the difference between a result having not been fetched from the downstream service and resulting in a null value versus having not ever attempted to fetch the value from the downstream service.

A similar but much more trivially simple case might be:
``` C#
    class Program
    {
        static int Main(string[] args)
        {
            var memoryCache = new MemoryCache(Options.Create<MemoryCacheOptions>(new MemoryCacheOptions()));

            while (true)
            {
                Console.Write("Enter a date/time value: ");
                var enteredString = Console.ReadLine();
                
                if (string.isNullOrEmpty(enteredString))
                    break;
                
                DateTime? parsedDateTime;
                if (!memoryCache.TryGetValue<DateTime?>(enteredString, out parsedDateTime))
                {
                    if (DateTime.TryParse(enteredString, out var tempParsedDateTime))
                        parsedDateTime = tempParsedDateTime
                    else 
                        parsedDateTime = null
                    
                    memoryCache.Set(enteredString, parsedDateTime);
                }
                
                if (parsedDateTime.HasValue)
                    Console.WriteLine($"Parsed date/time value: {parsedDateTime.Value}");
                else
                    Console.WriteLine($"Unable to parse date/time value.");
                    
                Console.WriteLine();
            }
			
            return 0;
        }
    }
```

PR #1275, opened 2019-03-18
Remove unnecessary checks

hishamco:
No need for those checks, because we already checked them in `StringLocalizer`

### DIFF
--- a/src/Localization/Abstractions/src/StringLocalizerExtensions.cs
+++ b/src/Localization/Abstractions/src/StringLocalizerExtensions.cs
@@ -23,11 +23,6 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(stringLocalizer));
             }
 
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
-            }
-
             return stringLocalizer[name];
         }
 
@@ -46,11 +41,6 @@ namespace Microsoft.Extensions.Localization
             if (stringLocalizer == null)
             {
                 throw new ArgumentNullException(nameof(stringLocalizer));
-            }
-
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
             }
 
             return stringLocalizer[name, arguments];


Eilon:
We typically keep these checks because that makes it clearer to the immediate caller that invalid data is being passed in.

hishamco:
If it's like that, what about `GetString()` 😄, I'm sure we remove such checks from many places in localization APIs in the old repo, but seems we forgot this one

Eilon:
There could be some inconsistencies, but that's the reason.

hishamco:
Still I don't see any value to make those checks, or we should make everything consistent

hishamco:
@ryanbrandenburg is  this by intention or we miss it before?!!

Eilon:
Nothing was missed, I closed the PR because we didn't intend to merge it.

hishamco:
Frankly I saw inconsistency in some of the APIs, namespaces, code ... etc, but even in major releases we cannot do anything .. is this the policy for all the Microsoft repos or ASP.NET Core once

Eilon:
This isn't about major releases, it's just that we felt this change isn't worth it. The current parameter checks are correct and should remain.

hishamco:
No no .. I didn't mean this, I knew it is trivial .. but I'm talking in general coz I opened some issues in the past and closed with unconvinced answer .. my question is if there is inconsistent namespace as example, I'm not sure why we can't fixed in major release even if it doesn't have a major value!! I expect major release is for breaking changes and fix what we did it wrong previously

Eilon:
Major versions do *allow* for breaking changes, but we have to weight the *value* of the change against the *cost*. The cost isn't just for ASP.NET Core itself, but for all the customers. Changing the namespace of a type that is used in hundreds (or thousands (or tens of thousands)) of apps might take us 5 seconds to do, but will cost "the world" hundreds of hours, for no appreciable gain. Not enough people were "harmed" by an incorrect namespace to make it worth costing hundreds of hours of fix-up time (or a lot more).

Issue #1286, opened 2019-03-20
UseWindowsService should enable EventLog logger by default


anurse:
When you are building a generic host app using Windows Service lifetime, console logging isn't really useful. The EventLog is a good place to log this stuff. We should enable the EventLog logger, or some other relevant logger, by default. We should probably only forward Warnings/Errors to the EventLog to avoid spamming it.

Things to consider:
* EventLog logger isn't a scenario we've really used heavily, it might be ugly
* We log to the "Application" log, which is a pain, but doing anything else requires Admin :(.

Tratcher:
Note I didn't have any permission issues setting the event viewer Source when running as Local System. When I make a base user then this is the error I get:
```
Exception Info: System.AggregateException: An error occurred while writing to logger(s). (The source was not found, but some or all event logs could not be searched.  To create the source, you need permission to read all event logs to make sure that the new source name is unique.  Inaccessible logs: Security, State.) ---> System.Security.SecurityException: The source was not found, but some or all event logs could not be searched.  To create the source, you need permission to read all event logs to make sure that the new source name is unique.  Inaccessible logs: Security, State.
   at System.Diagnostics.EventLog.FindSourceRegistration(String source, String machineName, Boolean readOnly, Boolean wantToCreate)
   at System.Diagnostics.EventLog.SourceExists(String source, String machineName, Boolean wantToCreate)
   at System.Diagnostics.EventLogInternal.VerifyAndCreateSource(String sourceName, String currentMachineName)
   at System.Diagnostics.EventLogInternal.WriteEvent(EventInstance instance, Byte[] data, Object[] values)
   at System.Diagnostics.EventLog.WriteEvent(EventInstance instance, Byte[] data, Object[] values)
   at System.Diagnostics.EventLog.WriteEvent(EventInstance instance, Object[] values)
   at Microsoft.Extensions.Logging.EventLog.WindowsEventLog.WriteEntry(String message, EventLogEntryType type, Int32 eventID, Int16 category) in D:\github\Extensions\src\Logging\Logging.EventLog\src\WindowsEventLog.cs:line 24
   at Microsoft.Extensions.Logging.EventLog.EventLogLogger.WriteMessage(String message, EventLogEntryType eventLogEntryType, Int32 eventId) in D:\github\Extensions\src\Logging\Logging.EventLog\src\EventLogLogger.cs:line 122
[...]

PR #1296, opened 2019-03-22
 Enable EventLog by default for WindowsService

Tratcher:
#1286 The application's name will be used as the default source.

Additions:
- Generic host now provides a default application name.
- EventLog logger now handles errors when you're not allowed to create sources.
- EventLog support for IOptions

### DIFF
--- a/src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
+++ b/src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
@@ -27,43 +27,24 @@ namespace Microsoft.Extensions.Hosting
         /// <returns></returns>
         public static IHostBuilder UseWindowsService(this IHostBuilder hostBuilder)
         {
-            return hostBuilder.UseWindowsService((context, settings) =>
-            {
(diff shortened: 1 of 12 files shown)


src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
```diff
@@ -21,11 +27,41 @@ public static class WindowsServiceLifetimeHostBuilderExtensions
...
+        /// and enables logging to the event log.
+        /// </summary>
+        /// <remarks>
+        /// This is context aware and will only activate if it detects the process is running
+        /// as a Windows Service.
+        /// </remarks>
+        /// <param name="hostBuilder"></param>
+        /// <param name="configureEventLog"></param>
+        /// <returns></returns>
+        public static IHostBuilder UseWindowsService(this IHostBuilder hostBuilder, Action<HostBuilderContext, EventLogSettings> configureEventLog)
```
pakrym: If you add IOptions<EventLogSettings> support to EventLogLogger provider you can avoid having this overload and make EventLog logger more consistent with others.

Issue #1041, opened 2019-01-31
HealthCheckPublisherOptions.Period sets the value of .Delay


guardrex:
`HealthCheckPublisherOptions.Period` sets the value of `HealthCheckPublisherOptions.Delay` ...

```csharp
services.Configure<HealthCheckPublisherOptions>(options =>
{
    options.Period = TimeSpan.FromSeconds(2);
});

services.AddSingleton<IHealthCheckPublisher, TestPublisher>();
```

`PublishAsync` is called on default 30-second intervals ... **_not the 2 seconds specified_**. When this setting is made `HealthCheckPublisherOptions.Delay` receives the value assigned to `Period`.

https://github.com/aspnet/Extensions/blob/release/2.2/src/HealthChecks/HealthChecks/src/HealthCheckPublisherOptions.cs

guardrex:
@pranavkm I think Ryan is OOF for the conference. Do you have a sec to take a look? This is for the doc updates on `IHealthCheckPublisher` options + example ..... I'm having a devil of a time setting the options cleanly, especially `Period`.

I put the whole sample here :point_right: https://github.com/guardrex/HealthChecksSample

The important files are:

* [Publisher (IHealthCheckPublisher impl)](https://github.com/guardrex/HealthChecksSample/blob/master/ReadinessPublisher.cs)
* [Liveness Probe Startup](https://github.com/guardrex/HealthChecksSample/blob/master/LivenessProbeStartup.cs) &ndash; `Period` is set on Line 48.

For this scenario the sample is run with the `--scenario liveness` switch.

It's not honoring my 2-second setting ... it's using the 30-second default period.

pranavkm:
@glennc could you review this? You're probably more familiar with the code here than I am.

guardrex:
In the meantime, I need to load up 3.0 preview anyway for Blazor work, and I'll find out if this strange behavior goes away with 3.0 and ditching that funky workaround.

guardrex:
Moved the app to 3.0 preview2. I was able to remove the workaround line ...

```csharp
services.TryAddEnumerable(
                ServiceDescriptor.Singleton(typeof(IHostedService), 
                    typeof(HealthCheckPublisherOptions).Assembly
                        .GetType(HealthCheckServiceAssembly)));
```

... but the `Period` is still defaulting to 30-seconds, and IDK where it's getting that 2-second setting for the delay given that it should be 5-seconds by default ...

[...]

PR #1312, opened 2019-03-27
Port fix for #1041 to 2.2

rynowak:
## Impact

It's not possible to set the `Period` of the health check publisher hosted service without using reflection. The wrong backing field is set.

This isn't very discoverable because the bug manifests in a setting not being honoured. You might think you are publishing health data every 30 seconds, but it's going to always use the default period.

## Workaround

The workaround is to use reflection. 

## Risk

The risk here is that someone is setting `Period` - which actually sets `Delay` and they are relying on this detail. If a user actually wanted to set `Delay`, they can already do that.

### DIFF
--- a/eng/PatchConfig.props
+++ b/eng/PatchConfig.props
@@ -18,6 +18,7 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(VersionPrefix)' == '2.2.5' ">
     <PackagesInPatch>
+      Microsoft.Extensions.HealthChecks;
     </PackagesInPatch>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


vivmishra:
Approved for 2.2.5
Wait for branch to open.

dougbu:
PR needs at least two things:
- approval from someone familiar with the changed code
- updates to PatchConfig.props

dougbu:
@rynowak he deadline for merging 2.x fixes is currently the 15th. Please get this approved and in soon so that we can start working on official builds a bit early.

rynowak:
@dougbu @pranavkm - one of you want to mash the like button on this trivial change?

PR #1313, opened 2019-03-27
Update Polly to 7.1.0 and 3.0.0

rynowak:
@reisenberger @Eilon

### DIFF
--- a/eng/Dependencies.props
+++ b/eng/Dependencies.props
@@ -64,8 +64,8 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)" />
     <LatestPackageReference Include="Moq" Version="4.10.0" />
     <LatestPackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <LatestPackageReference Include="Polly.Extensions.Http" Version="2.0.1" />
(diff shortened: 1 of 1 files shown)


eng/Dependencies.props
```diff
@@ -64,8 +64,8 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)" />
     <LatestPackageReference Include="Moq" Version="4.10.0" />
     <LatestPackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <LatestPackageReference Include="Polly.Extensions.Http" Version="2.0.1" />
-    <LatestPackageReference Include="Polly" Version="6.0.1" />
+    <LatestPackageReference Include="Polly.Extensions.Http" Version="3.0.0" />
```
Eilon: I don't see this one on NuGet.org. Is it not published yet?

rynowak: It is not, we're staging this for them to update their package 😀

Eilon: Cool. Well it's "approved" from my perspective. Once the target package is on NuGet.org and this change here gets checked in, Component Governance (CG) will auto-register it.

reisenberger:
[Polly.Extensions.Http/3.0.0](https://www.nuget.org/packages/Polly.Extensions.Http/3.0.0) is now available on nuget (for use with this PR)

PR #1324, opened 2019-03-28
Fix #8612 JSInterop throws truncated call stack

rynowak:
The fix for this is to use more ExceptionDispatchInfo! Basically
everwhere that we handle an exception is now an EDI. It's easy to pass
these around and they do the right thing as far as perserving the stack
in.

I de-factored this code a little bit to make all of this work, but it's
now pretty unambiguously correct upon inspection.

I started out wanting to get rid of the continue-with because we've
found that pretty error prone overall. However making this method
async-void started to ask more questions than it answered. What happens
if serialization of the exception fails?

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 
 namespace Microsoft.JSInterop
@@ -72,8 +73,10 @@ namespace Microsoft.JSInterop
                 ? null
                 : jsRuntimeBaseInstance.ArgSerializerStrategy.FindDotNetObject(dotNetObjectId);
 
+            // Using ExceptionDispatchInfo here throughout because we want to always preserve
+            // original stack traces.
             object syncResult = null;
-            Exception syncException = null;
+            ExceptionDispatchInfo syncException = null;
 
             try
             {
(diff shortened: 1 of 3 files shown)


src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -72,54 +73,49 @@ public static void BeginInvoke(string callId, string assemblyName, string method
...
+                {
+                    if (t.Exception != null)
+                    {
+                        var exception = t.Exception.GetBaseException();
+                        jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, ExceptionDispatchInfo.Capture(exception));
+                    }
+
+                    var result = TaskGenericsUtil.GetTaskResult(task);
+                    jsRuntimeBaseInstance.EndInvokeDotNet(callId, true, result);
+                }, TaskScheduler.Default);
```
rynowak: This should be `TaskScheduler.Current`. Will fix.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -285,15 +286,5 @@ private static Assembly GetRequiredLoadedAssembly(string assemblyName)
             return loadedAssemblies.FirstOrDefault(a => a.GetName().Name.Equals(assemblyName, StringComparison.Ordinal))
                 ?? throw new ArgumentException($"There is no loaded assembly with the name '{assemblyName}'.");
         }
-
-        private static Exception UnwrapException(Exception ex)
```
rynowak: This logic isn't as complex anymore because I'm not trying to build a one-size-fits-all *unwrapper*.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -175,9 +171,14 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
                 return methodInfo.Invoke(targetInstance, suppliedArgs);
             }
-            catch (Exception ex)
+            catch (TargetInvocationException tie) when (tie.InnerException != null)
             {
-                throw UnwrapException(ex);
+                ExceptionDispatchInfo.Capture(tie.InnerException).Throw();
+                throw null; // unreachable
+            }
+            catch (Exception)
```
rynowak: This isn't needed. Will remove.

rynowak:
Note that many of these tests are still disabled due to https://github.com/aspnet/AspNetCore-Internal/issues/1733

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -80,6 +81,10 @@ internal void EndInvokeDotNet(string callId, bool success, object resultOrExcept
             {
                 resultOrException = resultOrException.ToString();
             }
+            else if (!success && resultOrException is ExceptionDispatchInfo edi)
```
javiercn: nit: Isn't `resultOrException is ExceptionDispatchInfo edi` enough test?

rynowak: This is similar to the test above for exceptions. I don't see a reason to change this since there aren't many tests for exceptions in general.

rynowak:
> I would also reenable the tests if everything passes on the CI. If we find out they are flaky then we should make them [Flaky]

I'm going to send a separate PR to re-enable these tests.

PR #1329, opened 2019-03-28
Re-enable tests skipped in JSInterop

rynowak:
Fixes: aspnet/AspNetCore-Internal#1733

It's unclear what the cause of the original failures are but it was a
consistent failure. If this passes on the CI I will merge.


rynowak:
It is the fail. Something is actually wrong.

ryanbrandenburg:
`Object of type 'Microsoft.JSInterop.Tests.DotNetDispatcherTest+TestDTO' cannot be converted to type 'Microsoft.JSInterop.Tests.DotNetDispatcherTest+TestDTO'.` As I recall the speculation was that this had something to do with contexts when parsing JSON.

PR #1353, opened 2019-04-05
Switch to flaky instead of hard skip

HaoK:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2027

### DIFF
--- a/src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
+++ b/src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Testing;
 using Microsoft.AspNetCore.Testing.xunit;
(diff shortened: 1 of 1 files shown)


HaoK:
Note this latest failure was on windows, so just marking all of these tests as flaky across the board

src/Configuration/test/Config.FunctionalTests/ConfigurationTests.cs
```diff
@@ -463,9 +464,8 @@ public void OnLoadErrorCanIgnoreErrors()
             Assert.NotNull(provider);
         }
 
-        [ConditionalFact]
-        [OSSkipCondition(OperatingSystems.Linux, SkipReason = "File watching is flaky on non windows.")]
-        [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "File watching is flaky on non windows.")]
+        [Fact]
+        [Flaky("File watching is flaky (particularly on non windows.", FlakyOn.All)]
```
anurse: This is supposed to be a link to a GitHub issue tracking re-enabling it, or at least describing the details. If the test is permanently flaky, we should consider how valuable it actually is :)

PR #1361, opened 2019-04-06
Dispose configuration in host (Fixes #938)

### LINKED ISSUE #938: Dispose configuration in generic host & host builder
This is a follow-up to #861 which covered the configuration part that made configuration providers and the configuration root disposable. What’s left now is the actual code in `Host` and `HostBuilder` (as well as `WebHost` and `WebHostBuilder` for backwards compatibility) that disposes the configuration.

I have tried to look for a good way to do this and I’m honestly struggling a bit with this. The problem is overall that `Host` (unlike `WebHost`) does not have a direct reference to the configuration. The configuration is [built](https://github.com/aspnet/Extensions/blob/f2ef1b55d65feac6ad4e3dd8b04daa21f464631b/src/Hosting/Hosting/src/HostBuilder.cs#L177-L178) by the `HostBuilder` and [added to the service collection there](https://github.com/aspnet/Extensions/blob/f2ef1b55d65feac6ad4e3dd8b04daa21f464631b/src/Hosting/Hosting/src/HostBuilder.cs#L186), and then the built service provider is only [passed to the `Host`](https://github.com/aspnet/Extensions/blob/f2ef1b55d65feac6ad4e3dd8b04daa21f464631b/src/Hosting/Hosting/src/Internal/Host.cs#L23-L24). The host itself does not access the configuration itself though. This leaves us with a few options:

1. Retrieve the configuration in the `Host` to be able to dispose it. So either `Services.GetService<IConfiguration>()` (in the constructor or directly in `Dispose` before actually disposing the service collection), or by taking a direct dependency on `IConfiguration` in the constructor. E.g.:

   ```
   public void Dispose()
   {
       (Services.GetService<IConfiguration>()> as IDisposable)?.Dispose();
       (Services as IDisposable)?.Dispose();
   }
   ```

   This feels a bit odd though since `Host` does not actually use the configuration and from its design it also does not actually care about whether there is a configuration object or not. It would be perfectly fine to create a host without the `HostBuilder`, and without preparing a configuration object. So it wouldn’t feel perfectly right for the host to check for a configuration root and dispose it, just in case if there is one.

2. Introduce a concept of “objects that should be disposed with the host”. In addition to registering the configuration as an `IConfiguration` with the service collection, the host builder could also register the same instance with some wrapper that marks it to get disposed with the host. The host could then retrieve all the registered objects and dispose them. This could look like this:

   ```
   // in HostBuilder
   services.AddSingleton(_appConfiguration);
   services.AddSingleton(new DisposeWithHost(_appConfiguration));

   // in Host.Dispose
   foreach (var disposable in Services.GetServices<DisposeWithHost>())
   {
       (disposable.Instance as IDisposable)?.Dispose();
   }
   ```

   This would have the added benefit over the first solution that this would not introduce a direct dependency to the configuration inside the host. And the same concept could also be used for other things that should be disposed (e.g. the file providers inside `HostingEnvironment`).

3. Introduce a way for DI to be able to register a singleton instance while transferring the ownership to the DI container. Right now, singletons will get disposed if they are created by the DI container, i.e. when they aren’t registered with an instance but rather a factory or an implementation type. Because instances aren’t *owned* by the DI container, it will not dispose them automatically (as per aspnet/DependencyInjection#462).

   If there was a way to tell the DI container to register those instances as singletons while taking ownership, then the DI container could simply dispose of them. So the resposibility would be moved away from the host and the host builder:

   ```
   services.AddSingletonOwnedByContainer(_appConfiguration);
   ```

   What already works as a workaround here is to register a factory that statically returns the instance, but that does feel a bit wrong:

   ```
   services.AddSingleton(_ => _appConfiguration);
   ```

   On the other hand, this would be very easy to implement and wouldn’t require additional changes with the host since the service provider is already disposed properly.

---

Regardless of this, there is also the open question on what to do with the builder configuration itself. The host builders’ configuration is [used last to initialize the app configuration](https://github.com/aspnet/Extensions/blob/f2ef1b55d65feac6ad4e3dd8b04daa21f464631b/src/Hosting/Hosting/src/HostBuilder.cs#L172) and after that it is never read from again. A host can also be built just once, so we *could* dispose the configuration at that point. Since there’s also no way to publicly access the configuration from the builder (unlike the `WebHostBuilder`), it would be safe to do so at this point.

Also, whatever we end up doing for the generic host and host builder, we also need to transfer over to the web host and web host builder in AspNetCore.

I have working prototypes for the first two options, as well as the workaround of the third option. I just need some feedback on what approach we would like to follow here. Of course, if you have another suggestion, I am more than open to that!

/cc @davidfowl

poke:
This fixes #938: When disposing the host, the underlying configuration will also be disposed. Access to the configuration is provided by explicitly depending on the `IConfiguration` in `Host`. Since the host is internal and constructed by the `HostBuilder` through DI, this change does not affect users.

In addition, the host configuration that is created and managed by the `HostBuilder` is disposed when the `Host` is being built. At that point, the previous host configuration is only used to construct the app configuration and then every reference to the host configuration is replaced by the app configuration (e.g. in the `HostContext`). Since the host configuration is then never accessed again (the host can also not be built more than once) and since it also remains private to the host builder, the configuration can be safely disposed at that point.

This change combined with #928 will ultimately ensure that configuration providers will be properly disposed when the host gets disposed (or built).

A follow-up PR to AspNetCore will be made in a moment that backports this functionality to the legacy `WebHost` and `WebHostBuilder`.

/cc @davidfowl

### DIFF
--- a/src/Configuration/Config/ref/Microsoft.Extensions.Configuration.netstandard2.0.cs
+++ b/src/Configuration/Config/ref/Microsoft.Extensions.Configuration.netstandard2.0.cs
@@ -6,6 +6,7 @@ namespace Microsoft.Extensions.Configuration
     public static partial class ChainedBuilderExtensions
     {
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddConfiguration(this Microsoft.Extensions.Configuration.IConfigurationBuilder configurationBuilder, Microsoft.Extensions.Configuration.IConfiguration config) { throw null; }
+        public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddConfiguration(this Microsoft.Extensions.Configuration.IConfigurationBuilder configurationBuilder, Microsoft.Extensions.Configuration.IConfiguration config, bool shouldDisposeConfiguration) { throw null; }
     }
     public partial class ChainedConfigurationProvider : Microsoft.Extensions.Configuration.IConfigurationProvider, System.IDisposable
     {
@@ -21,6 +22,7 @@ namespace Microsoft.Extensions.Configuration
     {
         public ChainedConfigurationSource() { }
         public Microsoft.Extensions.Configuration.IConfiguration Configuration { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool ShouldDisposeConfiguration { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.Extensions.Configuration.IConfigurationProvider Build(Microsoft.Extensions.Configuration.IConfigurationBuilder builder) { throw null; }
     }
     public partial class ConfigurationBuilder : Microsoft.Extensions.Configuration.IConfigurationBuilder
--- a/src/Configuration/Config/ref/Microsoft.Extensions.Configuration.netstandard2.0.cs
+++ b/src/Configuration/Config/ref/Microsoft.Extensions.Configuration.netstandard2.0.cs
@@ -7,9 +7,10 @@ namespace Microsoft.Extensions.Configuration
     {
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddConfiguration(this Microsoft.Extensions.Configuration.IConfigurationBuilder configurationBuilder, Microsoft.Extensions.Configuration.IConfiguration config) { throw null; }
     }
-    public partial class ChainedConfigurationProvider : Microsoft.Extensions.Configuration.IConfigurationProvider
+    public partial class ChainedConfigurationProvider : Microsoft.Extensions.Configuration.IConfigurationProvider, System.IDisposable
     {
         public ChainedConfigurationProvider(Microsoft.Extensions.Configuration.ChainedConfigurationSource source) { }
+        public void Dispose() { }
         public System.Collections.Generic.IEnumerable<string> GetChildKeys(System.Collections.Generic.IEnumerable<string> earlierKeys, string parentPath) { throw null; }
         public Microsoft.Extensions.Primitives.IChangeToken GetReloadToken() { throw null; }
         public void Load() { }
--- a/src/Hosting/Hosting/src/HostBuilder.cs
+++ b/src/Hosting/Hosting/src/HostBuilder.cs
@@ -197,7 +197,8 @@ namespace Microsoft.Extensions.Hosting
 #pragma warning restore CS0618 // Type or member is obsolete
             services.AddSingleton<IHostEnvironment>(_hostingEnvironment);
             services.AddSingleton(_hostBuilderContext);
-            services.AddSingleton(_appConfiguration);
+            // register configuration as factory to make it dispose with the service provider
+            services.AddSingleton(_ => _appConfiguration);
 #pragma warning disable CS0618 // Type or member is obsolete
             services.AddSingleton<IApplicationLifetime>(s => (IApplicationLifetime)s.GetService<IHostApplicationLifetime>());
 #pragma warning restore CS0618 // Type or member is obsolete
@@ -225,6 +226,10 @@ namespace Microsoft.Extensions.Hosting
             {
                 throw new InvalidOperationException($"The IServiceProviderFactory returned a null IServiceProvider.");
             }
+
+            // resolve configuration explicitly once to mark it as resolved within the
+            // service provider, ensuring it will be properly disposed with the provider
+            _ = _appServices.GetService<IConfiguration>();
         }
     }
 }
--- a/src/Hosting/Hosting/src/HostBuilder.cs
+++ b/src/Hosting/Hosting/src/HostBuilder.cs
@@ -179,7 +179,7 @@ namespace Microsoft.Extensions.Hosting
         {
             var configBuilder = new ConfigurationBuilder()
                 .SetBasePath(_hostingEnvironment.ContentRootPath)
-                .AddConfiguration(_hostConfiguration);
+                .AddConfiguration(_hostConfiguration, shouldDisposeConfiguration: true);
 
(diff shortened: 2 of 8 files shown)


src/Hosting/Hosting/src/HostBuilder.cs
```diff
@@ -187,6 +187,9 @@ private void BuildAppConfiguration()
             }
             _appConfiguration = configBuilder.Build();
             _hostBuilderContext.Configuration = _appConfiguration;
+
+            // dispose host configuration since it is no longer accessible
```
Tratcher: Too soon, the host config is chained into the app config. You have to wait until host disposal.

poke: Oh. I was under the impression that `configurationBuilder.AddConfiguration(IConfiguration)` would actually _copy_ over the values from the source configuration. But it does actually link them. TIL!

Would it work for you to make the `ChainedConfigurationProvider` disposable and dispose of the chained configuration then? That would actually make this much nicer too.

poke: I went with that idea now and made it configurable to avoid breaking any existing code that would use the same configuration multiple times.

In the framework, the only places `AddConfiguration` is actually being used outside of the host builders are adding configuration sections which aren’t disposable and won’t trigger this. So maybe making the dispose configurable isn’t actually required.

src/Hosting/Hosting/src/Internal/Host.cs
```diff
@@ -17,16 +18,18 @@ internal class Host : IHost
...
+        private readonly IConfiguration _configuration;
         private readonly ILogger<Host> _logger;
         private readonly IHostLifetime _hostLifetime;
         private readonly ApplicationLifetime _applicationLifetime;
         private readonly HostOptions _options;
         private IEnumerable<IHostedService> _hostedServices;
 
-        public Host(IServiceProvider services, IHostApplicationLifetime applicationLifetime, ILogger<Host> logger,
-            IHostLifetime hostLifetime, IOptions<HostOptions> options)
+        public Host(IServiceProvider services, IConfiguration configuration, IHostApplicationLifetime applicationLifetime,
```
Tratcher: It smells that we'd need to dispose a service that was provided to us by DI. I think we can get the container to dispose it for us.
https://github.com/aspnet/Extensions/blob/b28d96763fdf2750f31936641313f3af24def0fc/src/Hosting/Hosting/src/HostBuilder.cs#L200
When registering the config, if you register it as a factory that returns the instance then the container should dispose the instance when it's done.

src/Configuration/Config/ref/Microsoft.Extensions.Configuration.netstandard2.0.cs
```diff
@@ -6,10 +6,12 @@ namespace Microsoft.Extensions.Configuration
     public static partial class ChainedBuilderExtensions
     {
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddConfiguration(this Microsoft.Extensions.Configuration.IConfigurationBuilder configurationBuilder, Microsoft.Extensions.Configuration.IConfiguration config) { throw null; }
+        public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddConfiguration(this Microsoft.Extensions.Configuration.IConfigurationBuilder configurationBuilder, Microsoft.Extensions.Configuration.IConfiguration config, bool shouldDisposeConfiguration) { throw null; }
```
Tratcher: @HaoK thoughts?

src/Hosting/Hosting/src/Internal/Host.cs
poke: I offered that before but @davidfowl [chose to inject it explicitly instead](https://github.com/aspnet/Extensions/issues/938#issuecomment-456709911).

Notet hat the `Host` also disposes the service provider (which is also provided by DI), so it’s already doing that in a way.

poke: (But I still do think that disposing it implicitly through the service provider would be a lot cleaner. Since the configuration is a singleton, the factory action would only be called once, right? So this wouldn’t actually have much of a drawback except for being a bit cryptic.)

Tratcher: Ah, I'll talk to him.

Tratcher: I talked to @davidfowl offline, let's try the factory approach. I don't think he read your original proposal all the way through.

poke: Alright, that works for me, thanks for clearing that up! (Note to self: Try to write shorter issues 😅)

Tratcher: Make sure you include a comment that the factory is used as a means of requesting disposal.

poke: Updated the PR. Had to keep the dependency on `IConfiguration` within `Host` though to make sure that there’s something that resolves the configuration. Otherwise it wouldn’t have been picked up reliably by the service provider.

Tratcher: How about resolving IConfiguration here instead?
https://github.com/aspnet/Extensions/blob/b28d96763fdf2750f31936641313f3af24def0fc/src/Hosting/Hosting/src/HostBuilder.cs#L222-L227

poke: Works for me – good idea to avoid having to rely on the host for that! 👍

src/Configuration/Config/ref/Microsoft.Extensions.Configuration.netstandard2.0.cs
HaoK: Seems fine, pattern has been to just add overloads like this that set source properties

Tratcher:
Thanks

davidfowl:
👏

poke:
That was quicker than expected now – thank you so much for the feedback! 😊

Seabizkit:
Not sure where the latest regarding this is, but wouldn't it make sense to add Dispose on the interface?

Issue #1337, opened 2019-04-01
Microsoft.Extensions.Hosting.WindowsServices - detect running as Windows Service?


drmcclelland:
Within my console application's Main() method, I am trying to detect whether the app is running as a Windows Service vs running as a console app.

When my code is running as a Windows Service, I want to do certain things - but if I'm just running the EXE from the command line I'd like it to do other things.

I have seen this approach used, but is requires an extra command line parameter:  https://github.com/stevejgordon/IHostedServiceAsAService/blob/de55562bb1e424973660892e71b67a6c8200bfc3/IHostedServiceAsAService/Program.cs#L11-L29

And I saw the IsWindowsService() static method in <code>Microsoft.Extensions.Hosting.WindowsServices</code>, but it's private:  https://github.com/aspnet/Extensions/blob/898f0947fa4271d08e88daea503be6fd8ffb3c4f/src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs#L54-L67

glennc:
@Tratcher Does it seem reasonable to make this an extension method on something like `IHostEnvironment` rather than its current home?

Tratcher:
No reason to make it an extension method, it's completely static. We could move it to a new type like ServiceHelpers.

drmcclelland:
Just tried to contribute my first PR #1369 to help out   :-)

PR #1369, opened 2019-04-08
Make IsWindowsService() visible publicly

drmcclelland:
Make IsWindowsService() visible publicly by moving it a new Microsoft.Extensions.Hosting.WindowsServices.ServiceHelpers type.

Addresses #1337

### DIFF
--- a/src/Hosting/WindowsServices/src/ServiceHelpers.cs
+++ b/src/Hosting/WindowsServices/src/ServiceHelpers.cs
@@ -6,8 +6,15 @@ using System.Runtime.InteropServices;
 
 namespace Microsoft.Extensions.Hosting.WindowsServices
 {
+    /// <summary>
+    /// Helper methods for Windows Services.
+    /// </summary>
     public static class ServiceHelpers
     {
+        /// <summary>
+        /// Check if the current process is hosted as a Windows Service. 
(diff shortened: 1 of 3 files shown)


Tratcher:
Housekeeping, you need to update the ref assemblies to add the new public API:
`Generated code is not up to date. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)`

src/Hosting/WindowsServices/src/ServiceHelpers.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Extensions.Hosting.WindowsServices
+{
+    public class ServiceHelpers
```
campersau: Should this be a `static` class?

drmcclelland: Thanks, updated code in PR.

Tratcher:
Thanks

PR #1375, opened 2019-04-09
Add Repeat attribute

jkotalik:
When diagnosing/fixing flaky tests, it would be great to easily be able to run a test multiple times to effectively "stress test" that the fix is working. 

The Deflake attribute can be applied to an assembly, class, or method to run that specific tests multiple times. This requires that a test uses LoggedTest (which is something we should probably apply to all of our tests anyways).


src/Logging/Logging.Testing/src/Xunit/DeflakeAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.Extensions.Logging.Testing
+{
+    /// <summary>
+    /// Runs a test multiple times to stress flaky tests that are believed to be fixed.
+    /// This can be used on an assembly, class, or method name.
+    /// Requires using <see cref="LoggedTest"/> to run.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false)]
+    public class DeflakeAttribute : Attribute
```
anurse: Cute name, but let's be a little more descriptive :). Perhaps `RepeatAttribute`?

jkotalik: We already have a `RetryAttribute` as a name, which will retry a test until it succeeds. I don't think `RepeatAttribute` is a great name as it's hard to know the difference between Repeat and Retry. But I don't have anything better...

anurse: I think Repeat is sufficient to me. The difference between `Repeat` and `Retry` in my mind is clear. `Repeat` re-runs the test no matter what. `Retry` indicates there was a first "try" and it failed.

src/Logging/Logging.Testing/src/Xunit/LoggedTestInvoker.cs
```diff
@@ -51,12 +51,13 @@ protected override object CreateTestClass()
             if (testClass is LoggedTestBase loggedTestBase)
             {
                 // Used for testing
+
```
JunTaoLuo: nit:            space

src/Logging/Logging.Testing/src/Xunit/DeflakeAttribute.cs
jkotalik: 🆙 📅

src/Logging/Logging.Testing/src/Xunit/LoggedTestRunner.cs
```diff
@@ -97,6 +103,40 @@ private async Task<decimal> InvokeTestMethodAsync(ExceptionAggregator aggregator
             return totalTime;
         }
 
+        private async Task<decimal> RunRepeatTestInvoker(ExceptionAggregator aggregator, ITestOutputHelper output, bool collectDump, RepeatAttribute repeatAttribute)
+        {
+            var repeatContext = new RetryContext()
```
JunTaoLuo: nit: Maybe use a different context? `RepeatContext` instead of `RetryContext`? Even though they would be similar let's not cause confusion.

src/Logging/Logging.Testing/src/Xunit/LoggedTestRunner.cs
```diff
@@ -54,6 +54,12 @@ private async Task<decimal> InvokeTestMethodAsync(ExceptionAggregator aggregator
 
             if (!typeof(LoggedTestBase).IsAssignableFrom(TestClass) || retryAttribute == null)
             {
+                var repeatAttribute = GetRepeatAttribute(TestMethod);
```
JunTaoLuo: Hmm should we explicitly prevent the case when Repeat and Retry are both added? Otherwise the failures might be hidden behind retry.

src/Logging/Logging.Testing/src/Xunit/LoggedTestInvoker.cs
```diff
@@ -51,12 +51,13 @@ protected override object CreateTestClass()
...
                 loggedTestBase.RetryContext = _retryContext;
 
                 if (_retryContext != null)
                 {
                     // Log retry attempt as warning
-                    if (_retryContext.CurrentIteration > 0)
+                    if (_retryContext.CurrentIteration > 0 && !string.IsNullOrEmpty(_retryContext.Reason))
```
JunTaoLuo: Another reason to use a different context. The reason may not be set when using the Retry attribute but we'd still want to log a warning.

jkotalik:
It would be a breaking change. Only reason I didn't remove it.

JunTaoLuo:
We're the only ones using it. Just make a pr to aspnetcore to remove all usage of it first and it'll be fine.

jkotalik:
I could also just do it in the update dependencies pr that would eventually come from extensions.

jkotalik:
@JunTaoLuo I removed the retry attribute.

src/Logging/Logging.Testing/src/Xunit/LoggedTestRunner.cs
```diff
@@ -49,82 +49,65 @@ protected override Task<decimal> InvokeTestMethodAsync(ExceptionAggregator aggre
 
         private async Task<decimal> InvokeTestMethodAsync(ExceptionAggregator aggregator, ITestOutputHelper output)
         {
-            var retryAttribute = GetRetryAttribute(TestMethod);
             var collectDump = TestMethod.GetCustomAttribute<CollectDumpAttribute>() != null;
-
-            if (!typeof(LoggedTestBase).IsAssignableFrom(TestClass) || retryAttribute == null)
+            var repeatAttribute = GetRepeatAttribute(TestMethod);
+            if (!typeof(LoggedTestBase).IsAssignableFrom(TestClass) || repeatAttribute == null)
```
JunTaoLuo: nit: new line before if

jkotalik:
I'm going to merge this. Whenever there is an Update Dependency PR to ASP.NET Core, I'll remove any uses of RetryTest (I only saw 2 in aspnetcore).

PR #1382, opened 2019-04-10
Update CODEOWNERS

natemcmaster:
FYI - @aspnet/build


natemcmaster:
Updated to remove @pakrym from DI and Logging

PR #1400, opened 2019-04-11
Expose property on CacheOptions to allow for access token authentication

johnayoub:
Summary of the changes
 - Add property to cache options to allow for access token authentication

### DIFF
--- a/src/Caching/SqlServer/src/DatabaseOperations.cs
+++ b/src/Caching/SqlServer/src/DatabaseOperations.cs
@@ -30,14 +30,17 @@ namespace Microsoft.Extensions.Caching.SqlServer
             "Connection string: {2}";
 
         public DatabaseOperations(
-            string connectionString, string schemaName, string tableName, ISystemClock systemClock)
+            string connectionString, string schemaName, string tableName, Func<string> accessTokenFunc, ISystemClock systemClock)
         {
             ConnectionString = connectionString;
             SchemaName = schemaName;
             TableName = tableName;
+            AccessTokenFunc = accessTokenFunc;
             SystemClock = systemClock;
             SqlQueries = new SqlQueries(schemaName, tableName);
         }
+
+        protected Func<string> AccessTokenFunc { get; }
 
         protected SqlQueries SqlQueries { get; }
 
@@ -49,9 +52,12 @@ namespace Microsoft.Extensions.Caching.SqlServer
 
         protected ISystemClock SystemClock { get; }
 
+        protected SqlConnection CreateSqlConnection() => new SqlConnection(ConnectionString)
+            {AccessToken = AccessTokenFunc?.Invoke()};
+
         public void DeleteCacheItem(string key)
         {
-            using (var connection = new SqlConnection(ConnectionString))
+            using (var connection = CreateSqlConnection())
             using (var command = new SqlCommand(SqlQueries.DeleteCacheItem, connection))
             {
                 command.Parameters.AddCacheItemId(key);
@@ -66,7 +72,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
         {
             token.ThrowIfCancellationRequested();
 
-            using (var connection = new SqlConnection(ConnectionString))
+            using (var connection = CreateSqlConnection())
             using (var command = new SqlCommand(SqlQueries.DeleteCacheItem, connection))
             {
                 command.Parameters.AddCacheItemId(key);
@@ -98,14 +104,14 @@ namespace Microsoft.Extensions.Caching.SqlServer
         {
             token.ThrowIfCancellationRequested();
 
-            await GetCacheItemAsync(key, includeValue: false, token:token);
+            await GetCacheItemAsync(key, includeValue: false, token: token);
         }
 
         public virtual void DeleteExpiredCacheItems()
         {
             var utcNow = SystemClock.UtcNow;
 
-            using (var connection = new SqlConnection(ConnectionString))
+            using (var connection = CreateSqlConnection())
             using (var command = new SqlCommand(SqlQueries.DeleteExpiredCacheItems, connection))
             {
                 command.Parameters.AddWithValue("UtcNow", SqlDbType.DateTimeOffset, utcNow);
@@ -123,7 +129,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
             var absoluteExpiration = GetAbsoluteExpiration(utcNow, options);
             ValidateOptions(options.SlidingExpiration, absoluteExpiration);
 
-            using (var connection = new SqlConnection(ConnectionString))
+            using (var connection = CreateSqlConnection())
             using (var upsertCommand = new SqlCommand(SqlQueries.SetCacheItem, connection))
             {
                 upsertCommand.Parameters
@@ -163,8 +169,8 @@ namespace Microsoft.Extensions.Caching.SqlServer
             var absoluteExpiration = GetAbsoluteExpiration(utcNow, options);
             ValidateOptions(options.SlidingExpiration, absoluteExpiration);
 
-            using (var connection = new SqlConnection(ConnectionString))
-            using(var upsertCommand = new SqlCommand(SqlQueries.SetCacheItem, connection))
+            using (var connection = CreateSqlConnection())
+            using (var upsertCommand = new SqlCommand(SqlQueries.SetCacheItem, connection))
             {
                 upsertCommand.Parameters
                     .AddCacheItemId(key)
@@ -212,7 +218,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
             TimeSpan? slidingExpiration = null;
             DateTimeOffset? absoluteExpiration = null;
             DateTimeOffset expirationTime;
-            using (var connection = new SqlConnection(ConnectionString))
+            using (var connection = CreateSqlConnection())
             using (var command = new SqlCommand(query, connection))
             {
                 command.Parameters
@@ -277,7 +283,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
             TimeSpan? slidingExpiration = null;
             DateTimeOffset? absoluteExpiration = null;
             DateTimeOffset expirationTime;
-            using (var connection = new SqlConnection(ConnectionString))
+            using (var connection = CreateSqlConnection())
             using (var command = new SqlCommand(query, connection))
             {
                 command.Parameters
--- a/src/Caching/SqlServer/src/DatabaseOperations.cs
+++ b/src/Caching/SqlServer/src/DatabaseOperations.cs
@@ -52,8 +52,13 @@ namespace Microsoft.Extensions.Caching.SqlServer
 
         protected ISystemClock SystemClock { get; }
 
-        protected SqlConnection CreateSqlConnection() => new SqlConnection(ConnectionString)
-            {AccessToken = AccessTokenFunc?.Invoke()};
+        protected SqlConnection CreateSqlConnection()
+        {
+            return new SqlConnection(ConnectionString)
+            {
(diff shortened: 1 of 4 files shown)


src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -49,9 +52,12 @@ internal class DatabaseOperations : IDatabaseOperations
 
         protected ISystemClock SystemClock { get; }
 
+        protected SqlConnection CreateSqlConnection() => new SqlConnection(ConnectionString)
+            {AccessToken = AccessTokenFunc?.Invoke()};
```
anurse: Generally we'd use a standard method body for this kind of method. Unless it's a really simple one-liner we tend to stick with standard methods instead of expression bodies.

src/Caching/SqlServer/src/SqlServerCacheOptions.cs
```diff
@@ -43,6 +43,11 @@ public class SqlServerCacheOptions : IOptions<SqlServerCacheOptions>
         /// </summary>
         public TimeSpan DefaultSlidingExpiration { get; set; } = TimeSpan.FromMinutes(20);
 
+        /// <summary>
+        /// A function, if provided, should return an access token for authentication
+        /// </summary>
+        public Func<string> AccessTokenFunc { get; set; }
```
anurse: Is there a reason this needs to be a function? Is the token expected to expire during while the application is running?

src/Caching/SqlServer/src/DatabaseOperations.cs
johnayoub: Made the change to adhere to the coding guidelines

src/Caching/SqlServer/src/SqlServerCacheOptions.cs
johnayoub: Yes that's a possibility. The tokens issued by Azure for managed identity expire after 1 hour.

anurse:
```suggestion
        public Func<string> AccessTokenProvider { get; set; }
```

This name aligns with what we use in other places. Also, I think it's **good** that this isn't async. We may need to remind users that they will have to fetch the access token asynchronously in a background component and swap the new token into a singleton value of some kind since they can't dynamically fetch an access token in the handler (because it's synchronous).

src/Caching/SqlServer/src/SqlServerCacheOptions.cs
```diff
@@ -43,6 +43,11 @@ public class SqlServerCacheOptions : IOptions<SqlServerCacheOptions>
         /// </summary>
         public TimeSpan DefaultSlidingExpiration { get; set; } = TimeSpan.FromMinutes(20);
 
+        /// <summary>
+        /// A function, if provided, should return an access token for authentication
```
anurse:
```suggestion
        /// Gets or sets an access token provider that will be called to return a token for each SQL connection.
```

Let's use similar language to https://github.com/aspnet/AspNetCore/blob/b1ca2c1155da3920f0df5108b9fedbe82efaa11c/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs#L87

src/Caching/SqlServer/src/MonoDatabaseOperations.cs
```diff
@@ -14,8 +14,8 @@ namespace Microsoft.Extensions.Caching.SqlServer
     internal class MonoDatabaseOperations : DatabaseOperations
     {
         public MonoDatabaseOperations(
-            string connectionString, string schemaName, string tableName, ISystemClock systemClock)
-            : base(connectionString, schemaName, tableName, systemClock)
+            string connectionString, string schemaName, string tableName, Func<string> accessTokenFunc, ISystemClock systemClock)
+            : base(connectionString, schemaName, tableName, accessTokenFunc, systemClock)
```
anurse:
```suggestion
            : base(connectionString, schemaName, tableName, accessTokenProvider, systemClock)
```

src/Caching/SqlServer/src/MonoDatabaseOperations.cs
```diff
@@ -14,8 +14,8 @@ namespace Microsoft.Extensions.Caching.SqlServer
     internal class MonoDatabaseOperations : DatabaseOperations
     {
         public MonoDatabaseOperations(
-            string connectionString, string schemaName, string tableName, ISystemClock systemClock)
-            : base(connectionString, schemaName, tableName, systemClock)
+            string connectionString, string schemaName, string tableName, Func<string> accessTokenFunc, ISystemClock systemClock)
```
anurse:
```suggestion
            string connectionString, string schemaName, string tableName, Func<string> accessTokenProvider, ISystemClock systemClock)
```

src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -30,15 +30,18 @@ internal class DatabaseOperations : IDatabaseOperations
             "Connection string: {2}";
 
         public DatabaseOperations(
-            string connectionString, string schemaName, string tableName, ISystemClock systemClock)
+            string connectionString, string schemaName, string tableName, Func<string> accessTokenFunc, ISystemClock systemClock)
```
anurse:
```suggestion
            string connectionString, string schemaName, string tableName, Func<string> accessTokenProvider, ISystemClock systemClock)
```

src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -30,15 +30,18 @@ internal class DatabaseOperations : IDatabaseOperations
...
         {
             ConnectionString = connectionString;
             SchemaName = schemaName;
             TableName = tableName;
+            AccessTokenFunc = accessTokenFunc;
```
anurse:
```suggestion
            AccessTokenProvider = accessTokenProvider;
```

src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -30,15 +30,18 @@ internal class DatabaseOperations : IDatabaseOperations
...
             SystemClock = systemClock;
             SqlQueries = new SqlQueries(schemaName, tableName);
         }
 
+        protected Func<string> AccessTokenFunc { get; }
```
anurse:
```suggestion
        protected Func<string> AccessTokenProvider { get; }
```

src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -49,9 +52,17 @@ internal class DatabaseOperations : IDatabaseOperations
 
         protected ISystemClock SystemClock { get; }
 
+        protected SqlConnection CreateSqlConnection()
+        {
+            return new SqlConnection(ConnectionString)
+            {
+                AccessToken = AccessTokenFunc?.Invoke()
```
anurse:
```suggestion
                AccessToken = AccessTokenProvider?.Invoke()
```

src/Caching/SqlServer/src/SqlServerCache.cs
```diff
@@ -74,6 +74,7 @@ public SqlServerCache(IOptions<SqlServerCacheOptions> options)
                     cacheOptions.ConnectionString,
                     cacheOptions.SchemaName,
                     cacheOptions.TableName,
+                    cacheOptions.AccessTokenFunc,
```
anurse:
```suggestion
                    cacheOptions.AccessTokenProvider,
```

src/Caching/SqlServer/src/SqlServerCache.cs
```diff
@@ -82,6 +83,7 @@ public SqlServerCache(IOptions<SqlServerCacheOptions> options)
...
```
anurse:
```suggestion
                    cacheOptions.AccessTokenProvider,
```

src/Caching/SqlServer/src/SqlServerCacheOptions.cs
johnayoub: Applied the name change as you suggested. This is the implementation that I had in mind for the AccessTokenProvider:  
```
AccessTokenProvider = () => {
  var azureServiceTokenProvider = new AzureServiceTokenProvider();

  return azureServiceTokenProvider.GetAccessTokenAsync("https://database.windows.net/").GetAwaiter().GetResult();
}
```

anurse: I was afraid you'd say that :). Calling `.GetResult()` on an awaiter or `.Wait`/`.Result` on a task will tank performance and make your app very difficult to scale. Blocking waits on Tasks is always a **bad idea**.

The scalable way to do this would be to have a background service call [`GetAuthenticationResultAsync`](https://github.com/Azure/azure-sdk-for-net/blob/0bf39aa206dbbe867619f4a4458d698d9e058e83/src/SdkCommon/AppAuthentication/Azure.Services.AppAuthentication/AzureServiceTokenProvider.cs#L264), store the token somewhere the callback can access and set a timer that will fire when the token is near expiry. Then when the timer fires, get a new token and atomically update the stored access token so that the next SQL connection uses it.

I think we're going to need to look in to this further to find the right design to make sure the SQL cache can use Azure Managed Service Identity properly.

johnayoub: hahaha :) this is the solution I have seen floating around for enabling Managed Identity for use with Entity Framework in the community. Internally the `AzureServiceTokenProvider` caches the token and refreshes if it has expired or is about to expire (5 minutes) so I figured I could reply on it since they're already taking care of multi-thread access. But I see your point about the potential performance hit if implemented as I suggested.

johnayoub: This is the implementation of the `AzureServiceTokenProvider` [Source Code](https://github.com/Azure/azure-sdk-for-net/blob/master/src/SdkCommon/AppAuthentication/Azure.Services.AppAuthentication/AzureServiceTokenProvider.cs)

Let me know your thoughts if the `Awaiter` in this case will be okay or just simply not a good idea :)

anurse: It's still universally a Bad Idea:tm: to call `.GetAwaiter().GetResult()`. Even if the underlying awaiter is "rarely" async. It's just a matter of good practice.

anurse:
Pinging some caching-adjacent people (and @davidfowl) to see if they have any thoughts: @JunTaoLuo @Tratcher.

Since Azure Managed Service Identity can't be encoded in the Connection String and requires updating a token multiple times throughout the application, I think a `Func` does make sense here. It does make me wonder if we should do a `Func<SqlConnection>` instead and just allow you to completely take over connection creation (to be a little future-proof).

src/Caching/SqlServer/src/SqlServerCacheOptions.cs
anurse: > this is the solution I have seen floating around for enabling Managed Identity for use with Entity Framework in the community

This is unfortunate. cc @ajcvickers for visibility. Might be good for the EF team to try to get out some "official" docs or guidance for using EF with [Azure Managed Service Identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/tutorial-windows-vm-access-sql) if there aren't already.

anurse:
Closing this as the PR has gotten stale. I think we'd still like to do this @johnayoub , but we're locking down public API surface for 3.0. We've got https://github.com/aspnet/Extensions/issues/1142 on the backlog to look at improving this, and would still be interested in a PR of some kind. Unfortunately we don't have a branch to receive post-3.0 changes in yet. I'd expect there will be an announcement at https://github.com/aspnet/Announcements when we do.

PR #1405, opened 2019-04-12
[2.2] Fix for assembly version not matching package version

natemcmaster:
Revert exclusions added in https://github.com/aspnet/Extensions/pull/1259 to NuGet package verifier which suppresses errors that should not have been ignored.

For future reference, assembly version mismatch errors should be taken seriously and not ignored. If we get assembly versions wrong, we can end up with all sorts of nasty problems with binding redirects, forced downgrades, runtime errors, shared framework mismatches, etc.

Required to unblock @BrennanConroy's work.

## Description

Normally, ASP.NET Core packages have the assembly version match the NuGet package version (at least the first three parts, x.y.z). Each patch of should have incremented the assembly versions but didn't. For example, the 2.2.4 release of Microsoft.Extensions.Configuration.KeyPerFile contains an an assembly with version `2.2.0.0`, not `2.2.4.0`.

## Customer impact

This will prevent us from having problems in the future if we have to patch any assemblies for a 2nd time.

## Regression?

Yes.

## Risk

Low. This restores behavior we had in 2.1 which somehow went missing during source code reorganization. It should have been caught earlier, but devs started suppressing the errors instead of fixing them.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -18,12 +18,15 @@
   <Import Project="eng\Baseline.Designer.props" />
 
   <PropertyGroup Condition=" '$(IsPackable)' != 'false' AND '$(IsServicingBuild)' == 'true' ">
-    <IsPackable>$(PackagesInPatch.Contains(' $(PackageId);'))</IsPackable>
+    <IsPackageInThisPatch Condition="'$(IsPackageInThisPatch)' == ''">$(PackagesInPatch.Contains(' $(PackageId);'))</IsPackageInThisPatch>
+    <!-- Suppress creation of .nupkg for servicing builds. -->
+    <IsPackable Condition=" '$(IsPackageInThisPatch)' != 'true' ">false</IsPackable>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


Directory.Build.targets
```diff
@@ -18,12 +18,15 @@
   <Import Project="eng\Baseline.Designer.props" />
 
   <PropertyGroup Condition=" '$(IsPackable)' != 'false' AND '$(IsServicingBuild)' == 'true' ">
-    <IsPackable>$(PackagesInPatch.Contains(' $(PackageId);'))</IsPackable>
+    <IsPackageInThisPatch Condition="'$(IsPackageInThisPatch)' == ''">$(PackagesInPatch.Contains(' $(PackageId);'))</IsPackageInThisPatch>
+    <!-- Suppress creation of .nupkg for servicing builds. -->
+    <IsPackable Condition=" '$(IsPackageInThisPatch)' != 'true' ">false</IsPackable>
```
dougbu: These two lines look like a nice optimization. Could they increase the risk of this change?

natemcmaster: Unlikely. I did this to improve readability, and also to align this section of code with whats in aspnet/AspNetCore. But if you spot a problem with it, let me know.

vivmishra:
Approved for 2.2.5

Issue #1348, opened 2019-04-04
Add doc comments to Host.CreateDefaultBuilder method


DamianEdwards:
It's used in the templates and currently has no doc comments. Given its prominence in all app types we should prioritize getting the docs added.

https://github.com/aspnet/Extensions/blob/master/src/Hosting/Hosting/src/Host.cs#L13

@anurse @glennc @Tratcher

PR #1409, opened 2019-04-12
Add doc comments to Host.CreateDefaultBuilder

Tratcher:
#1348 largely copied from it's predecessor WebHost.[CreateDefaultBuilder](https://github.com/aspnet/AspNetCore/blob/717fa3916716d881bcff38f2a0e30b0676f7dcbe/src/DefaultBuilder/src/WebHost.cs#L133-L149).

@DamianEdwards [Doc-a-thon]

### DIFF
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -8,11 +8,49 @@ using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Extensions.Hosting
 {
+    /// <summary>
+    /// Provides convenience methods for creating instances of <see cref="IHostBuilder"/> with pre-configured defaults.
+    /// </summary>
     public static class Host
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HostBuilder"/> class with pre-configured defaults.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Hosting/Hosting/src/Host.cs
```diff
@@ -8,11 +8,45 @@
...
+    /// <summary>
+    /// Provides convenience methods for creating instances of <see cref="IHostBuilder"/> with pre-configured defaults.
+    /// </summary>
     public static class Host
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HostBuilder"/> class with pre-configured defaults.
+        /// </summary>
+        /// <remarks>
+        ///   The following defaults are applied to the returned <see cref="HostBuilder"/>:
```
anurse: This totally doesn't format right in the docs :(. See the existing [`WebHost.CreateDefaultBuilder`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder?view=aspnetcore-2.2) docs...

If we want a list, we should use [`<list>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/list)

PR #1413, opened 2019-04-13
Remove target invocation exceptions

davidfowl:
- Use BindingFlags.DoNotWrapExceptions to invoke constructors

### DIFF
--- a/src/Shared/src/ActivatorUtilities/ActivatorUtilities.cs
+++ b/src/Shared/src/ActivatorUtilities/ActivatorUtilities.cs
@@ -403,16 +403,20 @@ namespace Microsoft.Extensions.Internal
                     }
                 }
 
+#if NETCOREAPP3_0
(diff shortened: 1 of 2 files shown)


src/Shared/src/ActivatorUtilities/ActivatorUtilities.cs
```diff
@@ -403,16 +403,20 @@ public object CreateInstance(IServiceProvider provider)
...
 
+#if NETCOREAPP3_0
+                return _constructor.Invoke(BindingFlags.DoNotWrapExceptions, binder: null, parameters: _parameterValues, culture: null);
+#else
                 try
                 {
                     return _constructor.Invoke(_parameterValues);
                 }
-                catch (TargetInvocationException ex)
+                catch (TargetInvocationException ex) when (ex.InnerException != null)
```
halter73: Is TargetInvocationException.InnerException ever in fact null?

davidfowl: No. I don't think so.

PR #1422, opened 2019-04-15
Delay logging scope creation

benaadams:
It would be good if there was something between logging totally off and logging totally allocately.

Scope creation is a pretty significant allocator, even if scopes are off and logging is at a minimum level
```
dotnet new web
```
Allocates as the following

image

Usage for this change

Change
```csharp
return logger.BeginScope(new HostingLogScope(httpContext, activityId));
```
to
```csharp
return logger.BeginScope(
    (state) => new HostingLogScope(state.httpContext, state.activityId), 
    (httpContext, activityId));
```
To drop the allocations when `CaptureScopes == false`

/cc @davidfowl

### DIFF
--- a/src/Logging/Logging.Abstractions/src/IScopeFuncLogger.cs
+++ b/src/Logging/Logging.Abstractions/src/IScopeFuncLogger.cs
@@ -1,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    public interface IScopeFuncLogger
+    {
+        IDisposable BeginScope<TState, TScopeState>(Func<TState, TScopeState> scopeCreator, in TState state);
+    }
+}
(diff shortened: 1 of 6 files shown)


src/Logging/Logging.Abstractions/src/IScopeFuncLogger.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    public interface IScopeFuncLogger
+    {
+        IDisposable BeginScope<TState, TScopeState>(Func<TState, TScopeState> scopeCreator, in TState state);
```
benaadams: Could be a default interface method on ILogger; however as its `netstandard2.0` I guess that wouldn't work?

#ifdef an interface? 🤔

davidfowl:
I'm not sure this is worth it to avoid the allocation in the case when `CaptureScopes` is false. This is rarely true. I'd rather try to improve things when capture scopes is on.

benaadams:
>  I'd rather try to improve things when capture scopes is on.

Would be good too :)

> I'm not sure this is worth it to avoid the allocation in the case when CaptureScopes is false. This is rarely true.

Trouble is there isn't a halfway cost between no logging; and any logging. If you just wanted events/stacktraces for Warning/Error/Critical; with the option to switch scopes on if anything interesting arises you can't do that currently... 

Well you can switch scopes on and off, but you always pay the cost for them even if they are off and you are doing any level of logging.

NinoFloris:
This got me thinking, can't we change scopes to apply on the way back up? 

Coding pattern is a bit messier but it would allow for the user or Logging itself to switch on scopes after the first error log (or some other condition). And then on way up it'll gather all request information in reverse order. That would be an awesome improvement. 

It would need some way to get around/suppress the EC restore that the AsyncMethodBuilder does though...

PR #1424, opened 2019-04-15
Use new CancellationToken API on netcoreapp3.0

davidfowl:
Implement Register by calling UnsafeRegister on netcoreapp3.0


Tratcher:
Which does what? The docs are blank.

PR #1431, opened 2019-04-16
Update the SDK

davidfowl:
I'd like to try out netstandard2.1 in various extensions projects

### DIFF
--- a/src/Hosting/samples/GenericHostSample/GenericHostSample.csproj
+++ b/src/Hosting/samples/GenericHostSample/GenericHostSample.csproj
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.0;net461</TargetFrameworks>
-    <RuntimeIdentifiers>win7-x64;linux-x64;osx-x64</RuntimeIdentifiers>
     <StartupObject>GenericHostSample.WindowsServiceControlled</StartupObject>
     <OutputType>Exe</OutputType>
     <LangVersion>latest</LangVersion>
--- a/global.json
+++ b/global.json
@@ -1,9 +1,9 @@
 {
   "sdk": {
-    "version": "3.0.100-preview3-010280"
+    "version": "3.0.100-preview4-011136"
   },
   "tools": {
-    "dotnet": "3.0.100-preview3-010280"
+    "dotnet": "3.0.100-preview4-011136"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19217.1",
--- a/src/Shared/src/Directory.Build.props
+++ b/src/Shared/src/Directory.Build.props
@@ -7,6 +7,7 @@
     <IsPackable>true</IsPackable>
     <IsShipping>false</IsShipping>
     <NoBuild>true</NoBuild>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <EnableDefaultItems>false</EnableDefaultItems>
     <IncludeBuildOutput>false</IncludeBuildOutput>


src/Hosting/samples/GenericHostSample/GenericHostSample.csproj
```diff
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.0;net461</TargetFrameworks>
-    <RuntimeIdentifiers>win7-x64;linux-x64;osx-x64</RuntimeIdentifiers>
```
Tratcher: Why? These were useful for testing the windows service.

davidfowl: Maybe, they're useful for net461 maybe but we now make native exes by default. I was also just trying to see if this would fix the issue I was running into but it didn't.

Tratcher: I was using them for standalone publish.

davidfowl: Why? You *had* to do that in the past to get an exe, not you don't. What other reason did you have to self contain deploying this app?

Tratcher: Services run as different users without access to the repo local runtime. It's easier to deploy the runtime with the service than to tell the service how to find it.

davidfowl: You can publish self contained without putting RIDS in the prodject file and doing a RID restore. That's been the case since 2.x.

davidfowl:
```
dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
.dotnet\sdk\3.0.100-preview4-011136\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(115,5): Error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.
PowerShell exited with code '1'.
```

Anybody know wtf this is?

natemcmaster:
This is how we hacked the SDK into producing '.Sources' packages.

davidfowl:
I see this was added:

https://github.com/dotnet/sdk/blob/8cb31b0fd0bb5d36b9781442865f2e26cf6fbedc/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L113-L116

@dsplaisted How do we build a package source only package? Is there anyway to disable this?

dsplaisted:
What is a package source only package?

The check was added to make sure that when you specify "no build", it doesn't actually run the Build target.  I'm not sure there's a point in disabling the check, if you are getting the error and want to disable the check, I think you'd get the same result by no longer specifying NoBuild.  However, I don't know anything about the hack Nate refers to to produce source packages, so that could be an exception.

natemcmaster:
A source package is a .nupkg which only contains *.cs content files, no .dll. I think <NoBuild> was set to suppress the C# compiler from ever running. These projects are using a .csproj as a hack to produce a .nupkg since we don’t have the ability to pack a nuspec directly.

dsplaisted:
I believe it is possible by now to use the Pack target to create a .nupkg from a .nuspec.

I think if you're getting the NoBuildRequested error, it means that some target that was running had a dependency on the Build target, so essentially you would be compiling the code anyway if the error wasn't there.

natemcmaster:
I’m not aware of any changes to nuget and packing. Sure, you can call the pack target and provide a nuspec….but you still need a .csproj file in order to invoke the Pack target, right? That’s what these dummy .csproj files are for.

dsplaisted:
Yes, you still need a dummy .csproj.

natemcmaster:
PR checks are passing now

davidfowl:
FINALLY! 😄

PR #1444, opened 2019-04-17
Update CODEOWNERS

natemcmaster:
Makes @dotnet-maestro-bot the owner of bot-managed files.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,6 +3,9 @@
 
 /azure-pipelines.yml                    @aspnet/build
 /eng/                                   @aspnet/build
+/eng/common/                            @dotnet-maestro-bot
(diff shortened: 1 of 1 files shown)


dougbu:
As in aspnet/AspNetCore#9463, I'd rather we hold off on this until we can get notified automatically.

dougbu:
Also, realized this means we would **not** be notified when someone other than the bot updates any of the files the bot would now own. That's a problem because we likely will miss changes that we should have blocked or corrected before merge e.g. Version.Details.xml changes that add dependencies but don't include needed `CoherentParentDependency` attributes.

natemcmaster:
Added @dougbu as well as the bot so you keep getting notifications on all bot PRs.

Issue #1265, opened 2019-03-16
AzureKeyVaultConfigurationProvider is not honoring ReloadInterval correctly


sukhong:
### Describe the bug
AzureKeyVaultConfigurationProvider is not honoring ReloadInterval correctly.

In https://github.com/aspnet/Extensions/blob/master/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs,

WaitForReload() function is taking _reloadInterval.Value.Milliseconds as delay time.

However, Value.Milliseconds do not corectly represent the whole TimeSpan, as it only contains milliseconds portion of it.

For example:
TimeSpan.FromMilliseconds(5000) returns 0 for its Milliseconds value, 
TimeSpan.FromMilliseconds(5010) will return 10 for Milliseconds value.

### To Reproduce
1. Using 3.0.0-preview4.19164.7 version of Microsoft.Extensions.Configuration.AzureKeyVault package,
2. Run this code:
`var azureKeyVaultConfirationOption = new AzureKeyVaultConfigurationOptions("[some keyvault]");
azureKeyVaultConfirationOption.ReloadInterval = TimeSpan.FromMilliseconds(5000);
config.AddAzureKeyVault(azureKeyVaultConfirationOption);`

3. This will attempt to reload the keys at every possible moment, since it is essentially waiting Task.Delay(0);

### Expected behavior
ReloadInterval should honor the whole TimeSpan, not just Milliseconds portion.

Tratcher:
Yes, this looks like a bug.

anurse:
This is done in https://github.com/aspnet/Extensions/pull/2000

PR #1459, opened 2019-04-18
AzureKeyVaultConfigurationProvider to respect ReloadInterval

arkein:
Summary:
 - Use `Task.Delay(TimeSpan, CancellationToken)`
 - Non-nullable private `_reloadInterval` to simplify code.
 - `.ConfigureAwait(false)` everywhere

Addresses #1265

### DIFF
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -46,7 +46,6 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
                 _reloadInterval = reloadInterval.Value;
             }
 
-            _pollingTask = null;
             _cancellationToken = new CancellationTokenSource();
         }
 
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -130,7 +130,7 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
             // schedule a polling task only if none exists and a valid delay is specified
             if (_pollingTask == null && _reloadInterval > TimeSpan.Zero)
             {
-                _pollingTask = Task.Run(PollForSecretChangesAsync);
+                _pollingTask = PollForSecretChangesAsync();
             }
         }
 
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
     /// </summary>
     internal class AzureKeyVaultConfigurationProvider : ConfigurationProvider, IDisposable
     {
-        private readonly TimeSpan? _reloadInterval;
+        private readonly TimeSpan _reloadInterval;
         private readonly IKeyVaultClient _client;
         private readonly string _vault;
         private readonly IKeyVaultSecretManager _manager;
@@ -36,14 +36,18 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
             _client = client ?? throw new ArgumentNullException(nameof(client));
             _vault = vault ?? throw new ArgumentNullException(nameof(vault));
             _manager = manager ?? throw new ArgumentNullException(nameof(manager));
-            if (reloadInterval != null && reloadInterval.Value <= TimeSpan.Zero)
+
+            if (reloadInterval != null)
             {
-                throw new ArgumentOutOfRangeException (nameof(reloadInterval), reloadInterval, nameof(reloadInterval) + " must be positive.");
+                if (reloadInterval.Value <= TimeSpan.Zero)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(reloadInterval), reloadInterval, nameof(reloadInterval) + " must be positive.");
+                }
+                _reloadInterval = reloadInterval.Value;
             }
 
             _pollingTask = null;
             _cancellationToken = new CancellationTokenSource();
-            _reloadInterval = reloadInterval;
         }
 
         public override void Load() => LoadAsync().ConfigureAwait(false).GetAwaiter().GetResult();
@@ -66,7 +70,7 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
 
         protected virtual async Task WaitForReload()
         {
-            await Task.Delay(_reloadInterval.Value, _cancellationToken.Token);
+            await Task.Delay(_reloadInterval, _cancellationToken.Token);
         }
 
         private async Task LoadAsync()
@@ -124,7 +128,7 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
             }
 
             // schedule a polling task only if none exists and a valid delay is specified
-            if (_pollingTask == null && _reloadInterval != null)
+            if (_pollingTask == null && _reloadInterval > TimeSpan.Zero)
             {
                 _pollingTask = Task.Run(PollForSecretChangesAsync);
             }
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
 
         protected virtual async Task WaitForReload()
         {
-            await Task.Delay(_reloadInterval.Value.Milliseconds, _cancellationToken.Token);
+            await Task.Delay(_reloadInterval.Value, _cancellationToken.Token);
         }
 
         private async Task LoadAsync()
@@ -126,7 +126,7 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
             // schedule a polling task only if none exists and a valid delay is specified
             if (_pollingTask == null && _reloadInterval != null)
             {
-                _pollingTask = PollForSecretChangesAsync();
+                _pollingTask = Task.Run(PollForSecretChangesAsync);
             }
         }
 
--- a/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
@@ -55,10 +55,10 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
         {
             while (!_cancellationToken.IsCancellationRequested)
             {
-                await WaitForReload();
+                await WaitForReload().ConfigureAwait(false);
                 try
                 {
-                    await LoadAsync();
+                    await LoadAsync().ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -66,7 +66,7 @@ private async Task PollForSecretChangesAsync()
 
         protected virtual async Task WaitForReload()
         {
-            await Task.Delay(_reloadInterval.Value.Milliseconds, _cancellationToken.Token);
+            await Task.Delay(_reloadInterval.Value, _cancellationToken.Token);
```
gfoidl: What if `_reloadInterval` is `null`?

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -126,7 +126,7 @@ private async Task LoadAsync()
             // schedule a polling task only if none exists and a valid delay is specified
             if (_pollingTask == null && _reloadInterval != null)
```
gfoidl:
```suggestion
            if (_pollingTask == null && _reloadInterval.HasValue)
```

But this is more a matter of taste, as the C#-compiler will do this for you too.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -126,7 +126,7 @@ private async Task LoadAsync()
...
             {
-                _pollingTask = PollForSecretChangesAsync();
+                _pollingTask = Task.Run(PollForSecretChangesAsync);
```
gfoidl: Is the `Task.Run` necessary? https://github.com/aspnet/Extensions/blob/43d453af6cb5a1087b95ab0b1d5249303a742d75/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs#L51-L55
So `PollForSecretChangesAsync` will `await` a `Task.Delay`. So the `Task.Run` is kind of over-engineering.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
gfoidl: To be precise: In this class this usage is safe, as `WaitForReload` is only called from `PollForSecretChangesAsync` and this (only) call is guarded: https://github.com/aspnet/Extensions/blob/43d453af6cb5a1087b95ab0b1d5249303a742d75/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs#L127-L129

But the access modifier for this method is `protected`, hence (in theory) a sub-class can call `WaitForReload`, which may result in a NRE.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
anurse: We generally prefer using `!= null` in our code style.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
arkein: Polling must run in background. Making a call to async method was the implicit way to achieve that behavior.

I've changed that to an explicit background call because for a short reload interval (<15ms) `WaitForReload()` returns immediately, leading to nested `LoadAsync()` call, which will attempt to create a new `_pollingTask` because it never got assigned in the parent `LoadAsync()` because the first `PollForSecretChangesAsync()` call has not ended yet. Eventually, it crashes the runtime with stack overflow.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
arkein: Agree. I've converted `_reloadInterval` private field from `Nullable<TimeSpan>` to a regular `TimeSpan` to avoid NRE altogether. Please review.

gfoidl: The non-nullable fields looks good.

Minor concern / thought: 
When not given as ctor-argument, its value is `TimeSpan.Zero`, so actually no "wait for reload". 
Should this be handled or is it fine to leave as is (as corner-case)?

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
gfoidl: > Polling must run in background. 

Agree.

> Making a call to async method was the implicit way to achieve that behavior.

Just the call to the async methods doesn't run in background / a different thread. Things start to get interesting at `await`. Then the compiler generated asnyc state-machine yields back to the caller, handing it back a task, which can be awaited.

> because it never got assigned in the parent 

I will be assigned in the parent, so the `if (_pollingTask == null` will prevent recursion and no SO will happen. Whether `PollForSecretChangesAsync` ends or not, isn't relevant in this case, as in https://github.com/aspnet/Extensions/blob/35d8d08f2c123f302222d5eca27523e494907b57/src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs#L59 the state-machine will yield and return the task (of the state-machine), so it is `!= null`.

Look at the [generated code](https://sharplab.io/#v2:EYLgHgbALAPgAgBgARwIwG4CwAoRLUB0AKgBYBOApgIYAmAlgHYDmWuyaArKznAMwoAmJAGEkOAN44k0pAAcydAG5UALhSSVaAewYAbAJ5IidALYUAyrKoMkAfUq6ttAJIM1ZZbqQBeI6YtWDAQAYmRaJgCydLq6dADOFADGOjRxABSoCACUrDJyCspqGtQ0OgYi1okUMap0OkRaANYUDOZaAK5kVXaJldW6tfVNLT5IDBQA7hUMVTUqdQwNza0dXRRpOVIy8kqq6nAQdrJaMYxMB7kyW9I7hfsAHCiHADJONACCcfozG9dIktg8nk6AAzJBpWzHU7MA4+XwMdoxJAAMmRdgcb1c7k8SAAfH4zJZrAQAFoUMJZP55AFA2lHE6xGGHXwABQZwS0ZHMSUoKmEJGsTAocU+30SG0utIAvn8Zdg/vItGpEmoaEhFHQyCp2lQvHBHrCAOpUOgqDlkABK1Tev0BMhpdLgAE4ngQACLVKj6CEYlxucmeAA0PT6cwWSxaxGGDE2dukcryCoKexQBsObJi5u5iV5/MFwtFP0pcf+VJkExI0XUaQAhLZejN+oNFtGCM44sJQwN5jorQBHdrC1XFumlkuj51IY2m81Wxy0CVl2kqMj6JdAh2j2mT160Qvi2NbmQJo9IXoqRIkcEAUTAVVkPZj6+pz7pAHo30hnEwGJyKK/43XE9AOwKUgA=), especially
```c#
    [AsyncStateMachine(typeof(<PollForSecretChangesAsync>d__5))]
    private Task PollForSecretChangesAsync()
    {
        <PollForSecretChangesAsync>d__5 stateMachine = default(<PollForSecretChangesAsync>d__5);
        stateMachine.<>4__this = this;
        stateMachine.<>t__builder = AsyncTaskMethodBuilder.Create();
        stateMachine.<>1__state = -1;
        AsyncTaskMethodBuilder <>t__builder = stateMachine.<>t__builder;
        <>t__builder.Start(ref stateMachine);        // see next snippet
        return stateMachine.<>t__builder.Task;     // <-- here the task is returned
    }

// Snippet from state-machine
                    default:
                        {
                            if (c._cancellationToken.IsCancellationRequested)
                            {
                                break;
                            }
                            awaiter = c.WaitForReload().GetAwaiter();
                            if (!awaiter.IsCompleted)
                            {
                                num = (<>1__state = 0);
                                <>u__1 = awaiter;
                                <>t__builder.AwaitUnsafeOnCompleted(ref awaiter, ref this);
                                return;
                            }
```
So in the state-machine a `awaiter` for `WaitForReload` is assigned, then the completed-event registered, followed by a return. All this is synchronously, but right after that return, the `Task` is returned.

Or you just try it:
```c#
            if (_pollingTask == null && _reloadInterval > TimeSpan.Zero)
            {
                _dummyCounter++;
                if (_dummyCounter > 1)
                    throw new Exception("should not be here ;-)");

                _pollingTask = PollForSecretChangesAsync();
            }
        }

        private int _dummyCounter;
```
Won't crash. No matter what `_reloadInterval` (or if you thread-safe increment the dummy-counter).

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -36,14 +36,18 @@ public AzureKeyVaultConfigurationProvider(IKeyVaultClient client, string vault,
...
             {
-                throw new ArgumentOutOfRangeException (nameof(reloadInterval), reloadInterval, nameof(reloadInterval) + " must be positive.");
+                if (reloadInterval.Value <= TimeSpan.Zero)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(reloadInterval), reloadInterval, nameof(reloadInterval) + " must be positive.");
+                }
+                _reloadInterval = reloadInterval.Value;
             }
 
             _pollingTask = null;
```
gfoidl: BTW: this can be omitted, as in .NET fields are initialized with `default`, which is for referency types `null`.
```suggestion

```

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
```diff
@@ -66,7 +70,7 @@ private async Task PollForSecretChangesAsync()
 
         protected virtual async Task WaitForReload()
         {
-            await Task.Delay(_reloadInterval.Value.Milliseconds, _cancellationToken.Token);
+            await Task.Delay(_reloadInterval, _cancellationToken.Token);
```
gfoidl:
```suggestion
            await Task.Delay(_reloadInterval, _cancellationToken.Token).ConfigureAwait(false);
```

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
arkein: I have to admit that your explanation totally makes sense; though I witnessed the stack overflow at my localhost... I must have been experimenting with `WaitForReload()` since it's quite meaningless async method that exists purely for unit testing.

I rolled that back, thanks for taking time to explain this.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
arkein: Corrected.

src/Configuration/Config.AzureKeyVault/src/AzureKeyVaultConfigurationProvider.cs
arkein: Corrected.

anurse:
As announced in https://github.com/aspnet/Announcements/issues/408, these packages are now part of the Azure SDK. If you're still interested in contributing this change, please open a new PR in https://github.com/Azure/azure-sdk-for-net .

PR #1476, opened 2019-04-19
skip flaky attribute tests, they're just too annoying

anurse:
These tests are causing all AzP builds to be yellow and I don't think they're providing value. Leaving them in so we can run them when making changes to `FlakyAttribute` behavior, but marking as skipped.

cc @rynowak @natemcmaster @davidfowl


anurse:
All is well again:

image

PR #1479, opened 2019-04-19
Lazy create ServiceProviderEngineScope.DisposeAsync statemachine

benaadams:
Last one in the common path

image

Not sure who code owners its for DI /cc @Tratcher @anurse


anurse:
Looks ok to me. @pakrym You have some familiarity here, do you know if we have some reasonable test coverage on this area? Assuming we're fairly well covered, I don't see a problem taking this.

pakrym:
LGTM

PR #1504, opened 2019-04-21
Fixes aspnet/Extensions#1503

ENikS:
 - Fixes an issue where dependencies are never instantiated due to lazy resolution of enumerable dependencies by the Unity container

aspnet/Extensions#1503

### DIFF
--- a/src/DependencyInjection/DI.Specification.Tests/src/Fakes/FakeDisposableCallbackOuterService.cs
+++ b/src/DependencyInjection/DI.Specification.Tests/src/Fakes/FakeDisposableCallbackOuterService.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Collections.Generic;
 
 namespace Microsoft.Extensions.DependencyInjection.Specification.Fakes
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/DI.Specification.Tests/src/Fakes/FakeDisposableCallbackOuterService.cs
```diff
@@ -13,10 +14,10 @@ public class FakeDisposableCallbackOuterService : FakeDisposableCallbackService,
             FakeDisposeCallback callback) : base(callback)
         {
             SingleService = singleService;
-            MultipleServices = multipleServices;
```
davidfowl: This can be null?

ENikS:
In one of CI tests enumerable is null. Without the null check it fails.

davidfowl:
> In one of CI tests enumerable is null. Without the null check it fails.

Which one?

ENikS:
I am not sure how to get logs for previous build. Do you want me to roll back so you could see it?

davidfowl:
Roll back locally and figure out which ones are passing null. These tests run super quickly, and you should be able to run them in VS.

ENikS:
I can't build locally at the moment, I don't have required build tools installed. The solution does not load projects.

I can temporary roll back the null check and let it fail CI checks.

src/DependencyInjection/DI.Specification.Tests/src/Fakes/FakeDisposableCallbackOuterService.cs
ENikS: https://github.com/aspnet/Extensions/pull/1504/checks?check_run_id=107248909

ENikS:
Let me know if you want the null check back or you would rather fix the CI test. 

Theoretically the enumerable should never be null in IoC environment. But I am assuming CI build does not use IoC for the tests.

PR #1563, opened 2019-04-25
Fix #519

### LINKED ISSUE #519: Are HttpMessageHandlers scoped per Typed HttpClient or a the internal HttpClient?
I have the following scenario:
* I have two different typed `HttpClient`s
* I have a customer `HttpMessageHandler` called `MyCustomHandler`
* I Configure them like this:
```
services
     .AddHttpClient<TeamCity.Client>()
     .AddHttpMessageHandler(() => new MyCustomHandler())

services
     .AddHttpClient<GitHub.Client>()
     .AddHttpMessageHandler(() => new MyCustomHandler())
```
* I notice that the handler attached to `TeamCity.Client` is actually intercepting requests made through `GitHub.Client` ?

I don't understand what is happening here?
1. Are those handlers attached and registered by type and thus they monitor every outgoing request even if it is not made through their typed http client?
1. If the answer to previous question is yes, then why you can ONLY configure handlers when you register (`AddHttpClient<>()`) named or typed clients?

rynowak:
Adds tracking for the relationship between named clients and the types
that they are bound to.

This allows us to throw a better in common cases that won't work:

- using clients with the same type name but different FQNs
- using the came client type twice with different names

### DIFF
--- a/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
+++ b/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
+using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using Microsoft.Extensions.Http;
@@ -321,6 +323,8 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(builder));
             }
 
+            ReserveClient(builder, typeof(TClient), builder.Name);
+
             builder.Services.AddTransient<TClient>(s =>
             {
                 var httpClientFactory = s.GetRequiredService<IHttpClientFactory>();
@@ -373,6 +377,8 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(builder));
             }
 
+            ReserveClient(builder, typeof(TClient), builder.Name);
+
             builder.Services.AddTransient<TClient>(s =>
             {
                 var httpClientFactory = s.GetRequiredService<IHttpClientFactory>();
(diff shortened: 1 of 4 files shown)


src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -514,5 +524,35 @@ public static IHttpClientBuilder SetHandlerLifetime(this IHttpClientBuilder buil
             builder.Services.Configure<HttpClientFactoryOptions>(builder.Name, options => options.HandlerLifetime = handlerLifetime);
             return builder;
         }
+
+        // See comments on HttpClientMappingRegistry.
+        private static void ReserveClient(IHttpClientBuilder builder, Type type, string name)
+        {
+            var registry = (HttpClientMappingRegistry)builder.Services.Single(sd => sd.ServiceType == typeof(HttpClientMappingRegistry)).ImplementationInstance;
```
rynowak: @davidfowl if you don't like this I have another cool idea https://github.com/aspnet/Extensions/commit/b44b4590453592d4df9de449e22f36f78c028f4a

davidfowl: This is horrible but it's less horrible than the other way.

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -47,6 +47,7 @@ public static IServiceCollection AddHttpClient(this IServiceCollection services)
             // Misc infrastructure
             //
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IHttpMessageHandlerBuilderFilter, LoggingHttpMessageHandlerBuilderFilter>());
+            services.TryAddSingleton(new HttpClientMappingRegistry());
```
davidfowl: Add a comment here saying that this *has* to be an instance or it'll break the upstream code.

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
Jetski5822: I really wish you had dont the `cool idea` way - just hit a bug with this in Orchard Core. The Cool Way would have worked :(

rynowak: I was trying to get @davidfowl to allow `ConditionalWeakTable` into the codebase for **years** 😆

PR #1620, opened 2019-04-29
Adding new Polly extension with policy key selection logic

shagilt:
Adding new Polly extension with policy key selection logic.
https://github.com/App-vNext/Polly/issues/624#issuecomment-478218169


src/HttpClientFactory/Polly/src/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -234,5 +234,55 @@ public static IHttpClientBuilder AddPolicyHandlerFromRegistry(this IHttpClientBu
             builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policy));
             return builder;
         }
+
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with a policy returned
+        /// by executing provided key selection logic <paramref name="policySelector"/>. Based on the
```
rynowak: Looks like this message just trailed off.

src/HttpClientFactory/Polly/src/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -234,5 +234,55 @@ public static IHttpClientBuilder AddPolicyHandlerFromRegistry(this IHttpClientBu
             builder.AddHttpMessageHandler(() => new PolicyHttpMessageHandler(policy));
             return builder;
         }
+
+        /// <summary>
+        /// Adds a <see cref="PolicyHttpMessageHandler"/> which will surround request execution with a policy returned
+        /// by executing provided key selection logic <paramref name="policySelector"/>. Based on the 
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="keySelector">To select a Policy key based on <seealso cref="HttpRequestMessage"/>.</param>
```
rynowak:
```suggestion
        /// <param name="keySelector">A delegate used to select a policy key based on the <see cref="HttpRequestMessage"/>.</param>
```

src/HttpClientFactory/Polly/src/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -234,5 +234,55 @@ public static IHttpClientBuilder AddPolicyHandlerFromRegistry(this IHttpClientBu
...
+        /// <param name="policySelector">Selects an <see cref="IAsyncPolicy{HttpResponseMessage}"/> to apply to the current request based on key selection.</param>
```
rynowak: This should be called `policyFactory` since it's used to instantiate and store the policy if it's not present.

src/HttpClientFactory/Polly/src/DependencyInjection/PollyHttpClientBuilderExtensions.cs
```diff
@@ -234,5 +234,55 @@ public static IHttpClientBuilder AddPolicyHandlerFromRegistry(this IHttpClientBu
...
+        /// <returns>An <see cref="IHttpClientBuilder"/> that can be used to configure the client.</returns>
+        /// <remarks>
+        /// <para>
+        /// See the remarks on <see cref="PolicyHttpMessageHandler"/> for guidance on configuring policies.
+        /// </para>
+        /// </remarks>
```
rynowak: Somewhere, this documentation needs to explain that the key is first used to look in the registry, and then the delegate is only called if the policy is not present.

rynowak:
Looks like you've got some build errors

rynowak:
Thanks @shagilt - will merge this once the build passed

rynowak:
Thanks!

PR #1623, opened 2019-04-30
Add the event log logger provider by default

davidfowl:
Fixes https://github.com/aspnet/Extensions/issues/1628

- This change turns on the event log logger provider by default if we're running on windows and sets a default filters for warning or above. We add the first so that it can be overridden by configuration.

This works when put in appsettings.json:

```JSON
{
  "Logging": {
    "LogLevel": {
      "Default": "Information",
      "Microsoft": "Warning",
      "Microsoft.Hosting.Lifetime": "Information"
    },
    "EventLog": {
      "LogLevel": {
        "Default": "Information"
      }
    }
  },
  "AllowedHosts": "*"
}
```

cc @natemcmaster This needs to be added to the ASP.NET Core shared framework. Do I need to wait until that change flows in to do that?

### DIFF
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -3,8 +3,10 @@
 
 using System.IO;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.EventLog;
 
 namespace Microsoft.Extensions.Hosting
 {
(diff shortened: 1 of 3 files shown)


natemcmaster:
> Do I need to wait until that change flows in to do that?

Nope. You can add it now. See https://github.com/aspnet/AspNetCore/blob/master/docs/SharedFramework.md#how-to-adjust-what-is-in-the-shared-framework

davidfowl:
> Nope. You can add it now. See https://github.com/aspnet/AspNetCore/blob/master/docs/SharedFramework.md#how-to-adjust-what-is-in-the-shared-framework

Great! But now I'm curious. This should fail on the PR that tries to get a new build of extensions if I don't make that change before right?

natemcmaster:
Yes, it should because there will be new transitive dependencies pulled in from M.Ext.Hosting. You'll need to add all of those to the shared framework, including any System.\* dependencies which are not already part of the base fx.

src/Hosting/Hosting/src/Host.cs
```diff
@@ -90,10 +92,26 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
...
             .ConfigureLogging((hostingContext, logging) =>
             {
+                var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+
+                // IMPORTANT: This needs to be added *before* configuration is loaded, this lets
+                // the defaults be overridden by the configuration.
+                if (isWindows)
+                {
+                    // Default the EventLogLoggerProvider to warning or above
+                    logging.AddFilter<EventLogLoggerProvider>(level => level >= LogLevel.Warning);
```
benaadams: This overrides any of the general category logging levels and has to be specifically overriden using an `"EventLog"` sub-setting. Is this desirable?

benaadams: e.g.
```
  "Logging": {
    "LogLevel": {
      "Microsoft.AspNetCore.Hosting.Diagnostics": "None"
    }
  },
```
Will not override it

yousiftouma: Is this documented anywhere? Had a lot of trouble figuring this out.

PR #1625, opened 2019-04-30
Shuffled guts of the EventLogLogger

davidfowl:
Fixes https://github.com/aspnet/Extensions/issues/1406

- Don't create a settings object or IEventLog per EventLogger
- Use a string builder internally instead of concating strings
- Clean up the formatting a bit
  - Added headings for various bits of info.
  - Added the event id to the message itself.
- Use .NET Runtime with a specific eventId as the default to avoid the warning about logging an unknown event id.
- Added more tests

PS: Somebackground on that Event Viewer issue we run into:

- https://www.jitbit.com/alexblog/266-writing-to-an-event-log-from-net-without-the-description-for-event-id-nonsense/
- https://stackoverflow.com/questions/3412463/description-for-event-id-from-source-cannot-be-found
- https://www.codeproject.com/Articles/4153/Getting-the-most-out-of-Event-Viewer

image

image

image

### DIFF
--- a/src/Logging/Logging.EventLog/src/EventLogLogger.cs
+++ b/src/Logging/Logging.EventLog/src/EventLogLogger.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Text;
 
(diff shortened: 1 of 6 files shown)


anurse:
Event **Log** Logger, right? Not Event **Source** Logger?

Issue #1549, opened 2019-04-24
HttpClientFactory logs - "StatusCode" property is inconsistent with other internal logs


sschneider1207:
## Info

Internal handlers used in the `HttpClientFactory` pattern, the [LoggingHttpMessageHandler](https://github.com/aspnet/Extensions/blob/master/src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandler.cs) and the [LoggingScopeHttpMessageHandler](https://github.com/aspnet/Extensions/blob/master/src/HttpClientFactory/Http/src/Logging/LoggingScopeHttpMessageHandler.cs), create log messages after successful http requests in the following formats

LoggingHttpMessageHandler:
```
Received HTTP response after {ElapsedMilliseconds}ms - {StatusCode}
```
LoggingScopeHttpMessageHandler:
```
End processing HTTP request after {ElapsedMilliseconds}ms - {StatusCode}
```
An example of these are:
```
Received HTTP response after 56.0044ms - OK
End processing HTTP request after 70.0862ms - OK
```

As you can see, the `StatusCode` property is sent as the textual representation of the status code instead of the integer version (e.g. `200` instead of `OK`).  

This conflicts with an internal log, the [HostingRequestFinishedLog](https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs), which generates logs in the following format:
```
Request finished in {0}ms {1} {2}
```
An example of this is:
```
Request finished in 605.5877ms 200 application/json
```

Here, the `StatusCode` property is the integer representation of the status code and not the text like it is in the above two message handlers.

## Problem

This conflict is causing an issue with dropped log messages in our system because we send our logs to Elasticsearch (via the [Serilog sink](https://github.com/serilog/serilog-sinks-elasticsearch)), and Elasticsearch requires consistent types for properties with the same name.  

In this case, if a `HostingRequestFinishedLog` log is sent first, Elasticsearch maps the `StatusCode` property to a `long` and all the logs from the two message handlers are discarded due to their `StatusCode` property being the wrong type.  Vice versa if the message handlers get their log in first.

I think the message handler logs should be consistent with the internal framework logs and send their `StatusCode` property as an integer, it would resolve this conflict and I generally think the integer is more useful anyway.  I'd be happy to submit a PR for this others agree.

Systemic33:
We have also identified this issue, so we would be very happy to see this completed.

[...]

PR #1629, opened 2019-04-30
Log HTTP status codes as integers

sschneider1207:
Modifies the `LoggingHttpMessageHandler` and `LoggingScopeHttpMessageHandler` to log status codes explicitly as integers in order to be consistent with [HostingRequestFinishedLog](https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs).  

This is necessary for logging stores that require consistent types for properties with the same name (e.g. Elasticsearch).  More details on the problem with the current logs are in the linked issue.

Addresses #1549

### DIFF
--- a/src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandler.cs
+++ b/src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandler.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Extensions.Http.Logging
                 EventIds.RequestStart,
                 "Sending HTTP request {HttpMethod} {Uri}");
 
-            private static readonly Action<ILogger, double, HttpStatusCode, Exception> _requestEnd = LoggerMessage.Define<double, HttpStatusCode>(
+            private static readonly Action<ILogger, double, int, Exception> _requestEnd = LoggerMessage.Define<double, int>(
                 LogLevel.Information,
                 EventIds.RequestEnd,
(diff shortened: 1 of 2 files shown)


Systemic33:
Hi,
Is there anything keeping this from being merged?

rynowak:
This is technically a breaking change, so we need to follow up on a little bit of process internally.

Systemic33:
If we only care about the API not breaking, then it could be implemented non-breaking with a new underlying Logging Action that uses int, e.g. `_requestEnd2` , and then use this instead in the `RequestEnd(ILogger logger, HttpResponseMessage response, TimeSpan duration)` with a cast to `int`.

Obviously that's a bit ugly 😅, but then the bug can be fixed immediately.

rynowak:
> If we only care about the API not breaking

We care about changes to behavior. Even though its an improvement it's still a breaking change - it can break folks existing log queries.

rynowak:
Thanks!

mle-ii:
This fix will fix an issue we just realized we were hitting with graylog.  StatusCode is sent as a number normally and this location sent it as a string.  When it gets indexed by the system the first one sent per index determines its type, if it's first set to a number value and then comes in as a string then it causes a parser/index failure and thus the message never makes it to our graylog instance.

http://docs.graylog.org/en/3.1/pages/indexer_failures.html

I'm going to guess this fixes more than it breaks.  When might this change make it to a released product, or has it already?

PR #1630, opened 2019-04-30
Add stream based config providers

HaoK:


src/Configuration/Config.Abstractions/src/IConfigurationProvider.cs
```diff
@@ -37,6 +37,12 @@ public interface IConfigurationProvider
         /// </summary>
         void Load();
 
+        /// <summary>
+        /// Use the following values as the configuration data for this provider.
+        /// </summary>
+        /// <param name="data">The configuration data this provider should use.</param>
+        void Use(IDictionary<string, string> data);
```
HaoK: This name kind of sucks, considered a Load() overload but that seemed confusing too, open to other ideas here...

Tratcher:
There's no reload detection with this model, correct?

How concerned are we about sync reading from async io sources?

src/Configuration/Config.Abstractions/src/IConfigurationProvider.cs
Tratcher: - AddData
- UseData
- SetData

HaoK:
Yeah no way to detect reloads automatically via these new apis, people who want reloadOnChange will need to continue using the old File based ones, or they could setup their own file watching and call reload on the config explicitly.

But yeah we probably do need to do something in StreamProvider.Load to handle it being called again, I'll add a test for that now

src/Configuration/Config.Ini/src/IniConfigurationStreamLoader.cs
```diff
@@ -0,0 +1,85 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.Extensions.Configuration.Ini
+{
+    /// <summary>
+    /// An INI file based <see cref="IConfigurationStreamLoader"/>.
+    /// </summary>
+    public class IniConfigurationStreamLoader : IConfigurationStreamLoader
```
ajcvickers: Should we ship this? (Using an ini file from a stream doesn't seem super useful to me, but then again the refactoring into this doesn't harm, so...)

src/Configuration/Config.Abstractions/src/IConfigurationProvider.cs
ajcvickers: `UseData` is my preference out of the suggestions, but I don't feel strongly about it.

src/Configuration/Config.Ini/src/IniConfigurationStreamLoader.cs
anurse: This is indirectly part of https://github.com/aspnet/Extensions/issues/1631 which would improve the Generic Host experience in other app models such as Xamarin where you don't always have a file path for a configuration file. If `.ini` is part of our supported set of config providers, I think it should work with an arbitrary stream too.

anurse:
> Yeah no way to detect reloads automatically via these new apis

Seems reasonable to me. If you're using a stream that comes from somewhere like an embedded resource or some app-model-dependent location, you'd also need to handle reload (if any) manually.

src/Configuration/Config.Ini/src/IniConfigurationStreamLoader.cs
ajcvickers: Fair enough

src/Configuration/Config.Abstractions/ref/Microsoft.Extensions.Configuration.Abstractions.netstandard2.0.cs
```diff
@@ -44,6 +44,7 @@ public partial interface IConfigurationProvider
         void Load();
         void Set(string key, string value);
         bool TryGet(string key, out string value);
+        void Use(System.Collections.Generic.IDictionary<string, string> data);
```
davidfowl: This is a breaking change... We shouldn't do that with extensions unless warranted. Why does this pass the bar?

davidfowl: OK so I spent some time looking at this and this breaks the IConfiguratonProvider abstraction generally. Pushing data into the IConfigurationProvider doesn't make sense as it's supposed to be aware of where it got the data from. It cannot be fed data.

src/Configuration/Config.Ini/src/IniConfigurationStreamLoader.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        /// <summary>
+        /// Read a stream of INI values into a key/value dictionary.
+        /// </summary>
+        /// <param name="stream">The stream of INI data.</param>
+        /// <returns></returns>
+        public static IDictionary<string, string> Read(Stream stream)
```
Tratcher: Why is Read public?

HaoK: Doesn't seem to hurt leaving the helper methods on the stream loaders public static if people wanted to use them directly.

src/Configuration/Config.Abstractions/src/IConfigurationStreamLoader.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace Microsoft.Extensions.Configuration
+{
+    /// <summary>
+    /// Loads configuration key/values from a stream into a provider.
+    /// </summary>
+    public interface IConfigurationStreamLoader
```
davidfowl: Why do we need this to be public?

HaoK: The base ConfigurationStreamLoader has a Loader property which is what the Json/Xml/Ini providers set to the appropriate type in the builder AddXyzStream() APIs

HaoK:
Breaking change removed, @davidfowl how does this look now?

src/Configuration/Config.Abstractions/src/IConfigurationStreamLoader.cs
davidfowl: I don’t think this interface should be public. Let’s make it internal and make the AddStream taking a loader internal as well. We can leave the Read helpers on the concrete types public

HaoK: So are you saying do internals visible to all the specific provider packages?

HaoK: The only drawback I see with that is other providers that aren't are own won't be able to expose a stream overload as easily but I guess they can just copy the code

davidfowl: I think the code sharing here is minimal and not worth exposing a new public method. The user facing methods are Add{format}Stream and that's it. `AddStream(IConfigurationStreamLoader)` overload seems unnecessary

HaoK: Ok so if we don't want this concept in abstractions, rather than going internal, I can probably just make `StreamConfigurationProvider` abstract with `abstract void Load(Stream stream)` that won't require any internals visible

HaoK:
Updated removing changes from abstractions, done via StreamConfigurationProvider abstract Load(Stream) instead, there's now one source/provider pair per AddXyzStream

src/Configuration/Config.Abstractions/src/IConfigurationStreamLoader.cs
davidfowl: Seems fine to me. I'd prefer that.

src/Configuration/Config.Json/src/JsonStreamConfigurationProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="source">The <see cref="JsonStreamConfigurationSource"/>.</param>
+        public JsonStreamConfigurationProvider(JsonStreamConfigurationSource source) : base(source) { }
+
+        /// <summary>
+        /// Loads json configuration key/values from a stream into a provider.
+        /// </summary>
+        /// <param name="stream">The json <see cref="Stream"/> to load configuration data from.</param>
+        public override void Load(Stream stream)
+        {
+            var data = JsonConfigurationFileParser.Parse(stream);
```
davidfowl: Why not set Data directly instead of calling Set?

HaoK: Sure, can switch to setting Data directly, its the same thing

HaoK: Oh yeah you mean just set the dictionary now yeah

PR #1632, opened 2019-05-01
Added information about using Castle.Windsor

schwarzie2478:
as an dependency injection container inside ASP.Net Core
 - Added reference in ReadMe.md
 - Added link to nuget package of Castle.Facilities.AspNetCore
 - Added sample code for usage of the container

### DIFF
--- a/eng/Dependencies.props
+++ b/eng/Dependencies.props
@@ -83,6 +83,7 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Stashbox.Extensions.Dependencyinjection" Version="2.6.3" />
     <LatestPackageReference Include="StructureMap.Microsoft.DependencyInjection" Version="1.4.0" />
     <LatestPackageReference Include="Unity.Microsoft.DependencyInjection" Version="2.1.3" />
+    <LatestPackageReference Include="Castle.Facilities.AspNetCore" Version="5.0.0" />
   </ItemGroup>
 
 </Project>
--- a/src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
+++ b/src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
 
         protected sealed override IServiceProvider CreateServiceProvider(IServiceCollection serviceCollection)
         {
-            foreach (var stackFrame in new StackTrace(1).GetFrames().Take(2))
+            foreach (var stackFrame in new System.Diagnostics.StackTrace(1).GetFrames().Take(2))
             {
                 if (SkippedTests.Contains(stackFrame.GetMethod().Name))
                 {
--- a/src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
+++ b/src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
 
         protected sealed override IServiceProvider CreateServiceProvider(IServiceCollection serviceCollection)
         {
-            foreach (var stackFrame in new System.Diagnostics.StackTrace(1).GetFrames().Take(2))
+            foreach (var stackFrame in new StackTrace(1).GetFrames().Take(2))
             {
                 if (SkippedTests.Contains(stackFrame.GetMethod().Name))
                 {
--- a/src/DependencyInjection/DI.External.Tests/test/Windsor.cs
+++ b/src/DependencyInjection/DI.External.Tests/test/Windsor.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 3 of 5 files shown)


schwarzie2478:
I could use some help getting this unit test to run, I was not succesful in getting my system configured succesfully to run the unit tests.  The build is succusful, I'm at a loss get the test running for Windsor.

src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
```diff
@@ -11,7 +11,7 @@ public abstract class SkippableDependencyInjectionSpecificationTests: Dependency
 
         protected sealed override IServiceProvider CreateServiceProvider(IServiceCollection serviceCollection)
         {
-            foreach (var stackFrame in new StackTrace(1).GetFrames().Take(2))
+            foreach (var stackFrame in new System.Diagnostics.StackTrace(1).GetFrames().Take(2))
```
dougbu: Why is this change necessary?

src/DependencyInjection/DI.External.Tests/test/Windsor.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public class WindsorDependencyInjectionSpecificationTests : SkippableDependencyInjectionSpecificationTests
+    {
+        private static readonly WindsorContainer Container = new WindsorContainer();
+        private class HomeController
+        { }
+        public override string[] SkippedTests => throw new NotImplementedException();
```
anurse: Throwing here is causing all the tests to fail. If you have no skipped tests you need to return an empty array:

```suggestion
        public override string[] SkippedTests => Array.Empty<string>();
```

src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
schwarzie2478: This namespace shouldn't be needed but I had the strange behavior that I couldn't get to resolve StackTrace unless I fully specified it. Using statement kept being ignored.  Both intellisense and the compiler complained about it.  I had the option enable to let VS use preview tooling and the latest preview4 build was just installed.

pakrym: This shouldn't be required.

pakrym:
As long as the test passes this seems to be a fine change, but it's not passing ATM.

davidfowl:
The tests all fail with 

```
System.ArgumentException : Facility of type 'Castle.Facilities.AspNetCore.AspNetCoreFacility' has already been registered with the container. Only one facility of a given type can exist in the container.
```

PS: Castle should drop the ASP.NET Core branding from this plugin.

src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
schwarzie2478: Removed it from the code.  Intellisence still complains but the code builds and the tests run ok.

src/DependencyInjection/DI.External.Tests/test/Windsor.cs
```diff
@@ -0,0 +1,40 @@
+using Castle.Facilities.AspNetCore;
```
Eilon: Missing license header in this file.

eng/Dependencies.props
```diff
@@ -83,6 +83,7 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Stashbox.Extensions.Dependencyinjection" Version="2.6.3" />
     <LatestPackageReference Include="StructureMap.Microsoft.DependencyInjection" Version="1.4.0" />
     <LatestPackageReference Include="Unity.Microsoft.DependencyInjection" Version="2.1.3" />
+    <LatestPackageReference Include="Castle.Facilities.AspNetCore" Version="5.0.0" />
```
davidfowl: This repository shouldn't depend on ASP.NET Core. Can you remove that from the container implementation? None of the other containers rely on ASP.NET Core here.

schwarzie2478: I'm not the maintainer of this package, but I opened an discussion on that repo page. [https://github.com/castleproject/Windsor/issues/487](url)

I'm interested to hear from them too...

src/DependencyInjection/DI.External.Tests/test/SkippableDependencyInjectionSpecificationTests.cs
ericstj: It's possible Intellisense was finding an internal StackTrace member in some other namespace...  Still intellisense and compile ought to be using the same algorithm.  If you can reproduce it we can file a bug against roslyn.

ericstj:
I believe this code has now moved to dotnet/runtime /cc @maryamariyan 

This PR will need to be opened there.  We should also avoid an upstack dependency on ASPNetCore as @davidfowl suggests if we add this to dotnet/runtime tests.

maryamariyan:
Yes DI.External tests are moved to dotnet/runtime repo:

https://github.com/dotnet/runtime/tree/master/src/libraries/Microsoft.Extensions.DependencyInjection/tests/DI.External.Tests

PR #1654, opened 2019-05-03
StyleCop Returns?

anurse:
Ok, so we've had our encounters with StyleCop in the past, but I think there may be some value in it. StyleCop is [now a set of Roslyn Analyzers **and Code Fixes**](https://github.com/DotNetAnalyzers/StyleCopAnalyzers). This is a (possibly crazy) PR that adds it to our repo and enables a **single rule**. @davidfowl 's favorite rule in fact, [`SA1000: KeywordsMustBeSpacedCorrectly`](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1000.md).

If we believe this is a terrible idea, I'll go away, but I don't think it's the best use of our time to have code reviewers nit picking things like spacing, `var` usage, license headers, etc. If we want to try it further, I can enable a few more core rules and see where that takes us. I can also bring it to aspnet/AspNetCore.

StyleCop has [a **lot** of rules](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/tree/master/documentation) but it is also [very configurable](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md). If we are cautious and introduce rules that mostly already align with what we intend as style conventions, I think this will add value. The goal is not to reshape our code around some specific style guidance but to use robots to check style adherance instead of spending code review time and iterations on it.

These errors show up in command-line builds:

image

And in VS (where they can also be `Ctrl-.`ed away):

image

image

Thoughts? @davidfowl @rynowak @Eilon

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -58,6 +58,10 @@
     <Reference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="All" />
+  </ItemGroup>
+
   <Import Project="eng\Dependencies.props" />
   <Import Project="eng\PatchConfig.props" />
   <Import Project="eng\ProjectReferences.props" />
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -62,6 +62,10 @@
     <Reference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="All" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)eng\StyleCop.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+
   <Import Project="eng\Dependencies.props" />
   <Import Project="eng\PatchConfig.props" />
   <Import Project="eng\ProjectReferences.props" />
--- a/Extensions.sln
+++ b/Extensions.sln
@@ -305,6 +305,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.Config
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.Configuration.NewtonsoftJson.Tests", "src\Configuration\Config.NewtonsoftJson\test\Microsoft.Extensions.Configuration.NewtonsoftJson.Tests.csproj", "{2FEE9416-E709-4EA7-82A3-4A34816F75FD}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "eng", "eng", "{E9C6F0EE-B104-4E82-8B21-4690F9083B08}"
+	ProjectSection(SolutionItems) = preProject
+		eng\StyleCop.ruleset = eng\StyleCop.ruleset
+	EndProjectSection
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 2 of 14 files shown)


anurse:
cc @aspnet/build for new dependency **if** we decide to try this out :)

davidfowl:
You should add code coverage first 😉

davidfowl:
Failing the build for style errors is evil so I don't support this, at all...

anurse:
> You should add code coverage first 😉

Homer Simpson stands up and tears a booklet in half while shouting "NEVER"

anurse:
> Failing the build for style errors is evil

Explain?

davidfowl:
I think issues like this are better solved with editor config like rules that light up in the IDE rather than build failures. Style errors just aren't important enough to have people pushing new commits to PRs and waiting an hour (or however long it takes tests to build and tests to pass). We don't have a problem with people writing code in the right style (except you and spaces between things 😄 ), we don't however have code coverage 😄

rynowak:
I think we should add `.editorconfig` and `dotnet-format` and then enable something like this based on what that supports.

I want to avoid having discussions about minor style issues that there's no debate over. @JamesNK follows me around and tells me about every extra blank line I've ever created. That matters to some people and they aren't wrong, but I have a lower sensitivity to it. I agree with him in principle, but I don't want to spend time on it. It's much better for the system to fix it more me.

I'm open to trying anything once, and I have a clear idea of what I think success looks like - **less time spent by the team on trivial stuff**. 

I like *more* ideas like having a minimal `editorconfig` and an agreement that it should be respected (using VS code cleanup + `dotnet-format`). I like the most ideas like having a bot to auto-format the world nightly and send a PR.

anurse:
> I think we should add .editorconfig and dotnet-format and then enable something like this based on what that supports

I looked in to `dotnet-format` and wasn't super happy with what I saw (very limited functionality, it just does whitespace reformatting as far as I could tell). It doesn't support most of the editorconfig options right now. Still, it's a start.

> Style errors just aren't important enough to have people pushing new commits to PRs and waiting an hour (or however long it takes tests to build and tests to pass).

This is actual feedback. Thanks. Though a "nit" comment about a style issue tends to yield exactly the same result. If all the feedback is "nit"s like this would you expect someone to update the PR?

> except you and spaces between things

eye roll

> I'm open to trying anything once, and I have a clear idea of what I think success looks like - **less time spent by the team on trivial stuff**.

This is what I'm going for. I find style "nits" tend to be unhelpful and trivial. I think it's a waste of the reviewers time when automation can take care of it.

> I like _more_ ideas like having a minimal `editorconfig` and an agreement that it should be respected (using VS code cleanup + `dotnet-format`). I like the most ideas like having a bot to auto-format the world nightly and send a PR.

I'm on board with something like that. It got [shot down last time](https://github.com/aspnet/Universe/issues/500) so I thought it wasn't generally an option we were interested in.

rynowak:
> I'm on board with something like that. It got shot down last time so I thought it wasn't generally an option we were interested in.

I think this is worth a revisit. What's great about me is that I'm willing to change my mind. Maybe other people want to be great too.

anurse:
Here's what `dotnet-format` yields on Extensions (with no changes to our existing editorconfig): https://github.com/aspnet/Extensions/compare/anurse/dotnet-format?expand=1

There are definitely some... interesting things in our code ;).

It does look like `dotnet-format` is considering work on supporting more editorconfig options but right now there's no need to add any for that purpose since we use the default for all the options `dotnet-format` supports. There's value in doing it for VS scenarios though.

davidfowl:
Check that thing in.

JamesNK:
> Failing the build for style errors is evil so I don't support this, at all... I'm supporting of using it at IDE time though.

+1. TS with SignalR had (has?) it cranked up to that max that the build would fail on the smallest infraction.

An `.editorconfig` would be useful because it tells Roslyn what do when refactoring, e.g. _underscoreFields

halter73:
> TS with SignalR had (has?) it cranked up to that max that the build would fail on the smallest infraction.

Working with SignalR's TS client recently, and I really appreciated the linter failing the build.

> Style errors just aren't important enough to have people pushing new commits to PRs and waiting an hour (or however long it takes tests to build and tests to pass). 

I think this is a straw man. We just need to make sure both local builds fail and PR checks fail quickly for style errors. No one expects changes that fail to even build locally to pass our PR checks, right?

davidfowl:
If you can make it fail local builds only, that’d be more ok. Still let’s start with an editor config

PR #1687, opened 2019-05-07
Remove the exception filter that is problematic for wasm AOT

lewing:
LLVM does not currently have IR to represent filter expressions and mono relies on LLVM IR when AOTing.  Removing the exception filter allows InvokeSynchronously to function properly in AOT and mixed (AOT+interpreter) modes for WebAssembly

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
@@ -171,10 +171,15 @@ namespace Microsoft.JSInterop
             {
                 return methodInfo.Invoke(targetInstance, suppliedArgs);
             }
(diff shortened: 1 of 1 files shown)


src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -171,10 +171,15 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
             {
                 return methodInfo.Invoke(targetInstance, suppliedArgs);
             }
-            catch (TargetInvocationException tie) when (tie.InnerException != null)
+            catch (TargetInvocationException tie)
```
rynowak: Add a comment so someone else doesn't *optimize* this?

PR #1692, opened 2019-05-09
Use Arcade to manage .NET Core installation and remove custom logger download

natemcmaster:
Removes copies of our dotnet-install scripts in favor of a new Arcade feature that installs .NET Core runtimes for us.

(kudos to @chcosta)


natemcmaster:
Blocked on https://github.com/dotnet/arcade/issues/2725

global.json
```diff
@@ -3,7 +3,12 @@
     "version": "3.0.100-preview5-011568"
   },
   "tools": {
-    "dotnet": "3.0.100-preview5-011568"
+    "dotnet": "3.0.100-preview5-011568",
+    "runtimes": {
+      "dotnet": [
+        "$(MicrosoftNETCoreAppPackageVersion)"
```
JunTaoLuo: global.json can now resolve version numbers from MSBuild files?

natemcmaster: It's a custom Arcade feature. It's not generally available to the rest of the world.

natemcmaster:
Now blocked on https://github.com/dotnet/arcade/issues/2730

natemcmaster:
CI checks on Windows timeout after an hour, which is bizarre. It looks like the new installer is getting the right bits, but I wonder if it's also doing something that is causing .NET Core to slow down like 100x on Windows.

Don't merge yet - let's investigate what's going on. Might be related to https://github.com/dotnet/arcade/issues/2744

Issue #1688, opened 2019-05-08
EnvironmentVariablesConfigurationProvider should support all types of connectionstrings 


onybo:
When setting a connection string on an Azure Webapp using Azure CLI the type must be specified.
The following types are supported:
`ApiHub`, `Custom`, `DocDb`, `EventHub`, `MySql`, `NotificationHub`, `PostgreSQL`, `RedisCache`, `SQLAzure`, `SQLServer`, `ServiceBus`
[az webapp config connection-string set](https://docs.microsoft.com/en-us/cli/azure/webapp/config/connection-string?view=azure-cli-latest#az-webapp-config-connection-string-set)

It seems that [Microsoft.Extensions.Configuration.EnvironmentVariables.EnvironmentVariablesConfigurationProvider](https://github.com/aspnet/Extensions/blob/master/src/Configuration/Config.EnvironmentVariables/src/EnvironmentVariablesConfigurationProvider.cs)  only support four of these: `MySql`, `SQLAzure`, `SQLServer` and `Custom`

This means that if I use Configuration.GetConnectionString("MyServiceBusConnectionString") and I have added a connection string of type ServiceBus on my Azure Website with the same name the code will unexpectedly fail. I will have to define the type of the connectionstring in azure to be Custom for this to work.

It would be much better if `EnvironmentVariablesConfigurationProvider` and the method `AzureEnvToAppEnv` supported the same types as the Azure CLI does.

These are the environment variable prefixes that I think should be added:
APIHUBCONNSTR_
DOCDBCONNSTR_
EVENTHUBCONNSTR_
NOTIFICATIONHUBCONNSTR_
POSTGRESQLCONNSTR_
REDISCACHECONNSTR_
SERVICEBUSCONNSTR_

I would be happy to make a PR for this should you agree that this is something that should be supported.

Tratcher:
Wow, that layering is really bizarre. I'd have expected there to be a library that read from config and dealt with this, not having it embedded in the provider itself.

poke:
Do we actually use those provider names for anything? I did a quick search across Extensions, AspNetCore, and EF Core, and outside of configuration tests, the `_ProviderName` config doesn’t appear to be used. What are those for?
And why would this configuration provider decide about what database provider I use? For example, it’s just by chance that [MySqlConnector](https://github.com/mysql-net/MySqlConnector) uses the same namespace for their MySQL provider (for migration and compatibility reasons). But they didn’t *have* to do it like that.

[...]

PR #1695, opened 2019-05-09
Added support for more connection string types

onybo:
Add support for all connection string types supported by Azure CLI
Fixes issue #1688

Addresses #1688


Tratcher:
@davidfowl this is not a well layered design, we should re-think it before expanding on it.

PR #1698, opened 2019-05-09
Obsolete TestPathUtilities

natemcmaster:
First step towards completing https://github.com/aspnet/Extensions/issues/1697.

Marking this as obsolete so we can better identify where this used and make updates as described in https://github.com/aspnet/Extensions/issues/1697.

This should also help devs avoid using this in any _new_ test code.

cc @aspnet/build @HaoK  - believe this should help us identify the cause of so many tests that don't work in Helix in the AspNetCore repo.


Directory.Build.props
```diff
@@ -55,6 +55,14 @@
     <Reference Include="Moq" />
   </ItemGroup>
 
+  <!-- Always copy content items into the build output directory. -->
```
dougbu: This doesn't happen "always". It only happens for test projects and even then the `<ItemDefinitionGroup>` only provides defaults.

src/Hosting/Hosting/test/HostBuilderTests.cs
```diff
@@ -118,10 +118,8 @@ public void CanConfigureAppConfigurationAndRetrieveFromDI()
         [Fact]
         public void CanConfigureAppConfigurationFromFile()
         {
-            // Needs to look in the project directory like VS would
-            var projectDir = Path.Combine(TestPathUtilities.GetRepoRootDirectory(), "src", "Hosting", "Hosting", "test");
             var hostBuilder = new HostBuilder()
-                .UseContentRoot(projectDir)
+                .UseContentRoot(AppContext.BaseDirectory)
```
dougbu: Isn't `AppContext.BaseDirectory` normally the bin/Debug/{framework} directory?

src/Hosting/Hosting/test/HostBuilderTests.cs
```diff
@@ -118,10 +118,8 @@ public void CanConfigureAppConfigurationAndRetrieveFromDI()
         [Fact]
         public void CanConfigureAppConfigurationFromFile()
         {
-            // Needs to look in the project directory like VS would
```
Tratcher: The intent of this test has been changed. It was emulating VS behavior where these config files are not published, but now you've changed everything to only test published. Is there any way to continue to test the VS scenario?

src/Hosting/Hosting/test/HostBuilderTests.cs
natemcmaster: It's the directory that contains the test assembly.

HaoK:
The other thing you should look for is calls to UseSolutionRelativeContentRoot since we don't typically have a sln on helix, I've been working around those by dropping a dummy sln in the publish directory https://github.com/aspnet/AspNetCore/blob/6a44aca6a204055ed9f18a7ed19634f85d4994db/src/Hosting/TestHost/src/WebHostBuilderExtensions.cs#L78

src/Hosting/Hosting/test/HostBuilderTests.cs
natemcmaster: Can you clarify what the VS scenario is? If that scenario is "Visual Studio sets "Directory.GetCurrentDirectory()" to the project directory", then this test doesn't actually assert that.

src/Hosting/Hosting/test/HostBuilderTests.cs
dougbu: Then this update will change the content root significantly. That may be your intent (given the update to copy everything to build and publish directories) but could break things looking for non-`@(Content)` files.

natemcmaster: That is my intent. Those tests should update to use `<Content>`. See the description in https://github.com/aspnet/Extensions/issues/1697 about what we need to do to tests like this.

src/Hosting/Hosting/test/HostBuilderTests.cs
Tratcher: Yes, VS sets the current director to the project dir *and* does not publish config files by default.

natemcmaster: If that's what you want to test, this test needs to be completely rewritten. What it's doing is a poor proxy for that scenario. If VS decided to change the current directory, this test would still pass. If the SDK started publishing this config files, the test would still pass. Nothing in this test code is asserting the _absence_ of the file in publish output, and nothing is asserting the cwd behavior of Visual Studio.

natemcmaster: I would prefer to leave this change in. This change makes the test compatible with Helix.

Tratcher: I won't block it, but we need to keep in mind where our tests diverge from the normal user experience.

src/TestingUtils/Microsoft.AspNetCore.Testing/test/TestPathUtilitiesTest.cs
```diff
@@ -1,31 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.IO;
-using Xunit;
-
-namespace Microsoft.AspNetCore.Testing
-{
-    public class TestPathUtilitiesTest
```
Tratcher: Removing the tests seems premature when you haven't actually removed any of the usage yet.

src/Hosting/test/FunctionalTests/ShutdownTests.cs
```diff
@@ -44,8 +44,11 @@ private async Task ExecuteShutdownTest(string testName, string shutdownMechanic)
             {
                 var logger = loggerFactory.CreateLogger(testName);
 
+                // TODO refactor deployers to not depend on source code
```
Tratcher: Issue?

natemcmaster: https://github.com/aspnet/AspNetCore/issues/10268

src/TestingUtils/Microsoft.AspNetCore.Testing/test/TestPathUtilitiesTest.cs
natemcmaster: Good point. Will restore these.

natemcmaster:
Updated with PR feedback. Planning to merge soon.

Re: UseSolutionRelativeContentRoot - @HaoK thanks for the tip! This API is probably okay for our customers to use when they are writing tests, but we have unique requirements, like distributed testing, which means the API should probably not be used in our test code.

PR #1703, opened 2019-05-10
I like trailing commas in config

halter73:
I searched the Extensions repo to see if there was any discussion about removing support for trailing commas in json config. I found this from @HaoK:

> @ajcvickers tests now are all passing, the trailing comma was a big issue in the new common provider tests you added, smallest change was to run it thru JSON.NET to clean it up. Thoughts?

https://github.com/aspnet/Extensions/pull/1028#issuecomment-461177591

My thoughts are that allowing trailing commas in config is the obviously better choice. And even if it's a close call, shouldn't we lean towards not breaking people? This broke the Kestrel "SampleApp" for instance.

Was there any other discussion about removing support for trailing commas in json config when switching from JSON.NET to System.Json.Text?

### DIFF
--- a/src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
+++ b/src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
@@ -24,8 +24,14 @@ namespace Microsoft.Extensions.Configuration.Json
         {
             _data.Clear();
 
+            var jsonReaderOptions = new JsonReaderOptions
+            {
+                CommentHandling = JsonCommentHandling.Skip,
+                AllowTrailingCommas = true,
+            };
+
             using (var reader = new StreamReader(input))
-            using (JsonDocument doc = JsonDocument.Parse(reader.ReadToEnd(), new JsonReaderOptions { CommentHandling = JsonCommentHandling.Skip }))
+            using (JsonDocument doc = JsonDocument.Parse(reader.ReadToEnd(), jsonReaderOptions))
             {
                 if (doc.RootElement.Type != JsonValueType.Object)
                 {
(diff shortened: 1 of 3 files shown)


src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
```diff
@@ -24,8 +24,14 @@ internal class JsonConfigurationFileParser
         {
             _data.Clear();
 
+            var jsonReaderOptions = new JsonReaderOptions
```
HaoK: We probably need to expose this on the AddJson level if we are going to stray from the defaults

JamesNK:
> Was there any other discussion about removing support for trailing commas in json config when switching from JSON.NET to System.Json.Text?

No. In fact I pushed to get a setting added to S.J.T for explicitly for config files.

halter73:
> We probably need to expose this on the AddJson level if we are going to stray from the defaults

Hopefully that can wait for another PR. We already stray from System.Json.Text defaults for config wrt comments.

anurse:
Change the file extension to `.json5` :trollface: 

Nah, I love it. Let's do it. Trail all the commas!

davidfowl:
Did you finally run the kestrel sample?

ajcvickers:
I'm more than fine with supporting trailing commas so we don't break people. When this was discussed in the past (along with the single quotes issue) the answer was no, because the new thing was supposed to be "cleaner" or something like that.

anurse:
The new parser doesn't support them by-default, which I believe was part of the discussion around making it a more "spec-adherent" parser. However, it does have the option to turn support on, which we take advantage of. I generally agree with both approaches that have been taken here (parser should be spec-adherent by default but consumers should be as flexible as their scenario allows).

halter73:
> Did you finally run the kestrel sample?

Yes. You know you could have submitted this PR back when you saw this issue first 😉.

PR #1704, opened 2019-05-11
Use System.Text.Json in Microsoft.JSInterop

pranavkm:


.editorconfig
```diff
@@ -13,7 +13,7 @@ insert_final_newline = true
 
 [*.cs]
 indent_size = 4
-dotnet_sort_system_directives_first = true:warning
+dotnet_sort_system_directives_first = true
```
pranavkm: Having the warning stops Remove & sort in VS to stop working

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRef.cs
```diff
@@ -1,66 +1,21 @@
...
 {
     /// <summary>
-    /// Wraps a JS interop argument, indicating that the value should not be serialized as JSON
-    /// but instead should be passed as a reference.
-    ///
-    /// To avoid leaking memory, the reference must later be disposed by JS code or by .NET code.
+    /// Provides convenience methods to produce a <see cref="DotNetObjectRef{TValue}" />.
     /// </summary>
-    public class DotNetObjectRef : IDisposable
+    public static class DotNetObjectRef
```
pranavkm: Let me know what you think about this design. We don't need the type to be generic, but it avoids cast in user code when receiving it from JS.

```C#
public static void SomeJSInteropMethod(DotNetObjectRef foo)
{
   var value = ((Foo)foo.Value).DoSomething();
}
```

versus 
```C#
public static void SomeJSInteropMethod(DotNetObjectRef<Foo> foo)
{
   var value = foo.Value.DoSomething();
}
```

We also don't need this factory method, but it's a convenient way to avoid a bit of typing when creating these reference instances.

```C#
var ref = new DotNetObjectRef<Foo>(foo);
```

versus

```C#
var ref = DotNetObjectRef.Create(foo);
```

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// </summary>
+    /// <typeparam name="TValue">The type of the value to wrap.</typeparam>
+    public sealed class DotNetObjectRef<TValue> : IDotNetObjectRef, IDisposable where TValue : class
+    {
+        private long? _trackingId;
+
+        /// <summary>
+        /// This A is for meant for JSON deserialization and should not be used by user code.
+        /// </summary>
+        [Obsolete("This API is meant for JSInterop infrastructure and should not be used by user code.")]
```
pranavkm: The empty ctor and TrackingId property are meant for the serializer and are not used by any of our code. I marked them as EditorBrowseableNever and also slapped an Obsolete attribute for good measure. It's not really obsolete, but at least it further deters accidental use.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace Microsoft.JSInterop
+{
+    /// <summary>
+    /// Wraps a JS interop argument, indicating that the value should not be serialized as JSON
+    /// but instead should be passed as a reference.
+    ///
+    /// To avoid leaking memory, the reference must later be disposed by JS code or by .NET code.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the value to wrap.</typeparam>
+    public sealed class DotNetObjectRef<TValue> : IDotNetObjectRef, IDisposable where TValue : class
```
pranavkm: Here's the interesting bit:

1) During serialization, we ignore the Value property and serialize the object id with a unique key: `{ "__dotnetObjectRef": 2334 }`
2) JavaScript looks for an object with the key and turns in to a `DotNetObjectRef` instance.
3) For a call in to .NET, when the serializer sets the `TrackingId` property, we lookup the object.
4) Additionally if the dispatcher notices a top-level argument that's shaped like DotNetObjectRef, but the receiver is not an instance of this type, we throw. At least for the most common cases where a user could mess up, we produce a decent error.

src/JSInterop/Microsoft.JSInterop/src/IJSRuntime.cs
```diff
@@ -13,20 +14,10 @@ public interface IJSRuntime
...
-
-        /// <summary>
-        /// Stops tracking the .NET object represented by the <see cref="DotNetObjectRef"/>.
-        /// This allows it to be garbage collected (if nothing else holds a reference to it)
-        /// and means the JS-side code can no longer invoke methods on the instance or pass
-        /// it as an argument to subsequent calls.
-        /// </summary>
-        /// <param name="dotNetObjectRef">The reference to stop tracking.</param>
-        /// <remarks>This method is called automatically by <see cref="DotNetObjectRef.Dispose"/>.</remarks>
-        void UntrackObjectRef(DotNetObjectRef dotNetObjectRef);
```
pranavkm: I removed this API. We recommendation would be to just dispose the instance since it would do exactly the same. Is there a good reason to keep this?

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -14,23 +16,16 @@ namespace Microsoft.JSInterop
     /// </summary>
     public abstract class JSRuntimeBase : IJSRuntime
     {
+        private protected readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions
```
pranavkm: Welcome to one of the many-many `JsonSerializerOptions`

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -88,33 +83,32 @@ internal void EndInvokeDotNet(string callId, bool success, object resultOrExcept
...
-                TaskGenericsUtil.SetTaskCompletionSourceResult(tcs, resultOrException);
-            }
-            else
-            {
-                TaskGenericsUtil.SetTaskCompletionSourceException(tcs, new JSException(resultOrException.ToString()));
+                if (succeeded)
+                {
+                    var resultType = TaskGenericsUtil.GetTaskCompletionSourceResultType(tcs);
+                    var result = asyncCallResult != null ?
+                        JsonSerializer.Parse(asyncCallResult.JsonElement.GetRawText(), resultType, JsonSerializerOptions) :
```
pranavkm: There's a feature request to get the serializer to support `JsonElement` which I plan on tackling in my free time. It would avoid all these round-tripping of JsonElement (utf-8 bytes) -> string (utf-16 chars) -> Object and should be a decent perf improvement

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
```diff
@@ -12,21 +12,32 @@ public static partial class DotNetDispatcher
         [Microsoft.JSInterop.JSInvokableAttribute("DotNetDispatcher.ReleaseDotNetObject")]
         public static void ReleaseDotNetObject(long dotNetObjectId) { }
     }
-    public partial class DotNetObjectRef : System.IDisposable
+    public static partial class DotNetObjectRef
```
rynowak: shenanigans.

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
```diff
@@ -12,21 +12,32 @@ public static partial class DotNetDispatcher
...
+        public DotNetObjectRef(TValue value) { }
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
+        [System.ObsoleteAttribute("This API is meant for JSInterop infrastructure and should not be used by user code.")]
+        [System.Text.Json.Serialization.JsonPropertyNameAttribute("__dotNetObject")]
+        public long TrackingId { get { throw null; } set { } }
+        [System.Text.Json.Serialization.JsonIgnoreAttribute]
+        public TValue Value { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public void Dispose() { }
-        public void EnsureAttachedToJsRuntime(Microsoft.JSInterop.IJSRuntime runtime) { }
+        public static implicit operator TValue (Microsoft.JSInterop.DotNetObjectRef<TValue> result) { throw null; }
```
rynowak: hmmm. This gives me pause. What scenarios do you imagine this being useful for?

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
```diff
@@ -12,21 +12,32 @@ public static partial class DotNetDispatcher
...
     }
     public partial interface IJSInProcessRuntime : Microsoft.JSInterop.IJSRuntime
     {
         T Invoke<T>(string identifier, params object[] args);
     }
     public partial interface IJSRuntime
     {
-        System.Threading.Tasks.Task<T> InvokeAsync<T>(string identifier, params object[] args);
-        void UntrackObjectRef(Microsoft.JSInterop.DotNetObjectRef dotNetObjectRef);
```
rynowak: why does this go away?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -133,54 +146,96 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
-            var serializerStrategy = runtime.ArgSerializerStrategy;
-            for (var i = 0; i < suppliedArgsLength; i++)
+            // There's no direct way to say we want to deserialize as an array with heterogenous
+            // entry types (e.g., [string, int, bool]), so we need to deserialize in two phases.
+            var jsonDocument = JsonDocument.Parse(argsJson);
+            var shouldDisposeJsonDocument = true;
+            try
             {
-                if (parameterTypes[i] == typeof(JSAsyncCallResult))
+                var suppliedArgsLength = jsonDocument.RootElement.GetArrayLength();
```
rynowak: worth hardening this for the case that it's not an array? I thinking about this is terms of security since this is expose via a websocket. It's good to call out and have a clear code-path for possible invalid data cases.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -133,54 +146,96 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
+                jsonDocument.Dispose();
                 throw;
             }
+
+
+            static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
+            {
+                // Check for incorrect use of DotNetObjectRef<T> at the top level. We know it's
+                // an incorrect use if there's a object that looks like { '__dotNetObject': <some number> },
+                // but we aren't assigning to DotNetObjectRef{T}.
```
rynowak: 👍

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -133,54 +146,96 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
+                return item.Type == JsonValueType.Object &&
+                     !typeof(IDotNetObjectRef).IsAssignableFrom(parameterType) &&
+                     item.TryGetProperty(DotNetObjectRefKey, out _);
```
rynowak: it would be better to make L237 the second condition and L236 the last one. `IsAssignableFrom` is expensive.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRef.cs
rynowak: I agree on all of these points. Immutable collections does a similar thing and it's decent. @SteveSandersonMS @javiercn any thoughts?

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
rynowak: The other alternative would be to write an analyzer for it, which is more work.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// (if there are no other references to it). Once the instance is disposed, it
+        /// can no longer be used in interop calls from JavaScript code.
+        /// </summary>
+        public void Dispose()
+        {
+            if (!Disposed)
+            {
+                Disposed = true;
+                // Make sure the object was constructed correctly.
+                DotNetObjectRefManager.Current.ReleaseDotNetObject(_trackingId.Value);
```
rynowak: Comment doesn't seem to match.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -133,54 +146,96 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
+                        // We will pass the JsonDocument instance to JAsyncCallResult and make JSRuntimeBase
+                        // responsible for disposing it.
+                        shouldDisposeJsonDocument = false;
+                        // For JS async call results, we have to defer the deserialization until
+                        // later when we know what type it's meant to be deserialized as
+                        suppliedArgs[index] = new JSAsyncCallResult(jsonDocument, item);
+                    }
+                    else if (IsIncorrectDotNetObjectRefUse(item, parameterType))
+                    {
+                        throw new InvalidOperationException($"In call to '{methodIdentifier}', parameter of type {parameterType.Name} at index {(index + 1)} must be wrapped in a {nameof(DotNetObjectRef<object>)} instance.");
```
SteveSandersonMS: I suggest we clarify that it's the method signature that's at fault, not the parameter value being supplied.

Also, shouldn't we be saying `DotNetObjectRef<SomeType>`, not `DotNetObjectRef<object>`?

```suggestion
                        throw new InvalidOperationException($"In call to '{methodIdentifier}', parameter of type '{parameterType.Name}' at index {(index + 1)} must be declared as type 'DotNetObjectRef<{parameterType.Name}>' to receive the incoming value.");
```

Please let me know if I'm misunderstanding this. We don't really want to force people to use `DotNetObjectRef<object>` in all cases if that's avoidable.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -235,11 +288,11 @@ private static (MethodInfo, Type[]) GetCachedMethodInfo(string assemblyName, str
             }
         }
 
-        private static IReadOnlyDictionary<string, (MethodInfo, Type[])> ScanAssemblyForCallableMethods(string assemblyName)
+        private static Dictionary<string, (MethodInfo, Type[])> ScanAssemblyForCallableMethods(string assemblyName)
         {
             // TODO: Consider looking first for assembly-level attributes (i.e., if there are any,
             // only use those) to avoid scanning, especially for framework assemblies.
-            var result = new Dictionary<string, (MethodInfo, Type[])>();
+            var result = new Dictionary<string, (MethodInfo, Type[])>(StringComparer.Ordinal);
```
SteveSandersonMS: Well spotted!

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRef.cs
SteveSandersonMS: This makes sense and looks good!

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefManager.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                return jsRuntimeBase.ObjectRefManager;
+            }
+        }
+
+        public long TrackObject(IDotNetObjectRef dotNetObjectRef)
+        {
+            lock (_storageLock)
+            {
+                // Assign an ID only if it doesn't already have one
```
SteveSandersonMS: I know you're inheriting this logic from the earlier version, but thinking about it now, this seems wrong. It's possible for two unrelated code paths to wrap the same object instance and pass it to JS as a `DotNetObject`. The two unrelated code paths will eventually dispose the reference.

But this will lead to a conflict, won't it? After the first receipient disposed the reference, the other may still be trying to use it.

So, it seems like we should always create a new wrapper and ID on each call.

SteveSandersonMS: If this reasoning is correct, then we shouldn't need `_trackedIdsByRef` at all either.

src/JSInterop/Microsoft.JSInterop/src/IJSRuntime.cs
SteveSandersonMS: This is interesting. Previously, it was the JS side that disposed the object references. The main reason for that was that when calling back to .NET, the .NET side would just receive the wrapped object instance, not the wrapper itself, so it had no easy way to dispose the wrapper.

With the new design, the .NET side does have an easy opportunity to dispose wrappers when handling any callback, so I definitely agree it's good for .NET to be able to dispose from its own side like you've added now.

I'm not certain what scenarios remain for disposing from the JS side. Maybe there are none. Even if there are, it could be achieved by calling a method on .NET to ask it to do the disposal. So unless anyone comes up with objects, it think it's good to proceed with your updated design where disposal only happens from the .NET side.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -207,9 +262,7 @@ public static void EndInvoke(long asyncHandle, bool succeeded, JSAsyncCallResult
         [JSInvokable(nameof(DotNetDispatcher) + "." + nameof(ReleaseDotNetObject))]
         public static void ReleaseDotNetObject(long dotNetObjectId)
```
SteveSandersonMS: Should this whole `[JSInvokable]` method still exist? It looks like elsewhere, you're removing support for issuing a disposal call from the JS side.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
SteveSandersonMS: Is there a risk that using `[Obsolete]` might trigger audit warnings telling us we're supposed to remove obsolete APIs at some point?

It might be sufficient to use `[EditorBrowsable(EditorBrowsableState.Never)]` only, since AFAIK nothing bad can happen if someone ignores us and uses it anyway. They just end up with an instance whose `_trackingId` is fixed at `null` and can never do anything useful, right?

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        /// <summary>
+        /// Gets the object instance represented by this wrapper.
+        /// </summary>
+        [JsonIgnore]
+        public TValue Value
+        {
+            get;
+            // Workaround for https://github.com/dotnet/corefx/issues/37536 and https://github.com/dotnet/corefx/issues/37567
+            // Once fixed, we should make this property private.
```
SteveSandersonMS: Could you make sure there's a bug tracking this, and that it's scheduled to be done before we ship 3.0? It's important that people don't actually write to `Value` as it completely weirdifies the programming model if they can.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        internal bool Disposed { get; private set; }
+
+        /// <summary>
+        /// This constructor is for meant for JSON serialization and should not be used by user code.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [JsonPropertyName(DotNetDispatcher.DotNetObjectRefKey)]
+        [Obsolete("This API is meant for JSInterop infrastructure and should not be used by user code.")]
+        public long TrackingId
```
SteveSandersonMS: Is there a longer-term plan for the JSON serializer to support serializing nonpublic fields given suitable metadata?

If not, we might consider actually calling the property `__dotNetObject` to make it extra unappealing.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+
+        object IDotNetObjectRef.Value => Value;
+
+        /// <summary>
+        /// Implictly converts the <see cref="DotNetObjectRef{TValue}" /> instance to the wrapping value.
+        /// </summary>
+        /// <param name="result">The <see cref="DotNetObjectRef{TValue}"/>.</param>
+        public static implicit operator TValue(DotNetObjectRef<TValue> result) => result.Value;
```
SteveSandersonMS: I have mixed feelings about this. I appreciate the effort to simplify usage, but if people are using the value as if it is an instance of `TValue`, they might forget there's a wrapper at all, and hence forget to dispose it.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
SteveSandersonMS: Also, considering that `DotNetObjectRefManager.Current.ReleaseDotNetObject` already no-ops if the value isn't tracked, maybe we don't need to track the `Disposed` flag on each instance.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+            get;
+            // Workaround for https://github.com/dotnet/corefx/issues/37536 and https://github.com/dotnet/corefx/issues/37567
+            // Once fixed, we should make this property private.
+            set;
+        }
+
+        internal bool Disposed { get; private set; }
+
+        /// <summary>
+        /// This constructor is for meant for JSON serialization and should not be used by user code.
```
campersau: This does not look like a `constructor`.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+    ///
+    /// To avoid leaking memory, the reference must later be disposed by JS code or by .NET code.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the value to wrap.</typeparam>
+    public sealed class DotNetObjectRef<TValue> : IDotNetObjectRef, IDisposable where TValue : class
+    {
+        private long? _trackingId;
+
+        /// <summary>
+        /// This A is for meant for JSON deserialization and should not be used by user code.
```
campersau:
```suggestion
        /// This API is for meant for JSON deserialization and should not be used by user code.
```

or `constructor`?

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,117 @@
...
+        private void ThrowIfDisposed()
+        {
+            if (Disposed)
+            {
+                throw new ObjectDisposedException(nameof(DotNetObjectRef<TValue>));
+            }
+        }
+
+        // Used by the RefManager to mark this instance as disposed once the object is no longer being tracked.
+        void IDotNetObjectRef.SetDisposed()
```
SteveSandersonMS: Could you clarify why we have this feature? I'm not sure what bad thing would happen if we let the `DotNetObjectRefManager` deal with disposal, and the `DotNetObjectRef<T>` wouldn't need to know whether it had been disposed.

Is it to stop people from passing the value to JS after it's already been disposed? Because I'm not sure what that protects people from, as their JS code can already hold onto instances after disposal and try to use them regardless. Is there another scenario I'm missing?

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
SteveSandersonMS: Cool

src/JSInterop/Microsoft.JSInterop/src/Json/Json.cs
```diff
@@ -1,39 +0,0 @@
...
-{
-    /// <summary>
-    /// Provides mechanisms for converting between .NET objects and JSON strings for use
-    /// when making calls to JavaScript functions via <see cref="IJSRuntime"/>.
-    /// 
-    /// Warning: This is not intended as a general-purpose JSON library. It is only intended
-    /// for use when making calls via <see cref="IJSRuntime"/>. Eventually its implementation
-    /// will be replaced by something more general-purpose.
-    /// </summary>
-    public static class Json
```
SteveSandersonMS: The removal of this class is my favourite part of this PR :)

src/JSInterop/Microsoft.JSInterop/src/IJSRuntime.cs
pranavkm: We still have a way to dispose the type from JavaScript - however it's an internal implementation detail. Calling `dotNetRef.Dispose()` works exactly the way it works today, it calls a JSInvokable method that passes in the object id. We look it up and dispose it. This seemed superfluous. Given that there isn't a way for `IJsRuntime` to track or create a `DotNetObjectRef`, this seemed unnecessary.

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
pranavkm: If you prefer specifying the type name:

```C#
RowModel model = dotNetObjectRef;
```

It's not absolutely necessary, but we have some precedence e.g. https://github.com/aspnet/Extensions/blob/master/src/Primitives/src/StringValues.cs#L42-L45

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
pranavkm: See https://github.com/aspnet/Extensions/pull/1704#discussion_r283906655. We don't need this API to implement disposing the ref. Users can dispose the actual `DotNetObjectRef<T>` instance which should stop tracking it.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
pranavkm: > Also, shouldn't we be saying `DotNetObjectRef<SomeType>`, not `DotNetObjectRef<object>`?

`nameof` ignores generic parameters, so you baically see "must be wrapped in 'DotnetObjectRef'`. Switching to your recommendation would produce a much nicer error.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
pranavkm: I left the support for disposing it from client, but removed support for untracking the object by calling `IJSRuntime.UntrackObjectRef`. In .NET, you can untrack it by disposing the reference. Are there any scenarios I'm missing where `UntrackObjectRef` is useful?

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: > It might be sufficient to use [EditorBrowsable(EditorBrowsableState.Never)] only, since AFAIK nothing bad can happen if someone ignores us and uses it anyway. 

Yup, I suppose. I'll remove the `Obsolete` warnings

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: https://github.com/aspnet/AspNetCore/issues/10298

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: > Is there a longer-term plan for the JSON serializer to support serializing nonpublic fields given suitable metadata?

Not entirely sure, but we I'll all for making it unappealing 👍

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: Okies. I'll take it out

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
anurse: If we need that (and I think getting rid of a weird `Obsolete` public property is worth it) then we should file it and raise it with them.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: > Is it to stop people from passing the value to JS after it's already been disposed

Sort of. But yeah, it seems unnecessary. Cleaning it up doe

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
pranavkm: Commonized this, so it's slightly less icky.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: @anurse it's likely that we'll use converters for this once the feature lands. As of now, having an ugly property that isn't visible to user code isn't super awful.

anurse: Cool, as long as we have *a* strategy for getting on the right path again and any dependencies that strategy has are tracked I'm happy :)

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
pranavkm: This API was removed.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -276,16 +334,48 @@ private static (MethodInfo, Type[]) GetCachedMethodInfo(string assemblyName, str
...
             // In some edge cases this might force developers to explicitly call something on the
             // target assembly (from .NET) before they can invoke its allowed methods from JS.
             var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies();
-            return loadedAssemblies.FirstOrDefault(a => a.GetName().Name.Equals(assemblyName, StringComparison.Ordinal))
-                ?? throw new ArgumentException($"There is no loaded assembly with the name '{assemblyName}'.");
+            return loadedAssemblies.FirstOrDefault(a => new AssemblyKey(a).Equals(assemblyKey))
+                ?? throw new ArgumentException($"There is no loaded assembly with the name '{assemblyKey.AssemblyName}'.");
         }
+
+        private readonly struct AssemblyKey : IEquatable<AssemblyKey>
```
pranavkm: I'm working around an issue with the arcade \ xunit: https://github.com/dotnet/arcade/issues/2816. Essentially the test has two instances of JSInterop.Tests in the AppDomain. The first one is only used by xunit and the latter by the test code.  

This results in weird errors:

> [A]TestDTO cannot be cast to [B]TestDTO. Type A originates from 'Microsoft.JSInterop.Tests, Version=42.42.42.42, Culture=neutral, PublicKeyToken=adb9793829ddae60' in the context 'Default' at location '/Users/vsts/agent/2.150.3/work/1/s/artifacts/bin/Microsoft.JSInterop.Tests/Debug/netcoreapp3.0/Microsoft.JSInterop.Tests.dll'. Type B originates from 'Microsoft.JSInterop.Tests, Version=42.42.42.42, Culture=neutral, PublicKeyToken=adb9793829ddae60' in the context 'Default' at location '/Users/vsts/agent/2.150.3/work/1/s/artifacts/bin/Microsoft.JSInterop.Tests/Debug/netcoreapp3.0/Microsoft.JSInterop.Tests.dll'.

However it seems generally OK to match by Assembly identity when it's available. If you'd like I can file an issue to clean this up later or just leave this as-is

pranavkm:
🆙 📅

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -264,17 +264,14 @@ module DotNet {
...
 
-  const dotNetObjectValueFormat = /^__dotNetObject\:(\d+)$/;
+  const dotNetObjectRefKey = '__dotNetObject';
   attachReviver(function reviveDotNetObject(key: any, value: any) {
-    if (typeof value === 'string') {
-      const match = value.match(dotNetObjectValueFormat);
-      if (match) {
-        return new DotNetObject(parseInt(match[1]));
-      }
+    if (value && typeof value === 'object' && value.hasOwnProperty(dotNetObjectRefKey)) {
```
pranavkm: I need to publish this package to npm. How does one go about doing this @aspnet/build ?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
rynowak: Ah, you're trying to solve *that* problem.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
natemcmaster: Send email with more details to our team so we can make sure get the setup done right. When does the package need to ship? Where? How does it get serviced?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
SteveSandersonMS: I still haven't yet understood exactly how one now disposes it from the client. What specific JS code achieves this? When I make sense of the change enough to understand that, I can probably answer your question :)

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -183,6 +164,85 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
+
+            // There's no direct way to say we want to deserialize as an array with heterogenous
+            // entry types (e.g., [string, int, bool]), so we need to deserialize in two phases.
+            var jsonDocument = JsonDocument.Parse(argsJson);
+            var shouldDisposeJsonDocument = true;
+            try
+            {
+                if (jsonDocument.RootElement.Type != JsonValueType.Array)
+                {
+                    throw new ArgumentNullException($"Expected a JSON array but got {jsonDocument.RootElement.Type}.");
```
SteveSandersonMS: Nit: `ArgumentException` instead of `ArgumentNullException`?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -276,16 +334,48 @@ private static (MethodInfo, Type[]) GetCachedMethodInfo(string assemblyName, str
...
+                AssemblyName = assemblyName;
+            }
+
+            public Assembly Assembly { get; }
+
+            public string AssemblyName { get; }
+
+            public bool Equals(AssemblyKey other)
+            {
+                if (Assembly != null && other.Assembly != null)
```
SteveSandersonMS: Is this condition backwards by mistake? It looks like you're trying to avoid property dereferences if either one is null, but instead it *only* dereferences the properties if either one is null.

SteveSandersonMS: I mean, if `other == null`, then we'll skip the `if` block. Then on L374, evaluating `other.AssemblyName` will throw a nullref.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefManager.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            return _trackedRefsById.TryGetValue(dotNetObjectId, out var dotNetObjectRef)
+                ? dotNetObjectRef.Value
+                : throw new ArgumentException($"There is no tracked object with id '{dotNetObjectId}'. Perhaps the DotNetObjectRef instance was already disposed.", nameof(dotNetObjectId));
+            
+        }
+
+        /// <summary>
+        /// Stops tracking the specified .NET object reference.
+        /// This overload is typically invoked from JS code via JS interop.
```
SteveSandersonMS: > This overload is typically invoked from JS code via JS interop.

Nit: Is this still true? I see it *can* be called via [this](https://github.com/aspnet/Extensions/pull/1704/files#diff-5bc9491fb100bce7a35de845877af2b1R268), but I thought the more common flow now was for this to be called via `someObjectRef.Dispose()`.

Also, being extra nitty, "overload" implies there are other overloads, but this seems to be the only one.

```suggestion
        /// This may be invoked either by disposing a DotNetObjectRef in .NET code, or via JS interop by calling "dispose" on the corresponding instance in JavaScript code
```

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -0,0 +1,85 @@
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public DotNetObjectRef()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="DotNetObjectRef{TValue}" />.
+        /// </summary>
+        /// <param name="value">The value to pass by reference.</param>
+        public DotNetObjectRef(TValue value)
```
SteveSandersonMS: I know you might not want to change this around further, but might there be value in making this constructor `internal`?

That would force people to always use `DotNetObjectRef.Create<T>()`, so we know there's only one entrypoint to this whole system.

src/JSInterop/Microsoft.JSInterop/src/IJSRuntime.cs
SteveSandersonMS: OK, now I understand how the disposal-from-JS is done.

I totally agree with all your reasoning here!

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
SteveSandersonMS: Update: Now I understand it. I agree with your reasoning - we don't need `UntrackObjectRef`.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
pranavkm: These are value types, so the instances (`other`) wouldn't be null. It's just trying to see if both items have the `Assembly` property set at which point we compare those.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
pranavkm: 👍

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
natemcmaster: Opened https://github.com/aspnet/AspNetCore-Internal/issues/2541 to track the work to enable CD.

PR #1714, opened 2019-05-14
Use Microsoft.NET.Compilers.Toolset

natemcmaster:
This allows us to consume compilers that are newer than what is bundled in the .NET Core SDK. This is valuable during previews because the compiler is building new features we want to take advantage of.

This upgrades our compiler to the version bundled in arcade, which right now is 3.2.0-beta1-19253-08. This should workaround https://github.com/aspnet/AspNetCore-Internal/issues/2476

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,6 +6,10 @@
 -->
 <Project>
   <Import Project="..\version.props" />
+  <PropertyGroup Label="Arcade settings">
+    <!-- Opt-in to using the version of the Roslyn compiler bundled with Arcade. -->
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
Have we applied this fix to the other repos? What do we do for AspNetCore which is not hasn't converted yet?

natemcmaster:
Only to EFCore. This same change wouldn't work  in AspNetCore because it's not on arcade yet

JunTaoLuo:
Right, so how should we mitigate the issue in AspNetCore?

natemcmaster:
The mitigation wasn't as obvious since AspNetCore isn't using the Arcade SDK, but here is the fix: https://github.com/aspnet/AspNetCore/pull/10415

PR #1726, opened 2019-05-17
Validate pipeline logger without telemetry

chcosta:
FYI @natemcmaster 

This is testing a version of the pipeline logger which has telemetry removed.

### DIFF
--- a/global.json
+++ b/global.json
@@ -11,7 +11,7 @@
     }
   },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19260.1",
+    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19267.3",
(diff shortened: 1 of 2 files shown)


global.json
```diff
@@ -11,7 +11,7 @@
     }
   },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19260.1",
+    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19267.3",
```
dougbu: Does this work correctly when you're not updating eng/Version.Details.xml ?

chcosta:
> is this intended to be merged

No, I'm not intending to merge this (I should have made it a draft PR, my bad).  I just wanted to validate that disabling telemetry resolved the problem.

global.json
chcosta: Yes, eng/version.details.xml is just used for darc to understand / track dependencies.  global.json is what the repo actually uses for these things.

PR #1735, opened 2019-05-21
Adding link to MySQL distributed caching

ChaosEngine:
Adding link to external 3rd party distributed cache for MySQL under supervision of Pomelo Foundation.
 - implementing abstraction similar to SqlServer and StackExchangeRedis caching
 - main dependency being of https://github.com/mysql-net/MySqlConnector

Addresses https://github.com/aspnet/AspNetCore.Docs/issues/11730
and creating PR as suggested

### DIFF
--- a/src/Caching/README.md
+++ b/src/Caching/README.md
@@ -2,3 +2,6 @@ Caching
 =======
 
 Contains libraries for in-memory caching and distributed caching. Includes distributed cache implementations for in-memory, Microsoft SQL Server, and Redis.
+
+Third party caching implementations:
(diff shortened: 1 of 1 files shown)


anurse:
We don't generally list alternate implementations like this on the README (since it's only shown in GitHub). I'd suggest sending a PR to our docs to add a list of third-party implementations there. We've done that before for other components with third-party implementations.

Here's the docs for Distributed Caching: https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-3.1

PR #1739, opened 2019-05-23
Use null ref types in JSInterop

pranavkm:


src/JSInterop/Microsoft.JSInterop/src/TaskGenericsUtil.cs
```diff
@@ -25,20 +24,24 @@ public static void SetTaskCompletionSourceException(object taskCompletionSource,
...
         {
-            var getter = _cachedResultGetters.GetOrAdd(task.GetType(), taskInstanceType =>
+            var taskInstanceType = task.GetType();
+            if (!_cachedResultGetters.TryGetValue(taskInstanceType, out var getter))
             {
                 var resultType = GetTaskResultType(taskInstanceType);
-                return resultType == null
-                    ? new VoidTaskResultGetter()
-                    : (ITaskResultGetter)Activator.CreateInstance(
+                getter = resultType == null ?
```
pranavkm: A bit unrelated to reference types, but this would avoids the delegate allocation

src/JSInterop/Microsoft.JSInterop/src/JSInProcessRuntimeBase.cs
```diff
@@ -22,7 +22,9 @@ public TValue Invoke<TValue>(string identifier, params object[] args)
             var resultJson = InvokeJS(identifier, JsonSerializer.ToString(args, JsonSerializerOptionsProvider.Options));
             if (resultJson is null)
             {
+#pragma warning disable CS8653 // A default expression introduces a null value for a type parameter.
```
pranavkm: I don't know how to use nullable + generics.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -22,7 +22,9 @@ public sealed class DotNetObjectRef<TValue> : IDotNetObjectRef, IDisposable wher
         /// This API is for meant for JSON deserialization and should not be used by user code.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
+#pragma warning disable CS8618 // Non-nullable field is uninitialized.
         public DotNetObjectRef()
```
pranavkm: The compiler complains that `Value` is uninitialized here. I don't want to make it null since it's expected not to be in the usual case users are interacting with it.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -32,21 +32,21 @@ public static class DotNetDispatcher
...
-            var targetInstance = (object)null;
+            var targetInstance = (object?)null;
             if (dotNetObjectId != default)
             {
                 targetInstance = DotNetObjectRefManager.Current.FindDotNetObject(dotNetObjectId);
             }
 
             var syncResult = InvokeSynchronously(assemblyName, methodIdentifier, targetInstance, argsJson);
-            if (syncResult == null)
+            if (syncResult is null)
```
rynowak: this is gross

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -32,21 +32,21 @@ public static class DotNetDispatcher
         /// <param name="dotNetObjectId">For instance method calls, identifies the target object.</param>
         /// <param name="argsJson">A JSON representation of the parameters.</param>
         /// <returns>A JSON representation of the return value, or null.</returns>
-        public static string Invoke(string assemblyName, string methodIdentifier, long dotNetObjectId, string argsJson)
```
rynowak: Should we be validating the arguments that are expected to be non-null?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -157,7 +157,7 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
                 if (tie.InnerException != null)
                 {
                     ExceptionDispatchInfo.Capture(tie.InnerException).Throw();
-                    throw null; // unreached
+                    throw tie.InnerException; // unreached
```
rynowak: meh, this was better before.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -48,7 +50,7 @@ internal DotNetObjectRef(TValue value)
         [EditorBrowsable(EditorBrowsableState.Never)]
         public long __dotNetObject
         {
-            get => _trackingId.Value;
+            get => _trackingId ?? throw new InvalidOperationException($"{nameof(DotNetObjectRef<TValue>)} has not been initialized.");
```
rynowak: Throwing exceptions in a getter makes using the debugger kinda a pain.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefOfT.cs
```diff
@@ -70,6 +72,12 @@ public long __dotNetObject
         /// </summary>
         public void Dispose()
         {
+            if (_trackingId == null)
+            {
+                // Do nothing.
+                return;
+            }
+
```
rynowak: should we null out `_trackingId` when its disposed?

src/JSInterop/Microsoft.JSInterop/src/JSInProcessRuntimeBase.cs
rynowak: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/nullable-reference-types#generics

Emphasis:

> If a type parameter is unconstrained or has only nullable constraints, the situation is a little more complex: this means that the corresponding type argument could be either nullable or non-nullable. The safe thing to do in that situation is to treat the type parameter as both nullable and non-nullable, giving warnings when either is violated.

rynowak: I think you can probably `!default` it.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -60,9 +60,9 @@ public Task<T> InvokeAsync<T>(string identifier, params object[] args)
         /// <param name="asyncHandle">The identifier for the function invocation, or zero if no async callback is required.</param>
         /// <param name="identifier">The identifier for the function to invoke.</param>
         /// <param name="argsJson">A JSON representation of the arguments.</param>
-        protected abstract void BeginInvokeJS(long asyncHandle, string identifier, string argsJson);
+        protected abstract void BeginInvokeJS(long asyncHandle, string identifier, string? argsJson);
 
-        internal void EndInvokeDotNet(string callId, bool success, object resultOrException)
```
rynowak: Should we add null check for `callId`?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
davidfowl: Nah, this is great!

PR #1742, opened 2019-05-23
 Allow flaky attribute on class and assembly

JunTaoLuo:
Addresses https://github.com/aspnet/Extensions/issues/1568

Tested to ensure adding a flaky attribute on class prevents all tests in that class from running.

### DIFF
--- a/src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
+++ b/src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Testing.xunit
     /// </para>
     /// </example>
     [TraitDiscoverer("Microsoft.AspNetCore.Testing.xunit.FlakyTestDiscoverer", "Microsoft.AspNetCore.Testing")]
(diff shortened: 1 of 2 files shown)


src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/FlakyAttribute.cs
```diff
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Testing.xunit
     /// </para>
     /// </example>
     [TraitDiscoverer("Microsoft.AspNetCore.Testing.xunit.FlakyTestDiscoverer", "Microsoft.AspNetCore.Testing")]
-    [AttributeUsage(AttributeTargets.Method)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly)]
```
dougbu: Need a test covering at least one of the newly supported uses of the attribute

anurse: We disabled all the FlakyAttribute tests because of the noise they produced. I don't think it's a good idea to bring them back.

PR #1759, opened 2019-05-28
Allow options validation with dependencies

poke:
This adds generic overloads to `OptionBuilder.Validate` that allow validation functions with up to five service dependencies. This is analog to what was done in aspnet/Options#236 where this was added for `Configure` and `PostConfigure`.

Options validation was added with aspnet/Options#266 but no generic overloads were included there. But in the end, having dependencies for validation (e.g. to check against different configurations) can be useful.

/cc @HaoK, @tebeco

### DIFF
--- a/src/Options/test/OptionsBuilderTest.cs
+++ b/src/Options/test/OptionsBuilderTest.cs
@@ -432,6 +432,51 @@ namespace Microsoft.Extensions.Options.Tests
             Assert.Equal("target", op.Virtual);
         }
 
+        [Fact]
+        public void ValidateWithDependencies()
+        {
+            var services = new ServiceCollection()
+                .AddSingleton<Counter>()
+                .AddTransient<SomeCounterConsumer>();
+            services.AddOptions<FakeOptions>().Configure(o => o.Message = "default");
+            services.AddOptions<FakeOptions>("0dep").Configure(o => o.Message = "Foo")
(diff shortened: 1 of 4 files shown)


tebeco:
Thx @poke ;)

src/Options/test/OptionsBuilderTest.cs
```diff
@@ -432,6 +432,50 @@ public void CanValidateOptionsThatDependOnOptions()
...
+            Assert.Equal("Foo 6 7 8 9", factory.Create("4dep").Message);
+            Assert.Equal("Foo 10 11 12 13 14", factory.Create("5dep").Message);
+
+            // factory caches configures
+            Assert.Equal("Foo 1 2", factory.Create("2dep").Message);
+
+            // new factory will re-execute validator
+            factory = sp.GetRequiredService<IOptionsFactory<FakeOptions>>();
+
+            var error1 = Assert.Throws<OptionsValidationException>(() => factory.Create("1dep"));
```
HaoK: Might be worth adding a comment saying that validation fails on this second pass through due to the counter services incrementing, otherwise it takes a bit of inspection to figure out why things would fail

poke: Good point, expanded the comment to explain that.

tebeco:
thx @poke @HaoK

PR #1760, opened 2019-05-28
Make Host.StartAsync cancellable by app lifetime

poke:
This fixes aspnet/AspNetCore#10137 (which still has to be moved to this repo).

Originally, `Host.StartAsync` would only start the web host which started very quickly. With the generic host and custom hosted services, it becomes more likely that the host startup process will take longer,
especially when hosted services need to do more work during startup (e.g. migrations). This increases the need for the startup process to be cancellable.

Hosts already have a concept of being externally cancellable, e.g. when running inside of the console or as a Windows service. This is done through the host application lifetime which can gracefully shut down the
application.

This change integrates the host application lifetime into the startup process to make sure that a stopped application lifetime will also cancel the startup process. That way, the application can shut down
immediately instead of having to wait for the startup to complete in order to terminate.

/cc @anurse, @Tratcher

### DIFF
--- a/src/Hosting/Hosting/src/Internal/Host.cs
+++ b/src/Hosting/Hosting/src/Internal/Host.cs
@@ -39,15 +39,18 @@ namespace Microsoft.Extensions.Hosting.Internal
         {
             _logger.Starting();
 
-            await _hostLifetime.WaitForStartAsync(cancellationToken);
+            using var combinedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _applicationLifetime.ApplicationStopping);
+            var combinedCancellationToken = combinedCancellationTokenSource.Token;
 
-            cancellationToken.ThrowIfCancellationRequested();
+            await _hostLifetime.WaitForStartAsync(combinedCancellationToken);
+
+            combinedCancellationToken.ThrowIfCancellationRequested();
             _hostedServices = Services.GetService<IEnumerable<IHostedService>>();
 
             foreach (var hostedService in _hostedServices)
             {
                 // Fire IHostedService.Start
-                await hostedService.StartAsync(cancellationToken).ConfigureAwait(false);
+                await hostedService.StartAsync(combinedCancellationToken).ConfigureAwait(false);
             }
 
             // Fire IHostApplicationLifetime.Started
--- a/src/Hosting/Hosting/test/Internal/HostTests.cs
+++ b/src/Hosting/Hosting/test/Internal/HostTests.cs
@@ -267,6 +267,92 @@ namespace Microsoft.Extensions.Hosting.Internal
         }
 
         [Fact]
+        public async Task CancellableStart_CancelledByApplicationLifetime()
+        {
+            var hostedService = new AsyncFakeHostedService();
+            var builder = CreateBuilder().ConfigureServices(services =>
+            {
+                services.AddSingleton<IHostedService>(hostedService);
+            });
+
+            using (var host = builder.Build())
+            {
+                var applicationLifetime = host.Services.GetService<IHostApplicationLifetime>();
+                var startTask = host.StartAsync();
+
+                // stop application
+                applicationLifetime.StopApplication();
+
+                // complete start task
+                hostedService.ContinueStart();
+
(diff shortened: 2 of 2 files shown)


src/Hosting/Hosting/test/Internal/HostTests.cs
```diff
@@ -731,9 +818,6 @@ public async Task HostedServiceCanInjectApplicationLifetime()
                    })
                    .Build())
             {
-                var lifetime = host.Services.GetRequiredService<IHostApplicationLifetime>();
-                lifetime.StopApplication();
```
poke: I’m not sure why this was there in this test. According to the name, this just checks that the lifetime can be injected into services, which will not require the application to be stopped.

src/Hosting/Hosting/test/Internal/HostTests.cs
```diff
@@ -709,13 +796,13 @@ public async Task HostStopApplicationDoesNotFireStopOnHostedService()
...
-
                 await host.StartAsync();
 
                 Assert.Equal(0, stoppingCalls);
                 Assert.Equal(0, disposingCalls);
+
+                var lifetime = host.Services.GetRequiredService<IHostApplicationLifetime>();
+                lifetime.StopApplication();
```
poke: Same as below, not sure what the original purpose of this was, but I left it in here to make sure that the `StopApplication` will not call the `StopAsync` of the hosted service.

src/Hosting/Hosting/src/Internal/Host.cs
```diff
@@ -39,19 +39,24 @@ public async Task StartAsync(CancellationToken cancellationToken = default)
         {
             _logger.Starting();
 
-            await _hostLifetime.WaitForStartAsync(cancellationToken);
+            using (var combinedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _applicationLifetime.ApplicationStopping))
```
Tratcher: Can we use C#8 in Extensions?
```suggestion
            using var combinedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _applicationLifetime.ApplicationStopping);
```
Less `{ }`

src/Hosting/Hosting/test/Internal/HostTests.cs
```diff
@@ -266,6 +266,93 @@ public async Task StartCanBeCancelled()
...
+            var hostedService = new AsyncFakeHostedService();
+            var builder = CreateBuilder().ConfigureServices(services =>
+            {
+                services.AddSingleton<IHostedService>(hostedService);
+            });
+
+            using (var host = builder.Build())
+            {
+                var applicationLifetime = host.Services.GetService<IHostApplicationLifetime>();
+                var startTask = Task.Run(() => host.StartAsync());
```
Tratcher: Task.Run isn't needed here, AsyncFakeHostedService.StartAsync doesn't block sync.

src/Hosting/Hosting/test/Internal/HostTests.cs
Tratcher: Move above the Asserts?

src/Hosting/Hosting/src/Internal/Host.cs
poke: I had that in a draft and then opted for a normal using block… 😅 

I don’t have a preference for either though, so I can change that back.

src/Hosting/Hosting/test/Internal/HostTests.cs
poke: Oh right, that’s a left-over from an early version where I was using that sync `ManualResetEvent` that the other tests use.

Tratcher:
Thanks

PureKrome:
Thanks @poke for the PR and the @Tratcher and @anurse for a quick turn around. Awesome effort :)

Just to confirm @Tratcher - this will be available in `Preview6` to test/check?

Tratcher:
@PureKrome correct. It should be available in the Master preview6 nightly SDK some time tomorrow. https://github.com/dotnet/core-sdk

PR #1776, opened 2019-06-02
AddHostedService overload with factory

BEagle1984:
- Added overload to AddHostedService extension method to allow construction via factory method


BEagle1984:
The build pipeline is failing some checks but it doesn't look like they are related to my commit. Please let me know if otherwise.

Tratcher:
Thanks

AgSync-Aaron:
I was about to submit an issue asking for this. This is going to make things easier to use the generic host. Thank you!

PR #1777, opened 2019-06-03
ConsoleLogger: add support for systemd '<pri>message' format

tmds:
Implements https://github.com/aspnet/Extensions/issues/647, https://github.com/aspnet/Extensions/issues/650

@glennc @davidfowl @anurse ptal

### DIFF
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -172,7 +172,8 @@ namespace Microsoft.Extensions.Logging.Console
 
                 // Queue log message
                 _queueProcessor.EnqueueMessage(new LogMessageEntry(
-                    message: logBuilder.ToString()
+                    message: logBuilder.ToString(),
+                    logAsError: logLevel >= Options.LogToStandardErrorThreshold
                 ));
 
                 void AppendAndReplaceNewLine(StringBuilder sb, string message)
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -73,120 +73,20 @@ namespace Microsoft.Extensions.Logging.Console
                 logBuilder = new StringBuilder();
             }
 
+            LogMessageEntry entry;
             if (Options.Format == ConsoleLoggerFormat.Default)
             {
-                // Example:
-                // INFO: ConsoleApp.Program[10]
-                //       Request received
-
-                var logLevelColors = GetLogLevelConsoleColors(logLevel);
-                var logLevelString = GetLogLevelString(logLevel);
-                // category and event id
-                logBuilder.Append(_loglevelPadding);
-                logBuilder.Append(logName);
-                logBuilder.Append("[");
-                logBuilder.Append(eventId);
-                logBuilder.AppendLine("]");
-
-                // scope information
-                GetScopeInformation(logBuilder, multiLine: true);
-
-                if (!string.IsNullOrEmpty(message))
-                {
-                    // message
-                    logBuilder.Append(_messagePadding);
-
-                    var len = logBuilder.Length;
-                    logBuilder.AppendLine(message);
-                    logBuilder.Replace(Environment.NewLine, _newLineWithMessagePadding, len, message.Length);
-                }
-
-                // Example:
-                // System.InvalidOperationException
-                //    at Namespace.Class.Function() in File:line X
-                if (exception != null)
-                {
-                    // exception message
-                    logBuilder.AppendLine(exception.ToString());
-                }
-
-                var timestampFormat = Options.TimestampFormat;
-                // Queue log message
-                _queueProcessor.EnqueueMessage(new LogMessageEntry(
-                    message: logBuilder.ToString(),
-                    timeStamp: timestampFormat != null ? DateTime.Now.ToString(timestampFormat) : null,
-                    levelString: logLevelString,
-                    levelBackground: logLevelColors.Background,
-                    levelForeground: logLevelColors.Foreground,
-                    messageColor: DefaultConsoleColor,
-                    logAsError: logLevel >= Options.LogToStandardErrorThreshold
-                ));
+                entry = CreateLogMessageDefault(logBuilder, logLevel, logName, eventId, message, exception);
             }
             else if (Options.Format == ConsoleLoggerFormat.Systemd)
             {
-                // systemd reads messages from standard out line-by-line in a '<pri>message' format.
-                // newline characters are treated as message delimiters, so we must replace them.
-                // Messages longer than the journal LineMax setting (default: 48KB) are cropped.
-                // Example:
-                // <6>ConsoleApp.Program[10] Request received
-
-                // loglevel
-                var logLevelString = GetSyslogSeverityString(logLevel);
-                logBuilder.Append(logLevelString);
-
-                // timestamp
-                var timestampFormat = Options.TimestampFormat;
-                if (timestampFormat != null)
-                {
-                    logBuilder.Append(DateTime.Now.ToString(timestampFormat));
-                }
-
-                // category and event id
-                logBuilder.Append(logName);
-                logBuilder.Append("[");
-                logBuilder.Append(eventId);
-                logBuilder.Append("]");
-
-                // scope information
-                GetScopeInformation(logBuilder, multiLine: false);
-
-                // message
-                if (!string.IsNullOrEmpty(message))
-                {
-                    logBuilder.Append(' ');
-                    // message
-                    var len = logBuilder.Length;
-                    AppendAndReplaceNewLine(logBuilder, message);
-                }
-
-                // exception
-                // System.InvalidOperationException at Namespace.Class.Function() in File:line X
-                if (exception != null)
-                {
-                    logBuilder.Append(' ');
-                    AppendAndReplaceNewLine(logBuilder, exception.ToString());
-                }
-
-                // newline delimiter
-                logBuilder.Append(Environment.NewLine);
-
-                // Queue log message
-                _queueProcessor.EnqueueMessage(new LogMessageEntry(
-                    message: logBuilder.ToString(),
-                    logAsError: logLevel >= Options.LogToStandardErrorThreshold
-                ));
-
-                void AppendAndReplaceNewLine(StringBuilder sb, string message)
-                {
-                    var len = sb.Length;
-                    sb.Append(message);
-                    sb.Replace(Environment.NewLine, " ", len, message.Length);
-                }
+                entry = CreateLogMessageSystemd(logBuilder, logLevel, logName, eventId, message, exception);
             }
             else
             {
                 throw new ArgumentOutOfRangeException(nameof(Options.Format));
             }
+            _queueProcessor.EnqueueMessage(entry);
 
             logBuilder.Clear();
             if (logBuilder.Capacity > 1024)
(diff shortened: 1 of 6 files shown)


tmds:
Tests pass except for:

> Generated code is not up to date. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)

Running `dotnet msbuild /t:GenerateReferenceSource` gives me an error:

> /tmp/tmp63fcf0c47e6547aba2fa60c16a931ca7.exec.cmd: line 15: TargetFramework: command not found

tmds:
> /tmp/tmp63fcf0c47e6547aba2fa60c16a931ca7.exec.cmd: line 15: TargetFramework: command not found

This leads to:
```diff
--- a/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.csproj
+++ b/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.csproj
@@ -1,12 +1,14 @@
+
 <!-- This file is automatically generated. -->
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
   </PropertyGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition=" == netstandard2.0">
     <Compile Include="Microsoft.Extensions.Logging.Console.netstandard2.0.cs" />
     <Reference Include="Microsoft.Extensions.Configuration.Abstractions"  />
     <Reference Include="Microsoft.Extensions.Logging.Configuration"  />
     <Reference Include="Microsoft.Extensions.Logging"  />
   </ItemGroup>
 </Project>
+
```

tmds:
Windows Release build failed:

> The process cannot access the file 'F:\workspace\_work\1\s\artifacts\SymStore\Release\Microsoft.Extensions.Configuration.CommandLine\netstandard2.0\Microsoft.Extensions.Configuration.CommandLine.pdb' because it is being used by another process.
F:\workspace\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19274.6\tools\SymStore.targets(61,5): error MSB3073: The command ""F:\workspace\_work\1\s\.packages\microsoft.diasymreader.pdb2pdb\1.1.0-beta1-62506-02\tools\Pdb2Pdb.exe" "F:\workspace\_work\1\s\artifacts\bin\Microsoft.Extensions.Configuration.CommandLine\Release\netstandard2.0\Microsoft.Extensions.Configuration.CommandLine.dll" /out "F:\workspace\_work\1\s\artifacts\SymStore\Release\Microsoft.Extensions.Configuration.CommandLine\netstandard2.0\Microsoft.Extensions.Configuration.CommandLine.pdb" /verbose /srcsvrvar SRC_INDEX=public" exited with code 2. [F:\workspace\_work\1\s\src\Configuration\Config.CommandLine\src\Microsoft.Extensions.Configuration.CommandLine.csproj]
##[error].packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19274.6\tools\SymStore.targets(61,5): error MSB3073: The command ""F:\workspace\_work\1\s\.packages\microsoft.diasymreader.pdb2pdb\1.1.0-beta1-62506-02\tools\Pdb2Pdb.exe" "F:\workspace\_work\1\s\artifacts\bin\Microsoft.Extensions.Configuration.CommandLine\Release\netstandard2.0\Microsoft.Extensions.Configuration.CommandLine.dll" /out "F:\workspace\_work\1\s\artifacts\SymStore\Release\Microsoft.Extensions.Configuration.CommandLine\netstandard2.0\Microsoft.Extensions.Configuration.CommandLine.pdb" /verbose /srcsvrvar SRC_INDEX=public" exited with code 2.

anurse:
> Running `dotnet msbuild /t:GenerateReferenceSource` gives me an error:
> 
> > /tmp/tmp63fcf0c47e6547aba2fa60c16a931ca7.exec.cmd: line 15: TargetFramework: command not found

paging @aspnet/build :)

anurse:
> The process cannot access the file 'F:\workspace_work\1\s\artifacts\SymStore\Release\Microsoft.Extensions.Configuration.CommandLine\netstandard2.0\Microsoft.Extensions.Configuration.CommandLine.pdb' because it is being used by another process.

Unfortunately this is a known problem with the `pdb2pdb` tool we use to convert symbols between formats. Rerunning is likely to flush this out.

anurse:
> /tmp/tmp63fcf0c47e6547aba2fa60c16a931ca7.exec.cmd

That's super weird... I *think* that's the Windows batch script MSBuild generates to run `Exec` tasks. It shouldn't be doing that at all on non-Windows... @tmds can you repro this consistently? By chance, is the temp file still hanging around and could you `cat` it?

anurse:
@BrennanConroy for context, @tmds is our champion over at Red Hat helping make .NET awesome on Linux :). We want to add an option to the console logger to support flowing logs to `syslog` (`systemd` automatically scrapes stdout and uses this format to apply prioritization to the messages).

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -73,53 +73,121 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
-            if (!string.IsNullOrEmpty(message))
+            if (Options.Format == ConsoleLoggerFormat.Default)
             {
-                // message
-                logBuilder.Append(_messagePadding);
+                // Example:
+                // INFO: ConsoleApp.Program[10]
+                //       Request received
+
+                var logLevelColors = GetLogLevelConsoleColors(logLevel);
```
anurse: Could we put the two formats into separate methods? i.e. `BuildDefaultMessage`/`BuildSystemdMessage`. Have each method return a `LogMessageEntry` and the outer code queues it.

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -73,53 +73,121 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
+                    levelString: logLevelString,
+                    levelBackground: logLevelColors.Background,
+                    levelForeground: logLevelColors.Foreground,
+                    messageColor: DefaultConsoleColor,
+                    logAsError: logLevel >= Options.LogToStandardErrorThreshold
+                ));
             }
+            else if (Options.Format == ConsoleLoggerFormat.Systemd)
+            {
+                // systemd reads messages from standard out line-by-line in a '<pri>message' format.
```
anurse: Is there a reference doc you could link here?

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -156,6 +224,26 @@ private static string GetLogLevelString(LogLevel logLevel)
             }
         }
 
+        private static string GetSyslogSeverityString(LogLevel logLevel)
+        {
+            switch (logLevel)
```
anurse: Would be nice to have a comment referencing a source for what these priority levels "mean".

src/Logging/Logging.Console/src/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.Console
+{
+    public enum ConsoleLoggerFormat
```
anurse: Let's start this off right with some `///` doc comments describing the enum and both options.

src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
```diff
@@ -7,6 +7,7 @@ public class ConsoleLoggerOptions
     {
         public bool IncludeScopes { get; set; }
         public bool DisableColors { get; set; }
+        public ConsoleLoggerFormat Format { get; set; }
```
anurse: Similar doc comment suggestion here. We're trying to get new types/members off on the right foot with `///` doc comments (even though we have a bunch of gaps in existing ones 😦)

src/Logging/Logging.Console/src/ConsoleLogger.cs
tmds: http://0pointer.de/blog/projects/journal-submit.html printf section is the best I found. It has the same info as the comment.

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -73,6 +73,31 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
+            {
+                entry = CreateLogMessageDefault(logBuilder, logLevel, logName, eventId, message, exception);
+            }
+            else if (Options.Format == ConsoleLoggerFormat.Systemd)
+            {
+                entry = CreateLogMessageSystemd(logBuilder, logLevel, logName, eventId, message, exception);
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException(nameof(Options.Format));
```
gfoidl: Should this be an `InvalidOperationException` instead?
So that it indicates a wrong option is passed in.

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -73,6 +73,31 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
                 logBuilder = new StringBuilder();
             }
 
+            LogMessageEntry entry;
+            if (Options.Format == ConsoleLoggerFormat.Default)
+            {
+                entry = CreateLogMessageDefault(logBuilder, logLevel, logName, eventId, message, exception);
```
gfoidl: I would name these `CreateDefaultLogMessage` and `CreateSystemdLogMessage`.

src/Logging/Logging.Console/src/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// <summary>
+        /// Default format.
+        /// </summary>
+        Default,
+        /// <summary>
+        /// systemd '&lt;pri&gt;message' format.
+        /// </summary>
+        Systemd
+    }
+}
```
gfoidl: Nit: newline

src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
```diff
@@ -9,12 +9,17 @@ public class ConsoleLoggerOptions
         public bool DisableColors { get; set; }
 
         /// <summary>
-        /// Gets or sets value indicating the minimum level of messaged that would get written to  <c>Console.Error</c>.
+        /// Gets or sets log message format. Defaults to <c>ConsoleLoggerFormat.Default</c>.
+        /// </summary>
+        public ConsoleLoggerFormat Format { get; set; }
```
gfoidl: I see it defaults to `Default`, but to make it more clear I'd prefer to set it explictely. 
```suggestion
        public ConsoleLoggerFormat Format { get; set; } = ConsoleLoggerFormat.Default;
```

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -109,23 +134,77 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
+
+            return new LogMessageEntry(
                 message: logBuilder.ToString(),
                 logAsError: logLevel >= Options.LogToStandardErrorThreshold
-            ));
+            );
 
-            logBuilder.Clear();
-            if (logBuilder.Capacity > 1024)
+            void AppendAndReplaceNewLine(StringBuilder sb, string message)
```
```suggestion
            static void AppendAndReplaceNewLine(StringBuilder sb, string message)
```
Just to be on the safe side for not capturing any locals from the parent (i.e. no DisplayClass (which will be a struct :wink:)).

src/Logging/Logging.Console/src/ConsoleLogger.cs
davidfowl: I don't think we should throw here. It's way too late. Throw on startup so random log statements don't start throwing.

src/Logging/Logging.Console/src/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Format of ConsoleLogger messages.
+    /// </summary>
+    public enum ConsoleLoggerFormat
+    {
+        /// <summary>
+        /// Default format.
+        /// </summary>
+        Default,
+        /// <summary>
+        /// systemd '&lt;pri&gt;message' format.
```
anurse: I think we can just describe what it does and leave the specific format to other docs. If you wanted to put in a `<remarks>` section about it that would be fine.

```suggestion
        /// Produces messages in a format suitable for systemd services.
```

src/Logging/Logging.Console/src/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Extensions.Logging.Console
+{
+    /// <summary>
+    /// Format of ConsoleLogger messages.
+    /// </summary>
+    public enum ConsoleLoggerFormat
+    {
+        /// <summary>
+        /// Default format.
```
anurse: Minor word-smithing to make it a little more prosaic :).

```suggestion
        /// Produces messages in a default format, suitable for console output.
```

src/Logging/Logging.Console/src/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.Console
+{
+    /// <summary>
+    /// Format of ConsoleLogger messages.
```
anurse: Let's get that hyperlinking goodness ;)

```suggestion
    /// Format of <see cref="ConsoleLogger" /> messages.
```

src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
```diff
@@ -9,12 +9,17 @@ public class ConsoleLoggerOptions
         public bool DisableColors { get; set; }
 
         /// <summary>
-        /// Gets or sets value indicating the minimum level of messaged that would get written to  <c>Console.Error</c>.
+        /// Gets or sets log message format. Defaults to <c>ConsoleLoggerFormat.Default</c>.
```
anurse: `<see>` should do the needful here with hyperlinks and all.

```suggestion
        /// Gets or sets log message format. Defaults to <see cref="ConsoleLoggerFormat.Default" />.
```

src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
anurse: Yeah, I'd agree. When the default is super-obvious (`int` being `0`, `bool` being `false`, etc.) I don't like initializing it, but here the default is only due to the `Default` value being first in the list.

src/Logging/Logging.Console/src/ConsoleLogger.cs
tmds: I've moved the check to the Options class. I've kept the exception in ConsoleLogger to get rid of the compiler error that entry is not assigned.

tmds:
Thanks for reviewing! I've addressed the comments.

src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
```diff
@@ -1,17 +1,32 @@
...
-        /// Gets or sets log message format. Defaults to <c>ConsoleLoggerFormat.Default</c>.
+        /// Gets or sets log message format. Defaults to <see cref="ConsoleLoggerFormat.Default" />.
         /// </summary>
-        public ConsoleLoggerFormat Format { get; set; }
+        public ConsoleLoggerFormat Format
+        {
+            get => _format;
+            set
+            {
+                if (!Enum.IsDefined(typeof(ConsoleLoggerFormat), value))
```
davidfowl: You could do:

if (value < Default || value > ConsoleLoggerFormat.Systemd)

tmds:
> /tmp/tmp63fcf0c47e6547aba2fa60c16a931ca7.exec.cmd: line 15: TargetFramework: command not found

@anurse this is a unix script, but things go wrong with the escaping (search for `'`):
```
#!/bin/sh
echo '
<Project Sdk=\"Microsoft.NET.Sdk\">
  <PropertyGroup>
    <TargetFrameworks>netstandard2.0</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup Condition=\"'$(TargetFramework)' == 'netstandard2.0'\">
    <Compile Include=\"Microsoft.Extensions.Logging.Console.netstandard2.0.cs\" />
    <Reference Include=\"Microsoft.Extensions.Configuration.Abstractions\"  />
    <Reference Include=\"Microsoft.Extensions.Logging.Configuration\"  />
    <Reference Include=\"Microsoft.Extensions.Logging\"  />
  </ItemGroup>
</Project>
' > /home/tmds/repos/AspNetExtensions/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.csproj
```

davidfowl:
@tmds Look at https://github.com/aspnet/Extensions/blob/master/docs/ReferenceAssemblies.md

src/Logging/Logging.Console/src/ConsoleLogger.cs
davidfowl: This code should be change now to just assert that the right format was set instead of throwing.

tmds: I'm not sure how you want to express the assert in code?

davidfowl: In the beginning of the method.

`Debug.Assert(Options.Format >= Default && Options.Format <= Systemd)`

BrennanConroy:
Thanks @tmds !

PR #1781, opened 2019-06-03
Options validation exception should use the failures for the message

HaoK:
Fixes https://github.com/aspnet/Extensions/issues/979

### DIFF
--- a/src/Options/DataAnnotations/src/DataAnnotationValidateOptions.cs
+++ b/src/Options/DataAnnotations/src/DataAnnotationValidateOptions.cs
@@ -48,10 +48,12 @@ namespace Microsoft.Extensions.Options
                     return ValidateOptionsResult.Success;
                 }
 
-                return ValidateOptionsResult.Fail(String.Join(Environment.NewLine,
-                    validationResults.Select(r => "DataAnnotation validation failed for members " +
-                        String.Join(", ", r.MemberNames) +
-                        " with the error '" + r.ErrorMessage + "'.")));
+                var errors = new List<string>();
+                foreach (var r in validationResults)
+                {
+                    errors.Add($"DataAnnotation validation failed for members: '{String.Join(",", r.MemberNames)}' with the error: '{r.ErrorMessage}'.");
+                }
+                return ValidateOptionsResult.Fail(errors);
             }
 
             // Ignored if not validating this instance.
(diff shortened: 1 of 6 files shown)


HaoK:
Whoops you were too fast reviewing this PR for me @ajcvickers I also added the fix for https://github.com/aspnet/Extensions/issues/1299 to this

src/Options/DataAnnotations/src/DataAnnotationValidateOptions.cs
```diff
@@ -48,10 +48,12 @@ public ValidateOptionsResult Validate(string name, TOptions options)
...
                 }
 
-                return ValidateOptionsResult.Fail(String.Join(Environment.NewLine,
-                    validationResults.Select(r => "DataAnnotation validation failed for members " +
-                        String.Join(", ", r.MemberNames) +
-                        " with the error '" + r.ErrorMessage + "'.")));
+                var errors = new List<string>();
+                foreach (var r in validationResults)
+                {
+                    errors.Add($"DataAnnotation validation failed for members: '{String.Join(",", r.MemberNames)}' with the error: '{r.ErrorMessage}'.");
```
ajcvickers: @blowdart Is it okay to have actual validation failure messages in exception text? Might not these messages have sensitive data?

blowdart: Ugh. No way to know.

ajcvickers: Agreed. So can they go in exception text or not?

HaoK: Its pretty useless if we don't show the actual validation error, that's kind of the whole point for this change, to put the validation error in their face...

HaoK: Otherwise we might as well just keep the old behavior which is you get an OptionsValidationException with the Failures inside and hard to discover

HaoK: Although I guess for data annotations specifically, we were already leaking the exception message, so maybe we need to resolve that either way

ajcvickers: Agreed it is unfriendly. Totally favor of this changes if @blowdart says its okay.

blowdart: PG decision, is the risk worth the rewards? Arthur seems to think yes.

ajcvickers: If the advise is anything other than, "no", then I say "yes!". But I'm going to make sure I keep track that you gave an implicit yes.

PR #1784, opened 2019-06-03
Fix generic IEnumerable callsite building 

pakrym:
Fixes: https://github.com/aspnet/Extensions/issues/1775

When we build call site for generic service inside IEnumerable (i.e. `IEnumerable<IFakeService<Something>>`) we have to iterate over all descriptors to try to find all applicable open-generic closures (in case there are registrations for `IFakeService<>`). 

The bug is that we incremented the index of cache location every time even if descriptor didn't represent a closure of requested type resulting in item inside IEnumerable being cached with a different key than standalone resolution.


pakrym:
You might want to backport this fix to 2.2.

anurse:
I'm going to try out the `auto-merge` label...

brace yourself

anurse:
I... don't think it's working...

PR #1796, opened 2019-06-05
ActivatorUtilities.CreateInstance: allocation optimization

### LINKED ISSUE #693: Improve performance of ActivatorUtilities.CreateInstance
Currently, it's 30x slower than invoking factory delegate. 
While it could never reach the performance of invoking the factory delegate there is a lot of possible improvements (especially around allocations) in `ActivatorUtilities.CreateInstance`

Yves57:
Some optimizations to try to fix #693.

Before:
```
|          Method |        Mean |     Error |    StdDev |         Op/s |  Gen 0 | Allocated |
|---------------- |------------:|----------:|----------:|-------------:|-------:|----------:|
|  CreateInstance | 1,317.58 ns | 5.6777 ns | 4.7411 ns |    758,965.3 | 0.0267 |     784 B |
```

After:
```
|          Method |        Mean |      Error |     StdDev |         Op/s |  Gen 0 | Allocated |
|---------------- |------------:|-----------:|-----------:|-------------:|-------:|----------:|
|  CreateInstance | 1,275.97 ns | 26.4718 ns | 44.9511 ns |    783,719.0 | 0.0134 |     448 B |
```

### DIFF
--- a/src/Shared/src/ActivatorUtilities/ActivatorUtilities.cs
+++ b/src/Shared/src/ActivatorUtilities/ActivatorUtilities.cs
@@ -43,43 +43,45 @@ namespace Microsoft.Extensions.Internal
             int bestLength = -1;
             var seenPreferred = false;
 
-            ConstructorMatcher bestMatcher = null;
+            ConstructorMatcher bestMatcher = default;
 
             if (!instanceType.GetTypeInfo().IsAbstract)
             {
                 foreach (var constructor in instanceType
                     .GetTypeInfo()
-                    .DeclaredConstructors
-                    .Where(c => !c.IsStatic && c.IsPublic))
+                    .DeclaredConstructors)
                 {
-                    var matcher = new ConstructorMatcher(constructor);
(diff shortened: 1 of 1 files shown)


src/Shared/src/ActivatorUtilities/ActivatorUtilities.cs
```diff
@@ -327,18 +329,16 @@ private static bool TryCreateParameterMap(ParameterInfo[] constructorParameters,
             return true;
         }
 
-        private class ConstructorMatcher
+        private struct ConstructorMatcher
         {
             private readonly ConstructorInfo _constructor;
             private readonly ParameterInfo[] _parameters;
             private readonly object[] _parameterValues;
-            private readonly bool[] _parameterValuesSet;
```
Yves57: `GetTypeInfo().IsAssignableFrom(null)` returns `false` (around line 357), so I don't see why we need `parameterValuesSet`.

anurse:
Thoughts @pakrym ? @davidfowl ?

pakrym:
Looks reasonable.

anurse:
Thanks for the contribution @Yves57 !

PR #1800, opened 2019-06-05
Added overload to UseWindowsService to configure EventLog settings.

erikmf12:
Added an overload to `Microsoft.Extensions.Hosting.WindowsServiceLifetimeHostBuilderExtensions.UseWindowsService()` that accepts an `Action<EventLogSettings>` and passes that action to the call to logging.AddEventLog().

This would allow developers to configure what the event log name will be without needing a call to .ConfigureLogging().

### DIFF
--- a/src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
+++ b/src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
@@ -50,5 +50,32 @@ namespace Microsoft.Extensions.Hosting
 
             return hostBuilder;
         }
+
+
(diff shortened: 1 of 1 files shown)


Tratcher:
Can't they just call `services.Configure<EventLogSettings>(...)`? I don't think we need this extension. @glennc

anurse:
Yeah, I'd be concerned about "coupling" `UseWindowsServices` to the EventLog logger. We do enable EventLog by default, but that's not something I want to couple in to the actual `UseWindowsServices` method.

Perhaps the docs on building Windows Services should include an example like @Tratcher referenced.

davidfowl:
Agreed

PR #1802, opened 2019-06-06
Resolve the JSInterop call if deserialization fails

pranavkm:


src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -173,10 +173,12 @@ private static object[] ParseArguments(string methodIdentifier, string argsJson,
 
             // There's no direct way to say we want to deserialize as an array with heterogenous
             // entry types (e.g., [string, int, bool]), so we need to deserialize in two phases.
-            var jsonDocument = JsonDocument.Parse(argsJson);
+            JsonDocument jsonDocument = default;
             var shouldDisposeJsonDocument = true;
             try
             {
+                jsonDocument = JsonDocument.Parse(argsJson);
```
BrennanConroy: I think you should be able to use the reader + serializer now instead of `JsonDocument` for better perf.

pranavkm: Curious, how does it work?

BrennanConroy: We could paired program a little if you'd like, or you could try it out on your own and use https://github.com/aspnet/AspNetCore/pull/10733 as a reference

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -94,10 +94,17 @@ internal void EndInvokeJS(long asyncHandle, bool succeeded, JSAsyncCallResult as
...
+                    try
+                    {
+                        var result = asyncCallResult != null ?
+                            JsonSerializer.Parse(asyncCallResult.JsonElement.GetRawText(), resultType, JsonSerializerOptionsProvider.Options) :
+                            null;
+                        TaskGenericsUtil.SetTaskCompletionSourceResult(tcs, result);
+                    }
+                    catch (Exception exception)
+                    {
+                        TaskGenericsUtil.SetTaskCompletionSourceException(tcs, new JSException(exception.Message, exception));
```
rynowak: Should this have a different message since we're wrapping it? Like `An exception occurred executing JS interop. See InnerException for details.`

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
pranavkm: Tracking via https://github.com/aspnet/AspNetCore/issues/10988

SteveSandersonMS:
It looks like you've increased the `try`/`catch` coverage in two places:

1. In `ParseArguments`, around `JsonDocument.Parse` (https://github.com/aspnet/Extensions/pull/1802/files#diff-5bc9491fb100bce7a35de845877af2b1R180)
2. In `EndInvokeJS`, around **both** `JsonSerializer.Parse` and `TaskGenericsUtil.SetTaskCompletionSourceResult` (https://github.com/aspnet/Extensions/pull/1802/files#diff-db9e84190c0e7d356a0a7caf51154b9bR97)

So my questions are:

* Do these represent two different failure cases? If so, could we have a unit test for each of them? There's only one new unit test so I'm guessing it only covers one of the scenarios.
* In the second of those two, since this now gets triggered by two different kinds of failures (one where JSON parsing fails, the other where the `SetTaskCompletionSourceResult` causes a synchronous exception), should we also have a further test to represent whichever one of these is not already in the existing unit test?

Sorry if that seems fussy, but since you have this in you mind right now anyway, hopefully it's not too bad to clarify those points (possibly with extra code if needed).

pranavkm:
@SteveSandersonMS added a test. 

They're slightly different in the change in `JSRuntimeBase` is the serializer failing to deserialize a user type. The other one is the args payload is malformed, which is likely a  client misbehaving.

src/JSInterop/Microsoft.JSInterop/test/DotNetDispatcherTest.cs
```diff
@@ -363,6 +363,22 @@ public void CannotInvokeWithIncorrectNumberOfParams()
...
+            var callId = "123";
+            var resultTask = jsRuntime.NextInvocationTask;
+            DotNetDispatcher.BeginInvoke(callId, thisAssemblyName, "InvocableStaticWithParams", default, "<xml>not json</xml>");
+
+            await resultTask; // This won't throw, it sets properties on the jsRuntime.
+
+            // Assert
+            var result = JsonDocument.Parse(jsRuntime.LastInvocationArgsJson).RootElement;
+            Assert.Equal(callId, result[0].GetString());
+            Assert.False(result[1].GetBoolean()); // Fails
```
SteveSandersonMS: Is there some reasonable way to assert that the error message reflects the reason (args were not valid JSON)?

PR #1804, opened 2019-06-06
Add SystemdServices package that provides Host integration for systemd.

tmds:
This package provides a IHostBuilder UseSystemdService method that:
- changes the Console logger to ConsoleLoggerFormat.Systemd
- provides SystemdServiceLifetime that triggers shutdown on SIGTERM
- and which uses the systemd notify protocol to inform systemd when the service
has started, and when it is stopping.

When the extension method is used outside systemd, it noops.

Contributes to https://github.com/aspnet/Extensions/issues/1241.


tmds:
The implementation mostly is a copy-paste from [Tmds.Systemd](https://github.com/tmds/Tmds.Systemd) adjusted to make it more DI friendly and following style of the WindowsServices package.

I'll add some tests that verify the socket peer receives the expected messages next week.

src/Hosting/SystemdServices/src/ServiceState.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Main process ID (PID) of the service, in case the service manager did not fork off the process itself.
+        /// </summary>
+        public static ServiceState MainPid(int value) => new ServiceState($"MAINPID={value}");
+
+        /// <summary>
+        /// Create custom ServiceState.
+        /// </summary>
+        public ServiceState(string state)
```
tmds: This currently includes a number of messages from https://www.freedesktop.org/software/systemd/man/sd_notify.html which are not used. I can leave those out, the user can construct them if desired using this constructor.

src/Hosting/SystemdServices/src/ISystemdNotifier.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// Provides support to notify systemd about the service status.
+    /// </summary>
+    public interface ISystemdNotifier
+    {
+        /// <summary>
+        /// Sends a notification to systemd.
+        /// </summary>
+        void Notify(ServiceState state, params ServiceState[] states);
```
tmds: I can simplify this to accept a single state.

tmds:
@glennc @davidfowl @anurse @BrennanConroy @Tratcher @gfoidl ptal

src/Hosting/SystemdServices/src/Microsoft.Extensions.Hosting.SystemdServices.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>.NET hosting infrastructure for Systemd Services.</Description>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
tmds: This is netstandard2.1 because of the `UnixDomainSocketEndPoint`. If desired, it can be made `netstandard2.0` by including an internal implementation.

src/Hosting/SystemdServices/src/ISystemdNotifier.cs
gfoidl: The single state makes sense to me.

src/Hosting/SystemdServices/src/ISystemdNotifier.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// <summary>
+        /// Returns whether systemd is configured to receive service notifications.
+        /// </summary>
+        bool IsEnabled { get; }
+    }
+}
```
gfoidl: Nit: newline (other files as well).

src/Hosting/SystemdServices/src/ServiceHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using System.Threading;
+
+namespace Microsoft.Extensions.Hosting.SystemdServices
+{
+    /// <summary>
+    /// Helper methods for systemd Services.
+    /// </summary>
+    public static class ServiceHelpers
```
gfoidl: `SystemdHelpers`? As `ServiceHelpers` is quite generic.

Can this class be `internal`?

src/Hosting/SystemdServices/src/ServiceHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        const string INVOCATION_ID = "INVOCATION_ID";
```
gfoidl:
```suggestion
        private const string INVOCATION_ID = "INVOCATION_ID";
```

src/Hosting/SystemdServices/src/SystemdNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+using System;
+using System.IO;
+using System.Text;
+using System.Net.Sockets;
+
+namespace Microsoft.Extensions.Hosting.SystemdServices
+{
+    public class SystemdNotifier : ISystemdNotifier
+    {
+        const string NOTIFY_SOCKET = "NOTIFY_SOCKET";
```
gfoidl:
```suggestion
        private const string NOTIFY_SOCKET = "NOTIFY_SOCKET";
```

src/Hosting/SystemdServices/src/SystemdNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <inheritdoc />
+        public void Notify(ServiceState state, ServiceState[] states)
+        {
+            if (!IsEnabled)
+            {
+                return;
+            }
+
+            byte[] data;
```
gfoidl: Can this buffer be rented from the `ArrayPool<byte>`? Would safe the allocation(s).

State is ASCII, so the rented size can be computed.
Same for the `states.Length != 0`-path. Size can be computed, rent the buffer, construct the MS from that buffer.

src/Hosting/SystemdServices/src/SystemdServiceHostBuilderExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// and configures console logging to the systemd format.
+        /// </summary>
+        /// <remarks>
+        ///   <para>
+        ///     This is context aware and will only activate if it detects the process is running
+        ///     as a systemd Service.
+        ///   </para>
+        ///   <para>
+        ///     The systemd service file must be configured with <c>Type=notify</c> to enable
+        ///     notifications. See https://www.freedesktop.org/software/systemd/man/systemd.service.html.
```
gfoidl: 👍 (especially for the link).

src/Hosting/SystemdServices/src/SystemdNotifier.cs
gfoidl: Or stackalloced if the size will be small.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
tmds: I took the same scope and name as the class from the WindowsServices package.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
tmds: I'll take a look at what is possible here based on the desired target tfm (netstandard2.0 or netstandard 2.1).

src/Hosting/SystemdServices/src/ServiceHelpers.cs
davidfowl: Yea it’s unfortunate we expose that method with such a generic name in the windows service package as well

gfoidl: Can this be fixed or is it too late?

davidfowl: It can be fixed but https://github.com/dotnet/corefx/issues/36842 isn't being fixed so we need something.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+        public static bool IsSystemdService()
+        {
+            // We use the INVOCATION_ID envirionment variable that was introduced in systemd 232 (released 2016-11-03).
+            // We clear the envvar so .NET child processes return false for this method.
+            if (s_invocationId == null)
+            {
+                string invocationId = Environment.GetEnvironmentVariable(INVOCATION_ID);
+                if (invocationId == null)
+                {
+                    invocationId = "";
```
dougbu: nit: `string.Empty`

src/Hosting/SystemdServices/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
+
+  <PropertyGroup>
+    <!-- Source build does not support Systemd. -->
```
dougbu: Hmm, there's no extra dependencies that force this. Why not?

src/Hosting/SystemdServices/src/ServiceHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        private static string s_invocationId;
+
+        /// <summary>
+        /// Check if the current process is hosted as a systemd Service.
+        /// </summary>
+        /// <returns><c>True</c> if the current process is hosted as a systemd Service, otherwise <c>false</c>.</returns>
+        public static bool IsSystemdService()
+        {
+            // We use the INVOCATION_ID envirionment variable that was introduced in systemd 232 (released 2016-11-03).
```
davidfowl: This seems so wrong to me. Why do we need to care about child processes not having this environment variable?

src/Hosting/SystemdServices/src/SystemdNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+        internal SystemdNotifier(string socketPath)
+        {
+            _socketPath = socketPath;
+        }
+
+        /// <inheritdoc />
+        public bool IsEnabled => _socketPath != null;
+
+        /// <inheritdoc />
+        public void Notify(ServiceState state, ServiceState[] states)
```
davidfowl: `Span<ServiceState>`

src/Hosting/SystemdServices/src/SystemdNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                data = ms.ToArray();
+            }
+
+            using (var socket = new Socket(AddressFamily.Unix, SocketType.Dgram, ProtocolType.Unspecified))
+            {
+                var endPoint = new UnixDomainSocketEndPoint(_socketPath);
+                socket.Connect(endPoint);
+
+                int rv = socket.Send(data);
```
davidfowl: 1. Async?
2. Errors?

src/Hosting/SystemdServices/src/SystemdNotifier.cs
davidfowl: +1 This method allocates too much. How often is it called in mainline uses cases?

src/Hosting/SystemdServices/test/UseSystemdServiceTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting.Internal;
+using Xunit;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public class UseSystemdServiceTests
+    {
+        [Fact]
+        public void DefaultsToOffOutsideOfService()
```
davidfowl: What's the strategy for testing this? Kestrel was using Docker to test the systemd activation feature https://github.com/aspnet/AspNetCore/tree/master/src/Servers/Kestrel/test/SystemdActivation, we should probably do the same here.

src/Hosting/SystemdServices/src/Microsoft.Extensions.Hosting.SystemdServices.csproj
natemcmaster: I don't think we have a compelling scenario for netstandard2.0 yet, so 2.1 is fine :)

davidfowl: I agree for now maybe. We can polyfill if the demand is high.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
gfoidl: I thought about async too, but it's called in a sync context, so there will be waiting anywhere.
Is `socket.SendAsync(...).GetAwaiter().GetResult()` a better option here?

davidfowl: It's currently called in a single place but this is a general purpose interface. We may need both sync and async versions.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
gfoidl: Aside from that, I mean can the name be changed to `SystemdHelpers` and `WindowsServiceHelpers`?

Thinking loud: Should https://github.com/dotnet/corefx/issues/36842 be extended for `IsRunningSystemdService`?

src/Hosting/SystemdServices/src/ISystemdNotifier.cs
davidfowl: We need Notify and NotifyAsync

davidfowl: Actually, these messages are extremely small... Maybe it's overkill?

src/Hosting/SystemdServices/src/Microsoft.Extensions.Hosting.SystemdServices.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>hosting</PackageTags>
+    <IsShipping>true</IsShipping>
```
anurse: Should we consider this for the linux shared framework? That would make it possible to enable it by default in the worker template (assuming it does what `.UseWindowsServices` does and only lights up if actually hosted in systemd).

That would also mean it needs to be in source-build.

Of course, I don't know exactly what (if any?) shared framework the worker template references. Does it reference `Microsoft.AspNetCore.App` or `Microsoft.NETCore.App`? If the latter, the point is moot, it's not going in to that shared fx in this release. (cc @glennc)

src/Hosting/SystemdServices/Directory.Build.props
anurse: It's not in a shared framework right now, so it doesn't **have** to be source-built. Having said that, it is a component for linux services which means it is probably desirable to land it in some kind of shared framework some day. Since it has no extra deps, I'd agree, let's try to put this in source-build now to avoid issues later.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
anurse: I think this should be `SystemdHelpers` and the Windows Service one should be `WindowsServicesHelpers` or something.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
anurse: If an env-var is the only way we can detect this, we might be stuck with this option.

For windows services, [we check the identity of the parent process](https://github.com/aspnet/Extensions/blob/9d6a021d72c2495d0944c4557c1847181d545061/src/Hosting/WindowsServices/src/ServiceHelpers.cs#L25). Is that a viable option here? IIRC, it's not trivial to detect your parent process in Linux though... Looks like it can be [read from `/proc/self/stat`](http://man7.org/linux/man-pages/man5/proc.5.html) if there isn't already an API in .NET Core for it.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
anurse: From earlier conversations with @glennc, I believe this is not a hot path at all. It's only used during an app state transition (starting, stopping, app-level errors, etc.).

`stackalloc`ing is reasonable but must only be done if the amount of space needed is finite and relatively small. Could always fall back to a rented buffer if the necessary space is too big.

Based on my understanding above, this doesn't feel like it needs super heavy optimization. One option is to use pre-defined byte arrays for the well-known values (see comment on ServiceState above).

src/Hosting/SystemdServices/src/ServiceState.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+
+namespace Microsoft.Extensions.Hosting.SystemdServices
+{
+    /// <summary>
+    /// Describes a service state change.
+    /// </summary>
+    public struct ServiceState
+    {
+        private string _state;
```
anurse: Since this is always going to be ASCII, what about storing the encoded bytes here instead of a `string`? Then there's no need to allocate when writing them to the socket. Unfortunately we can't store a `ReadOnlySpan<byte>` without making this a `ref struct` though, which means we don't get the static data optimization in C#...

Still, the constant values (`Ready`, etc.) could be stored as once-allocated `static readonly byte[]` fields.

src/Hosting/SystemdServices/src/ServiceState.cs
anurse: If you take the `byte[]` suggestion above, we can keep this ctor as a convenience method for users who are creating custom values. But we should also add a `ServiceState(byte[])` ctor.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
anurse: > Is `socket.SendAsync(...).GetAwaiter().GetResult()` a better option here?

Definitely not. I'd make this a `NotifyAsync` method and allow callers to decide what to do. For example, a caller could decide to fire-and-forget instead of blocking.

src/Hosting/SystemdServices/src/SystemdServiceLifetime.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            return Task.CompletedTask;
+        }
+
+        private void OnApplicationStarted()
+        {
+            Logger.LogInformation("Application started. Hosting environment: {envName}; Content root path: {contentRoot}",
+                Environment.EnvironmentName, Environment.ContentRootPath);
+
+            SystemdNotifier.Notify(ServiceState.Ready);
```
anurse: It would seem reasonable, if `Notify` was `async`, to fire-and-forget here (maybe with some code to capture and log errors?). Since `OnApplicationStarted` is synchronous, we don't really have other options here.

src/Hosting/SystemdServices/src/SystemdServiceLifetime.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private void OnApplicationStopping()
+        {
+            Logger.LogInformation("Application is shutting down...");
+
+            SystemdNotifier.Notify(ServiceState.Stopping);
```
anurse: fire-and-forget here too?

src/Hosting/SystemdServices/test/UseSystemdServiceTests.cs
anurse: Some of it could be tested without systemd at all. For example the notifier could be tested by creating a socket, flowing it down (preferably via internals rather than an env-var) and reading from the other end.

But I agree, docker for a simple integration test would be good as well.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
tmds: Q: I've been naming things here 'SystemdService' because the Windows package names things 'WindowsService'. The Service suffix could be dropped, e.g. `UseSystemdService` vs `UseSystemd`. I don't have a strong preference for either.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
tmds: > I believe this is not a hot path at all.

That's correct. This gets called once on startup, and once on shutdown.

> One option is to use pre-defined byte arrays for the well-known values 

+1

src/Hosting/SystemdServices/Directory.Build.props
tmds: I copied this over from the WindowsService, I'll remove it.

src/Hosting/SystemdServices/src/ServiceHelpers.cs
tmds: > This seems so wrong to me. Why do we need to care about child processes not having this environment variable?

The corresponding WindowsService function also won't return true for children.

> r windows services, we check the identity of the parent process. Is that a viable option here? IIRC, it's not trivial to detect your parent process in Linux though... Looks like it can be read from /proc/self/stat if there isn't already an API in .NET Core for it.

There are a number of options, the envvar is the simplest.
Other option: look at the parents, use cgroup names.
note: we want this to work also for systemd instances that run per user, that adds some complexity.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
tmds: We can treat this as a sync API. The buffer in the kernel is far larger than what we are trying to send.

src/Hosting/SystemdServices/test/UseSystemdServiceTests.cs
tmds: > For example the notifier could be tested by creating a socket, flowing it down (preferably via internals rather than an env-var) and reading from the other end.

I plan to add such a test.

> Kestrel was using Docker to test the systemd activation feature https://github.com/aspnet/AspNetCore/tree/master/src/Servers/Kestrel/test/SystemdActivation, we should probably do the same here.

Are you running these tests on regular basis?

src/Hosting/SystemdServices/src/Microsoft.Extensions.Hosting.SystemdServices.csproj
davidfowl: We don't do this for the windows service so we shouldn't do it here either....

src/Hosting/SystemdServices/src/ISystemdNotifier.cs
tmds: cfr earlier comment, it's overkill.

src/Hosting/SystemdServices/test/UseSystemdServiceTests.cs
davidfowl: They don't run at all, in fact, I need to file an issue to make them run again.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
gfoidl: > once on startup, and once on shutdown.

Then I would stick with the current code, as it will be the most readable one.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
gfoidl: Good to know. Can you add a comment for that fact please?

src/Hosting/SystemdServices/src/ServiceHelpers.cs
anurse: I'm fine with `UseSystemd`. It's `UseWindowsService` because "Windows Service" is the kind of executable, not necessarily because it's the "Windows" "background service" concept :). Plus `UseWindows` wouldn't be sufficient there.

src/Hosting/SystemdServices/src/SystemdNotifier.cs
anurse: > We can treat this as a sync API. The buffer in the kernel is far larger than what we are trying to send.

Hrm, we generally avoid operating under assumptions like that, but I see your point. Thoughts @davidfowl ?

src/Hosting/SystemdServices/src/ISystemdNotifier.cs
anurse: As long as somebody doesn't have a huge `Status` update :)

src/Hosting/SystemdServices/src/ServiceHelpers.cs
```diff
@@ -0,0 +1,41 @@
...
+            // We clear the envvar so .NET child processes return false for this method.
+            if (s_invocationId == null)
+            {
+                string invocationId = Environment.GetEnvironmentVariable(INVOCATION_ID);
+                if (invocationId == null)
+                {
+                    invocationId = "";
+                }
+                if (Interlocked.CompareExchange(ref s_invocationId, invocationId, null) == null)
```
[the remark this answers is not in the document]
Tratcher: Thinking about this more, this kind of state change doesn't belong in a status check method like this. Put it in the lifetime instead.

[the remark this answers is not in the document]
tmds: I'm going to remove clearing the envvar. After review comments, I think we can just keep the envvar and not worry about child processes seeing it too.

src/Hosting/SystemdServices/src/SystemdServiceLifetime.cs
```diff
@@ -0,0 +1,90 @@
...
+            _shutdownBlock.WaitOne();
+
+            // On Linux if the shutdown is triggered by SIGTERM then that's signaled with the 143 exit code.
+            // Suppress that since we shut down gracefully. https://github.com/aspnet/AspNetCore/issues/6526
+            System.Environment.ExitCode = 0;
+        }
+
+        public void Dispose()
+        {
+            _shutdownBlock.Set();
```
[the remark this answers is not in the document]
tmds: Interesting. I want to have a closer look at that issue after this. For now, I'll adopt the same implementation.

src/Hosting/Systemd/src/ServiceState.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// Service is beginning its shutdown.
+        /// </summary>
+        public static readonly ServiceState Stopping = new ServiceState("STOPPING=1");
+
+        /// <summary>
+        /// Create custom ServiceState.
+        /// </summary>
+        public ServiceState(string state)
+        {
+            _data = Encoding.UTF8.GetBytes(state) ?? throw new ArgumentNullException(nameof(state));
```
campersau:
```suggestion
            _data = Encoding.UTF8.GetBytes(state ?? throw new ArgumentNullException(nameof(state)));
```

src/Hosting/Systemd/src/SystemdHostBuilderExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        ///   <para>
+        ///     The systemd service file must be configured with <c>Type=notify</c> to enable
+        ///     notifications. See https://www.freedesktop.org/software/systemd/man/systemd.service.html.
+        ///   </para>
+        /// </remarks>
+        /// <param name="hostBuilder"></param>
+        /// <returns></returns>
+        public static IHostBuilder UseSystemd(this IHostBuilder hostBuilder)
+        {
+            if (SystemdHelpers.IsSystemdService ())
```
jkotalik:
```suggestion
            if (SystemdHelpers.IsSystemdService())
```

src/Hosting/Systemd/src/SystemdLifetime.cs
```diff
@@ -0,0 +1,90 @@
...
+            _shutdownBlock.WaitOne();
+
+            // On Linux if the shutdown is triggered by SIGTERM then that's signaled with the 143 exit code.
+            // Suppress that since we shut down gracefully. https://github.com/aspnet/AspNetCore/issues/6526
+            System.Environment.ExitCode = 0;
+        }
+
+        public void Dispose()
+        {
+            _shutdownBlock.Set();
```
jkotalik: What happens if SystemdLifetime isn't disposed? Wouldn't OnProcessExit hang?

halter73: OnProcessExit calls ApplicationLifetime.StopApplication which should dispose the DI container and therefore this. That said, it might not be a bad idea to timeout _shutdownBlock.WaitOne() and throw if it does timeout which a message explaining a SystemdLifetime didn't get disposed for some reason.

I'm guessing, thought I'm not sure, that the OnProcessExit call itself will eventually timeout (even if it's just because the process terminates) so shutdown doesn't actually hang forever. If that's true, we'd want to make sure we throw prior to that timeout. And we'd also want to make sure that exceptions thrown from OnProcessExit are easily observed.

In practice, I doubt this will come up much though, so I'm not too worried about it.

Tratcher: Yes it would hang. We have another thread on this: https://github.com/aspnet/Extensions/issues/1363

src/Hosting/Systemd/src/SystemdNotifier.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <inheritdoc />
+        public void Notify(ServiceState state)
+        {
+            if (!IsEnabled)
+            {
+                return;
+            }
+
+            using (var socket = new Socket(AddressFamily.Unix, SocketType.Dgram, ProtocolType.Unspecified))
```
davidfowl: Is it reasonable to open and close these on demand? Is there a universe where we'd want to keep the connection open because of sending other types of messages?

tmds: This is only for occasional messaging (now: two messages per application lifetime).
The native sd_notify implementation does the same thing: create and close socket per message.
We can change the scope of the socket to the SystemdNotifier instance if it ever makes sense.

davidfowl: Sounds good.

PR #1807, opened 2019-06-07
Add retries to HttpClientSlim on macOS.

jkotalik:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2637. If you need a backstory, see https://github.com/aspnet/KestrelHttpServer/issues/2635

### DIFF
--- a/src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
+++ b/src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
@@ -121,7 +121,9 @@ namespace Microsoft.AspNetCore.Testing
                     }
                 }
             }
-            throw new OperationCanceledException("Failed to connect, retry limit exceeded");
+
+            // This will never be hit.
+            throw new OperationCanceledException();
         }
 
         private static HttpStatusCode GetStatus(string response)
--- a/src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
+++ b/src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
@@ -115,7 +115,7 @@ namespace Microsoft.AspNetCore.Testing
                 {
                     return await retryBlock().ConfigureAwait(false);
                 }
-                catch (Exception)
+                catch (InvalidDataException)
                 {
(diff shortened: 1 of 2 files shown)


src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
```diff
@@ -94,6 +100,32 @@ private static async Task<string> ReadResponse(Stream stream)
...
+                {
+                    if (retry == retryCount - 1)
+                    {
+                        throw;
+                    }
+                }
+            }
+
+            // This will never be hit.
+            throw new OperationCanceledException();
```
Tratcher: NotImplementedException?

src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
```diff
@@ -94,6 +100,32 @@ private static async Task<string> ReadResponse(Stream stream)
             }
         }
 
+        private static async Task<string> RetryRequest(Func<Task<string>> retryBlock)
+        {
+#if MACOS
```
Tratcher: Doesn't this need to be a runtime check? This is packaged by TFM, not by OS.

jkotalik: Yeah, you are correct. It depends on runtime.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/HttpClientSlim.cs
```diff
@@ -94,6 +101,33 @@ private static async Task<string> ReadResponse(Stream stream)
...
+                retryCount = 3;
+            }
+
+            for (var retry = 0; retry < retryCount; retry++)
+            {
+                try
+                {
+                    return await retryBlock().ConfigureAwait(false);
+                }
+                catch (Exception)
```
halter73: Can we retry just when ReadResponse gets no data instead of retrying for any exception?

jkotalik: Sure, I'll just retry on `InvalidDataException` which is only thrown if there is no data.

PR #1814, opened 2019-06-09
Force StringValues.ToArray to copy the underlying array

marcelltoth:
### Summary of the changes
 - Added test demonstrating the problem
 - Reimplemented `ToArray` to be independent from `GetArrayValue`
 - Cleaned up `GetArrayValue` to look syntactically like the new `ToArray`. (I added comments to highlight the differences.)

Addresses #1813

### DIFF
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -225,11 +225,7 @@ namespace Microsoft.Extensions.Primitives
             switch (value)
             {
                 case string[] values:
-#if NETCOREAPP
-                    return values.AsSpan().ToArray();
-#else
-                    return (string[])values.Clone();
-#endif
+                    return new ReadOnlySpan<string>(values).ToArray();
                 case null:
                     return Array.Empty<string>();
                 default:
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -225,7 +225,11 @@ namespace Microsoft.Extensions.Primitives
             switch (value)
             {
                 case string[] values:
-                    return (string[]) values.Clone();
+#if NETCOREAPP
+                    return (string[]) values.AsSpan().ToArray();
+#else
+                    return (string[])values.Clone();
+#endif
                 case null:
                     return Array.Empty<string>();
                 default:
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -226,7 +226,7 @@ namespace Microsoft.Extensions.Primitives
             {
                 case string[] values:
 #if NETCOREAPP
-                    return (string[]) values.AsSpan().ToArray();
+                    return values.AsSpan().ToArray();
 #else
                     return (string[])values.Clone();
 #endif
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using Microsoft.Extensions.Internal;
 
@@ -218,7 +219,20 @@ namespace Microsoft.Extensions.Primitives
 
         public string[] ToArray()
         {
-            return GetArrayValue() ?? Array.Empty<string>();
+            // The same idea as GetArrayValue with slightly different semantics: This method always returns a copy of the array (just like the IEnumerable.ToArray extension method).
+
+            // Take local copy of _values so type checks remain valid even if the StringValues is overwritten in memory.
+            var value = _values;
+            switch (value)
+            {
+                case string[] values:
+                    return (string[]) values.Clone();
+                case null:
+                    return Array.Empty<string>();
+                default:
+                    // value not array, can only be string
+                    return new[] { Unsafe.As<string>(value) };
+            }
         }
 
         private string[] GetArrayValue()
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using System.Runtime.CompilerServices;
 using Microsoft.Extensions.Internal;
 
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -237,20 +237,19 @@ namespace Microsoft.Extensions.Primitives
 
         private string[] GetArrayValue()
         {
+            // The same idea as ToArray with slightly different semantics: This method reuses the existing array if possible.
+
             // Take local copy of _values so type checks remain valid even if the StringValues is overwritten in memory
             var value = _values;
-            if (value is string[] values)
+            switch (value)
             {
-                return values;
-            }
-            else if (value != null)
-            {
-                // value not array, can only be string
-                return new[] { Unsafe.As<string>(value) };
-            }
-            else
-            {
-                return null;
+                case string[] values:
+                    return values;
+                case null:
+                    return null;
+                default:
+                    // value not array, can only be string
+                    return new[] { Unsafe.As<string>(value) };
             }
         }
 
--- a/src/Primitives/test/StringValuesTests.cs
+++ b/src/Primitives/test/StringValuesTests.cs
@@ -522,5 +522,27 @@ namespace Microsoft.Extensions.Primitives
             Assert.True(StringValues.Equals(stringValues, expected));
             Assert.False(StringValues.Equals(stringValues, notEqual));
         }
+
+        [Fact]
+        public void ToArray_ConstructedFromString_CopiesValues()
+        {
+            var singleStringValue = new StringValues("abc");
+
+            var array = singleStringValue.ToArray();
+            array[0] = "bcd";
+
+            Assert.Equal("abc", singleStringValue[0]);
+        }
+
+        [Fact]
+        public void ToArray_ConstructedFromArray_CopiesValues()
+        {
+            var singleStringValue = new StringValues(new [] {"abc"});
+
+            var array = singleStringValue.ToArray();
+            array[0] = "bcd";
+
+            Assert.Equal("abc", singleStringValue[0]);
+        }
     }
 }


marcelltoth:
Alternative solutions considered:

- Change the behavior of `GetArrayValue` itself. This would change the behavior of `implicit operator string[] (StringValues value)`, for which **I assumed that we may want to keep the current behavior for performance reasons**. See the _Further considerations_ section of the issue. I might be wrong though, let me know if it is safe to change.
- Remove `ToArray` altogether and rely on the `IEnumerable` extension method. **This would break binary compatibility**, and also possibly lead to worse performance.

src/Primitives/src/StringValues.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
```
benaadams: Is this needed?

src/Primitives/src/StringValues.cs
```diff
@@ -218,25 +219,37 @@ static string GetJoinedStringValueFromArray(string[] values)
...
+            // The same idea as ToArray with slightly different semantics: This method reuses the existing array if possible.
+
             // Take local copy of _values so type checks remain valid even if the StringValues is overwritten in memory
             var value = _values;
-            if (value is string[] values)
-            {
-                return values;
-            }
-            else if (value != null)
+            switch (value)
```
benaadams: The `if` generates better asm than the `switch` [sharplab.io](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdzOgG0Aug1kB3XhjAALABQB5YACsYYBgMCtiSHDAAlAym5ibU5gkMuM6ubgweIWGRMYnGObnmYHgwTOQ29pnhuCD5BYnEAOzpVnZRlarCkhBcMB4RGvF1hcUMXBySkjWDQ/VNAIJQUNgAniwAouIYywA8VgB8fQMzZgAmMABm2ONYtccA9HfBoeGjEEHYiytoDEVcDN2SZYMYAlKy3GaNUYwRx2YwMACqXFw2HOgjmuF2ZQO7SiAF8jgVcfkidNoqSdKVygwAJLnbx+AJBHFkhJxOq8c7pdoMXi4Smw9q4CLgvKkuqQjwtWwRQWdbq9foikl1GCSXAlDlc54lACEAF5RuNJMKxbERQlIT0YfYjAikSi0Rj9hltXiCbllQVVeqWQU2cdLUa1AklcSaLigA)

src/Primitives/src/StringValues.cs
marcelltoth: Not at all, I missed it, thanks.

src/Primitives/src/StringValues.cs
marcelltoth: Maybe, but it is actually measurably better? If not, the switch is a easier to read / more expressive. I try to make a benchmark on it.

benaadams: If you are benchmarking 😉 then...
```csharp
return values.AsSpan().ToArray()
```
might be better than
```csharp
return (string[])values.Clone();
```

marcelltoth: I can give it a shot :D

marcelltoth: It looks like the switch is indeed faster in most workloads (except when the payload is null), at least on my machine, although I would say the difference is negligible:

> BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17763.529 (1809/October2018Update/Redstone5)
> Intel Core i5-6440HQ CPU 2.60GHz (Skylake), 1 CPU, 4 logical and 4 physical cores
> .NET Core SDK=3.0.100-preview5-011568
>   [Host]     : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT
>   DefaultJob : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT


| Method |     Data |      Mean |     Error |    StdDev |
|------- |--------- |----------:|----------:|----------:|
| Switch |        ? _(null)_ |  3.727 ns | 0.0643 ns | 0.0601 ns |
|     If |        ? _(null)_ |  3.692 ns | 0.0730 ns | 0.0683 ns |
| Switch | Array[2] | 73.998 ns | 1.2785 ns | 1.1333 ns |
|     If | Array[2] | 76.917 ns | 1.9549 ns | 1.8286 ns |
| Switch |      abc | 11.149 ns | 0.4064 ns | 0.5140 ns |
|     If |      abc | 12.266 ns | 0.2307 ns | 0.2158 ns |

marcelltoth: #### `AsSpan().ToArray()` is indeed about twice as fast as `Array.Clone()`

> BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17763.529 (1809/October2018Update/Redstone5)
> Intel Core i5-6440HQ CPU 2.60GHz (Skylake), 1 CPU, 4 logical and 4 physical cores
> .NET Core SDK=3.0.100-preview5-011568
>   [Host]     : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT
>   DefaultJob : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT


|      Method |     Mean |     Error |    StdDev |
|------------ |---------:|----------:|----------:|
|       Clone | 72.96 ns | 0.6034 ns | 0.5644 ns |
| SpanToArray | 27.97 ns | 0.3408 ns | 0.3188 ns |

It is nice to know this, thanks for the idea. This pretty much makes it the default go-to array clone solution on .NET Core then. I will update the code in the pull request and write about this somewhere.

src/Primitives/src/StringValues.cs
```diff
@@ -218,25 +218,41 @@ static string GetJoinedStringValueFromArray(string[] values)
...
-            return GetArrayValue() ?? Array.Empty<string>();
+            // The same idea as GetArrayValue with slightly different semantics: This method always returns a copy of the array (just like the IEnumerable.ToArray extension method).
+
+            // Take local copy of _values so type checks remain valid even if the StringValues is overwritten in memory.
+            var value = _values;
+            switch (value)
+            {
+                case string[] values:
+#if NETCOREAPP
+                    return (string[]) values.AsSpan().ToArray();
```
benaadams: Doesn't need the cast
```suggestion
                    return values.AsSpan().ToArray();
```

src/Primitives/src/StringValues.cs
```diff
@@ -218,25 +218,41 @@ static string GetJoinedStringValueFromArray(string[] values)
...
+
+            // Take local copy of _values so type checks remain valid even if the StringValues is overwritten in memory.
+            var value = _values;
+            switch (value)
+            {
+                case string[] values:
+#if NETCOREAPP
+                    return values.AsSpan().ToArray();
+#else
+                    return (string[])values.Clone();
```
benaadams: Can do a similar thing for .NET Standard; with the `System.Memory` package (which looks to already be included in the .csproj)
```suggestion
                    return new ReadOnlySpan<string>(array).ToArray();
```

marcelltoth: Let me check if that makes a difference in performance, I'm curious and I still have my test playground open.

benaadams: Actually `.AsSpan()` exists as an extension method in that package, so you should just be able to remove the `#if`

marcelltoth: Interestingly `ReadOnlySpan` is consistently faster than `Span` in this regard.

> BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17763.529 (1809/October2018Update/Redstone5)
> Intel Core i5-6440HQ CPU 2.60GHz (Skylake), 1 CPU, 4 logical and 4 physical cores
> .NET Core SDK=3.0.100-preview5-011568
>   [Host]     : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT
>   DefaultJob : .NET Core 3.0.0-preview5-27626-15 (CoreCLR 4.6.27622.75, CoreFX 4.700.19.22408), 64bit RyuJIT


|              Method |     Mean |     Error |    StdDev |
|-------------------- |---------:|----------:|----------:|
|               Clone | 82.88 ns | 1.9200 ns | 2.2857 ns |
|         SpanToArray | 34.36 ns | 0.7804 ns | 0.7300 ns |
| ReadOnlySpanToArray | 27.20 ns | 0.6373 ns | 0.9342 ns |

Looking at their sources there is only one notable difference between their constructors, this is only present in `Span`:
```
if ((object) default (T) == null && array.GetType() != typeof (T[]))
          ThrowHelper.ThrowArrayTypeMismatchException();
```
That `GetType()` might slow down things a slight bit. Does not matter much, but I can then just use `ReadOnlySpan` for both platforms, right?

benaadams: At a guess the covariance of the array is important for the Span type because you can update the elements; so the types need to be an exact match. For the ReadOnlySpan since you can't update the elements it doesn't matter...

> Does not matter much, but I can then just use ReadOnlySpan for both platforms, right?

Yes

benaadams:
> This may end up causing allocations to suddenly appear in various places in ASP.NET Core.

I was thinking about that; but I think `.ToString()` is mostly used over `.ToArray()` as it currently causes an allocation when you only have one header (which is the normal case)?

src/Primitives/src/StringValues.cs
NinoFloris: @benaadams Array.Clone, another one for the CoreFX improvement pile ;)

anurse:
What about adding a new method `CopyToArray` and documenting the quirkiness of `ToArray`? I recognize it's not the right behavior but this has a significant-enough risk of unknown performance regressions that I'd prefer to be cautious...

marcelltoth:
> 
> 
> What about adding a new method `CopyToArray` and documenting the quirkiness of `ToArray`? I recognize it's not the right behavior but this has a significant-enough risk of unknown performance regressions that I'd prefer to be cautious...

I was thinking about that but as we are shadowing `Enumerable.ToArray` this would still open up a surface for introducing bugs with seemingly innocent code. Think of this example:

```
IEnumerable<string> someStrings = someStringValues;
string[] copiedStrings = someStrings.ToArray();
Array.Sort(copeidString);
```

Now if I approve a "code style cleanup" like this:

```
var someStrings = someStringValues;
var copiedStrings = someStrings.ToArray();
Array.Sort(copeidString);
```

**this is a breaking change and a bug**. And I'm pretty confident that I wouldn't be the only code reviewer who'd let something like this go unnoticed.

I am a new contributor to this project (although I've been using it for years) so of course I can only make suggestions here, but in general I think the right path is to first make the code bug-free then go on to optimize the performance (while ensuring we do not change the behavior). And we already know of at least 1 bug in the framework that originates from this, possibly many more in both framework and user code. 

Of course I'm all for checking if `.ToArray` is used in performance critical paths before merging, and I would've done it myself, but being very new to contributing here I really do not feel confident in doing so.

src/Primitives/src/StringValues.cs
mburbea: Is it really faster to go through `.AsSpan().ToArray()` vs doing something like?
```
var copy= new T[orig.Length];
Array.Copy(orig, 0, copy, 0, copy.Length);
return copy;
```

davidfowl:
> I am a new contributor to this project (although I've been using it for years) so of course I can only make suggestions here, but in general I think the right path is to first make the code bug-free then go on to optimize the performance (while ensuring we do not change the behavior). And we already know of at least 1 bug in the framework that originates from this, possibly many more in both framework and user code.

It should be easy enough to audit the code to find the places where this is used. It's just not a priority at the moment. No way we're just going to merge this and then fix the regressions afterwards. That's an extremely risky way to go about things.

src/Primitives/src/StringValues.cs
marcelltoth: Interestingly yes. 
Check out my blog article where I test all kinds of different methods of copying an array: https://marcelltoth.net/article/41/fastest-way-to-copy-a-net-array it was inspired by this review.
The usual use case we have here is where the array is not empty but it only contains a handful of elements. In that scenario by my tests `Array.Copy` performs about 2x as slow. (See the second result set.)

PR #1820, opened 2019-06-11
Add tests for Systemd package

tmds:
This adds a test for SystemdNotifier introduced in https://github.com/aspnet/Extensions/pull/1804.

If desired, I can add an integration tests too.
It should be possible to write these as xunit tests that use the user-level systemd instance, so no docker/root required.
@davidfowl @anurse wdyt?

@glennc @BrennanConroy @Tratcher @gfoidl


davidfowl:
These are good.

src/Hosting/Systemd/test/UseSystemdTests.cs
```diff
@@ -22,5 +30,51 @@ public void DefaultsToOffOutsideOfService()
...
+                var endPoint = new UnixDomainSocketEndPoint(socketPath);
+                notifySocket.Bind(endPoint);
+
+                var systemdNotifier = new SystemdNotifier(socketPath);
+                systemdNotifier.Notify(state);
+
+                byte[] buffer = new byte[256];
+                using var cts = new CancellationTokenSource();
+                cts.CancelAfter(30000);
+                int nrBytesReceived = await notifySocket.ReceiveAsync(buffer, SocketFlags.None, cts.Token);
```
davidfowl: nit: read

tmds:
@davidfowl @anurse I've added some bits for integration testing. I need some help to get the test web application built in a way it is startable by systemd and using the libraries built here.

```
$ INVOCATION_ID=notnull .dotnet/dotnet  ./artifacts/bin/IntegrationTestApp/Debug/netcoreapp3.0/IntegrationTestApp.dll

Unhandled Exception: System.MissingMethodException: Method not found: 'Void Microsoft.Extensions.Logging.Console.ConsoleLoggerOptions.set_Format(Microsoft.Extensions.Logging.Console.ConsoleLoggerFormat)'.
   at Microsoft.Extensions.Hosting.SystemdHostBuilderExtensions.<>c.<UseSystemd>b__0_1(ConsoleLoggerOptions options)
   at Microsoft.Extensions.Options.ConfigureNamedOptions`1.Configure(String name, TOptions options)
   at Microsoft.Extensions.Options.OptionsFactory`1.Create(String name)
   at Microsoft.Extensions.Options.OptionsMonitor`1.<>c__DisplayClass11_0.<Get>b__0()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Extensions.Options.OptionsCache`1.GetOrAdd(String name, Func`1 createOptions)

```

src/Hosting/Systemd/IntegrationTestApp/IntegrationTestApp.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
Tratcher: Move to src/Hosting/Systemd/testassets/

tmds: @Tratcher I'm trying to find out how I can build and run this app so it uses libraries that are part of this repo. Is there an application in the repo already that I can have a look at?

Tratcher: Our location conventions matter a lot. Start by moving it to a testassets folder.
https://github.com/aspnet/AspNetCore/blob/master/src/Servers/testassets/ServerComparison.TestSites/ServerComparison.TestSites.csproj

tmds: I've moved the app to the testassets folder.
It ends up being built at `./artifacts/bin/IntegrationTestApp/Debug/netcoreapp3.0/IntegrationTestApp.dll`.
Trying to run that doesn't use the proper libraries:
```
INVOCATION_ID=not-empty .dotnet/dotnet ./artifacts/bin/IntegrationTestApp/Debug/netcoreapp3.0/IntegrationTestApp.dll

Unhandled Exception: System.MissingMethodException: Method not found: 'Void Microsoft.Extensions.Logging.Console.ConsoleLoggerOptions.set_Format(Microsoft.Extensions.Logging.Console.ConsoleLoggerFormat)'.
   at Microsoft.Extensions.Hosting.SystemdHostBuilderExtensions.<>c.<UseSystemd>b__0_1(ConsoleLoggerOptions options)
   at Microsoft.Extensions.Options.ConfigureNamedOptions`1.Configure(String name, TOptions options)
   at Microsoft.Extensions.Options.OptionsFactory`1.Create(String name)
   at Microsoft.Extensions.Options.OptionsMonitor`1.<>c__DisplayClass11_0.<Get>b__0()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Extensions.Options.OptionsCache`1.GetOrAdd(String name, Func`1 createOptions)
...
```

Tratcher: @aspnet/build ?

natemcmaster: It might be problematic to be running tests against AspNetCore APIs in this repo. The dependency flow of our repos is dotnet/core-setup -> Extensions -> AspNetCore -> core-sdk. To avoid a cycle, this repo is using an LKG version of the SDK and AspNetCore runtime. If you were to need a new API from or need to make a change that impacts AspNetCore, you would have to wait for the next stable or preview release to cycle that change back down to Extensions.

The other problem you might run into is the way the SDK trims assemblies against a FrameworkReference. I'm not 100% sure if that is what causes the missing method error right now, but this is where I would start looking. When the SDK sees the same assembly name in a project/package reference and a FrameworkReference, it chooses the higher assembly version or file version. It's probable that at some point this will be a problem for this test because local builds of this repo might be producing assemblies of a lower version than what is in the shared framework.

So, recommendation...consider moving tests that rely on AspNetCore into the AspNetCore repo. Also, binary logs are a great way to investigate issues related to conflict resolution of assemblies.

src/Hosting/Systemd/testassets/IntegrationTestApp/IntegrationTestApp.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsPackable>true</IsPackable>
```
natemcmaster: Can you remove this line? I don't think we want to ship this project as a package.

src/Hosting/Systemd/IntegrationTestApp/IntegrationTestApp.csproj
Tratcher: The test doesn't rely on AspNetCore... Is our mistake just using Microsoft.NET.Sdk.Web? I found one other test app in this repo and it uses Microsoft.Net.Sdk.
https://github.com/aspnet/Extensions/blob/master/src/Hosting/test/testassets/Microsoft.Extensions.Hosting.TestApp/Microsoft.Extensions.Hosting.TestApp.csproj

Tratcher: Oh, I missed the part with ConfigureWebHostDefaults and Startup. Those can be removed.

tmds: > Is our mistake just using Microsoft.NET.Sdk.Web?

:+1:, looking better!
```
 INVOCATION_ID=notnull .dotnet/dotnet  ./artifacts/bin/IntegrationTestApp/Debug/netcoreapp3.0/IntegrationTestApp.dll
<6>Microsoft.Hosting.Lifetime[0] Application started. Hosting environment: Production; Content root path: /home/tmds/repos/AspNetExtensions
```

src/Hosting/Systemd/test/UserService.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+
+        public void Start()
+        {
+            // In case the service wasn't properly stopped in a previous test run.
+            TryStop();
+
+            WriteUnitFile();
+            RunProcess("systemctl", "daemon-reload");
```
BrennanConroy: Isn't it supposed to be "sysctl"

tmds: sysctl is for settings, systemctl is for interacting with systemd.

tmds:
The test was added and it passes on my machine.
It gets skipped in CI because there is no `systemd --user` instance running.
Maybe the CI machines can be configured to run this?

The `SystemdStartStopWorks` test is up for review.

BrennanConroy:
> It gets skipped in CI because there is no `systemd --user` instance running.
> Maybe the CI machines can be configured to run this?

@aspnet/build Thoughts?

dougbu:
> @aspnet/build Thoughts?

Where are these tests attempting to run? If just in the main build, try moving them to Helix (where we have more Linux varieties).

BrennanConroy:
> try moving them to Helix

It seems that Extensions doesn't have helix o.O

dougbu:
This is odd because `systemd` should be the default on Ubuntu 16.04 and `systemd --user` should be enabled with the default `/etc/pam.d/system-login` configuration. I'm not sure why it's been disabled.

But, feel free to try enabling the service in .azure_pipelines.yml. (These agents are recreated for every job so you won't be making permanent changes.)

davidfowl:
This needs more docker? Does it change any global machine state?

dougbu:
@davidfowl with these agents, changing machine state isn't a big deal. The agents aren't reused.

davidfowl:
It's a huge deal when I run a build locally.

dougbu:
> It's a huge deal when I run a build locally.

My suggestion was to update azure-pipelines.yml, not to change the build scripts. Up to you to enable / disable `systemd --user` when you do / don't want to test this feature locally.

tmds:
> Does it change any global machine state?

The test writes the same config files each time. The test stops the service and cleans up the files.

Worst case (test process crash), the cleanup doesn't happen, then the files stick around, and the service (a no-op host) runs till the user logs out.

tmds:
> This is odd because systemd should be the default on Ubuntu 16.04 and systemd --user should be enabled with the default /etc/pam.d/system-login configuration. I'm not sure why it's been disabled.
But, feel free to try enabling the service in .azure_pipelines.yml. (These agents are recreated for every job so you won't be making permanent changes.)

@dougbu I don't know how to reproduce the environment to figure out what steps are needed.

davidfowl:
> Worst case (test process crash), the cleanup doesn't happen, then the files stick around, and the service (a no-op host) runs till the user logs out.

That sounds pretty safe. Systemd is installed by default so it shouldn't be an issue right?

tmds:
> Systemd is installed by default so it shouldn't be an issue right?

The test detects whether there is a `systemd --user` instance running. This seems not to be the case on the CI ubuntu machine. This should be the case on development machines.

This test passes on my development machine. It would be nice if we could get it to also run on the CI machine. If it is hard to figure out how to make that work, maybe we can merge the PR without that. I can't reproduce the CI machine configuration to figure out what needs to change in azure_pipelines.yml, but maybe @dougbu can help.

dougbu:
Suggest starting with adding a debugging step to the Ubuntu build job e.g. `systemctl is-system-running; systemctl --user is-system-running; systemd-analyze; systemd-analyze critical-chain`. I found commands like these in https://wiki.ubuntu.com/DebuggingSystemd#systemd_:_daemon.2C_app_startup_issues

pranavkm:
@tmat have you had a chance to follow @dougbu's suggestion?

tmds:
Not yet. I'll try to look into this by next week.

tmds:
@dougbu @pranavkm I looked into this, the systemd user deamon isn't running on the CI system.

Tratcher:
So what's next here?

tmds:
> So what's next here?

These tests work on a system with systemd user daemon, and are skipped on a system without.

I see my rebase on master has caused some failures which I'll look into.

I'm not sure what is causing the systemd user daemon to not be enabled on the CI machines.

anurse:
I think we need to get the @aspnet/aspdoi involved to talk about enabling systemd on our build agents or a helix queue or something so we can actually run these tests in CI.

tmds:
> talk about enabling systemd on our build agents or a helix queue or something so we can actually run these tests in CI.

The system level systemd instance is running. The user level instance isn't.
I'm not familiar with how this gets enabled/disabled.

anurse:
Update: We haven't been able to get anyone to investigate getting systemd up on the build agents, other things have been higher priority :(.

Anipik:
These project have been moved to https://github.com/dotnet/runtime/tree/main/src/libraries/Microsoft.Extensions.Hosting.Systemd

feel free to throw a pr in runtime repo

PR #1828, opened 2019-06-12
React to changes in Microsoft.NETCore.App packaging structure

natemcmaster:
Part of https://github.com/aspnet/AspNetCore/issues/9720

* React to splitting of Microsoft.NETCore.App into targeting and runtime packs.
* Update to preview6 SDK

cc @dagood

### DIFF
--- a/eng/Workarounds.AfterArcade.targets
+++ b/eng/Workarounds.AfterArcade.targets
@@ -38,8 +38,20 @@
     <PackageReference Remove="xunit.runner.console" />
   </ItemGroup>
 
-  <!-- Workaround https://github.com/dotnet/arcade/pull/2867 -->
-  <PropertyGroup Condition="'$(DotNetBuildFromSource)' == 'true'">
-    <BeforePack>$(BeforePack.Replace(';_AddSourcePackageSourceLinkFile', ''))</BeforePack>
-  </PropertyGroup>
+  <!-- Workaround until we can build our repo with a 3.0 Preview 7 SDK which has https://github.com/dotnet/core-sdk/pull/2401. -->
+  <ItemGroup Condition="'$(PreReleaseVersionLabel)' == 'preview7' ">
+    <KnownFrameworkReference Update="Microsoft.NETCore.App">
+      <!--
+        Temporary until we can update to build with the Preview7 SDK .
+        Using PreReleasePreviewNumber to time-bomb this so we can remove this before RTM.
+
+        This workaround was also applied to src\ProjectTemplates\test\Infrastructure\TemplateTests.props.in.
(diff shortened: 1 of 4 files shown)


eng/Workarounds.AfterArcade.targets
```diff
@@ -38,8 +38,20 @@
...
+      <!--
+        Temporary until we can update to build with the Preview7 SDK .
+        Using PreReleasePreviewNumber to time-bomb this so we can remove this before RTM.
+
+        This workaround was also applied to src\ProjectTemplates\test\Infrastructure\TemplateTests.props.in.
+      -->
+      <RuntimePackNamePatterns>Microsoft.NETCore.App.Runtime.**RID**</RuntimePackNamePatterns>
+      <AppHostPackNamePattern />
+      <AppHostRuntimeIdentifiers />
+      <PackagesToReference />
```
natemcmaster: cc @dsplaisted @dagood - is PackagesToReference still needed? I see it's still in core-sdk, but was unclear if there is anything in the package graph of NETCore.App that we need.

dougbu:
@natemcmaster please let us know once you've resolved conflicts and the build cleans up

natemcmaster:
I'm confused about how my proposed changes resulted in this error:

> src\Caching\Abstractions\src\Microsoft.Extensions.Caching.Abstractions.csproj(0,0): error NU1101: Unable to find package Microsoft.DotNet.GenAPI. No packages exist with this id in source(s): https://dotnet.myget.org/F/roslyn/api/v3/index.json, https://dotnet.myget.org/F/sourcelink/api/v3/index.json, https://www.myget.org/F/xunit/api/v3/index.json, nuget.org

I didn't touch the list of restore sources or the package ref to Microsoft.DotNet.GenAPI.

natemcmaster:
Updating the SDK might have changed the implicit restore feeds. Adding dotnetfeed back to restore sources solved the issue.

eng/Workarounds.AfterArcade.targets
dagood: I don't know, maybe @nguerrera is also familiar with `KnownFrameworkReference`s?

dsplaisted: We are planning to get rid of it, but it may still be needed for now depending on your scenario.  The main thing it does now is bring in the Microsoft.NETCore.Platforms package, which is needed if NuGet needs to understand RuntimeIdentifiers.

natemcmaster: Good to know. This repo doesn't have any projects using RuntimeIdentifier(s), so I think we're okay.

PR #1832, opened 2019-06-12
Rename variable to MicrosoftNETCoreAppRuntimeVersion

natemcmaster:
No functional change, but I think this naming is probably better.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -23,10 +23,6 @@
   <PropertyGroup Label="Automated">
     <!-- Packages from dotnet/core-setup -->
     <MicrosoftNETCoreAppRefPackageVersion>3.0.0-preview7-27812-01</MicrosoftNETCoreAppRefPackageVersion>
-    <!-- 
-       Win-x64 is used here because we have picked an arbitrary runtime identifier to flow the version of the latest NETCore.App runtime.
-       All Runtime.$rid packages should have the same version.
-    -->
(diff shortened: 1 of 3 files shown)


natemcmaster:
Note: the naming here intentionally "MicrosoftNETCoreAppRuntimeVersion", which does not use the word "package". The plan for 3.0 RMT is to remove Microsoft.NETCore.App. Since the runtime is no longer represented by a simple NuGet package, I was hoping to stop using this concept in our naming and code.

cc @sebastienros @jkotalik

eng/Versions.props
```diff
@@ -48,6 +44,13 @@
     <!-- Packages from dotnet/arcade -->
     <MicrosoftDotNetGenAPIPackageVersion>1.0.0-beta.19309.1</MicrosoftDotNetGenAPIPackageVersion>
   </PropertyGroup>
+  <PropertyGroup>
```
jkotalik:
```suggestion
  <PropertyGroup Label="Manual" >
```

jkotalik: Oops, I see now. Why not just put this version in the manual versions?

natemcmaster: It's not really a manual package version either. Maybe 'Label="Dependency version settings"'?

jkotalik: Sure. It's super nitpicky anyways. I just want us to have the same changes.

natemcmaster: Ok, updated to add a label.

PR #1833, opened 2019-06-13
Use even more Arcade and other csproj cleanups

natemcmaster:
This removes dead code and some of the workarounds we originally put in place to minimize the changes required to adopt Arcade in this repo.

Changes:
* Arcade's default for IsPackable is false, and that .csproj files should explicitly opt-in to producing packages. I've updated all .csproj files to use this. This is a safer default which helps us avoid accidentally shipping packages we didn't mean to.
* Version settings go in eng/Versions.props
* Instead of IsFinalBuild, use p:DotNetFinalVersionKind
* Other cleanups, like removing files that were only used by KoreBuild and updating README


src/FileProviders/Embedded/src/Microsoft.Extensions.FileProviders.Embedded.csproj
```diff
@@ -4,7 +4,7 @@
     <RootNamespace>Microsoft.Extensions.FileProviders</RootNamespace>
     <Description>File provider for files in embedded resources for Microsoft.Extensions.FileProviders.</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <NuspecFile>$(MSBuildProjectName).nuspec</NuspecFile>
```
dougbu: To confirm: Arcade finds well-named .nuspec files automatically?

natemcmaster: Oops, not sure why my editor did this. Unintentional change. I’ll re-review my PR to make sure there are no other mistakes.

PR #1834, opened 2019-06-13
Disable transitive project references in test projects

natemcmaster:
Part of https://github.com/aspnet/AspNetCore/issues/11169

This should help us ensure during patches that tests are running against the latest assemblies.

cc @HaoK


davidfowl:
This is pretty terrible. Why do we need to do this? Can we do something that doesn't impact the dev experience?

natemcmaster:
The problem I’m trying to solve only happens during patching. To allow the repo to release a subset of packages, projects use package references to the major.minor.0 version when the repo is servicing.

So, scenario: we need to service Primitives in 3.0.1, but we don’t want to cascade the effect of this and reship Config. The resulting dependency graph is
```
Config.Tests —project ref—> Config (3.0.1) —pkg ref—> Config.Abstractions (3.0.0) —pkg ref—> Primitives (3.0.0)
```
 Config.Tests runs against Primitives 3.0.0, not 3.0.1. This means we can’t be confident that the patch in Primitives 3.0.1 isn’t going to break something because we’re not testing the combination of latest packages.

Solutions: 
* we could ship every package on every patch instead of trying to incrementally patch the repo. So, Primitives 3.0.1 means we have to reship Config.Abstractions, Config, Logging, etc. the cascading effect.
* we add a direct dependency from Config.Tests to Primitives. The reference resolution will resolve to a ProjectRef to Primitives (3.0.1)
* we adopt a more powerful reference system like corefx’s.

Tratcher:
Oh my. Is there any kind of tooling we can use to make sure this stays current? Or that we got all of them this time?

HaoK:
For the patch/release branches without this fix, the dev experience was an basically a mirage since green PR checks weren't actually running against the actual coherent set of latest bits, that seems like a regression disaster waiting to happen.  Until I understood what was going on, I actually 'fixed' tests in a patch PR to validate the wrong behavior due to mismatched bits.

Is it possible to limit the dev exp regression to only the servicing branches?

natemcmaster:
> DisableTransitiveProjectReferences just make it not compile otherwise

Correct.

> Is it possible to limit the dev exp regression to only the servicing branches?

Sure, but it means every time we finish a stable release we have to do this change, and then revert, or add this change in servicing only and don’t merge it back to master. Seemed cleaner to just always do this.

natemcmaster:
So, @davidfowl are you okay if we move forward with this? I'm okay if we consider alternatives in the long run, but this seemed like the least expensive short term fix we can make to ensure our servicing tests are more reliable.

natemcmaster:
I'm planning to merge this before I leave today. If there are better ideas later for how to solve the problem, I won't be offended if you revert.

FYI @JunTaoLuo

davidfowl:
Yes it’s ok to leave for now

Issue #1463, opened 2019-04-18
Remove IOptionsFactory new constraint


NinoFloris:
Why was IOptionsFactory ever released with a `new()` constraint on the interface?!
It kills a lot of useful other construction patterns where you might not have an empty constructor for a good reason, (which is why you'd want a factory interface in the first place) 

And it also makes it painful to use with F# records (which don't have an empty constructor).

I'm not sure why this constraint wasn't put onto the `IServiceCollection.Configure` method type argument?

Could anybody explain?

HaoK:
Well the original IOptions had a `new()` constraint, and the default implementation of the OptionsFactory is using new().  The factory wasn't really meant changing how TOptions instances are created (which is always done via new), but more for applying all of the configurations to the instance.

HaoK:
Its main job is to apply these:

```
OptionsFactory(IEnumerable<IConfigureOptions<TOptions>> setups, IEnumerable<IPostConfigureOptions<TOptions>> postConfigures, IEnumerable<IValidateOptions<TOptions>> validations)
```

NinoFloris:
Would there still be some room to re-evaluate this? 

Like either removing the constraint from the interface and moving it up to the Configure method.
Or possibly detect another, unconstrained, factory interface during options creation?

You would make a lot of people with uncommon use cases happy with such a change, and the F# community would love it too.

NinoFloris:
@haok ping.

I'm willing to drive this break too, investigate breaking places, fix the code, write up a breaking change issue etcetera. But 3.0 is the best time to remove this constraint if possible.

HaoK:
Feel free to give it a shot so we can see what it looks like, but no guarantees this will be something we take.

HaoK:
Addressed by https://github.com/aspnet/Extensions/pull/2169

HaoK:
@anurse this is considered a 5.0 change since it was merged to master right?

anurse:
Correct. I put it in the placeholder `5.0.0-alpha1` milestone (just to track that it was in before any public preview releases), which isn't a committment to any specific release, just a placeholder until the schedule starts to take shape :).

NinoFloris:
@anurse did you get more info on backporting to 3.0?

anurse:
[...]

PR #1843, opened 2019-06-13
Remove all new() constraints from lower Options infrastructure types

NinoFloris:
Remove all new constraints from lower Options infrastructure types and move these to to the service collection extension methods.

 - Due to the open generic service registration  the unconstrained service collection extensions would previously throw an exception during first construction of IOptions<T> (via IOptionsFactory<T>) by the container if T didn't actually have an empty constructor. This is now moved to compile time for these extension methods. (this one)
```cs
services.TryAdd(ServiceDescriptor.Transient(typeof(IOptionsFactory<>), typeof(OptionsFactory<>)));
```
- Removing the `new()` constraints on all the Options types allows for new and creative implementations of IOptionsFactory. This is mainly motivated by cases where classes should always have validation before construction and other cases where there's no empty constructor (by default) like F# records.

This is a minor source level break for invalid `T`s which seems harmless. And removing a generic constraint should be binary compatible, so in terms of impact I'd say this is pretty optimal.

/cc @HaoK

Addresses #1463

@cartermp this is a great usability enhancement for F#. 
It allows for an `ImmutableOptionsFactory` to be supplied by Giraffe or Saturn that handles records and immutable update modifications compared to having to use mutable constructorless pocos.


NinoFloris:
@eilon @haok how would I go about fixing the CI issue?

>Generated code is not up to date. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)

NinoFloris:
So tests are failing again due to reference assembly issues, yet the files have been regenerated correctly, CI issue?

Eilon:
Adding @aspnet/build

dougbu:
The generated ref assembly project file looks odd.  Was it edited manually?

Then again, it's possible regenerating anywhere but on Windows has some issues. What is your dev platform @NinoFloris?

----
FYI it's not necessary to execute eng/scripts/GenerateReferenceAssemblies.ps1 when only changing a couple of projects. If you're editing the files manually, instead run `dotnet msbuild /t:GenerateReferenceSource` in any project directory where you've changed the public API.

NinoFloris:
@dougbu I'm running on macOS, I don't have powershell there so as a matter of fact only did a single project regenerate. I did manually fix up the TFM == netstandard2.0 condition as that was completely wrong (missing half of the condition and no quotes on the other half) but the extra newlines aren't mine.

src/Options/Options/ref/Microsoft.Extensions.Options.netstandard2.0.cs
```diff
@@ -6,14 +6,14 @@ namespace Microsoft.Extensions.DependencyInjection
     public static partial class OptionsServiceCollectionExtensions
     {
         public static Microsoft.Extensions.DependencyInjection.IServiceCollection AddOptions(this Microsoft.Extensions.DependencyInjection.IServiceCollection services) { throw null; }
-        public static Microsoft.Extensions.Options.OptionsBuilder<TOptions> AddOptions<TOptions>(this Microsoft.Extensions.DependencyInjection.IServiceCollection services) where TOptions : class { throw null; }
-        public static Microsoft.Extensions.Options.OptionsBuilder<TOptions> AddOptions<TOptions>(this Microsoft.Extensions.DependencyInjection.IServiceCollection services, string name) where TOptions : class { throw null; }
-        public static Microsoft.Extensions.DependencyInjection.IServiceCollection ConfigureAll<TOptions>(this Microsoft.Extensions.DependencyInjection.IServiceCollection services, System.Action<TOptions> configureOptions) where TOptions : class { throw null; }
+        public static Microsoft.Extensions.Options.OptionsBuilder<TOptions> AddOptions<TOptions>(this Microsoft.Extensions.DependencyInjection.IServiceCollection services) where TOptions : class, new() { throw null; }
```
pakrym: This is a breaking change and I don't think this is required.

pakrym:
I don't think we need to add `new()` constraint on extension methods as it makes things even more limiting than they were. Removing internal constrains on the other hand is good and makes IOptionsFactory slightly more of extensibility point than it is now.

src/Options/Options/ref/Microsoft.Extensions.Options.csproj
```diff
@@ -1,3 +1,4 @@
+
```
dougbu: Bet if you reverted all changes to this file, the Code Check job would pass. But, I lean toward a lot of this being an unnecessary breaking change (as @pakrym described).

HaoK:
We took https://github.com/aspnet/Extensions/pull/2169 instead

NinoFloris:
These notifications have slipped through, sorry, happy to see you got the fix in

PR #1845, opened 2019-06-14
Set ReferenceAssemblyAttribute in reference assemblies

natemcmaster:
Reference assemblies in .NET are supposed to have these two attributes, which prevent the runtime from loading and executing them.

For other examples of this, see https://github.com/dotnet/corefx/blob/41489a93acf3f36abcaaaea2003a8fdbb577cf35/eng/ReferenceAssemblies.props#L23-L28
and https://github.com/dotnet/standard/blob/14b6385e8640328e33883a919eca11594e22f21c/eng/versioning.targets#L48-L53

### DIFF
--- a/src/Shared/src/ReferenceAssemblyInfo.cs
+++ b/src/Shared/src/ReferenceAssemblyInfo.cs
@@ -1,0 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+// All reference assemblies should have the 0x70 flag which prevents them from loading
+// and the ReferenceAssemblyAttribute.
+[assembly:System.Runtime.CompilerServices.ReferenceAssembly]
+[assembly:System.Reflection.AssemblyFlags((System.Reflection.AssemblyNameFlags)0x70)]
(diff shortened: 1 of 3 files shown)


src/Shared/src/ReferenceAssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+// All reference assemblies should have the 0x70 flag which prevents them from loading
+// and the ReferenceAssemblyAttribute.
+[assembly:System.Runtime.CompilerServices.ReferenceAssembly]
+[assembly:System.Reflection.AssemblyFlags((System.Reflection.AssemblyNameFlags)0x70)]
```
davidfowl: What flags are these and why can't we use the enum?

natemcmaster: This flag sets AssemblyName.ProcessorArchitecture to None. There is no public API for this. https://github.com/dotnet/coreclr/blob/64ca544ecf55490675e72b853e98ebc8cc75a4fe/src/System.Private.CoreLib/src/System/Reflection/AssemblyName.CoreCLR.cs#L74

davidfowl: Add a comment and link to the source in our source?

natemcmaster: Done

PR #1856, opened 2019-06-17
Status property added

grahamehorner:
BackgroundService _executingTask.Status

### DIFF
--- a/src/Hosting/Abstractions/src/BackgroundService.cs
+++ b/src/Hosting/Abstractions/src/BackgroundService.cs
@@ -27,9 +27,11 @@ namespace Microsoft.Extensions.Hosting
         /// This property give information about the current status of the background service task,
         /// it allows the developer to see if the service is running or has faulted etc. handy for
         /// things like watchdog processes etc.
+        /// TaskStatus of Created may not be correct but it's the nearest value that represents the
+        /// TaskStatus when _executingTask is null
         /// </summary>
         public TaskStatus Status => 
-            _executingTask.Status;
+            _executingTask?.Status ?? TaskStatus.Created;
 
         /// <summary>
         /// Triggered when the application host is ready to start the service.
--- a/src/Hosting/Abstractions/src/BackgroundService.cs
+++ b/src/Hosting/Abstractions/src/BackgroundService.cs
@@ -24,6 +24,14 @@ namespace Microsoft.Extensions.Hosting
         protected abstract Task ExecuteAsync(CancellationToken stoppingToken);
 
         /// <summary>
+        /// This property give information about the current status of the background service task,
+        /// it allows the developer to see if the service is running or has faulted etc. handy for
+        /// things like watchdog processes etc.
+        /// </summary>
+        public TaskStatus Status => 
+            _executingTask.Status;
+
+        /// <summary>
         /// Triggered when the application host is ready to start the service.
         /// </summary>
         /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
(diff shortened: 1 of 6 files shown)


src/Hosting/Abstractions/src/BackgroundService.cs
```diff
@@ -23,6 +23,14 @@ public abstract class BackgroundService : IHostedService, IDisposable
         /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
         protected abstract Task ExecuteAsync(CancellationToken stoppingToken);
 
+        /// <summary>
+        /// This property give information about the current status of the background service task,
+        /// it allows the developer to see if the service is running or has faulted etc. handy for
+        /// things like watchdog processes etc.
+        /// </summary>
+        public TaskStatus Status =>
```
Tratcher: 1. This will null ref if StartAsync hasn't been called yet.
2. Even if it does fault, you can't see the exception.

Where do you expect this to be called from?

gfoidl:
For reference -> https://github.com/aspnet/Extensions/issues/1857

src/Hosting/Abstractions/src/BackgroundService.cs
rynowak: TaskStatus is also a pretty confusing enum. I'd want to know more about the expected usage of this.

grahamehorner: I would like to determine when a background service that has been registered with dependency injection is currently running or in the process of starting/shutting down or has stopped so I may perform some action(s) accordingly; for example I have a requirement to check that a number of services have ran to completion on a scheduled basis and report status accordingly; I also have a need to stop/restart a number running services as when a configuration change is detected.

Tratcher: BackgroundService isn't designed to be restarted. Do you mean restarting the entire host?

How do you expect to actually get to this type to call it? Resolve all of the IHostedSerivces and downcast?

grahamehorner: I would have expected that a developer be able to restart a BackgroundService? why/what would prevent the restarting of the BackgroundService task? I would typically access implementations of BackgroundService(s) using the interface or concrete implementation possibly via an implementation of something like a IHostedServiceManager and a collection of helper extension methods.  @Tratcher are you able to explain why BackgroundService aren't restartable ?

Tratcher: The contract for IHostedService is single use, managed entirely by the host, and there's no expectation that the application interact with them directly.

BackgroundService specifically isn't safe for restarting primarily due to the absence of any thread safety. It only expects to be called from the host in a thread safe way. Attempting control it from elsewhere in the app risks race conditions, deadlocks, etc..

Before continuing I suggest you go back and make an end-to-end sample of how you want to interact with hosted services.

grahamehorner: @davidfowl can you give an example of your use-case and/or pointer to work/enhancement you have mentioned; I'd like to fully understand the restrictions and/or the general views around a number of use cases where I believe the need for the BackgroundService task status is required to be exposed; irrespective of the ability starting/restarting/stopping the BackgroudService

anurse:
This got a bit stale, and I think we want to discuss it in more detail back on the issue linked (https://github.com/aspnet/Extensions/issues/1857). For example, how does someone get access to this `BackgroundService` to access the property? Downcast from `IHostedService`? Perhaps it would make more sense to expose status information as a first class concept.

PR #1864, opened 2019-06-18
Caching ResourceManagerStringLocalizer creation

### LINKED ISSUE #1264: ResourceManagerStringLocalizerFactory allocates too much
Here is an allocations snapshot for 50 requests of a sample web page rendering a blog post. It allocates more attributes than strings. More than 3MB of allocations for 50 requests (it depends on how many times the service is resolved).

image

The issue is in the `Create` method of this factory, which is called for every component that resolves an `IStringLocalizer` from DI, the result is cached, but creating the key of the cache entry is doing uncached reflection.

We could either cache by the type itself, or cache the reflection.

sebastienros:
Fixes #1264

As per the issue details, the current code is doing too much reflection. 
Here the `_localizerCache` is reused to add entries for the `Type` based initialization. The keys can't collide with the other usage of the dictionary.
`AssemblyQualifiedName` is used as the key to isolate all types in the dictionary.

### DIFF
--- a/src/Localization/Localization/src/ResourceManagerStringLocalizerFactory.cs
+++ b/src/Localization/Localization/src/ResourceManagerStringLocalizerFactory.cs
@@ -156,7 +156,7 @@ namespace Microsoft.Extensions.Localization
                 var assembly = typeInfo.Assembly;
 
                 return CreateResourceManagerStringLocalizer(assembly, baseName);
-            }
+            });
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


davidfowl:
@sebastienros Can you run a profile on this code so we can see the difference ?

ryanbrandenburg:
Did we ever get our profile proof?

JunTaoLuo:
Note that some sources edited in this PR has migrated to dotnet/aspnetcore. A new PR with the same changes for the affected sources should be opened in dotnet/aspnetcore. See https://github.com/dotnet/extensions/pull/2974 for details.

Issue #649, opened 2018-07-16
EventLogLogger Doesn't Dispose of System.Diagnostics.EventLog


jeff-bacon:
Microsoft.Extensions.Logging.EventLog.WindowsEventLog.cs creates a System.Diagnostics.EventLog which is IDisposable, but WindowsEventLog itself is not IDisposable.  This could be a potential memory leak.

davidfowl:
Fixed

PR #1865, opened 2019-06-18
Dispose EventLog in WindowsEventLog

Kahbazi:
Addresses #649

### DIFF
--- a/src/Logging/Logging.EventLog/src/EventLogLoggerProvider.cs
+++ b/src/Logging/Logging.EventLog/src/EventLogLoggerProvider.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Logging.EventLog
@@ -49,7 +50,10 @@ namespace Microsoft.Extensions.Logging.EventLog
 
         public void Dispose()
         {
-            _settings.Dispose();
+            if (_settings.EventLog is IDisposable disposableEventLog)
+            {
+                disposableEventLog.Dispose();
+            }
         }
 
         public void SetScopeProvider(IExternalScopeProvider scopeProvider)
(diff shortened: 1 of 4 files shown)


pakrym:
This is not the right fix. Recreating event log for every message is very inefficient.

davidfowl:
> This is not the right fix. Recreating event log for every message is very inefficient.

Right, the fix is to make the ILoggerProvider dispose any disposable resources the loggers might use.

src/Logging/Logging.EventLog/src/EventLogLoggerProvider.cs
```diff
@@ -49,6 +49,7 @@ public ILogger CreateLogger(string name)
 
         public void Dispose()
         {
+            _settings.Dispose();
```
davidfowl: Instead of making settings disposable can't we access the internal EventLog property here directly and dispose it?

Kahbazi: That was my first thought, but I change it because I thought every class should dispose its resources in case in future they have more resource to dispose.

Kahbazi:
Should I change `EventLogSettings.EventLog` type from `IEventLog` to `WindowsEventLog`?

src/Logging/Logging.EventLog/src/EventLogLoggerProvider.cs
davidfowl: I rather not make a settings type publicly disposable since the IEventLog being stored there is an implementation detail.

davidfowl:
> Should I change EventLogSettings.EventLog type from IEventLog to WindowsEventLog?

No it's an IEventLog so the tests can override it. Doing the cast is fine.

src/Logging/Logging.EventLog/src/EventLogSettings.cs
```diff
@@ -40,6 +40,14 @@ internal IEventLog EventLog
             set => _eventLog = value;
         }
 
+        public void Dispose()
+        {
+            if (EventLog is IDisposable disposableEventLog)
```
davidfowl: Move this to the provider

Kahbazi: Done.

Issue #1768, opened 2019-05-30
Change TryAdd IServiceCollection extensions to return "something"


DanielBryars:
### Is your feature request related to a problem? Please describe.

All the Try variants of the Add, AddTransient, AddSingleton etc return void: [Source](src/DependencyInjection/DI.Abstractions/src/Extensions/ServiceCollectionDescriptorExtensions.cs).

This means I can't chain the methods when building my dependencies, like I can with AddTransient<> etc.

This means it breaks the standard pattern TryXXX which returns a bool if it can't do what it's trying.

Returning void is the worst of both worlds.

### Describe the solution you'd like
Return IServiceCollection so that you can do the fluent method chaining.

### Describe alternatives you've considered
Return a bool. I don't like this, because there's very little you'd do with the answer. 

### Additional context
See my question on [so](https://stackoverflow.com/questions/56376546/why-doesnt-tryaddtransient-return-an-iservicecollection?noredirect=1#comment99354949_56376546) for some more discussion.

grahamehorner:
How about a TryGetOrAdd<TService>(impl) if the TService is not registered then it is added, otherwise the registered implementation is returned.

grahamehorner:
I would like to see an optional out bool added to the Try methods, that allows later evaluating the result if needed as well as the return type change to IServiceColletion. I'm not that bothered if the extension methods names differ slightly, but do agree flunte chain would be good.

khellang:
This is kinda funny. To quote myself from https://github.com/aspnet/DependencyInjection/pull/353#issuecomment-178206877 and https://github.com/aspnet/DependencyInjection/issues/447#issuecomment-248633002:

> Trust me - this'll come up time and time again :wink:

I doubt anything will happen here 😅

PR #1866, opened 2019-06-18
Return IServiceCollection in TryAdd methods

Kahbazi:
Addresses #1768


Eilon:
FYI this is a breaking change: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#signatures

Breaking changes are permitted in major versions (which this is), but need to be scrutinized.

Kahbazi:
This technically won't change return type, only adds one since they were all void and no code would be break with this update. Can we consider an exception here?

Eilon:
The return 'type' was `void` and is now something else. Yes, `System.Void` is a real type! Such a change is not *likely* to break code, but it certainly *can*. I'm not sure if it can break binary compatibility, but it can definitely cause compilation changes/breaks and runtime changes/breaks.

Ultimately the question that will be asked is: does the benefit of this change outweigh the cost.

davidfowl:
We're not going to make this breaking change.

PR #1871, opened 2019-06-19
fix #1723 by converting sql cache to new sqlclient

### LINKED ISSUE #1723: Consider changing Microsoft.Extensions.Caching.SqlServer to depend on Microsoft.Data.SqlClient and/or making it OOB
Creating this as a follow up from https://github.com/aspnet/AspNetCore/issues/10013#issuecomment-492811495. 

Currently Microsoft.Extensions.Caching.SqlServer depends on System.Data.SqlClient, which is still supported, but the new provider Microsoft.Data.SqlClient is where all new features of SQL Server will be supported. In the 3.0 time frame the main functional difference is Always Encrypted support in .NET Core.

I haven't checked if this ships only as a package or as part of the shared framework.

anurse:
This changes the `Microsoft.Extensions.Caching.SqlServer` package to use `Microsoft.Data.SqlClient`.

There is a big open question to be resolved though: The `Microsoft.Data.SqlClient` package does not currently support **.NET Framework 32-bit**. The tests run fine if forced to use 64-bit but if your app runs in 32-bit you **will** get `BadImageFormatException`s from `Microsoft.Data.SqlClient`

According to [the README](https://github.com/dotnet/SqlClient/blob/master/README.md) for the new SQL client, the plan is to support both 32 and 64 bit, so maybe this is OK, but I want Important People:tm: to know and be OK with this :).

**To be clear**: This is the support matrix:

| Arch \ Framework | **.NET Framework** | **.NET Core** |
| - | - | - |
| **x86** | `BadImageFormatException` | _Works_ |
| **x64** | _Works_ | _Works_ |

Fixes #1723

### DIFF
--- a/eng/tools/BaselineGenerator/Program.cs
+++ b/eng/tools/BaselineGenerator/Program.cs
@@ -152,6 +152,7 @@ namespace PackageBaselineGenerator
                 OmitXmlDeclaration = true,
                 Encoding = Encoding.UTF8,
                 Indent = true,
+                NewLineChars = "\n", // The files use LF, not CRLF.
             };
 
             using (var writer = XmlWriter.Create(output, settings))
@@ -212,6 +213,7 @@ namespace PackageBaselineGenerator
                     NewLineOnAttributes = false,
                     OmitXmlDeclaration = true,
                     WriteEndDocumentOnClose = true,
+                    NewLineChars = "\n", // The file uses LF, not CRLF!
                 };
 
                 using (var stream = File.OpenWrite(documentPath))
--- a/src/Caching/SqlServer/src/Microsoft.Extensions.Caching.SqlServer.csproj
+++ b/src/Caching/SqlServer/src/Microsoft.Extensions.Caching.SqlServer.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Distributed cache implementation of Microsoft.Extensions.Caching.Distributed.IDistributedCache using Microsoft SQL Server.</Description>
@@ -17,7 +17,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.Extensions.Caching.Abstractions" />
     <Reference Include="Microsoft.Extensions.Options" />
-    <Reference Include="System.Data.SqlClient" />
+    <Reference Include="Microsoft.Data.SqlClient" />
   </ItemGroup>
 
 </Project>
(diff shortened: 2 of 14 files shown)


anurse:
Ran the Database Functional tests manually. They all pass in `netcoreapp3.0` and when running `dotnet vstest --Platform:x64 --Framework:".NETFramework,Version=4.7.2" ..\..\..\..\artifacts\bin\Microsoft.Extensions.Caching.SqlServer.Tests\Debug\net472\Microsoft.Extensions.Caching.SqlServer.Tests.dll` (to force 64-bit .NET Framework). When running in VS, the `net472` tests run in 32-bit and thus fail.

Passing Database Functional Tests for `netcoreapp3.0` target

eng/tools/BaselineGenerator/Program.cs
```diff
@@ -212,6 +213,7 @@ private async Task<int> Run()
                     NewLineOnAttributes = false,
                     OmitXmlDeclaration = true,
                     WriteEndDocumentOnClose = true,
+                    NewLineChars = "\n", // The file uses LF, not CRLF!
```
anurse: I didn't end up needing to use this tool but when I first tried it, it changed every newline!

anurse:
Ping. It looks like we're going to go ahead and ship this with the limitation above. Please review so we can get this in for preview 7 and I can announce it (since it's a breaking change).

Tratcher:
Ref assemblies...

anurse:
Leslie Knope from the TV show "Parks and Recreation" glaring at a person off-camera

src/Caching/SqlServer/src/Microsoft.Extensions.Caching.SqlServer.csproj
```diff
@@ -1,9 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Distributed cache implementation of Microsoft.Extensions.Caching.Distributed.IDistributedCache using Microsoft SQL Server.</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <NoWarn>$(NoWarn);CS1591;PKG0001</NoWarn>
```
dougbu: Just curious: Where does this warning come from?

anurse: Removing the `System.Data.SqlClient` dependency is considered (rightly so) a breaking change:

```
C:\Users\anurse\code\aspnet\Extensions\eng\targets\ResolveReferences.targets(140,5): error PKG0001: Package references changed since the last release. This could be a breaking change. References removed: [C:\Users\anurse\code\aspnet\Extensions\src\Caching\SqlServer\src\Microsoft.Extensions.Caching.SqlServer.csproj]
C:\Users\anurse\code\aspnet\Extensions\eng\targets\ResolveReferences.targets(140,5): error PKG0001:  - System.Data.SqlClient [C:\Users\anurse\code\aspnet\Extensions\src\Caching\SqlServer\src\Microsoft.Extensions.Caching.SqlServer.csproj]
```

Prior art (other projects in this repo) suggests that suppressing is the way to handle that.

anurse:
> What prevents removing the System.Data.SqlClient dependency in this repo?

Nothing I think, will do when I fix the ref asms

anurse:
Rebasing and resolving conflicts...

anurse:
Announced: https://github.com/aspnet/Announcements/issues/370

PR #1874, opened 2019-06-21
[WIP] Strongly typed and accessible generic LogValues

Scooletz:
This PR introduces a new public API for `LogValues` that enables visiting them and obtaining values in a non-allocating manner (no calls to `.ToArray` that boxes structs into objects).This potentially, could enable some loggers to do not call `formatter` at all, and just serialize the call or at least use their generic methods (like `Serilog` https://github.com/serilog/serilog/blob/dev/src/Serilog/ILogger.cs#L99-L121 )

New APIs that are added:
 - `ILogValues` interface introduced, for all `LogValues` structs
 - `ITypedLogger` interface introduced

Remarks:
- this PR targets a non-exceptional flow of logging (`formatter` is still called for an exception)
- benchmarks do not include any kind of data processing, no serialization so they are skewed in a way that presents this work as removing a lot of overhead. In reality, it strongly depends on the implementation of the `ITypedLogger` and the data processing that is done in it (how it stores the data).

#### Benchmarks

I amended `FormattingBenchmarks` from the `Microsoft.Extensions.Logging.Performance` suite adding a non-formatting cases, where the new interface is used. The new cases are using `_Raw` suffix.

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17134
Intel Core i7-6700HQ CPU 2.60GHz (Skylake), 1 CPU, 8 logical cores and 4 physical cores
Frequency=2531247 Hz, Resolution=395.0622 ns, Timer=TSC
.NET Core SDK=3.0.100-preview6-012264
  [Host]     : .NET Core 3.0.0-preview7-27819-08 (CoreCLR 4.700.19.31804, CoreFX 4.700.19.31904), 64bit RyuJIT
  Job-GHYGVD : .NET Core 3.0.0-preview7-27819-08 (CoreCLR 4.700.19.31804, CoreFX 4.700.19.31904), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 3.0  
RunStrategy=Throughput  

```
|                       Method |      Mean |    Error |    StdDev |    Median |         Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|----------------------------- |----------:|---------:|----------:|----------:|-------------:|-------:|---------:|-------:|----------:|
|      TwoArguments_Formatting | 425.23 ns | 8.544 ns | 24.377 ns | 424.15 ns |  2,351,693.1 |   9.07 |     1.05 | 0.0038 |      80 B |
|             **TwoArguments_Raw**| 423.76 ns | 9.558 ns | 28.032 ns | 421.94 ns |  2,359,801.8 |   9.04 |     1.08 | 0.0038 |      80 B |
| TwoArguments_NoEx_Formatting | 399.61 ns | 9.939 ns | 28.196 ns | 402.59 ns |  2,502,462.4 |   8.53 |     1.04 | 0.0038 |      80 B |
|        **TwoArguments_NoEx_Raw** | 150.41 ns | 3.127 ns |  9.172 ns | 147.53 ns |  6,648,683.8 |   3.21 |     0.38 |      - |       0 B |
|                  NoArguments |  47.33 ns | 1.622 ns |  4.782 ns |  46.70 ns | 21,126,660.6 |   1.00 |     0.00 |      - |       0 B |


Scooletz:
Could sb help me with understanding the build failure?

gfoidl:
> Could sb help me with understanding the build failure?

you changed the public API, so the reference assembly need to be regenerated.

```
Generated code is not up to date. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)
```
So follow the steps from [here](https://github.com/aspnet/Extensions/blob/master/docs/ReferenceAssemblies.md#when-changing-public-api) to regenerate that assembly and check it in.

src/Logging/Logging.Abstractions/src/ILogValues.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Provides an interface for a state of <see cref="ILogger.Log{TState}"/> that wants to make their values visitable without boxing them to objects
```
gfoidl: Nit:
```suggestion
    /// Provides an interface for a state of <see cref="ILogger.Log{TState}"/> that wants to make their values visitable without boxing them to objects.
```

src/Logging/Logging.Abstractions/src/ILogValues.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// The original format of the log.
+        /// </summary>
+        string OriginalFormat { get; }
+
+        /// <summary>
+        /// Visits values of the log entry.
+        /// </summary>
+        /// <typeparam name="TVisitor">The type of the visitor.</typeparam>
+        /// <param name="visitor">The visitor to visit log values.</param>
+        void Accept<TVisitor>(ref TVisitor visitor)
```
gfoidl: Why is the `visitor` passed via `ref`?
If there's any strong reason therefore, please add a comment.

davidfowl:
> This potentially, could enable some loggers to do not call formatter at all, and just serialize the call.

This PR needs to be a full end to end with performance numbers to show the proof of concept.

Scooletz:
Thank you for chiming in @davidfowl @gfoidl 
Let me mark it as a draft and undraft it as soon as I have, both the right impl and numbers!

pakrym:
> This potentially, could enable some loggers to do not call formatter at all, and just serialize the call or

This is already possible using `IReadOnlyList<KVP>`.

Can you show a sample logger implementation that leverages this new feature.

Scooletz:
@pakrym `IReadOnlyList<KVP>` is boxing values to objects as KVP is `KeyValuePair<string,object>`. It'd be nice (if possible) to have a non-boxing alternative for values that are structs to be passed to loggers like `Serilog` https://github.com/serilog/serilog/blob/dev/src/Serilog/ILogger.cs#L99-L121 and leave the decision about boxing to the logger itself.

This is still WIP as I hit some limitations. I renamed the PR.

davidfowl:
You should fully implement the scenario in one of the loggers so we can see (with some performance numbers for before and after).

Scooletz:
I see. Thank you @davidfowl I provided initial benchmarks in the PR but will work on the real world scenario.

lodejard:
I'm a little confused. It seems like the strongly-typed values in Serilog do not extend more than the implementation of ILogger - just long enough to detect when IsEnabled is false, otherwise the values are immediately boxed into a new object[].

https://github.com/serilog/serilog/blob/8066ecff50ae04858e2d4f0c124471d8b02eedc6/src/Serilog/Core/Logger.cs#L247

Given that, is the goal just to avoid allocations and boxing until the destination like Serilog can decide if IsEnabled is true or not? If that's the case - that is already possible because the calls to `ILogger.Log<TState>` have zero allocations when TState is a struct like the one `LoggerMessage.Define` is based on.

Scooletz:
This is correct.@lodejard I started working on a PR for Serilog that pushes it down, without boxing but because of my limited availability I dropped it https://github.com/serilog/serilog/pull/1332. 
If there was a logger, being able to dissect the state in the `ILogger.Log<TState>` though, having these types public could enable building a faster one.

Anyway, I'm closing this for now.

PR #1879, opened 2019-06-25
Allow sanitation of dotnet interop exceptions

javiercn:
Adds a hook that gets called before serializing a .NET exception and that allows implementations to customize the object that gets sent to the client.

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
@@ -93,8 +93,6 @@ namespace Microsoft.JSInterop
             catch (Exception ex)
             {
                 syncException = ExceptionDispatchInfo.Capture(ex);
-                syncException.SourceException.Data.Add(JSRuntimeBase.AssemblyNameKey, assemblyName);
-                syncException.SourceException.Data.Add(JSRuntimeBase.MethodKey, methodIdentifier);
             }
 
             // If there was no callId, the caller does not want to be notified about the result
@@ -116,8 +114,6 @@ namespace Microsoft.JSInterop
                     if (t.Exception != null)
                     {
                         var exception = t.Exception.GetBaseException();
-                        exception.Data.Add(JSRuntimeBase.AssemblyNameKey, assemblyName);
-                        exception.Data.Add(JSRuntimeBase.MethodKey, methodIdentifier);
 
                         jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, ExceptionDispatchInfo.Capture(exception), assemblyName, methodIdentifier);
                     }
--- a/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
@@ -105,7 +105,7 @@ namespace Microsoft.JSInterop
             else if (syncException != null)
             {
                 // Threw synchronously, let's respond.
-                jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, syncException);
+                jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, syncException, assemblyName, methodIdentifier);
             }
             else if (syncResult is Task task)
             {
@@ -119,16 +119,16 @@ namespace Microsoft.JSInterop
                         exception.Data.Add(JSRuntimeBase.AssemblyNameKey, assemblyName);
                         exception.Data.Add(JSRuntimeBase.MethodKey, methodIdentifier);
 
-                        jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, ExceptionDispatchInfo.Capture(exception));
+                        jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, ExceptionDispatchInfo.Capture(exception), assemblyName, methodIdentifier);
                     }
 
                     var result = TaskGenericsUtil.GetTaskResult(task);
-                    jsRuntimeBaseInstance.EndInvokeDotNet(callId, true, result);
+                    jsRuntimeBaseInstance.EndInvokeDotNet(callId, true, result, assemblyName, methodIdentifier);
                 }, TaskScheduler.Current);
             }
             else
             {
-                jsRuntimeBaseInstance.EndInvokeDotNet(callId, true, syncResult);
+                jsRuntimeBaseInstance.EndInvokeDotNet(callId, true, syncResult, assemblyName, methodIdentifier);
             }
         }
 
--- a/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
@@ -93,6 +93,8 @@ namespace Microsoft.JSInterop
             catch (Exception ex)
             {
                 syncException = ExceptionDispatchInfo.Capture(ex);
+                syncException.SourceException.Data.Add(JSRuntimeBase.AssemblyNameKey, assemblyName);
+                syncException.SourceException.Data.Add(JSRuntimeBase.MethodKey, methodIdentifier);
             }
 
             // If there was no callId, the caller does not want to be notified about the result
@@ -114,6 +116,9 @@ namespace Microsoft.JSInterop
                     if (t.Exception != null)
                     {
                         var exception = t.Exception.GetBaseException();
+                        exception.Data.Add(JSRuntimeBase.AssemblyNameKey, assemblyName);
+                        exception.Data.Add(JSRuntimeBase.MethodKey, methodIdentifier);
+
                         jsRuntimeBaseInstance.EndInvokeDotNet(callId, false, ExceptionDispatchInfo.Capture(exception));
                     }
 
--- a/src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.ComponentModel;
 using System.Runtime.ExceptionServices;
 using System.Text.Json;
 using System.Threading;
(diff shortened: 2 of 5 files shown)


src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -62,18 +66,35 @@ public Task<T> InvokeAsync<T>(string identifier, params object[] args)
...
+        /// </summary>
+        /// <remarks>
+        /// This callback can be used in remote JS interop scenarios to sanitize exceptions that happen on the server to avoid disclosing
+        /// sensitive information to remote browser clients.
+        /// </remarks>
+        /// <param name="exception">The exception that occurred.</param>
+        /// <param name="assemblyName">The assembly for the invoked .NET method.</param>
+        /// <param name="methodIdentifier">The identifier for the invoked .NET method.</param>
+        /// <returns>An object containing information about the exception.</returns>
+        protected virtual object OnDotNetInvocationException(Exception exception, string assemblyName, string methodIdentifier) => exception;
```
javiercn: Remote JS runtime will override this and sanitize exceptions before serializing them.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -93,6 +93,8 @@ public static void BeginInvoke(string callId, string assemblyName, string method
             catch (Exception ex)
             {
                 syncException = ExceptionDispatchInfo.Capture(ex);
+                syncException.SourceException.Data.Add(JSRuntimeBase.AssemblyNameKey, assemblyName);
+                syncException.SourceException.Data.Add(JSRuntimeBase.MethodKey, methodIdentifier);
```
SteveSandersonMS: This mechanism seems kind of elaborate. What's the reason for not adding `assemblyName` and `methodIdentifier` as two extra params to `EndInvokeDotNet`?

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
SteveSandersonMS: Since the TS-side code expects the payload to be a `string` whenever `success == false`, would it be safer to declare the return type of this method to be `string`? Is there any scenario where we want it to be something else?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
javiercn: Yeah, lets do that, I thought end invoke dotnet was part of the public contract or called across boundaries, I realize now it's not, so I'll just do that.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: Whatever it is, will end up being serialized into a string. I don't want to limit ourselves in the future from returning an object here with a bit more information.

For example, the idea is we will return an error message crafted from the server, but we might as well return an object with Assembly, Method, Description and have the client build the actual error.

In the end the client receives a string that might contain json in it, does that make sense?

SteveSandersonMS: Ok, makes sense!

SteveSandersonMS:
I’m not so sure about using the non-void event for this. Events don’t provide any indication that only one thing can subscribe to them. This would be super confusing if people try to configure multiple listeners. 

I preferred the virtual method you had before.

javiercn:
@SteveSandersonMS I ended up doing something different. I didn't feel well exposing a public method on JSRuntime so I ended up adding an event to DotNetDispatcher that gets fired when an exception occurs.

We will simply register an event handler into DotNetDispatcher that sanitizes the exception in the circuit case.

I also have a crazy thought where DotNetDispatcher is just an implementation detail of how we allow .NET methods to be invoked through JS Interop and that it could live in the ASP.NET Core repo under the /shared folder and be referenced by client-side and server-side blazor independently.

The advantage is that it could be an internal class and we wouldn't be limited to change it in the future.

SteveSandersonMS:
We might be getting a bit off-track with this. `DotNetDispatcher` is meant to be in the JSInterop package and not tied to ASP.NET Core specifically so it can be consumed by alternate .NET WebAssembly UI frameworks.

> I didn't feel well exposing a public method on JSRuntime

It would be a `protected` method on `JSRuntimeBase`, wouldn't it? That seems like the natural fit for what we're doing here. An `event`, on the other hand, is pretty confusing because it comes with the implication that you can have multiple listeners.

javiercn:
> It would be a `protected` method on `JSRuntimeBase`, wouldn't it?

When I mean public I mean not internal/private. If you look at the API surface for JSRuntime it doesn't make any mention of .NET interop, we simply chose to implement .net interop that way.

For future reference, in general when I say public I mean public/protected something that a client can get to use either directly or through inheritance, so its something we need to account for the future in terms of breaking changes.

You are right that the event is something that can have multiple listeners, but I also don't think that its going to be a problem.

SteveSandersonMS:
> You are right that the event is something that can have multiple listeners, but I also don't think that its going to be a problem.

I think it's super weird to use an event for this, and I don't know why we'd choose to use that given that there is a natural way to model this with a virtual method.

As for expanding the API surface, I'm not suggesting we need to add anything to the `IJSRuntime` interface - the method can go onto `JSRuntimeBase` which is only seen by people implementing `IJSRuntime`.

javiercn:
@SteveSandersonMS How about a property with a delegate you can set. That's not an event and achieves the same purpose.

SteveSandersonMS:
We generally try to avoid using statics for configuration (as I'm sure you know - sorry if that sounds patronizing). But I recognize there isn't a good way of turning `DotNetDispatcher` into an instance class, or adding another DI service for JS interop exception sanitization, without also doing more async-local stuff given that the incoming calls have to start by hitting a static method for WASM purposes. I don't really want to add more async-local stuff either!

Perhaps the static settable delegate property you're suggesting is the best compromise. **If we do that, could you please also make it enforce the rule "you can only set this once ever"?** That rule will help people not to create bizarre situations where they use this as some kind of extra statefulness. Since the framework is always going to try to set that property, people won't be able to set it themselves either before or after because it would throw if they did, so that helps a lot.

@rynowak Please speak up if you have any concerns about this compromise. It is feasible to have an alternative, but it involves doing another async-local thing probably.

javiercn:
I think the static property is a good compromise. I was planning to add the set once behavior.

rynowak:
> Please speak up if you have any concerns about this compromise. It is feasible to have an alternative, but it involves doing another async-local thing probably.

Sorry for being late to the party, but I'm really not sure why we wouldn't bake this behaviour into `IJsRuntime` as a method to implement, or into an implementation of `IJsRuntime`.

For instance all of the exceptions that this code handles and sends back to JS *come from* a call into the JS runtime to be begin with.

I really see a static property as the absolute worst-case last-resort kind of thing. I don't think I've seen enough discussion or attempts at better solutions to convince me that's where we are.

SteveSandersonMS:
The reasoning is that IJsRuntime is a contract about .NET->JS interop. The contract doesn’t even imply that calls in the other direction exist. So it would be strange for it to have a method for dealing with exceptions on calls in the other direction. 

JsRuntimeBase is a particular implementation that internally knows about calls from JS to .NET, but its public API surface doesn’t mention that. Publicly, it is only offering a service to .NET code so it can call JS. The fact that internally it’s where the other side also lives is a private implementation detail. 

All this said, I don’t feel ultra strongly. We could reason that JsRuntimeBase is a concrete type that *is* involved in JS-to-.NET calls even if that’s not mentioned on the public API yet, and hence just put a protected virtual method for this on it (as suggested above). 

I am also willing to accept the static delegate on the grounds that we can ensure this only gets used by the framework. It’s not a delightful compromise, hence this whole discussion.

javiercn:
@SteveSandersonMS @rynowak alternative approach, what if instead of adding a static property we simply pass an additional callback as argument to DotNetDispatcher.BeginInvoke that is the "exception sanitizer" that way is not a static and doesn't have to be in JSRuntime base either.

I'm happy going back to having it on JSRuntime too, I like this more because it doesn't have the drawback of the static and it also doesn't spread the .NET concept to the JS side of things, but I'll leave it up to you as I don't have strong feelings either way.

Thoughts?

SteveSandersonMS:
@javiercn I think that would be viable, but TBH I lean towards preferring the `protected virtual` method on `JSRuntimeBase`.

My reasoning is that exception sanitization is not really an obvious thing to be specifying on a per-call basis to `BeginInvoke` - it doesn't feel natural that the developer would be thinking they should supply this information at that point (they just want to invoke some method). If we routinely added more parameters to `BeginInvoke` each time we thought of another thing you might want to control, it would quickly get out of hand.

So for long-term sanity, I'd be more comfortable hiving this concern off into its own separate virtual method.

javiercn:
@stevesandersonms then jsruntime base it is, although as I said, no developer will invoke this ever as it is infrastructure for the platform.

But for that same reason I don’t care it is on jsruntimebase either

rynowak:
> doesn't spread the .NET concept to the JS side of things, but I'll leave it up to you as I don't have strong feelings either wa

I don't really see this as a drawback? 

It's obvious why the JS -> .NET infrastructure isn't there for you to **invoke** from .NET, but it's also obvious that JS -> .NET interop is part of the implementation. Maybe we should lean into this more, and there are some APIs that should be defined.

javiercn:
> It's obvious why the JS -> .NET infrastructure isn't there for you to **invoke** from .NET, but it's also obvious that JS -> .NET interop is part of the implementation. Maybe we should lean into this more, and there are some APIs that should be defined.

Maybe, although I'm not sure what you are suggesting. That said, adding new APIs is something we can do later. For now, we'll just stick with adding a method to JSRuntimeBase and we can add more things when we have scenarios for it.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -62,18 +62,31 @@ public Task<T> InvokeAsync<T>(string identifier, params object[] args)
...
+        /// </summary>
+        /// <remarks>
+        /// This callback can be used in remote JS interop scenarios to sanitize exceptions that happen on the server to avoid disclosing
+        /// sensitive information to remote browser clients.
+        /// </remarks>
+        /// <param name="exception">The exception that occurred.</param>
+        /// <param name="assemblyName">The assembly for the invoked .NET method.</param>
+        /// <param name="methodIdentifier">The identifier for the invoked .NET method.</param>
+        /// <returns>An object containing information about the exception.</returns>
+        protected virtual object OnDotNetInvocationException(Exception exception, string assemblyName, string methodIdentifier) => exception.ToString();
```
rynowak: I kinda wonder about this signature being fragile, but I don't have any better ideas. I think this is good for right now and we can look at it again as part of the API review of this area.

PR #1880, opened 2019-06-25
Adds an opt-in default timeout for async JS calls

javiercn:


src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -26,6 +25,11 @@ public abstract class JSRuntimeBase : IJSRuntime
 
         internal DotNetObjectRefManager ObjectRefManager { get; } = new DotNetObjectRefManager();
 
+        /// <summary>
+        /// Gets or sets the default timeout for asynchronous JavaScript calls.
+        /// </summary>
+        protected TimeSpan? DefaultAsyncCallTimeout { get; set; }
```
javiercn: RemoteJSRuntime will set this value based on circuit options.

javiercn: Its opt-in so that there is no perf impact in blazor client-side.

rynowak: I think it's smart to have the timeout be optional, I also think we'll want the same behavior in both places for consistency (unless people are actually hitting the timeout in real scenarios).

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -58,6 +65,22 @@ public Task<T> InvokeAsync<T>(string identifier, params object[] args)
...
         }
 
+        private async Task<T> GetResultTask<T>(long taskId, string methodIdentifier, Task<T> task)
+        {
+            await Task.WhenAny(task, Task.Delay(DefaultAsyncCallTimeout.Value));
+
+            if (!task.IsCompleted)
+            {
+                _pendingTasks.TryRemove(taskId, out _);
+                throw new JSException($"The call to {methodIdentifier} exceded the maximum allowed wait time.");
```
rynowak: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#using-a-timeout

This should also throw `OperationCanceledException`

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
```diff
@@ -59,8 +59,10 @@ public static partial class JSRuntime
     public abstract partial class JSRuntimeBase : Microsoft.JSInterop.IJSRuntime
     {
         protected JSRuntimeBase() { }
+        protected System.TimeSpan? DefaultAsyncCallTimeout { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         protected abstract void BeginInvokeJS(long asyncHandle, string identifier, string argsJson);
         public System.Threading.Tasks.Task<T> InvokeAsync<T>(string identifier, params object[] args) { throw null; }
+        protected virtual object OnDotNetInvocationException(System.Exception exception, string assemblyName, string methodIdentifier) { throw null; }
```
rynowak: Where did this come from?

javiercn: From another PR, this is based on, which adds the ability to sanitize exceptions. I just forgot to update ref assemblies on it. This one builds on top of that one

javiercn:
I'll fix the cancellation token stuff and add tests, I wanted to make sure you were ok with the approach before going further

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: Don't you think a more specific exception might make troubleshooting easier (specially when seeing exceptions on deployed applications)

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: Why do you think its useful to have a timeout on the client?

rynowak: When one environment is more restrictive than the other, people will write components that only function in the less restrictive environment. We've seen this already with the sync-context + dispatching.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
rynowak: https://docs.microsoft.com/en-us/dotnet/api/system.timeoutexception?view=netframework-4.8

rynowak: Either way it shouldn't be `JSException` - that's the one we use to represent an exception thrown by js code.

rynowak:
>  I wanted to make sure you were ok with the approach before going further

If you wanted to, you could eliminate some overhead by using a single TCS - but in general the approach is going to be different code with the same result. @davidfowl already wrote a novel about this.

The other thing to investigate is what does the JS side see when something is cancelled due to the timeout. So if you timeout after 30 sec, but the JS side completes at 35 sec, what does it see? Exception?

We should probably also be using https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

javiercn:
> We should probably also be using https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

Yeah, I noticed, I just didn't want to distract the focus of this PR with the change, but I'll add it.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
SteveSandersonMS: A default 60-second timeout, even in client-side Blazor, sounds like the right thing to do to me too.

There is some risk that people will have a bad time with it, e.g., they are waiting for user input in some 3rd-party JS UI such as a dialog, and during development and testing they always click the button in time. But then in production they find real users are getting errors because they are responding to the prompt too slowly. The developer would need to know they have to disable the timeout or pick a longer value in this case.

However I'm a little more comfortable with that risk, than with the other risk that they assume it's no timeout by default and are unable to port their components between the two environments.

SteveSandersonMS:
I know this is really annoying from an API cleanliness point of view, but doesn't there also have to be a way of disabling or changing the timeout on a per-call basis?

Consider the [Payment Request "show" API](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/show). It returns a `Promise` that completes when the user has finished interacting with the native "enter card details" dialog. We still want this to be callable through JS interop APIs, don't we?

Now also consider if we have a 60-second timeout by default (maybe that's even our built-in default). The developer absolutely does *not* want the payment request call to fail if a user takes more than 60 seconds to type in their card details. This user was just about to give them money! You'd probably want to allow 15 minutes, if not unlimited time until the end of the circuit, to complete this.

You may argue that remote JS interop isn't a good fit for long-running promises like this. And that's kind of true, but I don't think it's viable for us to tell people they either can't use those APIs or they have to implement their own custom async model.

So I'm not thrilled about complicating the API surface, but it seems there's no getting away from adding an extra optional `timeout` param to `InvokeAsync`, e.g.:

```cs
public Task<T> InvokeAsync<T>(TimeSpan timeout, string identifier, params object[] args) { ... }
```

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: Hmm, should we add a cancellation token parameter and apply the timeout only when a cancellation token is not explicitly provided?

I would be more comfortable with that solution as that would give users the ability to configure the timeout on a per-call to match their expectations. For example, in the case you bring up, a user might choose to pass a custom cancellation token to wait indefinitely or until something else happens.

We will need to make the timeout property public I believe if we want it to be configurable in the WebAssembly case. And I would still allow users to set it to null if they want to disable it alltogether.

**So to summarize my proposal:**
* Add cancellation token parameter users can pass to explicitly control when an operation gets cancelled.
* Make the time-out property public so that it can be configured in WebAssembly and have it default to 60 seconds.
  * Alternatively we can keep the property protected and have it be configured indirectly through other mechanisms.
    * On SSB through circuit options.
    * On CSB through a static property.
  * @SteveSandersonMS What pattern do you suggest to configure this in WebAssembly. Startup code?
* Allow for the property to be set to null to disable time-outs.

SteveSandersonMS: I agree, cancellation token does sound better!

> What pattern do you suggest to configure this in WebAssembly. Startup code?

I'd have a look at whether it's easy enough to use `services.Configure` and `Options<T>`, where `T` is some new type called something like `WebAssemblyJSRuntimeOptions`.

If that turns out to be inconvenient or bloats the app size by pulling in loads of options-related functionality, a rough but acceptable-in-the-short-term option would be having `Microsoft.AspNetCore.Blazor` contain an extension method on `IJSRuntime` called `SetDefaultTimeout` that writes to a `static` property on the internal `WebAssemblyJSRuntime` class. If we take this approach, it's something we'd want to improve in the future when formalizing the approach to options/config in client-side Blazor.

javiercn:
A few updates:
* Add an overload that takes in a cancellation token. Sadly no params for that overload; An array needs to be built explicitly.
* You can pass in a cancellation token explicitly or use the default cancellation set up by the platform. We will configure both .NET and WASM to wait for 1 minute before cancelling the task.
* Added tests to confirm that works as desired.
* Removed the exception that the platform would throw when the task has already been cancelled and simply ignore the results.

dougbu:
> @javiercn javiercn requested a review from dougbu as a code owner

I don't see any changes in the build infrastructure. Is this review request out of date?

javiercn:
@dougbu I think so, I changed the PR base and you have been added by mistake

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -20,40 +22,99 @@ public abstract class JSRuntimeBase : IJSRuntime
...
-            // the operation.
-
             var taskId = Interlocked.Increment(ref _nextPendingTaskId);
-            var tcs = new TaskCompletionSource<T>();
+            var tcs = new TaskCompletionSource<T>(TaskContinuationOptions.RunContinuationsAsynchronously);
+            if (cancellationToken != default)
+            {
+                _cancellationRegistrations[taskId] = cancellationToken.Register(() =>
+                {
+                    tcs.TrySetCanceled();
```
pranavkm: Should you pass in the `CancellationToken` here?

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -20,40 +22,99 @@ public abstract class JSRuntimeBase : IJSRuntime
...
-                _pendingTasks.TryRemove(taskId, out _);
+                CleanupRegistrations(taskId);
                 throw;
             }
         }
 
+        private void CleanupRegistrations(long taskId)
+        {
+            _pendingTasks.TryRemove(taskId, out _);
+            _cancellationRegistrations.TryRemove(taskId, out var registration);
```
pranavkm: This should  guard against this returning `false`

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: That's the point of that? Just curious cause I can't figure out from the documentation.

SteveSandersonMS: I think it's just to avoid having to capture `tcs` in the closure, which costs an allocation.

In this case it doesn't help us avoid an allocation, since we also have to capture `taskId` anyway.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
SteveSandersonMS: I agree that would be a bit better than having the comment on L86

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -88,9 +149,13 @@ internal void EndInvokeJS(long asyncHandle, bool succeeded, JSAsyncCallResult as
             {
                 if (!_pendingTasks.TryRemove(asyncHandle, out var tcs))
                 {
-                    throw new ArgumentException($"There is no pending task with handle '{asyncHandle}'.");
+                    // We should simply return if we can't find an id for the invocation.
+                    // This likely means that the method that initiated the call defined a timeout and stopped waiting.
+                    return;
                 }
 
+                CleanupRegistrations(asyncHandle);
```
SteveSandersonMS: This is a bit confusing. It's referred to as `taskId` (not `asyncHandle`) everywhere else. Could this be updated to match?

SteveSandersonMS: BTW I know that inconsistency was my fault.

src/JSInterop/Microsoft.JSInterop/test/JSRuntimeBaseTest.cs
```diff
@@ -38,6 +39,70 @@ public void DispatchesAsyncCallsWithDistinctAsyncHandles()
...
+        [Fact]
+        public async Task InvokeAsync_CancelsAsyncTask_AfterDefaultTimeout()
+        {
+            // Arrange
+            var runtime = new TestJSRuntime();
+            runtime.DefaultTimeout = TimeSpan.FromSeconds(1);
+
+            // Act
+            var task = runtime.InvokeAsync<object>("test identifier 1", "arg1", 123, true);
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
SteveSandersonMS: Is there a risk of this being flaky? Are the continuations guaranteed to execute on the same thread and in the same order they were queued? I wouldn't have guessed that `cancellationToken.Register`'s callback necessarily ran on the same thread the call was made from.

Maybe just make the delay on L51 a bit longer than the one on L47.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -20,40 +22,99 @@ public abstract class JSRuntimeBase : IJSRuntime
...
             }
         }
 
+        private void CleanupRegistrations(long taskId)
+        {
+            _pendingTasks.TryRemove(taskId, out _);
+            _cancellationRegistrations.TryRemove(taskId, out var registration);
+
+            // dispose will noop when registration == default
+            registration.Dispose();
```
SteveSandersonMS: AFAICT this line isn't covered by any unit test. Is there some way we can prove in tests that we're correctly unregistering from the CT once the JS call itself completes (successfully or unsuccessfully)?

SteveSandersonMS:
At some point in the future, we might consider also sending a notification to the JS side if the .NET CT fires, to tell it that it can give up on whatever it was doing.

However, there isn't a single standard way to represent a cancellation request for a JS `Promise`, so it's unclear how developers would make any practical use of such a feature.

So I think this is something we can at least defer until the future and see what kinds of customer requests we get.

javiercn:
> At some point in the future, we might consider also sending a notification to the JS side if the .NET CT fires, to tell it that it can give up on whatever it was doing.
> 
> However, there isn't a single standard way to represent a cancellation request for a JS `Promise`, so it's unclear how developers would make any practical use of such a feature.
> 
> So I think this is something we can at least defer until the future and see what kinds of customer requests we get.

We would simply reject the promise with a message saying it was cancelled. That's how I've seen it done in the past.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: I'm not sure if we can inspect the state of cancellation tokens. I can try and inherit from them, but I've seen it fired and disposing means unregister AFAIK.

SteveSandersonMS:
> We would simply reject the promise with a message saying it was cancelled. That's how I've seen it done in the past.

In the .NET-to-JS case, we're the *consumer* of the promise (or more precisely, the "thenable"), not the producer. We can't reject it. Even if we could, it wouldn't have a useful effect, which would be cancelling whatever operation it represents (e.g., a `fetch`).

src/JSInterop/Microsoft.JSInterop/test/JSRuntimeBaseTest.cs
javiercn: What I think you think it could happen is that task.Delay gets awaited for 1 second and between that an the call to await task the cancellation token doesn't get a chance to execute, is that assessment correct? 
It all depends on how the thread scheduler decides to execute the next task. If it decides to continue executing something else and then go and execute the continuation for task before the register callback then there's nothing we can do.

Adding more time would help reduce that chance in that place, but I'm wondering if there's something we can do to the task scheduler to force this.

In any case, I'll add more time here to mitigate that

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: Sounds good

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: > I think it's just to avoid having to capture `tcs` in the closure, which costs an allocation.
> 
> In this case it doesn't help us avoid an allocation, since we also have to capture `taskId` anyway.

I think pranav is referring to passing the cancellation token to trySetCancelled, which I'm not sure how would avoid the allocation as that would require receiving it as a parameter to Register

src/JSInterop/Microsoft.JSInterop/test/JSRuntimeBaseTest.cs
SteveSandersonMS: > What I think you think it could happen is that task.Delay gets awaited for 1 second and between that an the call to await task the cancellation token doesn't get a chance to execute, is that assessment correct?

Yes - both calls are scheduled to occur after 1 second, and may happen on different threads, so which one affects/accesses the shared state first is a bit of a gamble.

> Adding more time would help reduce that chance in that place, but I'm wondering if there's something we can do to the task scheduler to force this.

Instead of waiting for 1 second, we could wait for `task` itself to complete, then assert that it failed with a `TaskCancellationException`.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: I haven't been able (looking at the corefx source code) to determine why passing the cancellation token is important, as AFAIK it doesn't get reflected publicly in any place. (It ends up on an internal Task.CancellationToken property)

javiercn:
> In the .NET-to-JS case, we're the _consumer_ of the promise (or more precisely, the "thenable"), not the producer. We can't reject it. Even if we could, it wouldn't have a useful effect, which would be cancelling whatever operation it represents (e.g., a `fetch`).

You are right.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: Will do

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: done

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: That said, I added it because its uber cheap

src/JSInterop/Microsoft.JSInterop/test/JSRuntimeBaseTest.cs
javiercn: Yeah, you are right, the task.delay there does nothing. This is why I shouldn't be writing tests at night 😄

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: There's no way AFAIK we can know for sure we are correctly disposing registrations.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
javiercn: I don't like this name as I would prefer something more explicit or less, like DefaultTimeout, I don't feel it conveys enough information to know what you are setting a timeout for, That said, I'm not a native english speaker so I've switched it to your suggestion.

Issue #1808, opened 2019-06-07
Console logs timestamps in local timezone instead of UTC


danmosemsft:
Moved from Corefx. Issue by @tomkerkhove

When configuring the Console logger to output timestamps as following:
```csharp
Host.CreateDefaultBuilder(args)
    .ConfigureAppConfiguration((hostingContext, config) =>
    {
        config.AddEnvironmentVariables();
    })
    .ConfigureLogging((hostBuilderContext, loggingBuilder) =>
    {
        loggingBuilder.AddConsole(consoleLoggerOptions => consoleLoggerOptions.TimestampFormat = "[HH:mm:ss]");
    })
    .ConfigureServices(services =>
    {
        services.AddHostedService<OrdersQueueProcessor>();
    });
```

The generated logged message uses my local timezone rather than UTC without an option to change this.

Would be good to be able to influence this and use UTC by default.

anurse:
It would definitely have to be an option, but I don't see this as a problem.

The simplest thing would certainly be to just add a `WriteTimestampInUtc` option that can be set to `true` to opt-in to this behavior. The main concern I have is just a possible explosion of options. I don't think this would be unreasonable though. There are more powerful logging systems out there (such as Serilog) that give you a LOT of flexibility about the log messages and we're not out to replicate their functionality.

tomkerkhove:
I'm not saying it's a problem but indeed suggesting that we should be able to use UTC.
Having this as a property is fine for me!

Other logging systems provide more robust features indeed but those do not require to be in here, however, I think timestamps are essential.

tomkerkhove:
This one should be done

anurse:
Yep, thanks for the reminder!

PR #1896, opened 2019-06-27
 Provide capability to use UTC time notation for console logs

tomkerkhove:
Provide capability to use UTC time notation for console logs via `ConsoleLoggerOptions.UseUtcTimezone`.

This is turned off by default to be backwards compatible.

Addresses #1808

In terms of testing I would suggest to provide a similar test to `WriteCore_LogsCorrectTimestamp` which uses UTC option and verifies stamp or a similar one - Thoughts?


src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
```diff
@@ -37,5 +37,10 @@ public ConsoleLoggerFormat Format
         /// Gets or sets format string used to format timestamp in logging messages. Defaults to <c>null</c>.
         /// </summary>
         public string TimestampFormat { get; set; }
+
+        /// <summary>
+        /// Gets or sets indication whether or not UTC timezone should be used to for timestamps in logging messages. Defaults to <c>false</c>.
+        /// </summary>
+        public bool UseUtcTimezone { get; set; } = false;
```
gfoidl:
```suggestion
        public bool UseUtcTimezone { get; set; }
```
`false` is already the default.

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -137,11 +137,16 @@ private LogMessageEntry CreateDefaultLogMessage(StringBuilder logBuilder, LogLev
                 logBuilder.AppendLine(exception.ToString());
             }
 
+            string timestamp = null;
             var timestampFormat = Options.TimestampFormat;
+            if (timestampFormat != null)
+            {
+                timestamp = Options.UseUtcTimezone ? DateTime.UtcNow.ToString(timestampFormat) : DateTime.Now.ToString(timestampFormat);
```
```suggestion
                var dateTime = Options.UseUtcTimezone ? DateTime.UtcNow : DateTime.Now;
                timestamp = dateTime.ToString(timestampFormat);
```
But I think this is more a matter of taste.

src/Logging/Logging.Console/src/ConsoleLoggerOptions.cs
tomkerkhove: True but I wanted to be more explicit - However it doesn't really make sense so I'll remove it!

src/Logging/Logging.Console/src/ConsoleLogger.cs
tomkerkhove: Fine for me, will send update asap

tomkerkhove:
@gfoidl I've incorporated your suggested and added a test similar to the existing one.

However, they do not verify the output so I'm curious how you would approach it?
Should I just deserialize them and verify the timezone? Open to suggestions.

src/Logging/test/ConsoleLoggerTest.cs
```diff
@@ -382,14 +382,50 @@ public void WriteCore_LogsCorrectTimestamp(ConsoleLoggerFormat format, LogLevel
...
+            var t = SetUp(new ConsoleLoggerOptions { TimestampFormat = "yyyyMMddHHmmss ", Format = format, UseUtcTimezone = true});
+            var levelPrefix = t.GetLevelPrefix(level);
+            var logger = t.Logger;
+            var sink = t.Sink;
+            var ex = new Exception("Exception message" + Environment.NewLine + "with a second line");
+
+            // Act
+            logger.Log(level, 0, _state, ex, _defaultFormatter);
+
+            // Assert
```
gfoidl: Hm, there is no assert that UTC is written, so this test is more or less a duplication of the already existing test for normal datetime. (just see you commented on that too :wink:).

> Should I just deserialize them and verify the timezone?

That's possible (but then you have to assert / take into account that current timezone != UTC).

Maybe it's better to (at least in my projects I would do) extract the creation of the `timestamp` in `ConsoleLogger` to a separate method that is testable?

Something like 
```c#
private static string GetTimestamp(string timestampFormat, bool useUtcTimezone)
    => GetTimestamp(timestampFormat, useUtcTimezone, DateTime.Now);

internal static string GetTimestamp(string timestampFormat, bool useUtcTimezone, DateTime now)
{
    if (timestampFormat == null)
        return null;

    if (useUtcTimezone)
        now = now.ToUniversalTime();

    return now.ToString(timestampFormat);
}
```
Then the second method is testable.

tomkerkhove: That's one approach but who says that method is being called? We have no guarantee about that actually so not sure if that's going to cover everything.

gfoidl: You're right. So your approach is a better end-to-end test.

tomkerkhove: So what I'd propose then would be:
1. Change format to `yyyyMMddHHmmsszzz`
2. Get tracked timestamp via regex similar to https://github.com/aspnet/Extensions/pull/1896/files#diff-a27bf784d2879ac5f61a0d1a8573fc80R418
3. Parse to DateTime and check timezone

Thoughts?

I'm not a regex expert so help would be ideal on that 😅

tomkerkhove: This should be there, except for explicitly setting the culture to a non-UTC compatible one for the non-UTC test.

I tried to use the folling but it didn't work:
```csharp
var cultureInfo = new CultureInfo("en-US");
CultureInfo.DefaultThreadCurrentCulture = cultureInfo;
CultureInfo.DefaultThreadCurrentUICulture = cultureInfo;
```

tomkerkhove:
Looks like my PR is messed up due to the rebase, going to start with a clean branch.

tomkerkhove:
Closing in favor of #2420 with a clean branch.

PR #1898, opened 2019-06-27
Update corefx dependencies

BrennanConroy:
Had to manually change the versions because of certain people being people...

These are the "officially decided on" versions we're using from the release channel.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -10,69 +10,69 @@
 -->
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Win32.Registry" Version="4.6.0-preview7.19326.9" CoherentParentDependency="Microsoft.NETCore.App.Runtime.win-x64">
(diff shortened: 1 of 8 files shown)


anurse:
Let's take this ASAP so we can let it settle.

anurse:
Can we manually take this to aspnet/AspNetCore ASAP? I'd like the breaking changes resolved as quickly as possible for obvious reasons 😛.

BrennanConroy:
😢  I guess I can

anurse:
I mean, if the automation is working and `EveryBuild` this may just be baby-sitting :).

dougbu:
What exactly led to the need for this manual update? I understand the reaction changes but not why the Maestro bot didn't use these versions automatically.

If this is about pulling from '.NET Core 3 Release' instead of '.NET Core 3 Dev', later CoreFx and core-setup packages will undo the changes here. We should only be using '.NET Core 3 Dev' on the master branch of any of our repos.

PR #1900, opened 2019-06-27
Add missing doc comments

ryanbrandenburg:
Filling in some missing doc comments.

Partially addresses: https://github.com/aspnet/AspNetCore/issues/11506


src/Caching/Abstractions/src/MemoryCacheEntryExtensions.cs
```diff
@@ -24,8 +25,9 @@ public static class MemoryCacheEntryExtensions
         /// <summary>
         /// Sets the size of the cache entry value.
         /// </summary>
-        /// <param name="options"></param>
-        /// <param name="size"></param>
+        /// <param name="options">The option to set the size of.</param>
```
Eilon: Is this ambiguous or misleading? It's not the size of the option. It's the size of the cache entry.

src/Hosting/Abstractions/src/IHostBuilder.cs
```diff
@@ -48,31 +48,31 @@ public interface IHostBuilder
         /// <summary>
         /// Overrides the factory used to create the service provider.
         /// </summary>
-        /// <typeparam name="TContainerBuilder"></typeparam>
-        /// <param name="factory"></param>
+        /// <typeparam name="TContainerBuilder">The type of builder.</typeparam>
+        /// <param name="factory">The factory to register/</param>
```
Eilon: Typo: slash instead of dot.

src/Hosting/Hosting/src/HostingHostBuilderExtensions.cs
```diff
@@ -152,7 +152,7 @@ public static IHostBuilder UseConsoleLifetime(this IHostBuilder hostBuilder, Act
         /// Enables console support, builds and starts the host, and waits for Ctrl+C or SIGTERM to shut down.
         /// </summary>
         /// <param name="hostBuilder">The <see cref="IHostBuilder" /> to configure.</param>
-        /// <param name="cancellationToken"></param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the health check.</param>
```
Eilon: Health check? I think this comment was erroneously copied in many many places. Please fix in all places.

src/Logging/Logging.Abstractions/src/EventId.cs
```diff
@@ -3,42 +3,80 @@
 
 namespace Microsoft.Extensions.Logging
 {
+    /// <summary>
+    /// Identifies a logging event.
```
Eilon: I think it would be very useful to add some explanatory content here because I can never remember what matters here: does the int matter, does the string matter, do both matter?

By reading the code, it looks like the int is required and is what should be unique, and that the string is optional and doesn't need to be unique? Is that even true? Does anyone know? 😄 We should figure it out and document it here.

src/Logging/Logging.Abstractions/src/EventId.cs
```diff
@@ -3,42 +3,80 @@
...
+        /// Implicitly creates an EventId from the given <see cref="int"/>.
+        /// </summary>
+        /// <param name="i">The <see cref="int"/> to convert to an EventId.</param>
         public static implicit operator EventId(int i)
         {
             return new EventId(i);
         }
 
+        /// <summary>
+        /// Checks if two specified <see cref="EventId"/>'s have the same value.
```
Eilon: See above comment about confusion on what "matters" here. From the code, it seems two EventIds are equal if their ids are equal. The names are ignored?

src/Logging/Logging.Abstractions/src/EventId.cs
```diff
@@ -3,42 +3,80 @@
...
         public string Name { get; }
 
+        /// <inheritdoc />
         public override string ToString()
         {
             return Name ?? Id.ToString();
         }
 
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
```
Eilon: Per earlier comments, what does "equal" truly mean here? Definitely not obvious.

src/Logging/Logging.Abstractions/src/NullLogger.cs
```diff
@@ -10,6 +10,9 @@ namespace Microsoft.Extensions.Logging.Abstractions
     /// </summary>
     public class NullLogger : ILogger
     {
+        /// <summary>
+        /// Returns an instance of <see cref="NullLogger"/>.
```
Eilon: Not just `an` instance. It's `the` instance. Maybe, `Returns the shared instance of BLAH`?

src/Logging/Logging.Abstractions/src/NullLoggerFactory.cs
```diff
@@ -9,6 +9,14 @@ namespace Microsoft.Extensions.Logging.Abstractions
     /// </summary>
     public class NullLoggerFactory : ILoggerFactory
     {
+        /// <summary>
+        /// Creates a new <see cref="NullLoggerFactory"/> instance.
+        /// </summary>
+        public NullLoggerFactory() { }
```
Eilon: Weird. I feel like this was meant to be private, but wasn't. It'd be a breaking change to make it private. Should it be private? And if so, worth the break?

src/Logging/Logging/src/FilterLoggingBuilderExtensions.cs
```diff
@@ -11,58 +11,181 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     public static class FilterLoggingBuilderExtensions
     {
+        /// <summary>
+        /// Adds a log filter to the factory.
+        /// </summary>
+        /// <param name="builder">The <see cref="ILoggingBuilder"/> to add the filter to.</param>
+        /// <param name="filter">The filter to be added.</param>
```
Eilon: These freakin' callbacks of type `Func<string, string, LogLevel, bool>` are not clear at all. What are the 3 params there? Category, *something*, and LogLevel?

src/Logging/Logging/src/LoggerFactory.cs
```diff
@@ -69,6 +91,11 @@ private void RefreshFilters(LoggerFilterOptions filterOptions)
             }
         }
 
+        /// <summary>
+        /// Creates an <see cref="ILogger"/> with the given <paramref name="categoryName"/>.
+        /// </summary>
+        /// <param name="categoryName">The categoryName for messages produced by the logger.</param>
```
Eilon: In the prose, use `category name` (it's just English here, not a code identifier).

src/Logging/Logging/src/LoggingBuilderExtensions.cs
```diff
@@ -12,23 +12,40 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     public static class LoggingBuilderExtensions
     {
+        /// <summary>
+        /// Sets a minimum <see cref="LogLevel"/> requirement for log messages.
```
Eilon: `... for log messages to be logged`

src/ObjectPool/src/IPooledObjectPolicy.cs
```diff
@@ -3,10 +3,23 @@
...
     public interface IPooledObjectPolicy<T>
     {
+        /// <summary>
+        /// Create a <typeparamref name="T"/>.
+        /// </summary>
+        /// <returns>The <typeparamref name="T"/> which was created.</returns>
         T Create();
 
+        /// <summary>
+        /// Put an object in the pool.
```
Eilon: `Return an object to the pool.`

src/ObjectPool/src/IPooledObjectPolicy.cs
```diff
@@ -3,10 +3,23 @@
...
+        /// </summary>
+        /// <param name="obj">The object to put in the pool.</param>
```
Eilon: `... to return to the pool`

src/ObjectPool/src/ObjectPool.cs
```diff
@@ -3,15 +3,31 @@
...
 namespace Microsoft.Extensions.ObjectPool
 {
+    /// <summary>
+    /// A pool of objects.
+    /// </summary>
+    /// <typeparam name="T">The type of objects to pool.</typeparam>
     public abstract class ObjectPool<T> where T : class
     {
+        /// <summary>
+        /// Returns an object from the pool.
```
Eilon: `Gets an object from the pool.` It's confusing to use the word `Return` in the `Get` method when there's also a `Return` method!

src/ObjectPool/src/ObjectPool.cs
```diff
@@ -3,15 +3,31 @@
...
+        /// </summary>
+        /// <returns>An object from the pool.</returns>
         public abstract T Get();
 
+        /// <summary>
+        /// Put an object in the pool.
```
Eilon: See earlier comment.

src/ObjectPool/src/ObjectPool.cs
Eilon: Also, I think this isn't even quite true. All this does is guarantee you'll get an object of the desired type. If there are any available in the pool, you get one from the pool. If the pool is empty, you get a new one. I think it's crucial to describe this.

src/Logging/Logging/src/FilterLoggingBuilderExtensions.cs
anurse: Category and Provider (Console/Debug/AzureAppService/EventSource/Etc.) are the two, I don't know which order though. At some point with these, `Func<,,,,>` loses it's charm...

src/Logging/Logging.Abstractions/src/NullLoggerFactory.cs
ryanbrandenburg: If you want to file an issue I think that's a reasonable change, but since this is a docs PR I'm trying to keep functional changes out of it.

src/ObjectPool/src/IPooledObjectPolicy.cs
ryanbrandenburg: I phrased it this way because it's not clear to me that we have any guarantee that the object given to this method was in the pool in the first place (in which case the method is misnamed too, but that's a different story). Is there such a guarantee?

Eilon: No such guarantee as far as I know. But the method name is `Return`, so I think you want to use that term and avoid introducing a new one. It's fine to add clarifying remarks if needed.

src/Logging/Logging.Abstractions/src/NullLoggerFactory.cs
Eilon: Ah, looks like there's already a bug on that: https://github.com/aspnet/Extensions/issues/662. I agree it shouldn't be done as part of this PR; it was merely an observation.

src/Hosting/Hosting/src/HostingHostBuilderExtensions.cs
```diff
@@ -50,9 +50,21 @@ public static IHostBuilder UseContentRoot(this IHostBuilder hostBuilder, string
             });
         }
 
+        /// <summary>
+        /// Specify the <see cref="IServiceProvider"/> to be the default one.
```
Eilon: @Tratcher - can you review these docs? I'm not familiar with their intentions.

src/Logging/Logging.Abstractions/src/EventId.cs
```diff
@@ -67,7 +67,7 @@ public override string ToString()
         }
 
         /// <summary>
-        /// Indicates whether the current object is equal to another object of the same type.
+        /// Indicates whether the current object is equal to another object of the same type. Two events are equal if they the same Id.
```
Eilon: `.. if they **have** the same id.` (Note that `id` is lowercase. Or change it to a `<ref>` thingy if you want to specify the `Id` property.)

src/ObjectPool/src/ObjectPool.cs
```diff
@@ -10,9 +10,9 @@ namespace Microsoft.Extensions.ObjectPool
     public abstract class ObjectPool<T> where T : class
     {
         /// <summary>
-        /// Returns an object from the pool.
+        /// Gets an object from the pool is one is available, otherwise creates one.
```
Eilon: `... **if** one is available...`

ryanbrandenburg:
:up::date:

src/ObjectPool/src/DefaultObjectPoolProvider.cs
```diff
@@ -5,10 +5,17 @@
 
 namespace Microsoft.Extensions.ObjectPool
 {
+    /// <summary>
+    /// A provider of <see cref="DefaultObjectPool{T}"/>'s.
```
rynowak: Not super correct, this default implementation of `ObjectPoolProvider`. We don't guarantee that it returns `DefaultObjectPool`

src/ObjectPool/src/IPooledObjectPolicy.cs
```diff
@@ -3,10 +3,23 @@
...
+        /// Create a <typeparamref name="T"/>.
+        /// </summary>
+        /// <returns>The <typeparamref name="T"/> which was created.</returns>
         T Create();
 
+        /// <summary>
+        /// Return an object to the pool.
+        /// </summary>
+        /// <param name="obj">The object to return to the pool.</param>
+        /// <returns><code>true</code> if returned to the pool.</returns>
```
rynowak: This needs some more info.

You implement this method to run some processing when an object was returned to the pool. You can *reset* the state of an object. You also can return false from here if it's not possible/desirable for the pool to keep the object.

src/ObjectPool/src/DefaultObjectPool.cs
```diff
@@ -8,6 +8,10 @@
 
 namespace Microsoft.Extensions.ObjectPool
 {
+    /// <summary>
+    /// Default implementation of <see cref="ObjectPool{T}"/>.
+    /// </summary>
+    /// <typeparam name="T">The type to pool objects for.</typeparam>
```
rynowak: We should capture the detail that this implementation keeps a cache of *retained objects*. This means that if you check out twenty objects, and return twenty objects then `maximumRetained` will be kept by the pool, and `20 - maximumRetained` will be available to be GCed.

src/ObjectPool/src/ObjectPoolProvider.cs
```diff
@@ -3,13 +3,24 @@
 
 namespace Microsoft.Extensions.ObjectPool
 {
+    /// <summary>
+    /// A provider of <see cref="ObjectPool{T}"/>'s.
```
rynowak: Apostraphe is wrongzo, this is a pluralisation.

Eilon: In these instance's I tend to phrase it as `A provider of <Foo> instances.` thus avoiding the trap of incorrect apostrophization's.

ryanbrandenburg:
@rynowak I'd really like your blessing to my ObjectPool changes since it seems like that's the area where there's the most substantive stuff to get wrong.

ryanbrandenburg:
👀 @rynowak

PR #1907, opened 2019-07-01
Capture stack trace in FileConfigurationProvider

HaoK:
Fixes https://github.com/aspnet/Extensions/issues/1892

### DIFF
--- a/src/Configuration/Config.FileExtensions/src/FileConfigurationProvider.cs
+++ b/src/Configuration/Config.FileExtensions/src/FileConfigurationProvider.cs
@@ -69,7 +69,8 @@ namespace Microsoft.Extensions.Configuration
                     {
                         error.Append($" The physical path is '{file.PhysicalPath}'.");
                     }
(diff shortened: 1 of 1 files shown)


src/Configuration/Config.FileExtensions/src/FileConfigurationProvider.cs
```diff
@@ -125,7 +126,7 @@ private void HandleException(Exception e)
             }
             if (!ignoreException)
             {
-                throw e;
+                ExceptionDispatchInfo.Capture(e).Throw();
```
Tratcher: Shouldn't the capture happen on line 90?
https://github.com/aspnet/Extensions/pull/1907/files#diff-61c1423e04275a1e7c00297557873270R90

HaoK: Yeah, updated to pass in the dispatch info instead and updated the other place this gets called

PR #1915, opened 2019-07-02
Add InvokeAsync with cancellation token overload to IJSRuntime

javiercn:


src/JSInterop/Microsoft.JSInterop/src/IJSRuntime.cs
```diff
@@ -19,5 +20,15 @@ public interface IJSRuntime
...
+
+        /// <summary>
+        /// Invokes the specified JavaScript function asynchronously.
+        /// </summary>
+        /// <typeparam name="TValue">The JSON-serializable return type.</typeparam>
+        /// <param name="identifier">An identifier for the function to invoke. For example, the value <code>"someScope.someFunction"</code> will invoke the function <code>window.someScope.someFunction</code>.</param>
+        /// <param name="args">JSON-serializable arguments.</param>
+        /// <param name="cancellationToken">A cancellation token to signal the cancellation of the operation.</param>
+        /// <returns>An instance of <typeparamref name="TValue"/> obtained by JSON-deserializing the return value.</returns>
+        Task<TValue> InvokeAsync<TValue>(string identifier, IEnumerable<object> args, CancellationToken cancellationToken = default);
```
rynowak: What's the contract of this? Does providing a cancellation token override the default timeout? What if they pass `default`? Does that still override the timeout?

javiercn: * No cancellation token (params overload) uses the default timeout.
* Passing a cancellation token overrides the default timeout.
* Passing the default means no timeout

rynowak: Then this needs to be non-default parameter since the decision to pass a CT is really significant. There's a really important difference between these overloads.

javiercn: Fair enough, I'll send another PR

rynowak: I think we should get an ack from @GrabYourPitchforks as well (don't have to block on that).

javiercn: We can review that on the meeting next week.

PR #1916, opened 2019-07-02
Explicitly reference Roslyn compiler toolset version that will ship in preview 7

dougbu:
- see also aspnet/EntityFrameworkCore#16370 and aspnet/EntityFrameworkCore#16385 discussions
- grab latest from the '.NET Core 3 Release' channel


dougbu:
@wtgodbe new subscription for 'release/3.0-preview7' is '51725046-aeea-47d4-e680-08d6fb36d5a8'

dougbu:
Thanks @wtgodbe❕ (I'll fix up the property name in 'master')

PR #1945, opened 2019-07-05
Fix memory leak

OpenSourceAries:
 - Move event registration out of event callback to avoid memory leak(recursive event registration)

### DIFF
--- a/src/Logging/Logging.Console/src/ConsoleLoggerProvider.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLoggerProvider.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.Logging.Console
             _options = options;
             _loggers = new ConcurrentDictionary<string, ConsoleLogger>();
 
+            _optionsReloadToken = _options.OnChange(ReloadLoggerOptions);
             ReloadLoggerOptions(options.CurrentValue);
 
             _messageQueue = new ConsoleLoggerProcessor();
@@ -44,8 +45,6 @@ namespace Microsoft.Extensions.Logging.Console
             {
                 logger.Value.Options = options;
             }
-
-            _optionsReloadToken = _options.OnChange(ReloadLoggerOptions);
         }
 
         /// <inheritdoc />
--- a/src/Logging/Logging.Console/src/ConsoleLoggerProvider.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLoggerProvider.cs
@@ -23,8 +23,8 @@ namespace Microsoft.Extensions.Logging.Console
             _options = options;
             _loggers = new ConcurrentDictionary<string, ConsoleLogger>();
 
-            _optionsReloadToken = _options.OnChange(ReloadLoggerOptions);
             ReloadLoggerOptions(options.CurrentValue);
+            _optionsReloadToken = _options.OnChange(ReloadLoggerOptions);
 
             _messageQueue = new ConsoleLoggerProcessor();
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
--- a/src/Logging/test/ConsoleLoggerTest.cs
+++ b/src/Logging/test/ConsoleLoggerTest.cs
@@ -1002,7 +1002,7 @@ namespace Microsoft.Extensions.Logging.Test
         }
 
         [Fact]
-        public void ConsoleLoggerOptions_DisableColors_IsReloaded()
+        public void ConsoleLoggerOptions_TimeStampFormat_IsReloaded()
         {
             // Arrange
             var monitor = new TestOptionsMonitor(new ConsoleLoggerOptions());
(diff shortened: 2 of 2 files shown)


BrennanConroy:
Are you actually seeing an issue?

This is how reload registration works. When the token is triggered you need to get a new token and register the callback for when the next options change happens.

OpenSourceAries:
Sorry, not seeing related issues.

My English is not good enough to explain, Hope the Demo could explain my thought.

``` 
public class FakeOptionsMonitor
{
    internal event Action<string> _onChange;

    public void SimulationSourceChanged(string newOptions)
    {
        _onChange?.Invoke(newOptions);
    }

    public IDisposable OnChange(Action<string> listener)
    {
        var disposable = new ChangeTrackerDisposable(this, listener);
        _onChange += disposable.OnChange;
        return disposable;
    }
}

internal class ChangeTrackerDisposable : IDisposable
{
    private readonly Action<string> _listener;
    private readonly FakeOptionsMonitor _monitor;

    public ChangeTrackerDisposable(FakeOptionsMonitor monitor, Action<string> listener)
    {
        _listener = listener;
        _monitor = monitor;
    }

    public void OnChange(string name) => _listener.Invoke(name);

    public void Dispose() => _monitor._onChange -= OnChange;
}
```

Main:

```
public static void ReloadFakeOptionsChange(string fakeNewOptions)
{
    Console.WriteLine($"Change: {fakeNewOptions}");
    _optionsReloadToken = _fakeOptionsMonitor.OnChange(ReloadFakeOptionsChange);
}

private static FakeOptionsMonitor _fakeOptionsMonitor;
private static IDisposable _optionsReloadToken;
private static void Main(string[] args)
{
    _fakeOptionsMonitor = new FakeOptionsMonitor();
    ReloadFakeOptionsChange("Initiliaze");

    string input = string.Empty;
    while (!string.IsNullOrEmpty(input = Console.ReadLine()))
    {
        Console.WriteLine($"Input: {input}");
        _fakeOptionsMonitor.SimulationSourceChanged(input);
    }
}
```

Test result:  image

Modiy Main:

```
public static void ReloadFakeOptionsChange(string fakeNewOptions)
{
    Console.WriteLine($"Change: {fakeNewOptions}");
}

private static FakeOptionsMonitor _fakeOptionsMonitor;
private static IDisposable _optionsReloadToken;
private static void Main(string[] args)
{
    _fakeOptionsMonitor = new FakeOptionsMonitor();
    _optionsReloadToken = _fakeOptionsMonitor.OnChange(ReloadFakeOptionsChange);
    ReloadFakeOptionsChange("Initiliaze");

    string input = string.Empty;
    while (!string.IsNullOrEmpty(input = Console.ReadLine()))
    {
        Console.WriteLine($"Input: {input}");
        _fakeOptionsMonitor.SimulationSourceChanged(input);
    }
}
```

Or

```
public static void ReloadFakeOptionsChange(string fakeNewOptions)
{
    Console.WriteLine($"Change: {fakeNewOptions}");
    _optionsReloadToken?.Dispose();
    _optionsReloadToken = _fakeOptionsMonitor.OnChange(ReloadFakeOptionsChange);
}

private static FakeOptionsMonitor _fakeOptionsMonitor;
private static IDisposable _optionsReloadToken;
private static void Main(string[] args)
{
    _fakeOptionsMonitor = new FakeOptionsMonitor();
    ReloadFakeOptionsChange("Initiliaze");

    string input = string.Empty;
    while (!string.IsNullOrEmpty(input = Console.ReadLine()))
    {
        Console.WriteLine($"Input: {input}");
        _fakeOptionsMonitor.SimulationSourceChanged(input);
    }
}
```

Test result: image


It's OK to close this PR if you feel I am wrong or you have better solution.

rynowak:
In your first example, you're relying on `ChangeTrackerDisposable .Dispose` to unhook an event handler, but nothing calls dispose on it. 

Since this line runs each time a change is triggered you're going to end up subscribing N times.
```C#
_optionsReloadToken = _fakeOptionsMonitor.OnChange(ReloadFakeOptionsChange);
```

Each time that line runs, you create a new instance of `ChangeTrackerDisposable` but since you didn't dispose the old one, it's still alive and still listening.

OpenSourceAries:
@rynowak  yes, this is what I want to fix through this PR.

OpenSourceAries:
I think the first example could cover the implementation of _ConsoleLoggerProvider_ before this PR.

BrennanConroy:
I just realized this is using `IOptionsMonitor` and not `ChangeToken`. I believe options monitor will handle the underlying change token so registration only needs to happen once.

src/Logging/Logging.Console/src/ConsoleLoggerProvider.cs
```diff
@@ -23,6 +23,7 @@ public ConsoleLoggerProvider(IOptionsMonitor<ConsoleLoggerOptions> options)
             _options = options;
             _loggers = new ConcurrentDictionary<string, ConsoleLogger>();
 
+            _optionsReloadToken = _options.OnChange(ReloadLoggerOptions);
```
BrennanConroy: Could you move this to be below `ReloadLoggerOptions`?

OpenSourceAries: Yes. This is just to keep consistent with other logger providers(such as [Extensions/src/Logging/Logging.AzureAppServices/src/BatchingLoggerProvider.cs](https://github.com/aspnet/Extensions/blob/master/src/Logging/Logging.AzureAppServices/src/BatchingLoggerProvider.cs))

BrennanConroy:
Thanks for finding and fixing this!

PR #1990, opened 2019-07-09
In JSInterop calls, use expected value for 'this'. Fixes #1477

### LINKED ISSUE #1477: JSInterop cannot invoke instance methods
### Describe the bug
JSInterop cannot invoke instance methods on javascript objects.

### To Reproduce
1. Using version Blazor preview4
2. Run this code `jsruntime.Invoke<object>("localStorage.setItem", "key", "value");`
3. JS Console (in firefox) prints
```
Microsoft.JSInterop.JSException: 'setItem' called on an object that does not implement interface Storage. blazor.webassembly.js:1:32027
invokeJSFromDotNet@http://localhost:60352/_framework/blazor.webassembly.js:1:5671
```

This happens because https://github.com/aspnet/Extensions/blob/bbe51d0fda2649105515b9fa450aca4948bcf9c9/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts#L160 uses apply and passes `null` as the instance argument for the invocation.

### Expected behavior
The `setItem` method is called on the `localStorage` instance and not called as a static method

refs https://github.com/aspnet/AspNetCore/issues/5563

SteveSandersonMS:
Previously, if you did this:

```cs
jsInterop.InvokeAsync<string>("a.b", arg1, arg2)
```

it would be equivalent to `window.a.b.apply(null, [arg1, arg2])`. That's OK as long as the function doesn't require any particular value for `this`. But unfortunately some static functions do expect a `this` value.

After this PR, the same call becomes equivalent to `window.a.b.apply(window.a.b, [arg1, arg2])`, or more simply `a.b(arg1, arg2)`, which is much more likely to match the developer's intent.

As a realistic example, it means you can do things like:

```cs
jsInterop.InvokeAsync<string>("localStorage.getItem", "myKey")
```

... and it will just work without needing any JS wrapper library.

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
+++ b/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
@@ -142,14 +142,6 @@ module DotNet {
    */
   export const jsCallDispatcher = {
     /**
-     * Finds the JavaScript function matching the specified identifier.
-     *
-     * @param identifier Identifies the globally-reachable function to be returned.
-     * @returns A Function instance.
-     */
-    findJSFunction,
-
-    /**
      * Invokes the specified synchronous JavaScript function.
      *
      * @param identifier Identifies the globally-reachable function to invoke.
@@ -227,8 +219,10 @@ module DotNet {
 
     let result: any = window;
     let resultIdentifier = 'window';
+    let lastSegmentValue: any;
     identifier.split('.').forEach(segment => {
       if (segment in result) {
+        lastSegmentValue = result;
         result = result[segment];
         resultIdentifier += '.' + segment;
       } else {
@@ -237,6 +231,8 @@ module DotNet {
     });
 
     if (result instanceof Function) {
+      result = result.bind(lastSegmentValue);
+      cachedJSFunctions[identifier] = result;
       return result;
     } else {
       throw new Error(`The value '${resultIdentifier}' is not a function.`);


src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -141,14 +141,6 @@ module DotNet {
    * Receives incoming calls from .NET and dispatches them to JavaScript.
    */
   export const jsCallDispatcher = {
-    /**
-     * Finds the JavaScript function matching the specified identifier.
-     *
-     * @param identifier Identifies the globally-reachable function to be returned.
-     * @returns A Function instance.
-     */
-    findJSFunction,
```
SteveSandersonMS: This was just not used. As in, there was no reason to declare it publicly since it was only used internally.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -237,6 +231,8 @@ module DotNet {
     });
 
     if (result instanceof Function) {
+      result = result.bind(lastSegmentValue);
+      cachedJSFunctions[identifier] = result;
```
SteveSandersonMS: Previously we also had a bug where it never cached the lookup results so it always recreated them on each call.

javiercn: Is it reasonable to have a test for this?

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -237,6 +231,8 @@ module DotNet {
     });
 
     if (result instanceof Function) {
+      result = result.bind(lastSegmentValue);
```
javiercn: Can we have a test for this? Maybe an E2E test?

SteveSandersonMS: Yes, but not in this repo

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
SteveSandersonMS: I'll see if this is a reasonable thing to add to JS interop E2E tests (though of course they are in a different repo).

javiercn: That's fine, I think we still want to know if this breaks.

PR #1999, opened 2019-07-10
Consistently handle prefixes in AddEnvironmentVariables

declspec:
 - Resolves an issue where Connection Strings from Azure App Services are not able to be imported when you call `AddEnvironmentVariables(string prefix)`. 

`AzureEnvToAppEnv(...)` will always yield dictionary entries with keys starting with `ConnectionString:` for Connection Strings detected from Azure. The environment variable filter will then immediately filter these out as they can never start with the user-defined `_prefix` unless it's either empty or literally `ConnectionString:`. 

Addresses: https://github.com/aspnet/Extensions/issues/1306

### DIFF
--- a/src/Configuration/Config.EnvironmentVariables/test/EnvironmentVariablesTest.cs
+++ b/src/Configuration/Config.EnvironmentVariables/test/EnvironmentVariablesTest.cs
@@ -85,23 +85,24 @@ namespace Microsoft.Extensions.Configuration.EnvironmentVariables.Test
         {
             var dict = new Hashtable()
             {
-                {"CUSTOMCONNSTR_db1", "CustomConnStr"},
-                {"SQLCONNSTR_db2", "SQLConnStr"},
-                {"MYSQLCONNSTR_db3", "MySQLConnStr"},
-                {"SQLAZURECONNSTR_db4", "SQLAzureConnStr"},
+                {"CUSTOMCONNSTR_PREFIX_db1", "CustomConnStr"},
+                {"SQLCONNSTR_PREFIX_db2", "SQLConnStr"},
+                {"MYSQLCONNSTR_PREFIX_db3", "MySQLConnStr"},
+                {"SQLAZURECONNSTR_PREFIX_db4", "SQLAzureConnStr"},
                 {"CommonEnv", "CommonEnvValue"},
             };
-            var envConfigSrc = new EnvironmentVariablesConfigurationProvider("ConnectionStrings:");
+            var envConfigSrc = new EnvironmentVariablesConfigurationProvider("PREFIX_");
 
             envConfigSrc.Load(dict);
 
-            Assert.Equal("CustomConnStr", envConfigSrc.Get("db1"));
-            Assert.Equal("SQLConnStr", envConfigSrc.Get("db2"));
-            Assert.Equal("System.Data.SqlClient", envConfigSrc.Get("db2_ProviderName"));
-            Assert.Equal("MySQLConnStr", envConfigSrc.Get("db3"));
-            Assert.Equal("MySql.Data.MySqlClient", envConfigSrc.Get("db3_ProviderName"));
-            Assert.Equal("SQLAzureConnStr", envConfigSrc.Get("db4"));
-            Assert.Equal("System.Data.SqlClient", envConfigSrc.Get("db4_ProviderName"));
+            Assert.Equal("CustomConnStr", envConfigSrc.Get("ConnectionStrings:db1"));
+            Assert.Equal("SQLConnStr", envConfigSrc.Get("ConnectionStrings:db2"));
+            Assert.Equal("System.Data.SqlClient", envConfigSrc.Get("ConnectionStrings:db2_ProviderName"));
+            Assert.Equal("MySQLConnStr", envConfigSrc.Get("ConnectionStrings:db3"));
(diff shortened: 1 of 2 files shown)


declspec:
This will apply the prefix filtering to the Connection Strings as well.
e.g
```cs
.AddEnvironmentVariables(prefix: "TEST_");
```

Will only capture ConnectionStrings starting with `TEST_`, i.e.:

```
ENV=
CUSTOMCONNSTR_Database=test <- not captured
CUSTOMCONNSTR_TEST_Database=test1 <- maps to ConnectionStrings:Database=test1
```

If this is not the desired behaviour, i.e. all ConnectionStrings should be included (essentially ignoring the prefix), then that'd be straightforward to implement as well.

anurse:
It seems like a reasonable change to me. Looking at the test, it does seem like it's approaching it the wrong way.

FYI, this layer is a bit ugly and we do want to refactor this at some point. Having the environment variable config provider have magic understanding of these prefixes is a bit ugly (I know @Tratcher dislikes it 😄). Doesn't mean we shouldn't fix a bug in it though!

Thoughts @HaoK @Tratcher ?

Tratcher:
Stop patching it and replace it?

Also, CUSTOMCONNSTR_TEST_Database is the wrong ordering for this behavior, it should be TEST_CUSTOMCONNSTR_Database. E.g. prefix should always be evaluated first.

anurse:
> Also, CUSTOMCONNSTR_TEST_Database is the wrong ordering for this behavior, it should be TEST_CUSTOMCONNSTR_Database. E.g. prefix should always be evaluated first.

I'm confused, isn't that completely counter to the scenario? As unpleasant as the layering is here, there is a specific scenario in mind and that scenario is broken. In Azure, you specify a connection string called `PREFIX_FOO`:

image

And it manifests as an environment variable: `MYSQLCONNSTR_PREFIX_FOO = MyConnectionString`

There's no way in this scenario for the environment variable to end up as `PREFIX_MYSQLCONNSTR_FOO = MyConnectionString`.

If you configure `PREFIX_` as a prefix in the environment variables provider, it correctly filters App Settings using that prefix, but **does not filter** Connection Strings and, in fact, completely fails to load any connection strings.

The desired behavior is for connection strings and app settings to be treated as they are entered into the portal.

> Stop patching it and replace it?

Yes, but that's not something we can do in 3.0 😄. The scenario exists and is in use. Honestly, the layering doesn't concern me a lot, it's just a convention that aligns with how Azure behaves.

src/Configuration/Config.EnvironmentVariables/test/EnvironmentVariablesTest.cs
```diff
@@ -85,23 +85,24 @@ public void LoadKeyValuePairsFromAzureEnvironmentWithPrefix()
...
             envConfigSrc.Load(dict);
 
-            Assert.Equal("CustomConnStr", envConfigSrc.Get("db1"));
-            Assert.Equal("SQLConnStr", envConfigSrc.Get("db2"));
-            Assert.Equal("System.Data.SqlClient", envConfigSrc.Get("db2_ProviderName"));
-            Assert.Equal("MySQLConnStr", envConfigSrc.Get("db3"));
-            Assert.Equal("MySql.Data.MySqlClient", envConfigSrc.Get("db3_ProviderName"));
-            Assert.Equal("SQLAzureConnStr", envConfigSrc.Get("db4"));
-            Assert.Equal("System.Data.SqlClient", envConfigSrc.Get("db4_ProviderName"));
+            Assert.Equal("CustomConnStr", envConfigSrc.Get("ConnectionStrings:db1"));
```
HaoK: Can we leave existing tests alone and add a new test for the new behavior?

declspec: @HaoK This test as it currently stands breaks with the changes

declspec:
> Stop patching it and replace it?
> 
> Also, CUSTOMCONNSTR_TEST_Database is the wrong ordering for this behavior, it should be TEST_CUSTOMCONNSTR_Database. E.g. prefix should always be evaluated first. Otherwise you'll never be able to properly separate these layers.

@Tratcher I mentioned this a bit in my comment [here](https://github.com/aspnet/Extensions/issues/1306#issuecomment-510098005), what your suggesting would be the most consistent behaviour for the prefix; it's just a simple pre-check on the environment variables and then anything that happens after that is a separate concern. 

However, as @anurse mentioned, the `ConnectionString` environment variables are often set through a UI like Azure's Portal which does the mapping to an environment variable for you so you don't have an opportunity to set a prefix to the beginning of it.

src/Configuration/Config.EnvironmentVariables/test/EnvironmentVariablesTest.cs
HaoK: Right which makes this change an regression then

declspec: Yes, it would be a regression. Do you know why anyone wanted this behaviour in the first place? Supporting this one weird case (unwrapping a magically-mapped Azure ConnectionString) is what's caused every other prefix to be completely broken.

pranavkm:
@anurse I moved this PR to 5.0.0-preview1, Could we get a resolution on accepting this?

PR #2000, opened 2019-07-10
Use TimeSpan value for reload interval

henkmollema:
`Milliseconds` returns the wrong, `TotalMilliseconds` should be used. But we can pass in the `TimeSpan` value directly as well.


henkmollema:
cc @pakrym as he finished and merged #1054

anurse:
Thanks!

henkmollema:
Welcome! Hopefully this prevents setting a few servers on fire. :smile:

PR #2002, opened 2019-07-10
Bring back required export jsCallDispatcher.findJSFunction

SteveSandersonMS:
I got this wrong yesterday. I removed this export because I thought it was unused. In fact, it's not used from any code in the `extensions` or `aspnetcore` repos, but it *is* used by Mono WebAssembly itself.

I've added a comment now to clarify why this is required.


javiercn:
This is yet another reason to have jsinterop within asp.net core 😢 (Not blaming you for anything)

SteveSandersonMS:
Agreed. Or at the very least we should have some E2E tests of its usage within Mono WebAssembly in this repo.

PR #2043, opened 2019-07-15
[JSInterop] Updates the JSInterop abstractions to better support logging and diagnostics.

javiercn:
This PR is related to https://github.com/aspnet/AspNetCore/pull/12182. See that PR for more context on the practical applications of these changes.

* Allows the server to directly dispatch completions from a method hub by calling EndInvoke directly.
  * The client is responsible for handling exceptions thrown by this method.
  * We do it this way so that the server can log issues delivering the exception.
* Allows the server to better log completions for JS -> .NET async calls by overriding EndInvokeDotNet.
  * Removes the need for a specific endpoint to sanitize .NET exceptions.

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
+++ b/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
@@ -19,7 +19,24 @@ module DotNet {
    * @param dispatcher An object that can dispatch calls from JavaScript to a .NET runtime.
    */
   export function attachDispatcher(dispatcher: DotNetCallDispatcher) {
+    if (!dispatcher) {
+      throw new Error("dispatcher can't be null or undefined.");
+    }
+
     dotNetDispatcher = dispatcher;
+    if (!(dotNetDispatcher.endInvokeDotNetFromJS)) {
+      // We didn't get a concrete implementation, so provide a default one
+      dotNetDispatcher.endInvokeDotNetFromJS = (asyncHandle: number, succeeded: boolean, resultOrError: any) => {
+        return getRequiredDispatcher().beginInvokeDotNetFromJS(
+          0,
+          'Microsoft.JSInterop',
+          'DotNetDispatcher.EndInvoke',
+          null,
+          succeeded ?
+            JSON.stringify([asyncHandle, succeeded, resultOrError], argReplacer) :
+            JSON.stringify([asyncHandle, succeeded, resultOrError]));
+      };
+    }
   }
 
   /**
@@ -75,7 +92,7 @@ module DotNet {
     try {
       const argsJson = JSON.stringify(args, argReplacer);
       getRequiredDispatcher().beginInvokeDotNetFromJS(asyncCallId, assemblyName, methodIdentifier, dotNetObjectId, argsJson);
-    } catch(ex) {
+    } catch (ex) {
       // Synchronous failure
       completePendingCall(asyncCallId, false, ex);
     }
@@ -116,7 +133,7 @@ module DotNet {
   export interface DotNetCallDispatcher {
     /**
      * Optional. If implemented, invoked by the runtime to perform a synchronous call to a .NET method.
-     * 
+     *
      * @param assemblyName The short name (without key/version or .dll extension) of the .NET assembly holding the method to invoke. The value may be null when invoking instance methods.
      * @param methodIdentifier The identifier of the method to invoke. The method must have a [JSInvokable] attribute specifying this identifier.
      * @param dotNetObjectId If given, the call will be to an instance method on the specified DotNetObject. Pass null or undefined to call static methods.
@@ -135,6 +152,15 @@ module DotNet {
      * @param argsJson JSON representation of arguments to pass to the method.
      */
     beginInvokeDotNetFromJS(callId: number, assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, argsJson: string): void;
+
+    /**
+     * Invoked by the runtime to begin an asynchronous call to a .NET method.
+     *
+     * @param callId A value identifying the asynchronous operation. This value should be passed back in a later call from .NET to JS.
+     * @param succeded Whether the operation succeeded or not.
+     * @param resultOrError The result or the error from the async operation.
+     */
+    endInvokeDotNetFromJS?(callId: number, succeeded: boolean, resultOrError: any): void;
   }
 
   /**
@@ -183,8 +209,8 @@ module DotNet {
         // On completion, dispatch result back to .NET
         // Not using "await" because it codegens a lot of boilerplate
         promise.then(
-          result => getRequiredDispatcher().beginInvokeDotNetFromJS(0, 'Microsoft.JSInterop', 'DotNetDispatcher.EndInvoke', null, JSON.stringify([asyncHandle, true, result], argReplacer)),
-          error => getRequiredDispatcher().beginInvokeDotNetFromJS(0, 'Microsoft.JSInterop', 'DotNetDispatcher.EndInvoke', null, JSON.stringify([asyncHandle, false, formatError(error)]))
+          result => getRequiredDispatcher().endInvokeDotNetFromJS!(asyncHandle, true, result),
+          error => getRequiredDispatcher().endInvokeDotNetFromJS!(asyncHandle, false, formatError(error))
         );
       }
     },
@@ -219,7 +245,7 @@ module DotNet {
       return error ? error.toString() : 'null';
     }
   }
-  
+
   function findJSFunction(identifier: string): Function {
     if (cachedJSFunctions.hasOwnProperty(identifier)) {
       return cachedJSFunctions[identifier];
@@ -247,7 +273,7 @@ module DotNet {
     }
   }
 
-  class DotNetObject {    
+  class DotNetObject {
     constructor(private _id: number) {
     }
 
@@ -268,14 +294,14 @@ module DotNet {
     }
 
     public serializeAsArg() {
-      return {__dotNetObject: this._id};
+      return { __dotNetObject: this._id };
     }
   }
 
   const dotNetObjectRefKey = '__dotNetObject';
   attachReviver(function reviveDotNetObject(key: any, value: any) {
     if (value && typeof value === 'object' && value.hasOwnProperty(dotNetObjectRefKey)) {
-        return new DotNetObject(value.__dotNetObject);
+      return new DotNetObject(value.__dotNetObject);
     }
 
     // Unrecognized - let another reviver handle it
--- a/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
+++ b/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
@@ -24,19 +24,6 @@ module DotNet {
     }
 
     dotNetDispatcher = dispatcher;
-    if (!dotNetDispatcher.endInvokeDotNetFromJS) {
-      // We didn't get a concrete implementation, so provide a default one
-      dotNetDispatcher.endInvokeDotNetFromJS = (asyncHandle, succeeded, resultOrError, replacer) => {
-        return getRequiredDispatcher().beginInvokeDotNetFromJS(
-          0,
-          'Microsoft.JSInterop',
-          'DotNetDispatcher.EndInvoke',
-          null,
-          succeeded ?
-            JSON.stringify([asyncHandle, succeeded, resultOrError], replacer) :
-            JSON.stringify([asyncHandle, succeeded, resultOrError]));
-      };
-    }
   }
 
   /**
@@ -154,14 +141,13 @@ module DotNet {
     beginInvokeDotNetFromJS(callId: number, assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, argsJson: string): void;
 
     /**
-     * Invoked by the runtime to begin an asynchronous call to a .NET method.
+     * Invoked by the runtime to complete an asynchronous JavaScript function call started from .NET
      *
      * @param callId A value identifying the asynchronous operation. This value should be passed back in a later call from .NET to JS.
      * @param succeded Whether the operation succeeded or not.
-     * @param resultOrError The result or the error from the async operation.
-     * @param argReplacer A function that replaces [DotNetObject] references with the appropriate value during serialization.
+     * @param resultOrError The serialized result or the serialized error from the async operation.
      */
-    endInvokeDotNetFromJS?(callId: number, succeeded: boolean, resultOrError: any, argReplacer: (key: string, value: any) => any): void;
+    endInvokeJSFromDotNet(callId: number, succeeded: boolean, resultOrError: any): void;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -124,35 +124,37 @@ private async Task<T> InvokeWithDefaultCancellation<T>(string identifier, IEnume
...
         /// </summary>
-        /// <remarks>
-        /// This callback can be used in remote JS interop scenarios to sanitize exceptions that happen on the server to avoid disclosing
-        /// sensitive information to remote browser clients.
-        /// </remarks>
-        /// <param name="exception">The exception that occurred.</param>
-        /// <param name="assemblyName">The assembly for the invoked .NET method.</param>
-        /// <param name="methodIdentifier">The identifier for the invoked .NET method.</param>
-        /// <returns>An object containing information about the exception.</returns>
-        protected virtual object OnDotNetInvocationException(Exception exception, string assemblyName, string methodIdentifier) => exception.ToString();
```
javiercn: I decided to change the extensibility because it made logging harder and had some missing pieces of data `(dotNetObjectId)`

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -19,7 +19,24 @@ module DotNet {
    * @param dispatcher An object that can dispatch calls from JavaScript to a .NET runtime.
    */
   export function attachDispatcher(dispatcher: DotNetCallDispatcher) {
+    if (!dispatcher) {
+      throw new Error("dispatcher can't be null or undefined.");
+    }
+
     dotNetDispatcher = dispatcher;
+    if (!(dotNetDispatcher.endInvokeDotNetFromJS)) {
```
SteveSandersonMS:
```suggestion
    if (!dotNetDispatcher.endInvokeDotNetFromJS) {
```

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -135,6 +152,16 @@ module DotNet {
      * @param argsJson JSON representation of arguments to pass to the method.
      */
     beginInvokeDotNetFromJS(callId: number, assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, argsJson: string): void;
+
+    /**
+     * Invoked by the runtime to begin an asynchronous call to a .NET method.
```
SteveSandersonMS: The description doesn't appear to match the method name

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -135,6 +152,16 @@ module DotNet {
...
+     *
+     * @param callId A value identifying the asynchronous operation. This value should be passed back in a later call from .NET to JS.
+     * @param succeded Whether the operation succeeded or not.
+     * @param resultOrError The result or the error from the async operation.
+     * @param argReplacer A function that replaces [DotNetObject] references with the appropriate value during serialization.
```
SteveSandersonMS: This layering doesn't look right. Why don't we pass a pre-serialized value to `endInvokeDotNetFromJS`? That way the implementer doesn't need to know about the concept of `argReplacer` and can't accidentally do the serialization incorrectly.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -135,6 +152,16 @@ module DotNet {
...
+     */
+    endInvokeDotNetFromJS?(callId: number, succeeded: boolean, resultOrError: any, argReplacer: (key: string, value: any) => any): void;
```
SteveSandersonMS: Should this actually be called `endInvokeJSFromDotNet`? I can't understand the meaning of it otherwise. The default implementation works by making a call into .NET, which wouldn't make sense if we were completing a call that originated in JS (as the completion would be going to JS, not to .NET).

SteveSandersonMS: Also TBH I'd prefer for this to be a required property on the interface, and not to have a default implementation here. The implementation can be provided by the platform - we already have a way to plug in an implementation for WebAssembly separately from server-side Blazor. I don't see why we'd have a default implementation for this but not for `beginInvokeDotNetFromJS`.

So altogether perhaps this should be declared as:

```suggestion
    endInvokeJSFromDotNet(callId: number, succeeded: boolean, result: string): void;
```

... where the platform's job is simply to pass through the `callId` and `result` string to `DotNetDispatcher.EndInvoke` on the .NET side.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -19,7 +19,24 @@ module DotNet {
    * @param dispatcher An object that can dispatch calls from JavaScript to a .NET runtime.
    */
   export function attachDispatcher(dispatcher: DotNetCallDispatcher) {
+    if (!dispatcher) {
+      throw new Error("dispatcher can't be null or undefined.");
```
SteveSandersonMS: Should be single quotes, but TBH I think we'd be fine without this check. In most places in TypeScript we rely on compile-time strict nullness checking, since runtime checks aren't super robust anyway (just because a value is nonnull doesn't mean it has anything like the shape you're expecting at runtime).

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -244,6 +245,24 @@ static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
...
+        {
+            // We don't need to notify the client if a JS interop call fails to be processed on the server.
+            // The client needs to catch potential exceptions. We do it this way so that we can log errors
+            // processing the callback completion.
+            var parsedArgs = ParseArguments(
+                nameof(EndInvoke),
+                arguments,
+                EndInvokeParameterTypes);
+
+            EndInvoke((long)parsedArgs[0], (bool)parsedArgs[1], (JSAsyncCallResult)parsedArgs[2]);
```
SteveSandersonMS: Can we remove the other overload of `EndInvoke`, or at least make it private and not have `[JSInvokable]`, so there's only one entrypoint? I think now that `endInvokeJSFromDotNet` is becoming part of the platform contract, it's the platform's job to have some way of calling `EndInvoke`, and this library doesn't need to provide any `[JSInvokable]` version of it.

I know that means that a naive implementation in WebAssembly will be double-serializing the args, but we're always free in the future to special case that so that WebAssembly calls `EndInvoke(string arguments)` directly without going through JS interop.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -244,6 +245,24 @@ static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
...
+        /// <summary>
+        /// Receives notification that a call from .NET to JS has finished, marking the
+        /// associated <see cref="Task"/> as completed.
+        /// </summary>
+        /// <param name="arguments">The serialized arguments for the callback completion.</param>
+        public static void EndInvoke(string arguments)
+        {
+            // We don't need to notify the client if a JS interop call fails to be processed on the server.
+            // The client needs to catch potential exceptions. We do it this way so that we can log errors
+            // processing the callback completion.
```
SteveSandersonMS: This comment block is phrased strangely, because it's phrased in terms of server-side Blazor, which this library isn't meant to be coupled to.

Could this be rephrased in terms of .NET/JS interop?

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -124,35 +124,37 @@ private async Task<T> InvokeWithDefaultCancellation<T>(string identifier, IEnume
         protected abstract void BeginInvokeJS(long taskId, string identifier, string argsJson);
 
         /// <summary>
-        /// Allows derived classes to configure the information about an exception in a JS interop call that gets sent to JavaScript.
+        /// Completes an async JS interop call from JavaScript to DotNet.
```
SteveSandersonMS:
```suggestion
        /// Completes an async JS interop call from JavaScript to .NET
```

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: Ok, lets do that.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: I think I tried that, I don't remember the issue I ran into, I'll give it another try. I think it is because I wanted access to the raw error on the server to log it.

I can pass in callId, succeded, and Json.stringify([callId, succeeded, resultOrError], argReplacer) as arguments to this method, but then on the server I don't have a good way to simply access the error message without parsing the payload, which is what I wanted to avoid.

I think we might be ok simply logging serializedArgs on the sever, if that's the case we can simply pass in the serialized args here directly..

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: I'll update the comment.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: I removed the check for this (and the default implementation) I only added it because it was complaining below. That said, it can't be single quotes due to the single quote in the middle of the string. Normally when you have such quote is acceptable to use double quotes to surround the string.

We don't have ESLint setup in this repo, we could if we wanted to, but I'm not planning to do it now.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
SteveSandersonMS: > I think we might be ok simply logging serializedArgs on the sever,

👍

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
SteveSandersonMS: `\'` 😄

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
javiercn: I got rid of the [JSInvokable] on EndInvoke by providing a custom platform method inside Mono.WebAssembly.JSInterop

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: done

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: done

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: done

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
javiercn: got rid of it.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
javiercn: I rephrased it inside the comments

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
javiercn: done

javiercn:
@SteveSandersonMS anything else left here?

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -135,6 +135,15 @@ module DotNet {
      * @param argsJson JSON representation of arguments to pass to the method.
      */
     beginInvokeDotNetFromJS(callId: number, assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, argsJson: string): void;
+
+    /**
+     * Invoked by the runtime to complete an asynchronous JavaScript function call started from .NET
+     *
+     * @param callId A value identifying the asynchronous operation. This value should be passed back in a later call from .NET to JS.
```
SteveSandersonMS:
```suggestion
     * @param callId A value identifying the asynchronous operation.
```

src/JSInterop/Microsoft.JSInterop/src/JSAsyncCallResult.cs
```diff
@@ -23,8 +22,7 @@ namespace Microsoft.JSInterop.Internal
     /// <summary>
     /// Intended for framework use only.
     /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public sealed class JSAsyncCallResult
+    internal sealed class JSAsyncCallResult
```
SteveSandersonMS: Hooray!

src/JSInterop/Microsoft.JSInterop/test/DotNetDispatcherTest.cs
```diff
@@ -574,6 +575,27 @@ protected override string InvokeJS(string identifier, string argsJson)
...
+                {
+                    resultOrError = ex.ToString();
+                }
+                else if (!success && resultOrError is ExceptionDispatchInfo edi)
+                {
+                    resultOrError = edi.SourceException.ToString();
+                }
+
+                var args = JsonSerializer.Serialize(new[] { callId, success, resultOrError }, new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase });
+                BeginInvokeJS(0, "DotNet.jsCallDispatcher.endInvokeDotNetFromJS", args);
```
SteveSandersonMS: This seems super weird for a test class to be doing. It's pretty misleading to put production-like code in a test fake, as it suggests to test code that this behavior is inherent (e.g., in the base class). It's exceedingly strange for the unit tests to be making assertions that it does this JS interop call when that's not inherent to the contract any more.

Instead of doing this, could we have it behave similarly to the other overrides on the test class, i.e., log the args and let the unit tests make assertions about them?

src/JSInterop/Microsoft.JSInterop/test/JSRuntimeBaseTest.cs
```diff
@@ -283,14 +283,24 @@ public class BeginInvokeAsyncArgs
...
+                }
+                else if (!success && resultOrError is ExceptionDispatchInfo edi)
                 {
-                    return OnDotNetException(exception, assemblyName, methodName);
+                    resultOrError = edi.SourceException.ToString();
                 }
 
-                return base.OnDotNetInvocationException(exception, assemblyName, methodName);
+                var args = JsonSerializer.Serialize(new[] { callId, success, resultOrError }, new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase });
+                BeginInvokeJS(0, "DotNet.jsCallDispatcher.endInvokeDotNetFromJS", args);
```
SteveSandersonMS: Again since this is not part of the contract, I don't think the test class should do this. Can we update the unit tests to describe the contract as it now exists, not as it was before?

Issue #1323, opened 2019-03-28
Support Stop_Pending service status in Microsoft.Extensions.Hosting.WindowsServices


xqrzd:
Currently when stopping a service in Microsoft.Extensions.Hosting.WindowsServices, the service is immediately reports to SCM that it's stopped, even though the process may still be running. A fix would be to simply wait in ServiceBase.OnStop https://github.com/aspnet/Extensions/blob/master/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs#L72 (maybe wait on IHostApplicationLifetime.ApplicationStopped?)

I'm happy to attempt a PR if this is worth doing. The same thing could be done on startup if it's useful.

Tratcher:
Yes, please give it a try and let us know if it works as expected.

PR #2048, opened 2019-07-16
Wait for host to shutdown before marking service stopped

xqrzd:
When the ApplicationStopping event is fired (either by SCM or the application), transition the service to StopPending. Then, wait in OnStop() until ApplicationStopped is fired, to keep the service in the StopPending state.

Task.Run is used in ApplicationStopping to avoid a deadlock because it waits for ApplicationStopped.

Addresses #1323

### DIFF
--- a/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
+++ b/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
 {
     public class WindowsServiceLifetime : ServiceBase, IHostLifetime
     {
-        private readonly TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>();
+        private readonly TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
         private readonly ManualResetEventSlim _delayStop = new ManualResetEventSlim();
 
         public WindowsServiceLifetime(IHostEnvironment environment, IHostApplicationLifetime applicationLifetime, ILoggerFactory loggerFactory)
--- a/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
+++ b/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
@@ -87,14 +87,7 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
         {
             ApplicationLifetime.StopApplication();
             // Wait for the host to shutdown before marking service as stopped.
-            if (_hostOptions == null)
-            {
-                _delayStop.Wait();
-            }
-            else
-            {
-                _delayStop.Wait(_hostOptions.ShutdownTimeout);
-            }
+            _delayStop.Wait(_hostOptions.ShutdownTimeout);
             base.OnStop();
         }
 
--- a/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
+++ b/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
@@ -6,6 +6,7 @@ using System.ServiceProcess;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.Hosting.WindowsServices
 {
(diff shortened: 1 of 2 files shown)


src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -11,7 +11,8 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
 {
     public class WindowsServiceLifetime : ServiceBase, IHostLifetime
     {
-        private TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>();
+        private readonly TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>();
```
Tratcher:
```suggestion
        private readonly TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
```

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -56,7 +62,7 @@ private void Run()
 
         public Task StopAsync(CancellationToken cancellationToken)
         {
-            Stop();
+            ApplicationLifetime.StopApplication();
```
Tratcher: StopAsync is part of IHostLifetime, not ServiceBase. This is called when the app is already shutting down and we need to signal ServiceBase. You should not have needed to change this.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -35,6 +36,11 @@ public Task WaitForStartAsync(CancellationToken cancellationToken)
             ApplicationLifetime.ApplicationStopping.Register(() =>
             {
                 Logger.LogInformation("Application is shutting down...");
+                Task.Run(Stop); // Transition the service to StopPending.
```
Tratcher: Should not be needed, see StopAsync.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
xqrzd: I can add it back, is calling it in Task.Run ok? Otherwise it deadlocks because the host waits for StopAsync before triggering ApplicationStopped, and this method won't return until ApplicationStopped is fired.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
xqrzd: It isn't necessary, but it has the nice effect of marking the service as StopPending if the service shuts itself down. For instance, without calling Stop as soon as the application is stopping, this application won't display a StopPending status as it's stopping,

```csharp
class Program
{
    static async Task Main(string[] args)
    {
        using var cts = new CancellationTokenSource(5000);
        using var host = Host.CreateDefaultBuilder(args)
            .UseWindowsService()
            {
                services.AddHostedService<MyService>();
                services.Configure<HostOptions>(opt => opt.ShutdownTimeout = TimeSpan.FromSeconds(30));
            })
            .Build();

        await host.RunAsync(cts.Token);
    }
}

public class MyService : IHostedService
{
    public Task StartAsync(CancellationToken cancellationToken)
    {
        return Task.CompletedTask;
    }

    public async Task StopAsync(CancellationToken cancellationToken)
    {
        await Task.Delay(5000);
    }
}
```

This would be easier if ServiceBase had a SetStatus method to arbitrarily change the service's status.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
Tratcher: You want to move the Task.Run(Stop) from ApplicationStopping to here? That's OK so long as you add comments as to why.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -72,6 +78,7 @@ protected override void OnStart(string[] args)
         protected override void OnStop()
         {
             ApplicationLifetime.StopApplication();
+            _delayStop.Wait();
```
Tratcher: This should either time out or you should add IDisposible to this class and also call _delayStop.Set() there. There are error conditions where ApplicationStopped might not be called but Dispose would be.

Tratcher:
Code check failure, the reference assemblies need to be updated: "Generated code is not up to date. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)"

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -72,7 +86,26 @@ protected override void OnStart(string[] args)
         protected override void OnStop()
         {
             ApplicationLifetime.StopApplication();
+            // Wait for the host to shutdown before marking service as stopped.
+            if (_hostOptions == null)
```
Tratcher: _hostOptions should never be null.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
xqrzd: I ended up doing both to be safe

Tratcher:
Thanks

PR #2049, opened 2019-07-16
Use JsonConverter for DotNetObjectRef

pranavkm:


src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRef.cs
```diff
@@ -15,7 +15,8 @@ public static class DotNetObjectRef
         /// <returns>An instance of <see cref="DotNetObjectRef{TValue}" />.</returns>
         public static DotNetObjectRef<TValue> Create<TValue>(TValue value) where TValue : class
         {
-            return new DotNetObjectRef<TValue>(value);
+            var objectId = DotNetObjectRefManager.Current.TrackObject(value);
```
pranavkm: We were doing this in the ctor of `DotNetObjectRef<T>` in the past. However, we don't actually need the ref instance to be cached - it's never used. This just makes it much cleaner.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRefJsonConverterFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.JSInterop
+{
+    internal sealed class DotNetObjectReferenceJsonConverterFactory : JsonConverterFactory
+    {
+        public override bool CanConvert(Type typeToConvert)
+        {
+            Debug.Assert(
+                typeToConvert.IsGenericType && typeToConvert.GetGenericTypeDefinition() == typeof(DotNetObjectRef<>),
+                "We expect this to only be called for DotNetObjectRef<T> instances.");
+            return true;
```
rynowak: This kind of thing is broken. It won't actually throw an exception or prevent this from happening. It would be better to just write the implementation correctly (return false for unsupported).

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectReferenceJsonConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+            var dotNetObjectId = reader.GetInt64();
+
+            if (!reader.Read())
+            {
+                // We need to read all the data that was given to us.
+                throw new InvalidDataException("Invalid DotNetObjectRef JSON.");
+            }
+
+            var value = (TValue)DotNetObjectRefManager.Current.FindDotNetObject(dotNetObjectId);
+            return new DotNetObjectRef<TValue>(dotNetObjectId, value);
```
rynowak: Generally these APIs can be used in a simpler way. It's find to just let the serializer throw for things like the wrong property type.

Use JsonException here instead of InvalidDataException, but most of the time the reader should be throwing for you.

src/JSInterop/Microsoft.JSInterop/src/IDotNetObjectRef.cs
```diff
@@ -7,6 +7,5 @@ namespace Microsoft.JSInterop
 {
     internal interface IDotNetObjectRef : IDisposable
```
rynowak: Is this still used?

src/JSInterop/Microsoft.JSInterop/test/DotNetObjectReferenceJsonConverterTest.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Microsoft.JSInterop.Tests
+{
+    public class DotNetObjectReferenceJsonConverterTest
+    {
+        [Fact]
+        public Task Read_ReadsJson() => WithJSRuntime(_ =>
+        {
+            // Arrange
+            // Throw-away value
+            DotNetObjectRef.Create(new TestModel());
```
rynowak: What's the purpose of doing this?

pranavkm: Making sure we're reviving the right object.

src/JSInterop/Microsoft.JSInterop/src/IDotNetObjectRef.cs
pranavkm: We use the marker interface in the dispatcher: https://github.com/aspnet/Extensions/blob/master/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs#L243. We could have it do the same check as the one in the factory

src/JSInterop/Microsoft.JSInterop/test/DotNetObjectReferenceJsonConverterTest.cs
rynowak: Sorry I super don't get it 😆 this isn't used anywhere - what's the side-effect that's important?

src/JSInterop/Microsoft.JSInterop/src/IDotNetObjectRef.cs
rynowak: That's fine, I just wanted to make sure it wasn't a mistake.

src/JSInterop/Microsoft.JSInterop/test/DotNetObjectReferenceJsonConverterTest.cs
pranavkm: I can add a more detailed comment. Creating adds an id which is what we use to look up things. I want to verify we're restoring the right instance on deserialization.

pranavkm:
🆙 📅

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectReferenceJsonConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    internal sealed class DotNetObjectReferenceJsonConverter<TValue> : JsonConverter<DotNetObjectRef<TValue>> where TValue : class
+    {
+        private static JsonEncodedText DotNetObjectRefKey => DotNetDispatcher.DotNetObjectRefKey;
+
+        public override DotNetObjectRef<TValue> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            long dotNetObjectId = 0;
+
+            while (reader.Read() && reader.TokenType != JsonTokenType.EndObject)
```
rynowak: One thing about this - I don't think it would detect an incomplete object. `{ __id: "foo"`

src/JSInterop/Microsoft.JSInterop/test/DotNetObjectReferenceJsonConverterTest.cs
rynowak: Still waiting for that more detailed comment.

rynowak: How does this do anything? - it seems like it validates some code we didn't write doesn't exist.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectReferenceJsonConverter.cs
pranavkm: The reader should throw at this point - it'll attempt to read till the end of token and complain when it doesn't find a terminator.

src/JSInterop/Microsoft.JSInterop/test/DotNetObjectReferenceJsonConverterTest.cs
pranavkm: It's validating that we're dereferencing the correct instance when we track more than one instance. Do you have suggestions about how we would validate this or what text would help you understand this test better?

rynowak: It would make a lot more sense if you compared the object ids to show that they are different then. A comment like `throw away value` should be replace with "we're doing this for the side effect of making sure that more than one value is being tracked"

Issue #1363, opened 2019-04-07
Hang when calling start/stop on IHost


glennc:
The program below appears to hang after logging the application is shutting down message and I don't understand why. The OnProcessExit event appears to be waiting here https://github.com/aspnet/Extensions/blob/master/src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs#L79 but I don't know why that would be or if that is a problem or just a red herring. 

```csharp
    public class Program
    {
        public static async Task Main(string[] args)
        {
            var host = CreateHostBuilder(args).Build();
            await host.StartAsync();
            await host.StopAsync();
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureServices(services => services.AddHostedService<Worker>());
    }
```

@Tratcher @davidfowl

glennc:
The StopAsync isn't actually required to reproduce, if you let the process exit the same thing happens.

Tratcher:
It's not actually unblocked until Dispose.
https://github.com/aspnet/Extensions/blob/4b123acda735f7a7e20b62d9cfafabe5707290a0/src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs#L99

davidfowl:
That might be problematic.

Tratcher:
Yeah. We did it that way to try to let app cleanup run on app exit events.

glennc:
OK, that's surprising. Can it not be tied to something like applicationLifetime.Stopped that gets raised after the host has finished stopping all the hosted services?

This means if you're not calling Run then you want to put the host in a using, which I don't think we've ever told anyone they'd have to do.

xqrzd:
It was done there to allow any hosted services / dependencies to be disposed before the process exits. Prior to that change on Linux sigterm would effectively kill the process. If the event was tied to applicationLifetime.Stopped it would allow any hosted services to stop, but not be disposed. I think a proper fix requires a .net runtime change, something like `DontExitProcessUntilMainMethodHasReturned` or something like that.

Tratcher:
Note to self: Could address this with a finalizer?

davidfowl:
I'm not sure that would work. The host is rooted by the DI container which is rooted by the callback, which is stored in a static.

tmds:
```
        private void OnProcessExit(object sender, EventArgs e)
        {
            ApplicationLifetime.StopApplication();
[...]

PR #2053, opened 2019-07-18
Add logging for the hanging case when calling start/stop

maryamariyan:
 - Created a new console app project under `src/Hosting/samples` that reproduces the hang on start/stop.
 - Added logging for when we get the hanging case.

Note: This PR does not fix the hang, instead it makes the tactical fix explained in [this comment](https://github.com/aspnet/Extensions/issues/1363#issuecomment-511966883)

Addresses #1363 

cc: @anurse @Tratcher


src/Hosting/test/testassets/Microsoft.Extensions.Hosting.TestApp/Program.cs
```diff
@@ -25,42 +25,39 @@ public static void Main(string[] args)
...
+                var config = host.Services.GetRequiredService<IConfiguration>();
+                var lifetime = host.Services.GetRequiredService<IHostApplicationLifetime>();
 
-            if (config["STARTMECHANIC"] == "Run")
-            {
-                host.Run();
-            }
-            else if (config["STARTMECHANIC"] == "WaitForShutdown")
-            {
-                using (host)
```
maryamariyan: the using was down below here, I moved it up.

maryamariyan:
most of the test code already had using block around IHostBuilder.Build() calls, I changed all missing ones I could find.

src/Hosting/samples/SampleForHang/Program.cs
```diff
@@ -0,0 +1,22 @@
+
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+
+namespace SampleForHang
```
Tratcher: Adding a new project for this wasn't necessary, remove it. A project can have multiple Program classes, there's an attribute in the project for which one to run.
https://github.com/aspnet/Extensions/blob/ab1b0aa16453b7c5876365f5dc6d476153873af0/src/Hosting/samples/GenericHostSample/ProgramFullControl.cs#L10

src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
```diff
@@ -76,6 +76,10 @@ private void OnApplicationStopping()
         private void OnProcessExit(object sender, EventArgs e)
         {
             ApplicationLifetime.StopApplication();
+            if(!_shutdownBlock.WaitOne(TimeSpan.FromMinutes(1)))
```
Tratcher: @anurse how long would we really need to wait here? The default shutdown timeout is only 5s. The longer you wait the more likely someone will just kill it and never see the log.
Could we pull the HostOptions from DI and just use that timeout?
https://github.com/aspnet/Extensions/blob/ab1b0aa16453b7c5876365f5dc6d476153873af0/src/Hosting/Hosting/src/HostOptions.cs#L16

anurse: Hrm... given that we have that timeout, I wonder if we should just block for that duration and then exit rather than this whole thing about logging and waiting indefinitely.

But yeah, at the least, we can use it as the timeout before logging.

@maryamariyan you can get access to the `HostOptions` object @Tratcher mentioned by adding a constructor parameter to `ConsoleLifetime` of type `IOptions<HostOptions>`. The DI system will automatically inject it for you and you can access properties on the `HostOptions` through that.

src/Hosting/samples/SampleForHang/Program.cs
anurse: Ah, I forgot about that, thanks!

src/Shared/testassets/CreateHostBuilderPatternTestSite/Program.cs
```diff
@@ -9,7 +9,7 @@ public class Program
     {
         public static void Main(string[] args)
         {
-            var webHost = CreateHostBuilder(args).Build();
+            using (var webHost = CreateHostBuilder(args).Build()) { }
```
Tratcher: This is causing a strange compiler error on the CI. Run build.cmd locally and see if you can track it down.
```
Program.cs(12,20): error CS1674: 'IHost': type used in a using statement must be implicitly convertible to 'System.IDisposable' or implement a suitable 'Dispose' method. [F:\workspace\_work\1\s\src\Shared\testassets\CreateHostBuilderPatternTestSite\CreateHostBuilderPatternTestSite.csproj]
[error]Program.cs(12,20): error CS1674: 'IHost': type used in a using statement must be implicitly convertible to 'System.IDisposable' or implement a suitable 'Dispose' method.
Program.cs(12,20): error CS1674: 'IHost': type used in a using statement must be implicitly convertible to 'System.IDisposable' or implement a suitable 'Dispose' method. [F:\workspace\_work\1\s\src\Shared\testassets\CreateHostBuilderPatternTestSite\CreateHostBuilderPatternTestSite.csproj]
[error]Program.cs(12,20): error CS1674: 'IHost': type used in a using statement must be implicitly convertible to 'System.IDisposable' or implement a suitable 'Dispose' method.
```

Edit: I see the issue. This test is testing the class structure of Program.CreateHostBuilder, but it uses mocks for the types.
https://github.com/aspnet/Extensions/blob/ae950eba800c2c70da9a57bf183bea75f0849d31/src/Shared/testassets/MockHostTypes/IHost.cs
You can just revert this file.

src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
```diff
@@ -19,14 +19,15 @@ public class ConsoleLifetime : IHostLifetime, IDisposable
         private CancellationTokenRegistration _applicationStartedRegistration;
         private CancellationTokenRegistration _applicationStoppingRegistration;
 
-        public ConsoleLifetime(IOptions<ConsoleLifetimeOptions> options, IHostEnvironment environment, IHostApplicationLifetime applicationLifetime)
-            : this(options, environment, applicationLifetime, NullLoggerFactory.Instance) { }
+        public ConsoleLifetime(IOptions<ConsoleLifetimeOptions> options, IHostEnvironment environment, IHostApplicationLifetime applicationLifetime, IOptions<HostOptions> hostOptions)
```
Tratcher: Changing this signature requires updating ref assemblies.
https://github.com/aspnet/Extensions/blob/master/docs/ReferenceAssemblies.md

maryamariyan: expected to change both netstandard 2.0 and netcoreapp3.0? or just netcoreapp3.0? the script auto-populated for both.

Tratcher: Yeah, both.

Tratcher:
Thanks

PR #2061, opened 2019-07-19
Use Utf8JsonReader in DotNetDispatcher

pranavkm:


BrennanConroy:
Numbers?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -172,74 +171,63 @@ private static object[] ParseArguments(string methodIdentifier, string argsJson,
                 return Array.Empty<object>();
             }
 
-            // There's no direct way to say we want to deserialize as an array with heterogenous
-            // entry types (e.g., [string, int, bool]), so we need to deserialize in two phases.
-            var jsonDocument = JsonDocument.Parse(argsJson);
-            var shouldDisposeJsonDocument = true;
-            try
+            var utf8JsonBytes = Encoding.UTF8.GetBytes(argsJson).AsSpan();
```
BrennanConroy: You should be able to avoid this allocation by stackalloc/renting and using
https://github.com/dotnet/corefx/blob/7431e4938c461f6e24cf3122c0b9ca4eb27e4dde/src/Common/src/CoreLib/System/Text/Unicode/Utf8.cs#L42

pranavkm: This project targets netstandard2.0. I don't that API exists?

BrennanConroy: Or even better
https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Text/UTF8Encoding.cs#L371

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -172,74 +171,63 @@ private static object[] ParseArguments(string methodIdentifier, string argsJson,
...
-                if (suppliedArgsLength != parameterTypes.Length)
-                {
-                    throw new ArgumentException($"In call to '{methodIdentifier}', expected {parameterTypes.Length} parameters but received {suppliedArgsLength}.");
-                }
+            // Check if we have the right number of tokens
+            var suppliedArgsLength = 0;
+            while (reader.Read() && reader.TokenType != JsonTokenType.EndArray)
+            {
+                suppliedArgsLength++;
+                reader.Skip();
```
ahsonkhan: nit: I think this can be written a bit differently to avoid the double parsing of the json data.

Currently we are doing one pass to validate, and then the other to do the actual work.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -172,74 +171,63 @@ private static object[] ParseArguments(string methodIdentifier, string argsJson,
...
-                    {
-                        throw new InvalidOperationException($"In call to '{methodIdentifier}', parameter of type '{parameterType.Name}' at index {(index + 1)} must be declared as type 'DotNetObjectRef<{parameterType.Name}>' to receive the incoming value.");
-                    }
-                    else
-                    {
-                        suppliedArgs[index] = JsonSerializer.Deserialize(item.GetRawText(), parameterType, JsonSerializerOptionsProvider.Options);
-                    }
+            // Reset the reader
+            reader = new Utf8JsonReader(utf8JsonBytes);
+            reader.Read();
```
ahsonkhan: nit: Add `Debug.Assert(reader.TokenType == JsonTokenType.StartArray)`

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -172,74 +171,63 @@ private static object[] ParseArguments(string methodIdentifier, string argsJson,
...
                 // an incorrect use if there's a object that looks like { '__dotNetObject': <some number> },
                 // but we aren't assigning to DotNetObjectRef{T}.
-                return item.ValueKind == JsonValueKind.Object &&
-                    item.TryGetProperty(DotNetObjectRefKey.EncodedUtf8Bytes, out _) &&
-                     !typeof(IDotNetObjectRef).IsAssignableFrom(parameterType);
+                if (objectRefReader.Read() &&
+                    objectRefReader.TokenType == JsonTokenType.PropertyName &&
+                    objectRefReader.ValueTextEquals(DotNetObjectRefKey.EncodedUtf8Bytes))
+                {
+                    // The JSON payload looks has the expected shape.
```
ahsonkhan: nit: This comment is a bit confusing to read.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -172,74 +171,63 @@ private static object[] ParseArguments(string methodIdentifier, string argsJson,
...
-                // Always dispose the JsonDocument in case of an error.
-                jsonDocument?.Dispose();
-                throw;
+                suppliedArgs[i] = JsonSerializer.Deserialize(ref reader, parameterType, JsonSerializerOptionsProvider.Options);
             }
 
-            static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
+            return suppliedArgs;
+
+            static bool IsIncorrectDotNetObjectRefUse(Type parameterType, Span<byte> utf8JsonBytes, JsonReaderState readerState)
```
ahsonkhan: Why not pass the reader directly rather than passing in the data/state? If you don't want the mutations to the reader be retained, pass by value (rather than ref).

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -263,16 +251,24 @@ static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
...
+
+            reader.Read();
+            var taskId = reader.GetInt64();
 
-        private static void EndInvoke(long asyncHandle, bool succeeded, JSAsyncCallResult result)
-            => ((JSRuntimeBase)JSRuntime.Current).EndInvokeJS(asyncHandle, succeeded, result);
+            reader.Read();
+            var success = reader.GetBoolean();
+
+            jsRuntimeBase.EndInvokeJS(taskId, success, ref reader);
```
ahsonkhan: Maybe read once more before passing the refer to `EndInvokeJS`? Looks like both branches within `EndInvokeJS` do a `.Read()` first anyway which we can remove.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -139,40 +138,49 @@ private async Task<T> InvokeWithDefaultCancellation<T>(string identifier, IEnume
...
-                    var exceptionText = asyncCallResult?.JsonElement.ToString() ?? string.Empty;
+                    SetJSException(exception);
+                }
+            }
+            else
+            {
+                try
+                {
+                    jsonReader.Read();
+                    var exceptionText = jsonReader.TokenType == JsonTokenType.Null ? string.Empty : jsonReader.GetString();
```
ahsonkhan: `jsonReader.GetString()` no longer throws for `TokenType ==  JsonTokenType.Null` but rather returns a null string. 

Just in case that matters here. Maybe passing in null as the `exceptionText` to `JSException` is OK (rather than string.Empty).

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
ahsonkhan: Is cross-compiling/multi-targeting feasible? If perf/allocation is super critical, maybe use the char* overload for Encoding.UTF8 on older platforms and the span one on newer ones.

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
ahsonkhan: If you pass in null to Exception, `Exception.Message` is `Exception of type 'System.Exception' was thrown.`

If you pass in string.Empty, `Exception.Message` is empty.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
SteveSandersonMS: @pranavkm We can stackalloc on netstandard2.0, it just has to be done inside a method marked as unsafe. I think we'd be OK with the "unsafe" for now (this will move to netstandard2.1 later). However I'm not sure if additional checks are needed to avoid problems if we receive an especially long string. Brennan's other suggestion may be better!

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
SteveSandersonMS: I agree, this does seem a bit odd

SteveSandersonMS: Also maybe I'm misunderstanding the APIs, but does `reader.Skip` skip over complex objects or does it just skip over individual tokens? How many tokens does it think are in the input `["a", ["b1", "b2"]]`, or is that sort of thing not possible here?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
SteveSandersonMS: Also it might be nice to make this *not* be a local method to make it super clear that we're not expecting to capture any locals in a closure, since that would be a bad perf situation.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -263,16 +251,24 @@ static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
         /// </exception>
         public static void EndInvoke(string arguments)
         {
-            var parsedArgs = ParseArguments(
```
SteveSandersonMS: I think the logic here could definitely use some extra comments now. Something to depict what shape of data we're expecting to receive here, like:

```
// arguments must be a valid JSON string of the form:
//     [taskId, success, result]
// where 'taskId' is a number (parsed as long), 'success' is a boolean, and 'result' is passed through to jsRuntimeBase.EndInvokeJS to be read and parsed
```

src/JSInterop/Microsoft.JSInterop/src/JSRuntimeBase.cs
```diff
@@ -139,40 +138,49 @@ private async Task<T> InvokeWithDefaultCancellation<T>(string identifier, IEnume
...
                     TaskGenericsUtil.SetTaskCompletionSourceException(tcs, new JSException(exceptionText));
                 }
+                catch (Exception exception)
+                {
+                    SetJSException(exception);
```
SteveSandersonMS: Any reason to have separate try/catch blocks for the two success/fail branches, or could we have a single try/catch wrapping them both?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
ahsonkhan: `reader.Skip` skips over complex objects (not individual tokens).

So, if your reader was positioned at StartArray (or Object), it would skip to EndArray (or Object). If your reader was positioned at String, or other primitive token (like Null, True, False, Number), it wouldn't do anything.

pranavkm:
Some numbers:

|  |  Before  | After   |
|---|---|---|
| `string` | image  | image  |   
| `byte[]`| image   | image  |   
| `object[]`| image | N/A |

The `byte[]` allocation is a wash at this point and using `stackalloc` saves about 100 bytes \ JS operation (100 kb over 2k operations). It doesn't seem particularly worthwhile considering considering you have to cross-compile. Particularly given that the plan is to transmit the args as a utf8 byte array (https://github.com/aspnet/AspNetCore/issues/12280), the added complexity doesn't seem worthwhile at this point.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -362,7 +370,20 @@ private static Assembly GetRequiredLoadedAssembly(AssemblyKey assemblyKey)
...
+
+            if (WorkAroundTestBug)
+            {
+                // Workaround for https://github.com/dotnet/arcade/issues/2816
+                assembly = loadedAssemblies.LastOrDefault(a => new AssemblyKey(a).Equals(assemblyKey));
+            }
+            else
+            {
+                assembly = loadedAssemblies.FirstOrDefault(a => new AssemblyKey(a).Equals(assemblyKey));
+
```
BrennanConroy:
```suggestion
```

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -165,81 +164,64 @@ private static object InvokeSynchronously(string assemblyName, string methodIden
...
-                jsonDocument?.Dispose();
-                throw;
+                // Either we received more parameters than we expected or the JSON is malformed.
+                throw new JsonException($"Unexpected JSON token {reader.TokenType}. Ensure that the call to `{methodIdentifier}' is supplied with exactly '{parameterTypes.Length}' parameters.");
             }
 
-            static bool IsIncorrectDotNetObjectRefUse(JsonElement item, Type parameterType)
+            return suppliedArgs;
+
+            // Note that the JsonReader instance is intentionally not passed by ref (or asn in parameter) since we want a copy of the original reader.
```
BrennanConroy:
```suggestion
            // Note that the JsonReader instance is intentionally not passed by ref (or an in parameter) since we want a copy of the original reader.
```

src/JSInterop/Microsoft.JSInterop/test/DotNetDispatcherTest.cs
```diff
@@ -390,6 +486,178 @@ public void CannotInvokeWithIncorrectNumberOfParams()
...
+
+            // Assert
+            Assert.Collection(
+                result,
+                v => Assert.Equal(4, v),
+                v => Assert.Null(v));
+        }
+
+        [Fact]
+        public void ParseArguments_UsesStackForSmallJsonPayloads()
```
BrennanConroy: I don't understand this one

src/JSInterop/Microsoft.JSInterop/test/DotNetDispatcherTest.cs
```diff
@@ -390,6 +486,163 @@ public void CannotInvokeWithIncorrectNumberOfParams()
...
+            var task = jsRuntime.InvokeAsync<TestDTO>("somemethod");
+
+            DotNetDispatcher.ParseEndInvokeArguments(jsRuntime, $"[{jsRuntime.LastInvocationAsyncHandle}, true, {{\"intVal\": 7}}]");
+
+            Assert.True(task.IsCompletedSuccessfully);
+            Assert.Equal(7, task.Result.IntVal);
+        }
+
+        [Fact]
+        public void ParseEndInvokeArguments_UsesStackForSmallPayloads()
```
BrennanConroy: Same question, what does this mean?

Is it a remnant from wanting to test the stackalloc stuff?

pranavkm: Yup. Removed in the update

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -354,6 +360,8 @@ private static (MethodInfo, Type[]) GetCachedMethodInfo(AssemblyKey assemblyKey,
             return result;
         }
 
+        internal static bool WorkAroundTestBug;
```
SteveSandersonMS: I don't understand this. Am I missing something, or is its value always `true`? If so, why do we even have this flag and the `if`/`else` case below? Why not just always do L378?

SteveSandersonMS: Oh wait, I see how it only gets set to `true` in tests.

Still, is there any disadvantage in always using the logic on L378?

pranavkm: Not particularly. I didn't think it was necessary to change product code, but in the most ordinary cases, you're unlikely to see more than two instances of the same assembly so this doesn't change much.

pranavkm:
@SteveSandersonMS \ @javiercn would you like to have another look at this before I merge it in?

Issue #953, opened 2019-01-16
StringValues: Inconsistent hashcode for single-valued instances depending on construction


acolomba:
### Describe the bug

Two ```StringValues``` instances initialized in different ways with the same _single_ value may or may not produce the same hashcode depending on the actual value they hold. 

Or, given a fixed string value, they may produce the same or different hashcodes across different .NET runtime executions.

### To Reproduce

1. .NET Core 2.1.503, Microsoft.Extensions.Primitive 2.1.1

2. Run this unit test repeatedly.

```
var sv1 = new StringValues("value");
var sv2 = new StringValues(new [] { "value" });

Assert.Equal(sv1.GetHashCode(), sv2.GetHashCode());
```

The assertion will sporadically fail.

### Expected behavior

The assertion should succeed consistently.

```StringValues.Equals``` is always true for the many forms of ```sv1.Equals(sv2)```,  so right now we sometimes end up with equal instances with different hashcodes, breaking the contract between ```.Equals``` and ```.GetHashCode```.

### Root cause

For single-valued instances, ```StringValues.GetHashCode``` calculates the hashcode differently according to the internal representation:

* ```sv1.GetHashCode()``` returns the hashcode of the one string value stored in the ```_value``` internal field.
    * result: ```"value".GetHashCode()```

* ```sv2.GetHashCode()``` uses a ```HashCodeCombiner``` for the one value stored in the first element of the internal ```_values``` array field. The combiner accumulates hashcodes into a long, and then returns the hashcode of that long.
    * result: ```((long)"value".GetHashCode()).GetHashCode()```

Now, the hashcode of a long is:

```
(unchecked((int)((long)m_value)) ^ (int)(m_value >> 32))
```

This returns m_value if m_value is > 0 so the double ```.GetHashCode()``` above is ok when ```"value".GetHashCode()``` is in that range, and the two hashcodes will be the same. But a negative ```"value".GetHashCode()``` ends up producing a different value altogether.

In other words, two ```StringValues``` instances initialized in different ways with the same single value may or may not produce the same hashcode depending on whether the hashcode of the value they hold, by virtue of being negative, ends up setting the sign bit of a long.

Also, because hashcodes are seeded differently at every .NET runtime execution, across different runs ```StringValues``` hashcodes that were equal in one run won't be in another run.

### Suggested fix

This can be fixed in several ways, but I think the fix that meshes best with the style of this code would be to go from:

```
        public override int GetHashCode()
        {
            if (_values == null)
[...]

PR #2062, opened 2019-07-19
Correct GetHashCode behavior for StringValues

maryamariyan:

- Single value hashcode should return same value as
a Single item array with that same value

Addresses #953

cc: @anurse

### DIFF
--- a/src/Primitives/test/StringValuesTests.cs
+++ b/src/Primitives/test/StringValuesTests.cs
@@ -231,12 +231,16 @@ namespace Microsoft.Extensions.Primitives
         }
 
         [Fact]
-        public void GetHashCode_NullVsArrayWithNullItem_DifferentHashCodes()
+        public void GetHashCode_NullCases_DifferentHashCodes()
         {
             var sv1 = new StringValues((string)null);
             var sv2 = new StringValues(new[] { (string)null });
             Assert.NotEqual(sv1, sv2);
-            Assert.Equal(sv1.GetHashCode(), sv2.GetHashCode());
+            Assert.NotEqual(sv1.GetHashCode(), sv2.GetHashCode());
+
+            var sv3 = new StringValues((string[])null);
+            Assert.Equal(sv1, sv3);
+            Assert.Equal(sv1.GetHashCode(), sv3.GetHashCode());
         }
 
         [Fact]
--- a/src/Primitives/test/StringValuesTests.cs
+++ b/src/Primitives/test/StringValuesTests.cs
@@ -222,6 +222,24 @@ namespace Microsoft.Extensions.Primitives
         }
 
         [Fact]
+        public void GetHashCode_SingleValueVsArrayWithOneItem_SameHashCode()
+        {
+            var sv1 = new StringValues("value");
+            var sv2 = new StringValues(new[] { "value" });
+            Assert.True(sv1.Equals(sv2));
+            Assert.Equal(sv1.GetHashCode(), sv2.GetHashCode());
+        }
+
+        [Fact]
+        public void GetHashCode_SingleValueVsArrayWithTwoItems_DifferentHashCodes()
+        {
+            var sv1 = new StringValues("value");
+            var sv2 = new StringValues(new[] { "value", "value" });
+            Assert.False(sv1.Equals(sv2));
+            Assert.NotEqual(sv1.GetHashCode(), sv2.GetHashCode());
+        }
+
+        [Fact]
         public void ImplicitStringArrayConverter_Works()
         {
             string[] nullStringArray = null;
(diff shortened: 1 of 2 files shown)


maryamariyan:
This is not ready yet.. still working through the test cases.

src/Primitives/test/StringValuesTests.cs
```diff
@@ -221,6 +221,33 @@ public void ImplicitStringConverter_Works()
...
+            Assert.Equal(sv1.GetHashCode(), sv2.GetHashCode());
+        }
+
+        [Fact]
+        public void GetHashCode_NullVsArrayWithNullItem_DifferentHashCodes()
+        {
+            var sv1 = new StringValues((string)null);
+            var sv2 = new StringValues(new[] { (string)null });
+            Assert.NotEqual(sv1, sv2);
+            Assert.Equal(sv1.GetHashCode(), sv2.GetHashCode());
```
maryamariyan: Do we expect the two hash codes to be equal in this use case?

Tratcher: No? Maybe factor in the array length?

anurse: @Tratcher why not have them be equal? The idea here is to treat a single item array as the same as the single item itself.

anurse:
Whoops, @maryamariyan please wait for @Tratcher to approve before merging. Didn't mean to click Approve :)

src/Primitives/test/StringValuesTests.cs
Tratcher: `(string)null` and `(string[])null` are considered equal default null values, but `new[] { (string)null }` is not, you had to actually specify that value.

Tratcher: https://github.com/aspnet/Extensions/blob/c89d414a99d735ae8e6f8b5c46ea4669b38dd09d/src/Primitives/src/StringValues.cs#L436-L454

Tratcher: It doesn't matter if the hash codes are the same if Equals still fails.

maryamariyan: Do we want Equals and GetHashCode to have a coherent behavior? If so, then would Equals need to change to include GetHashCode in its calculation?

maryamariyan: I found the MemberData `FilledStringValuesWithExpected` which is used in a test called `Equals_StringArray`, showing different cases where two StringValues can be equal. I am guessing whatever the final implementation of GetHashCode will be has to make sure it returns the same hashcode for those instances that Equals returns true for:

```
       public static TheoryData<StringValues, string[]> FilledStringValuesWithExpected
        {
            get
            {
                return new TheoryData<StringValues, string[]>
                {
                    { default(StringValues), new string[0] },
                    { StringValues.Empty, new string[0] },
                    { new StringValues(string.Empty), new[] { string.Empty } },
                    { new StringValues("abc"), new[] { "abc" } },
                    { new StringValues(new[] { "abc" }), new[] { "abc" } },
                    { new StringValues(new[] { "abc", "bcd" }), new[] { "abc", "bcd" } },
                    { new StringValues(new[] { "abc", "bcd", "foo" }), new[] { "abc", "bcd", "foo" } },
                    { string.Empty, new[] { string.Empty } },
                    { "abc", new[] { "abc" } },
                    { new[] { "abc" }, new[] { "abc" } },
                    { new[] { "abc", "bcd" }, new[] { "abc", "bcd" } },
                    { new[] { "abc", "bcd", "foo" }, new[] { "abc", "bcd", "foo" } },
                    { new[] { null, "abc", "bcd", "foo" }, new[] { null, "abc", "bcd", "foo" } },
                    { new[] { "abc", null, "bcd", "foo" }, new[] { "abc", null, "bcd", "foo" } },
                    { new[] { "abc", "bcd", "foo", null }, new[] { "abc", "bcd", "foo", null } },
                    { new[] { string.Empty, "abc", "bcd", "foo" }, new[] { string.Empty, "abc", "bcd", "foo" } },
                    { new[] { "abc", string.Empty, "bcd", "foo" }, new[] { "abc", string.Empty, "bcd", "foo" } },
                    { new[] { "abc", "bcd", "foo", string.Empty }, new[] { "abc", "bcd", "foo", string.Empty } }
                };
            }
        }
```

Tratcher: https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=netframework-4.8#remarks
"Two objects that are equal return hash codes that are equal. However, the reverse is not true: equal hash codes do not imply object equality, because different (unequal) objects can have identical hash codes."

maryamariyan:
@Tratcher @anurse I will be reviving this PR today.

What is left to do is go over existing test cases where we consider two StringValues to be equal and to modify GetHashCode algorithm so that it can return the same value for them.

maryamariyan:
I updated an existing test in my latest commit that tests Equals method for string and string arrays also making sure if the two elements are equal then their hash codes are also equal.

@Tratcher please let me know if you want anything else to be checked as well or if there's anything I missed to address.

maryamariyan:
@anurse I don't have permissions to merge.

anurse:
Thanks @maryamariyan !

PR #2079, opened 2019-07-24
Catch all exceptions from Exit shutdown

jkotalik:
Same as https://github.com/aspnet/AspNetCore/pull/12518 except in master and with generic host.

TODO I think I can test this one with a bad logger. I'll take a stab at it.

### DIFF
--- a/src/Logging/Logging.Console/src/ConsoleLoggerProcessor.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLoggerProcessor.cs
@@ -41,7 +41,11 @@ namespace Microsoft.Extensions.Logging.Console
             }
 
             // Adding is completed so just log the message
-            WriteMessage(message);
+            try
+            {
+                WriteMessage(message);            
+            }
+            catch (Exception) { }
         }
 
         // for testing
@@ -49,24 +53,18 @@ namespace Microsoft.Extensions.Logging.Console
         {
             var console = message.LogAsError ? ErrorConsole : Console;
 
-            try
+            if (message.TimeStamp != null)
             {
-                if (message.TimeStamp != null)
-                {
-                    console.Write(message.TimeStamp, message.MessageColor, message.MessageColor);
-                }
-
-                if (message.LevelString != null)
-                {
-                    console.Write(message.LevelString, message.LevelBackground, message.LevelForeground);
-                }
-
-                console.Write(message.Message, message.MessageColor, message.MessageColor);
-                console.Flush();
+                console.Write(message.TimeStamp, message.MessageColor, message.MessageColor);
             }
-            catch (Exception)
+
+            if (message.LevelString != null)
             {
+                console.Write(message.LevelString, message.LevelBackground, message.LevelForeground);
             }
+
+            console.Write(message.Message, message.MessageColor, message.MessageColor);
+            console.Flush();
         }
 
         private void ProcessLogQueue()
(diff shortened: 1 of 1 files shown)


src/Hosting/Hosting/src/Internal/ConsoleLifetime.cs
```diff
@@ -78,11 +78,21 @@ private void OnApplicationStopping()
...
+            {
+                ApplicationLifetime.StopApplication();
+                if(!_shutdownBlock.WaitOne(HostOptions.ShutdownTimeout))
+                {
+                    Logger.LogInformation("Waiting for the host to be disposed. Ensure all 'IHost' instances are wrapped in 'using' blocks.");
+                }
+            }
+            // When hosting with IIS in-process, we detach the Console handle on main thread exit.
+            // Console.WriteLine may throw here as we are logging to console on ProcessExit.
+            // We catch and ignore all exceptions here. Do not log to Console in thie exception handler.
```
anurse: Is the Console Logger susceptible to this issue? It should be doing the actual console write from a background thread.

I'm actually not 100% convinced we should completely avoid logging in this situation since IMO loggers generally shouldn't throw (i.e. even when disposed they should just become black holes).

jkotalik: It looks like it is susceptible: https://github.com/aspnet/Extensions/blob/master/src/Logging/Logging.Console/src/ConsoleLoggerProcessor.cs#L44

jkotalik: Do you prefer me to fix the console logger or do this fix here?

anurse: Hrrrmmmmmhhhghgghhrhrrmrmrmrrmmmmm....

Console logger I think. Loggers shouldn't throw.

jkotalik: It does feel weird to add a try/catch around every log message in the console logger. Could it affect perf?

anurse: Just put it around the call to `WriteMessage`. Other call sites already `try..catch` it.

As to perf, everything is already pretty littered with `try..catch`es. If they affect perf, that damage is done ;)

jkotalik: Actually nvrm, I now realize my original fix wouldn't have helped. Console logs are written on a background thread, so the thread that would have crashed would be the one to crash the process.

jkotalik:
🆙 📅 . I could hack together a test where I replace the IConsole, but I would need to make _messageQueue internal and do a bit of other magic.

src/Logging/Logging.Console/src/ConsoleLoggerProcessor.cs
```diff
@@ -49,18 +49,24 @@ internal virtual void WriteMessage(LogMessageEntry message)
         {
             var console = message.LogAsError ? ErrorConsole : Console;
 
-            if (message.TimeStamp != null)
+            try
```
BrennanConroy: Not sure we need to wrap this, `ProcessLogQueue` already has a try catch and if the console throws we complete the blockingcollection since there was a problem.

And the only other call to `WriteMessage` is directly from the console logger. We might still want to avoid throwing directly from the console logger too, in that case we should move this try catch to `EnqueueMessage`.

PR #2081, opened 2019-07-24
Fix GetResourcePrefix() that causes an issue with using Rootnamespace

hishamco:
Addresses issues https://github.com/aspnet/AspNetCore/issues/10639


ryanbrandenburg:
@hishamco your changes look good (the failure is very unlikely to be your fault, should pass on re-run), but I'm a little nervous that we don't have a more "functional" test of this behavior. The tests we have (and that you've updated) mostly seem to confirm that the innards of the code as-is matches our expected output, rather than that the end-user gets the correct thing. I wrote a test [on this branch](https://github.com/aspnet/Extensions/tree/rybrande/Localization/RootNamespace) that tests the actual end results, if you could look it over and cherry-pick it into your branch I think we'd be good to merge this in.

hishamco:
> (the failure is very unlikely to be your fault, should pass on re-run)

Pardon, I'm still unable to run the source from VS 2019 :(

> I'm a little nervous that we don't have a more "functional" test of this behavior

There's a functional test in AspNetCore repo, but seems you forgot to change the namespace, that's why it doesn't fail, have a look to my another PR https://github.com/aspnet/AspNetCore/pull/12556, so:

1. Shall we copy that sample to this repo
2. Leave it with my changes in both repos
3.Take your branch into account

> if you could look it over and cherry-pick it into your branch I think we'd be good to merge this in.

I didn't use cherry-pick before, but I 'll try

hishamco:
BTW can we consider this as breaking change while it's behavioral?

ryanbrandenburg:
Good catch, but I'd like to have a functional test here as well since it doesn't require much work to enable it and gives us a quicker turn-around in the future if we break this again.

That's a good point about this being breaking, I'll have to bring it up with @mkArtakMSFT and see where we land.

hishamco:
I 'll add new functional test or use your ..

> That's a good point about this being breaking, I'll have to bring it up with @mkArtakMSFT and see where we land.

Announcement may needed after the merge

hishamco:
One more thing we should revise which tests are suitable here and there, coz you knew the was in a lovely once repo 😄

hishamco:
cherry-pick was very easy with GitKraken 😄

Let me do a final check before you merge

ryanbrandenburg:
To be clear, if this is considered a breaking change then merging the PR is likely to be delayed by the process of handling that.

src/Localization/Localization/test/Model.cs
```diff
@@ -0,0 +1,8 @@
+// This namespace for test resources with alternative RootNamespace
```
ryanbrandenburg: This file needs to be moved into the new folder I created to separate the tests which mess with RootNamespace from those which do not.

hishamco: What if:
- Create `test-assets`
- Convert the `LocalizationTest` to ClassLibrary
- Add the test as a part of `Microsoft.Extensions.Localization.Tests`

This quite similar to what we did before into AspNetCore repo

ryanbrandenburg: The test isn't of loading resources from a class-library, Something like what you suggest probably makes sense if we have more (and more complicated tests) for this kind of Localization thing in the future, but we only have the one, and it can be fixed by moving this file into `src/Localization/Localization/test/Microsoft.Extensions.Localization.Tests`.

hishamco: Ok, what I need to do now is moving the `model` class into `Microsoft.Extensions.Localization.Tests`. Am I right?

ryanbrandenburg: Yes

hishamco:
@ryanbrandenburg :shipit: after @mkArtakMSFT approved

hishamco:
@mkArtakMSFT any update on this?

pranavkm:
Thanks for the PR!

hishamco:
It's my pleasure to improve the localization as much as I can

hishamco:
@ryanbrandenburg I think we need to create announcement for this breaking change

/cc @mkArtakMSFT

jeremyVignelles:
Hello,
I faced the very same issue with my asp.net core 3.1 app.
Since it has been flagged as a `breaking change`, I assume that it hasn't landed in 3.1, so is there a known workaround?

hishamco:
> I assume that it hasn't landed in 3.1

I don't think so, please steps to reproduce

jeremyVignelles:
I managed to find a workaround:
- Copy https://github.com/aspnet/Extensions/blob/2303aa23ddf5a2ffff866911a55f4f88782a215b/src/Localization/Localization/src/ResourceManagerStringLocalizerFactory.cs into your project (I changed name and namespaces to avoid conflicts)
- Register the new StringLocalizerFactory into the service provider, after AddLocalization:

```cs
            services.AddLocalization(options => options.ResourcesPath = "Resources");
            services.AddSingleton<IStringLocalizerFactory, Net5ResourceManagerStringLocalizerFactory>();
```

My resource is now resolved as expected.

Sorry for the noise.

hishamco:
I didn't know what's the issue that are you facing?

jeremyVignelles:
I had a project where I wanted to use request localization with asp.net 3.1, but RootNamespace was different from AssemblyName, and it failed to resolve my resources, despite the [assembly: RootNamespace()] attribute.
After replacing the IStringLocalizerFactory implementation, it started working fine

hishamco:
Please file a new issue, so I can have a look if there's a real bug

jeremyVignelles:
@hishamco : Here is a repro project: https://github.com/jeremyVignelles/repro-localization
The relevant line is here : https://github.com/jeremyVignelles/repro-localization/blob/5f0fc80ac58a6321c78f4ce0e743635809f20889/TestLocalization/Program.cs#L14 . If I use the ResourceManagerStringLocalizerFactory from 3.1, it doesn't work as expected.

hishamco:
@pranavkm I looked to the merge date 8 Aug in both `release/3.0` and `release/3.1` but I didn't find the commit?!! Any idea?

jeremyVignelles:
ping? Will this PR be backported to 3.1 ?

hishamco:
I'm still waiting for @parnavkm reply

/cc @mkArtakMSFT

PR #2090, opened 2019-07-26
Multitarget extensions packages to remove dependencies for source build

JunTaoLuo:
Addresses https://github.com/aspnet/Extensions/issues/2056

cc @crummel @dseefeld


JunTaoLuo:
FYI not ready for review, FileProviders.Embedded has msbuild props/targets that need to be added for netcoreapp3.0

src/FileProviders/Embedded/src/Microsoft.Extensions.FileProviders.Embedded.csproj
```diff
@@ -3,7 +3,8 @@
   <PropertyGroup>
     <RootNamespace>Microsoft.Extensions.FileProviders</RootNamespace>
     <Description>File provider for files in embedded resources for Microsoft.Extensions.FileProviders.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
JunTaoLuo: This breaking pack somehow. Need to investigate.

JunTaoLuo:
@anurse who would be a good candidate to review this? The main change is to multitarget some of the packages built for source build with netcoreapp3.0 and remove dependencies. However there are a few areas that needed some changes:

1. The packaging of FilesProviders.Embedded had to be changed since the TFMs we target is conditional
2. For the projects where we are adding the netcoreapp3.0 tfm, I also made sure the tests test both the netcoreapp3.0 and netstandard 2.0 binaries. Luckily most tests already test on netcoreapp3.0 and net472 so the coverage is mostly sufficient. However, I guess we might also want to test on netcoreapp2.2 to ensure we are testing netstandard2.0 binaries in Core. Do you think that would be useful?
3. I had to update the test logic in Config.UserSecrets to accommodate more tfm for test coverage.

src/Configuration/Config.UserSecrets/test/MsBuildTargetTest.cs
```diff
@@ -41,14 +41,12 @@ public MsBuildTargetTest(ITestOutputHelper output)
         }
 
         [Theory]
-        [InlineData(".csproj", ".cs")]
-        [InlineData(".fsproj", ".fs")]
-        public void GeneratesAssemblyAttributeFile(string projectExt, string sourceExt)
+        [InlineData(".csproj", ".cs", "netcoreapp2.2")] // Testing with netcoreapp2.2 to verify netstandard 2.0 binaries
```
JunTaoLuo: Hmm actually this doesn't seem right.

JunTaoLuo:
cc @pranavkm @ajaybhargavb to take a look at FileProviders changes
cc @HaoK  for Config changes
cc @BrennanConroy for Logging changes
cc @Tratcher for Hosting changes
cc @rynowak for everything

src/Configuration/Config.UserSecrets/test/Microsoft.Extensions.Configuration.UserSecrets.Tests.csproj
```diff
@@ -11,9 +11,13 @@
...
     <Reference Include="Microsoft.Extensions.Configuration" />
 
+    <Reference Include="Newtonsoft.Json" Condition="'$(TargetFramework)' == 'net472'"/>
+  </ItemGroup>
+
+  <ItemGroup>
     <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute">
-      <_Parameter1>TargetFramework</_Parameter1>
-      <_Parameter2>$(TargetFramework)</_Parameter2>
+        <_Parameter1>TargetFramework</_Parameter1>
```
HaoK: nit: whitespace

src/Configuration/Config.UserSecrets/test/Microsoft.Extensions.Configuration.UserSecrets.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFrameworks>netcoreapp3.0;net472</TargetFrameworks>
```
HaoK: just curious why user secrets gets a net472?

JunTaoLuo: I added this to the tests so that it will exercise the netstandard2.0 binaries. I see this pattern elsewhere in our tests in this repo.

pranavkm: FYI @davidfowl \ @rynowak. Since we're cross-compiling these binaries, they're being tested on net4x to get coverage for the .NETStandard codepath. 

@JunTaoLuo does this need to do anything special to not run on xplat?

src/Caching/Abstractions/ref/Microsoft.Extensions.Caching.Abstractions.csproj
```diff
@@ -1,10 +1,14 @@
 <!-- This file is automatically generated. -->
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
pranavkm: Spoke to @JunTaoLuo offline. We need to make sure this TFM works when we're source building netcoreapp3.1. I'd personally prefer not to cross-compile 3 different TFMs just for source build.

src/Caching/Abstractions/ref/Microsoft.Extensions.Caching.Abstractions.csproj
```diff
@@ -1,10 +1,14 @@
...
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <Compile Include="Microsoft.Extensions.Caching.Abstractions.netstandard2.0.cs" />
     <Reference Include="Microsoft.Extensions.Primitives"  />
   </ItemGroup>
+<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
```
dougbu: Why add a netcoreapp3.0 target in projects that reference exactly the same packages without that TFM?
  - I could see this if all ref/ projects use this approach while all non-ref/ projects use `$(DotNetBuildFromSource)`. But, don't we also need to source build the ref/ projects _only_ for netcoreapp3.0? 
  - Put another way, when building for source build, what will narrow the list of target frameworks in ref/ projects to just netcoreapp3.0?
  - Note I didn't confirm all ref/ projects are unconditionally multitargeted

Can you or GenAPI reduce the duplication by removing the conditions and using e.g. `<Compile Include="Microsoft.Extensions.Caching.Abstractions.$(TargetFramework).cs" />`?
  - Overall condition of the single reference / compile list could just be ` '$(TargetFramework)' != '' `

Separately, might help to update the issues to explicitly say we're going with netcoreapp3.0 and not netstandard2.1

src/Configuration/Config.Json/src/Microsoft.Extensions.Configuration.Json.csproj
```diff
@@ -18,8 +19,11 @@
   <ItemGroup>
     <Reference Include="Microsoft.Extensions.Configuration" />
     <Reference Include="Microsoft.Extensions.Configuration.FileExtensions" />
-    <Reference Include="System.Text.Json" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <Reference Include="System.Threading.Tasks.Extensions" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <Reference Include="System.Text.Json" />
+    <Reference Include="System.Threading.Tasks.Extensions" />
```
dougbu: Finally, a few references that are newly conditional 😃

src/Configuration/Config.UserSecrets/test/Microsoft.Extensions.Configuration.UserSecrets.Tests.csproj
JunTaoLuo: No, we don't need any additional logic for that. It's already handled for us, which is why most of the tests in this repo targets both TFMs.

src/Caching/Abstractions/ref/Microsoft.Extensions.Caching.Abstractions.csproj
JunTaoLuo: To my understanding, the ref/ projects are not built in source-build: https://github.com/dotnet/source-build/pull/503. @crummel @dagood is that still accurate?

dagood: @JunTaoLuo That PR was about removing some source-build infra that was producing some  ref assembly packages in a way that didn't work very well. Now we have a new way of producing ref assembly packages that works better. But `ref/` projects are unrelated--they've always built in any repo that has the `ref/` `src/` pattern to my knowledge.

JunTaoLuo: Ah, I misunderstood the PR. Hmm I need to think about this more. I regenerated the reference assemblies without considering the side effects.

One way to do this in source build might be to rerun our GenerateReferenceAssemblies script in source build.

JunTaoLuo: Oh and w.r.t 
> Why add a netcoreapp3.0 target in projects that reference exactly the same packages without that TFM?

We need to multi-target projects that have multi-targeted dependencies, otherwise the SDK will complain:
```
C:\gh\Extensions\.dotnet\sdk\3.0.100-preview7-012821\Microsoft.Common.CurrentVersion.targets(1653,5): error : Project 'C:\gh\Extensions\src\Primitives\src\Microsoft.Extensions.Primitives.csproj' targets 'netcoreapp3.0'. It cannot be referenced by a project that targets '.NETStandard,Version=v2.0'. [C:\gh\Extensions\src\Caching\Abstractions\src\Microsoft.Extensions.Caching.Abstractions.csproj]
```

For example, the above error will show up if you multi-target M.E.Primitives with netstandard2.0 and netcoreapp3.0 but keep M.E.Caching.Abstractions only at netstandard2.0.

src/Caching/Abstractions/ref/Microsoft.Extensions.Caching.Abstractions.csproj
JunTaoLuo: I spoke with @crummel, turns out they are tracking it in https://github.com/dotnet/source-build/issues/759. Which is currently triaged for 3.1.

JunTaoLuo:
I don't see an appropriate label but this has been approved by shiproom.

PR #2093, opened 2019-07-27
Update branding to 5.0.0-alpha1

dougbu:


dougbu:
This is blocked until we branch for preview8

PR #2099, opened 2019-07-29
Move exception to setter for UsePollingFileWatcher

maryamariyan:
Getters on properties shouldn't throw exceptions.

 - Moves exception to setter for UsePollingFileWatcher

Addresses #561

cc: @anurse

### DIFF
--- a/src/FileProviders/Physical/src/PhysicalFileProvider.cs
+++ b/src/FileProviders/Physical/src/PhysicalFileProvider.cs
@@ -84,16 +84,19 @@ namespace Microsoft.Extensions.FileProviders
         /// </value>
         public bool UsePollingFileWatcher
         {
-            get => _usePollingFileWatcher ?? false;
+            get
+            {
+                if (_usePollingFileWatcher == null)
+                {
+                    ReadPollingEnvironmentVariables();
+                }
+                return _usePollingFileWatcher ?? false;;
+            }
             set
             {
                 if (_fileWatcher != null)
                 {
                     throw new InvalidOperationException($"Cannot modify {nameof(UsePollingFileWatcher)} once file watcher has been initialized.");
-                }
-                if (_usePollingFileWatcher == null)
-                {
-                    ReadPollingEnvironmentVariables();
                 }
                 _usePollingFileWatcher = value;
             }
--- a/src/FileProviders/Physical/src/PhysicalFileProvider.cs
+++ b/src/FileProviders/Physical/src/PhysicalFileProvider.cs
@@ -84,21 +84,19 @@ namespace Microsoft.Extensions.FileProviders
         /// </value>
         public bool UsePollingFileWatcher
         {
-            get
+            get => _usePollingFileWatcher ?? false;
+            set
             {
                 if (_fileWatcher != null)
                 {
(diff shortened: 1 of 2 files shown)


src/FileProviders/Physical/src/PhysicalFileProvider.cs
```diff
@@ -84,21 +84,19 @@ public PhysicalFileProvider(string root, ExclusionFilters filters)
...
+            set
             {
                 if (_fileWatcher != null)
                 {
                     throw new InvalidOperationException($"Cannot modify {nameof(UsePollingFileWatcher)} once file watcher has been initialized.");
                 }
-
                 if (_usePollingFileWatcher == null)
                 {
                     ReadPollingEnvironmentVariables();
```
anurse: I think we need to keep this `if` clause in the `get` accessor. It's just the exception that needs to move.

src/FileProviders/Physical/test/PhysicalFileProviderTests.cs
```diff
@@ -1378,6 +1378,63 @@ public async Task WildCardToken_RaisesEventsWhenFileSystemWatcherDoesNotFire()
...
+                            // Act / Assert
+                            Assert.Throws<InvalidOperationException>(() => { provider.UsePollingFileWatcher = true; });
+                        }
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public void UsePollingFileWatcher_FileWatcherNotNull_GetterEfzUp()
```
BrennanConroy: `EfzUp` ??

maryamariyan: Please ignore. I thought I removed that.

src/FileProviders/Physical/src/PhysicalFileProvider.cs
```diff
@@ -86,19 +86,20 @@ public bool UsePollingFileWatcher
         {
             get
             {
-                if (_fileWatcher != null)
-                {
-                    throw new InvalidOperationException($"Cannot modify {nameof(UsePollingFileWatcher)} once file watcher has been initialized.");
-                }
-
                 if (_usePollingFileWatcher == null)
```
BrennanConroy: I think we also want to keep:
```csharp
if (_fileWatcher != null)
{
    return false;
}
```

src/FileProviders/Physical/src/PhysicalFileProvider.cs
```diff
@@ -86,19 +86,20 @@ public bool UsePollingFileWatcher
...
                 {
                     ReadPollingEnvironmentVariables();
                 }
-
-                return _usePollingFileWatcher.Value;
+                return _usePollingFileWatcher ?? false;;
```
BrennanConroy:
```suggestion
                return _usePollingFileWatcher ?? false;
```

src/FileProviders/Physical/src/PhysicalFileProvider.cs
BrennanConroy: I think the test you removed was good and will verify this check is here

maryamariyan:
@anurse is this PR ok to merge? I don't have permission to merge myself.

Issue #2023, opened 2019-07-12
Make `AddHttpClient()` register the default client in DI as `HttpClient`


rynowak:
[This method](https://github.com/aspnet/Extensions/blob/11cf90103841c35cbefe9afb8e5bf9fee696dd17/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs#L22) is a bit of a trap. It registers the `IHttpClientFactory` service, but doesn't do much else that's useful.

Now that Blazor registers an HttpClient by default (in client-side), it seems obvious that we should do that by default. 

We basically would add the following by default:
```
services.AddTransient<HttpClient>(s =>
{
    return s.GetRequiredService<IHttpClientFactory>().CreateClient(string.Empty));
});
```

This could be overwritten by anything the user wants to register if it comes after `AddHttpClient()`.

PR #2123, opened 2019-07-31
Making 'AddHttpClient()' register default client in DI as 'HttpClient'

cjaliaga:
AddHttpClient() now registers a default client in DI as HttpClient by default.
Added test.

Addresses #2023

### DIFF
--- a/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
+++ b/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
@@ -52,6 +52,12 @@ namespace Microsoft.Extensions.DependencyInjection
             // because we access it by reaching into the service collection.
             services.TryAddSingleton(new HttpClientMappingRegistry());
 
+            //Register default client as HttpClient
+            services.AddTransient(s =>
+            {
+                return s.GetRequiredService<IHttpClientFactory>().CreateClient(string.Empty);
+            });
+
             return services;
         }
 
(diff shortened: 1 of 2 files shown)


rynowak:
Great! Thanks.

Note that this will show up in 5.0 (full release in 2020).

Pilchie:
Awesome thanks! This tripped me up recently.

cjaliaga:
Always happy to help :smile: Me too, I was like, where is my HttpClient if I just added them :confused:

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -52,6 +52,12 @@ public static IServiceCollection AddHttpClient(this IServiceCollection services)
             // because we access it by reaching into the service collection.
             services.TryAddSingleton(new HttpClientMappingRegistry());
 
+            //Register default client as HttpClient
+            services.AddTransient(s =>
```
martincostello: Just spotted this PR and I wonder, should it use `TryAddTransient()` instead?

Existing apps might already manually configure this in pre-5.0.0 (I've definitely done it in places), and then when upgraded to 5.0.0 forget to remove it and then have two `HttpClient` dependencies registered instead of one?

rynowak: I'd be OK with that change, it's definitely an improvment. Want to send a PR?

martincostello: Sure, I’ll fire one up in a bit.

martincostello: #2131

PR #2131, opened 2019-08-04
Do not add extra HttpClient registration

martincostello:
Tweak behaviour added in #2123 so that `HttpClient` is not registered against the service collection again if a registration already exists.


rynowak:
Thanks!

PR #2137, opened 2019-08-05
Update DI and Hosting to use Microsoft.Bcl.AsyncInterfaces

PinpointTownes:
Fixes https://github.com/aspnet/Extensions/issues/1467.

/cc @Tratcher @anurse (I believe you own the Hosting stack)
/cc @davidfowl @Eilon (not sure who's responsible of DI these days 😄)


davidfowl:
This is a good change but we don’t want any unnecessary package refs by default on netcoreapp3.0 which means we’ll need to multi target in more places (like hosting).  @JunTaoLuo was working on that.

PinpointTownes:
@davidfowl both the hosting abstractions and implementations packages already target `netstandard20` and `netcoreapp30`, with the BCL async package excluded for `netcoreapp30`.

AFAICT, most of the packages here reference `Microsoft.Extensions.DependencyInjection.Abstractions`, which doesn't need the BCL package (only `Microsoft.Extensions.DependencyInjection` does).

The only stack that uses `Microsoft.Extensions.DependencyInjection` is logging (both `Microsoft.Extensions.Logging` and `Microsoft.Extensions.Logging.Testing` reference Microsoft.Extensions.DependencyInjection`). These 2 packages could be easily updated to include a `netstandard21` TFM.

Would you like me to do that?

Tratcher:
@JunTaoLuo is already adding multi-targeting in https://github.com/aspnet/Extensions/pull/2090/files#diff-e607480dfcb295da19cb5dc533726055.

PinpointTownes:
Perfect, I'll leave this PR as-is, then. Let me know if there's anything I help with.

anurse:
@Tratcher @davidfowl to be clear, this hasn't yet been submitted for 3.0 approval. Please make sure to check in with me before merging anything. I do have concerns about the churn this will create in 3.0 and the very short runway we have to make sure we get this right.

PinpointTownes:
@anurse it's your call, but I'd be interested in reading more about your concerns.

To be clear, this PR doesn't add new code, it simply realigns the .NET Standard source with the .NET Core one by removing conditional compilation constants that previously excluded async disposal support from .NET Standard and using the BCL async package. It's exactly what you guys did in ASP.NET Core with the connections abstractions and the SignalR .NET client (which is why I assumed it was an oversight).

It's also worth mentioning this area was already covered by many tests, that will now ensure async disposal works just fine on .NET Framework.

anurse:
> @anurse it's your call

It's not actually my call, but it is indeed our call ;).

> It's exactly what you guys did in ASP.NET Core with the connections abstractions and the SignalR .NET client (which is why I assumed it was an oversight).

We implemented `IAsyncDisposable` there but it's not quite the same as these changes which are actively using those interfaces in DI.

In general, I agree that this change makes sense and was probably just an oversight. We just need to work through the risk. For example, an issue just arose from this discussion (though it's not directly relevant to this PR and exists in the `netcoreapp3.0` version): https://github.com/aspnet/AspNetCore/issues/12918

PinpointTownes:
Roger.

> For example, an issue just arose from this discussion (though it's not directly relevant to this PR and exists in the netcoreapp3.0 version): aspnet/AspNetCore#12918

Nice. Glad it helped find a bug there 😄

davidfowl:
See https://github.com/aspnet/Extensions/pull/2090. This is in the same bucket and is less risky than the linked PR

PinpointTownes:
Looks like there are merge conflicts. Do you want me to rebase this PR?

PinpointTownes:
@davidfowl @anurse did you guys eventually come to an agreement? 😄

Pilchie:
@PinpointTownes they both happen to be out right now, but I'll get back to you by Monday at the latest. Sorry for the delay.

PinpointTownes:
Thanks @Pilchie!

No hurry, I just wanted to make sure this was still considered for the final 3.0 release, since it enables an important scenario (i.e async disposables on .NET Standard 2.0-only platforms like .NET Framework).

PinpointTownes:
@JunTaoLuo in the same vein: https://github.com/aspnet/AspNetCore/issues/13343 (I'm pretty sure the Extensions repo also has problematic calls)

PinpointTownes:
Rebased and resolved the conflicts in the props files 😄

PinpointTownes:
Hum, the CI failure looks like a flaky test 🤕

JunTaoLuo:
@PinpointTownes Can you make the changes I mentioned in the previous comment? We need to additionally cross-target netstandard2.1 for M.E.Hosting, M.E.DependencyInjection and M.E.Hosting.WindowsServices. The last one is needed to ensure the dependency isn't referenced in an app that references it running on core.

PinpointTownes:
@JunTaoLuo I'm OOO today but feel free to directly push that to my fork so it's added to this PR.

Otherwise, I'll take a look tonight 😄

Pilchie:
Note: Do to the expanding scope of this and lack of runway, we decided NOT to take this for 3.0, but we will plan to take it for 3.1 instead.

I believe @anurse had a couple of concerns with places we might have other bugs around async disposal too.

PinpointTownes:
That's extremely unfortunate. Are you at least going to update DI to dispose `IAsyncDisposable` services in a blocking way on .NET Standard 2.0?

Pilchie:
> Are you at least going to update DI to dispose IAsyncDisposable services in a blocking way on .NET Standard 2.0?

Not for 3.0 at this point.

JunTaoLuo:
Thanks @PinpointTownes! Unfortunately, this will only be available starting in 3.1

PinpointTownes:
Thanks!

PR #2157, opened 2019-08-07
Fix build break by using API that's available across net472 and netcoreapp3.0

JunTaoLuo:
Fixing break by #2090. I'm not sure why the original PR built cleanly though.


JunTaoLuo:
I'm going to merge immediately to fix the build. Will investigate why this only showed up in the official build instead of the PR check separately.

PR #2163, opened 2019-08-09
Change JSInterop to avoid using async locals

pranavkm:


src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -23,15 +24,32 @@ public static class DotNetDispatcher
         private static readonly ConcurrentDictionary<AssemblyKey, IReadOnlyDictionary<string, (MethodInfo, Type[])>> _cachedMethodsByAssembly
             = new ConcurrentDictionary<AssemblyKey, IReadOnlyDictionary<string, (MethodInfo, Type[])>>();
 
+        private static readonly AsyncLocal<JSRuntimeBase> _currentJSRuntime = new AsyncLocal<JSRuntimeBase>();
```
pranavkm: Wanted to see what it would look like to remove JSRuntime.Current. We only have one piece of code that currently references it (`ReleaseDotNetObject`) which we want to avoid using interop for, so that should clean this up well.

WebAssemblyJSRuntime ends up using a static for it's statically exposed APIs, but that seems reasonable .

Blazor's E2E tests pass, so I don't think I've missed anything. It does involve an API change, perhaps it might be something we want to consider right away rather than sit on it. Thoughts?


-------------------------------

On a related note, I also had a look at trying to removing references to JSInterop & System.Text.Json from Microsoft.AspNetCore.Components. The former is doable today, but the latter is blocked by https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/ElementReference.cs#L15. With the new `JsonSerializerOptions` property we have the ability to have the serialization behavior definied by the runtime, but we'd have to make the ctor and the `Id` property getter public for the server project to be able to read \ write it correctly.

rynowak: How does this halp? This is still an async local 😬

pranavkm: We noted this in the API Review, but the plan is to make ReleaseDotNetObject explicit like the other JSRuntime operations \ event handling changes rather than rely on JSInterop for it. I just didn't want to bother with that change as part of this investigation. Once we address that, we would no longer require this.

pranavkm:
This is good to go. Here's the follow PR in AspNetCore: https://github.com/aspnet/AspNetCore/pull/13154 that I was able to build & verify locally

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -66,7 +66,7 @@ module DotNet {
     }
   }
 
-  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, ...args: any[]): Promise<T> {
+  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, args: any[] | null): Promise<T> {
```
pranavkm: I made this change as part of https://github.com/aspnet/Extensions/commit/d6bfc28e2104066dc363bf79bed39d6580977b26, but it breaks interop. Cannot use an args array as a parameter that accepts another args array

src/JSInterop/Mono.WebAssembly.Interop/src/MonoWebAssemblyJSRuntime.cs
```diff
@@ -16,6 +16,25 @@ namespace Mono.WebAssembly.Interop
...
+        /// <summary>
+        /// Gets the <see cref="MonoWebAssemblyJSRuntime"/> used to perform operations using <see cref="DotNetDispatcher"/>.
+        /// </summary>
+        private static MonoWebAssemblyJSRuntime Instance { get; set; }
+
+        /// <summary>
+        /// Initializes the <see cref="MonoWebAssemblyJSRuntime"/> to be used to perform operations using <see cref="DotNetDispatcher"/>.
+        /// </summary>
+        /// <param name="jsRuntime">The <see cref="MonoWebAssemblyJSRuntime"/> instance.</param>
+        protected static void Initialize(MonoWebAssemblyJSRuntime jsRuntime)
```
pranavkm: We want this extensibility to configure the `ElementRefJsonConverter` here: https://github.com/aspnet/AspNetCore/pull/13154/files#diff-d74d8a63556f3a7f0b5be4814c52aeccR19, but it has to be static. I'm open to suggestions. We could also revisit this post 3.0 since this is client-side blazor and we have some additional time to figure this out.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectReferenceOfT.cs
```diff
@@ -50,10 +47,36 @@ internal long ObjectId
...
+            {
+                _jsRuntime = jsRuntime;
+                _objectId = jsRuntime.TrackObjectReference(this);
+            }
+            else if (!ReferenceEquals(_jsRuntime, jsRuntime))
+            {
+                throw new InvalidOperationException($"{GetType().Name} is already being tracked by a different instance of {nameof(JSRuntime)}. A common cause is caching an instance of {nameof(DotNetObjectReference<TValue>)}" +
+                    $" globally. Consider creating instances of {nameof(DotNetObjectReference<TValue>)} at the JSInterop callsite instead");
+            }
+
```
rynowak: It's probably worth a comment here about why it's safe for a value to be tracked multiple times by the same runtime.

src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
```diff
@@ -16,35 +16,39 @@ namespace Microsoft.JSInterop
...
+        protected internal JsonSerializerOptions JsonSerializerOptions { get; }
 
         /// <summary>
-        /// Sets the current JS runtime to the supplied instance.
-        ///
-        /// This is intended for framework use. Developers should not normally need to call this method.
+        /// Gets or sets the default timeout for asynchronous JavaScript calls.
         /// </summary>
-        /// <param name="instance">The new current <see cref="IJSRuntime"/>.</param>
-        public static void SetCurrentJSRuntime(IJSRuntime instance)
```
rynowak: 👏

src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
```diff
@@ -191,5 +195,29 @@ internal void EndInvokeJS(long taskId, bool succeeded, ref Utf8JsonReader jsonRe
                 TaskGenericsUtil.SetTaskCompletionSourceException(tcs, new JSException(message, exception));
             }
         }
+
+        #region DotNetObjectReference
```
rynowak: meh. We don't typically use regions.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectReferenceOfT.cs
```diff
@@ -50,10 +47,36 @@ internal long ObjectId
             get
             {
                 ThrowIfDisposed();
+                Debug.Assert(_objectId != 0, "Accessing ObjectId without tracking is always incorrect.");
+
                 return _objectId;
             }
         }
 
+        internal long TrackObjectReference(JSRuntime jsRuntime)
```
rynowak: Does it make sense to change this pattern? This method calls `jsRuntime.TrackObjectReference(this)`. It's a little odd to have the visitor pattern in use without polymorphism. 

So what I had in mind would be something like making the JSRuntime+ObjectId internal set properties, and then moving all of this logic to the JSRuntime class.

src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
rynowak: After getting to the tests, I see why you put this in a region. I left a suggestion elsewhere about this.

src/JSInterop/Microsoft.JSInterop/test/Infrastructure/DotNetDispatcherTest.cs
```diff
@@ -162,166 +167,177 @@ public void CannotInvokeUnsuitableMethods(string methodIdentifier)
...
+            DotNetDispatcher.EndInvokeJS(jsRuntime, argsJson);
 
             // Assert
             var ex = await Assert.ThrowsAsync<JSException>(async () => await task);
             Assert.Empty(ex.Message);
-        });
+        }
 
         [Fact]
-        public Task CanInvokeInstanceMethodWithParams() => WithJSRuntime(jsRuntime =>
```
rynowak: 👏

src/JSInterop/Mono.WebAssembly.Interop/src/MonoWebAssemblyJSRuntime.cs
rynowak: Why does it need to be static? 

So the problem here is you want this class to be subclassable, but it also has a static singleton property. The only winning move is not to play.

If you don't have a static property here, then this becomes an instance method, and it's wicked simple.

src/JSInterop/Mono.WebAssembly.Interop/src/MonoWebAssemblyJSRuntime.cs
```diff
@@ -34,11 +53,11 @@ protected override void BeginInvokeJS(long asyncHandle, string identifier, strin
 
         // Invoked via Mono's JS interop mechanism (invoke_method)
         private static string InvokeDotNet(string assemblyName, string methodIdentifier, string dotNetObjectId, string argsJson)
-            => DotNetDispatcher.Invoke(assemblyName, methodIdentifier, dotNetObjectId == null ? default : long.Parse(dotNetObjectId), argsJson);
+            => DotNetDispatcher.Invoke(Instance, assemblyName, methodIdentifier, dotNetObjectId == null ? default : long.Parse(dotNetObjectId), argsJson);
```
rynowak: Ah yeah. That's *nasty*.

src/JSInterop/Microsoft.JSInterop/src/JSRuntime.cs
```diff
@@ -191,5 +196,48 @@ internal void EndInvokeJS(long taskId, bool succeeded, ref Utf8JsonReader jsonRe
                 TaskGenericsUtil.SetTaskCompletionSourceException(tcs, new JSException(message, exception));
             }
         }
+
+        internal long TrackObjectReference<TValue>(DotNetObjectReference<TValue> dotNetObjectReference) where TValue : class
```
pranavkm: Updated to do this.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectReferenceOfT.cs
```diff
@@ -50,8 +46,30 @@ internal long ObjectId
             get
             {
                 ThrowIfDisposed();
+                Debug.Assert(_objectId != 0, "Accessing ObjectId without tracking is always incorrect.");
```
rynowak: will this cause issues in the debugger?

pranavkm: Hmm, let me check.

pranavkm: Seems to ignore it.

PR #2165, opened 2019-08-09
Update documentation for IHttpClientFactory

ps-tb:
 - IHttpClientFactory XML docs don't match documentation suggestions
 - Change incorporates suggestions in documentation


ps-tb:
This is specifically referencing this section of the docs: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-2.2#httpclient-and-lifetime-management

rynowak:
Thanks!

PR #2168, opened 2019-08-11
Fix unintended breaking change

dougbu:
- blocking aspnet/AspNetCore#1293 dependency update PR
- commit 3ec8c35e450c lost `AssemblyName` substitution
  - tasks assembly in package incorrectly named ".Manifest.Task.dll"


dougbu:
Intend to get this in as soon as possible to fix the package and unblock dependency flow. If you're around, have a look @JunTaoLuo

PR #2169, opened 2019-08-11
Remove internal new() constraints from options infrastructure

pakrym:
### Motivation:

Currently it's very hard to use Options with option types that do not have a default constructor, doing so requires copying and modifying both OptionManger and OptionFactory source which is quite painful. I think this limitation is artificial side effect of the implementation and could be relaxed.

Examples of usages that would benefit from this adjustment:
https://github.com/microsoft/ApplicationInsights-aspnetcore/blob/04b5485d4a8aa498b2d99c60bdf8ca59bc9103fc/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs (note how it's not even a correct implementation, it misses IPostConfigureOptions and IValidateOptions calls)

https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Microsoft.Extensions.Azure/src/Internal/ClientOptionsFactory.cs

This PR removes internal new() constraints and adds a virtual `OptionsFactory<T>.CreateInstance` method that allows to inheriting and overriding the instance creation logic.

### Breaking change:

This is not a hard breaking change because constraints are being relaxed. Using IOptions<T> where T is NOT `new()`  without providing custom factory would still result in exception but it would move from DI service tree building time (throws when closing generic) to `CreateInstance` method call inside the `OptionsFactory`.

### Performance considerations:

Compiler generates the same code for Activator.CreateInstance<T>() and new(): https://sharplab.io/#v2:EYLgHgbALANALiAlgGwD4AEBMBGAsAKHQGYACLEgYRIG8CT6zT0oSBZAHgBUA+ACgEoadBiICCAYziIAbgEM4AewBOAOgpKApvI0BJAHYBnOLL3iNXPvwDcw+gF9bJR8TItW2CwJIB3ABYbNEk4QEj0NbwFHWnwRETDvIIEbGIYHfDsgA===

### 3.0

I know it's kind of late but can this go into 3.0?

### DIFF
--- a/src/Options/Options/ref/Microsoft.Extensions.Options.netstandard2.0.cs
+++ b/src/Options/Options/ref/Microsoft.Extensions.Options.netstandard2.0.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Extensions.Options
         string Name { get; }
         Microsoft.Extensions.Primitives.IChangeToken GetChangeToken();
     }
-    public partial interface IOptionsFactory<TOptions> where TOptions : class, new()
+    public partial interface IOptionsFactory<TOptions> where TOptions : class
     {
         TOptions Create(string name);
     }
@@ -120,11 +120,11 @@ namespace Microsoft.Extensions.Options
         TOptions Get(string name);
         System.IDisposable OnChange(System.Action<TOptions, string> listener);
     }
-    public partial interface IOptionsSnapshot<out TOptions> : Microsoft.Extensions.Options.IOptions<TOptions> where TOptions : class, new()
+    public partial interface IOptionsSnapshot<out TOptions> : Microsoft.Extensions.Options.IOptions<TOptions> where TOptions : class
     {
         TOptions Get(string name);
     }
-    public partial interface IOptions<out TOptions> where TOptions : class, new()
+    public partial interface IOptions<out TOptions> where TOptions : class
     {
         TOptions Value { get; }
     }
@@ -139,7 +139,7 @@ namespace Microsoft.Extensions.Options
     public static partial class Options
     {
         public static readonly string DefaultName;
-        public static Microsoft.Extensions.Options.IOptions<TOptions> Create<TOptions>(TOptions options) where TOptions : class, new() { throw null; }
+        public static Microsoft.Extensions.Options.IOptions<TOptions> Create<TOptions>(TOptions options) where TOptions : class { throw null; }
     }
     public partial class OptionsBuilder<TOptions> where TOptions : class
     {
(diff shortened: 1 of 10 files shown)


HaoK:
Looks reasonable, IOptionsFactory was added in 2.0 and a virtual CreateInstance method is a big improvment over the `new()` constraint

anurse:
Adding an API to 3.0 is going to require some information about customer impact. @pakrym 	I noticed you *have* a workaround, am I correct in saying that in order to work around it you have to reimplement chunks of options? It looks like you basically had to re-implement the `IOptionsFactory` interface entirely (`ClientOptionsFactory`) and provide your own custom implementation of `IOptionsMonitor`.

pakrym:
> @pakrym I noticed you have a workaround, am I correct in saying that in order to work around it you have to reimplement chunks of options? It looks like you basically had to re-implement the IOptionsFactory interface entirely (ClientOptionsFactory) and provide your own custom implementation of IOptionsMonitor.

Yes, the workaround exists but it requires the reimplementation of a large chunk of Options and the result often doesn't work correctly.

HaoK:
What doesn't work correctly when you implement your custom IOptionsXyz?  I would imagine you have full control if completely take over the specific implementations of a specific options type?

HaoK:
Doesn't this seems like something safer to do for 5.0 at this point rather than trying to get it into 3.0?

HaoK:
I guess its the new() constraint that causes lots of issues, since you can't get around that easily

pakrym:
> I would imagine you have full control if completely take over the specific implementations of a specific options type?

Right, but it's very hard to get everything done correctly to support all features of Options like named options, change tokens, etc. etc.

I'm not trying to push this super hard, if you feel 5 would be safer I'm fine with that.

HaoK:
Yeah something like this seems like it would be safer to go in early in a major release since its hard to tell if this would result in any subtle impact, it doesn't seem like it would, but that seems like famous last words to me :)

anurse:
I do think it would be a challenge to get this in for 3.0. I can send up a flare and see what comes out of it though. Let's get this in to master now (well, once there's an approving review ;)) and we can always backport.

anurse:
Go ahead and merge to master when you're ready @HaoK, since @pakrym may not have permission anymore 😛.

HaoK:
Thanks @pakrym !

pakrym:
If you need help porting this back to 3.0, I'm here for you 😉

anurse:
I'm not worried about the mechanics of porting here, the code hasn't changed much :).

anurse:
This smells like a breaking change (since it's a change to an interface). Is it?

I guess the question is what code did you need to have in order to implement this. If you had to have a `new()` constraint on your implementation as well, then we're probably OK since leaving that constraint in your implementation is fine. This change would just allow you to remove it. If implementing the interface somehow allowed you to _infer_ the `new()` constraint such that you could just do `new T()` somewhere in your implementation, then this would break that and you wouldn't be able to `new T()` any more.

I think we're OK, but may need to take a quick look to confirm.

pakrym:
Good point, I don't think I know the case where the constraint is not forced to be repeated and is just inferred.

anurse:
It looks ok, from some sharplab spelunking. You do indeed need to have the constraint on your implementing class so removing it from the interface doesn't appear to cause a problem. It might be weird in DI if you register a type that expects to be able to call `new T()` but I think that scenario is not a huge concern.

PR #2170, opened 2019-08-12
LoggerFactory Configurable ScopeProvider

benaadams:
It currently near impossible to stay on the default ExecutionContext without switching off logging in its entirety due to the scopes stored in the AsyncLocal in `LoggerExternalScopeProvider`; even if Scopes are off.

The only other-ways around it are to implement your own `ILoggerFactory` and ditch `LoggerFactory` or only use loggers that don't implement `ISupportExternalScope`. Both approaches are unsatisfactory as they don't allow you reuse logging.

Allowing it to be set allows `NullExternalScopeProvider` to be set instead and something along the lines of:
```csharp
{
    services.AddSingleton(NullExternalScopeProvider.Instance);
    services.Replace(ServiceDescriptor.Singleton<ILoggerFactory, LoggerFactory>(
        (services) =>
        {
            var factory = new LoggerFactory();
            factory.ScopeProvider = NullExternalScopeProvider.Instance;
            return factory;
        }));
})
```

Ideally it would get the from DI so `services.AddSingleton(NullExternalScopeProvider.Instance);` would replace it.

/cc @anurse @davidfowl @sebastienros


TheXenocide:
Just thought I'd stop by and mention that I had a similar issue and have submitted PR #2434 that takes the DI route instead, in case you're interested. I would have had issues with this approach not utilizing other constructors via DI for the LoggerFactory (I'm also adding additional scope, as opposed to removing it, but the premise of replacing the IExternalScopeProvider remains the same).

PR #2176, opened 2019-08-13
Special case Disposing DotNetObjectReferences

pranavkm:
This removes a public JSInvokable method required for disposing DotNetObjectReferences

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
+++ b/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
@@ -66,7 +66,11 @@ module DotNet {
     }
   }
 
-  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, args: any[]): Promise<T> {
+  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, ...args: any[]): Promise<T> {
+    if (assemblyName && dotNetObjectId) {
+      throw new Error(`For instance method calls, assemblyName should be null. Received '${assemblyName}'.`) ;
+    }
+
     const asyncCallId = nextAsyncCallId++;
     const resultPromise = new Promise<T>((resolve, reject) => {
       pendingAsyncCalls[asyncCallId] = { resolve, reject };
@@ -269,10 +273,7 @@ module DotNet {
     }
 
     public dispose() {
-      const promise = invokeMethodAsync<any>(
-        'Microsoft.JSInterop',
-        'DotNetDispatcher.ReleaseDotNetObject',
-        this._id);
+      const promise = invokePossibleInstanceMethodAsync<any>(null, '__Dispose', this._id);
       promise.catch(error => console.error(error));
     }
 
--- a/src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netcoreapp3.0.cs
+++ b/src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netcoreapp3.0.cs
@@ -8,8 +8,6 @@ namespace Microsoft.JSInterop
         public static void BeginInvoke(string callId, string assemblyName, string methodIdentifier, long dotNetObjectId, string argsJson) { }
         public static void EndInvoke(string arguments) { }
         public static string Invoke(string assemblyName, string methodIdentifier, long dotNetObjectId, string argsJson) { throw null; }
-        [Microsoft.JSInterop.JSInvokableAttribute("DotNetDispatcher.ReleaseDotNetObject")]
-        public static void ReleaseDotNetObject(long dotNetObjectId) { }
     }
     public static partial class DotNetObjectRef
     {
@@ -18,7 +16,7 @@ namespace Microsoft.JSInterop
     public sealed partial class DotNetObjectRef<TValue> : System.IDisposable where TValue : class
     {
         internal DotNetObjectRef() { }
-        public TValue Value { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public TValue Value { get { throw null; } }
         public void Dispose() { }
     }
     public partial interface IJSInProcessRuntime : Microsoft.JSInterop.IJSRuntime
--- a/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
@@ -38,7 +38,7 @@ namespace Microsoft.JSInterop
             // the targeted method has [JSInvokable]. It is not itself subject to that restriction,
             // because there would be nobody to police that. This method *is* the police.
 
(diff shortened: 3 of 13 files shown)


src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -66,7 +66,11 @@ module DotNet {
     }
   }
 
-  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, args: any[]): Promise<T> {
+  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, ...args: any[]): Promise<T> {
+    if (assemblyName && dotNetObjectId) {
+      throw new Error(`For instance method calls, assemblyName should be null. Received '${assemblyName}'.`) ;
```
javiercn: Isn't the text a bit misleading? You don't know if they are trying to call an instance method or a static one, (based on the check) shouldn't the message reflect that?

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -127,30 +127,38 @@ public static void BeginInvoke(string callId, string assemblyName, string method
...
+                {
+                    // The client executed dotNetObjectReference.dispose(). Dispose the reference and exit.
+                    objectReference.Dispose();
+                    return default;
+                }
+
+                assemblyKey = new AssemblyKey(objectReference.Value.GetType().Assembly);
             }
 
+
```
javiercn: Remove extra blank line.

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -127,30 +127,38 @@ public static void BeginInvoke(string callId, string assemblyName, string method
...
-            }
-            else
-            {
-                assemblyKey = new AssemblyKey(assemblyName);
+                if (string.Equals("__Dispose", methodIdentifier, StringComparison.Ordinal))
+                {
+                    // The client executed dotNetObjectReference.dispose(). Dispose the reference and exit.
+                    objectReference.Dispose();
+                    return default;
+                }
```
javiercn: Could we do this though `[JSInvokableAttribute("__Dispose")]` instead?

javiercn: Nevermind, I think you can't. Could we make this string a constant in all places?

javiercn:
> This removes a public JSInvokable method required for disposing DotNetObjectReferences

How does it work in the WASM case?

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
pranavkm: We effectively have the same check on the server. You're expected to either have an assemblyName + methodIdentifier (so we discover a JSInvocable globally) or a methodIdentifier + dotNetObjectId (so we discover it on the type). 

This is where I copied the text from: https://github.com/aspnet/Extensions/pull/2176/files#diff-5bc9491fb100bce7a35de845877af2b1R141

pranavkm:
> How does it work in the WASM case?

It's effectively the same code path. WASM also goes through the dispatcher which is where the code change is in. Am I missing something?

javiercn:
> It's effectively the same code path. WASM also goes through the dispatcher which is where the code change is in. Am I missing something?

Nope, you are right, I didn't realize it was an instance method.

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
```diff
@@ -66,7 +66,11 @@ module DotNet {
     }
   }
 
-  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, args: any[]): Promise<T> {
+  function invokePossibleInstanceMethodAsync<T>(assemblyName: string | null, methodIdentifier: string, dotNetObjectId: number | null, ...args: any[]): Promise<T> {
+    if (assemblyName && dotNetObjectId) {
```
rynowak: Is this just hardening? Or was there a bug here?

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netcoreapp3.0.cs
```diff
@@ -18,7 +16,7 @@ public static partial class DotNetObjectRef
     public sealed partial class DotNetObjectRef<TValue> : System.IDisposable where TValue : class
     {
         internal DotNetObjectRef() { }
-        public TValue Value { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: 😲

src/JSInterop/Microsoft.JSInterop/src/DotNetDispatcher.cs
```diff
@@ -127,30 +127,38 @@ public static void BeginInvoke(string callId, string assemblyName, string method
...
 
+
             var (methodInfo, parameterTypes) = GetCachedMethodInfo(assemblyKey, methodIdentifier);
 
             var suppliedArgs = ParseArguments(methodIdentifier, argsJson, parameterTypes);
 
             try
             {
-                return methodInfo.Invoke(targetInstance, suppliedArgs);
+                return methodInfo.Invoke(objectReference?.Value, suppliedArgs);
```
rynowak: It's probably worth a comment here that the `objectReference?.Value` is expected to be null in the non-instance-method case. I had to read this a few times to convince myself it's right.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRef.cs
```diff
@@ -15,8 +15,7 @@ public static class DotNetObjectRef
         /// <returns>An instance of <see cref="DotNetObjectRef{TValue}" />.</returns>
         public static DotNetObjectRef<TValue> Create<TValue>(TValue value) where TValue : class
         {
-            var objectId = DotNetObjectRefManager.Current.TrackObject(value);
-            return new DotNetObjectRef<TValue>(objectId, value);
+            return new DotNetObjectRef<TValue>(DotNetObjectRefManager.Current, value);
```
rynowak: Is this going to fight with your other change?

src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts
pranavkm: It's just hardening. I was confused looking at the function name, because I assumed it only called `instance` methods.

src/JSInterop/Microsoft.JSInterop/src/DotNetObjectRef.cs
pranavkm: Yup. But it's easier to review these things piecemeal.

PR #2188, opened 2019-08-16
Simplify JSRuntime method signature

pranavkm:


src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netcoreapp3.0.cs
```diff
@@ -50,7 +50,7 @@ public abstract partial class JSRuntime : Microsoft.JSInterop.IJSRuntime
         protected System.TimeSpan? DefaultAsyncTimeout { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         protected internal System.Text.Json.JsonSerializerOptions JsonSerializerOptions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         protected abstract void BeginInvokeJS(long taskId, string identifier, string argsJson);
-        protected internal abstract void EndInvokeDotNet(string callId, bool success, object resultOrError, string assemblyName, string methodIdentifier, long dotNetObjectId);
+        protected internal abstract void EndInvokeDotNet(Microsoft.JSInterop.Infrastructure.JSCallInfo callInfo, in Microsoft.JSInterop.Infrastructure.DotNetInvocationResult dispatchResult);
```
pranavkm: The design in the API review had things like error info in the `CallInfo` item. But that felt like conflating inputs and outputs. This turned out to be slightly more reasonable design. We can produce a `JSCallInfo` at the onset and pass it thru the stack. It has everything we need (so far) to correctly identify a call, wheras the latter is the success \ failure discriminated union.

I haven't, as yet, done the work to correctly populate the "Error Kind". As of now, we catch the result of the entire invocation - it's a fair bit of work to get fine grained error kinds that identify if it's a parse versus invocation versus bad object id. I wanted to get the API design in first, we can always do the identification bit in 3.1 since it doesn't affect the shape of this API.

src/JSInterop/Microsoft.JSInterop/src/Infrastructure/JSCallInfo.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.JSInterop.Infrastructure
+{
+    /// <summary>
+    /// Information about a JSInterop call from JavaScript to .NET.
+    /// </summary>
+    public readonly struct JSCallInfo
+    {
+        public JSCallInfo(string methodIdentifier, string assemblyName, long? dotNetObjectId, long? callId)
```
pranavkm: I also tweaked this to use the correct representation for data  types. We were reading everything as a string today and then parsing them in to longs. Just feels wrong. It's fairly trivial to add support for `long?` to BlazorPack, so I changed these data types

pranavkm: While we're here, the order in which these parameters arrive is kinda odd:

https://github.com/aspnet/Extensions/blob/master/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts#L62

`callId, assemblyName, methodIdentifier, dotNetObjectId, argsJson`

`callId` doesn't apply to sync code and may be null, `assemblyName` and `objectId` are discriminated (only one of the two can be specified). `methodIdentifier` is the only value that must always be present. I changed the order in the C# since it feels correct, but left it alone in TypeScript because it seems like unnecessary work.

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
```diff
@@ -72,8 +72,31 @@ namespace Microsoft.JSInterop.Infrastructure
...
+    }
+    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+    public readonly partial struct DotNetInvocationInfo
+    {
+        private readonly object _dummy;
+        private readonly int _dummyPrimitive;
+        public DotNetInvocationInfo(string assemblyName, string methodIdentifier, long dotNetObjectId, string callId) { throw null; }
+        public string AssemblyName { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public string CallId { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public long DotNetObjectId { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: Wasn't this nullable before? Or are we using 0 to track that?

src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetInvocationInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+        public string AssemblyName { get; }
+
+        /// <summary>
+        /// Gets the identifier of the method to be invoked. This is the value specified in the <see cref="JSInvokableAttribute"/>.
+        /// </summary>
+        public string MethodIdentifier { get; }
+
+        /// <summary>
+        /// Gets the object identifier for instance method calls.
+        /// This value is <see langword="null"/> when <see cref="AssemblyName"/> is specified.
```
rynowak: This doc is wrongzo.

src/JSInterop/Microsoft.JSInterop/ref/Microsoft.JSInterop.netstandard2.0.cs
pranavkm: We've been using 0 for null values. I had it changed to be nullable in my draft PR, but it requires a change to the hub protocol. At this point, it's better to play it safe and keep the data types the same.

src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetInvocationInfo.cs
pranavkm: Yup

PR #2210, opened 2019-08-21
Add switch to enable expected and unexpected arguments in any order

dougbu:
- unblocks work on aspnet/AspNetCore#4923
  - arguments for inside and outside men of service reference doc gen tool are mixed by default
  - users may add either argument type to the end of the outside man's command line
- e.g. "command --unexpected unexpectedValue --expected" can now set the "expected" option
  - only "--unexpected" and "unexpectedValue" are added to RemainingArguments in that case

- default behaviour of the command-line parser is unchanged to avoid breaking existing applications
- new switch is supported only when calling `CommandLineApplication` constructor for top-level commands
  - `dotnet-getdocument` (the outside man) has no (sub)commands and expanding scope would increase churn

nits: take VS suggestions in changed files

### DIFF
--- a/src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
+++ b/src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
@@ -13,14 +13,23 @@ namespace Microsoft.Extensions.CommandLineUtils
 {
     internal class CommandLineApplication
     {
-        // Indicates whether the parser should throw an exception when it runs into an unexpected argument.
-        // If this field is set to false, the parser will stop parsing when it sees an unexpected argument, and all
-        // remaining arguments, including the first unexpected argument, will be stored in RemainingArguments property.
+        // Indicates whether the parser should throw an exception when it runs into an unexpected argument. If this is
+        // set to true (the default), the parser will throw on the first unexpected argument. Otherwise, all unexpected
+        // arguments (including the first) are added to RemainingArguments.
         private readonly bool _throwOnUnexpectedArg;
 
-        public CommandLineApplication(bool throwOnUnexpectedArg = true)
+        // Indicates whether the parser should check remaining arguments for command or option matches after
+        // encountering an unexpected argument. Ignored if _throwOnUnexpectedArg is true (the default). If
+        // _throwOnUnexpectedArg and this are both false, the first unexpected argument and all remaining arguments are
+        // added to RemainingArguments. If _throwOnUnexpectedArg is false and this is true, only unexpected arguments
+        // are added to RemainingArguments -- allowing a mix of expected and unexpected arguments, commands and
+        // options.
+        private readonly bool _continueAfterUnexpectedArg;
+
+        public CommandLineApplication(bool throwOnUnexpectedArg = true, bool continueAfterUnexpectedArg = false)
         {
             _throwOnUnexpectedArg = throwOnUnexpectedArg;
+            _continueAfterUnexpectedArg = continueAfterUnexpectedArg;
             Options = new List<CommandOption>();
             Arguments = new List<CommandArgument>();
             Commands = new List<CommandLineApplication>();
@@ -145,6 +154,7 @@ namespace Microsoft.Extensions.CommandLineUtils
                     {
                         shortOption = arg.Substring(1).Split(new[] { ':', '=' }, 2);
                     }
+
                     if (longOption != null)
                     {
                         processed = true;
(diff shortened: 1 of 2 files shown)


BrennanConroy:
Is this for 3.1? Should it go through ask-mode?

dougbu:
@BrennanConroy I need to chat w/ @Pilchie about exactly where this should be targeted

src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -153,24 +163,40 @@ public int Execute(params string[] args)
...
+                            var ignoreContinueAfterUnexpectedArg = false;
+                            if (string.IsNullOrEmpty(longOptionName) &&
+                                !command._throwOnUnexpectedArg &&
+                                AllowArgumentSeparator)
                             {
-                                // skip over the '--' argument separator
+                                // Skip over the '--' argument separator then consume all remaining arguments. All
+                                // remaining arguments are unconditionally stored for further use.
                                 index++;
+                                ignoreContinueAfterUnexpectedArg = true;
```
pranavkm: Could we change this to say 

```
command.RemainingArguments.AddRange(new ArraySegment<string>(args, index, args.Length - index));
```

it's much easier to follow and you don't need to introduce a new parameter to this method.

src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -153,24 +163,40 @@ public int Execute(params string[] args)
...
                             }
 
-                            HandleUnexpectedArg(command, args, index, argTypeName: "option");
                             break;
                         }
 
                         // If we find a help/version option, show information and stop parsing
                         if (command.OptionHelp == option)
                         {
+                            arguments?.Dispose();
```
pranavkm: Is this required for 3.0? I'm worried a change like this in the release late might be destabilizing.

dougbu:
### More context
The command-line parser is unsuitable for use in service reference tools without this change. Mixing known (intended for the “outside man” executable) and unknown (just passed through to the “inside man” that runs in the context of the user’s app) command-line arguments doesn’t work; the existing code (and still, by default) stops looking for known command options as soon as it hits something unknown. The result is errors every time a user builds their Web API project because the outside man doesn’t see all its options.

Even if I reordered the options in the MSBuild glue to put known (outside man) arguments first, the experience is degraded. If a user adds an option to the end of the command that’s intended for both executables e.g. with `<OpenApiGenerateDocumentsOptions>--quiet</OpenApiGenerateDocumentsOptions>`, `--quiet` will only be passed through to the inside man – minimizing its usefulness.

Without addressing this particular TODO in aspnet/AspNetCore#4923, it’ll also be more difficult to test the code (aspnet/AspNetCore#4914).

src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
dougbu: Considered that but duplicating the code factored into `HandleUnexpectedArg(...)` would make maintenance more difficult, requires another `break` and wasn't as clear.

src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
dougbu: This was a VS suggestion I took because it made sense. `arguments` is a local variable (see https://github.com/aspnet/Extensions/pull/2210/files#diff-b972a436637005ce5a174421167fd548R138) referencing a disposable object. My preference is not to leave the VS message glowing in 3.0.

Also, what could this destabilize?

pranavkm: Let's remove this if it's not absolutely required. We don't have a way to test the impact of this change in all the places this gets used and it doesn't seem required for service reference to work.

dougbu:
:up::date: to rebase and address @pranavkm's comments

src/Shared/test/CommandLineApplicationTests.cs
```diff
@@ -530,6 +924,91 @@ public void ArgumentSeparator(string[] input, string[] expectedRemaining, string
...
+            // Help and Version options never get values; parsing ends when encountered.
+            Assert.False(optHelp.HasValue());
+            Assert.False(optVersion.HasValue());
+        }
+
+        [Theory]
+        [InlineData(new[] { "--", "a", "--top", "val" }, new[] { "--", "a", "--top", "val" }, null, false)]
+        [InlineData(new[] { "--", "--help" }, new[] { "--", "--help" }, null, false)]
+        [InlineData(new[] { "--", "--version" }, new[] { "--", "--version" }, null, false)]
+        [InlineData(new[] { "unexpected", "--", "--version" }, new[] { "unexpected", "--", "--version" }, null, false)]
```
dougbu: @BrennanConroy the behaviour for this case would be the same with `AllowArgumentSeparator = true` because nothing checks what's being added to `RemainingArguments` after hitting `unexpected`. There's a `dotnet-watch` test covering this in AspNetCore. But, I doubt the quirky behaviour is intentional i.e. devs setting `AllowArgumentSeparator = true` would be surprised to find `--` in `RemainingArguments` and it may cause user-visible oddities. Suggest filing an issue somewhere.

dougbu:
Any other comments @rynowak or @BrennanConroy❔

dougbu:
Rerunning OSX leg because failed test shouldn't have run; it's already marked as flaky. Likely hit one of the cases where traits are ignored in VSTest.

src/Shared/test/CommandLineApplicationTests.cs
BrennanConroy: I'm not sure there's an actionable issue here. The dotnet watch help text is very clear that **anything** after an unexpected argument is passed to the child process.

PR #2225, opened 2019-08-24
Add XML Docs for pubternal types

Pilchie:
Part of https://github.com/aspnet/AspNetCore/issues/3656


Tratcher:
@davidfowl why didn't we finish making these fully internal?

PR #2251, opened 2019-08-29
Make JsonConfigurationFileParser public

wayne-o:
The JSON config code depends on the fact that the JSON is coming from a file. This is not always the case and making this file public makes it easy for others to parse JSON which is not file based and use it as config.

 - Make JsonConfigurationFileParser public

### DIFF
--- a/src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
+++ b/src/Configuration/Config.Json/src/JsonConfigurationFileParser.cs
@@ -9,7 +9,7 @@ using System.Text.Json;
 
 namespace Microsoft.Extensions.Configuration.Json
 {
-    internal class JsonConfigurationFileParser
+    public class JsonConfigurationFileParser
     {
         private JsonConfigurationFileParser() { }
 


clement911:
We need exactly that. In our scenario, we are loading the JSON directly from a data store and need to parse the JSON string loaded in memory.

anurse:
Is copying this code into your own project not sufficient? Adding a new public API has a very high cost and I don't expect we'd be willing to take that cost here. The benefit is minimal over just copying the code into your project. The config file format is very simple and not likely to change. A public API needs to be reviewed to decide if it is in a good state to be supported for many more years, and we'd be heavily restricted in what changes we could make to the implementation (to avoid breaking consumers of this now-public API).

clement911:
Personally we can copy the code and that's fine but I suspect that loading config from JSON that is not stored in a file is not that uncommon.
That being said, this class doesn't need to be public necessarily.
Maybe there could be a `AddJsonStream `in addition to the existing `AddJsonFile `on the `IConfigurationBuilder`.

anurse:
> Maybe there could be a `AddJsonStream `in addition to the existing `AddJsonFile `on the `IConfigurationBuilder`.

Isn't there [already one](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.jsonconfigurationextensions.addjsonstream?view=dotnet-plat-ext-3.0)?

clement911:
Is there?
I don't see it there: https://github.com/aspnet/Extensions/blob/master/src/Configuration/Config.Json/src/JsonConfigurationExtensions.cs

It doesn't seem exist in asp.net core 2.2. Was this added in 3.0?
The link you sent refers to '.NET Platform Extensions'. What is this? I'm quite confused.

anurse:
Correct, it was added in 3.0.0-preview6.

clement911:
I see, well we don't really need this anymore then since we'll be upgrading to 3.0 in the next few weeks.

anurse:
Awesome! Glad to hear it!

I'm going to close this as we don't believe that adding a new public API is the right choice here.

PR #2289, opened 2019-09-06
Avoid resetting console color when the color has not changed

Kahbazi:


src/Logging/Logging.Console/src/AnsiLogConsole.cs
```diff
@@ -23,26 +26,33 @@ public AnsiLogConsole(IAnsiSystemConsole systemConsole)
         public void Write(string message, ConsoleColor? background, ConsoleColor? foreground)
         {
             // Order: backgroundcolor, foregroundcolor, Message, reset foregroundcolor, reset backgroundcolor
-            if (background.HasValue)
+            bool backgroundColorChanged = false;
+            bool foregroundColorChanged = false;
+
+            string backgroundColor = GetBackgroundColorEscapeCode(background.Value);
```
gfoidl: What happens when `background` is `null`?

src/Logging/Logging.Console/src/AnsiLogConsole.cs
```diff
@@ -23,26 +26,33 @@ public AnsiLogConsole(IAnsiSystemConsole systemConsole)
...
+            if (background.HasValue && _defaultBackgroundColor != backgroundColor)
             {
-                _outputBuilder.Append(GetBackgroundColorEscapeCode(background.Value));
+                backgroundColorChanged = true;
+                _outputBuilder.Append(backgroundColor);
             }
 
-            if (foreground.HasValue)
+            string foregroundColor = GetForegroundColorEscapeCode(foreground.Value);
```
gfoidl: What happens when `foreground` is `null`?

Kahbazi:
My main goal was to avoid resetting color when the foreground or background is actually the default color for example Debug and Trace level has the default color for Windows Console.
But you are right about the bugs, I will fix them.

src/Logging/Logging.Console/src/AnsiLogConsole.cs
```diff
@@ -23,26 +26,39 @@ public AnsiLogConsole(IAnsiSystemConsole systemConsole)
...
             {
-                _outputBuilder.Append("\x1B[39m\x1B[22m"); // reset to default foreground color
+                _outputBuilder.Append(_defaultForegroundColor); // reset to default foreground color
             }
 
-            if (background.HasValue)
+            if (backgroundColorChanged)
             {
-                _outputBuilder.Append("\x1B[49m"); // reset to the background color
+                _outputBuilder.Append(_defaultBackgroundColor); // reset to the background color
```
gfoidl:
```suggestion
                _outputBuilder.Append(_defaultBackgroundColor); // reset to default background color
```

I see you've taken the comment from the code before, but it should be aligned to the foreground-case.

src/Logging/Logging.Console/src/AnsiLogConsole.cs
```diff
@@ -11,6 +11,9 @@ namespace Microsoft.Extensions.Logging.Console
     /// </summary>
     internal class AnsiLogConsole : IConsole
     {
+        private const string _defaultForegroundColor = "\x1B[39m\x1B[22m";
```
gfoidl:
```suggestion
        private const string DefaultForegroundColor = "\x1B[39m\x1B[22m";
```
I'm not 100% sure about naming guidelines for consts in this repo, but I assume it should be PascalCase.

Kahbazi: 👍

anurse:
For the Windows one I'm OK with this, but for the ANSI one I have some concerns.

It's important to keep in mind here that the Console Logger is not the exclusive owner of the Console. Other code could be writing to the console and might write dangling ANSI color codes *in between* log messages. We reset the colors to the default at the end of the message, but nothing prevents another part of the app from coming along and writing console codes.

> My main goal was to avoid resetting color when the foreground or background is actually the default color for example Debug and Trace level has the default color for Windows Console.

What does that gain us? I have some concerns about possible scenarios this could break so I'd like to know what the improvement we get from this is. Is it more performant? Is there a functional issue being fixed?

I think I'd prefer that we're extra protective and always write color codes (even if they are the default) before *and after* writing log messages. There's certainly still a risk of interleaving

PR #2336, opened 2019-09-12
Support netcoreapp3.1 TFM

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore/issues/13581.


src/Caching/Memory/ref/Microsoft.Extensions.Caching.Memory.netcoreapp.cs
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Caching.Distributed
```
Tratcher: The old ref assembly files stick around?

JunTaoLuo: ah yes, I need to clean those up.

PR #2342, opened 2019-09-13
Add support for JSInvokable methods on generic types

pranavkm:
Prior to this change, DotNetDispatcher cached the MethodInfo on the
generic type definition. Using this would have required MethodInfo.MakeGenericMethod before the method was invoked.
We could separately cache the result of this to avoid the reflection cost per invocation.

Alternatively we could cache static and non-static MethodInfo instances separately which is what this change attempts to do.
The big difference in the outcome is that this requires instance (non-static) JSInvokable methods to be only unique named within
the type hierarchy as opposed to across all static and instance JSInvokable methods in an assembly.


src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetDispatcher.cs
```diff
@@ -301,7 +305,47 @@ private static (MethodInfo, Type[]) GetCachedMethodInfo(AssemblyKey assemblyKey,
...
+        {
+            var type = objectReference.Value.GetType();
+            var assemblyMethods = _cachedMethodsByType.GetOrAdd(type, ScanTypeForCallableMethods);
+            if (assemblyMethods.TryGetValue(methodIdentifier, out var result))
+            {
+                return result;
+            }
+            else
+            {
+                throw new ArgumentException($"The type '{type.Name}' does not contain a public invokable method with [{nameof(JSInvokableAttribute)}(\"{methodIdentifier}\")].");
```
pranavkm: Would exposing the type name be problematic in any way? For instance methods, reporting the object id isn't useful.

src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetDispatcher.cs
```diff
@@ -312,35 +356,22 @@ private static (MethodInfo, Type[]) GetCachedMethodInfo(AssemblyKey assemblyKey,
...
             var invokableMethods = GetRequiredLoadedAssembly(assemblyKey)
                 .GetExportedTypes()
-                .SelectMany(type => type.GetMethods(
-                    BindingFlags.Public |
-                    BindingFlags.DeclaredOnly |
-                    BindingFlags.Instance |
-                    BindingFlags.Static))
-                .Where(method => method.IsDefined(typeof(JSInvokableAttribute), inherit: false));
+                .SelectMany(type => type.GetMethods(BindingFlags.Public | BindingFlags.Static))
+                .Where(method => !method.ContainsGenericParameters && method.IsDefined(typeof(JSInvokableAttribute), inherit: false));
```
pranavkm: According to https://docs.microsoft.com/en-us/dotnet/api/system.reflection.methodbase.containsgenericparameters?view=netframework-4.8, methods containing generic parameters cannot be invoked:

>  If the ContainsGenericParameters property returns true, the method cannot be invoked.

This gives us the opportunity to treat these pretty much the same as non-public methods with `JsInvokable` and provide a more meaningful message. Without this check, you get strange errors from System.Text.Json.

src/JSInterop/Microsoft.JSInterop/test/Infrastructure/DotNetDispatcherTest.cs
```diff
@@ -790,6 +861,18 @@ public static async Task<string> AsyncThrowingMethod()
             }
         }
 
+        public class GenericType<TValue>
```
pranavkm: These scenarios work ✔️

src/JSInterop/Microsoft.JSInterop/test/Infrastructure/DotNetDispatcherTest.cs
```diff
@@ -790,6 +861,18 @@ public static async Task<string> AsyncThrowingMethod()
...
+        {
+            [JSInvokable] public string EchoStringParameter(string input) => input;
+            [JSInvokable] public TValue EchoParameter(TValue input) => input;
+        }
+
+        public class GenericMethodClass
```
pranavkm: These scenarios throw ❌

PR #2344, opened 2019-09-13
Fix packing on *nix systems

JunTaoLuo:
Without this the nupkgs packed on *nix systems will be missing files. It seems like the globbing pattern `**` doesn't work correctly in MSbuild on *nix so I'm moving the globbing into the nuspec. This is required to fix source-build. Keeping this as a draft until I get positive confirmation that this fixes the issue.

### DIFF
--- a/src/FileProviders/Embedded/src/Microsoft.Extensions.FileProviders.Embedded.csproj
+++ b/src/FileProviders/Embedded/src/Microsoft.Extensions.FileProviders.Embedded.csproj
@@ -27,9 +27,7 @@
 
   <ItemGroup>
     <NuspecProperty Include="AssemblyName=$(AssemblyName)" />
-    <NuspecProperty Include="OutputBinary=$(OutputPath)**\$(AssemblyName).dll" />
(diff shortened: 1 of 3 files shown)


dougbu:
>  if @dougbu is making changes to support stable builds

I'm only working in AspNetCore. Believe @mmitche and @JohnTortugo confirmed Extensions was fine wrt. stable builds.

Pilchie:
In that case, yes, lets keep it as a patch if that's okay with the source build folks.

JunTaoLuo:
Got confirmation that this resolves the issue for source build and that it's okay to make this fix in 3.1 instead of 3.0

JunTaoLuo:
Yea that's fine. It should resolve to use the netstandard2.0 props when the app targets netcoreapp3.1. This is the same pattern used in the Razor SDK: https://github.com/aspnet/AspNetCore-Tooling/tree/master/src/Razor/src/Microsoft.NET.Sdk.Razor

JunTaoLuo:
Note, I'll open another PR for a similar fix in AspNetCore

PR #2352, opened 2019-09-14
Add a perf scenario for DotNetDispatcher

pranavkm:
I tried playing around with using `ObjectResultExecutor` to avoid the reflection invoke in `DotNetDispatcher`. It does seem like a good change. We could consider a strategy where it does a reflection invoke the first few times around, but produces a compiled delegate if it's more frequently called. I recall DI has (❓) a pattern like this. 

# Vanilla benchmark

 Method |     Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
------- |---------:|----------:|----------:|----------:|-------:|----------:|
 Baseline | 1.220 us | 0.0103 us | 0.0080 us | 820,004.1 | 0.0057 |     240 B |
 Invoke | 1.002 us | 0.0064 us | 0.0060 us | 998,215.6 | 0.0038 |     200 B |
 
 
# 5000 operations per invoke
 Method |     Mean |    Error |   StdDev |        Op/s |   Gen 0 | Allocated |
------- |---------:|---------:|---------:|------------:|--------:|----------:|
 Baseline | 1.197 us | 0.0099 us | 0.0093 us | 835,661.3 | 31.2500 |   1.14 MB | 
 Invoke | 964.1 ns | 11.89 ns | 11.12 ns | 1,037,280.3 | 23.4375 | 976.56 KB |

### DIFF
--- a/src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetDispatcher.cs
+++ b/src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetDispatcher.cs
@@ -13,6 +13,8 @@ using System.Threading.Tasks;
 
 namespace Microsoft.JSInterop.Infrastructure
 {
+    using InvokeMethod = Func<object, object[], object>;
+
     /// <summary>
     /// Provides methods that receive incoming calls from JS to .NET.
     /// </summary>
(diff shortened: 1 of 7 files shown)


src/JSInterop/Microsoft.JSInterop/src/Infrastructure/DotNetDispatcher.cs
```diff
@@ -371,7 +374,8 @@ private static (MethodInfo methodInfo, Type[] parameterTypes) GetCachedMethodInf
                         $"the [JSInvokable] attribute.");
                 }
 
-                result.Add(identifier, (method, parameterTypes));
+                var invoker = ObjectMethodExecutor.GetExecutor(method, default);
```
rynowak: We need to be really careful with this. It's likely for WASM that using reflection is the implementation we want due to size.

SteveSandersonMS: Indeed. The numbers suggest < 20% speed/memory differences, so I wouldn't say this stands out as a good tradeoff.

pranavkm:
Fair enough. Let's get the infrastructure for measuring the WASM size impact before we try this out.

rynowak:
Can we add the test but remove the product code change?

PR #2358, opened 2019-09-16
Refactor xUnit extensibility

rynowak:
Adds our own hook for before/after logic that's more usable, called
`ITestMethodLifecycle`. This provides access to a context object
including the information about the test and the output helper. This can
be implemented by attributes or by the class itself.

The goal (and result) of this, is that we have a single *test executor*
extensibility point that provides all of the features we need. We should
use this everywhere we need features xUnit doesn't have.

Adding a new extensibility point (`ITestMethodLifecycle`) allows us to
do this without turning all of these features into a giant monolith.

---

Also updated our existing extensibility to use this new hook.

I did as much cleanup as a could to remove duplication from logging and
keep it loosly coupled. I didn't want to tease this apart completely
because the scope of this PR is already pretty large.


src/Logging/Logging.Testing/src/CollectDumpAttribute.cs
```diff
@@ -12,7 +17,23 @@ namespace Microsoft.Extensions.Logging.Testing
     /// This currently only works in Windows environments
     /// </remarks>
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
-    public class CollectDumpAttribute : Attribute
+    public class CollectDumpAttribute : Attribute, ITestMethodLifecycle
```
rynowak: This is probably a good place to start reviewing as any. This is an example of the kind of thing that's possible now.

rynowak:
Included a representative set of folks. Please take a look with some urgency. My goal is for no one to feel like this is a surprise when it hits aspnet/AspNetCore.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/RepeatContext.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Testing
+{
+    public class RepeatContext
+    {
+        private static AsyncLocal<RepeatContext> _current = new AsyncLocal<RepeatContext>();
```
rynowak: Open to suggestions about this. We have tests that test repeat, so we want this to be exposed somewhere. We could also just dumb down that test to use a dictionary or some-such and then we don't need this.

rynowak: I didn't create it so I don't want to ruin it.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/TestFileOutputContext.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace Microsoft.AspNetCore.Testing
+{
+    /// <summary>
+    /// Provides access to file storage for the running test. Get access by
+    /// implementing <see cref="ITestMethodLifecycle"/>, and accessing <see cref="TestContext.FileOutput"/>.
+    /// </summary>
+    /// <remarks>
+    /// Requires defining <see cref="AspNetTestFramework"/> as the test framework.
+    /// </remarks>
+    public sealed class TestFileOutputContext
```
rynowak: I tried to centralize as much as possible from `AssemblyTestLog`, but it's not really realistic to make it totally clean. 

This thing is conceptually oriented around test methods and test classes. 
`AssemblyTestLog` lets you pass in arbitrary strings.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/AspNetTestAssemblyRunner.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Testing
+{
+    public class AspNetTestAssemblyRunner : XunitTestAssemblyRunner
```
rynowak: OK, you have reached **the fun part**.

There's like 9 levels of extensibility we need to implement between the top level (framework), and the lowest level (invoker). Most of them are just pass-through.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/AspNetTestMethodRunner.cs
```diff
@@ -0,0 +1,53 @@
...
+            CancellationTokenSource cancellationTokenSource,
+            object[] constructorArguments)
+            : base(testMethod, @class, method, testCases, diagnosticMessageSink, messageBus, aggregator, cancellationTokenSource, constructorArguments)
+        {
+            _constructorArguments = constructorArguments;
+        }
+
+        protected override Task<RunSummary> RunTestCaseAsync(IXunitTestCase testCase)
+        {
+            if (testCase.GetType() == typeof(XunitTestCase))
```
rynowak: This is how we take control of the test execution without needing to implement a custom discoverer.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/ConditionalTheoryDiscoverer.cs
```diff
@@ -63,5 +64,24 @@ protected override IEnumerable<IXunitTestCase> CreateTestCasesForDataRow(ITestFr
...
+        protected override IEnumerable<IXunitTestCase> CreateTestCasesForSkippedDataRow(
+            ITestFrameworkDiscoveryOptions discoveryOptions,
+            ITestMethod testMethod,
+            IAttributeInfo theoryAttribute,
+            object[] dataRow,
+            string skipReason)
+        {
+            return new[]
+            {
+                new WORKAROUND_SkippedDataRowTestCase(DiagnosticMessageSink, discoveryOptions.MethodDisplayOrDefault(), discoveryOptions.MethodDisplayOptionsOrDefault(), testMethod, skipReason, dataRow),
```
rynowak: https://github.com/xunit/xunit/issues/1782

src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
```diff
@@ -102,5 +103,19 @@ public virtual void Dispose()
             _initializationException?.Throw();
             _testLog.Dispose();
         }
+
+        Task ITestMethodLifecycle.OnTestStartAsync(TestContext context, CancellationToken cancellationToken)
```
Tratcher: Can I use OnTestStartAsync to conditionally skip a test?
https://github.com/aspnet/AspNetCore/issues/11301

src/TestingUtils/Microsoft.AspNetCore.Testing/src/TestContext.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Microsoft.AspNetCore.Testing
+{
+    /// <summary>
+    /// Provides access to contextual information about the running tests. Get access by
+    /// implementing <see cref="ITestMethodLifecycle"/>.
+    /// </summary>
+    /// <remarks>
+    /// Requires defining <see cref="AspNetTestFramework"/> as the test framework.
+    /// </remarks>
+    public sealed class TestContext
```
Tratcher: How about Theory data?

src/TestingUtils/Microsoft.AspNetCore.Testing/src/RepeatContext.cs
jkotalik: Sure, that's fine.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/AspNetTestMethodRunner.cs
jkotalik: What type would a testCase be if the test was skipped?

rynowak: A test case that's skipped is one of two things:
1. Using `[ConditionalFact]` or similar it's a **subclass** of `XunitTestCase` and we would not hijake it
2. Using `[Fact(Skip = ...)]` it would be `XunitTestCase`, we would hijack it but it would still result in a skip

src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
rynowak: Unfortunately no. I didn't try to build control-flow into these methods.

src/TestingUtils/Microsoft.AspNetCore.Testing/src/TestContext.cs
rynowak: What do you mean? This context is set per-line-of-theory-data because that's xUnit's granularity. Each line-of-theory-data results in a new instance of the test class, etc.

Tratcher: I mean the data for that instance of the theory.

rynowak: yes then.

Issue #2353, opened 2019-09-14
CallSiteValidator doesn't cache results of the walk


davidfowl:
When ScopeValidation is on for the dependency injection container, we walk to resolution graph and find places where service lifetimes are mismatched (like singleton -> scoped). This logic currently doesn't mark visited services which leads to re-calculating results repeatedly which ends up being O(N!).

Run this example with scope validation on:
https://github.com/sergey-kolodiy/AspNetCore.DI

PR #2374, opened 2019-09-21
DI - Cache CallSiteValidator results during walk

brandondahler:
Caches the results of CallSiteValidator during its walk
 - Adds results to either _scopedServices or _nonScopedServices for each new service encountered instead of only adding at the top-level.
- Also caches results for implementation types as well since those would be the same as the service results.
- There's extra complexity around caching the results of services resolved under a EnumerableCallSite -- the service type should not be considered or results stored for it for any call sites that are direct children of a EnumerableCallSite.
- Tests added around the EnumerableCallSiteComplexity, name are... uhh... long, open to renaming them if you got better names in mind 🤷‍♂ 

Addresses #2353

# Performance Comparison
As measured by CallSiteValidatorBenchmark.cs:
## Before
BeforeUpdate

## After
AfterUpdate

### DIFF
--- a/src/DependencyInjection/DI/perf/CallSiteValidatorBenchmark.cs
+++ b/src/DependencyInjection/DI/perf/CallSiteValidatorBenchmark.cs
@@ -1,0 +1,186 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+using Microsoft.Extensions.DependencyInjection.ServiceLookup;
+
+namespace Microsoft.Extensions.DependencyInjection.Performance
+{
+    public class CallSiteValidatorBenchmark
+    {
+        private ServiceCallSite _callSite;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var services = new ServiceCollection();
+            services.AddTransient<A>();
+            services.AddTransient<B>();
+            services.AddTransient<C>();
+            services.AddTransient<D>();
+            services.AddTransient<E>();
+            services.AddTransient<F>();
+            services.AddTransient<G>();
+            services.AddTransient<H>();
+            services.AddTransient<I>();
+            services.AddTransient<J>();
+            services.AddTransient<K>();
+            services.AddTransient<L>();
+            services.AddTransient<M>();
+            services.AddTransient<N>();
+            services.AddTransient<O>();
+            services.AddTransient<P>();
+
+            var callSiteFactory = new CallSiteFactory(services.ToArray());
+
+            _callSite = callSiteFactory.GetCallSite(typeof(A), new CallSiteChain());
+        }
+
+        [Benchmark()]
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/DI/perf/CallSiteValidatorBenchmark.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        private class A
+        {
+            public A(B b, C c, D d, E e, F f, G g, H h, I i, J j, K k, L l)
+            {
+
+            }
+
+            [MethodImpl(MethodImplOptions.NoInlining)]
```
brandondahler: Nit - Remove

src/DependencyInjection/DI/src/ServiceLookup/CallSiteValidator.cs
```diff
@@ -11,15 +12,66 @@ internal class CallSiteValidator: CallSiteVisitor<CallSiteValidator.CallSiteVali
...
+                    {
+                        _nonScopedServices.Add(callSite.ServiceType);
+                    }
+                }
+            }
+
+            return scoped;
         }
 
+
```
brandondahler: Nit - Remove.

src/DependencyInjection/DI/test/ServiceProviderValidationTests.cs
```diff
@@ -110,6 +111,66 @@ public void GetService_DoesNotThrow_WhenScopeFactoryIsInjectedIntoSingleton()
             Assert.NotNull(result);
         }
 
+
```
brandondahler: Nit - Remove

src/DependencyInjection/DI/src/ServiceLookup/CallSiteValidator.cs
```diff
@@ -11,13 +12,69 @@ internal class CallSiteValidator: CallSiteVisitor<CallSiteValidator.CallSiteVali
...
+        // Cache already-checked services that resulted in null.
+        private readonly HashSet<Type> _nonScopedServices = new HashSet<Type>();
+
         public void ValidateCallSite(ServiceCallSite callSite)
         {
-            var scoped = VisitCallSite(callSite, default);
+            VisitCallSite(callSite, default);
+        }
+
+        protected override Type VisitCallSite(ServiceCallSite callSite, CallSiteValidatorState argument)
```
brandondahler: Nit - rename argument to state.

src/DependencyInjection/DI/src/ServiceLookup/CallSiteValidator.cs
```diff
@@ -11,13 +12,69 @@ internal class CallSiteValidator: CallSiteVisitor<CallSiteValidator.CallSiteVali
         // Keys are services being resolved via GetService, values - first scoped service in their call site tree
         private readonly ConcurrentDictionary<Type, Type> _scopedServices = new ConcurrentDictionary<Type, Type>();
 
+        // Cache already-checked services that resulted in null.
+        private readonly HashSet<Type> _nonScopedServices = new HashSet<Type>();
```
brandondahler: Should we just be using ServiceCacheKey instead?  Same for _scoped services?

Issue #2345, opened 2019-09-13
ASP.NET Core Dependency Injection container uses excessive amount of memory and CPU when large dependency graph is created


sergey-kolodiy:
### Description
ASP.NET Core Dependency Injection container uses excessive amount of memory and CPU when large dependency graph is created.

### How To Reproduce
I created a [repository](https://github.com/sergey-kolodiy/AspNetCore.DI) with a small project that reproduces the issue. It targets .NET Framework 4.7.2, but the issue is also reproducible for .NET Core 2.2.

1. Build the solution and run `AspNetCore.DI.Default` project.
2. Using Fiddler or some other tool of your choice, send multiple simultaneous HTTP requests to `http://localhost:5000/api/values` API endpoint. The issue is not always reproduced if you send a single HTTP request and then just wait for the response.
3. Observe resources consumption. You should see significant increase in both CPU and memory consumption:
![image](https://user-images.githubusercontent.com/741576/64855546-cabe0f00-d628-11e9-8523-01e74f913e48.png)
4. For comparison, run `AspNetCore.DI.Autofac` project that uses Autofac DI container instead of the default one and send multiple HTTP API requests again. You won't see those CPU and memory peaks this time.

### Expected behavior
Obviously the default DI container shouldn't use that much CPU and memory resources.

### So, what's going on?
Web API from the example above uses MVC middleware only, but, in addition to MVC services, it also registers a bunch of services with `Scoped` lifetime that are organized in a large graph (without any circular dependencies), and it seems like the default ASP.NET Core DI container is having some issues handling it.

In the request processing thread, the dependency graph is created and injected into the `ValuesController` just fine, and the API returns HTTP 200. But, it looks like for every dependency graph encountered, default DI container (or ASP.NET Core runtime) also spawns a separate background thread that performs some sort of initial processing and/or caching of that dependency graph. So, in theory this operation shouldn't affect request processing threads (at least when the dependency graph is relatively small).

However, it seems like this graph processing algorithm complexity grows exponentially, so the bigger dependency graph is, the more time and memory is consumed by this background operation. On my machine I had peaks up to 100% CPU (completely freezed) and more that 11GB memory consumption, which is not acceptable for production.

After this background process is done, resource consumption slowly decreases back to normal, and it doesn't start again when I invoke API methods on that controller, but for other controllers with different dependency graphs it starts again.

This issue is not reproducible when using Autofac DI container.

davidfowl:
[...]

PR #2375, opened 2019-09-22
[WIP] DI - Update ILEmitCallSiteAnalyzer to not recurse entire trees unnecessarily.

brandondahler:
Draft of expected changes that would be needed for #2345 
 - Need input from code owners about whether this is the correct direction to take.

Addresses #2345


brandondahler:
Side note - as-is the complexity of Transient is still at least 2^n.

PR #2400, opened 2019-09-25
Updating the documentation for SourceName

amgdy:
Updated the default event source name documentation to be **.NET Runtime** instead of **Application** to match the code execution


amgdy:
Fix for #2446

Issue #2403, opened 2019-09-26
Support value types in dependency injection


shaynevanasperen:
### There is already an ignored test in the source code relating to support for value types
There is a unit test named `WorksWithStructServices` which is ignored with the reason `"We don't support value task services currently"` over here: https://github.com/aspnet/Extensions/blob/master/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs#L183

### It would be great if we could get that unit test working
I would like to be able to use value types with dependency injection, to support:
* A **follow-up** feature for being able to resolve instances of `ValueTuple<T>` (and the other 7 varieties: `ValueTuple<T1, ... T8>`). The idea is that you'd be able to register services as normal (say `Foo` and `Bar`), and then somewhere in application code you could require a dependency of `ValueTuple<Foo, Bar>` and have the DI container automatically return a _transient_ instance of `ValueTuple<Foo, Bar>` initialized with services `Foo` and `Bar` resolved from the container.
* Just being able to register and resolve value types, because why not?

### Justification
I've authored [a library](https://github.com/shaynevanasperen/Magneto) which exposes a collection of abstract generic base classes, where consumers of the library close the type with type arguments. One of the type arguments represents a **"context"** which needs to be resolved from an `IServiceProvider`, and I wanted to support consumers of the library using a  `ValueTuple<T1, T2, ... Tn>` as their **"context"** type (so that they don't need to create a special class just to wrap a few arbitrary dependencies).

I actually implemented a solution for this by wrapping the `IServiceProvider` in my library with a small class that can do this resolution of `ValueTuple` types:
```cs
class ValueTupleServiceProvider : IServiceProvider
{
    readonly IServiceProvider _serviceProvider;

    public ValueTupleServiceProvider(IServiceProvider serviceProvider) =>
        _serviceProvider = serviceProvider;

    public object GetService(System.Type serviceType) =>
        serviceType.IsValueTupleType() // extension method defined elsewhere
            ? Activator.CreateInstance(serviceType,
                serviceType.GetValueTupleItemTypes() // extension method defined elsewhere
                    .Select(GetService).ToArray())
            : _serviceProvider.GetService(serviceType);
}
```
[...]

PR #2404, opened 2019-09-26
Add support for value types in dependency injection

shaynevanasperen:
This pull request adds support for value types in:
- `ExpressionsServiceProviderEngine`  and
- `ILEmitServiceProviderEngine`*

*(including **CLR** bug for `ServiceLifetime.Scoped` when targeting `netcoreapp5.0`)

Summary of the changes:
- Copy-paste typo fix in `ILEmitServiceProviderEngine.cs`
- Add two extra calls to `Convert` in `ExpressionResolverBuilder.cs`
- Update `Convert` in `ExpressionResolverBuilder.cs` to also check for `expression.Type.IsValueType`
- Emit `OpCodes.Box` as appropriate in `ILEmitResolverBuilder.cs`
- Update `ServiceProviderContainerTests.cs` to no longer skip tests for value types (structs)

There's a small problem with this though: I've had to exclude one specific scenario for the tests, as it fails with an error stating that the CLR detected an invalid program. I need some help to figure out what's wrong in this specific case and devise a solution. Is this perhaps an indication of an _actual_ bug in the CLR for `netcoreapp5.0`?

Addresses #2403

### DIFF
--- a/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
+++ b/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
@@ -10,11 +10,19 @@ using Microsoft.Extensions.DependencyInjection.Specification;
 using Microsoft.Extensions.DependencyInjection.Specification.Fakes;
 using Microsoft.Extensions.DependencyInjection.Tests.Fakes;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Microsoft.Extensions.DependencyInjection.Tests
 {
     public abstract class ServiceProviderContainerTests : DependencyInjectionSpecificationTests
     {
+        private readonly ITestOutputHelper _outputHelper;
+
+        protected ServiceProviderContainerTests(ITestOutputHelper outputHelper)
+        {
+            _outputHelper = outputHelper;
+        }
+
         [Fact]
         public void RethrowOriginalExceptionFromConstructor()
         {
@@ -180,12 +188,19 @@ namespace Microsoft.Extensions.DependencyInjection.Tests
             Assert.NotNull(provider.CreateScope());
         }
 
-        [Theory(Skip = "We don't support value task services currently")]
+        [Theory]
         [InlineData(ServiceLifetime.Transient)]
         [InlineData(ServiceLifetime.Scoped)]
         [InlineData(ServiceLifetime.Singleton)]
         public void WorksWithStructServices(ServiceLifetime lifetime)
         {
+#if NETCOREAPP5_0
+            if (GetType() == typeof(ServiceProviderILEmitContainerTests) && lifetime == ServiceLifetime.Scoped)
+            {
+                _outputHelper.WriteLine("We don't support scoped value type services when using the ILEmitServiceProviderEngine in netcoreapp5.0 currently");
+                return;
+            }
+#endif
             IServiceCollection serviceCollection = new ServiceCollection();
             serviceCollection.Add(new ServiceDescriptor(typeof(IFakeService), typeof(StructFakeService), lifetime));
             serviceCollection.Add(new ServiceDescriptor(typeof(StructService), typeof(StructService), lifetime));
--- a/src/DependencyInjection/DI/src/ServiceLookup/Expressions/ExpressionResolverBuilder.cs
+++ b/src/DependencyInjection/DI/src/ServiceLookup/Expressions/ExpressionResolverBuilder.cs
@@ -100,7 +100,9 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
                     ScopeParameter);
             }
 
-            return Expression.Lambda<Func<ServiceProviderEngineScope, object>>(VisitCallSite(callSite, null), ScopeParameter);
+            return Expression.Lambda<Func<ServiceProviderEngineScope, object>>(
+                Convert(VisitCallSite(callSite, null), typeof(object)),
+                ScopeParameter);
         }
 
         protected override Expression VisitRootCache(ServiceCallSite singletonCallSite, object context)
@@ -186,7 +188,7 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         private static Expression Convert(Expression expression, Type type)
         {
             // Don't convert if the expression is already assignable
-            if (type.GetTypeInfo().IsAssignableFrom(expression.Type.GetTypeInfo()))
+            if (!expression.Type.IsValueType && type.GetTypeInfo().IsAssignableFrom(expression.Type.GetTypeInfo()))
             {
                 return expression;
             }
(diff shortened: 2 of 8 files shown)


src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
```diff
@@ -180,12 +188,19 @@ public void ScopeDispose_PreventsServiceResolution()
...
 
-        [Theory(Skip = "We don't support value task services currently")]
+        [Theory]
         [InlineData(ServiceLifetime.Transient)]
         [InlineData(ServiceLifetime.Scoped)]
         [InlineData(ServiceLifetime.Singleton)]
         public void WorksWithStructServices(ServiceLifetime lifetime)
         {
+#if NETCOREAPP5_0
+            if (GetType() == typeof(ServiceProviderILEmitContainerTests) && lifetime == ServiceLifetime.Scoped)
```
shaynevanasperen: As mentioned in the pull request summary/description, this specific scenario fails with an exception:

```cs
System.InvalidProgramException : Common Language Runtime detected an invalid program.
   at ResolveService(ILEmitResolverBuilderRuntimeContext , ServiceProviderEngineScope )
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.Tests.ServiceProviderContainerTests.WorksWithStructServices(ServiceLifetime lifetime)
```

So until there's a solution for that, I've made it skip this specific scenario.

I'm very curious to know why this specific scenario fails **only** when running the unit tests targeting `netcoreapp5.0`. Perhaps there's a change in the CLR which has introduced a bug?

shaynevanasperen: @pakrym and @davidfowl I think the two of you, and perhaps @brandondahler would be interested in this.

brandondahler:
As mentioned in the line comment, I added this code and the test now succeeds.

``` patch
diff --git a/src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs b/src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
index bccc39888..9e4428371 100644
--- a/src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
+++ b/src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs
@@ -151,6 +151,11 @@ protected override object VisitConstructor(ConstructorCallSite constructorCallSi
             foreach (var parameterCallSite in constructorCallSite.ParameterCallSites)
             {
                 VisitCallSite(parameterCallSite, argument);
+
+                if (parameterCallSite.ServiceType.IsValueType)
+                {
+                    argument.Generator.Emit(OpCodes.Unbox_Any, parameterCallSite.ServiceType);
+                }
             }
             argument.Generator.Emit(OpCodes.Newobj, constructorCallSite.ConstructorInfo);
             return null;
```

_____

I wanted to make a more general statement that you'll probably need to look more closely at individual ways that the value types and/or objects are created or marshaled in the IL and ensure that we are boxing/unboxing as appropriate.

This specific instance is a good example where we're presumably producing IL with undefined behavior, but it is only being caught in the .net core runtime.  We'll need to be sure that we're producing valid and well-defined IL for all possible paths through this code.  Generally, we should not rely solely on the runtime flagging instances of badly formed IL.     

I have to defer to members of the aspnet team for better expertise and/or expectations of proof that the new IL is syntactically correct and well-defined.

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
brandondahler: I don't believe it is a CLR bug, I would presume it is that the newer runtime is better flagging IL that has (presumably) undefined behavior*.  I say this is because with a simple correction in the IL being generated, I was able to make the test succeed.

See my comment below for a change that prevents this exception from being thrown.

_* I am not an authoritative or even particularly well-informed source, so take my opinion with a grain of salt._

pakrym:
There is a reason I didn't fix the implementation when writing a test - this feature is a perf trap in so many ways even if implemented perfectly:
1. Resolving struct via GetService/GetService<T> would incur an allocation per resolution
2. Registering a struct via factory would incur an allocation per factory invocation.
3. Resolving a struct via an interface would incur an allocation per resolution.
4. Scope value cache is a `Dictionary<key, object>` so it's an allocation to store a struct into a cache.

But I'll leave deciding if it's worth having to @davidfowl and @anurse 

As far as implementation goes:
1. InvalidProgramException is definitely not a CLR bug and should be fixed.
1. I think there is a ton of extra boxing/unboxing happening especially in scoped service implementations: for example scoped resolver functions return `object` https://github.com/aspnet/Extensions/blob/0c5c499d05b9e2e76003f3198ef451ca48b7ab6d/src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs#L97, also a result local inside the scope resolver is defined as object https://github.com/aspnet/Extensions/blob/0c5c499d05b9e2e76003f3198ef451ca48b7ab6d/src/DependencyInjection/DI/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs#L315 
1. Needs way more coverage than the single test we have. Something like the following matrix:
	a. Resolved via GetService or injected into another object
	b. Resolved via struct type or via an interface it implements
	c. Registered as singleton/scoped/transient
	d. Registered as type/constant/factory
	e. Structs injected into classes, classes into structs, structs into structs

anurse:
I'll be honest, with @pakrym 's perf analysis and the issues with getting the IL right above, I'm not sure the benefit is outweighing the cost here.

If you want to be able to register an arbitrary value type in DI, there is always a workaround of embedding the value type in a class, which makes the application developer clearly aware of the allocation model in use (since they have to `new` up the class and generally that is known to cause heap allocation).

The original issue refers to the justification of being able to allow resolving `ValueTuple<...Tn>` (where `Tn` are other services in DI) and I certainly see where having to create a custom class in this situation isn't ideal. I believe `Tuple<...Tn>` could be used today, if I'm not interpreting things incorrectly, which may be an option.

If the goal is just to support `ValueTuple<...Tn>` as a constructor parameter when injecting, then we may have other options. One option, which I have **not fully thought through** (cc @davidfowl @pakrym), would be to solve this at the injection site by allowing it to directly support constructor arguments like `ValueTuple<...Tn>` and constructing the necessary tuple at the time the constructor is invoked instead of actually registering the `ValueTuple` in the container. That's a little special-casey though since it's **not** what we did for things like `IEnumerable`. But it would mean that instead of boxing and unboxing `ValueTuple`s all over the place, we're allocating a single `ValueTuple` right above where we call the constructor on the stack and thus don't need to box.

pakrym:
We tried not to add features to MEDI that are not supported by other containers. I wrote a small thing way back https://gist.github.com/pakrym/3a026eb03386467050da1ddadda7d4e5

brandondahler:
I agree with @pakrym -- I'm personally not seeing the value compared to the complexity of implementing it effective and efficiently.

anurse:
> We tried not to add features to MEDI that are not supported by other containers.

This is a very good point I had not initially considered.

Issue #1689, opened 2019-05-08
TimestampFormat is not validated, consequent exception doesn't show up in console


NinoFloris:
There's a direct call to `ToString` here, but if TimestampFormat is not valid it'll throw. 

https://github.com/aspnet/Extensions/blob/master/src/Logging/Logging.Console/src/ConsoleLogger.cs#L111-L121

However you actually won't see the exception of it in your console until that field is valid again (are old messages cached?)

Running on 3.0.100-preview5

EDIT:
Not only do you not get the exceptions, it dismantles kestrel completely. Causing connection hangs followed by connection resets.

NinoFloris:
Also it seems Standard Date and Time Format Strings don't work once there's any extra character beside the format is supplied. So this works `"u"` but this doesn't `"'['u']'"` respectively logging `2019-05-08 20:11:12Z` and `[u]`.

Though to really get a converted Z 8601 time (so converted into utc) doesn't seem possible with `u`. You just get your local timezone time with a Z character at the end. `K` helps a bit here by at least presenting the actual timezone, but true conversion into UTC time doesn't seem possible.
Would an additional option be possible? `"TimestampKind": "Utc"`

anurse:
So the issue here is that you don't see the exception anywhere? That does seem unpleasant. Perhaps we should catch exceptions here and force a useful message into the queue (something like "Error formatting log message: {blah...}"

NinoFloris:
It also stalls kestrel completely for some reason (exception escaping to lower levels?)

PR #2420, opened 2019-09-29
Provide capability to use UTC time notation for console logs

tomkerkhove:
Provide capability to use UTC time notation for console logs via `ConsoleLoggerOptions.UseUtcTimezone`.

This is turned off by default to be backwards compatible.

Addresses #1808

In terms of testing I would suggest to provide a similar test to `WriteCore_LogsCorrectTimestamp` which uses UTC option and verifies stamp or a similar one - Thoughts?

This replaces #1896 and is now based off of a clean branch.

### DIFF
--- a/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
+++ b/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Extensions.Logging.Console
         public bool IncludeScopes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.Extensions.Logging.LogLevel LogToStandardErrorThreshold { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public string TimestampFormat { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool UseUtcTimezone { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
     }
     [Microsoft.Extensions.Logging.ProviderAliasAttribute("Console")]
     public partial class ConsoleLoggerProvider : Microsoft.Extensions.Logging.ILoggerProvider, Microsoft.Extensions.Logging.ISupportExternalScope, System.IDisposable
--- a/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
+++ b/src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Extensions.Logging.Console
         public bool IncludeScopes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.Extensions.Logging.LogLevel LogToStandardErrorThreshold { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public string TimestampFormat { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
-        public bool UseUtcTimezone { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool UseUtcTimestamp { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
     }
     [Microsoft.Extensions.Logging.ProviderAliasAttribute("Console")]
     public partial class ConsoleLoggerProvider : Microsoft.Extensions.Logging.ILoggerProvider, Microsoft.Extensions.Logging.ISupportExternalScope, System.IDisposable
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -141,7 +141,8 @@ namespace Microsoft.Extensions.Logging.Console
             var timestampFormat = Options.TimestampFormat;
             if (timestampFormat != null)
             {
-                timestamp = Options.UseUtcTimezone ? DateTime.UtcNow.ToString(timestampFormat) : DateTime.Now.ToString(timestampFormat);
+                var dateTime = Options.UseUtcTimezone ? DateTime.UtcNow : DateTime.Now;
+                timestamp = dateTime.ToString(timestampFormat);
             }
 
             return new LogMessageEntry(
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -137,11 +137,16 @@ namespace Microsoft.Extensions.Logging.Console
                 logBuilder.AppendLine(exception.ToString());
             }
 
+            string timestamp = null;
             var timestampFormat = Options.TimestampFormat;
+            if (timestampFormat != null)
+            {
+                timestamp = Options.UseUtcTimezone ? DateTime.UtcNow.ToString(timestampFormat) : DateTime.Now.ToString(timestampFormat);
+            }
 
             return new LogMessageEntry(
                 message: logBuilder.ToString(),
-                timeStamp: timestampFormat != null ? DateTime.Now.ToString(timestampFormat) : null,
+                timeStamp: timestamp,
                 levelString: logLevelString,
                 levelBackground: logLevelColors.Background,
                 levelForeground: logLevelColors.Foreground,
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -31,27 +31,11 @@ namespace Microsoft.Extensions.Logging.Console
             _newLineWithMessagePadding = Environment.NewLine + _messagePadding;
         }
 
-        internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor)
+        internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor, ISystemClock systemClock = null)
         {
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
-            }
-
-            _name = name;
+            _name = name ?? throw new ArgumentNullException(nameof(name));
             _queueProcessor = loggerProcessor;
-        }
-
-        internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor, ISystemClock systemClock)
-        {
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
-            }
-
-            _name = name;
-            _queueProcessor = loggerProcessor;
-            _systemClock = systemClock;
+            _systemClock = systemClock ?? new SystemClock();
         }
 
         internal IExternalScopeProvider ScopeProvider { get; set; }
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -219,7 +219,7 @@ namespace Microsoft.Extensions.Logging.Console
 
         private DateTime GetCurrentDateTime()
         {
-            return Options.UseUtcTimezone ? DateTime.UtcNow : DateTime.Now;
+            return Options.UseUtcTimestamp ? DateTime.UtcNow : DateTime.Now;
         }
 
         public bool IsEnabled(LogLevel logLevel)
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Diagnostics;
 using System.Text;
-using Microsoft.Extensions.Internal;
 
 namespace Microsoft.Extensions.Logging.Console
 {
@@ -19,7 +18,6 @@ namespace Microsoft.Extensions.Logging.Console
 
         private readonly string _name;
         private readonly ConsoleLoggerProcessor _queueProcessor;
-        private readonly ISystemClock _systemClock = new SystemClock();
 
         [ThreadStatic]
         private static StringBuilder _logBuilder;
@@ -31,11 +29,10 @@ namespace Microsoft.Extensions.Logging.Console
             _newLineWithMessagePadding = Environment.NewLine + _messagePadding;
         }
 
-        internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor, ISystemClock systemClock = null)
+        internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor)
         {
             _name = name ?? throw new ArgumentNullException(nameof(name));
             _queueProcessor = loggerProcessor;
-            _systemClock = systemClock ?? new SystemClock();
         }
 
         internal IExternalScopeProvider ScopeProvider { get; set; }
@@ -217,8 +214,7 @@ namespace Microsoft.Extensions.Logging.Console
 
         private DateTime GetCurrentDateTime()
         {
-            var utcNow = _systemClock.UtcNow.DateTime;
-            return Options.UseUtcTimestamp ? utcNow : utcNow.ToLocalTime();
+            return Options.UseUtcTimestamp ? DateTime.UtcNow : DateTime.Now;
         }
 
         public bool IsEnabled(LogLevel logLevel)
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -31,7 +31,12 @@ namespace Microsoft.Extensions.Logging.Console
 
         internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor)
         {
-            _name = name ?? throw new ArgumentNullException(nameof(name));
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            _name = name;
             _queueProcessor = loggerProcessor;
         }
 
--- a/src/Logging/Logging.Console/src/ConsoleLogger.cs
+++ b/src/Logging/Logging.Console/src/ConsoleLogger.cs
@@ -217,7 +217,8 @@ namespace Microsoft.Extensions.Logging.Console
 
         private DateTime GetCurrentDateTime()
         {
-            return Options.UseUtcTimestamp ? _systemClock.UtcNow.DateTime : DateTime.Now;
+            var utcNow = _systemClock.UtcNow.DateTime;
+            return Options.UseUtcTimestamp ? utcNow : utcNow.ToLocalTime();
         }
 
         public bool IsEnabled(LogLevel logLevel)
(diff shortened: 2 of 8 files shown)


tomkerkhove:
Tagging @gfoidl since he reviewed it before.

Added one more comment in my other PR: https://github.com/aspnet/Extensions/pull/1896#discussion_r329352016

tomkerkhove:
Tagging @gfoidl given he reviewed my initial PR.

I've added a comment on my initial PR to tackle culture to avoid hitting the current test issue:
https://github.com/aspnet/Extensions/pull/1896#discussion_r329352016

src/Logging/test/ConsoleLoggerTest.cs
```diff
@@ -377,19 +378,65 @@ public void WriteCore_LogsCorrectTimestamp(ConsoleLoggerFormat format, LogLevel
...
-                {
-                    Assert.Single(sink.Writes);
-                    Assert.Matches("^<\\d>\\d{14}\\s[^\\s]", sink.Writes[0].Message);
-                    Assert.StartsWith(levelPrefix, sink.Writes[0].Message);
-                }
-                break;
+                    {
+                        Assert.Single(sink.Writes);
+                        Assert.StartsWith(levelPrefix, sink.Writes[0].Message);
+                        var regexMatch = Regex.Match(sink.Writes[0].Message, "^<\\d>(\\d{4}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2})\\s[^\\s]");
```
gfoidl: Apply the similar `Assert.Matches` as above.

src/Logging/test/ConsoleLoggerTest.cs
```diff
@@ -377,19 +378,65 @@ public void WriteCore_LogsCorrectTimestamp(ConsoleLoggerFormat format, LogLevel
...
-                    Assert.Matches("^\\d{14}\\s$", sink.Writes[0].Message);
-                    Assert.StartsWith(levelPrefix, sink.Writes[1].Message);
-                }
-                break;
+                    {
+                        Assert.Equal(3, sink.Writes.Count);
+                        Assert.StartsWith(levelPrefix, sink.Writes[1].Message);
+                        Assert.Matches("^\\d{4}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\s$", sink.Writes[0].Message);
+                        var parsedDateTime = DateTimeOffset.Parse(sink.Writes[0].Message.Trim());
+                        Assert.NotEqual(0, parsedDateTime.Offset.Hours);
```
gfoidl: This causes a [test-failure](https://dev.azure.com/dnceng/public/_build/results?buildId=370092&view=ms.vss-test-web.build-test-results-tab&runId=11249960&resultId=100145&paneView=debug).

src/Logging/test/ConsoleLoggerTest.cs
```diff
@@ -413,27 +460,27 @@ public void WriteCore_LogsCorrectMessages(ConsoleLoggerFormat format, LogLevel l
...
-                    Assert.Equal(2, sink.Writes.Count);
-                    Assert.Equal(
-                        levelPrefix + ": test[0]" + Environment.NewLine +
-                        "      This is a test, and {curly braces} are just fine!" + Environment.NewLine +
-                        "System.Exception: Exception message" + Environment.NewLine +
-                        "with a second line" + Environment.NewLine,
-                        GetMessage(sink.Writes));
-                }
-                break;
+                    {
```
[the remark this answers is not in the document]
tomkerkhove: This is more due to my local R# config, will revert them.

src/Logging/test/ConsoleLoggerTest.cs
```diff
@@ -377,19 +378,65 @@ public void WriteCore_LogsCorrectTimestamp(ConsoleLoggerFormat format, LogLevel
...
-                {
-                    Assert.Equal(3, sink.Writes.Count);
-                    Assert.Matches("^\\d{14}\\s$", sink.Writes[0].Message);
-                    Assert.StartsWith(levelPrefix, sink.Writes[1].Message);
-                }
-                break;
+                    {
+                        Assert.Equal(3, sink.Writes.Count);
+                        Assert.StartsWith(levelPrefix, sink.Writes[1].Message);
+                        Assert.Matches("^\\d{4}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\D\\d{2}\\s$", sink.Writes[0].Message);
```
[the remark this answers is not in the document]
tomkerkhove: Agreed, was wondering why this approach was used so more than happy to change it :D

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: Yes, see my comment here: https://github.com/aspnet/Extensions/pull/1896#discussion_r329352016

It works locally but my guess is that the build agent runs in UTC local which makes it fail. Ideally the tests sets the locale to a non-UTC one to enforce this but seeking help for this as I'm not an expert on that topic.

gfoidl: Oh, you're fast, I'm still thinking for a good reliable approach 😃

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: I'd love to but then I cannot do `regexMatch.Groups` to retrieve the stamp in systemd logs.

src/Logging/test/ConsoleLoggerTest.cs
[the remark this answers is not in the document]
tomkerkhove: Done!

src/Logging/test/ConsoleLoggerTest.cs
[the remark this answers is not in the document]
tomkerkhove: Reverted

src/Logging/test/ConsoleLoggerTest.cs
gfoidl: Hm, I think it either boils down to https://github.com/aspnet/Extensions/pull/1896#discussion_r300340699 or to add an `IDateTimeProvider` that is injected to the console-logger and can be mocked.
To change the current timezone can't be done with (pure) .NET, and I think we shouldn't change the system's timezone just for this test either.

The easiest would be to add `IDateTimeProvider`, but this comes with the cost of another level of indirection (and an interface-call -- which may harm perf, I have an issue with perf of console-logger in mind).
Perf-wise instead of the interface and abstract class (with sealed overrides in the concrete implementation) could be used.
Just add an ctor-overload (internal) which accepts an instance of `IDateTimeProvider`, that can be used for testing.
The default ctor uses a concrete default implementation.

That's quite some change in the code-base. If there are other places where this could be used to, I think it's worth it (maybe there's already such an abstraction lying around, didn't check that).

Otherwise just go with the first mentioned approach. To address your concern "who says that method is being called?" you could
* just rely on review :wink:
* for debug build implement a callback which you check, so like
```c#
// ConsoleLogger
private static string GetTimestamp(string timestampFormat, bool useUtcTimezone)
    => GetTimestamp(timestampFormat, useUtcTimezone, DateTime.Now);

#if DEBUG
public static Action GetTimestampCalled;
#endif

internal static string GetTimestamp(string timestampFormat, bool useUtcTimezone, DateTime now)
{
#if DEBUG
    GetTimestampCalled?.Invoke();
#endif

    if (timestampFormat == null)
        return null;

    if (useUtcTimezone)
        now = now.ToUniversalTime();

    return now.ToString(timestampFormat);
}
```
```c#
// Test, only run in DEBUG
    bool getTimestampCalled = false;
    void OnGetTimestampCalled() => getTimestampCalled = true;

    ConsoleLogger.GetTimestampCalled += OnGetTimestampCalled;
    try
    {
        string actual = logger.Log();

        Assert.NotNull(actual);
        Assert.True(getTimestampCalled);
    }
    finally
    {
        // de-registration may not be necessary in tests, but it prevents memory leaks (static event)
        ConsoleLogger.GetTimestampCalled -= OnGetTimestampCalled;
    }
```

I'm not absolutely sure what approach is better.

src/Logging/test/ConsoleLoggerTest.cs
gfoidl: Ups...then it's fine.

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: This feels a bit odd to me given we're changing the library just for sake of testing but if that's what you guys want, I can do that.

That said, I can take a crack at the `IDateTimeProvider` approach and see how that goes, if you want.

gfoidl: Even with the `IDateTimeProvider`-approach outlined above, we don't have a guarantee it is called (except we use a mock).
Now I think this will be over-engineering for this "simple" test. Right now I'll just go with https://github.com/aspnet/Extensions/pull/1896#discussion_r300340699, as it is quite simple (in code and tests).

PS: I'm not on the team, so I don't know what they want either :wink:

src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
```diff
@@ -24,6 +24,7 @@ public partial class ConsoleLoggerOptions
         public bool IncludeScopes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.Extensions.Logging.LogLevel LogToStandardErrorThreshold { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public string TimestampFormat { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool UseUtcTimezone { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
```
anurse: Time to open the naming thread :)

I think think should include some indication as to *where* we are using the timezone (i.e. the timestamp). Perhaps this:

```suggestion
        public bool UseUtcTimestamp { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
```

anurse: cc @davidfowl @rynowak

src/Logging/test/ConsoleLoggerTest.cs
anurse: > It works locally but my guess is that the build agent runs in UTC local which makes it fail

All our build agents run in Azure, which sets the local time zone to UTC. We also need the tests to be reliable in places that do that, so we need them to pass on UTC time zones.

This is indeed a challenging one to automate, especially since all our CI automation runs in UTC already so toggling this option is a no-op. Changing the locale of the machine during the build is error-prone at best.

We do have an [`ISystemClock` abstraction we've used in places](https://github.com/aspnet/Extensions/blob/master/src/Caching/Abstractions/src/Internal/ISystemClock.cs). It might be worth looking at a similar approach here.

src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
tomkerkhove: Give it's in `ConsoleLoggerOptions` I think it's pretty straightforward but happy to change it to `ConsoleLoggerOptions.UseUtcTimestampNotation`, `ConsoleLoggerOptions.UseUtcTimestampMessageNotation` or whatever you prefer, it's just a suggestion!

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: I'll try to see if I can make some sense out of it, I'm not that much of an expert on that topic 😅 

I'm assuming you want to go the same route as above where `ISystemClock` is injected via an internal ctor or what do you have in mind exactly?

src/Logging/Logging.Console/ref/Microsoft.Extensions.Logging.Console.netcoreapp.cs
anurse: I think `UseUtcTimezone` is sufficient here. I just think `UseUtcTimezone` wasn't entirely clear as to what we were affecting.

tomkerkhove: `UseUtcTimestamp` is fine for me, ok for you @davidfowl & @rynowak ?

davidfowl: 👍

tomkerkhove: Done!

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: The approach worked, thanks for the tip.

However, there is one issue with the references but I'll add a comment there.

src/Logging/test/ConsoleLoggerTest.cs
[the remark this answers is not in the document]
gfoidl: Spacing-issue arose again :wink: Please fix.

src/Logging/Logging.Console/src/Microsoft.Extensions.Logging.Console.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <Compile Include="../../shared/*.cs" />
+    <ProjectReference Include="..\..\..\Caching\Abstractions\src\Microsoft.Extensions.Caching.Abstractions.csproj" />
```
tomkerkhove: @anurse I had to add this reference to be able to use `ISystemClock`. This feels dirty and should be moved out of `Microsoft.Extensions.Caching.Abstractions.csproj` but not sure if 1) you agree and 2) this should be done via this PR.

src/Logging/test/Microsoft.Extensions.Logging.Tests.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFrameworks>$(DefaultNetCoreTargetFramework);net472</TargetFrameworks>
   </PropertyGroup>
 
+  <ItemGroup>
```
tomkerkhove: @anurse I had to add this reference to be able to use `ISystemClock`, `SystemClock` & `TestClock`. This feels dirty and should be moved out of `Microsoft.Extensions.Caching.Abstractions.csproj` but not sure if 1) you agree and 2) this should be done via this PR.

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -40,6 +42,18 @@ internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor)
             _queueProcessor = loggerProcessor;
         }
 
+        internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor, ISystemClock systemClock)
```
gfoidl: Just forward the other ctor to this one:
```c#
internal ConsoleLogger(string name, ConsoleLoggerProcessor loggerProcessor) : this(name, loggerProcessor, new SystemClock()) {}
```
No need to duplicate logic.

Or as they are internal, just merge them with default param `systemClock = null` and the `_systemClock = systemClock ?? new SystemClock();`.

src/Logging/Logging.Console/src/ConsoleLogger.cs
```diff
@@ -210,6 +231,11 @@ static void AppendAndReplaceNewLine(StringBuilder sb, string message)
             }
         }
 
+        private DateTime GetCurrentDateTime()
+        {
+            return Options.UseUtcTimestamp ? _systemClock.UtcNow.DateTime : DateTime.Now;
```
```suggestion
            var utcNow = _systemClock.UtcNow.DateTime;
            return Options.UseUtcTimestamp ? utcNow : utcNow.ToLocalTime();
```
But I don't know...

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: Actually I got confused and [`ISystemClock`](https://github.com/aspnet/Extensions/blob/master/src/Caching/Abstractions/src/Internal/ISystemClock.cs) won't cut it given it only supports `Utc` and not local time.

What I can do is extend it so that it contains a `Now` property, but wanted to check with you guys first if that's ok.

src/Logging/test/ConsoleLoggerTest.cs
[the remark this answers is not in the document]
tomkerkhove: I will, sorry man!

src/Logging/Logging.Console/src/ConsoleLogger.cs
tomkerkhove: Sounds good to me, sloppy change on my side - Sorry!

tomkerkhove:
Not exactly given the TestClock uses a stamp from 2013 so I can rely on that.

Will incorporate the required changes ASAP.

src/Logging/Logging.Console/src/Microsoft.Extensions.Logging.Console.csproj
davidfowl: This is no go. We can't have this assembly depend on caching. Unfortunately the ISystemClock was never made to be a proper abstraction.

tomkerkhove: Hence why I'm suggesting to move it out given we need it for this test, or an alternative approach has to be provided..

tomkerkhove:
I'll hold off making any more changes until you've agreed upon an approach to test this @anurse  @davidfowl given I've received mixed suggestions/guidance on this.

src/Logging/test/ConsoleLoggerTest.cs
anurse: Actually, why not just compare the offset to `DateTimeOffset.Now.Offset` (and `DateTimeOffset.UtcNow.Offset` in the other case)? Yes, it means that on our build agents it's always testing against the UTC offset but that's because our agents run with UTC as the local time zone. In other builds (local dev builds, etc.) that won't be the case.

I think that may be the best we can do without introducing a lot of extra abstractions, internal code, mocks, etc.

src/Logging/Logging.Console/src/Microsoft.Extensions.Logging.Console.csproj
anurse: Agreed, we can't take a reference. I think we can just rely on some level of manual testing and code review here and simplify the test (commented elsewhere) to just compare the offset of the generated timestamp to the system-reported local time zone when `UseUtcTimestamp` is `false` (even if that local time zone is UTC).

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: Actually that's a good point - Will change this later today/tomorrow!

anurse: Thanks! Testing things involving `DateTime` is notoriously tricky after all ;).

src/Logging/test/Microsoft.Extensions.Logging.Tests.csproj
tomkerkhove: Removed

src/Logging/Logging.Console/src/Microsoft.Extensions.Logging.Console.csproj
tomkerkhove: Removed

src/Logging/test/ConsoleLoggerTest.cs
tomkerkhove: My pleasure, things should have been changed now.

That is correct, but it should not be a reason for creating such a sloppy PR like this one - Sorry for that!

tomkerkhove:
All the required changes should be done and are good to go as far as I see.

Thank you for the reviews @gfoidl @anurse @davidfowl - Feel free to let me know if I need to change something!

Thank you for your patience, this was not one of my best/cleanest PRs :/

src/Logging/Logging.Console/src/Microsoft.Extensions.Logging.Console.csproj
```diff
@@ -15,7 +15,7 @@
 
   <ItemGroup>
     <Compile Include="../../shared/*.cs" />
-
+
```
gfoidl: Nit: remove that extra space

tomkerkhove: Done!

anurse:
LGTM. @davidfowl new API here (option to enable UTC timestamp in the console logger), any review?

tomkerkhove:
Any feedback @davidfowl ?

anurse:
I think the API is fine. I'll kick the build, so that we have a fresh build to confirm we won't break anything, and we should be able to merge this.

tomkerkhove:
Thanks @anurse!

tomkerkhove:
Thanks for the review @davidfowl !

tomkerkhove:
🎉

NinoFloris:
@tomkerkhove thanks for your PR! ran into this in https://github.com/dotnet/extensions/issues/1689 as well

Issue #2409, opened 2019-09-26
Use DI to Configure IExternalScopeProvider


TheXenocide:
### Is your feature request related to a problem? Please describe.
I am trying to collate ambient and global diagnostic data to be provided as structured event data with logs and have found that, despite the fact that an `IExternalScopeProvider` interface exists, there's no way to configure the default `LoggerFactory` to use anything but the default LoggerExternalScopeProvider. I literally only need to supply one extra state object to loggers that support external scope which, ideally, would involve inheriting the default provider and overriding `ForEachScope` (or some similar such facility that would let me tie in to all scope-aware loggers to provide one additional global-but-contextual set of metadata).

### Describe the solution you'd like
The default `LoggerExternalScopeProvider` should be registered using the same `TryAdd`/`Singleton` behavior seen in `LoggingServiceCollectionExtensions.AddLogging` as is used for other default logging interface implementations and `LoggerFactory` should receive it as an injected constructor parameter.

### Describe alternatives you've considered
In the meantime (unless there's a more advisable solution) I think I'm going to wind up creating a decorator (since none of the methods I need to override are virtual) implementation of `ILoggerFactory` whose only purpose is to wrap the implementation of `LoggerFactory` so that, after calling the default implementation's `AddProvider` method I can call `SetScopeProvider` to replace the one the default implementation used with a decorator implementation of `IExternalScopeProvider` in which the implementation of `ForEachScope` will call the callback with a state object gathered from our custom contextual diagnostic scope and global values either before or after calling the default implementation.

### Additional context
From the archived repository: https://github.com/aspnet/Logging/issues/875
Related: https://github.com/aspnet/Extensions/issues/644
Question that prompted my deep dive: https://github.com/aspnet/Extensions/issues/2398
PR containing an alternative non-DI solution: https://github.com/aspnet/Extensions/pull/2170

TheXenocide:
[...]

PR #2434, opened 2019-10-01
IExternalScopeProvider can now be injected (#2409)

TheXenocide:
- Added IExternalScopeProvider parameter to LoggerFactory to allow injection.
- Registers default implementation if not otherwise registered.

Addresses #2409

Note: this is currently based on the release/3.0 branch since I verified that it would solve the problems I'm having with the current stable release. I can rebase to 3.1 if necessary.

### DIFF
--- a/src/Logging/Logging.Abstractions/ref/Microsoft.Extensions.Logging.Abstractions.netstandard2.0.cs
+++ b/src/Logging/Logging.Abstractions/ref/Microsoft.Extensions.Logging.Abstractions.netstandard2.0.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Extensions.Logging
     public partial class LoggerExternalScopeProvider : Microsoft.Extensions.Logging.IExternalScopeProvider
     {
         public LoggerExternalScopeProvider() { }
-        public void ForEachScope<TState>(System.Action<object, TState> callback, TState state) { }
+        public virtual void ForEachScope<TState>(System.Action<object, TState> callback, TState state) { }
         public System.IDisposable Push(object state) { throw null; }
     }
     public static partial class LoggerFactoryExtensions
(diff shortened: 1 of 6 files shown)


davidfowl:
> Note: this is currently based on the release/3.0 branch since I verified that it would solve the problems I'm having with the current stable release. I can rebase to 3.1 if necessary.

Best to assume that everything goes into master. There's an approval process for things to make it into 3.1 and this doesn't meet the bar.

TheXenocide:
Just so I understand what you mean about 3.1, is that meaning our team definitely won't be able to use IExternalScopeProvider until .NET 5.0? I don't mind rebasing to master, just wondering what the implications are for our adoption schedule and what kinds of requests *do* meet the bar for 3.1 (I'm guessing this just means the scope of work is already finalized)? Other than that, one thing this PR does is mark a method as virtual, though I'm wondering if that's not worth discussing further (possibly in the original feature request; not really sure where would be more appropriate)? Thanks for your time.

davidfowl:
> Just so I understand what you mean about 3.1, is that meaning our team definitely won't be able to use IExternalScopeProvider until .NET 5.0? I don't mind rebasing to master, just wondering what the implications are for our adoption schedule and what kinds of requests do meet the bar for 3.1 (I'm guessing this just means the scope of work is already finalized)? Other than that, one thing this PR does is mark a method as virtual, though I'm wondering if that's not worth discussing further (possibly in the original feature request; not really sure where would be more appropriate)? Thanks for your time.

Yes that's correct. The 3.1 bar is *extremely high*, you can think about it like a patch for 3.0 (mostly bug fixes and critical feedback from 3.0). This is a new feature that introduces a new API which needs to be reviewed and signed off and doesn't meet that bar.

TheXenocide:
global.json in master is set to .NET 5 Alpha already, which I haven't found anywhere (and I'm not particularly excited about building all of .NET from source just to modify a few lines here and there). Is there guidance for how to work against master?

TheXenocide:
Okay, found the installer for latest SDK bits

TheXenocide:
I haven't forgotten about this, but now that we've worked around the issue at the office, any contribution I make will likely be from home (harder to find spare time, but probably better for keeping my build environments isolated since I really don't want to install the preview SDK on my work machine).

PR #2442, opened 2019-10-02
Fixing a typo in ChainedBuilderExtensions

francisspor:
The class description comment had a typo, so I cleaned that up and added some more details to what the extension methods are fore.

### DIFF
--- a/src/Configuration/Config/src/ChainedBuilderExtensions.cs
+++ b/src/Configuration/Config/src/ChainedBuilderExtensions.cs
@@ -7,7 +7,7 @@ using System.Collections.Generic;
 namespace Microsoft.Extensions.Configuration
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


src/Configuration/Config/src/ChainedBuilderExtensions.cs
```diff
@@ -7,7 +7,7 @@
 namespace Microsoft.Extensions.Configuration
 {
     /// <summary>
-    /// IConfigurationBuilder extension methods for the chaind configuration provider.
+    /// Extension methods for adding <see cref="IConfiguration"/> to a <see cref="IConfigurationBuilder"/>.
```
Eilon: Consider changing `a` to `an` (because it's spoken as `an I Configuration Builder).

anurse:
Thanks @francisspor !

PR #2449, opened 2019-10-03
Add OSMinVersionAttribute

Tratcher:
@anurse Testing changes only.

This is a test infrastructure improvement to support some 3.0 and 3.1 feature work that we've done in AspNetCore such as https://github.com/aspnet/AspNetCore/issues/13356, https://github.com/aspnet/AspNetCore/issues/14382.

These features only work on very recent versions of Win10 and we need to mark them accordingly.

This is also intended as a replacement for the version logic on OSSkipCondition where you had to [explicitly list](https://github.com/aspnet/AspNetCore/blob/894d9af96f007340e975e77bbe21c27a5b5be690/src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs#L27) every version to skip. That doesn't scale when we're supporting/skipping so many versions of Windows. We can convert tests to use this new attribute instead and cut the version logic from OSSkipCondition.

For now this is restricted to Windows but it could be extended to Mac if there's a need.

### DIFF
--- a/src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/OSMinVersionAttribute.cs
+++ b/src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/OSMinVersionAttribute.cs
@@ -1,0 +1,81 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Testing
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = true)]
+    public class OSMinVersionAttribute : Attribute, ITestCondition
+    {
+        private readonly OperatingSystems _targetOS;
+        private readonly Version _minVersion;
+        private readonly OperatingSystems _currentOS;
+        private readonly Version _currentVersion;
+        private readonly bool _skip;
+
+        /// <summary>
+        /// Used to indicate the minimum version a test can run on for the given operating system. 
+        /// Also add <see cref="OSSkipConditionAttribute"/> to skip other operating systems.
+        /// </summary>
+        /// <param name="targetOS">The OS to check for a version. Only Windows is currently supported.</param>
+        /// <param name="minVersion">The minimum OS version NOT to skip.</param>
+        public OSMinVersionAttribute(OperatingSystems targetOS, string minVersion) :
+            this(targetOS, Version.Parse(minVersion), GetCurrentOS(), GetCurrentOSVersion())
+        {
+        }
+
+        // to enable unit testing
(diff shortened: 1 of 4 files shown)


src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/OSMinVersionAttribute.cs
```diff
@@ -0,0 +1,81 @@
...
+                throw new NotImplementedException(targetOS.ToString());
+            }
+
+            _targetOS = targetOS;
+            _minVersion = minVersion;
+            _currentOS = currentOS;
+            _currentVersion = currentVersion;
+
+            _skip = _targetOS == _currentOS && _minVersion > _currentVersion;
+            SkipReason = $"The test cannot run on this operating system version '{currentVersion}'.";
```
halter73: Nit: might be convenient to put the _targetOS in the SkipReason for if/when we eventually start supporting multiple OSs.

src/TestingUtils/Microsoft.AspNetCore.Testing/test/OSMinVersionTest.cs
```diff
@@ -0,0 +1,73 @@
...
+            Assert.False(
+                RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
+                Environment.OSVersion.Version.ToString().StartsWith("6.1"),
+                "Test should not be running on Win7 or Win2008R2.");
+        }
+
+        [ConditionalTheory]
+        [OSMinVersion(OperatingSystems.Windows, WindowsVersions.Win8)]
+        [InlineData(1)]
+        public void TestTheorySkipWindows(int arg)
```
halter73: Nit: This name doesn't follow the convention of the other tests

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/OSMinVersionAttribute.cs
Tratcher: @rynowak have you noticed that the skip reasons don't show up in VS or on the command line anymore? That's what I was seeing today.

Tratcher: The target and current OS must match for the test to be skipped. We usually know what the current OS was.

src/TestingUtils/Microsoft.AspNetCore.Testing/test/OSMinVersionTest.cs
Tratcher:
```suggestion
        public void RunTheory_Win8DoesNotRunOnWin7(int arg)
```

src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/OSMinVersionAttribute.cs
rynowak: No I have not noticed that.

PR #2464, opened 2019-10-05
Support configuring the Azure Key Vault IConfigurationSource as optional.

danielcweber:
So far, most of the various configuration sources in aspnet/Extensions can be configured as optional, i.e. they won't throw in case configurations cannot be accessed. This PR adds this feature for Azure Key Vault.

There are 3 scenarios where secret retrieval from Key Vault might fail and you might want to continue anyway:

- Invalid or unreachable host name. This is much like a file that isn't there. Maybe you haven't set up a Key Vault yet and your host name says "https://invalid" or something else. 
- Failure to obtain a valid access token. This happens locally when the user hasn't logged into Azure with "az login" or in an Azure deployment when managed identity isn't configured. You might not want your app to crash then.
- Insufficient permissions (List/Get Secrets).

### DIFF
--- a/src/Configuration/Config.AzureKeyVault/src/KeyVaultClientWrapper.cs
+++ b/src/Configuration/Config.AzureKeyVault/src/KeyVaultClientWrapper.cs
@@ -1,6 +1,10 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Azure.KeyVault;
 using Microsoft.Azure.KeyVault.Models;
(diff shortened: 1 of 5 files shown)


src/Configuration/Config.AzureKeyVault/src/KeyVaultClientWrapper.cs
```diff
@@ -11,21 +15,64 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
...
+            {
+                return await _keyVaultClientImplementation.GetSecretsAsync(vault);
+            }
+            catch (System.Net.Http.HttpRequestException)
+            {
+                //Host unknown or not reachable.
+                if (!_optional)
+                    throw;
+            }
+            catch (Azure.Services.AppAuthentication.AzureServiceTokenProviderException)
```
Tratcher: These are invalid config conditions, not a problem with availability.

src/Configuration/Config.AzureKeyVault/src/KeyVaultClientWrapper.cs
```diff
@@ -11,21 +15,64 @@ namespace Microsoft.Extensions.Configuration.AzureKeyVault
...
         {
-            return _keyVaultClientImplementation.GetSecretsAsync(vault);
+            try
+            {
+                return await _keyVaultClientImplementation.GetSecretsAsync(vault);
+            }
+            catch (System.Net.Http.HttpRequestException)
+            {
+                //Host unknown or not reachable.
+                if (!_optional)
```
Tratcher: Style: always use `{ }`

src/Configuration/Config.AzureKeyVault/src/KeyVaultClientWrapper.cs
danielcweber: Right, not a problem with availability, but not necessarily invalid config conditions either I think. I get this exception when running code locally and I haven't logged into Azure with "az login". So it's not an invalid config condition but...maybe the scenario here is "I want my code to run even though I haven't logged into Azure".

src/Configuration/Config.AzureKeyVault/src/KeyVaultClientWrapper.cs
danielcweber: Will revise.

danielcweber:
Edited the initial post and mentioned the 3 scenarious that would make you want to continue vs. throwing.

HaoK:
These packages are going away and moving to azure sdk maintained packages, see https://github.com/aspnet/Announcements/issues/409  feel free to reopen this PR in the new azure integration packages over there.

Issue #1799, opened 2019-06-05
Two more test cases for DependencyInjectionSpecificationTests to test disposable singletons


peter-perot:
.NET Core dependency injection is a "tracking" container, i.e. it tracks disposable objects and disposes them if the container is disposed or the scope is left.

When a disposable singleton is registered, the container **only tracks the object if it is created by a factory**, in order to dispose it if the container itself is disposed. If the singleton is **added as an existing instance, the container does not track the object**, and it does not dispose it. This behavior is correct since disposable singletons which are added as an existing instance are considered "externally owned".

However, some 3rd party containers which implement the .NET Core abstractions behave differently here, e.g. Autofac.

I propose to file two additional test cases to check the behavior mentioned above:

```csharp
using System;
using System.Collections.Generic;
using Xunit;

namespace Microsoft.Extensions.DependencyInjection.Specification
{
  public abstract class DependencyInjectionSpecificationTests2 : DependencyInjectionSpecificationTests
  {
    [Fact]
    public void SingletonServiceCreatedFromFactoryIsDisposedWhenContainerIsDisposed()
    {
      // Arrange
      var collection = new TestServiceCollection();
      collection.AddSingleton(_ => new TestDisposable());
      var provider = this.CreateServiceProvider(collection);

      // Act
      var service = provider.GetService<TestDisposable>();
      ((IDisposable)provider).Dispose();

      // Assert
      Assert.True(service.IsDisposed);
    }

    [Fact]
    public void SingletonServiceCreatedFromInstanceIsNotDisposedWhenContainerIsDisposed()
    {
      // Arrange
      var collection = new TestServiceCollection();
      collection.AddSingleton(new TestDisposable());
      var provider = this.CreateServiceProvider(collection);

      // Act
      var service = provider.GetService<TestDisposable>();
      ((IDisposable)provider).Dispose();

      // Assert
      Assert.False(service.IsDisposed);
    }

    internal class TestDisposable : IDisposable
    {
      public bool IsDisposed { get; private set; }

      public void Dispose()
      {
        this.IsDisposed = true;
      }
    }

    internal class TestServiceCollection : List<ServiceDescriptor>, IServiceCollection
    {
    }
  }
}
```

anurse:
This seems reasonable to me (thoughts @davidfowl @pakrym ?). I don't have a ton of context on the DI spec tests. Would you like to open a PR?

davidfowl:
[...]

PR #2465, opened 2019-10-05
Add disposable singleton test cases

tomakita:
Added two test cases that check whether or not disposable singletons are disposed as expected after dependency injection and disposal.

Addresses #1799

NOTE: All credit for this idea and code go to @peter-perot.  I merely submitted the PR, since the issue was opened 4 months ago and has since had the help-wanted label added.  If @peter-perot would like to submit this PR and have this one deleted, that's fine too.

### DIFF
--- a/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
+++ b/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
@@ -384,7 +384,7 @@ namespace Microsoft.Extensions.DependencyInjection.Tests
 
             public void Dispose()
             {
-                this.IsDisposed = true;
+                IsDisposed = true;
             }
         }
 
--- a/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
+++ b/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
@@ -346,6 +346,52 @@ namespace Microsoft.Extensions.DependencyInjection.Tests
                 exception.Message);
         }
 
+        [Fact]
+        public void SingletonServiceCreatedFromFactoryIsDisposedWhenContainerIsDisposed()
+        {
+            // Arrange
+            var collection = new TestServiceCollection();
+            collection.AddSingleton(_ => new TestDisposable());
+            var provider = this.CreateServiceProvider(collection);
+
+            // Act
+            var service = provider.GetService<TestDisposable>();
+            ((IDisposable)provider).Dispose();
+
+            // Assert
+            Assert.True(service.IsDisposed);
+        }
+
+        [Fact]
+        public void SingletonServiceCreatedFromInstanceIsNotDisposedWhenContainerIsDisposed()
+        {
+            // Arrange
+            var collection = new TestServiceCollection();
+            collection.AddSingleton(new TestDisposable());
+            var provider = this.CreateServiceProvider(collection);
+
+            // Act
+            var service = provider.GetService<TestDisposable>();
+            ((IDisposable)provider).Dispose();
+
+            // Assert
+            Assert.False(service.IsDisposed);
+        }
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
```diff
@@ -346,6 +346,52 @@ public void ProviderScopeDisposeThrowsWhenOnlyDisposeAsyncImplemented()
                 exception.Message);
         }
 
+        [Fact]
+        public void SingletonServiceCreatedFromFactoryIsDisposedWhenContainerIsDisposed()
+        {
+            // Arrange
+            var collection = new TestServiceCollection();
+            collection.AddSingleton(_ => new TestDisposable());
+            var provider = this.CreateServiceProvider(collection);
```
davidfowl: Style nit: We don't use `this.`

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
```diff
@@ -346,6 +346,52 @@ public void ProviderScopeDisposeThrowsWhenOnlyDisposeAsyncImplemented()
...
+        {
+            public bool IsDisposed { get; private set; }
+
+            public void Dispose()
+            {
+                this.IsDisposed = true;
+            }
+        }
+
+        internal class TestServiceCollection : List<ServiceDescriptor>, IServiceCollection
```
davidfowl: What's this for?

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
tomakita: I've removed `this` in https://github.com/tomakita/Extensions/commit/0672203deab9b0ac7ef9660b0ab1f4bd88dfe958.  Thanks for pointing this out.

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
tomakita: It was meant to be an implementation of `IServiceCollection` (inheriting from `List<ServiceDescriptor>` is an easy way of implementing `IServiceCollection` for this purpose) that we could use for our tests, but this comment has made me realize that there is a default implementation of `IServiceCollection` called [`ServiceCollection`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicecollection?view=aspnetcore-3.0) that can be used for exactly this purpose.  I feel silly now.

Thanks for pointing this out, too!

Unrelated, but also in https://github.com/tomakita/Extensions/commit/0672203deab9b0ac7ef9660b0ab1f4bd88dfe958: I've changed the accessibility of `TestDisposable` to private, since it doesn't need to be internal.

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
```diff
@@ -346,6 +346,48 @@ public void ProviderScopeDisposeThrowsWhenOnlyDisposeAsyncImplemented()
...
+            Assert.False(service.IsDisposed);
+        }
+
+        private class FakeDisposable : IDisposable
+        {
+            public bool IsDisposed { get; private set; }
+
+            public void Dispose()
+            {
+                this.IsDisposed = true;
```
davidfowl: remove `this.`

tomakita: Done in https://github.com/tomakita/Extensions/commit/93099d524ba9fd729db88ef5645b4c5891ee3525

tomakita: Done.

PR #2481, opened 2019-10-09
Unify OS version xunit attributes

Tratcher:
It turns out that me of last week thinks a lot like me of last year.
https://github.com/aspnet/Extensions/pull/2449 added OSMinVersionAttribute, but last year I added MinimumOSVersionAttribute that does almost the same thing. The main functional difference is that the old one didn't distinguish between win10 versions. The old one also did registry checks that are no longer needed.

At least I've improved as a developer since last year :grin:, this time there were test, more granular version checks, etc..

This PR merges the new attribute back into the old one with code improvements and applies the new tests to the old attribute.


Tratcher:
@anurse test infrastructure improvements that support 3.1 feature work.

anurse:
Infra-only is tell mode for 3.1 anyway, so I'm fine with this (once approved, green builds, etc.).

PR #2483, opened 2019-10-10
Compile using ref assemblies

JunTaoLuo:
https://github.com/aspnet/AspNetCore-Internal/issues/1773


src/Logging/Logging.Testing/src/Microsoft.Extensions.Logging.Testing.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Helpers for writing tests that use Microsoft.Extensions.Logging. Contains null implementations of the abstractions that do nothing, as well as test implementations that are observable.</Description>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
JunTaoLuo: This breaks .NET Framework tests. I was trying this to fix builds but this approach won't work. To summarize the issue:

1. xunit.extensibility.execution has a netstandard1.1 assembly called xunit.execution.dotnet.dll and a net452 assembly called xunit.execution.desktop.dll
2. We want to produce Microsoft.AspNetCore.Testing (MAT) reference assemblies to ensure all users of it reference only 3.0.0.0 ref assemblies.
3. MAT references xunit.extensibility.execution and only the netstandard2.0 reference assembly is generated. There's no ref assembly generated for net461.
4. Microsoft.Extensions.Logging.Testing (MELT) references MAT and can't build property for net461 since there's a transitive reference to xunit.execution.dotnet.dll via MAT's netstandard2.0 reference assembly. xunit.execution.dotnet.dll can't be resolved since MELT brings in xunit.execution.desktop.dll instead.

Is there a way to resolve this conflict @dougbu @nguerrera? Since there are no changes here I think we might be able to use the released 3.0.0 MELT? It's just a testing package anyway.

dougbu: Why does this project need to change? Why not generate a separate .NET 4.6.1 ref/ assembly?

JunTaoLuo: Ah I thought we can only generate ref assemblies for .NET Core. Turns out we can generate ref assemblies for .NET Framework as well. Looks like that works.

eng/targets/ResolveReferences.targets
```diff
@@ -74,16 +73,33 @@
       <IsSharedSource Condition="'%(IsSharedSource)' == '' AND $([System.String]::new('%(Identity)').EndsWith('.Sources'))">true</IsSharedSource>
     </Reference>
 
+    <!-- Packages which are implicitly defined by the .NET Core SDK. -->
     <_ImplicitPackageReference Include="@(PackageReference->WithMetadataValue('IsImplicitlyDefined', 'true'))" />
-    <_ExplicitPackageReference Include="@(PackageReference)" Exclude="@(_ImplicitPackageReference)" />
+    <!-- Capture a list of references which were set explicitly in the project. -->
+    <_AllowedExplicitPackageReference Include="@(PackageReference->WithMetadataValue('AllowExplicitReference', 'true'))" />
+    <_AllowedExplicitPackageReference Include="FSharp.Core" Condition="'$(MSBuildProjectExtension)' == '.fsproj'" />
```
wtgodbe: Do we know if there might be any other literal/explicit references like this one (now or in the future)?

eng/targets/ResolveReferences.targets
```diff
@@ -97,15 +113,42 @@
     <Reference Remove="@(_ProjectReferenceByAssemblyName)" />
   </ItemGroup>
 
-  <Target Name="ResolveCustomReferences" BeforeTargets="CollectPackageReferences;ResolveAssemblyReferencesDesignTime;ResolveAssemblyReferences" Condition=" '$(TargetFramework)' != '' AND '$(EnableCustomReferenceResolution)' == 'true'">
+  <!--
+    This target resolves remaining Referene items to Packages, if possible. If not, they are left as Reference items fo the SDK to resolve.
```
wtgodbe: nit, but `Referene` -> `Reference, & `fo` -> `for`

eng/targets/ResolveReferences.targets
```diff
@@ -74,16 +73,33 @@
       <IsSharedSource Condition="'%(IsSharedSource)' == '' AND $([System.String]::new('%(Identity)').EndsWith('.Sources'))">true</IsSharedSource>
     </Reference>
 
+    <!-- Packages which are implicitly defined by the .NET Core SDK. -->
     <_ImplicitPackageReference Include="@(PackageReference->WithMetadataValue('IsImplicitlyDefined', 'true'))" />
-    <_ExplicitPackageReference Include="@(PackageReference)" Exclude="@(_ImplicitPackageReference)" />
+    <!-- Capture a list of references which were set explicitly in the project. -->
+    <_AllowedExplicitPackageReference Include="@(PackageReference->WithMetadataValue('AllowExplicitReference', 'true'))" />
+    <_AllowedExplicitPackageReference Include="FSharp.Core" Condition="'$(MSBuildProjectExtension)' == '.fsproj'" />
+    <_ExplicitPackageReference Include="@(PackageReference)" Exclude="@(_ImplicitPackageReference);@(_AllowedExplicitPackageReference)" />
```
wtgodbe: Could you add some comments explaining what these items are/why they're needed? `_ExplicitPackageReference`, `_UnusedProjectReferenceProvider`, `_CompilationOnlyReference`, `_InvalidReferenceToNonSharedFxAssembly`

eng/targets/ResolveReferences.targets
JunTaoLuo: Actually I think this one wouldn't be needed. I copied changes from the AspNetCore but I don't think this one applies.

src/Shared/src/ProviderAliasUtilities/Microsoft.Extensions.ProviderAliasUtilities.Sources.csproj
```diff
@@ -0,0 +1,7 @@
+<Project>
```
dougbu: This is fine but only if it is _not_ creating a new (useless) package

eng/targets/ResolveReferences.targets
```diff
@@ -47,19 +47,18 @@
       Projects should only use the project references instead of baseline package references when:
```
dougbu: Changes here
1. copy over latest fixes from aspnet/AspNetCore
2. does new stuff to handle the new ref/-dependent compilation requirement
3. addresses @wtgodbe's feedback

Please categorize because I'd like to focus on (2).

eng/targets/ResolveReferences.targets
```diff
@@ -47,19 +47,18 @@
...
-
       We don't use project references between components in servicing builds between compontents to preserve the baseline as much as possible.
     -->
     <UseProjectReferences Condition=" '$(UseProjectReferences)' == '' AND '$(IsServicingBuild)' != 'true' ">true</UseProjectReferences>
     <UseProjectReferences Condition=" '$(UseProjectReferences)' == '' AND '$(IsImplementationProject)' != 'true' ">true</UseProjectReferences>
     <UseProjectReferences Condition=" '$(UseProjectReferences)' == '' ">false</UseProjectReferences>
 
     <!-- Specifies if reference assembly projects should be resolved for Reference items -->
-    <ReferenceReferenceAssemblies Condition=" '$(ReferenceReferenceAssemblies)' == '' AND '$(IsReferenceAssemblyProject)' == 'true'">true</ReferenceReferenceAssemblies>
+    <ReferenceReferenceAssemblies Condition=" '$(ReferenceReferenceAssemblies)' == '' AND ('$(IsReferenceAssemblyProject)' == 'true' OR '$(IsImplementationAssemblyProject)' == 'true')">true</ReferenceReferenceAssemblies>
```
JunTaoLuo: @dougbu (2)

eng/targets/ResolveReferences.targets
```diff
@@ -47,19 +47,18 @@
...
     <ReferenceReferenceAssemblies Condition=" '$(ReferenceReferenceAssemblies)' == '' ">false</ReferenceReferenceAssemblies>
 
     <!-- Specifies if implementation assembly projects should be resolved for Reference items -->
-    <ReferenceImplementationAssemblies Condition=" '$(ReferenceImplementationAssemblies)' == '' AND '$(IsReferenceAssemblyProject)' != 'true'">true</ReferenceImplementationAssemblies>
+    <ReferenceImplementationAssemblies Condition=" '$(ReferenceImplementationAssemblies)' == '' AND '$(IsReferenceAssemblyProject)' != 'true' AND '$(IsImplementationAssemblyProject)' != 'true'">true</ReferenceImplementationAssemblies>
```
JunTaoLuo: @dougbu (2)

src/Shared/src/ProviderAliasUtilities/Microsoft.Extensions.ProviderAliasUtilities.Sources.csproj
JunTaoLuo: It is being produced. Will remove.

Issue #2431, opened 2019-10-01
Aspnet Core 3.0 DI Bug when constructor value type with default value


cyrusdargahi:
### Description

When a class is registered with service collection as transient and that class constructor has a parameter with a value type, but that parameter has a default value, then when it is resolved in runtime, the first two times it works as expected and the parameter is set to the default value, but on subsequent service resolutions the parameter will be given "bizarre" values that are not declare anywhere within the application code.

I have created a repo with an example and explanation https://github.com/cyrusdargahi/aspnetcore3DIBug

### To Reproduce
Build and run the application. On first and the second load in the browser you should see the output 0.01.

### Error
On any page reloads after the (exactly) first two, weird thing start to happen. The output will now be a bizarre number taken from somewhere that isn't defined anywhere in the application code.

### Expected behavior
Expected result is the page should display the constant 0.01 result on any page (re)load.

davidfowl:
Value types are unsupported in the DI container.

SplitThePotCyrus:
I know that @davidfowl but this is a breaking change from the behavior since asp.net core 2.2 (or even further back). We have a production solution that had similar code and we got very unpleasant result after migrating to asp.net core 3.0 (almost cost us a lot of money due to this _small_ bug/breaking change)

In this case the constructor has a default value so no value need to be injected. But somehow the DI instead injects very strange values, after the second time it gets resolved. 

At least it should not behave as it does right now. Either it should throw an exception or it should work consistently (not only the first two times that instance is resolved). Also this should be documented as a breaking change, if no code action is taken.

cyrusdargahi:
I know that @davidfowl but this is a breaking change from the behavior since asp.net core 2.2 (or even further back). We have a production solution that had similar code and we got very unpleasant result after migrating to asp.net core 3.0 (almost cost us a lot of money due to this small bug/breaking change)

In this case the constructor has a default value so no value need to be injected. But somehow the DI instead injects very strange values, after the second time it gets resolved.

At least it should not behave as it does right now. Either it should throw an exception or it should work consistently (not only the first two times that instance is resolved). Also this should be documented as a breaking change, if no code action is taken.

davidfowl:
> At least it should not behave as it does right now. Either it should throw an exception or it should work consistently (not only the first two times that instance is resolved). Also this should be documented as a breaking change, if no code action is taken.

Agreed.

brandondahler:
[...]

PR #2501, opened 2019-10-12
DI: Fix IL builder's undefined behavior for defaulted value types.

brandondahler:
If a service contains a constructor that has value types that are assigned default values, ILEmitResolverBuilder will produce IL with undefined behavior.  At runtime, this causes the value to sometimes resolve as the correct value and other times resolve as a random (uninitialized) value.

Addresses #2431

### DIFF
--- a/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
+++ b/src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
@@ -121,11 +121,17 @@ namespace Microsoft.Extensions.DependencyInjection.Tests
         {
             var serviceCollection = new ServiceCollection();
             serviceCollection.AddSingleton<IFakeService, FakeService>();
-            serviceCollection.AddSingleton<ClassWithServiceAndOptionalArgsCtorWithStructs>();
+            serviceCollection.AddTransient<ClassWithServiceAndOptionalArgsCtorWithStructs>();
 
             var provider = CreateServiceProvider(serviceCollection);
-            var service = provider.GetService<ClassWithServiceAndOptionalArgsCtorWithStructs>();
-            Assert.NotNull(service);
+
+            for (int i = 0; i < 100; i++)
+            {
+                var service = provider.GetService<ClassWithServiceAndOptionalArgsCtorWithStructs>();
+
+                Assert.NotNull(service);
+                Assert.Equal(default, service.DateTimeDefault);
+            }
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
```diff
@@ -121,11 +121,17 @@ public void ResolvesServiceMixedServiceAndOptionalStructConstructorArguments()
         {
             var serviceCollection = new ServiceCollection();
             serviceCollection.AddSingleton<IFakeService, FakeService>();
-            serviceCollection.AddSingleton<ClassWithServiceAndOptionalArgsCtorWithStructs>();
+            serviceCollection.AddTransient<ClassWithServiceAndOptionalArgsCtorWithStructs>();
```
davidfowl: Why not add a new test?

brandondahler: Rationale was that I wasn’t testing a new behavior but a quirk in the existing implementation. In this context swapping to transient from singleton doesn’t affect what primary behavior we’re testing while enabling me to validate the quirk is fixed. 

I can easily fork the test to a separate one if you’d prefer.

pakrym: Yeah, fork it into a separate test and I wonder if we should verify all the parameter values.

pakrym:
Change looks good to me, let's fix the test.

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
```diff
@@ -128,6 +128,33 @@ public void ResolvesServiceMixedServiceAndOptionalStructConstructorArguments()
...
+        [Fact]
+        public void ResolvesServiceMixedServiceAndOptionalStructConstructorArgumentsReliably()
+        {
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton<IFakeService, FakeService>();
+            serviceCollection.AddTransient<ClassWithServiceAndOptionalArgsCtorWithStructs>();
+
+            var provider = CreateServiceProvider(serviceCollection);
+
+            for (int i = 0; i < 100; i++)
```
pakrym: Can you please also add a comment about why this loop is here

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
brandondahler: Done

src/DependencyInjection/DI/test/ServiceProviderContainerTests.cs
brandondahler: Added

PR #2503, opened 2019-10-13
Make WASM packages stay preview releases in 3.1.

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/14623


dougbu:
Because the Mono.WebAssembly.Interop package was released as 3.0.0 **GA**, this change does not seem useful. Does it really meet the 3.1 bar?

pranavkm:
Yes, the 3.1.0 version will RTM as part of blazor-wasm.

dougbu:
Why is this change needed then? And, if it's needed and approved, why not do it as part of stabilizing versions for 3.1.0 GA?

mkArtakMSFT:
> Because the Mono.WebAssembly.Interop package was released as 3.0.0 **GA**, this change does not seem useful. Does it really meet the 3.1 bar?

That was a mistake and because there is no scenario where customers would take direct dependency on that package, I've unlisted 3.0 version of it. As @pranavkm said, we'll RTM 3.1.0 version of it with Blazor WASM release.

PR #2543, opened 2019-10-21
Provide new EventId implicitly with tuple syntax

### LINKED ISSUE #2542: More convenient EventId via tuple syntax
### Is your feature request related to a problem? Please describe.
It's best to have unique, named EventId values for each log location but the number of `new EventId` statements grows quickly.

### Describe the solution you'd like
```csharp
_logger.LogWarning((42, "SomethingFailed"), "Something failed because {Reasons}",
    result.Error.Reason);
```
is the same as
```csharp
_logger.LogWarning(new EventId(42, "SomethingFailed"), "Something failed because {Reasons}",
    result.Error.Reason);
```

### Describe alternatives you've considered
An EventId static method, but `EventId.Named(42, "SomethingFailed")` is not much different from `new EventId(42, "SomethingFailed")`

Additional log message overloads, but `_logger.LogWarning(42, "SomethingFailed", "Something failed because {Reasons}", error.Message)` would have the same method signature as a message that takes two string arguments. Even if you parsed the second and third argument to guess that the third was meant to be the format string - intellisense would still tell the user otherwise.

### Additional context
I'll send along a PR also if this seems like an idea to consider. It's purely additive, so the new c# tuple syntax isn't required. There is also a second idea which builds on this to make pre-declaring high-performance messages via `LoggerMessage.Define` much more practical.

lodejard:
An `(int, string)` tuple may be provided where an `EventId` parameter is expected.
- Encourages use of named events by reducing typing
- Compatible with EventId flavor of existing extension methods

Fixes #2542

### DIFF
--- a/src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
+++ b/src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
@@ -170,13 +170,13 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks
 
         internal static class EventIds
         {
-            public static readonly EventId HealthCheckProcessingBegin = new EventId(100, "HealthCheckProcessingBegin");
-            public static readonly EventId HealthCheckProcessingEnd = new EventId(101, "HealthCheckProcessingEnd");
+            public static readonly EventId HealthCheckProcessingBegin = (100, nameof(HealthCheckProcessingBegin));
+            public static readonly EventId HealthCheckProcessingEnd = (101, nameof(HealthCheckProcessingEnd));
 
-            public static readonly EventId HealthCheckBegin = new EventId(102, "HealthCheckBegin");
-            public static readonly EventId HealthCheckEnd = new EventId(103, "HealthCheckEnd");
-            public static readonly EventId HealthCheckError = new EventId(104, "HealthCheckError");
-            public static readonly EventId HealthCheckData = new EventId(105, "HealthCheckData");
+            public static readonly EventId HealthCheckBegin = (102, nameof(HealthCheckBegin));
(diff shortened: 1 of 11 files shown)


src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
```diff
@@ -170,13 +170,13 @@ private static void ValidateRegistrations(IEnumerable<HealthCheckRegistration> r
 
         internal static class EventIds
         {
-            public static readonly EventId HealthCheckProcessingBegin = new EventId(100, "HealthCheckProcessingBegin");
-            public static readonly EventId HealthCheckProcessingEnd = new EventId(101, "HealthCheckProcessingEnd");
+            public static readonly EventId HealthCheckProcessingBegin = (100, nameof(HealthCheckProcessingBegin));
```
Kahbazi: https://github.com/aspnet/AspNetCore/pull/11379#discussion_r295441810

lodejard: Oh, I see what you mean. Although if the shape of logging is considered a contract that shouldn't change - there's probably several details that might be changed accidentally as easily. Using string instead of nameof doesn't seem like a strict enforcement... 

Just thinking out loud, I wonder if there would be a convenient way to validate that logging contract hasn't changed in the unit tests?

JunTaoLuo:
Note that some sources edited in this PR has migrated to dotnet/aspnetcore. A new PR with the same changes for the affected sources should be opened in dotnet/aspnetcore. See https://github.com/dotnet/extensions/pull/2974 for details.

Issue #2568, opened 2019-10-28
Windows service ServiceBase access for pause/continue support


artiomchi:
### Is your feature request related to a problem? Please describe.
Currently there's a simple way to run a netcore3 app as a windows service using the `.UseWindowsService()` extension. The issue is that this extension uses it's own `ServiceBase` class, which is the only class that receives pause and continue events, shutdown events and custom service commands.

### Describe the solution you'd like
Add a second generic extension method that allows a custom class to be passed to `.UseWindowsService<>()`, which will be based on the existing `WindowsServiceLifetime`, retaining all functionality, but allowing extra customisation.

### Describe alternatives you've considered
I've thought of altering the `WindowsServiceLifetime` class to receive a `ICollection<ServiceBase>` from DI, and register them alongside itself in the `ServiceBase.Run()` call, but that will mean the `ApplicationLifetime.ApplicationStarted` will be called only after the `WindowsServiceLifetime` service has started, not when all services have.

Also, it seems to make sense to only have one service registered, and move "service" functionality to hosted service classes instead.

### Additional context
I've added an implementation to my fork, branched off `release/3.0`, since I wasn't sure which branch would be best to start the fix from: [feature/windows-service-custom-class](https://github.com/artiomchi/Extensions/tree/feature/windows-service-custom-class), with the fix being in this commit: [d9bafd6
](https://github.com/artiomchi/Extensions/commit/d9bafd63027ba352af6839b0f0003a8e504eb519)

If you let me know whether I should do it from `master`, or any other branch, I can make a PR for this.

Tratcher:
Can you give some code examples here showing the proposed usage?

Yes, we primarily work in master.

artiomchi:
I forgot to add, to use this new class, it'd have to be registered when setting up the host:
```csharp
Host.CreateDefaultBuilder(args)
    .UseWindowsService<CustomService>()
```

artiomchi:
Close, This should only be registered when running within the context of a web service. So the full solution becomes:

```csharp
Host.CreateDefaultBuilder(args)
    .UseWindowsService()
    .ConfigureServices((hostContext, services) =>
    {
        if (WindowsServiceHelpers.IsWindowsService())
            services.AddSingleton<IHostLifetime, CustomService>();
    })
```

I just thought to suggest a more streamlined option, as well as extend the documentation, so that making a windows service that can handle power/pause events is more obvious.

If this is considered unnecessary, then I'm happy to cancel the PR and close the issue

[...]

PR #2572, opened 2019-10-29
Adding support for custom Windows service lifetime class support

artiomchi:
Adds the ability to handle windows service lifetime events by
 - Allowing to register a subclass of `WindowsServiceLifetime` as host lifetime

Addresses #2568


artiomchi:
I've manually updated the reference assemblies, since I don't have the 3.1 SDK installed.

Let me know if that's not OK please, or if anyone could re-generate them using the msbuild task

anurse:
You should be able to accomplish the same thing with these two lines:

```csharp
hostBuilder
    .UseWindowsService(...)
    .ConfigureServices(s => s.AddSingleton<IHostLifetime, MyCustomLifetime>())
```

Closing as we don't think this adds enough value to make a public API change.

PR #2579, opened 2019-10-30
Enable SDL validation in rel/3.1

wtgodbe:
Part of https://github.com/aspnet/AspNetCore-Internal/issues/3293

@riarenas @JohnTortugo PTAL

CC @aspnet/build

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -297,3 +297,17 @@ stages:
       # Symbol validation isn't being very reliable lately. This should be enabled back
       # once this issue is resolved: https://github.com/dotnet/arcade/issues/2871
       enableSymbolValidation: false
+      # This is to enable SDL runs part of Post-Build Validation Stage
+      SDLValidationParameters:
+        enable: true
+        continueOnError: false
+        params: ' -SourceToolsList @("policheck","credscan")
+        -TsaInstanceURL $(_TsaInstanceURL)
+        -TsaProjectName $(_TsaProjectName)
+        -TsaNotificationEmail $(_TsaNotificationEmail)
+        -TsaCodebaseAdmin $(_TsaCodebaseAdmin)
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -296,4 +296,18 @@ stages:
...
       # once this issue is resolved: https://github.com/dotnet/arcade/issues/2871
-      enableSymbolValidation: false
\ No newline at end of file
+      enableSymbolValidation: false
+      # This is to enable SDL runs part of Post-Build Validation Stage
+      SDLValidationParameters:
+        enable: true
+        continueOnError: false
+        params: ' -SourceToolsList @("policheck","credscan")
+        -TsaInstanceURL $(_TsaInstanceURL)
```
riarenas: I  believe these variables for Arcade are coming from a variable group that is referenced here 
https://github.com/dotnet/arcade/blob/e2e223ceb1a140788fa8e48c6b90ad6bb5064419/eng/common-variables.yml#L31

riarenas:
@sunandabalu can probably review these better than Cesar or I can, but left some comments.

azure-pipelines.yml
wtgodbe: Looks like I need to add https://github.com/dotnet/winforms/blob/15eb866a9db9a5e9856ad8072e83acdd778770ee/azure-pipelines.yml#L17-L19

riarenas: Just make sure that all the values match. I think the expectation is that repos have their own variable groups so that say, aspnetcore bugs don't get filed under the winforms area path.

wtgodbe:
I validated that this works w/ https://dev.azure.com/dnceng/internal/_build/results?buildId=408162. @sunandabalu if everything else looks good I'll mass-merge this & the other equivalent PRs in the other ASP.Net repos

PR #2584, opened 2019-10-31
Update BlazorWASMPreReleaseVersionLabel to preview3

wtgodbe:
@mkArtakMSFT @pranavkm should this stay at preview2, or keep in line with preview3?

CC @mmitche


mkArtakMSFT:
This should be `preview3`! The change is correct. And it should stay so when 3.1 RTMs.

wtgodbe:
Noted, thanks @mkArtakMSFT!

PR #2595, opened 2019-11-03
Change CtorUsed to reflect the correct ctor types arguments

kiapanahi:
Summary of the changes
 - A small change in CtorUsed string property

The CtorUsed is being set to `"IFakeService, string, string"` in the `public ClassWithAmbiguousCtors(IFakeService service, string data1, int data2)` constructor but it should have been set to `"IFakeService, string, int"`. The PR addresses this small issue.


ericstj:
Closing since this code no longer lives in this repo.  If this is still of interest you can submit a PR to https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.DependencyInjection/tests/DI.Specification.Tests/Fakes/ClassWithAmbiguousCtors.cs.

ericstj:
cc @maryamariyan @eerhardt

PR #2621, opened 2019-11-09
chmod on activate.sh

rynowak:
Marks script as executable


rynowak:
nvm this was intentionally not executable

davidfowl:
ah you're supposed to dot include it

BrennanConroy:
`source activate.sh` on unix

PR #2626, opened 2019-11-10
Set ServiceName vis UseWindowsService

### LINKED ISSUE #2612: Allow setting the ServiceName property in `UseWindowsService` (via Options)
From discussions in https://github.com/aspnet/Extensions/issues/2138

We should allow a user to configure the service name through the `.UseWindowsService` extension method on `IHostBuilder`. This should probably be done through an options type, so the API change here would be:

```diff
namespace Microsoft.Extensions.Hosting
{
    public static class WindowsServiceLifetimeHostBuilderExtensions
    {
        public static IHostBuilder UseWindowsService(this IHostBuilder hostBuilder);
+++     public static IHostBuilder UseWindowsService(this IHostBuilder hostBuilder, Action<WindowsServiceLifetimeOptions> configure);

    }

+++ public class WindowsServiceLifetimeOptions
+++ {
+++     public string ServiceName { get; set; }
+++     // Possible other options...
+++ }
}
```

Kahbazi:
Fixes #2612


src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
```diff
@@ -26,6 +26,22 @@ public static class WindowsServiceLifetimeHostBuilderExtensions
         /// <param name="hostBuilder">The <see cref="IHostBuilder"/> to operate on.</param>
         /// <returns>The same instance of the <see cref="IHostBuilder"/> for chaining.</returns>
         public static IHostBuilder UseWindowsService(this IHostBuilder hostBuilder)
+        {
+            return UseWindowsService(hostBuilder, _ => { });
```
gfoidl: Maybe just pass `null` as action down, then check in the other overload for `!= null`?

src/Hosting/WindowsServices/src/WindowsServiceLifetimeOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.Extensions.Hosting
+
+{
+    public class WindowsServiceLifetimeOptions : HostOptions
+    {
+        /// <summary>
+        /// The name used to identify the service to the system.
+        /// </summary>
+        public string ServiceName { get; set; } = string.Empty;
+
```
gfoidl: Nit: remove empty blank line

src/Hosting/WindowsServices/src/WindowsServiceLifetimeHostBuilderExtensions.cs
anurse: We do this pattern a lot, so I'd suggest whatever we do in other places like this :). I *think* we do it this way (passing in an empty delegate), but I don't have a concrete example handy.

gfoidl: In a quick search I didn't find an example either. I prefer consistency too.

anurse: I think I'd keep it as an empty delegate. It seems to be the thing we generally do in these situations. There may be a super-small perf impact (calling an empty delegate vs a null check) but this is during start-up so I'm not super worried.

I'd rather just keep `null`s out of our code as much as possible, so this code as-is seems good to me.

src/Hosting/WindowsServices/src/WindowsServiceLifetimeOptions.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Extensions.Hosting
+
+{
+    public class WindowsServiceLifetimeOptions : HostOptions
```
Tratcher: I don't know how well this derivation will work. You did this to get ShutdownTimeout, right? But now you have two different options objects.

@HaoK How well does `Configure<TOptions>` work with derived types?

You might be better of injecting both options separately.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -14,9 +14,9 @@ public class WindowsServiceLifetime : ServiceBase, IHostLifetime
     {
         private readonly TaskCompletionSource<object> _delayStart = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
         private readonly ManualResetEventSlim _delayStop = new ManualResetEventSlim();
-        private readonly HostOptions _hostOptions;
+        private readonly WindowsServiceLifetimeOptions _options;
 
-        public WindowsServiceLifetime(IHostEnvironment environment, IHostApplicationLifetime applicationLifetime, ILoggerFactory loggerFactory, IOptions<HostOptions> optionsAccessor)
+        public WindowsServiceLifetime(IHostEnvironment environment, IHostApplicationLifetime applicationLifetime, ILoggerFactory loggerFactory, IOptions<WindowsServiceLifetimeOptions> optionsAccessor)
```
Tratcher: This requires a ref assembly update. See https://github.com/aspnet/Extensions/blob/master/docs/ReferenceAssemblies.md

src/Hosting/WindowsServices/src/WindowsServiceLifetimeOptions.cs
HaoK: Configure doesn't look at inheritance of types at all (other than what you get for free from the ctor), but only the Configures that are registered against the exact TOptions type are run

Tratcher: I was afraid of that. Ok, this model won't work. Split WindowsServiceLifetimeOptions and HostOptions and inject them separately.

anurse:
@Tratcher we don't have the PR labeller bot running here so just a brief reminder we should run this through API review before merging. I applied `api-suggestion`. Flip it to `api-ready-for-review` we're happy with the API.

src/Hosting/WindowsServices/ref/Microsoft.Extensions.Hosting.WindowsServices.netstandard2.0.cs
```diff
@@ -16,7 +22,7 @@ public static partial class WindowsServiceHelpers
     }
     public partial class WindowsServiceLifetime : System.ServiceProcess.ServiceBase, Microsoft.Extensions.Hosting.IHostLifetime
     {
-        public WindowsServiceLifetime(Microsoft.Extensions.Hosting.IHostEnvironment environment, Microsoft.Extensions.Hosting.IHostApplicationLifetime applicationLifetime, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Hosting.HostOptions> optionsAccessor) { }
+        public WindowsServiceLifetime(Microsoft.Extensions.Hosting.IHostEnvironment environment, Microsoft.Extensions.Hosting.IHostApplicationLifetime applicationLifetime, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Hosting.HostOptions> hostOptionsAccessor, Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Hosting.WindowsServiceLifetimeOptions> windowsServiceOptionsAccessor) { }
```
pranavkm: We can make this change without making this a breaking change. Just add an additional ctor that adds an `IOptions<WindowsServiceLifetimeOptions>` and have the other ctor pass in null.

pranavkm: Also let's leave the first parameter named `optionsAccessor`. Avoids needless confusion.

anurse:
> @tmds is there a similar concept we would want for SystemD?

My understanding is any kind of "service name" is entirely controlled in the config files, but I may be wrong.

tmds:
> My understanding is any kind of "service name" is entirely controlled in the config files.

That's correct. This is the `SyslogIdentifier`.
When not specified in the config file, systemd defaults to the app's filename. This works nicely with the shared apphost. Before the apphost, all apps logged under `dotnet` (since they were executed as `dotnet myassembly.dll`).

With systemd's structured logging API (not used by ASP.NET Core), the application can specify the syslog identifier.

src/Hosting/WindowsServices/ref/Microsoft.Extensions.Hosting.WindowsServices.netstandard2.0.cs
anurse: @Kahbazi can you make this change? Basically, leave the existing constructor and add this as a new constructor instead of a replacement. The existing constructor should remain unmodified (even parameter names), to avoid API breaks.

anurse:
@tmds Cool, good to know. Thanks!

src/Hosting/WindowsServices/ref/Microsoft.Extensions.Hosting.WindowsServices.netstandard2.0.cs
Kahbazi: @anurse I just did it in https://github.com/aspnet/Extensions/pull/2626/commits/5ce2efc27955ec6d95f607b64156fdd57ed859a9

src/Hosting/WindowsServices/src/WindowsServiceLifetimeOptions.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Extensions.Hosting
+
```
Tratcher:
```suggestion
```

Tratcher:
Thanks

PR #2627, opened 2019-11-10
Add event name for hosting logs

Kahbazi:

### DIFF
--- a/src/Hosting/Hosting/src/Internal/LoggerEventIds.cs
+++ b/src/Hosting/Hosting/src/Internal/LoggerEventIds.cs
@@ -1,17 +1,19 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.Logging;
+
 namespace Microsoft.Extensions.Hosting.Internal
 {
     internal static class LoggerEventIds
     {
-        public const int Starting = 1;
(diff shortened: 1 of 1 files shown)


src/Hosting/Hosting/src/Internal/LoggerEventIds.cs
```diff
@@ -1,17 +1,19 @@
...
     {
-        public const int Starting = 1;
-        public const int Started = 2;
-        public const int Stopping = 3;
-        public const int Stopped = 4;
-        public const int StoppedWithException = 5;
-        public const int ApplicationStartupException = 6;
-        public const int ApplicationStoppingException = 7;
-        public const int ApplicationStoppedException = 8;
+        public static EventId Starting = new EventId(1, "Starting");
```
gfoidl:
```suggestion
        public static readonly EventId Starting = new EventId(1, "Starting");
```
Others as well.

anurse:
Agree with @gfoidl 's suggestion. Otherwise looks good to me.

src/Hosting/Hosting/src/Internal/LoggerEventIds.cs
Kahbazi: oh yeah, I forgot that!

anurse:
@Tratcher any concerns here? I'm good to just merge it but would like a quick ACK from you (feel free to just merge it yourself)

Tratcher:
Thanks

PR #2630, opened 2019-11-11
Add startvs for each solution

Kahbazi:


anurse:
Seems fine to me, thanks!

PR #2634, opened 2019-11-11
update jsinterop NPM package ID for 5.0

anurse:
(Recommend viewing with Show Whitespace off)

A matching change will be needed in aspnet/AspNetCore I believe, so do not merge until everything is ready there and we're ready to flow through :).


anurse:
I'll look at kicking off this whole flow early next week. Don't want to do it today and have stuff broken over the weekend :).

anurse:
I believe that we can just merge this safely. AspNetCore references a specific package download URL so merging this should NOT break that process. I'll give it a go and will follow the internal build closely.

PR #2649, opened 2019-11-12
Adding Optional Serialization / Deserialization Capability

khaledobaid:
Adding two functions (Serialize / Deserialize) to support processing data before and after storing inside redis. I implemented this solution in my project as saving byte[] binary data could cause issues related to the replication. while this could binary-unsafe issue related to the replication module.

Usage : This is very simple example of using Base64 encoding for the binary data used in Startup.cs

```
services.AddStackExchangeRedisCache(options =>
            {
                options.Configuration = "127.0.0.1";
                options.InstanceName = "redisdb";
                 // Optional functions
                options.Serialize = (input) => { return Convert.ToBase64String(input); };
                options.Deserialize = (input) => { return Convert.FromBase64String(input); };
            });

### DIFF
--- a/src/Caching/StackExchangeRedis/src/RedisCache.cs
+++ b/src/Caching/StackExchangeRedis/src/RedisCache.cs
@@ -95,13 +95,17 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
 
             var absoluteExpiration = GetAbsoluteExpiration(creationTime, options);
 
+            RedisValue updatedValue = value;
+            if (_options.Serialize != null)
+                updatedValue = _options.Serialize.Invoke(value);
+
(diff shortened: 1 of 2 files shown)


khaledobaid:
It is giving 
`Generated code is not up to date. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)`

Can you please let me know what i can do ?

anurse:
This is a relatively significant feature request. Can you open a new issue to discuss the underlying requirement you have so we can discuss what might be the right approach?

In general, we need to discuss the requirements and design of any new API before considering a PR, so I'm going to close this for now. Please feel free to open an issue and once we land on a decision we can consider a PR.

PR #2679, opened 2019-11-18
Redo #2677 to get latest Arcade without removing extra files

dougbu:


dougbu:
Wow, an actual reason to use `rebase and merge` with multiple commits 😺

PR #2683, opened 2019-11-19
Update branding to 3.0.2

JunTaoLuo:


JunTaoLuo:
@dougbu yes. Since you mentioned that change should be tell mode but this ones isn't.

PR #2684, opened 2019-11-19
Use nameof

Youssef1313:


JunTaoLuo:
Note that the sources edited in this PR has migrated to dotnet/aspnetcore. This PR should closed and a new PR with the same changes should be opened in dotnet/aspnetcore.

PR #2695, opened 2019-11-21
Write flaky test logs to a Flaky directory

JunTaoLuo:
A quick fix to parts of https://github.com/aspnet/AspNetCore/issues/17243. The changes are not exactly extensible to anything other than `Flaky` attributes but it should be sufficient for now.

### DIFF
--- a/src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
+++ b/src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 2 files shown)


BrennanConroy:
I don't see anything different in the artifacts

src/Logging/Logging.Testing/src/LoggedTest/LoggedTestBase.cs
```diff
@@ -63,12 +64,17 @@ public virtual void Initialize(TestContext context, MethodInfo methodInfo, objec
                 var logLevelAttribute = methodInfo.GetCustomAttribute<LogLevelAttribute>()
                                         ?? methodInfo.DeclaringType.GetCustomAttribute<LogLevelAttribute>()
                                         ?? methodInfo.DeclaringType.Assembly.GetCustomAttribute<LogLevelAttribute>();
+                var flakyAttribute = methodInfo.GetCustomAttribute<FlakyAttribute>()
+                                        ?? methodInfo.DeclaringType.GetCustomAttribute<FlakyAttribute>()
+                                        ?? methodInfo.DeclaringType.Assembly.GetCustomAttribute<FlakyAttribute>();
```
dougbu: Curious: Doesn't the generic method have an overload supporting `inherit: true`?

JunTaoLuo:
Need to go back to the drawing board to resolve the underlying issues.

JunTaoLuo:
I'm leaning towards just having an option for not deleting existing logs and reasonable defaults. We'll retain logs on CI and delete on local runs. PR incoming.

Issue #2699, opened 2019-11-22
IHostApplicationLifetime not notified of shutdown


peval27:
### Describe the bug
When running the application as a Windows Service, if I shutdown the PC, the app is not notified of the shutdown or of the stop. IHostApplicationLifetime.ApplicationStopping or ApplicationStopped are not fired (and it's where I log about app stopping/stopped)

### To Reproduce
1. Using version '3.0.1' of package 'Microsoft.Extensions.Hosting.WindowsServices'
2. Create a Windows Service following instructions [here ](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-3.0&tabs=visual-studio). In particular make sure you call .UseWindowsService

### Expected behavior
App notified of system shutdown, so it can stop in a graceful manner.

peval27:
I had a look at the source code and it looks like WindowsServiceLifetime.cs doesn't override OnShutdown

peval27:
I can confirm the bug, since I've downloaded the source code and added 
```
protected override void OnShutdown()
{
     ApplicationLifetime.StopApplication();
     base.OnShutdown();                
}
```

and the app is now notified as expected during machine shutdown/restart.

anurse:
@peval27 that code seems good to us. Would be you up for turning it in to a PR?

Tratcher:
From the docs it sounds like you'd have to set CanShutdown as well?
https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicebase.onshutdown?redirectedfrom=MSDN&view=netcore-3.0#System_ServiceProcess_ServiceBase_OnShutdown

peval27:
> 
> 
> From the docs it sounds like you'd have to set CanShutdown as well?
> https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicebase.onshutdown?redirectedfrom=MSDN&view=netcore-3.0#System_ServiceProcess_ServiceBase_OnShutdown

not sure why it's being called then. happy to add to the constructor.

peval27:
> 
> 
> @peval27 that code seems good to us. Would be you up for turning it in to a PR?

will give it a go!

PR #2705, opened 2019-11-22
Handle operating system shutdown by overriding OnShutdown()

peval27:
Summary of the changes
 - overridden ServiceBase.OnShutdown() to get notified on OS shutdown

Addresses #2699

### DIFF
--- a/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
+++ b/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
                 throw new ArgumentNullException(nameof(optionsAccessor));
             }
             _hostOptions = optionsAccessor.Value;
+            CanShutdown = true;
         }
 
         private IHostApplicationLifetime ApplicationLifetime { get; }
@@ -89,6 +90,12 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
             // Wait for the host to shutdown before marking service as stopped.
             _delayStop.Wait(_hostOptions.ShutdownTimeout);
             base.OnStop();
+        }
+        
+        protected override void OnShutdown()
+        {
+             ApplicationLifetime.StopApplication();
+             base.OnShutdown();                
         }
 
         protected override void Dispose(bool disposing)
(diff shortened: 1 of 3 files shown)


anurse:
I was about to ask "can you add a test" but then I realized that would be **very** tricky ;).

peval27:
Yeah I don't see too much existing unit testing here

peval27:
BTW I saw you added milestone 5.0.0. what's the ETA? The released one is 3, so I assume quite a long time here. Wouldn't be possible to squash his big fix in a maintenance 3.0.x

anurse:
> Yeah I don't see too much existing unit testing here

Well, and testing something that requires shutting down the machine doesn't really lend itself to an automated test ;).

The full roadmap can be found here: https://github.com/dotnet/core/blob/master/roadmap.md

> BTW I saw you added milestone 5.0.0. what's the ETA? The released one is 3, so I assume quite a long time here. Wouldn't be possible to squash his big fix in a maintenance 3.0.x

5.0 is the next release after 3.1 (which is locked down for release in December), we're skipping 4.x to avoid conflicts with .NET Framework. The ETA is November 2020.

We don't generally add functionality in a maintenance release. We never provided support for the `OnShutdown` event in the past so this isn't really a regression. Can you provide any details on what scenarios are blocked in your environment? Given that this can be worked around by registering your own version of `WindowsServiceLifetime` in the container (based on this fixed code), I think this wouldn't meet the servicing bar.

src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs
```diff
@@ -90,6 +91,12 @@ protected override void OnStop()
             _delayStop.Wait(_hostOptions.ShutdownTimeout);
             base.OnStop();
         }
+        
+        protected override void OnShutdown()
+        {
+             ApplicationLifetime.StopApplication();
```
Tratcher: If OnShutdown blocks will that delay shutdown? Does the application have time to actually shut down before the OS kills it? Compare to OnStop.

peval27: Good point. will fix

peval27: good point, will fix it.

Tratcher:
Thanks

Issue #2077, opened 2019-07-24
Adding a typed HTTP client twice, throws exception


leastprivilege:
**EDIT by @rynowak** - see below for shiproom template.
----

Calling

`services.AddHttpClient<SomeType>()`

more than once throws an exception in 3.0 preview 7 - this used to work in 2.x.

Is that the intended behavior?

rynowak:
This is intended to throw in 3.0 and the reason is that we've gotten lots and lots of bug reports where users would do things like this:

```
services.AddHttpClient<Foo.SomeType>().ConfigureSomething(...);
services.AddHttpClient<Bar.SomeType>().ConfigureSomethingElse(...);
```

These will result in all of the configuration being applied to a single *logical client* because http client factory uses named options based on the type name (not the fully-qualified type name).

It would be possible for us to make this work in the future if necessary. It's clear that how this actually works is different from what users believe. The change in 3.0 was made to make this problem clear when it occurs.

leastprivilege:
So what's the recommendation then - this pattern seems common:

* a services.AddXXX extensions methods adds a bunch of services plus some default (typed) http clients
* this returns a builder
* the builder has some helpers to override the standard http clients

This is at least how we did it in IdentityServer. I am open to changing that.

leastprivilege:
..oh and what's the reason to not use the fully qualified type name?

rynowak:
> 
> 
> So what's the recommendation then - this pattern seems common:
> 
>     * a services.AddXXX extensions methods adds a bunch of services plus some default (typed) http clients
> 
>     * this returns a builder
> 
>     * the builder has some helpers to override the standard http clients
> 
> 
> This is at least how we did it in IdentityServer. I am open to changing that.

This hasn't come up before so I don't have a ready-made solution for it. It makes sense.

ImoutoChan:
What is expected behaviour when I do something like this:

```
services.AddHttpClient<ISomeInterface, SomeType>();
services.AddHttpClient<ISomeInterface, OtherType>();
services.AddHttpClient<ISomeInterface, OneMoreType>();
```

and then consume them with `IEnumarable<ISomeInterface>` in some constructor? This scenario worked fine before 3.0.

I found workaround and made it work like this:

```
services.AddHttpClient<SomeType>();
services.AddHttpClient<OtherType>();
services.AddHttpClient<OneMoreType>();

services.AddTransient<ISomeInterface>(provider => provider.GetRequiredService<SomeType>());
[...]

PR #2710, opened 2019-11-24
Fixes for some servicing-proposed HttpClient Factory issues

rynowak:
#### Description

In 3.0 we added validation to HttpClient Factory's registration code path to prevent some common mistakes. This new validation code blocks some valid use cases that we didn't know about.

#### Customer Impact

aspnet/AspNetCore#13346 Attempting to call `AddTypedClient<T>` twice on the same builder with different types throws an exception. This is a valuable shorthand way of expressing a configuration once and then binding it to multiple types.

#2077 Attempting to call `AddHttpClient<T>` twice with the same `T` throws an exception. This pattern is used when both library code and user code want to collaborate on configuration of a client type.
		
#### Regression?

Yes, this is a regression from 2.2
		
#### Risk

Low. This add special casing to our validation code path to allow more patterns. The impact of this is that cases that are explicitly blocked by an exception in 3.0 (but were allowed in 2.2) will be allowed again.

### DIFF
--- a/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
+++ b/src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
@@ -532,7 +532,12 @@ namespace Microsoft.Extensions.DependencyInjection
             Debug.Assert(registry != null);
 
             // Check for same type registered twice. This can't work because typed clients have to be unique for DI to function.
-            if (registry.TypedClientRegistrations.TryGetValue(type, out var otherName))
+            if (registry.TypedClientRegistrations.TryGetValue(type, out var otherName) &&
+
+                // Allow duplicate registrations with the same name. This is usually someone calling "AddHttpClient" once
+                // as part of a library, and the consumer of that library doing the same thing to add further configuration.
+                // See: https://github.com/aspnet/Extensions/issues/2077
+                !string.Equals(name, otherName, StringComparison.Ordinal))
             {
                 var message =
                     $"The HttpClient factory already has a registered client with the type '{type.FullName}'. " +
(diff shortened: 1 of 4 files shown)


bladefist:
Looks like this didn't make it into 3.1, is it coming soon?

VinceEeckhout:
Any news when 3.1.1 might be released?

mkArtakMSFT:
These were approved for 3.1.2, so updating the milestone accordingly /cc @jamshedd

johnkors:
Will the fix also work for the combo of typed+named clients?

```csharp
services.AddHttpClient<MyHttpClient>("client1", config.GetSection("client1config"));
services.AddHttpClient<MyHttpClient>("client2", config.GetSection("client2config"));
services.AddHttpClient<MyHttpClient>("client3", config.GetSection("client3config"));

var httpClient = _clientFactory.CreateClient("client1"); // MyHttpClient using client1config
```

VinceEeckhout:
Nowpe... I already tried this out myself. Seems like .net core 3 only
relies on the type that is provided.

lwansbrough:
Proposal: Add validation to the validation to prevent mistakes.

vivmishra:
Moved to Mar as per Tactics. Will need to be explicitly approved for Feb, if required.

Pilchie:
Pulling back into 3.1.2.

Pilchie:
@mkArtakMSFT @rynowak - can we get someone to review this and get it merged?

dougbu:
Has this been reviewed offline? If it's ready, please someone to sign off then get it in before 4PM today 😃

johnkors:
Is there a preview-nuget we can install to test this?

kjkrum:
Has this change been released? Calling `AddTypedClient<T>` more than once on the same builder works in a debug build, but the same project crashes as described [here](https://github.com/dotnet/aspnetcore/issues/13346) when it's deployed in production.

I'm using Microsoft.Extensions.Hosting 3.1.7, if that's the relevant package.

pranavkm:
**Hi, it looks like you are posting on a closed issue/PR/commit!**

We're very likely to lose track of your bug/feedback/question unless you:
1. Open a new issue
2. Explain very clearly what you need help with
3. If you think you have found a bug, include detailed repro steps so that we can investigate the problem

Thanks!

Issue #1121, opened 2019-02-15
Execution time reported for HTTP request by logger is lower than in reality


ilya-spv:
### Describe the bug

I make an HTTP requests to REST API that returns quite a lot of data. I have now used HttpClientFactory from .Net Core to get the data and it works fine. But what has puzzled me is a discrepancy between my timing and internal timing that comes from the library itself.

Here is an example message from the library:

End processing HTTP request after 130311.0094ms - OK

The thing is: it really takes 20-30 more seconds to really download all the data even though the library has written it's all finished (To execute `client.GetAsync()` method).

I assume what happens here is that library is very eager to report right after the header is received without waiting for the body to get downloaded.

This really sound like a bug, but I am not sure if it is. Maybe someone has a better explanation on why this is happening?

### To Reproduce
1. Using version '2.2.0.' of package 'Microsoft.Extensions.Http'
2. Run this code: Register HttpClient with HttpClientFactory to get logging, make a call that takes a lot of time and returns lots of data ('client.GetAsync()') and check how much time does it take
3. The time it takes is significantly higher than what the library itself reports in the log statement

### Expected behavior
Library should log the end request when it is really finished, that is all the data is received.

### Additional context
I am not 100% sure this is a bug, but would like at least to have som explanation on why this is happening then.

P.S. Here is an example of my code:

```
Stopwatch watch = new Stopwatch();
watch.Start();
var response = await Client.GetAsync(url);
watch.Stop();
Console.WriteLine($"Elapsed time {watch.ElapsedMilliseconds} ms");
```
Elapsed time in my log is way higher than in the system message

P.P.S I have tried running with HttpCompletionOption.ResponseContentRead but it really did not change the outcome, it's still a big difference.

rynowak:
That's interesting, do you have ability to profile this and see where the time is spent?

ilya-git:
I can try, but I assume the reporting happens right after the header is received, and it takes some time to receive everything. But I don't really know, and probably profiling won't help either since it's not CPU usage

rynowak:
Most profilers have an option to profile 'thread time' instead of 'cpu time'. In particular I know that the timeline mode of JetBrains `dotTrace` supports this sort of thing, and they have a very generous trial period.

If you're to provide some data, or a simple repro I'd be happy to investigate this. 

One other question - are you running in an environment with a sync context? Does it make a difference if you do `await Client.GetAsync(url).ConfigureAwait(false)`?

[...]

PR #2718, opened 2019-11-26
Make it clear we're processing response HEADERS

rynowak:
Fixes: #1121

Just a small tweak to logging to make it clear that this is the elapsed
time to get the response headers.


src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandlerBuilderFilter.cs
```diff
@@ -51,4 +51,4 @@ public Action<HttpMessageHandlerBuilder> Configure(Action<HttpMessageHandlerBuil
             };
         }
     }
-}
```
rynowak: VS did this. Must be important.

JamesNK: I will allow it.

I only complain about whitespace in code. I'm agnostic to ending files with blank line or not 😝

PR #2726, opened 2019-11-27
Specify user secret directory with environment variable DOTNET_USER_SECRETS_DIR

mgeramb:
Provide the possibilty to specify the user secret directory with the environment variable DOTNET_USER_SECRETS_DIR. 
This can be used for development teams to store the secrets in a shared folder and protect them with the windows security.

### DIFF
--- a/src/Configuration/Config.UserSecrets/src/PathHelper.cs
+++ b/src/Configuration/Config.UserSecrets/src/PathHelper.cs
@@ -42,9 +42,13 @@ namespace Microsoft.Extensions.Configuration.UserSecrets
 
             const string userSecretsFallbackDir = "DOTNET_USER_SECRETS_FALLBACK_DIR";
 
+            var root = Environment.GetEnvironmentVariable("DOTNET_USER_SECRETS_DIR"); // Allow setting a user secret directory with an environment variable. This can be used for development teams to share secrets projected by windows security
+            if (!string.IsNullOrEmpty(root))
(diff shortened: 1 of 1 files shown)


Tratcher:
Why not use one of the normal config providers in this scenario and read the environment variable in Startup.cs?

That said, is an environment variable granular enough when you're working on multiple projects?

mgeramb:
> Why not use one of the normal config providers in this scenario and read the environment variable in Startup.cs?

Because then the tool support with the `dotnet user-secrets` will not work

> That said, is an environment variable granular enough when you're working on multiple projects?

Good question, maybe it makes sense to check environment variables in this order:

- `DOTNET_USER_SECRETS_DIR `
- `DOTNET_USER_SECRETS_DIR_<USER_SECRET_ID>`
- `APPDATA`
- …

But the question is, how the user secret id should be mapped to the environment variable, because not all characters are allowed here. E.g. in a linux bash, only charasters, digits and underscores can be used.

Tratcher:
Since this is a feature request that requires additional design, please open a corresponding issue.

anurse:
Closing this *for now*. As @Tratcher said, we really need to talk about a way to design this properly, based on the underlying problem (sharing user secrets). There are a lot of concerns we need to review before we could make a change to support that scenario and just adding support for an env var here might not be the right approach.

Please do file an issue and we can discuss the underlying request!

mgeramb:
@Tratcher, @anurse Done: https://github.com/aspnet/Extensions/issues/2739

PR #2734, opened 2019-11-29
IsSystemdService: make check stricter to avoid false positives

tmds:
Checking the INVOCATION_ID envvar is causing false positives.
Instead, we now return true only when the direct parent process is 'systemd'.

Fixes https://github.com/aspnet/Extensions/issues/2525

cc @Tragetaschen @anurse @onyxmaster @Tratcher

### DIFF
--- a/src/Hosting/Systemd/src/SystemdHelpers.cs
+++ b/src/Hosting/Systemd/src/SystemdHelpers.cs
@@ -14,8 +14,6 @@ namespace Microsoft.Extensions.Hosting.Systemd
     /// </summary>
     public static class SystemdHelpers
     {
-        private const string INVOCATION_ID = "INVOCATION_ID";
-
         private static bool? _isSystemdService;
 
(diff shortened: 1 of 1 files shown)


anurse:
Technically breaking so we should run this by the API/Breaking Change review group (cc @aspnet/asp-net-api-reviews). I think it's warranted though (and the set of people it would break is pretty small: scenarios where `INVOCATION_ID` is set, the direct parent is **not** systemd, **and** the desired behavior is to act as a systemd service).

halter73:
Thanks @tmds! @anurse do you think we should backport this?

anurse:
The fact that it is *technically* a breaking change makes me a little nervous. But it is customer-reported and has no workaround. We also strongly believe there is no *intended* scenario that would be broken.

Let's take it to Tactics. @halter73 could you open a backport PR? I can fill in the justification details.

Issue #814, opened 2018-12-19
Add an overload with a Func<IServiceProvider, T> to AddHttpClient<TClient, TImplementation>


andre-ss6:
We have a multi-tenant web app that calls to our multi-tenant Web Api. I'm changing the api client used in the web app to use IHttpClientFactory, using typed clients. However, as it's a multi-tenant software, we have logic to call to different URLs depending on where the request is coming from. Something like this:

```csharp
public class MyApiClient : IMyApiClient
{
   public MyApiClient(HttpClient client, string baseUrl) {}
}
```

I used to do DI like this before:

```csharp
services.AddScoped<IMyApiClient>(sp => 
{
   // decide which url to use in the client
   return new MyApiClient(url);
}
```

However, now with `.AddHttpClient`, since there's no overload that accepts a `Func<IServiceCollection, T>`. there's no way I can do that anymore, and I'm left with three non-optimal solutions:

1) Change to use named clients instead of typed clients. However, that's not good not only because well, it's named clients, but also because I already have plenty of dependencies to `IMyApiClient` in a lot of places.

2) Copy the source code from `.AddHttpClient` and do the base URL selection just like before.

3) Create a whole new class just to represent the API base URL so I can register it on the IoC container and ask for that instead of a string on the `MyApiClient` constructor.

Note that I can't just set the HttpClient base URL using the `Action<HttpClient> configureClient` overload for two reasons:

1) In order to select the base URL, I need another dependency. Thus, I would need an instance of `IServiceProvider`.
2) Even if I could change my logic to not need any dependency when selecting the base URL, it still wouldn't work, because the base URL can't be changed after the HttpClient has executed a request.

## EDIT: 
I ended up going with option 2. This is what I came up with:

```csharp
services.AddHttpClient("MyApiClient")
    .AddTransientHttpErrorPolicy(p => p.WaitAndRetryAsync(4, attempt => TimeSpan.FromSeconds(Math.Pow(1.45, attempt)) + TimeSpan.FromMilliseconds(_xoroshiro.Next(0, 100))))
    .AddTransientHttpErrorPolicy(p => p.CircuitBreakerAsync(5, TimeSpan.FromSeconds(15)));

services.AddTransient<IMyApiClient>(sp =>
{
    // logic to select URL
    var context = sp.GetRequiredService<IHttpContextAccessor>().HttpContext;
    string[] host = context.Request.Host.Host.Replace("someurl", "").Split(new char[] { '.' });
[...]

PR #2741, opened 2019-12-04
Add more client factory overloads

rynowak:
Fixes: #814

This change makes a few existing features more discoverable by adding
first-class overloads on `AddHttpClient`.


src/HttpClientFactory/Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs
```diff
@@ -760,5 +760,197 @@ public static IHttpClientBuilder AddHttpClient<TClient>(this IServiceCollection
...
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (factory == null)
+            {
+                throw new ArgumentNullException(nameof(factory));
+            }
+
+            var name = TypeNameHelper.GetTypeDisplayName(typeof(TClient), fullName: false);
```
pranavkm: Huh, I didn't realize that's what's happening under the covers. Out of curiosity, why this over `typeof(TClient).Name`?

rynowak: Nested classes and generics. These names show up in log category names, so we want them to be recognizable.

PR #2759, opened 2019-12-05
Remove requirement for order of objects in returned enumerable in DependencyInjection.Tests

ENikS:
Unity container fails [ResolvesMixedOpenClosedGenericsAsEnumerable](https://github.com/aspnet/Extensions/blob/0aa6213e1c4abd0cf45189e93751bf38b9f8c07c/src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs#L719) test. The test expects enumerable to have elements in the order consistent with the order of descriptors in the Service Collection.

Unity unable to keep the same order because it stores constructable types and generic type definitions in different repositories. As a result, relations between these mixed types are lost once types are registered.

I am proposing to change the check to verify existence of the instance in the returned collection and perhaps verify default resolve as well.

### DIFF
--- a/src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
+++ b/src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs
@@ -730,6 +730,7 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
             var serviceProvider = CreateServiceProvider(serviceCollection);
 
             var enumerable = serviceProvider.GetService<IEnumerable<IFakeOpenGenericService<PocoClass>>>().ToArray();
+            var service = serviceProvider.GetService<IFakeOpenGenericService<PocoClass>>();
 
             // Assert
             Assert.Equal(3, enumerable.Length);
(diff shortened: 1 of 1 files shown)


ENikS:
I am assuming that after registering collection like this:
```c#
serviceCollection.AddSingleton(typeof(IFakeOpenGenericService<PocoClass>), typeof(FakeService));
serviceCollection.AddSingleton(typeof(IFakeOpenGenericService<>), typeof(FakeOpenGenericService<>));
serviceCollection.AddSingleton<IFakeOpenGenericService<PocoClass>>(instance);
```
resolving `IFakeOpenGenericService<PocoClass>` should return the `instance`. The test is passing in Visual Studio on windows. Am I missing something?

ericstj:
Closing this PR as the PR branch appears to be deleted.  If this is still of interest please resubmit a PR in https://github.com/dotnet/runtime where this code now lives:
https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.DependencyInjection/tests/DI.Specification.Tests/DependencyInjectionSpecificationTests.cs

/cc @maryamariyan @eerhardt

Issue #2252, opened 2019-08-29
Microsoft.Extensions.Caching.Memory.CacheExtensions.GetOrCreateAsync is lacking ConfigureAwait(false)


ghelyar:
https://github.com/aspnet/Extensions/blob/master/src/Caching/Abstractions/src/MemoryCacheExtensions.cs#L119

This causes deadlocks under certain scenarios (e.g. asp.net context, when an outer awaitable is blocking synchronously)

ladeak:
Any updates on this issue?

PR #2767, opened 2019-12-08
GetOrCreateAsync's factory should use ConfigureAwait

ladeak:
GetOrCreateAsync's factory should use with ConfigureAwait false, as some UI applications might block on the UI thread.
 - Added ConfigureAwait(false) in GetOrCreateAsync
 - Added a test to cover the issue.
 - Added two helper classes for the tests. Helper classes comments' indicated their original source.

Not sure if tests falls into the 'too hard' category, because of the helper classes.

Addresses #2252

### DIFF
--- a/src/Caching/Directory.Build.targets
+++ b/src/Caching/Directory.Build.targets
@@ -1,0 +1,8 @@
+﻿<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.targets))\Directory.Build.targets" />
+
+  <PropertyGroup Condition="'$(IsShipping)' == 'true'">
+    <CodeAnalysisRuleset>$(MSBuildThisFileDirectory)AspNet.Extensions.ruleset</CodeAnalysisRuleset>
+  </PropertyGroup>
+
+</Project>
--- a/eng/Dependencies.props
+++ b/eng/Dependencies.props
@@ -41,6 +41,7 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion)" />
     <LatestPackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
     <LatestPackageReference Include="Microsoft.DotNet.GenAPI" Version="$(MicrosoftDotNetGenApiPackageVersion)" />
+    <LatestPackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="$(MicrosoftCodeAnalysisFxCopAnalyzersVersion)" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' != '.NETFramework' " Label="MSBuild">
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -105,6 +105,8 @@
     <MicrosoftCodeAnalysisCSharpPackageVersion>2.8.0</MicrosoftCodeAnalysisCSharpPackageVersion>
     <MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>2.8.0</MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
     <MicrosoftExtensionsDependencyModelPackageVersion>2.1.0</MicrosoftExtensionsDependencyModelPackageVersion>
+    <!-- Used for FxCop rules during build -->
+    <MicrosoftCodeAnalysisFxCopAnalyzersVersion>2.9.8</MicrosoftCodeAnalysisFxCopAnalyzersVersion>
     <!-- Stable dotnet/corefx packages no longer updated for .NET Core 3 -->
     <SystemMemoryPackageVersion>4.5.2</SystemMemoryPackageVersion>
     <SystemNetHttpPackageVersion>4.3.2</SystemNetHttpPackageVersion>
--- a/src/Caching/Abstractions/src/DistributedCacheExtensions.cs
+++ b/src/Caching/Abstractions/src/DistributedCacheExtensions.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Extensions.Caching.Distributed
         /// <returns>A task that gets the string value from the stored cache key.</returns>
         public static async Task<string> GetStringAsync(this IDistributedCache cache, string key, CancellationToken token = default(CancellationToken))
         {
-            var data = await cache.GetAsync(key, token);
+            var data = await cache.GetAsync(key, token).ConfigureAwait(false);
             if (data == null)
             {
                 return null;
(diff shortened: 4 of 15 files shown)


ladeak:
@anurse and @Tratcher please review thid PR.

ladeak:
@Pilchie please suggest reviewers

src/Caching/Memory/test/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -121,6 +122,34 @@ public async Task GetOrCreateAsync_AddsNewValue()
...
+                {
+                    await Task.Delay(1).ConfigureAwait(false);
+                    return new object();
+                }).GetAwaiter().GetResult();
+                return result;
+            }
+
+            var sut = Task.Factory.StartNew(SystemUnderTestMethod, CancellationToken.None, TaskCreationOptions.None, singleThreadedScheduler);
+
+            var exception = await Record.ExceptionAsync(() => sut.TimeoutAfter(TimeSpan.FromSeconds(3)));
```
anurse: What's the purpose of recording the exception just to check it's `null`? I'd think `await sut.TimeoutAfter(...)` would be sufficient because if it throws, the test will fail and will display the exception details.

src/Caching/Memory/test/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -121,6 +122,34 @@ public async Task GetOrCreateAsync_AddsNewValue()
             Assert.Same(obj, result);
         }
 
+        [Fact]
+        public async Task GetOrCreateAsync_DoesNotBlock()
```
anurse:
```suggestion
        public async Task GetOrCreateAsync_AvoidsCapturingSynchronizationContextWhenAwaitingUseCallback()
```

It certainly can *block* depending on what the user callback does.

davidfowl:
@rynowak had some thoughts about ConfigureAwait in these libraries, specifically that its not worth trying to test all of the code paths and instead use an analyzer to make sure all the places have `ConfigureAwait(false)`.

cc @JamesNK

ladeak:
@davidfowl , I am happy to with the analyzer instead of tests, could you suggest one?

src/Caching/Memory/test/MemoryCacheSetAndRemoveTests.cs
ladeak: Got it, just being explicit on what is being asserted. Will update as suggested.

JamesNK:
I have setup Microsoft.CodeAnalysis.FxCopAnalyzers in https://github.com/grpc/grpc-dotnet

It comes with a lot of analyzers. I have disabled most of them but ConfigureAwait false is still enabled: https://github.com/grpc/grpc-dotnet/blob/master/Grpc.DotNet.ruleset

src/Caching/Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
+﻿<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.targets))\Directory.Build.targets" />
+
+  <PropertyGroup>
+    <CodeAnalysisRuleset Condition="'$(IsShipping)' == 'true'">$(MSBuildThisFileDirectory)AspNet.Extensions.ruleset</CodeAnalysisRuleset>
+  </PropertyGroup>
+
```
ladeak: I am new to build.props and build.targets, and not sure why adding the package reference here instead of the csproj does not work as expected, but the same condition does not seem to filter out sample and tests projects.
```
<ItemGroup Condition...>
  <Reference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" />
</ItemGroup>
```

ladeak:
I suppose I need to run GenerateReferenceSource or GenerateReferenceAssemblies to resolve build issues.

ladeak:
When Directory.Build.target defines the build Reference to FxCopAnalyzer and as well as CodeAnalysisRuleset, I get a build error because ref/*.csproj-s are not updated.

> eng/targets/ResolveReferences.targets(200,5): error MSB3245: Could not resolve this reference. Could not locate the package or project for "Microsoft.CodeAnalysis.FxCopAnalyzers". Did you update baselines and dependencies lists? See docs/ReferenceResolution.md for more details.

When I  move
```xml
  <ItemGroup>
    <Reference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" />
  </ItemGroup>
```
to Directory.Build.props, and run the script that updates the ref/ projects, then build.cmd, I get the following error:

> C:\Users\...\source\repos\Extensions\.dotnet\sdk\5.0.100-alpha1-015536\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets(256,5): error NETSDK1023: A PackageReference for 'Microsoft.CodeAnalysis.FxCopAnalyzers' was included in your project. This package is implicitly referenced by the .NET SDK and you do not typically need to reference it from your project. For more information, see https://aka.ms/sdkimplicitrefs [C:\Users\...\source\repos\Extensions\src\Caching\Abstractions\ref\Microsoft.Extensions.Caching.Abstractions.csproj]

What would be the suggested way to resolve this issue?

rynowak:
Looks like we probably have to add this to the references list: https://github.com/aspnet/Extensions/blob/master/docs/ReferenceResolution.md#example-adding-a-new-dependency

@anurse @JamesNK - can you help with that?

anurse:
Step 1 and 2 there should be all that's needed, I believe. Adding the `<Reference>` element to the project and then the `<LatestPackageReference>` element to the `eng/Dependencies.props` file. While this package does comes from "another dotnet team", it's [dotnet/roslyn-analyzers](https://github.com/dotnet/roslyn-analyzers) which is not part of our dependency flow process, so there's no need to go through step 3.

@ladeak let me know if you have any issues making those updates!

ladeak:
@anurse , I have added the Dependency to the "External dependencies":
https://github.com/aspnet/Extensions/pull/2767/files#diff-8c508e8b957801ef7a859521e59d58b9 but still returns a build error:
```ini
eng/targets/ResolveReferences.targets(200,5): error MSB3245: Could not resolve this reference. Could not locate the package or project for "Microsoft.CodeAnalysis.FxCopAnalyzers". Did you update baselines and dependencies lists? See docs/ReferenceResolution.md for more details.
```

ladeak:
@anurse Could you suggest, I did those two steps as suggested, but I still get errors.

src/Caching/Abstractions/ref/Microsoft.Extensions.Caching.Abstractions.csproj
```diff
@@ -6,9 +6,11 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <Compile Include="Microsoft.Extensions.Caching.Abstractions.netstandard2.0.cs" />
     <Reference Include="Microsoft.Extensions.Primitives"  />
+    <Reference Include="Microsoft.CodeAnalysis.FxCopAnalyzers"  />
```
Tratcher: Should this be showing up in the ref assemblies? Will it show up as a dependency from the official package?

ladeak: I do not think it should.

ladeak:
@anurse , @tracychms please complete reviews :)

Tratcher:
Thanks

Issue #528, opened 2018-10-09
Blacklist headers from logging


rasmus:
In our infrastructure, logging e.g. `Authorization` headers is frowned upon. I would like to be able to blacklist headers, making sure certain headers isn't copied to any logs.

If logging were "secure by default", then maybe `Authorization` and `Cookie` headers could be blacklisted by default.

rynowak:
Done! #2772

PR #2772, opened 2019-12-11
Blacklist HTTP headers from logging

a-tsymbal:
Addresses #528

### DIFF
--- a/src/HttpClientFactory/Http/ref/Microsoft.Extensions.Http.netcoreapp.cs
+++ b/src/HttpClientFactory/Http/ref/Microsoft.Extensions.Http.netcoreapp.cs
@@ -18,8 +18,8 @@ namespace Microsoft.Extensions.DependencyInjection
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Func<System.IServiceProvider, System.Net.Http.HttpMessageHandler> configureHandler) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Func<System.Net.Http.HttpMessageHandler> configureHandler) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler<THandler>(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder) where THandler : System.Net.Http.HttpMessageHandler { throw null; }
-        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Collections.Generic.ICollection<string> redactedLoggedHeaderNames) { throw null; }
-        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Predicate<string> isSensitiveHeader) { throw null; }
+        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Collections.Generic.IEnumerable<string> redactedLoggedHeaderNames) { throw null; }
+        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Func<string, bool> shouldRedactHeaderValue) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder SetHandlerLifetime(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.TimeSpan handlerLifetime) { throw null; }
     }
     public static partial class HttpClientFactoryServiceCollectionExtensions
@@ -55,7 +55,7 @@ namespace Microsoft.Extensions.Http
         public System.TimeSpan HandlerLifetime { get { throw null; } set { } }
         public System.Collections.Generic.IList<System.Action<System.Net.Http.HttpClient>> HttpClientActions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
         public System.Collections.Generic.IList<System.Action<Microsoft.Extensions.Http.HttpMessageHandlerBuilder>> HttpMessageHandlerBuilderActions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
-        public System.Predicate<string> IsSensitiveHeader { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
+        public System.Func<string, bool> ShouldRedactHeaderValue { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
         public bool SuppressHandlerScope { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
     }
     public abstract partial class HttpMessageHandlerBuilder
@@ -81,13 +81,15 @@ namespace Microsoft.Extensions.Http.Logging
 {
     public partial class LoggingHttpMessageHandler : System.Net.Http.DelegatingHandler
     {
-        public LoggingHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger, System.Predicate<string> isSensitiveHeader) { }
+        public LoggingHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger) { }
+        public LoggingHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger, Microsoft.Extensions.Http.HttpClientFactoryOptions options) { }
         [System.Diagnostics.DebuggerStepThroughAttribute]
         protected override System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage> SendAsync(System.Net.Http.HttpRequestMessage request, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
     public partial class LoggingScopeHttpMessageHandler : System.Net.Http.DelegatingHandler
     {
-        public LoggingScopeHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger, System.Predicate<string> isSensitiveHeader) { }
+        public LoggingScopeHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger) { }
+        public LoggingScopeHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger, Microsoft.Extensions.Http.HttpClientFactoryOptions options) { }
         [System.Diagnostics.DebuggerStepThroughAttribute]
         protected override System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage> SendAsync(System.Net.Http.HttpRequestMessage request, System.Threading.CancellationToken cancellationToken) { throw null; }
     }
--- a/src/HttpClientFactory/Http/ref/Microsoft.Extensions.Http.netcoreapp.cs
+++ b/src/HttpClientFactory/Http/ref/Microsoft.Extensions.Http.netcoreapp.cs
@@ -18,8 +18,9 @@ namespace Microsoft.Extensions.DependencyInjection
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Func<System.IServiceProvider, System.Net.Http.HttpMessageHandler> configureHandler) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Func<System.Net.Http.HttpMessageHandler> configureHandler) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler<THandler>(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder) where THandler : System.Net.Http.HttpMessageHandler { throw null; }
+        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Collections.Generic.ICollection<string> redactedLoggedHeaderNames) { throw null; }
+        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Predicate<string> isSensitiveHeader) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder SetHandlerLifetime(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.TimeSpan handlerLifetime) { throw null; }
-        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder SetLogSensitiveHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Collections.Generic.ICollection<string> logSensitiveHeaders) { throw null; }
     }
     public static partial class HttpClientFactoryServiceCollectionExtensions
     {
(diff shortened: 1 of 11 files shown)


a-tsymbal:
I will add tests for this later, just want to have some feedback before tests.

src/HttpClientFactory/Http/ref/Microsoft.Extensions.Http.netcoreapp.cs
```diff
@@ -19,6 +19,7 @@ public static partial class HttpClientBuilderExtensions
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Func<System.Net.Http.HttpMessageHandler> configureHandler) { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder ConfigurePrimaryHttpMessageHandler<THandler>(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder) where THandler : System.Net.Http.HttpMessageHandler { throw null; }
         public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder SetHandlerLifetime(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.TimeSpan handlerLifetime) { throw null; }
+        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder SetLogSensitiveHeaders(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Collections.Generic.ICollection<string> logSensitiveHeaders) { throw null; }
```
rynowak: I think this is overall looking pretty good. 

If we keep the current design approach. I have some suggestions about naming/terminology that I think will make things clearer.

```
public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaderValues(this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, System.Collections.Generic.ICollection<string> logSensitiveHeaders)

public System.Collections.Generic.IList<string> RedactedLoggedHeaderNames { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
```

I like using the term *redact* here - because it's exactly what this does. It still logs whether or not the header was present, but doesn't log the value.

----

Another design idea that comes up here is to make this a predicate-based system. The usage would look like:

```
services.AddHttpClient("Foo").RedactLoggedHeaders((header) =>
{
    if (header == "Authorize")
    {
        return false;
    }

    return true;
});
```

This is nice because the app developer can write any code they want, and we don't need to add any additional features. We could keep the functionality you added in this PR (deny-list) and just layer it on top as an overload.

```
services.AddHttpClient("Foo").RedactLoggedHeaders(new []{ "Authorize", });
```

If we wanted to pivot this to support an allow-list, you could have that too, without really changing the logic.

```
services.AddHttpClient("Foo").AllowLoggedHeaders(new []{ "Content-Type", "Content-Length", });
```

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -525,6 +526,35 @@ public static IHttpClientBuilder SetHandlerLifetime(this IHttpClientBuilder buil
             return builder;
         }
 
+        /// <summary>
+        /// Sets the HTTP headers collection which values should not be logged.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="logSensitiveHeaders">The HTTP headers collection which values should not be logged.</param>
+        /// <returns></returns>
```
rynowak: Reminder to all of us to revisit doc comments once we lock on the design.

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
```diff
@@ -90,10 +92,27 @@ public override string ToString()
 
                 for (var i = 0; i < Values.Count; i++)
                 {
+                    var isLogSensitiveHeader = false;
                     var kvp = Values[i];
                     builder.Append(kvp.Key);
                     builder.Append(": ");
 
+                    for (int headerIndex = 0; headerIndex < _logSensitiveHeaders.Count; headerIndex++)
```
rynowak: I think we'd really like to avoid the O(MN) here nested loop in the case that the list of logged headers is really large. I think this could really be simpler and more performant with a hashset.

src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandlerBuilderFilter.cs
```diff
@@ -41,12 +50,15 @@ public Action<HttpMessageHandlerBuilder> Configure(Action<HttpMessageHandlerBuil
                 var outerLogger = _loggerFactory.CreateLogger($"System.Net.Http.HttpClient.{loggerName}.LogicalHandler");
                 var innerLogger = _loggerFactory.CreateLogger($"System.Net.Http.HttpClient.{loggerName}.ClientHandler");
 
+                var options = _optionsMonitor.Get(builder.Name);
+                var logSensitiveHeaders = new ReadOnlyCollection<string>(_optionsMonitor.Get(builder.Name).LogSensitiveHeaders);
```
rynowak: Duplicate call to `_optionsMonitor.Get(...)` here.

a-tsymbal: 😊

a-tsymbal:
Thank you for your review @rynowak!
I was about to ask for help with naming, `RedactedLoggedHeaderNames` looks better.

I will handle all the comments as soon as i can.

a-tsymbal:
Updated PR according to the feedback @rynowak.

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
```diff
@@ -94,15 +96,23 @@ public override string ToString()
...
+                        foreach (var value in (IEnumerable<object>)kvp.Value)
+                        {
+                            builder.Append(value);
+                            builder.Append(", ");
+                        }
 
-                    // Remove the extra ', '
-                    builder.Remove(builder.Length - 2, 2);
-                    builder.AppendLine();
+                        // Remove the extra ', '
```
khellang: Is it worth pulling this into an extension method on `StringBuilder` and calling `StringBuilder.AppendJoin` on `!NETSTANDARD2_0`?

davidfowl:
@rynowak this needs an API review. I've added the tag.

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
a-tsymbal: It might be useful to have a similar API on the `StringBuilder` as we already have on the `String` for joining elements.

rynowak:
@a-tsymbal  - this looks pretty good. The next step here is the team is going to take this through API review. I'm going to give you another code review pass assuming that the API review won't change the shape significantly. 

We'll review this Monday 12/16

a-tsymbal:
I think I forgot to initialize `public Predicate<string> IsSensitiveHeader { get; set; }` with `= (header) => false;`, otherwise exception will be thrown when headers aren't configured.

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -485,6 +486,60 @@ public static IHttpClientBuilder AddTypedClient<TClient>(this IHttpClientBuilder
...
+            builder.Services.Configure<HttpClientFactoryOptions>(builder.Name, options =>
+            {
+                options.IsSensitiveHeader = isSensitiveHeader;
+            });
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Sets the HTTP headers collection which values should be redacted before logging.
```
rynowak:
```suggestion
        /// Sets the collection of HTTP headers names for which values should be redacted before logging.
```

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -485,6 +486,60 @@ public static IHttpClientBuilder AddTypedClient<TClient>(this IHttpClientBuilder
...
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="redactedLoggedHeaderNames">The HTTP headers collection which values should not be logged.</param>
```
rynowak:
```suggestion
        /// <param name="redactedLoggedHeaderNames">The collection of HTTP headers names for which values should be redacted before logging.</param>
```

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -485,6 +486,60 @@ public static IHttpClientBuilder AddTypedClient<TClient>(this IHttpClientBuilder
...
+        /// <returns>The <see cref="IHttpClientBuilder"/>.</returns>
+        public static IHttpClientBuilder RedactLoggedHeaders(this IHttpClientBuilder builder, ICollection<string> redactedLoggedHeaderNames)
```
rynowak: Suggestion: make this `IEnumerable<string>`. There's no real advantage of `ICollection<>` for this use case

src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandler.cs
```diff
@@ -19,15 +19,22 @@ namespace Microsoft.Extensions.Http.Logging
     public class LoggingHttpMessageHandler : DelegatingHandler
     {
         private ILogger _logger;
+        private readonly Predicate<string> _isSensitiveHeader;
 
-        public LoggingHttpMessageHandler(ILogger logger)
+        public LoggingHttpMessageHandler(ILogger logger, Predicate<string> isSensitiveHeader)
```
rynowak: This is a breaking API change. This needs to be a separate constructor to avoid the unneccesary break.

Another idea, should this take `HttpClientFactoryOptions` instead? Then we don't have to keep adding constructors if we add more options that impact logging.

src/HttpClientFactory/Http/src/Logging/LoggingScopeHttpMessageHandler.cs
```diff
@@ -15,15 +15,22 @@ namespace Microsoft.Extensions.Http.Logging
     public class LoggingScopeHttpMessageHandler : DelegatingHandler
     {
         private ILogger _logger;
+        private readonly Predicate<string> _isSensitiveHeader;
 
-        public LoggingScopeHttpMessageHandler(ILogger logger)
+        public LoggingScopeHttpMessageHandler(ILogger logger, Predicate<string> isSensitiveHeader)
```
rynowak: Same comments as the other handler. We avoid breaking changes like this unless really necessary, and it's not necessary here.

src/HttpClientFactory/Http/test/Logging/HttpHeadersLogValueTest.cs
```diff
@@ -0,0 +1,55 @@
...
+                "Request Headers:" + Environment.NewLine +
+                "secureHeader1: *" + Environment.NewLine +
+                "unsecureHeader1: value1" + Environment.NewLine +
+                "unsecureHeader2: value2" + Environment.NewLine +
+                "secureHeader2: *" + Environment.NewLine,
+                result);
+        }
+
+        private class TestHttpHeaders : HttpHeaders { }
+    }
```
rynowak: I'd also like to see an integration test or two - you can use a patterrn like [this](https://github.com/aspnet/Extensions/blob/49bf9920e4e9079b02e2f775b2bc923abf2be4d1/src/HealthChecks/HealthChecks/test/HealthCheckPublisherHostedServiceTest.cs#L165) to hijack logging for testing.

blowdart:
I just have one comment here, the PR swaps between redacted (in the header list) and sensitive (in the check methods), consistency would be nice. I would prefer redacted, because you can redact for lots of reasons, not just security sensitivity.

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
khellang: It doesn't have to be an extension method, but it could still be worth #ifdef'ing to call the "proper" `AppendJoin`, no? I don't really know how often this is called, but appending and then removing doesn't sound as optimized as [`AppendJoin`](https://github.com/dotnet/runtime/blob/4359ebfc9943608b34f411366b1e544ac45702b7/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs#L1297).

Tratcher:
Any thoughts on generalizing the terminology even more? There are other reasons to avoid logging headers besides their sensitive nature, such as them being large or not useful in logs. E.g. A User-Agent header is both large and repetitive.
Suggestions:
RedactLoggedHeaders -> SuppressLoggedHeaderValue
IsSensitiveHeader  => ShouldSuppressHeaderValueLog or ShouldLogHeaderValue

On another note, it's odd that we'll log HttpClient headers but refuse to log server headers. @blowdart would an opt-in or opt-out option like this be sufficient caution to allow logging headers in the servers?

rynowak:
> On another note, it's odd that we'll log HttpClient headers but refuse to log server headers.

It's not odd that things written by different people at different times do different things 😆

rynowak:
> Any thoughts on generalizing the terminology even more? There are other reasons to avoid logging headers besides their sensitive nature, such as them being large or not useful in logs. E.g. A User-Agent header is both large and repetitive.

Using the term `redact` here instead of something else is on purpose instead of something more general. The presence of the header is still logged, the value is blanked out.

rynowak:
> I just have one comment here, the PR swaps between redacted (in the header list) and sensitive (in the check methods), consistency would be nice. I would prefer redacted, because you can redact for lots of reasons, not just security sensitivity.

Agreed on my part.

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
rynowak: that would be fine

rynowak:
Here's a thing for API Review:

```C#
namespace Microsoft.Extensions.DependencyInjection
{
    public static class HttpClientBuilderExtensions
    {
        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(
            this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder,
            System.Collections.Generic.ICollection<string> redactedLoggedHeaderNames) 
            { throw null; }

        public static Microsoft.Extensions.DependencyInjection.IHttpClientBuilder RedactLoggedHeaders(
            this Microsoft.Extensions.DependencyInjection.IHttpClientBuilder builder, 
            System.Predicate<string> shouldRedactHeaderValue) 
            { throw null; }
    }
}

namespace Microsoft.Extensions.Http
{
    public class HttpClientFactoryOptions
    {
        public System.Predicate<string> ShouldRedactHeaderValue { get; set; }
    }
}

namespace Microsoft.Extensions.Http.Logging
{
    public partial class LoggingHttpMessageHandler : System.Net.Http.DelegatingHandler
    {
        public LoggingHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger, HttpClientFactoryOptions options) 
        { }
    }

    public partial class LoggingScopeHttpMessageHandler : System.Net.Http.DelegatingHandler
    {
        public LoggingHttpMessageHandler(Microsoft.Extensions.Logging.ILogger logger, HttpClientFactoryOptions options) 
        { }
    }
}
```

blowdart:
@Tratcher if we can come up with a sensible set of redacted defaults I'd be ok doing the same server side.

rynowak:
If it helps the discussion about server, we only log headers at trace level. This PR adds to the ability to further filter the set of values logged when you're at the highest possible level of logging.

Tratcher:
Logged https://github.com/aspnet/AspNetCore/issues/17908 for the server discussion.

pranavkm:
The API suggestion @rynowak points out https://github.com/aspnet/Extensions/pull/2772#issuecomment-566138695 here is API approved. Please update your PR to match the suggested API shape.

src/HttpClientFactory/Http/src/Logging/LoggingHttpMessageHandler.cs
a-tsymbal: Please review my changes again, i see only one constructor in the API review code, but you suggest not to introduce breaking changes, could you please suggest what should I do with those constructors?

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
a-tsymbal: Could you please check whether I used `#if` correctly?

src/HttpClientFactory/Http/src/DependencyInjection/HttpClientBuilderExtensions.cs
```diff
@@ -485,6 +486,62 @@ public static IHttpClientBuilder AddTypedClient<TClient>(this IHttpClientBuilder
             return builder;
         }
 
+        /// <summary>
+        /// Sets the <see cref="Func{T, R}"/> which determines whether to redact the HTTP header value before logging.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/>.</param>
+        /// <param name="shouldRedactHeaderValue">The <see cref="Func{T, R}"/> which determines whether redact the HTTP header value before logging.</param>
```
rynowak:
```suggestion
        /// <param name="shouldRedactHeaderValue">The <see cref="Func{T, R}"/> which determines whether to redact the HTTP header value before logging.</param>
```

src/HttpClientFactory/Http/src/HttpClientFactoryOptions.cs
```diff
@@ -69,6 +69,11 @@ public TimeSpan HandlerLifetime
             }
         }
 
+        /// <summary>
+        /// The <see cref="Func{T, R}"/> which determines whether redact the HTTP header value before logging.
```
rynowak:
```suggestion
        /// The <see cref="Func{T, R}"/> which determines whether to redact the HTTP header value before logging.
```

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
rynowak: Looks right. `#if NETCOREAPP ` is fine. You can remove the TODO

src/HttpClientFactory/Http/test/Logging/HttpHeadersLogValueTest.cs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http.Headers;
+using Microsoft.Extensions.Http.Logging;
+using Xunit;
+
+namespace Microsoft.Extensions.Http.Tests.Logging
```
rynowak: nit: you can remove the `.Tests` from the namespace

src/HttpClientFactory/Http/test/Logging/HttpHeadersLogValueTest.cs
rynowak: @a-tsymbal are you able to add an integration test? I can help with this if you get stuck

a-tsymbal: @rynowak It would be great if you can help a little bit.

rynowak:
@a-tsymbal - I'll take a crack at integration tests for this later today

rynowak:
Sent a PR to your branch: https://github.com/a-tsymbal/Extensions/pull/1

src/HttpClientFactory/Http/src/Logging/HttpHeadersLogValue.cs
```diff
@@ -94,15 +96,28 @@ public override string ToString()
...
+                    if (_shouldRedactHeaderValue(kvp.Key))
                     {
-                        builder.Append(value);
-                        builder.Append(", ");
+                        builder.Append("*");
+                        builder.AppendLine();
                     }
+                    else
+                    {
+#if NETCOREAPP // TODO: or use !NETSTANDARD2_0?
```
rynowak: Please remove the TODO comment from here

Issue #2662, opened 2019-11-14
StringValues casting undefined


breyed:
Implicitly casting various values of a `StringValues` object leads to undefined results.

``` c#
string fromList = new StringValues("a","b");
string fromEmpty = StringValues.Empty;
```

Will `fromList` be comma-separated, newline-separated, or something else?

Will `fromEmpty` be an empty string? Will it be null? Something else?

The results should be well defined. `implicit operator StringValues(string value)` needs an XML comment so that the semantics appear in the metadata-generated class definition in Visual Studio when you hit F12 on `StringValues`. Likely, other members need docs, too. Currently, `StringValues` has a summary, but that's all.

Marusyk:
Can I take this one?

benbhall:
I think it is `public static implicit operator string(StringValues values)` that needs the XML to address the particular issue demonstrated.

I've had a go at that and also added XML to everything else. Will pop a PR in.

PR #2780, opened 2019-12-12
XML documentation comments for StringValues.

benbhall:
- Add XML comment to clarify behaviour casting to string as per issue.
- Add XML across StringValues members.
- Minor tidy with some expression bodies.

Addresses #2662

### DIFF
--- a/src/Primitives/src/StringValues.cs
+++ b/src/Primitives/src/StringValues.cs
@@ -15,40 +15,76 @@ namespace Microsoft.Extensions.Primitives
     /// </summary>
     public readonly struct StringValues : IList<string>, IReadOnlyList<string>, IEquatable<StringValues>, IEquatable<string>, IEquatable<string[]>
     {
+        /// <summary>
+        /// A readonly instance of the <see cref="StringValues"/> struct whose value is an empty string array.
+        /// </summary>
+        /// <remarks>
+        /// In application code, this field is most commonly used to safely represent a <see cref="StringValues"/> that has null string values.
+        /// </remarks>
         public static readonly StringValues Empty = new StringValues(Array.Empty<string>());
 
         private readonly object _values;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StringValues"/> structure using the specified string.
+        /// </summary>
+        /// <param name="value">A string value.</param>
         public StringValues(string value)
         {
             _values = value;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StringValues"/> structure using the specified array of strings.
+        /// </summary>
+        /// <param name="values">A string array.</param>
         public StringValues(string[] values)
         {
             _values = values;
         }
 
+        /// <summary>
+        /// Defines an implicit conversion of a given string to a <see cref="StringValues"/>.
+        /// </summary>
+        /// <param name="value">A string to implicitly convert.</param>
         public static implicit operator StringValues(string value)
         {
             return new StringValues(value);
         }
 
+        /// <summary>
+        /// Defines an implicit conversion of a given string array to a <see cref="StringValues"/>.
+        /// </summary>
+        /// <param name="values">A string array to implicitly convert.</param>
         public static implicit operator StringValues(string[] values)
         {
             return new StringValues(values);
         }
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


benbhall:
Ah okay. I'm probably distracted by the issue being addressed around undefined behaviour, so I've tried to take the surprise out of interface where there could be confusion.

Is it a worthwhile piece? i.e. shall I have another bash at it? Any further tips welcome. At least I got it right for `Count` :)

anurse:
> Is it a worthwhile piece?

I think docs for this are definitely worthwhile. It's the type used as the "value" for HTTP headers, and some doc clarity would be very valuable.

benbhall:
Okay second attempt in @anurse.

I've tried to adhere to your advice re documenting **how to use** the members and just used `<remark>` to supplement:

- `Empty`
- The `implicit operator string (StringValues values)` I believe #2662 concerned
- `ToArray()` re violation principle of least astonishment raised in #1814 

Hoping to pick up the latter when I've finished with these docs if that's okay.

src/Primitives/src/StringValues.cs
```diff
@@ -112,6 +159,10 @@ private static string OutOfBounds()
             return Array.Empty<string>()[0]; // throws
         }
 
+        /// <summary>
+        /// Converts the value of the current <see cref="StringValues"/> object to its equivalent string representation.
+        /// </summary>
+        /// <returns>A string representation of the value of the current <see cref="StringValues"/> object.</returns>
```
Tratcher: With multiple entries being comma separated.

src/Primitives/src/StringValues.cs
```diff
@@ -240,6 +299,11 @@ private string[] GetArrayValue()
             }
         }
 
+        /// <summary>
+        /// Returns the zero-based index of the first occurrence of a item in the <see cref="StringValues" />.
```
Tratcher:
```suggestion
        /// Returns the zero-based index of the first occurrence of an item in the <see cref="StringValues" />.
```

src/Primitives/src/StringValues.cs
```diff
@@ -433,6 +520,12 @@ public static StringValues Concat(string value, in StringValues values)
             return new StringValues(combined);
         }
 
+        /// <summary>
+        /// Determines whether two specified <see cref="StringValues"/> objects have the same values.
```
Tratcher:
```suggestion
        /// Determines whether two specified <see cref="StringValues"/> objects have the same values in the same order. 
```

src/Primitives/src/StringValues.cs
```diff
@@ -15,40 +15,76 @@ namespace Microsoft.Extensions.Primitives
...
         public static implicit operator StringValues(string[] values)
         {
             return new StringValues(values);
         }
 
+        /// <summary>
+        /// Defines an implicit conversion of a given <see cref="StringValues"/> to a string.
+        /// </summary>
+        /// <remarks>
+        /// Returns <c>null</c> where <see cref="StringValues"/> has been initialized from an empty string array or is <see cref="StringValues.Empty"/>.
```
Tratcher: with multiple values joined as a comma separated string.

src/Primitives/src/StringValues.cs
```diff
@@ -216,6 +267,14 @@ static string GetJoinedStringValueFromArray(string[] values)
             }
         }
 
+        /// <summary>
+        /// Creates a string array from the current <see cref="StringValues"/> object.
+        /// </summary>
+        /// <returns>A string array represented by this instance.</returns>
+        /// <remarks>
+        /// <para>If the <see cref="StringValues"/> contains a string internally, it is copied to a new array.</para>
```
Tratcher:
```suggestion
        /// <para>If the <see cref="StringValues"/> contains a single string internally, it is copied to a new array.</para>
```

src/Primitives/src/StringValues.cs
benbhall: Good spot.

src/Primitives/src/StringValues.cs
benbhall: Yep good idea will push something up now.

Tratcher:
Thanks

PR #2781, opened 2019-12-12
[Release/3.0] Set IsShipping=false for all TestingUtils projects

wtgodbe:
Without this, the ref project(s) in this dir get the default value of `IsShipping=true` from Arcade: https://github.com/dotnet/arcade/blob/0c9b0442f2ce607a90615a7bfad4ead069150bc6/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets#L18. This means that other projects which compile against ref projects in this folder (i.e., `Microsoft.Extensions.Logging.Testing` compiles against `Microsoft.AspNetCore.Testing`) will get non-suffixed versions for their dependencies. When we ship stable, `Microsoft.Extensions.Logging.Testing` lists a dependency on a stable version of `Microsoft.AspNetCore.Testing`, which does not exist.

We should think of a better way to resolve this issue - for example, AspNetCore calculates a global default for `IsShipping`, which would prevent this type of issue: https://github.com/aspnet/AspNetCore/blob/20fc1adf2ada01a4a1fc7934706d2b641acbb9f4/Directory.Build.props#L23

It's also worth noting that `Microsoft.AspNetCore.Testing` is unique in that it is a non-shipping project that has a ref assembly. We should understand why that ref assembly is needed, and decide if it can be removed (removing it caused some errors that seemed non-trivial to fix). The ref assembly was added in this PR: https://github.com/aspnet/Extensions/pull/2483

CC @dougbu @JunTaoLuo @mmitche

### DIFF
--- a/src/TestingUtils/Directory.Build.props
+++ b/src/TestingUtils/Directory.Build.props
@@ -4,5 +4,6 @@
   <PropertyGroup>
     <!-- These projects depend on a 3rd party source -->
     <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
+    <IsShipping>false</IsShipping>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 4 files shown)


wtgodbe:
I'll be back at my desk in around ~45 minutes, let's wait for @dougbu's signoff before merging

Issue #741, opened 2018-12-13
Sql Cache In-Memory OLTP?


dauger:
Has having the option of using Sql Server 2014's in-memory OLTP instead of traditional tables been evaluated? It seems like it might be a good fit for a cache, but I'll admit that I have not done a deep dive on in-memory OLTP's capabilities & restrictions yet.

yahorsi:
Seems it might be perfect to use in memory oltp feature in ASP.NET Core Distributed Cache (SqlServer provider)

xantari:
Here is my version of SQL In Memory table capable distributed cache implementation with only slight variations for your out of the box SQL Cache provider.

You can view it here: https://github.com/xantari/Microsoft.Extensions.Caching.SqlServerInMemory

Issue #1894, opened 2019-06-27
Add support for “in-memory tables” in Microsoft.Extensions.Caching.SqlServer


epignosisx:
Currently, the Microsoft.Extensions.Caching.SqlServer library does not play well with SQL Server in-memory tables. The reason is that the data type DateTimeOffset is not supported by IM tables. 

The library would need to change to a different data type like DateTime2.

xantari:
@epignosisx I created a branch based off of the Microsoft.Extensions.Caching.SqlServer library that supports this: https://github.com/xantari/Microsoft.Extensions.Caching.SqlServerInMemory

xantari:
Related: https://github.com/dotnet/aspnetcore/issues/17640

anurse:
Duplicate of https://github.com/dotnet/extensions/issues/741

PR #2791, opened 2019-12-15
Add support In-Memory Table For SQL Server Cache 

ricoisme:
 - use object instead of explicit type

I think In-memory table better than disk table for Caching item.

My sql's script of test as below

`
--Add MEMORY_OPTIMIZED_DATA filegroup to the database.
ALTER DATABASE ricoCacheDB
ADD FILEGROUP MemoryFG CONTAINS MEMORY_OPTIMIZED_DATA

--Add file to the MEMORY_OPTIMIZED_DATA filegroup.
ALTER DATABASE ricoCacheDB
ADD FILE
	( NAME = ricoCacheDB_File1,
	  FILENAME = N'/var/opt/mssql/data/ricoCacheDB_File1')
TO FILEGROUP MemoryFG

--Drop table if it already exists.
IF OBJECT_ID('dbo.AllCacheMem','U') IS NOT NULL
	DROP TABLE [dbo].AllCacheMem
GO

--Create memory optimized table and indexes on the memory optimized table.
CREATE TABLE [dbo].[AllCacheMem]
(
	[Id] [nvarchar] (449) NOT NULL,
	[Value] [varbinary] (max) NOT NULL,
	[ExpiresAtTime] varchar(64) not null,
	[SlidingExpirationInSeconds] [bigint] NULL,
	[AbsoluteExpiration] varchar (64) null,
	CONSTRAINT PK_AllCache PRIMARY KEY NONCLUSTERED (id),
	INDEX IX_ExpiresAtTime (ExpiresAtTime)
) WITH (MEMORY_OPTIMIZED = ON, DURABILITY = SCHEMA_ONLY)
GO
`

### DIFF
--- a/src/Caching/SqlServer/src/DatabaseOperations.cs
+++ b/src/Caching/SqlServer/src/DatabaseOperations.cs
@@ -98,7 +98,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
         {
             token.ThrowIfCancellationRequested();
 
-            await GetCacheItemAsync(key, includeValue: false, token:token);
+            await GetCacheItemAsync(key, includeValue: false, token: token);
         }
 
         public virtual void DeleteExpiredCacheItems()
@@ -164,7 +164,7 @@ namespace Microsoft.Extensions.Caching.SqlServer
             ValidateOptions(options.SlidingExpiration, absoluteExpiration);
 
             using (var connection = new SqlConnection(ConnectionString))
-            using(var upsertCommand = new SqlCommand(SqlQueries.SetCacheItem, connection))
+            using (var upsertCommand = new SqlCommand(SqlQueries.SetCacheItem, connection))
             {
                 upsertCommand.Parameters
                     .AddCacheItemId(key)
@@ -228,7 +228,8 @@ namespace Microsoft.Extensions.Caching.SqlServer
                     {
                         var id = reader.GetFieldValue<string>(Columns.Indexes.CacheItemIdIndex);
 
-                        expirationTime = reader.GetFieldValue<DateTimeOffset>(Columns.Indexes.ExpiresAtTimeIndex);
+                        var expirationTimeStr = reader.GetFieldValue<string>(Columns.Indexes.ExpiresAtTimeIndex);
+                        DateTimeOffset.TryParse(expirationTimeStr, out expirationTime);
 
                         if (!reader.IsDBNull(Columns.Indexes.SlidingExpirationInSecondsIndex))
                         {
@@ -238,8 +239,10 @@ namespace Microsoft.Extensions.Caching.SqlServer
 
                         if (!reader.IsDBNull(Columns.Indexes.AbsoluteExpirationIndex))
                         {
-                            absoluteExpiration = reader.GetFieldValue<DateTimeOffset>(
+                            var absoluteExpirationStr = reader.GetFieldValue<string>(
                                 Columns.Indexes.AbsoluteExpirationIndex);
+                            DateTimeOffset.TryParse(absoluteExpirationStr, out var currentAbsoluteExpiration);
+                            absoluteExpiration = currentAbsoluteExpiration;
                         }
 
                         if (includeValue)
@@ -294,8 +297,9 @@ namespace Microsoft.Extensions.Caching.SqlServer
                     {
                         var id = await reader.GetFieldValueAsync<string>(Columns.Indexes.CacheItemIdIndex, token);
 
-                        expirationTime = await reader.GetFieldValueAsync<DateTimeOffset>(
+                        var expirationTimeStr = await reader.GetFieldValueAsync<string>(
                             Columns.Indexes.ExpiresAtTimeIndex, token);
+                        DateTimeOffset.TryParse(expirationTimeStr, out expirationTime);
 
                         if (!await reader.IsDBNullAsync(Columns.Indexes.SlidingExpirationInSecondsIndex, token))
                         {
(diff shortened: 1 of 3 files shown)


mfields3:
Would much prefer to use an In-Memory Table for caching purposes.
Is there an ETA on this?

ricoisme:
My quick performance test via benchmarkdotnet as below
image

[reference my blog](https://medium.com/ricos-note/add-support-in-memory-for-microsoft-extensions-caching-sqlserver-e86ed4b4492)

src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -238,8 +239,10 @@ protected virtual byte[] GetCacheItem(string key, bool includeValue)
 
                         if (!reader.IsDBNull(Columns.Indexes.AbsoluteExpirationIndex))
                         {
-                            absoluteExpiration = reader.GetFieldValue<DateTimeOffset>(
+                            var absoluteExpirationStr = reader.GetFieldValue<object>(
                                 Columns.Indexes.AbsoluteExpirationIndex);
+                            DateTimeOffset.TryParse(absoluteExpirationStr?.ToString(), out var currentAbsoluteExpiration);
```
ericstj, line 244: Is it save to convert to string representation?  Won't this result in loss of information like time-zone?

src/Caching/SqlServer/src/DatabaseOperations.cs
```diff
@@ -238,8 +239,10 @@ protected virtual byte[] GetCacheItem(string key, bool includeValue)
 
                         if (!reader.IsDBNull(Columns.Indexes.AbsoluteExpirationIndex))
                         {
-                            absoluteExpiration = reader.GetFieldValue<DateTimeOffset>(
+                            var absoluteExpirationStr = reader.GetFieldValue<object>(
```
ericstj, line 242: If this is a string it should be fetched as a string.

src/Caching/SqlServer/src/MonoSqlParameterCollectionExtensions.cs
```diff
@@ -22,23 +22,22 @@ internal static class MonoSqlParameterCollectionExtensions
...
-
         public static SqlParameterCollection AddAbsoluteExpirationMono(
                     this SqlParameterCollection parameters,
                     DateTimeOffset? utcTime)
         {
             if (utcTime.HasValue)
             {
                 return parameters.AddWithValue(
-                    Columns.Names.AbsoluteExpiration, SqlDbType.DateTime, utcTime.Value.UtcDateTime);
+                    Columns.Names.AbsoluteExpiration, SqlDbType.VarChar, utcTime.Value.UtcDateTime.ToString());
```
ericstj, line 35: Is this changing the schema of the table?  How does that work with  existing tables that had a different schema?

ericstj:
Looks like https://github.com/dotnet/extensions/pull/2806 deletes much of this as well so perhaps this is no longer needed.

src/Caching/SqlServer/src/DatabaseOperations.cs
ricoisme: It is save and never loss relate  information of datetime.

src/Caching/SqlServer/src/MonoSqlParameterCollectionExtensions.cs
```diff
@@ -22,23 +22,22 @@ internal static class MonoSqlParameterCollectionExtensions
             this SqlParameterCollection parameters,
             DateTimeOffset utcTime)
         {
-            return parameters.AddWithValue(Columns.Names.ExpiresAtTime, SqlDbType.DateTime, utcTime.UtcDateTime);
+            return parameters.AddWithValue(Columns.Names.ExpiresAtTime, SqlDbType.VarChar, utcTime.UtcDateTime.ToString());
```
tarekgh, line 25: If the stored date is going to parsed later then this not good to store the date using `DateTime.ToString()` as it will depend on the current culture and parsing can fail later. It is better either to use InvariantCulture when formatting it `utcTime.UtcDateTime.ToString(CultureInfo.InvariantCulture)` or just format the date using "o" or "r" and when parsing using the same specifiers `utcTime.UtcDateTime.ToString("o")`.

src/Caching/SqlServer/src/DatabaseOperations.cs
ericstj: @tarekgh captured my concern here: https://github.com/dotnet/extensions/pull/2791#discussion_r504328413

src/Caching/SqlServer/src/MonoSqlParameterCollectionExtensions.cs
Tratcher: If this is a breaking schema change then that seems like a non-starter. Or is there also a migration plan proposed?

ericstj:
> Can you describe more about why you are changing the type used for storing expiration and how this relates to an the title of this PR (support for in-memory table)?

I'll answer my own question.  As mentioned in https://github.com/dotnet/extensions/issues/1894 it looks like the reason this doesn't work with in-memory tables is due to it's use of DateTimeOffset which I guess is not supported for in memory tables. 

I believe this PR will fix: https://github.com/dotnet/extensions/issues/741

I still think we should understand the outcome of discussion in https://github.com/dotnet/extensions/pull/2806 before taking this change.

ricoisme (commented):
> > Can you describe more about why you are changing the type used for storing expiration and how this relates to an the title of this PR (support for in-memory table)?
> 
> I'll answer my own question. As mentioned in #1894 it looks like the reason this doesn't work with in-memory tables is due to it's use of DateTimeOffset which I guess is not supported for in memory tables.
> 
> I believe this PR will fix: #741
> 
> I still think we should understand the outcome of discussion in #2806 before taking this change.

You are right. The DateTimeOffset  type is not supported in IM table.
In addition,I have used varchar instead of DateTimeOffset is work and no any problem in real-world.

JunTaoLuo:

xantari:
Can someone explain why my issue https://github.com/dotnet/extensions/issues/1894  was removed? It documented in detail the issue with the current SQL Server cache provider not using the correct data types. And I just learned that there is no plan to fix this as indicated today here: https://github.com/dotnet/aspnetcore/issues/28378

ericstj:
@Anipik was https://github.com/dotnet/extensions/issues/1894 transferred to dotnet/runtime?

Anipik:
No issue was deleted/removed during the archiving process. It must be transferred to dotnet/runtime or dotnet/aspnetcore, 
@xantari do u remember the title of the issue ?

Issue #1268, opened 2019-03-17
Performance improvement for Microsoft.Extensions.Caching.SqlServer


MarkCBB:
### Problem/Performance improvement
There are variables never used but fetched from the database and converted in the code. This means two unnecessary costs: fetching the value from the database and converting the value.

For more details about the impact in performance see the section Additional context.

NB: I'm continuing based on [this answer](https://github.com/aspnet/Caching/pull/406#issuecomment-428639431) written by @muratg in a pull request opened some time ago in the previous location of this repository.

### Proposed solution
The files changed in [this Pull Request](https://github.com/aspnet/Caching/pull/406/files) addressed this performance optimization. Removes the usage of the non used variables and modifies the query to don't select from the database those values when are not needed.

If you consider to merge these changes I can send another pull request for the current location of the repository.

### Describe alternatives you've considered
This improvement requires the removal of some code sections and the optimization of the query sent to the database. No more alternatives.

### Additional context
I've created [this performance test](https://github.com/MarkCBB/test-performance-sqlserver-cache) -easy to set up and run- in order to know if the improvement makes any difference based in execution times. In my opinion there are variations making this improvement meaningful. In the readme.md file you'll find how to run this test and more detailed times.

Find below a resume of the execution times in milliseconds:

* Average. Optimized: 23643, non optimized 25276 (optimized is 1633 faster)
* Slowest. Optimized: 28287, non optimized 31622 (optimized is 3335 faster)
* Fastest. Optimized: 21845, non optimized 23004 (optimized is 1159 faster)

muratg:
cc @davidfowl

MarkCBB:
This issue and its related pull request implements some housekeeping, code
cleaning and a few performance improvement. I understand the low priority
as the gain is low, but honestly, I don’t see why this pull request should
be closed losing the work done. Also, recently I’ve updated the branch
resolving all conflicts with the last changes in master branch.
May be is time to review the pull request, merge it and then close the
issue ;)
Please, could you take a look or keep this pull request alive?

JunTaoLuo:
The linked PR has been merged.

PR #2806, opened 2019-12-24
Performance improvement by cleaning code and query tuning

MarkCBB:
 - Removing unused columns in query
 - Cleaning useless code
 - Choosing the right query when selecting

Addresses #1268


anurse:
Can you please include data indicating the performance improvement? We need to see data showing the value of the improvement in order to consider a change like this.

MarkCBB:
Hello @anurse,
You'll find some data at the end of the description in [the related issue](https://github.com/aspnet/Extensions/issues/1268).

src/Caching/SqlServer/src/Columns.cs
```diff
@@ -22,7 +22,7 @@ public static class Indexes
             public const int ExpiresAtTimeIndex = 1;
             public const int SlidingExpirationInSecondsIndex = 2;
             public const int AbsoluteExpirationIndex = 3;
```
anurse: Are these three constants needed now that they aren't included the in the query at all?

MarkCBB: Indeed, my bad, I should realize that these constants are no longer needed. In fact, the constant at the line 21 is not needed either. I'll remove them in a commit.

The only constant that would remain is `CacheItemValueIndex`. Even that is the only one that remains, I'm in favor to keep it because adds some level of consistency and makes easier to track and refactor the code. Just finding the references is possible to know where this column is used.

MarkCBB:
Hello @anurse. Indeed, the gain in performance is not too high. I don’t have real-world examples because I realized these improvements just reading the code. This code was my reference to develop [a package to use MongoDB as a cache database](https://www.nuget.org/packages/MarkCBB.Extensions.Caching.MongoDB/) based on your approach for these extensions.

In my opinion this change is nice to have because: keeps the code cleaner removing unused blocks, removes unused fields from the query (as you said), improves a little bit the performance which is always positive, and is not too complex.

However, given your comments and the low impact, I totally understand the low priority. Feel free to ask me to close this pull request or just leave it open until you can take some action ;)

src/Caching/SqlServer/src/MonoDatabaseOperations.cs
```diff
@@ -98,12 +79,9 @@ protected override async Task<byte[]> GetCacheItemAsync(string key, bool include
             }
 
             byte[] value = null;
-            TimeSpan? slidingExpiration = null;
-            DateTimeOffset? absoluteExpiration = null;
-            DateTimeOffset expirationTime;
             using (var connection = new SqlConnection(ConnectionString))
             {
-                var command = new SqlCommand(SqlQueries.GetCacheItem, connection);
+                var command = new SqlCommand(query, connection);
```
eerhardt, line 84: Do we have some sort of test for this? I'm surprised this bug snuck through... it seems like a good place to test.

eerhardt (approved):
This change looks good to me.

cc @Tratcher

src/Caching/SqlServer/src/MonoDatabaseOperations.cs
MarkCBB: Good point. Not so easy. Let me explain: `protected DatabaseOperations.GetCacheItem` (Async and not Async) functions are called only by public `DatabaseOperations.GetCacheItem[Async]` that makes the call passing the parameter `includeValue` to true, and `DatabaseOperations.RefreshCacheItem[Async]` which sets the parameter `includeValue` to false (as doesn’t need the value). 

When the parameter `includeValue` is true GetCacheItem opens a dataReader with two commands in the sql string: one to update the expiration date and another one to get and return the cached value (or null if not found). When the parameter `includeValue` is false, only opens the dataReader with a single update command in the sql string to update the expiration date and always returns null. 

It is difficult to test the value of the sql string inside the function or the commands executed in the database depending on the value of the parameter. 

In my opinion, the cleanest approach is not related to testing, but to refactor a bit the code. Roughly explained, I would get rid of the boolean parameter `includeValue` and let the function `GetCacheItemAsync` just with the implementation for refresh the expiration date and get the value. Next, implement the `RefreshCacheItem` into another method/function exclusively written for that. 

Then each method will use the proper command string and implementation without any conditional. It is not a so complex refactoring.

Given the “not so high” priority in this project and the limited impact in performance and features (this is more housekeeping), I would vote to make it in another pull request when there is time to do it.

Sorry for this loong explanation but I couldn't make it shorter to propose another approach instead of modifying the tests.

eerhardt: ok, thanks for the explanation.

eerhardt:
@Tratcher - It looks like this is ready to be merged. Any feedback before merging this?

Tratcher (changes requested):
I think this PR is highlighting a much bigger problem. It looks like sliding expiration was never completely implemented. For example look at RefreshCacheItem, I don't think it works. The values you removed are needed to calculate the new expiration time. Why new expiration wasn't being calculated and pushed back to the database I don't know.

MarkCBB:
> I think this PR is highlighting a much bigger problem. It looks like sliding expiration was never completely implemented. For example look at RefreshCacheItem, I don't think it works. The values you removed are needed to calculate the new expiration time. Why new expiration wasn't being calculated and pushed back to the database I don't know.

As far as I know `RefreshCacheItem` works. Calls `GetCacheItem` with `includeValue` parameter set to false. Therefore the query that will run is the one coming from this property `SqlQueries.GetCacheItemWithoutValue` and this property returns the value set in [this line](https://github.com/dotnet/extensions/blob/229898d9ec8cb6c37b183caee2de449a47c6430c/src/Caching/SqlServer/src/SqlQueries.cs#L65) which is the value coming from UpdateCacheItemFormat property set [here](https://github.com/dotnet/extensions/blob/229898d9ec8cb6c37b183caee2de449a47c6430c/src/Caching/SqlServer/src/SqlQueries.cs#L17). This query is an update that implements the logic to compute and set the new expiration time.

This update is also sent when the value is fetched (`includeValue` parameter set to true in the call to `GetCacheItem`) because the sql query contains two commands: the update and the select to get the value. Doing so you update and get the value in a single roundtrip to the database, which I think is a good and well performant approach. [Here](https://github.com/dotnet/extensions/blob/229898d9ec8cb6c37b183caee2de449a47c6430c/src/Caching/SqlServer/src/SqlQueries.cs#L64) is where the both sql commands are set to the property (the [update](https://github.com/dotnet/extensions/blob/229898d9ec8cb6c37b183caee2de449a47c6430c/src/Caching/SqlServer/src/SqlQueries.cs#L17) and the [select](https://github.com/dotnet/extensions/blob/229898d9ec8cb6c37b183caee2de449a47c6430c/src/Caching/SqlServer/src/SqlQueries.cs#L31)).

I checked some time ago and I'm almost sure that did work properly. Also, I think there are some test cases covering that feature, but I should run it again in a real application to make 100% sure. 

From my point of view, the only fact that could make it fail is if the transaction mode is set to explicit but normally IMPLICIT_TRANSACTIONS is OFF by default. Also, this would make fail everything, because there are not transactions in this code.

Finally, the code is not the cleanest state and it could be a good idea to refactor it, but IMO this should be done in another pull request. [See this discussion.](https://github.com/dotnet/extensions/pull/2806#discussion_r504834724)

Tratcher:
Ooh, I hadn't seen that the expiration update was embedded in the query, that's slick.

Tratcher (approved):

Issue #1085, opened 2019-02-07
[Feature request] Add ReloadOnChange For KeyPerFile Configuration


Kahbazi:
Can we have reload on change for KeyPerFile configuration?

davidfowl:
We should look into supporting this. When updating config maps mapped to volumes on k8s we don't react to those changes today.

cc @glennc @rynowak

Kahbazi:
@davidfowl Can I send a PR for this?

davidfowl:
Yep!

HaoK:
Fixed by https://github.com/dotnet/extensions/pull/2808, thanks again @Kahbazi

PR #2808, opened 2019-12-28
Add ReloadOnChange to KeyPerFile configuration provider

Kahbazi:
Addresses #1085

### DIFF
--- a/src/Configuration/Config.KeyPerFile/ref/Microsoft.Extensions.Configuration.KeyPerFile.netcoreapp.cs
+++ b/src/Configuration/Config.KeyPerFile/ref/Microsoft.Extensions.Configuration.KeyPerFile.netcoreapp.cs
@@ -6,6 +6,7 @@ namespace Microsoft.Extensions.Configuration
     public static partial class KeyPerFileConfigurationBuilderExtensions
     {
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddKeyPerFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, System.Action<Microsoft.Extensions.Configuration.KeyPerFile.KeyPerFileConfigurationSource> configureSource) { throw null; }
+        public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddKeyPerFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, string directoryPath) { throw null; }
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddKeyPerFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, string directoryPath, bool optional) { throw null; }
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddKeyPerFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, string directoryPath, bool optional, bool reloadOnChange) { throw null; }
     }
--- a/src/Configuration/Config.KeyPerFile/ref/Microsoft.Extensions.Configuration.KeyPerFile.netcoreapp.cs
+++ b/src/Configuration/Config.KeyPerFile/ref/Microsoft.Extensions.Configuration.KeyPerFile.netcoreapp.cs
@@ -7,13 +7,15 @@ namespace Microsoft.Extensions.Configuration
     {
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddKeyPerFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, System.Action<Microsoft.Extensions.Configuration.KeyPerFile.KeyPerFileConfigurationSource> configureSource) { throw null; }
         public static Microsoft.Extensions.Configuration.IConfigurationBuilder AddKeyPerFile(this Microsoft.Extensions.Configuration.IConfigurationBuilder builder, string directoryPath, bool optional) { throw null; }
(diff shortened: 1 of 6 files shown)


src/Configuration/Config.KeyPerFile/ref/Microsoft.Extensions.Configuration.KeyPerFile.netcoreapp.cs
```diff
@@ -24,6 +26,8 @@ public partial class KeyPerFileConfigurationSource : Microsoft.Extensions.Config
         public System.Func<string, bool> IgnoreCondition { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
         public string IgnorePrefix { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
         public bool Optional { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
+        public int ReloadDelay { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
+        public bool ReloadOnChange { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
         public Microsoft.Extensions.Configuration.IConfigurationProvider Build(Microsoft.Extensions.Configuration.IConfigurationBuilder builder) { throw null; }
```
rynowak: Will need API review (even if its the same as what other providers do).

src/Configuration/Config.KeyPerFile/src/KeyPerFileConfigurationBuilderExtensions.cs
```diff
@@ -19,6 +19,18 @@ public static class KeyPerFileConfigurationBuilderExtensions
         /// <param name="optional">Whether the directory is optional.</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
         public static IConfigurationBuilder AddKeyPerFile(this IConfigurationBuilder builder, string directoryPath, bool optional)
+            => builder.AddKeyPerFile(directoryPath, optional, reloadOnChange: false);
```
rynowak: question - do our other config providers do reloading by default? Looks like JSON does not: https://github.com/aspnet/Extensions/blob/master/src/Configuration/Config.Json/src/JsonConfigurationExtensions.cs#L38

davidfowl: We should be consistent so it should be off by default.

src/Configuration/Config.KeyPerFile/src/KeyPerFileConfigurationProvider.cs
```diff
@@ -30,12 +48,17 @@ private static string TrimNewLine(string value)
         /// Loads the docker secrets.
```
Kahbazi: Should this summary change to something general?

src/Configuration/Config.KeyPerFile/src/KeyPerFileConfigurationBuilderExtensions.cs
```diff
@@ -19,6 +19,18 @@ public static class KeyPerFileConfigurationBuilderExtensions
         /// <param name="optional">Whether the directory is optional.</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
         public static IConfigurationBuilder AddKeyPerFile(this IConfigurationBuilder builder, string directoryPath, bool optional)
```
Kahbazi: Should I also add another overload with only directoryPath as input, with optional and reloadOnChange set to false to match other providers?

davidfowl: Yes

src/Configuration/Config.KeyPerFile/src/KeyPerFileConfigurationBuilderExtensions.cs
HaoK: Yeah this is expensive and has some unexpected side effects so generally reload should be off by default unless people explicitly make it enabled

pranavkm:
The API looks pretty identical to the file configuration source, so that looks good. @rynowak \ @HaoK could one of you drive getting the implementation approved and merged?

HaoK:
Sure i can drive this

src/Configuration/Config.KeyPerFile/test/KeyPerFileTests.cs
```diff
@@ -217,6 +217,61 @@ void ReloadLoop()
             Assert.Equal("Foo", options.Text);
         }
 
+        [Fact]
```
HaoK: Can you add a test where nothing exists to start, and the directory/files get created later so its the nothing found to something case as opposed to just modification scenarios?

src/Configuration/Config.KeyPerFile/src/KeyPerFileConfigurationProvider.cs
HaoK: Sure, loads the configuration values

src/Configuration/Config.KeyPerFile/test/KeyPerFileTests.cs
Kahbazi: @Haok I added the test.

HaoK:
Looks good, thanks @Kahbazi !

Issue #2809, opened 2019-12-29
Configuration > Binder - Support for binding by custom attribute names. 


Coderrob:
### Is your feature request related to a problem? Please describe.
I am trying to use JSON configuration settings that have names such as "some.flag" that get loaded into the IConfiguration dictionary. The key itself is mapped via JsonPropertyAttribute on my classes. The Binder project does not know of anything outside of the objects property names when it attempts to lookup the value by property name in the configuration dictionary.

### Describe the solution you'd like
What would be useful is to scan the custom attributes to identify either a "known" name or allow for specifying attribute types to use for lookup with binder options.

### Describe alternatives you've considered
Alternative is to use private properties that act as a setter for the desired properties that have the same matching name from JSON. The period in the name doesn't help in my case.

PR #2810, opened 2019-12-29
Add custom attribute property name lookup to configuration binder.

Coderrob:
Took a swag at addressing the gap between the Json source and the Binder projects. The Json settings I work with contain names that don't match the setting classes. Those are mapped by JsonProperty or DataMember where a name is given that can be mapped.

The binder project, rightfully, doesn't know about Json or any other formats. The one area where it could be built up more is around looking at the attributes associated with the properties of the object it's attempting to bind to the configuration dictionary. This is an attempt to try to just that.

 - Added new BindingOption flag for binding by attribute names
 - Broke out the property value lookup into a standalone function
 - Added a function to return "name" values from custom attribute data from the PropertyType

Addresses #2809 (in this specific format)

### DIFF
--- a/src/Configuration/Config.Binder/src/BinderOptions.cs
+++ b/src/Configuration/Config.Binder/src/BinderOptions.cs
@@ -13,5 +13,11 @@ namespace Microsoft.Extensions.Configuration
         /// If true, the binder will attempt to set all non read-only properties.
         /// </summary>
         public bool BindNonPublicProperties { get; set; }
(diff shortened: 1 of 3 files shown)


Coderrob:
Not entirely sure why all but linux tests are failing for the CI. I've built locally and run the project tests. No new files were added and project reference changes were not required. Will wipe and pull PR branch locally again to re-run tests.

PR #2817, opened 2020-01-01
Remove ref project from Microsoft.AspNetCore.Testing

JunTaoLuo:
Adresses https://github.com/aspnet/Extensions/issues/2784


JunTaoLuo:
I'm curious whether there are any other ramifications for this change. We'll need to test building with this change in AspNetCore.

eng/ProjectReferences.props
```diff
@@ -56,7 +56,6 @@
     <ProjectReferenceProvider Include="Microsoft.Extensions.Options.DataAnnotations" ProjectPath="$(RepoRoot)src\Options\DataAnnotations\src\Microsoft.Extensions.Options.DataAnnotations.csproj" RefProjectPath="$(RepoRoot)src\Options\DataAnnotations\ref\Microsoft.Extensions.Options.DataAnnotations.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Options" ProjectPath="$(RepoRoot)src\Options\Options\src\Microsoft.Extensions.Options.csproj" RefProjectPath="$(RepoRoot)src\Options\Options\ref\Microsoft.Extensions.Options.csproj" />
     <ProjectReferenceProvider Include="Microsoft.Extensions.Primitives" ProjectPath="$(RepoRoot)src\Primitives\src\Microsoft.Extensions.Primitives.csproj" RefProjectPath="$(RepoRoot)src\Primitives\ref\Microsoft.Extensions.Primitives.csproj" />
-    <ProjectReferenceProvider Include="Microsoft.AspNetCore.Testing" ProjectPath="$(RepoRoot)src\TestingUtils\Microsoft.AspNetCore.Testing\src\Microsoft.AspNetCore.Testing.csproj" RefProjectPath="$(RepoRoot)src\TestingUtils\Microsoft.AspNetCore.Testing\ref\Microsoft.AspNetCore.Testing.csproj" />
```
dougbu: I've lost some context. Why turn off `$(IsProjectReferenceProvider)` for Microsoft.AspNetCore.Texting? That is, why not just change `$(IsProjectReferenceProvider)`?

src/Logging/Logging.Testing/src/Microsoft.Extensions.Logging.Testing.csproj
```diff
@@ -18,7 +18,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Reference Include="Microsoft.AspNetCore.Testing" />
+    <ProjectReference Include="..\..\..\TestingUtils\Microsoft.AspNetCore.Testing\src\Microsoft.AspNetCore.Testing.csproj" />
```
dougbu: Why isn't this redundant?

JunTaoLuo: Why would it be? This is a src project.

dougbu: Oops

eng/ProjectReferences.props
JunTaoLuo: I'm not quite understanding what you are saying here? I did change `IsProjectReferenceProvider` to false for this project.

dougbu: I fixed my topo after making the original comment. Think you're responding to the email notification rather than what's here. In any case, it seems unnecessary to change more than `$(HasReferenceAssembly)`.

JunTaoLuo: >In any case, please update Microsoft.AspNetCore.Testing to remove the ref/ project and change other projects to use an @(ProjectReference) item when they depend on Microsoft.AspNetCore.Testing.
To make it clear what's going on later, please also mark Microsoft.AspNetCore.Testing with <IsProjectReferenceProvider>false</IsProjectReferenceProvider>.

We wanted to ensure that any references to M.A.Testing to use `ProjectReference` and that a `Reference` to the project would fail.

dougbu: Thanks for the reminder.

JunTaoLuo:
FYI, I spoke with @wtgodbe and we already have workarounds in place so we don't need this PR for correctness. It's for cleaning up only. I won't be putting this under servicing consider, but will include it in a future patch as a tell mode infrastructure change.

Pilchie:
@JunTaoLuo Are we taking this for 3.1.2?

dougbu:
@JunTaoLuo Aren't the "workarounds" hacks that we should cleanup outside and separate from a security update? Since we have this ready, my preference is to take this now.

JunTaoLuo:
Yes we should be taking this in 3.1.2. It's a cleanup item.

PR #2819, opened 2020-01-03
Preserve functional test logs on CI

JunTaoLuo:
WIP. Testing approach in CI.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -111,11 +111,17 @@ stages:
           inputs:
             command: custom
             arguments: 'locals all -clear'
-        - script: eng\common\cibuild.cmd -configuration $(_BuildConfig) -prepareMachine $(_BuildArgs) /bl:artifacts/log/build.windows.binlog
+        - script: eng\common\cibuild.cmd
+            -configuration $(_BuildConfig)
+            -prepareMachine
+            $(_BuildArgs)
+            /bl:artifacts/log/build.windows.binlog
           displayName: Build and Publish
-        - script: eng\scripts\ci-flaky-tests.cmd -configuration $(_BuildConfig) /bl:artifacts/log/build.windows.flakytests.binlog
+        - script: eng\scripts\ci-flaky-tests.cmd -configuration $(_BuildConfig)
           displayName: Run Flaky Tests
           continueOnError: true
+        - script: dotnet msbuild eng/repo.targets /t:EnsureFunctionalTestLogsPreserved
+          displayName: Ensure functional test logs preserved
         - powershell: eng\scripts\KillProcesses.ps1
           displayName: Kill processes
           condition: always()
@@ -190,6 +196,7 @@ stages:
             --prepareMachine
             -p:BuildNodeJS=true
             $(_OfficialBuildArgs)
+            /bl:artifacts/log/build.linux.binlog
           displayName: Build
         - script: eng/scripts/ci-flaky-tests.sh --configuration $(_BuildConfig)
           displayName: Run Flaky Tests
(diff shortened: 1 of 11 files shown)


azure-pipelines.yml
```diff
@@ -111,11 +111,17 @@ stages:
...
+        - script: eng\common\cibuild.cmd
+            -configuration $(_BuildConfig)
+            -prepareMachine
+            $(_BuildArgs)
+            /bl:artifacts/log/build.windows.binlog
           displayName: Build and Publish
         - script: eng\scripts\ci-flaky-tests.cmd -configuration $(_BuildConfig)
           displayName: Run Flaky Tests
           continueOnError: true
+        - script: dotnet msbuild eng/repo.targets /t:EnsureFunctionalTestLogsPreserved
```
JunTaoLuo: I suppose I can add the verification to macOS and Linux runs as well, but I don't think it adds much.

eng/repo.targets
```diff
@@ -52,4 +52,16 @@
           SkipNonexistentTargets="true"
           SkipNonexistentProjects="true" />
   </Target>
+
+  <Target Name="EnsureFunctionalTestLogsPreserved">
+    <PropertyGroup>
+      <FunctionalLogDir>$(ArtifactsDir)log\Microsoft.Extensions.Logging.Testing.Tests\$(DefaultNetCoreTargetFramework)\AssemblyTestLogTests\</FunctionalLogDir>
+      <FlakyTestLog>$(FunctionalLogDir)FunctionalLogs_LogsPreservedFromFlakyRun.log</FlakyTestLog>
```
JunTaoLuo: The only complaint I have is that this is fragile against renames, but oh well.

src/Logging/Logging.Testing/src/build/Microsoft.Extensions.Logging.Testing.props
```diff
@@ -9,15 +9,21 @@
   <Target Name="SetLoggingTestingAssemblyAttributes"
     BeforeTargets="GetAssemblyAttributes"
     Condition="'$(GenerateLoggingTestingAssemblyAttributes)' != 'false'">
+    <PropertyGroup>
+      <PreserveExistingLogsInOutput Condition="'$(PreserveExistingLogsInOutput)' == '' AND '$(ContinuousIntegrationBuild)' == 'true'">true</PreserveExistingLogsInOutput>
```
JunTaoLuo: The idea is that we will preserve existing logs on CI but delete existing logs locally by default. These can always be overridden if needed.

azure-pipelines.yml
BrennanConroy: Meh

src/TestingUtils/Microsoft.AspNetCore.Testing/ref/Microsoft.AspNetCore.Testing.net46.cs
```diff
@@ -295,6 +295,7 @@ public sealed partial class TestFileOutputContext
         public string TestName { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
         public static string GetAssemblyBaseDirectory(System.Reflection.Assembly assembly, string baseDirectory = null) { throw null; }
         public static string GetOutputDirectory(System.Reflection.Assembly assembly) { throw null; }
+        public static bool GetPreserveExistingLogsInOutput(System.Reflection.Assembly assembly) { throw null; }
```
dougbu: You won't need to do this if the ref/ assembly removal happens first. But, I guess that has to wait for branches to open ☹️

JunTaoLuo: Yup

PR #2822, opened 2020-01-03
Set argument if no option has found

Kahbazi:
I changed the behavior so the `CommandLineApplication` couldn't found the available option, it tries to add the `arg` to the `Arguments` list.

Addresses https://github.com/aspnet/AspNetCore/issues/18095

### DIFF
--- a/src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
+++ b/src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
@@ -26,10 +26,13 @@ namespace Microsoft.Extensions.CommandLineUtils
         // options.
         private readonly bool _continueAfterUnexpectedArg;
 
-        public CommandLineApplication(bool throwOnUnexpectedArg = true, bool continueAfterUnexpectedArg = false)
+        private readonly bool _treatUnmatchedOptionsAsArguments;
+
+        public CommandLineApplication(bool throwOnUnexpectedArg = true, bool continueAfterUnexpectedArg = false, bool treatUnmatchedOptionsAsArguments = false)
         {
             _throwOnUnexpectedArg = throwOnUnexpectedArg;
             _continueAfterUnexpectedArg = continueAfterUnexpectedArg;
+            _treatUnmatchedOptionsAsArguments = treatUnmatchedOptionsAsArguments;
             Options = new List<CommandOption>();
             Arguments = new List<CommandArgument>();
             Commands = new List<CommandLineApplication>();
@@ -136,7 +139,7 @@ namespace Microsoft.Extensions.CommandLineUtils
             CommandLineApplication command = this;
             CommandOption option = null;
             IEnumerator<CommandArgument> arguments = null;
-            var argumentsAssigned = true;
+            var argumentsAssigned = false;
 
             for (var index = 0; index < args.Length; index++)
             {
@@ -162,7 +165,7 @@ namespace Microsoft.Extensions.CommandLineUtils
                         var longOptionName = longOption[0];
                         option = command.GetOptions().SingleOrDefault(opt => string.Equals(opt.LongName, longOptionName, StringComparison.Ordinal));
 
-                        if (option == null)
+                        if (option == null && _treatUnmatchedOptionsAsArguments)
                         {
                             if (arguments == null)
                             {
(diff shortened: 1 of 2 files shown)


src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -161,6 +162,25 @@ public int Execute(params string[] args)
                         var longOptionName = longOption[0];
                         option = command.GetOptions().SingleOrDefault(opt => string.Equals(opt.LongName, longOptionName, StringComparison.Ordinal));
 
+                        if (option == null)
```
anurse: I missed that this was in shared code. I think we should make this conditional on a boolean being passed in to the `CommandLineApplication`, something like `treatUnmatchedOptionsAsArguments`.

Kahbazi:
It's obvious but I just wanted to point out with this solution we won't be able to pass an option as an argument. Are we ok with that?
For example we can't set `--project` as a key for user-secret because it is an option.

anurse:
I think that's OK. There are limits to what we can do here. As I noted in the issue, the file can be directly edited.

src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -317,7 +320,7 @@ public int Execute(params string[] args)
                     option = null;
                 }
 
-                if (!processed && argumentsAssigned)
+                if (!processed && !argumentsAssigned)
```
anurse: What's the reason this was inverted?

Kahbazi: I named this variable wrong. `argumentsAssigned` was true when no variable was set. Although everything was working right, I just flip everything to make the name right and hence invert this check!

anurse: Ah, I missed that you added this boolean because I was only looking at the most recent commit ;). Thanks for the clarification!

src/Shared/test/CommandLineApplicationTests.cs
```diff
@@ -1166,5 +1166,28 @@ public void ThrowsExceptionOnInvalidOption()
 
             Assert.Equal($"Unrecognized option '{inputOption}'", exception.Message);
         }
+
+        [Fact]
+        public void TreatUnmatchedOptionsAsArguments()
```
anurse: Can you:

1. Add an actual option to this test and validate that it **does** get properly matched.
2. Add a second test that is essentially the same but asserts the behavior when `treatUnmatchedOptionsAsArguments` is explicitly set to `false`? (It should fail because of the unmatched options).

src/Shared/src/CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -26,10 +26,13 @@ internal class CommandLineApplication
         // options.
         private readonly bool _continueAfterUnexpectedArg;
 
-        public CommandLineApplication(bool throwOnUnexpectedArg = true, bool continueAfterUnexpectedArg = false)
+        private readonly bool _treatUnmatchedOptionsAsArguments;
+
+        public CommandLineApplication(bool throwOnUnexpectedArg = true, bool continueAfterUnexpectedArg = false, bool treatUnmatchedOptionsAsArguments = false)
```
Kahbazi: @anurse  Does this count as a breaking change? I don't know if I pick the best option on how to pass `treatUnmatchedOptionsAsArguments`. On the other hand I don't think many users are using `CommandLineApplication`.

anurse: No, this component is shared-source and not published to NuGet. There's a third-party (though it was written by a now-former team member) [fork of it](https://github.com/natemcmaster/CommandLineUtils), but this code shouldn't be visible outside of our repos and that fork has long-since drifted away from this source.

anurse:
I think this is good! We're having some quirks with the build pipeline because of the repo move, but I'll work on getting this merged.

Pilchie:
@mmitche resolved the issue with the pipelines.

Pilchie:
Note - might need an admin to override the pending `Extensions-ci` check :(

Issue #2149, opened 2019-08-07
BackgroundService blocked the execution of whole host


vova-lantsov-dev:
### Describe the bug
When I run the specific `foreach` cycle in `BackgroundService`-derived class, it blocks the whole host from starting. Even second hosted service doesn't start. When I comment `foreach` cycle, everything works as expected.

### To Reproduce
TargetFramework: netcoreapp2.1
Version: 2.1.12

Use following hosted service: [NotificationRunner.cs](https://gist.github.com/vova-lantsov-dev/7b3b820df1c7a6b86dac6bb0593c8155)
And singleton: [NotificationService.cs](https://gist.github.com/vova-lantsov-dev/788c5816a137a81e0c2af5f4655b21bc)

### Expected behavior
BackgroundService.ExecuteAsync should work in background without blocking even if it has blocking code. As you can see in [NotificationService](https://gist.github.com/vova-lantsov-dev/788c5816a137a81e0c2af5f4655b21bc) class, cancellation of enumerator is based on IApplicationLifetime.ApplicationStopping, but anyway it shouldn't affect the host startup because BackgroundService is expected to run in background :)

### Screenshots
When `foreach` cycle exists
![image](https://user-images.githubusercontent.com/45858087/62523649-ecafbf00-b83c-11e9-855e-e0ba541d0c3e.png)
Then execution is blocked on this cycle
![image](https://user-images.githubusercontent.com/45858087/62525002-4d3ffb80-b83f-11e9-8ce2-713dda7e79ad.png)

But when `foreach` cycle is commented
![image](https://user-images.githubusercontent.com/45858087/62525057-647ee900-b83f-11e9-84d0-47f896735b98.png)
Then execution continues as expected
![image](https://user-images.githubusercontent.com/45858087/62525220-ae67cf00-b83f-11e9-83a9-c7ca1c45b82c.png)
(But why twice?)

### Additional context
```
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-preview7-012821
 Commit:    6348f1068a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\

Host (useful for support):
  Version: 3.0.0-preview7-27912-14
  Commit:  4da6ee6450

.NET Core SDKs installed:
  2.1.700 [C:\Program Files\dotnet\sdk]
[...]

PR #2823, opened 2020-01-03
Allow ExecuteAsync cancellation propagation

### LINKED ISSUE #2811: BackgroundService should provide a linked token to ExecuteAsync
# Describe the bug

To allow cancellation occurring on start to flow through [ExecuteAsync](https://github.com/aspnet/Extensions/blob/master/src/Hosting/Abstractions/src/BackgroundService.cs#L33), ```BackgroundService``` should provide a linked token to ```ExecuteAsync```

### To Reproduce
1. Put ```await Task.Delay(Timeout.Infinite, stoppingToken);``` inside ```ExecuteAsync```
2. Do ```Ctrl+C```
3. App will wait indefinitely

### Expected behavior

* Cancel occurred

### Solution

```csharp
_stoppingCts  = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
_executingTask = ExecuteAsync(_stoppingCts.Token);
```

I can do a PR if needed

klettier:
 - Add test
 - Link the token provided to ```StartAsync``` to the one provided to the ```_executingTask ```

fixes #2811

### DIFF
--- a/src/Hosting/Abstractions/src/BackgroundService.cs
+++ b/src/Hosting/Abstractions/src/BackgroundService.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -13,7 +13,7 @@ namespace Microsoft.Extensions.Hosting
     public abstract class BackgroundService : IHostedService, IDisposable
     {
         private Task _executingTask;
-        private readonly CancellationTokenSource _stoppingCts = new CancellationTokenSource();
+        private CancellationTokenSource _stoppingCts;
 
         /// <summary>
         /// This method is called when the <see cref="IHostedService"/> starts. The implementation should return a task that represents
@@ -29,6 +29,9 @@ namespace Microsoft.Extensions.Hosting
         /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
         public virtual Task StartAsync(CancellationToken cancellationToken)
         {
+            // Create linked token
+            _stoppingCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
             // Store the task we're executing
             _executingTask = ExecuteAsync(_stoppingCts.Token);
 
--- a/src/Hosting/Abstractions/src/BackgroundService.cs
+++ b/src/Hosting/Abstractions/src/BackgroundService.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Extensions.Hosting
         /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
         public virtual Task StartAsync(CancellationToken cancellationToken)
         {
-            // Create linked token
+            // Create linked token to allow cancelling executing task from provided token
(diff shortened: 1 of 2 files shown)


src/Hosting/Abstractions/src/BackgroundService.cs
```diff
@@ -29,6 +29,9 @@ public abstract class BackgroundService : IHostedService, IDisposable
         /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
         public virtual Task StartAsync(CancellationToken cancellationToken)
         {
+            // Create linked token
```
Tratcher: This comment is redundant with the method name. How about explaining why?

src/Hosting/Abstractions/src/BackgroundService.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Extensions.Hosting
     public abstract class BackgroundService : IHostedService, IDisposable
     {
         private Task _executingTask;
-        private readonly CancellationTokenSource _stoppingCts = new CancellationTokenSource();
+        private CancellationTokenSource _stoppingCts;
```
Tratcher: Dispose now has to handle the possibility that StartAsync was never called to initialize _stoppingCts. It should be updated to `_stoppingCts?.Cancel();`.

sertunc:
Thanks a lot

klettier:
Guys, do you when this will be the released ?

I have currently a case (**with current nuget 3.1.0**) that prevent me from cancelling the ```RunAsync``` without this change.

Bellow you can find an example of code that will hang forever.

```csharp
[Fact]
public async Task Cancel()
{
    var cancellationTokenSource = new CancellationTokenSource();

    await Host.CreateDefaultBuilder()
                {
                    services.AddHostedService(s => new CancellerService(cancellationTokenSource));
                })
              .Build()
              .RunAsync(cancellationTokenSource.Token);
}

class CancellerService : BackgroundService
{
    private readonly CancellationTokenSource tokenSource;
    System.Threading.Timer timer;

    public CancellerService(CancellationTokenSource tokenSource)
    {
        this.tokenSource = tokenSource;
        timer = new System.Threading.Timer(TimerCallback, null, TimeSpan.FromSeconds(5), TimeSpan.Zero);
    }

    void TimerCallback(object state)
    {
        tokenSource.Cancel();
    }

    protected override Task ExecuteAsync(CancellationToken stoppingToken)
    {
        var blocking = new BlockingCollection<int>();
        
        blocking.Take(stoppingToken);

        return Task.CompletedTask;
    }

    public override Task StopAsync(CancellationToken cancellationToken)
    {
        return base.StopAsync(cancellationToken);
    }
}
```

Thx

Tratcher:
This will be included in the next major release (5.0) that has its first preview coming soon. The bar is very high for adding any patches to 3.1, it has to be a high impact issue with no workaround.

A workaround for the example above would be to add `await Task.Yield()` at the start of ExecuteAsync. That's a known issue we're addressing separately. https://github.com/dotnet/extensions/issues/2149

klettier:
> This will be included in the next major release (5.0) that has its first preview coming soon. The bar is very high for adding any patches to 3.1, it has to be a high impact issue with no workaround.
> 
> A workaround for the example above would be to add `await Task.Yield()` at the start of ExecuteAsync. That's a known issue we're addressing separately. #2149

Thx a lot

PR #2824, opened 2020-01-04
Remove async await from BackgroundService.StopAsync

Kahbazi:


davidfowl:
Why?

Kahbazi:
To improve performance by removing async state machine. But I was thinking out of context, now that you've asked, I think it's not that important because it's only called once at the end of application lifecycle. :confused:

Tratcher:
Agreed.

PR #2829, opened 2020-01-06
Add case sensitivity test for options

HaoK:


HaoK:
Minor PR adding a test to explicitly show that named options are case sensitive, from doc issue: https://github.com/aspnet/AspNetCore.Docs/pull/16439#event-2928444842

PR #2844, opened 2020-01-10
Guard for Custom Implementation of IHostApplicationLifetime 

### LINKED ISSUE #2825: NullReferenceException with custom implementation of IHostApplicationLifetime
### Describe the bug
A custom implementation of `IHostApplicationLifetime` causes a `NullReferenceException`.

### To Reproduce
```csharp
public class Program
{
    public static void Main(string[] args)
    {
        CreateHostBuilder(args).Build().Run();
    }

    public static IHostBuilder CreateHostBuilder(string[] args) =>
        Host.CreateDefaultBuilder(args)
            .ConfigureServices((hostContext, services) =>
            {
                services.AddSingleton<IHostApplicationLifetime, CustomApplicationLifetime>();
                services.AddHostedService<Worker>();

            });
}

public class CustomApplicationLifetime : IHostApplicationLifetime
{
    public CancellationToken ApplicationStarted => default;

    public CancellationToken ApplicationStopping => default;

    public CancellationToken ApplicationStopped => default;

    public void StopApplication()
    {
    }
}
```

### Additional context
This is where error is happening.
https://github.com/aspnet/Extensions/blob/ecc4d9fe9381ac65848968b672eb6098ec2a2cae/src/Hosting/Hosting/src/Internal/Host.cs#L39

There's no null checking for `_applicationLifetime`. In order to fix this we could store `IHostApplicationLifetime` instead of `ApplicationLifetime` and use cast in `StartAsync(...)` and `StopAsync(...)`

Kahbazi:
- Added a gaurd for IHostApplicationLifetime
- Removed null checks

Fixes #2825

### DIFF
--- a/src/Hosting/Hosting/src/Internal/Host.cs
+++ b/src/Hosting/Hosting/src/Internal/Host.cs
@@ -25,6 +25,10 @@ namespace Microsoft.Extensions.Hosting.Internal
         {
             Services = services ?? throw new ArgumentNullException(nameof(services));
             _applicationLifetime = (applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime))) as ApplicationLifetime;
+            if (_applicationLifetime is null)
+            {
+                throw new ArgumentException("The only acceptable implementation of Microsoft.Extensions.Hosting.IHostApplicationLifetime is Microsoft.Extensions.Hosting.Internal.ApplicationLifetime", nameof(applicationLifetime));
+            }
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _hostLifetime = hostLifetime ?? throw new ArgumentNullException(nameof(hostLifetime));
             _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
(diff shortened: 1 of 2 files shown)


src/Hosting/Hosting/src/Internal/Host.cs
```diff
@@ -25,6 +25,10 @@ internal class Host : IHost, IAsyncDisposable
         {
             Services = services ?? throw new ArgumentNullException(nameof(services));
             _applicationLifetime = (applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime))) as ApplicationLifetime;
+            if (_applicationLifetime is null)
```
Tratcher: `is null` vs `== null`?

src/Hosting/Hosting/src/Internal/Host.cs
```diff
@@ -25,6 +25,10 @@ internal class Host : IHost, IAsyncDisposable
...
+            {
+                throw new ArgumentException("The only acceptable implementation of Microsoft.Extensions.Hosting.IHostApplicationLifetime is Microsoft.Extensions.Hosting.Internal.ApplicationLifetime", nameof(applicationLifetime));
```
Tratcher:
```suggestion
                throw new ArgumentException("Replacing IHostApplicationLifetime is not supported.", nameof(applicationLifetime));
```

Tratcher:
Thanks

PR #2882, opened 2020-01-18
Revert "Remove ref project from Microsoft.AspNetCore.Testing"

wtgodbe:
Reverts dotnet/extensions#2817

This broke AspNetCore (see https://github.com/dotnet/aspnetcore/pull/18434):

> HttpsTests.cs(118,10): error CS1705: Assembly 'Microsoft.Extensions.Logging.Testing' with identity 'Microsoft.Extensions.Logging.Testing, Version=3.1.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' uses 'Microsoft.AspNetCore.Testing, Version=3.1.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' which has a higher version than referenced assembly 'Microsoft.AspNetCore.Testing' with identity 'Microsoft.AspNetCore.Testing, Version=3.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' [/Users/runner/runners/2.163.1/work/1/s/src/Servers/Kestrel/test/InMemory.FunctionalTests/InMemory.FunctionalTests.csproj]

Reverting it in order to get a candidate build for 3.1.2. We can always un-revert it if there's an obvious fix that lets us take this change without jeopardizing the release schedule.

CC @JunTaoLuo @dougbu @Pilchie @mmitche


dougbu:
Huh? @JunTaoLuo did this to fix a problem that sounds a lot like dotnet/aspnetcore#18434. Would prefer to revert the workaround for the same problem checked into dotnet/aspnetcore.

Put another way, the "Remove ref project from Microsoft.AspNetCore.Testing" change was correct and addressed more than just a useless project.

wtgodbe:
> Huh? @JunTaoLuo did this to fix a problem that sounds a lot like dotnet/aspnetcore#18434. Would prefer to revert the workaround for the same problem checked into dotnet/aspnetcore.

@dougbu do you know offhand what the workaround in AspNetCore is? I'd also prefer to revert that, but wasn't able to remember what (if anything) we did to workaround this in AspNetCore, which presumably caused the failure in that repo once we ingested John's change

dougbu:
> do you know offhand 

No, it's not coming to me. I remember lots of warning suppressions but there was something else (perhaps instead of the warning suppressions?). I'll look tomorrow if you haven't figured it out by then.

PR #2894, opened 2020-01-22
Pin Cryptography.Cng

JunTaoLuo:
We decided to pin Cng as part of the discussion at https://github.com/dotnet/aspnetcore/pull/18364#discussion_r368208938.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -38,9 +38,9 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>4ac4c0367003fe3973a3648eb0715ddb0e3bbcea</Sha>
     </Dependency>
-    <Dependency Name="System.Security.Cryptography.Cng" Version="4.6.1" CoherentParentDependency="Microsoft.NETCore.App.Runtime.win-x64">
+    <Dependency Name="System.Security.Cryptography.Cng" Version="4.6.0" Pinned="true">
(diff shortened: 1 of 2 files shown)


eng/Version.Details.xml
```diff
@@ -38,9 +38,9 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>4ac4c0367003fe3973a3648eb0715ddb0e3bbcea</Sha>
     </Dependency>
-    <Dependency Name="System.Security.Cryptography.Cng" Version="4.6.1" CoherentParentDependency="Microsoft.NETCore.App.Runtime.win-x64">
+    <Dependency Name="System.Security.Cryptography.Cng" Version="4.6.0" Pinned="true">
```
JunTaoLuo: Does this need to be moved as per: https://github.com/dotnet/aspnetcore/pull/18364#discussion_r369765860

dougbu: @mmitche are you implying by your approval that we do **not** need to move pinned dependencies below non-pinned ones from the same repo anymore?

wtgodbe: I believe @mmitche has pushed a fix that renders the ordering unnecessary here - can you confirm Matt?

wtgodbe: He confirmed in the AspNetCore PR, so this should be good to go

PR #2899, opened 2020-01-23
[release/3.0] Pin dependency on Microsoft.Win32.Registry

wtgodbe:
Part of https://github.com/dotnet/aspnetcore/issues/18519

We depend on this package in order to get the ref assembly out of it, which will not be serviced. We should pin the dependency in order to avoid a conflict between the implementation assembly in the package from patch `x.y.z+1`, and the shared framework from patch `x.y.z+2` (where the package is not patched).

CC @JunTaoLuo @dougbu @ericstj @nguerrera @mmitche 

Also CC @Pilchie to confirm we can take this as tell-mode for March

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -10,7 +10,7 @@
 -->
 <Dependencies>
   <ProductDependencies>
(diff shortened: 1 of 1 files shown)


Pilchie:
> Also CC @Pilchie to confirm we can take this as tell-mode for March

Yes, sounds good to me.

PR #2904, opened 2020-01-24
fix typo from "retruned" to "returned" in HealthReport.Entries XML Comments

npnelson:
fixed typo in XML comments


Pilchie:
Thanks!

PR #2928, opened 2020-02-03
[release/2.1] Switch to Windows Server queues

dougbu:


dougbu:
@Pilchie :ok: to use tell-mode for this in our next release? And, what are your thoughts on doing the same thing for the next 3.0.x?

Pilchie:
Yes to tell-mode.  I would do it for 3.0.x if we end up servicing 3.0.x again, but hopefully we won't. 🤞

PR #2940, opened 2020-02-05
[Host] Allow disabling reloadOnChange for Host's CreateDefaultBuilder

kjerk:
* Add ASPNETCORE_BUILDER_CONFIG_RELOAD environment flag to allow disabling live reload in builder.
* Add two tests to assert two different cases (live reload on, live reload off) work.

Addresses https://github.com/dotnet/extensions/issues/2939

### DIFF
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
@@ -71,9 +70,8 @@ namespace Microsoft.Extensions.Hosting
             builder.ConfigureAppConfiguration((hostingContext, config) =>
             {
                 var env = hostingContext.HostingEnvironment;
-
-                var reloadOnChange = Convert.ToBoolean(
-                    Environment.GetEnvironmentVariable("ASPNETCORE_BUILDER_CONFIG_RELOAD") ?? "true");
+                
+                var reloadOnChange = hostingContext.Configuration.GetValue("HOSTBUILDER_CONFIG_RELOAD", true);
 
                 config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: reloadOnChange)
                       .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: reloadOnChange);
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Extensions.Hosting
             {
                 var env = hostingContext.HostingEnvironment;
                 
-                var reloadOnChange = hostingContext.Configuration.GetValue("HOSTBUILDER_CONFIG_RELOAD", true);
+                var reloadOnChange = hostingContext.Configuration.GetValue("hostbuilder:configreload", true);
 
                 config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: reloadOnChange)
                       .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: reloadOnChange);
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
@@ -71,8 +72,11 @@ namespace Microsoft.Extensions.Hosting
             {
                 var env = hostingContext.HostingEnvironment;
 
-                config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
-                      .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: true);
+                var reloadOnChange = Convert.ToBoolean(
+                    Environment.GetEnvironmentVariable("ASPNETCORE_BUILDER_CONFIG_RELOAD") ?? "true");
+
+                config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: reloadOnChange)
+                      .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: reloadOnChange);
 
                 if (env.IsDevelopment() && !string.IsNullOrEmpty(env.ApplicationName))
                 {
--- a/src/Hosting/Hosting/src/Host.cs
+++ b/src/Hosting/Hosting/src/Host.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Extensions.Hosting
             {
                 var env = hostingContext.HostingEnvironment;
                 
-                var reloadOnChange = hostingContext.Configuration.GetValue("hostbuilder:configreload", true);
+                var reloadOnChange = hostingContext.Configuration.GetValue("hostBuilder:reloadConfigOnChange", defaultValue: true);
 
                 config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: reloadOnChange)
                       .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: reloadOnChange);
--- a/src/Hosting/Hosting/test/HostTests.cs
+++ b/src/Hosting/Hosting/test/HostTests.cs
@@ -150,7 +150,7 @@ namespace Microsoft.Extensions.Hosting
 
             config.GetReloadToken().RegisterChangeCallback(o =>
             {
-                //configReloadedCancelTokenSource.Cancel();
+                configReloadedCancelTokenSource.Cancel();
             }, null);
             // Wait for up to 10 seconds, if config reloads at any time, cancel the wait.
             await Task.WhenAny(Task.Delay(10000, configReloadedCancelToken)); // Task.WhenAny ignores the task throwing on cancellation.
--- a/src/Hosting/Hosting/test/HostTests.cs
+++ b/src/Hosting/Hosting/test/HostTests.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics.Tracing;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -83,9 +84,9 @@ namespace Microsoft.Extensions.Hosting
         }
 
         [Fact]
-        public void CreateDefaultBuilder_ConfigJsonDoesNotReload()
+        public async Task CreateDefaultBuilder_ConfigJsonDoesNotReload()
         {
-            var reloadFlagConfig = new Dictionary<string, string>() {{ "HOSTBUILDER_CONFIG_RELOAD", "false" }};
+            var reloadFlagConfig = new Dictionary<string, string>() {{ "hostbuilder:configreload", "false" }};
             var appSettingsPath = Path.Combine(Path.GetTempPath(), "appsettings.json");
 
             string SaveRandomConfig()
@@ -110,15 +111,15 @@ namespace Microsoft.Extensions.Hosting
             Assert.Equal(dynamicConfigMessage1, config["Hello"]);
 
             var dynamicConfigMessage2 = SaveRandomConfig();
-            Task.Delay(1000).Wait(); // Give reload time to fire if it's going to.
+            await Task.Delay(1000); // Give reload time to fire if it's going to.
             Assert.NotEqual(dynamicConfigMessage1, dynamicConfigMessage2); // Messages are different.
             Assert.Equal(dynamicConfigMessage1, config["Hello"]); // Config did not reload
         }
 
         [Fact]
-        public void CreateDefaultBuilder_ConfigJsonDoesReload()
+        public async Task CreateDefaultBuilder_ConfigJsonDoesReload()
         {
-            var reloadFlagConfig = new Dictionary<string, string>() { { "HOSTBUILDER_CONFIG_RELOAD", "true" } };
+            var reloadFlagConfig = new Dictionary<string, string>() { { "hostbuilder:configreload", "true" } };
             var appSettingsPath = Path.Combine(Path.GetTempPath(), "appsettings.json");
 
             string SaveRandomConfig()
@@ -143,7 +144,16 @@ namespace Microsoft.Extensions.Hosting
             Assert.Equal(dynamicConfigMessage1, config["Hello"]);
 
             var dynamicConfigMessage2 = SaveRandomConfig();
-            Task.Delay(1000).Wait(); // Give reload time to fire if it's going to.
+
+            var configReloadedCancelTokenSource = new CancellationTokenSource();
+            var configReloadedCancelToken = configReloadedCancelTokenSource.Token;
+
+            config.GetReloadToken().RegisterChangeCallback(o =>
+            {
+                //configReloadedCancelTokenSource.Cancel();
+            }, null);
+            // Wait for up to 10 seconds, if config reloads at any time, cancel the wait.
+            await Task.WhenAny(Task.Delay(10000, configReloadedCancelToken)); // Task.WhenAny ignores the task throwing on cancellation.
             Assert.NotEqual(dynamicConfigMessage1, dynamicConfigMessage2); // Messages are different.
             Assert.Equal(dynamicConfigMessage2, config["Hello"]); // Config DID reload from disk
         }
--- a/src/Hosting/Hosting/test/HostTests.cs
+++ b/src/Hosting/Hosting/test/HostTests.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Extensions.Hosting
         [Fact]
         public async Task CreateDefaultBuilder_ConfigJsonDoesNotReload()
         {
-            var reloadFlagConfig = new Dictionary<string, string>() {{ "hostbuilder:configreload", "false" }};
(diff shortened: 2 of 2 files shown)


src/Hosting/Hosting/src/Host.cs
```diff
@@ -71,8 +72,11 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
             {
                 var env = hostingContext.HostingEnvironment;
 
-                config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
-                      .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: true);
+                var reloadOnChange = Convert.ToBoolean(
+                    Environment.GetEnvironmentVariable("ASPNETCORE_BUILDER_CONFIG_RELOAD") ?? "true");
```
Tratcher: Rather than directly reading an environment variable, how about reading from the host's config (which does include environment variables already, but could also be configured other ways). See hostingContext.Configuration. https://github.com/dotnet/extensions/blob/24751691256010d4fac8d16afc56e6d9db12c00a/src/Hosting/Abstractions/src/HostBuilderContext.cs#L27
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#configure-the-host-builder-with-configurehostconfiguration

kjerk: Oh cool, I was worried about config not being 'ripe' by this point, but that does make sense since it's pulled in already, thanks.

Pushing an update that
* Switches option pull from `Environment` to `hostingContext.Configuration`
* Changes flag name to `DOTNET_HOSTBUILDER_CONFIG_RELOAD` to disambiguate and fit the [config.AddEnvironmentVariables prefix](https://github.com/dotnet/extensions/blob/master/src/Hosting/Hosting/src/Host.cs#L63)
* Update tests to use `DOTNET_HOSTBUILDER_CONFIG_RELOAD`

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -85,21 +85,23 @@ public void CreateDefaultBuilder_EnablesValidateOnBuild()
         [Fact]
         public void CreateDefaultBuilder_ConfigJsonDoesNotReload()
         {
-            Environment.SetEnvironmentVariable("ASPNETCORE_BUILDER_CONFIG_RELOAD", "false");
+            Environment.SetEnvironmentVariable("DOTNET_HOSTBUILDER_CONFIG_RELOAD", "false");
```
Tratcher: Tests shouldn't alter environment variables, it can have side-effects on other tests. Since we're using the config system now provide the key via AddInMemoryCollection.
https://github.com/dotnet/extensions/blob/126381c5526d59eb573386ffc642820091555c32/src/Hosting/Hosting/test/HostBuilderTests.cs#L39-L45

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -81,6 +82,56 @@ public void CreateDefaultBuilder_EnablesValidateOnBuild()
             Assert.Throws<AggregateException>(() => hostBuilder.Build());
         }
 
+        [Fact]
+        public void CreateDefaultBuilder_ConfigJsonDoesNotReload()
+        {
+            Environment.SetEnvironmentVariable("DOTNET_HOSTBUILDER_CONFIG_RELOAD", "false");
+            Environment.CurrentDirectory = Path.GetTempPath();
```
Tratcher: Changing CurrentDirectory is also not a good idea. Why did you need to do that? So you'd have somewhere to write your appsettings.json? It would be better to change the host's ContentRoot to the temp path.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.1#contentrootpath

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -81,6 +82,56 @@ public void CreateDefaultBuilder_EnablesValidateOnBuild()
...
+            }
+
+            var dynamicConfigMessage1 = SaveRandomConfig();
+            var host = Host.CreateDefaultBuilder().Build();
+            var config = host.Services.GetRequiredService<IConfiguration>();
+
+            Assert.Equal(dynamicConfigMessage1, config["Hello"]);
+
+            var dynamicConfigMessage2 = SaveRandomConfig();
+            Task.Delay(1000).Wait(); // Give reload time to fire if it's going to.
```
Tratcher: This is going to be unreliable. You can directly register for the change callback.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfiguration.getreloadtoken?view=dotnet-plat-ext-3.1#Microsoft_Extensions_Configuration_IConfiguration_GetReloadToken

src/Hosting/Hosting/src/Host.cs
```diff
@@ -70,9 +70,11 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
             builder.ConfigureAppConfiguration((hostingContext, config) =>
             {
                 var env = hostingContext.HostingEnvironment;
+                
+                var reloadOnChange = hostingContext.Configuration.GetValue("HOSTBUILDER_CONFIG_RELOAD", true);
```
Tratcher: The host uses top level keys like `contentRoot`.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.1#contentrootpath
`configReload` should be adequate here.

src/Hosting/Hosting/test/HostTests.cs
kjerk: I will update the config to polyfill from a config source like an in memory collection instead as you are suggesting, but this will then drop the assertion that DOTNET_HOSTBUILDER_CONFIG_RELOAD works, and moves it to being untested, and unverifiable. I'm not sure of a good solution here as this means that any tests that would assert that `config.AddEnvironmentVariables(prefix: "DOTNET_");` works in `CreateDefaultBuilder()`, or anywhere else are untestable :(

src/Hosting/Hosting/test/HostTests.cs
kjerk: Yes just trying to sandbox. I didn't know content root would redirect configuration flatfile loading, I was just worried about it since the filepath is [hard coded without prefix](https://github.com/dotnet/extensions/blob/master/src/Hosting/Hosting/src/Host.cs#L74-L75). Thanks, I'll update this.

src/Hosting/Hosting/test/HostTests.cs
kjerk: This is where we're going to run into a hard problem, the feature is testing that we've disabled (default builder's) filesystem modification reloads, so even if I subscribe to the notification being raised, that's already downhill from the filesystem notification needing to have triggered, or not, and I'm still going to have to wait X time for that event right?

I can assert that the event fired, but since the backend is asynchronous, in the instance that we're trying to assert that this did happen, it's not only the same, but I need to wait anyway, otherwise what I've done is test immediately before the change had time to propagate, and in the case we're trying to assert that the filesystem event did not fire, then I still need to have waited a reasonable but non-making-the-tests-unbearable amount of time for it to have done so.

src/Hosting/Hosting/src/Host.cs
kjerk: Is there some prefix I can add to that to disambiguate? I was trying to follow the `{ASPNETCORE/DOTNET}_FEATURE_DESCRIPTION` nomenclature of other aspnetcore and dotnet flags sort of like [DOTNET_USE_POLLING_FILE_WATCHER](https://github.com/dotnet/aspnetcore/blob/6cc9f6f130af4ed0e7f321b144265cfbcec0ceee/src/Tools/dotnet-watch/src/CommandLineOptions.cs#L26)

This would mean that `DOTNET_CONFIGRELOAD` specified in an environment variable would trip the behavior added in this PR, and that to me sounds like it's entirely disabling config reloading, which is not quite the case, as one's own calls to `AddJsonFile` are left alone.

Tratcher: We can at least use the hierarchical config pattern.
Also, config keys are not case sensitive. Changed for readability.
```suggestion
            var reloadFlagConfig = new Dictionary<string, string>() {{ "hostbuilder:configreload", "false" }};
```

src/Hosting/Hosting/test/HostTests.cs
Tratcher: Fair point. The test for disabled reload can wait one second and ensure check that config did not change. Just make sure to do so properly: (.Wait() can cause threading issues even in the tests.)
```suggestion
            await Task.Delay(1000); // Give reload time to fire if it's going to.
```

For the positive test we must use the event and a long timeout. The test machines have strange load characteristics and are subject to 1-10s random delay.

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -81,6 +82,72 @@ public void CreateDefaultBuilder_EnablesValidateOnBuild()
             Assert.Throws<AggregateException>(() => hostBuilder.Build());
         }
 
+        [Fact]
+        public void CreateDefaultBuilder_ConfigJsonDoesNotReload()
```
Tratcher:
```suggestion
        public async Task CreateDefaultBuilder_ConfigJsonDoesNotReload()
```

src/Hosting/Hosting/test/HostTests.cs
kjerk: Okay cool, that makes sense. I tried to hybridize this approach you mentioned. I increased the timeout to be 10 seconds since as you mentioned this can be volatile, but used the ReloadToken to cancel the Task if the ReloadTokoen fires, since we know that means we're ready to proceed regardless.

src/Hosting/Hosting/src/Host.cs
kjerk: Sounds good, this has been updated in my most recent commit https://github.com/dotnet/extensions/pull/2940/commits/de16b88781a0a8171a6268a43f8b1211679bdc8b

src/Hosting/Hosting/test/HostTests.cs
kjerk: Changed, thanks.

kjerk:
Sorry it took a while to get back around to this one, and thanks for the feedback. I pushed https://github.com/dotnet/extensions/pull/2940/commits/de16b88781a0a8171a6268a43f8b1211679bdc8b to resolve (I think) any dangling parts from the conversations.

Changelist
* Update config key to `hostbuilder:configreload`
* Changed both tests to `async Task` instead of being non async methods.
* For the positive test case, increase wait time on test, listen for `config.GetReloadToken()` to fire and if it does, cancel any waiting and proceed.
* Cancelling a task throws, so rather than write a full try/catch for the whole thing, wrap `Task.Delay()` in `Task.WhenAny()` which will ignore the throw and function as normal to keep it terse.

src/Hosting/Hosting/src/Host.cs
```diff
@@ -70,9 +70,11 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
             builder.ConfigureAppConfiguration((hostingContext, config) =>
             {
                 var env = hostingContext.HostingEnvironment;
+                
+                var reloadOnChange = hostingContext.Configuration.GetValue("hostbuilder:configreload", true);
```
Tratcher:
```suggestion
                var reloadOnChange = hostingContext.Configuration.GetValue("hostbuilder:configreload", defaultValue: true);
```

anurse: Maybe change the name a little too? Descriptive is better than terse IMO. `hostBuilder:reloadConfigOnChange` ?

Tratcher: I'll take care of that and get this merged.

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -81,6 +83,81 @@ public void CreateDefaultBuilder_EnablesValidateOnBuild()
             Assert.Throws<AggregateException>(() => hostBuilder.Build());
         }
 
+        [Fact]
+        public async Task CreateDefaultBuilder_ConfigJsonDoesNotReload()
+        {
+            var reloadFlagConfig = new Dictionary<string, string>() {{ "hostbuilder:configreload", "false" }};
```
Tratcher:
```suggestion
            var reloadFlagConfig = new Dictionary<string, string>() {{ "hostbuilder:reloadConfigOnChange", "false" }};
```

src/Hosting/Hosting/test/HostTests.cs
```diff
@@ -81,6 +83,81 @@ public void CreateDefaultBuilder_EnablesValidateOnBuild()
...
+            var dynamicConfigMessage2 = SaveRandomConfig();
+            await Task.Delay(1000); // Give reload time to fire if it's going to.
+            Assert.NotEqual(dynamicConfigMessage1, dynamicConfigMessage2); // Messages are different.
+            Assert.Equal(dynamicConfigMessage1, config["Hello"]); // Config did not reload
+        }
+
+        [Fact]
+        public async Task CreateDefaultBuilder_ConfigJsonDoesReload()
+        {
+            var reloadFlagConfig = new Dictionary<string, string>() { { "hostbuilder:configreload", "true" } };
```
Tratcher:
```suggestion
            var reloadFlagConfig = new Dictionary<string, string>() { { "hostbuilder:reloadConfigOnChange", "true" } };
```

Tratcher:
Thanks @kjerk

kjerk:
> Thanks @kjerk

Thanks everybody.
Seems like a small change but this helps our prod deployments in Centos a lot.

davidfowl:
The name of this configuration entry isn't ideal. Maybe it should be host and not hostbuilder?

src/Hosting/Hosting/src/Host.cs
```diff
@@ -70,9 +70,11 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
             builder.ConfigureAppConfiguration((hostingContext, config) =>
             {
                 var env = hostingContext.HostingEnvironment;
+                
+                var reloadOnChange = hostingContext.Configuration.GetValue("hostBuilder:reloadConfigOnChange", defaultValue: true);
```
am11, line 74: Should this also be set to https://github.com/dotnet/runtime/blob/79ae74f5ca5c8a6fe3a48935e85bd7374959c570/src/libraries/Microsoft.Extensions.Configuration.FileExtensions/src/FileConfigurationSource.cs#L33?

Apparently this condition is true: https://github.com/dotnet/runtime/blob/79ae74f5ca5c8a6fe3a48935e85bd7374959c570/src/libraries/Microsoft.Extensions.Configuration.FileExtensions/src/FileConfigurationProvider.cs#L34 even with `"hostBuilder": {"reloadConfigOnChange": false }` in appsettings.json.

kernel implementation without inotify support throw the following exception on application startup and there seems to be no way to turn (this quite an <ins>un</ins>important) feature off:

```
Unhandled exception. System.IO.IOException: Function not implemented
   at System.IO.FileSystemWatcher.StartRaisingEvents()
   at System.IO.FileSystemWatcher.StartRaisingEventsIfNotDisposed()
   at System.IO.FileSystemWatcher.set_EnableRaisingEvents(Boolean value)
   at Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.TryEnableFileSystemWatcher()
   at Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.CreateFileChangeToken(String filter)
   at Microsoft.Extensions.FileProviders.PhysicalFileProvider.Watch(String filter)
   at Microsoft.Extensions.Configuration.FileConfigurationProvider.<.ctor>b__1_0()
   at Microsoft.Extensions.Primitives.ChangeToken.OnChange(Func`1 changeTokenProducer, Action changeTokenConsumer)
   at Microsoft.Extensions.Configuration.FileConfigurationProvider..ctor(FileConfigurationSource source)
   at Microsoft.Extensions.Configuration.Json.JsonConfigurationSource.Build(IConfigurationBuilder builder)
   at Microsoft.Extensions.Configuration.ConfigurationBuilder.Build()
   at Microsoft.Extensions.Hosting.HostBuilder.BuildAppConfiguration()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at mymvc1.Program.Main(String[] args)
```

davidfowl:
What did we merge here? Did we change the key to host vs hostBuilder?

src/Hosting/Hosting/src/Host.cs
am11: got the global opt-out knob working with: https://github.com/dotnet/runtime/commit/457ed114aebe9798fd58bbf9733194e51b734896; `DOTNET_USE_POLLING_FILE_WATCHER=1` in unikernel environment completely disables inotify. sadly, it wasn't backported to net5 so i switched to net6..

kjerk: At least for HostBuilder, at the time this was being changed if you used CreateDefaultBuilder() you were getting live reloading whether you wanted it or not because it was [hard coded to true](https://github.com/dotnet/extensions/pull/2940/files/f0d48a3cae6fabe0712bf0601e1b7b312d3d3b5e#diff-0c3da3ec1ae00a375bab50fb4a1a452d745345c234877033508add5b21ffb7e1L74-R77)

If Source.ReloadOnChange is populated by an env var and is available by the time ConfigureAppConfiguration() fires then that would work a bit better.

kjerk:
Looking back at this it looks like the config change was mentioned a few months after it was merged/closed (Mar 19).

I think iirc at the time when [referencing the docs, while _Host_.blah](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-5.0#set-up-a-host) was used as a static factory method, most things were referencing IHostBuilder, CreateHostBuilder, CreateDefaultBuilder, etc etc, and the only other ENV var override I could find referenced the implementation class, so I just defaulted to hostBuilder, though host is a bit nicer and agnostic sounding.

To make the change would require subsequent changes and maybe another PR?

Also because I myself had trouble finding the tiny link at the top of this PR, the original issue this is resolving is here: https://github.com/dotnet/extensions/issues/2939 which gives some details on what problem this was addressing.

Tratcher:
https://github.com/dotnet/runtime/blob/02ad981eab46339b6abef5b97b3db761d22b6455/src/libraries/Microsoft.Extensions.Hosting/src/HostingHostBuilderExtensions.cs#L237
It would be `DOTNET_hostBuilder_reloadConfigOnChange=true`

PR #2941, opened 2020-02-06
Disposing CTS on cancel 

zubyk:
Call CancellationTokenSource.Dispose() on task stop.
Link cancellationToken to exists _stoppingCts  if you call StartAsync() when task already started.

### DIFF
--- a/src/Hosting/Abstractions/src/BackgroundService.cs
+++ b/src/Hosting/Abstractions/src/BackgroundService.cs
@@ -29,26 +29,49 @@ namespace Microsoft.Extensions.Hosting
         /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
         public virtual Task StartAsync(CancellationToken cancellationToken)
         {
-            // Only if task not started before
+            CancellationToken stoppingCtsToken;
+
+            // Only if _stoppingCts not created before or created and canceled
             if (_stoppingCts?.IsCancellationRequested != false)
             {
-                // Create linked token to allow cancelling executing task from provided token
-                _stoppingCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                _stoppingCts = CancellationTokenSource.CreateLinkedTokenSource();
+
+                stoppingCtsToken = _stoppingCts.Token;
 
                 // Clear resources on cancel
-                _stoppingCts.Token.Register(() =>
+                stoppingCtsToken.Register(() =>
                 {
                     _stoppingCts.Dispose();
                 });
             }
             else
             {
-                // link token to allow cancelling executing task from provided token
-                cancellationToken.Register(() => _stoppingCts.Cancel());
+                try
+                {
+                    // If _stoppingCts already disposed
+                    stoppingCtsToken = _stoppingCts.Token;
+                }
+                catch (ObjectDisposedException)
+                {
+                    return Task.CompletedTask;
+                }
             }
 
+            // link token to allow cancelling executing task from provided token
+            cancellationToken.Register(() =>
+            {
+                try
+                {
+                    // If _stoppingCts already disposed
+                    _stoppingCts.Cancel();
+                }
+                catch (ObjectDisposedException)
+                {
+                }
+            });
+            
             // Store the task we're executing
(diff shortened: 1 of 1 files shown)


src/Hosting/Abstractions/src/BackgroundService.cs
```diff
@@ -29,11 +29,49 @@ public abstract class BackgroundService : IHostedService, IDisposable
...
 
+            // Only if _stoppingCts not created before or created and canceled
+            if (_stoppingCts?.IsCancellationRequested != false)
+            {
+                _stoppingCts = CancellationTokenSource.CreateLinkedTokenSource();
+
+                stoppingCtsToken = _stoppingCts.Token;
+
+                // Clear resources on cancel
+                stoppingCtsToken.Register(() =>
```
Tratcher: It's not necessary to dispose a token that's already cancelled (or that's in the process of being cancelled), it just causes unnecessary ObjectDisposedExceptions for everyone.

zubyk: hmm ... in all my last code I relied on this article [https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.dispose?view=netcore-3.1](url):
_Always call Dispose before you release your last reference to the CancellationTokenSource. Otherwise, the resources it is using will not be freed until the garbage collector calls the CancellationTokenSource object's Finalize method._

Since CTS is never released in the original code, this could cause potential memory leaks.
Also in original code In the StartAsync() call CTS always rewriten with new instance (and new executedTask). And when you call StopAsync() you cancel only last CTS and last created executedTask.
Please correct me if I`m wrong...

Tratcher: Calling Dispose is important if you're not going to call Cancel because yes it can leak memory (all the registered callbacks). Once cancelled, all the registered callbacks are released.

zubyk: I checked the sources of CTS, the Dispose() method clears other resources than Cancel().
But yes. I think you're right. This is not an infrastructure problem. Even if no one calls Dispose(), the GC can collect garbage when the CTS finalizer is called.

What do you think about:

> Also in original code In the StartAsync() call CTS always rewriten with new instance (and new executedTask). And when you call StopAsync() you cancel only last CTS and last created executedTask.

need to fix this?
maybe for BackgroundService.cs this is not such a big problem, since usually StartAsync() is called once.

Tratcher: BackgroundService is not designed to be started more than once, so that's not a concern.

src/Hosting/Abstractions/src/BackgroundService.cs
```diff
@@ -29,11 +29,49 @@ public abstract class BackgroundService : IHostedService, IDisposable
...
         public virtual Task StartAsync(CancellationToken cancellationToken)
         {
-            // Create linked token to allow cancelling executing task from provided token
-            _stoppingCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            CancellationToken stoppingCtsToken;
 
+            // Only if _stoppingCts not created before or created and canceled
+            if (_stoppingCts?.IsCancellationRequested != false)
+            {
+                _stoppingCts = CancellationTokenSource.CreateLinkedTokenSource();
```
zubyk: shame, shame on me! this line need to fix `_stoppingCts = new CancellationTokenSource();`

src/Hosting/Abstractions/src/BackgroundService.cs
zubyk: ok. Last try. maybe rewrite Dispose() with 
`public virtual void Dispose()
        {
                using(_stoppingCts)
                {
                      _stoppingCts?.Cancel();
                }
        }`

Tratcher: *shrug* not sure it accomplishes anything since the token gets cancelled anyways.

zubyk:
@Tratcher: It's not necessary to dispose a token that's already cancelled (or that's in the process of being cancelled)

PR #2954, opened 2020-02-10
Changed volatile variable in Redis cache

rmja:
...from _connection to _cache. 

 - The declared volatile variable is currently _connection, but the variable that can gain from being volatile is _cache, as that is the one that is tested in Connect(Async).

None

### DIFF
--- a/src/Caching/StackExchangeRedis/src/RedisCache.cs
+++ b/src/Caching/StackExchangeRedis/src/RedisCache.cs
@@ -29,8 +29,8 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
         private const string DataKey = "data";
         private const long NotPresent = -1;
 
-        private volatile ConnectionMultiplexer _connection;
-        private IDatabase _cache;
+        private ConnectionMultiplexer _connection;
+        private volatile IDatabase _cache;
 
         private readonly RedisCacheOptions _options;
         private readonly string _instance;


anurse:
> * The declared volatile variable is currently _connection, but the variable that can gain from being volatile is _cache, as that is the one that is tested in Connect(Async).

Can you clarify what you mean here? Is there a thread-safety issue?

rmja:
The _cache variable is tested twice in the Connect method, but if it is not marked as volatile, then the second test may be unpredictable according to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile if multiple threads are connecting in parallel.

anurse:
The second test is done within the lock which means we know that only one thread could be accessing the value. I'm still not entirely sure what `volatile` is actually buying us here.

rmja:
I am not entirely sure, but if both threads arrive at the Wait at the same time after testing _cache for null. Then when the first thread assigns to _cache, is it then guaranteed that the other thread will re-read the value of _cache after it gets the lock? Isn't it allowed to use the stall value from the previous read before the lock?

anurse:
`SemaphoreSlim.Wait` takes a lock prior to returning, which performs a volatile read (of internal lock state) and should ensure all outstanding writes have completed. Generally `volatile` is completely unnecessary as long as **all** reads and writes to the variable are done within locks.

rmja:
Yes, but in this case, there is a read _before_ the lock is obtained.

If you think that the volatile can be safely removed, then I can create a new PR removing the current volatile specifier on the _connection variable.

eerhardt:
@rmja - sorry this has sat idle for so long. Any chance you can resolve the conflicts that have occurred since you made this PR?

JunTaoLuo:

PR #2957, opened 2020-02-12
Fix LoggingEventSource to handle null strings

sywhang:
Fix issue initially reported in this issue: https://github.com/microsoft/perfview/issues/1092.

`EventSource.WriteEventCore` does not play nicely with null strings passed directly.

This fixes https://github.com/dotnet/extensions/issues/2958. 

cc @davidfowl @josalem

### DIFF
--- a/src/Logging/Logging.EventSource/src/LoggingEventSource.cs
+++ b/src/Logging/Logging.EventSource/src/LoggingEventSource.cs
@@ -125,6 +125,10 @@ namespace Microsoft.Extensions.Logging.EventSource
         {
             if (IsEnabled())
             {
+                LoggerName ??= "";
+                EventName ??= "";
+                FormattedMessage ??= "";
+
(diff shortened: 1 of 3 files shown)


davidfowl:
We should file a bug here

sywhang:
Filed https://github.com/dotnet/extensions/issues/2958

davidfowl:
Do we have any tests for this?

sywhang:
I can add some with this PR.

davidfowl:
Can we make this change in one place SetEventData? Just do value ?? “”

josalem:
You would need to make the swap before you fix the `char*` I think, so it might require a refactor of how `SetEventData` is used.  The `char*` needs to be fixed until `WriteEventCore` returns.

davidfowl:
I got a pretty sweet hack for this

Issue #2962, opened 2020-02-12
IHostApplicationLifetime has incorrect doc comments for ApplicationStopped/ApplicationStopping


willdean:
The documentation for IHostApplicationLifetime, which is just the doc-comments from the source code, contains the following text for **both** `ApplicationStopping` **and** `ApplicationStopped`:

> Triggered when the application host is performing a graceful shutdown. Shutdown will block until this event completes.

I think there are a few problems here:

1. The text shouldn't be the same for both properties - I guess it's correct for `ApplicationStopping` but wrong for `ApplicationStopped`?
2. Presumably one can't block shutdown at all by the time ApplicationStopped happens - it's too late.
3. Personally I find the "...will block until this event completes." text rather unclear, when there isn't an 'event' here at all - there's just a cancellation token.   Issue #2827 makes me think that I'm not the only person unsure precisely what this sentence implies in terms of behaviour.  The documentation for `CancellationToken.Register` manages to describe registering a callback on a CT without using the word 'event' at all - would it be clearer to mirror this terminology here?

I'd be happy to put in a PR, if my assumptions are correct:

* "Stopping" means a graceful shutdown has begun.  
* The shutdown will be blocked until any callbacks registered on ApplicationStopping have returned.
* "Stopped" means that graceful shutdown has completed.

anurse:
A few tweaks but yes, in general you are correct. The "Stopped" event still *blocks the app from exiting* (by not returning from `StopAsync`) but doesn't allow you to block the shutdown of the various services in the app. Stopping happens before those services are shut down, so you can block shutdown in that one.

A PR clarifying this would be great!

PR #2963, opened 2020-02-13
Clarify doc-comments in IHostApplicationLifetime

willdean:
- Correct/clarify comments on ApplicationStopping and ApplicationStopped

I was a little doubtful about "any callbacks" vs "all callbacks", which might be stronger/clearer?

Addresses #2962

### DIFF
--- a/src/Hosting/Abstractions/src/IHostApplicationLifetime.cs
+++ b/src/Hosting/Abstractions/src/IHostApplicationLifetime.cs
@@ -16,14 +16,14 @@ namespace Microsoft.Extensions.Hosting
         CancellationToken ApplicationStarted { get; }
 
         /// <summary>
-        /// Triggered when the application host is performing a graceful shutdown.
-        /// Shutdown will block until this event completes.
+        /// Triggered when the application host is beginning a graceful shutdown.
+        /// Shutdown will block until any callbacks registered on this token have returned.
         /// </summary>
         CancellationToken ApplicationStopping { get; }
 
         /// <summary>
-        /// Triggered when the application host is performing a graceful shutdown.
-        /// Shutdown will block until this event completes.
+        /// Triggered when the application host has completed a graceful shutdown.
+        /// The application will not exit until any callbacks registered on this token have returned.
         /// </summary>
         CancellationToken ApplicationStopped { get; }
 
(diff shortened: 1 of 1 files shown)


src/Hosting/Abstractions/src/IHostApplicationLifetime.cs
```diff
@@ -16,14 +16,14 @@ public interface IHostApplicationLifetime
         CancellationToken ApplicationStarted { get; }
 
         /// <summary>
-        /// Triggered when the application host is performing a graceful shutdown.
-        /// Shutdown will block until this event completes.
+        /// Triggered when the application host is beginning a graceful shutdown.
+        /// Shutdown will block until any callbacks registered on this token have returned.
```
anurse:
```suggestion
        /// Shutdown will block until all callbacks registered on this token have completed.
```

I think "all" is best here. Also "completed" reads better IMO.

src/Hosting/Abstractions/src/IHostApplicationLifetime.cs
```diff
@@ -16,14 +16,14 @@ public interface IHostApplicationLifetime
         CancellationToken ApplicationStarted { get; }
 
         /// <summary>
-        /// Triggered when the application host is performing a graceful shutdown.
-        /// Shutdown will block until this event completes.
+        /// Triggered when the application host is beginning a graceful shutdown.
```
anurse:
```suggestion
        /// Triggered when the application host is starting a graceful shutdown.
```

nit: beginning -> starting (reads better IMO)

src/Hosting/Abstractions/src/IHostApplicationLifetime.cs
```diff
@@ -16,14 +16,14 @@ public interface IHostApplicationLifetime
...
+        /// Shutdown will block until any callbacks registered on this token have returned.
         /// </summary>
         CancellationToken ApplicationStopping { get; }
 
         /// <summary>
-        /// Triggered when the application host is performing a graceful shutdown.
-        /// Shutdown will block until this event completes.
+        /// Triggered when the application host has completed a graceful shutdown.
+        /// The application will not exit until any callbacks registered on this token have returned.
```
anurse:
```suggestion
        /// The application will not exit until all callbacks registered on this token have completed.
```

nit: any -> all, returned -> completed.

src/Hosting/Abstractions/src/IHostApplicationLifetime.cs
willdean: Lol!  I deliberately chose 'beginning' rather than 'starting' because I felt there was a jar between the 'stopping' in the property name and having 'starting' in the description.   "It's starting the stopping but you can stop the stopping before it starts if you stop in the callback."
But obviously I will defer to you!

PR #2980, opened 2020-02-15
Document that StringValue ctor accepts a null string

breyed:


Tratcher:
@JunTaoLuo would merging this disrupt your migration work?

JunTaoLuo:
We won't be blocking in this repo for migration. Well just do a pass post migration to ensure we didn't miss anything.

FYI this PR is affecting a binary that's moving to runtime.

Tratcher:
Thanks

PR #2994, opened 2020-02-19
Make Mono.WebAssembly.Interop non-shipping

pranavkm:


wtgodbe:
Good for 3.1.3

wtgodbe:
> D:\a\1\s\eng\targets\ResolveReferences.targets(226,5): error : Project shouldn't have reference assembly but folder exists D:\a\1\s\src\JSInterop\Mono.WebAssembly.Interop\src/../ref/ [D:\a\1\s\src\JSInterop\Mono.WebAssembly.Interop\src\Mono.WebAssembly.Interop.csproj]

@dougbu do we need to remove the ref assembly from this project now that it's non-shipping?

dougbu:
> do we need to remove the ref assembly from this project now that it's non-shipping?

Only shared framework assemblies should have ref/ assemblies. We're enforcing that in dotnet/aspnetcore but don't have the same in place in this repo. But, as long as the ref/ assembly doesn't make it into Microsoft.Internal.Extensions.Ref, having a ref/ project here is only wasteful.

dougbu:
Ooops, I didn't read the error closely enough. Guess we have the error in this repo too. Need to remove the ref/ project.

pranavkm:
Spoke to @wtgodbe offline. In 3.2, we moved this package to the AspNetCore repo and it might be consolidated with another package after we're done with API reviews. Removing it from 3.1 seems reasonable.

wtgodbe:
LGTM. @dougbu @JunTaoLuo any concerns?

PR #3025, opened 2020-02-26
Fix HttpClientFactory deadlocks for 2.1 version

### LINKED ISSUE #3023: HttpClientFactory deadlocks in single-threaded SynchronizationContext for 2.1.*
Is Microsoft.Extensions.* version 2.1.* in LTS support? If so is it possible to fix https://github.com/dotnet/extensions/issues/999 in 2.1 package?

Kahbazi:
Fixes #3023


Kahbazi:
@rynowak Is it possible to patch this into 2.1*?

PR #3032, opened 2020-02-27
Flow singleton IServiceProvider through to DataAnnotationValidateOptions

jackhorton:
Summary of the changes
 - Passes the `IServiceProvider` used to resolve the `IValidateOptions` singleton service through to the `ValidationContext` that is received by the `ValidationAttribute`s. This lets `ValidationAttribute` implementations call `ValidationContext.GetService` in order to resolve dependencies to help them do their validation at runtime.
 - Does not add any tests because it didn't look like DataAnnotationValidateOptions had any to begin with.
 - Added the new ctor parameter as an optional parameter so that existing custom users of this class didn't break

Addresses #N/A (should I create a bug first?)


hhoangnl:
What happened to this change?
Will it still be merged to dotnet:master?
It will solve an issue that I am having now with resolving services inside a custom validation attribute.

PR #3040, opened 2020-02-29
Remove Serilog dependency in extensions

JunTaoLuo:
@maryamariyan Here's the changes that will remove the serilog dependencies in extensions. I don't know how you'd like to incorporate these changes into your work. Do you want to cherry-pick these changes on your branch? Or merge this PR then merge the commit over to your branch? Or some other way?

FYI, the first commit only removes the dependencies. The second substitutes the file logging we had with XUnit logging. I'm not sure what kind of logging you would prefer in dotnet/runtime so I kept the commits separate. Feel free to use one or both commits.

BTW @anurse I'll work on a PR in aspnetcore to migrate the removed files in this PR.


Tratcher:
`error : Flaky test functional logs not found.`

JunTaoLuo:
@maryamariyan Note that I had to remove a few more things in Logging.Testing. So the changes in dotnet/runtime will need to be updated when you get to it.

PR #3063, opened 2020-03-06
Add support for int keyed dictionaries to configuration binder.

beppemarazzi:
Up to now, only string or enum keys are supported for dictionaries in `ConfigurationBinder`.
This simple proposal adds support for integer keys:

- int
- uint
- short
- ushort


anurse:
Was an issue filed to discuss this addition? A change adding functionality like this really should be discussed and designed in an issue before a PR can be reviewed. If there is an issue with discussion, can you link to it from this PR?

beppemarazzi:
@anurse: I think a lot of people have this issue. i.e. look this #3052

thorgeirk11:
Please support this feature... Also, why is not just using System.Json.Text? That now has native support for this kind of binding.

PR #3085, opened 2020-03-14
Fixed variable name typo

jeuxjeux20:
The variable was named `factoryLamda`, which is a bit wrong because the only definition I've found for "lamda" is "_London Academy of Music and Dramatic Art_". I'm pretty sure this has nothing to do with music and art, so I just replaced it with `factoryLambda` 😄


Pilchie:
Thanks for the contribution @jeuxjeux20 !

PR #3086, opened 2020-03-16
Added double-checked locking to CreateLogger method

akovac35:
Added double-checked locking to minimize existing locking performance impact in the CreateLogger method.

 - Detail 1
 - Detail 2


akovac35:
Checked documentation about thread safety and the suggested change is not valid because Dictionary does not support reads while writing.

PR #3094, opened 2020-03-19
Revert SDK version

BrennanConroy:
Newer SDK brings in the net5.0 TFM. AspNetCore isn't ready for that yet. Trying to downgrade SDK so Extensions produces netcoreapp5.0 again.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,9 +1,9 @@
 {
   "sdk": {
-    "version": "5.0.100-preview.3.20168.3"
+    "version": "5.0.100-preview.1.20106.1"
   },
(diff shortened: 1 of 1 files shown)


BrennanConroy:
@Pilchie @wtgodbe Are we interested in taking this?

It will make the AspNetCore ingestion easier instead of having to do all the net5.0 work in the dependency update PR.

wtgodbe:
Looks like we landed on taking this change: https://github.com/dotnet/aspnetcore/pull/19981#issuecomment-601263364

PR #3097, opened 2020-03-20
Remove dependency on netstandard.library

wtgodbe:
This package isn't being produced out of dotnet/runtime anymore - we should just use the default version that is bundled with the SDK rather than listing a dependency on it.

CC @dagood @Pilchie @ericstj

Branch created by following the instructions in https://github.com/dotnet/aspnetcore-internal/blob/master/docs/engineering/servicing-and-preview-fixes.md

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -74,10 +74,6 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>c39af4fc45c05e862815c34239b33de3b0259705</Sha>
     </Dependency>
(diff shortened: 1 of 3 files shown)


wtgodbe:
Is the merge conflict expected after following the doc's instructions?

wtgodbe:
Rebased this to be strictly against master, since the 3.1 change is unneeded - @dougbu PTAL

dougbu:
To confirm: The plan is to keep the .NET Standard dependency in 'release/3.1' but remove it in 'master'?

wtgodbe:
> To confirm: The plan is to keep the .NET Standard dependency in 'release/3.1' but remove it in 'master'?

That's correct

PR #3098, opened 2020-03-20
[release/3.1] Remove dependency on netstandard.library

wtgodbe:
This package isn't being produced out of dotnet/runtime anymore - we should just use the default version that is bundled with the SDK rather than listing a dependency on it.

CC @dagood @Pilchie @ericstj

Branch created by following the instructions in https://github.com/dotnet/aspnetcore-internal/blob/master/docs/engineering/servicing-and-preview-fixes.md

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,14 +6,13 @@
 -->
 <Project>
   <PropertyGroup Label="Version settings">
-    <MajorVersion>3</MajorVersion>
-    <MinorVersion>1</MinorVersion>
-    <PatchVersion>3</PatchVersion>
+    <MajorVersion>5</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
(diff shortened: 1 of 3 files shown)


wtgodbe:
Is the merge conflict expected after following the doc's instructions?

dagood:
> This package isn't being produced out of dotnet/runtime anymore

dotnet/runtime isn't relevant to 3.1, and I'm not sure if the idea behind removing this dependency makes sense here. I don't know how likely we are to build a netstandard patch but I believe it would be built as part of the 3.1 product in dotnet/core-setup if it happened.

dougbu:
@wtgodbe reminder to resolve conflicts this PR is hitting.

eng/Versions.props
```diff
@@ -6,14 +6,13 @@
 -->
 <Project>
   <PropertyGroup Label="Version settings">
-    <MajorVersion>3</MajorVersion>
-    <MinorVersion>1</MinorVersion>
-    <PatchVersion>3</PatchVersion>
+    <MajorVersion>5</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>0</PatchVersion>
```
dougbu: Wow, really broken.

wtgodbe:
@dougbu I had questions about this in one of the other PRs: https://github.com/dotnet/aspnetcore/pull/20039#issuecomment-603471725

And about @dagood's concerns here: https://github.com/dotnet/aspnetcore/pull/20043#issuecomment-603508261

Issue #718, opened 2018-12-13
[RedisCache] Allow injection of ConnectionMultiplexer


epignosisx:
Currently the ```RedisCache``` creates the ```ConnectionMultiplexer``` and does not allow access to it. This prevents the application from reusing the ```ConnectionMultiplexer``` for other operations like PubSub.

Granted, the application could create another ```ConnectionMultiplexer```, but the recommendation is to have only one per application.

I was thinking if it would be possible to have a separate constructor for ```RedisCache``` that would receive the ```ConnectionMultiplexer```. I understand that so far **Microsoft.Extensions.Caching.Redis** did not leak any implementation details about its usage of **StackExchange.Redis**. However, I noticed that **Microsoft.AspNetCore.DataProtection.Redis** leaks **StackExchange.Redis** [types here](https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection.Redis/RedisXmlRepository.cs#L26). So, I'm unclear about how you feel about keeping **StackExchange.Redis** as an implementation detail.

kinosang:
For Data Protection we can use `.PersistKeysToStackExchangeRedis(ConnectionMultiplexer, RedisKey)`, how about adding an overloading such as `AddStackExchangeRedisCache(ConnectionMultiplexer, InstanceName)`?

hjkl950217:
I also encountered the same problem, for some redis data, I need to use IConnectionMultiplexer

bidianqing:
@muratg Any update to this?

muratg:
I'm no longer working with the ASP.NET team, so paging @anurse. @DamianEdwards and @davidfowl here.

anurse:
We don't have plans to do this for 3.0, but it is on our backlog to consider for a future version.

bidianqing:
https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Services/Basket/Basket.API/Startup.cs#L78

bragma:
Hi, it seems this feature won't be available anytime soon. Is there a workaround for this? I need to use the redis cache both via IDistributedCache and with native redis commands. Thanks!

anurse:
@bragma I don't disagree at all that it would be a good plan for us to allow injecting it, but can you provide more context on why you require that your native redis commands and the redis cache use the same **instance** of the `ConnectionMultiplexer`?

kinosang:
[...]

PR #3140, opened 2020-04-02
Allow injection of ConnectionMultiplexer

bladepan:
Summary of the changes
 - Add ConnectionFactory property in RedisCacheOptions
 - RedisCache class uses ConnectionFactory to create connection objects when it is set

Addresses #718

### DIFF
--- a/src/Caching/StackExchangeRedis/src/RedisCache.cs
+++ b/src/Caching/StackExchangeRedis/src/RedisCache.cs
@@ -30,11 +30,12 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
         private const string DataKey = "data";
         private const long NotPresent = -1;
 
-        private volatile ConnectionMultiplexer _connection;
+        private volatile IConnectionMultiplexer _connection;
         private IDatabase _cache;
         private bool _disposed;
 
-        private readonly RedisCacheOptions _options;
+        private readonly Func<Task<IConnectionMultiplexer>> _asyncConnectionFactory;
+        private readonly Func<IConnectionMultiplexer> _connectionFactory;
         private readonly string _instance;
 
         private readonly SemaphoreSlim _connectionLock = new SemaphoreSlim(initialCount: 1, maxCount: 1);
@@ -46,10 +47,37 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
                 throw new ArgumentNullException(nameof(optionsAccessor));
             }
 
-            _options = optionsAccessor.Value;
+            RedisCacheOptions options = optionsAccessor.Value;
+ 
+             // This allows partitioning a single backend cache for use with multiple apps/services.
+            _instance = options.InstanceName ?? string.Empty;
 
-            // This allows partitioning a single backend cache for use with multiple apps/services.
-            _instance = _options.InstanceName ?? string.Empty;
+            if (options.ConnectionFactory != null)
+            {
+                _asyncConnectionFactory = options.ConnectionFactory;
+                _connectionFactory = () => _asyncConnectionFactory().Result;
+            }
+            else
+            {
+                _asyncConnectionFactory = async () =>
+                {
+                    if (options.ConfigurationOptions != null)
+                    {
+                        return await ConnectionMultiplexer.ConnectAsync(options.ConfigurationOptions).ConfigureAwait(false);
+                    }
+
+                    return await ConnectionMultiplexer.ConnectAsync(options.Configuration).ConfigureAwait(false);
+                };
+                _connectionFactory = () =>
+                {
+                    if (options.ConfigurationOptions != null)
+                    {
+                        return ConnectionMultiplexer.Connect(options.ConfigurationOptions);
+                    }
+
+                    return ConnectionMultiplexer.Connect(options.Configuration);
+                };
+            }
         }
 
         public byte[] Get(string key)
--- a/src/Caching/StackExchangeRedis/src/RedisCache.cs
+++ b/src/Caching/StackExchangeRedis/src/RedisCache.cs
@@ -30,15 +30,11 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
         private const string DataKey = "data";
         private const long NotPresent = -1;
 
-        private volatile IConnectionMultiplexer _connection;
-        private IDatabase _cache;
+        private readonly IConnectionMultiplexer _connection;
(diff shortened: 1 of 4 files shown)


anurse:
Oof, sorry about the three message spam. But the main point of the bots message still applies. In this case, the code is all staying in extensions so we can leave this open!

anurse:
Ok, fixed the bot :). We don't need the warning anymore since the code that's moving has already moved. Sorry again!

bladepan:
The failed checks do not seem to be related to the changes. what should I do?

anurse:
@bladepan don't worry, we'll respin it as needed.

@JunTaoLuo these look like they *might* be related to removal of code ported to runtime maybe?

JunTaoLuo:
The CodeCheck failure indicates that the ref assemblies were not updated. Please run ./eng/scripts/GenerateReferenceAssemblies.ps1 to update them. The test errors looks weird, I'm not sure why the assemblies are not loaded in Debug builds but are loaded in the Release builds. Either way, it's not related to the source moves since this is 3.1

anurse:
Ah, I missed that this is 3.1. @bladepan please retarget this branch to `master`. We don't take feature changes in patch releases, so the earliest this could land is .NET 5.

bladepan:
@anurse , retargeted.

bladepan:
@anurse , @Tratcher , ping.

anurse:
@bladepan FYI there will be some delays in getting to this PR as we do the migrations discussed in https://github.com/aspnet/Announcements/issues/411 . There's still plenty of time to get this in for 5.0 :).

ransagy:
Is there any update here? This will be a great help in my scenario (custom lifetime/expiration management for cached session store items in ASP.NET auth stack) without modifying the infra in other wonky ways.

adamhathcock:
I think this doesn't go far enough.  It shouldn't cache the connection multiplexer to allow pooling of multiplexers

ransagy:
Isn't the entire point NOT to pool or use multiple ConnectionMultiplexer instances? As the class manages the pooling in itself.

adamhathcock:
My understanding is a ConnectionMultiplexer is a single TCP connection and can multiplex redis commands but it seems to be desirable have a pool of connections just like a database pool.

I'm using https://imperugo.gitbook.io/stackexchange-redis-extensions/configuration to pool ConnectionMulitplexers

src/Caching/StackExchangeRedis/src/RedisCache.cs
```diff
@@ -45,10 +46,37 @@ public RedisCache(IOptions<RedisCacheOptions> optionsAccessor)
...
+ 
+             // This allows partitioning a single backend cache for use with multiple apps/services.
+            _instance = options.InstanceName ?? string.Empty;
 
-            // This allows partitioning a single backend cache for use with multiple apps/services.
-            _instance = _options.InstanceName ?? string.Empty;
+            if (options.ConnectionFactory != null)
+            {
+                _asyncConnectionFactory = options.ConnectionFactory;
+                _connectionFactory = () => _asyncConnectionFactory().Result;
```
eerhardt, line 57: Is sync-over-async appropriate here? Is it possible this will cause problems?

bladepan: i don't see much problem here. this sync factory will be called by sync methods only. 
either we do this or we have a async-over-sync, which is less desirable IMO.

bladepan:
@adamhathcock , you can use a ConnectionMultiplexer talking to a redis cluster, so it is not just one tcp connection. The connection management logic should be done inside StackExchange.Redis. Besides, you can implement pooling of ConnectionMultiplexers by implementing a IConnectionMultiplexer, so this change does not prevent pooling of ConnectionMultiplexers.

adamhathcock:
@bladepan again, my understanding is that a single ConnectionMultiplexer is a single TCP connection.  It multiplexes multiple Redis commands, hence the name.

A pool could be implemented under `IConnectionMultiplexer` but that feels wrong.  Whether or not StackExchange.Redis should do pooling itself, that's not my call but it just doesn't which is why I personally use another library to handle pooling.

src/Caching/StackExchangeRedis/src/RedisCache.cs
eerhardt: > i don't see much problem here

See https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html, and especially the `Resources` section towards the bottom. It explains why blocking on an async operation leads to issues.

Previously, when calling the sync methods, it will call `ConnectionMultiplexer.Connect` (the sync method) which is the right thing to do. Now we only have a `Func<Task<IConnectionMultiplexer>> ConnectionFactory;` async method, which is not right to call from the synchronous path.

bladepan: How about adding a IConnectionMultiplexer field in RedisCacheOptions?

bladepan:
ping @eerhardt

src/Caching/StackExchangeRedis/src/RedisCache.cs
```diff
@@ -30,26 +30,38 @@ public class RedisCache : IDistributedCache, IDisposable
...
+            _instance = options.InstanceName ?? string.Empty;
 
-            // This allows partitioning a single backend cache for use with multiple apps/services.
-            _instance = _options.InstanceName ?? string.Empty;
+            _connection = options.ConnectionMultiplexer;
+            if (_connection == null)
+            {
+                if (options.ConfigurationOptions != null)
+                {
+                    _connection = ConnectionMultiplexer.Connect(options.ConfigurationOptions);
```
eerhardt, line 56: It looks like @Tratcher first did it this way in https://github.com/aspnet/Caching/commit/38dd3caf812728ccc0f8e99c6ad2aec064892255#diff-9ed70d9e570b33d5dc0bc574eb1d3c175d6d11397886ca242f646c6be97712dcR39, but then moved it out of the constructor during the rest of the initial implementation. See https://github.com/aspnet/Caching/pull/23/.

I assume there are a few problems with establishing the connection in the construtor:

1. What happens when it fails?
2. It could take a long time, when constructors are typically fast.
3. Not possible to be `async` any more.

eerhardt:
Have we settled on an approach in this area? #3473 is trying to address the same issue, but in a different way. It would be great if we could come to a conclusion on the best way forward here.

JunTaoLuo:

PR #3142, opened 2020-04-02
Add new Polly DI extension method for HttpClientFactory

martincostello:
This pull request adds the following new overload to the Polly extensions for HttpClientFactory:

```csharp
public static IServiceCollection AddPolicyRegistry(
    Action<IServiceProvider, IPolicyRegistry<string>> configureRegistry)
```

The existing two registrations allow a policy registry to be registered and then configured directly, but there isn't an overload that takes a delegate and provides the `IServiceProvider` to allow the policies to be configured based on other registered services, such as settings loaded from `IConfiguration`.

I've added a unit test that demonstrates the use case of using another service to configure the policies at the point the policy registry dependency is first resolved.

I've also tidied up a few minor trivial things I found while making the change, which I can split into a separate PR if necessary.

### DIFF
--- a/src/HttpClientFactory/Polly/src/DependencyInjection/PollyServiceCollectionExtensions.cs
+++ b/src/HttpClientFactory/Polly/src/DependencyInjection/PollyServiceCollectionExtensions.cs
@@ -61,5 +61,42 @@ namespace Microsoft.Extensions.DependencyInjection
 
             return registry;
         }
+
+        /// <summary>
+        /// Registers an empty <see cref="PolicyRegistry"/> in the service collection with service types
+        /// <see cref="IPolicyRegistry{String}"/>, and <see cref="IReadOnlyPolicyRegistry{String}"/> and
+        /// uses the specified delegate to configure it.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="configureRegistry">A delegate that is used to configure an <see cref="IPolicyRegistry{String}"/>.</param>
+        /// <returns>The provided <see cref="IServiceCollection"/>.</returns>
+        public static IServiceCollection AddPolicyRegistry(this IServiceCollection services, Action<IServiceProvider, IPolicyRegistry<string>> configureRegistry)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (configureRegistry == null)
+            {
+                throw new ArgumentNullException(nameof(configureRegistry));
+            }
+
+            // Create an empty registry, configure it and register it as an instance.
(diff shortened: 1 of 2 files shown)


martincostello:
@karelz Would you be able to take a look at this PR please? It seems to have fallen into a limbo state with the repo reorganisation.

karelz:
Sorry for the delay @martincostello . Sadly, we do not have anyone currently with knowledge of this code base. We should be able to look at it (incl. general code transfer) around mid July once we are sure we can deliver all critical features and bug fixes for 5.0.
Will that work for you?

martincostello:
No problem - thanks!

martincostello:
Is this too late for 5.0 now?

karelz:
We are still finishing 5.0 this week and a bit next week. Then we will have to ramp up on the code base from scratch, so I don't expect we will be able to make it happen with high confidence in 5.0 :(

karelz:
Sorry, we finally got our grip on the code base (thanks @CarnaViire!) and got to this PR.
It seems to be specifically about Polly, which should be still owned by ASP.NET team -- @Pilchie who should take a look at Polly part of HttpClientFactory?

@martincostello sorry for the delay, you have been caught unfortunately in the middle of ownership transition where things are a bit messy :(

halter73:
Original PR that added support for policy registry: https://github.com/aspnet/HttpClientFactory/pull/66

@martincostello Did you copy this pattern from anywhere? To me this seems like it's trying to work around the fact that the IPolicyRegistry isn't configured using the ASP.NET Core's [Options pattern](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1#use-di-services-to-configure-options). If it was, you'd be able to configure policies based on other registered services [using Configure](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1#use-di-services-to-configure-options).

I'm thinking maybe we should update AddPolicyRegistry to use the Options pattern instead.

martincostello:
@halter73 Yeah, I based it on the overloads that the various methods for configuring `IHttpClientFactory` have.

halter73:
@davidfowl @Tratcher What are your thoughts about using the Options pattern here?

Tratcher:
`Action<IServiceProvider, IPolicyRegistry<string>>` does resemble an Options pattern. @halter73 do you have a rough idea of what an Options version of this would look like?

halter73:
I haven't fully thought it through, but I figure the `IPolicyRegistry​<​string​>` could be a property in an option type.

JunTaoLuo:

martincostello:
Opened https://github.com/dotnet/aspnetcore/pull/28283.

PR #3160, opened 2020-04-07
Johluo/remove migrated files

JunTaoLuo:
Cleaning up and addressing more parts of https://github.com/dotnet/aspnetcore/issues/19254

I've looked at the samples and given my thoughts on what is valuable at https://github.com/dotnet/AspNetCore.Docs/issues/17646.


.github/CODEOWNERS
```diff
@@ -10,5 +10,3 @@
 /src/Caching/                           @tratcher @anurse
```
JunTaoLuo, line 10: we should probably hand these off too @dotnet/extensions-migration .

Directory.Build.props
```diff
@@ -14,9 +14,6 @@
     <IsTestAssetProject Condition="$(RepoRelativeProjectDir.Contains('testassets'))">true</IsTestAssetProject>
     <IsSampleProject Condition="$(RepoRelativeProjectDir.Contains('samples'))">true</IsSampleProject>
 
-    <IsReferenceAssemblyProject>false</IsReferenceAssemblyProject>
```
JunTaoLuo, line 17: @dotnet/aspnet-build I wanted to double check removing reference assembly related logic here. Since everything left is package only, I think we can remove all ref assemblies logic right?

src/DependencyInjection/DI.External.Tests/test/Autofac.cs
```diff
@@ -1,20 +0,0 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
JunTaoLuo, line 1: @maryamariyan I've removed all DI spec and external tests since they will be moved to runtime. If you need the source for reference you can find it in the commit history.

Directory.Build.props
wtgodbe: Makes sense to me

Directory.Build.props
```diff
@@ -73,5 +70,4 @@
 
   <Import Project="eng\Dependencies.props" />
   <Import Project="eng\ProjectReferences.props" />
-  <Import Project="eng\targets\Csharp.Refs.props" Condition="'$(IsReferenceAssemblyProject)' == 'true'" />
```
dougbu, line 76: Hey @JunTaoLuo why is eng/targets/Csharp.Refs.props still in this repo?

JunTaoLuo: No reason, i just forgot to remove it.

.github/CODEOWNERS
3GDXC: @JunTaoLuo are you able to provide details as to where the source for logging & logging abstractions now live? as the master branch no longer contains the source as far as I can tell?

Tratcher: https://github.com/dotnet/runtime/blob/6ef538d46fc971f351a22cbe854717f4ada59d89/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/ILogger.cs

anurse: See https://github.com/aspnet/Announcements/issues/411 for the full list of packages and where the code lives after this migration is complete.

3GDXC: @Tratcher @anurse thanks for pointing me to the announcement and new location for the source; it may be good to also update the README.md accordingly with a more information and/or to redirect others to the new location for source; as people may still think the namespace/package relates to this repo and not the runtime repo if they have missed the announcement.

JunTaoLuo: @bradygaster how far along are you with updating the readme?

poizan42: https://github.com/aspnet/Logging still points to https://github.com/dotnet/extensions which is quite confusing now

Tratcher: @poizan42 did you see the updated ReadMe in extensions?

poizan42: Yes, but the description at https://github.com/aspnet/Logging should probably still be updated to point to https://github.com/dotnet/runtime/tree/master/src/libraries/Microsoft.Extensions.Logging or something

PR #3165, opened 2020-04-09
Remove DiagnosticAdapter

JunTaoLuo:
According to the migrations documentation, we are actually removing DiagnosticAdapter.


rynowak:
Amazing. Best DiagnosticSource Adapter change in years 🥂

wtgodbe:
@ericstj is there a replacement for DiagnosticsAdapter? We're still using it in aspnetcore, should we just switch to the 5.0.0 version?

https://github.com/dotnet/aspnetcore/issues/35463

ericstj:
You can you use the last stable version from dotnet/extensions.  IIRC it was the consensus of @rynowak @anurse @Pilchie that this assembly wasn't providing much value and should be deprecated.  Can you just remove usage?

PR #3179, opened 2020-04-14
Package references updates

JunTaoLuo:
Addresses item in https://github.com/dotnet/aspnetcore/issues/19254

As we discussed previously, we wanted to understand the changes in out package graph and make appropriate announcements. I made a quick analyzer that found the following.

I've noticed that dotnet/runtime does a few things differently:
1. If no code from a package dependency is actually used, it is not included as a dependency in the package produced. This can lead to breaking changes.
2. If code from a transitive dependency is used, it will be included as a direct dependency. Though this results in a different set of dependencies listed in the package, I believe it does not cause a breaking change, especially given that we service all of those dependencies.

I've opened this as a PR so we can review each change if necessary. However, I don't plan to check this in.

### DIFF
--- a/docs/ReferenceChanges.md
+++ b/docs/ReferenceChanges.md
@@ -1,0 +1,50 @@
+Package: Microsoft.Extensions.Configuration.Binder
+  TFM: .NETStandard2.0
+    Removed:
+      Microsoft.Extensions.Configuration
+Package: Microsoft.Extensions.Configuration.Json
+  TFM: .NETStandard2.0
+    Removed:
+      System.Threading.Tasks.Extensions
+Package: Microsoft.Extensions.DependencyInjection
+  TFM: .NETFramework4.6.1
+    Added:
+      System.Threading.Tasks.Extensions
+  TFM: .NETStandard2.0
+    Added:
+      System.Threading.Tasks.Extensions
+Package: Microsoft.Extensions.Hosting.Abstractions
+  TFM: .NETCoreApp5.0
+    Removed:
+      Microsoft.Extensions.Logging.Abstractions
+  TFM: .NETStandard2.0
+    Removed:
+      Microsoft.Extensions.Logging.Abstractions
+    Added:
+      System.Threading.Tasks.Extensions
+Package: Microsoft.Extensions.Logging
+  TFM: .NETCoreApp5.0
+    Removed:
+      Microsoft.Extensions.Configuration.Binder
+  TFM: .NETStandard2.0
+    Removed:
+      Microsoft.Extensions.Configuration.Binder
+Package: Microsoft.Extensions.Logging.Console
+  TFM: .NETCoreApp5.0
+    Removed:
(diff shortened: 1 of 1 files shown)


docs/ReferenceChanges.md
```diff
@@ -0,0 +1,50 @@
+Package: Microsoft.Extensions.Configuration.Binder
+  TFM: .NETStandard2.0
+    Removed:
+      Microsoft.Extensions.Configuration
```
JunTaoLuo, line 4: In this case, it turns out only code from M.E.Configuration.Abstraction is used. So the package removed the M.E.Configuration dependency and added a M.E.Configuration.Abstractions dependency.

docs/ReferenceChanges.md
```diff
@@ -0,0 +1,50 @@
...
+  TFM: .NETFramework4.6.1
+    Added:
+      System.Threading.Tasks.Extensions
+  TFM: .NETStandard2.0
+    Added:
+      System.Threading.Tasks.Extensions
+Package: Microsoft.Extensions.Hosting.Abstractions
+  TFM: .NETCoreApp5.0
+    Removed:
+      Microsoft.Extensions.Logging.Abstractions
```
JunTaoLuo, line 19: This is a change we noticed during our ingestion.

docs/ReferenceChanges.md
```diff
@@ -0,0 +1,50 @@
...
+    Removed:
+      Microsoft.Extensions.Configuration
+Package: Microsoft.Extensions.Configuration.Json
+  TFM: .NETStandard2.0
+    Removed:
+      System.Threading.Tasks.Extensions
+Package: Microsoft.Extensions.DependencyInjection
+  TFM: .NETFramework4.6.1
+    Added:
+      System.Threading.Tasks.Extensions
```
JunTaoLuo, line 12: I just checked, these were in the previous closure as well since Microsoft.Bcl.AsyncInterfaces included a reference to it so this is a false positive. I didn't include that package in the set of packages I was analyzing. The same goes for M.E.Hosting.Abstraction.

docs/ReferenceChanges.md
```diff
@@ -0,0 +1,41 @@
...
+  TFM: .NETCoreApp5.0
+    Removed:
+      Microsoft.Extensions.Configuration.Abstractions
+  TFM: .NETStandard2.0
+    Removed:
+      Microsoft.Extensions.Configuration.Abstractions
+Package: Microsoft.Extensions.Logging.EventLog
+  TFM: .NETFramework4.6.1
+    Removed:
+      System.Diagnostics.EventLog
```
anurse, line 33: Is this because the type is just part of .NET Framework 4.6.1? (@ericstj ?)

ericstj: Yes

JunTaoLuo:
Announcement made in https://github.com/aspnet/Announcements/issues/415

PR #3195, opened 2020-04-21
[release/3.1] Add npm package to BAR artifacts

anurse:
This (attempts) to add the JS Interop NPM package to the shipping artifacts in BAR. The ones generated by dotnet/aspnetcore are and it makes our release process much more complicated that I have to go and find the right artifacts from AzDO instead of the files just being present in the drop share.

I say 'attempts' above because I've been hitting my head against a wall trying to "test" this. I *think* it does the right thing but can't figure out the MSBuild incantations necessary to actually generate an asset manifest to **see** if it works. Perhaps @wtgodbe or someone else from @dotnet/aspnet-build can help?

We'll know it's working if the `tgz` files appear in [barviz for dotnet-extensions](https://maestro-prod.westus2.cloudapp.azure.com/129/https:%2F%2Fgithub.com%2Fdotnet%2Fextensions/latest/graph), like the `jar`, `pom` and `tgz` files [already do for dotnet-aspnetcore](https://maestro-prod.westus2.cloudapp.azure.com/129/https:%2F%2Fgithub.com%2Fdotnet%2Faspnetcore/latest/graph)

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -1,0 +1,17 @@
+<Project>
+    <PropertyGroup>
+        <!-- The one use of ArtifactsDir in Publish.proj adds an additional slash, confusing itself. -->
+        <ArtifactsDir Condition=" HasTrailingSlash('$(ArtifactsDir)') ">$(ArtifactsDir.Substring(0, $([MSBuild]::Subtract($(ArtifactsDir.Length), 1))))</ArtifactsDir>
+
+        <PublishDependsOnTargets>$(PublishDependsOnTargets);_AddNpmPackagesToManifest</PublishDependsOnTargets>
+    </PropertyGroup>
+    <Target Name="_AddNpmPackagesToManifest">
+        <ItemGroup>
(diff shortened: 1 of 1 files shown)


anurse:
Marking tell-mode as this is infrastructure only. This is **not** needed for 3.1.4

anurse:
@wtgodbe I think my best bet for "testing" this is to yolo merge after your branding PR and watch the internal builds carefully. I'm confident it hasn't broken anything but I think the publishing I'm trying to get working (adding assets to BAR) doesn't really run until it's on the primary branch, right?

wtgodbe:
@anurse you could also push this to a branch in the AzDO repo, then add that branch to the "general testing" channel w/ DARC, then run an internal build of that

anurse:
Oh ho, that's a good point. I'll try that.

anurse:
Welp, that didn't work and nor did it produce any useful binlogs :(. The binlog only seems to show running the Test target, no publishing at all (my build: https://dev.azure.com/dnceng/internal/_build/results?buildId=643157&view=results)

anurse:
I did use the wrong artifact path though, so let's try that :).

PR #3205, opened 2020-04-25
[release/3.1] fix : failureStatus

witskeeper:
use HealthCheckRegistration.FailureStatus when CheckHealthAsync failure;

the "failureStatus" param in HealthChecksBuilderAddCheckExtensions.Add never used;

### DIFF
--- a/src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
+++ b/src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs
@@ -121,7 +121,7 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks
                 {
                     var duration = stopwatch.GetElapsedTime();
                     entry = new HealthReportEntry(
-                        status: HealthStatus.Unhealthy,
+                        status: registration.FailureStatus,
                         description: "A timeout occured while running check.",
                         duration: duration,
                         exception: ex,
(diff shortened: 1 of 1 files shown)


rynowak:
@witskeeper - is there a bug tracking this change?

witskeeper:
> @witskeeper - is there a bug tracking this change?

nope

witskeeper:
> @witskeeper - is there a bug tracking this change?

do I need to create one？

ericstj:
@JunTaoLuo looks like this is a servicing PR.

JunTaoLuo:
Yes, unfortunately, I'm not familiar enough with the component to deem whether this meets the bar for servicing. @Pilchie who would be the best person to make this call?

Pilchie:
@witskeeper - can you describe the end effect of this?  How did you find it?  Is it an issue for you in production?

By default we don't fix things in servicing branches unless there is an issue that customers can't really work around that is blocking them in production.

witskeeper:
> @witskeeper - can you describe the end effect of this? How did you find it? Is it an issue for you in production?
> 
> By default we don't fix things in servicing branches unless there is an issue that customers can't really work around that is blocking them in production.

SEE:
https://github.com/dotnet/aspnetcore/issues/22503

JunTaoLuo:
@Pilchie This dropped off my radar but should we include this in our upcoming servicing build? It looks like there are no workarounds and the fix is trivial.

Pilchie:
I still haven't seen what scenario is blocked by this.  Without that, and an explanation of what workarounds have been investigated and why they are insufficient, it doesn't meet a servicing level bar.

witskeeper:
This can be close   case   my  pr  merged  in   .NET 5.0  RC . 
see
dotnet/aspnetcore#22503
dotnet/aspnetcore#21175

Pilchie:
Thanks - closing.

PR #3208, opened 2020-04-27
Feature/consolelifetime bypass cancel

weitzhandler:
When trying to use Generic Host in an Android app, calling `Host.CreateDefaultHost()` to initiate it.

When calling `host.Start()`, calls `ConsoleLifetime.WaitForStartAsync(token)` under the hood, which calls `Console.CancelKeyPress += OnCancelKeyPress;`, that throws a `PlatformNotSupportedException`.

I tried copying the `ConsoleLifetime` file to my project, and removed that line. Then at service config I added it to the service registration:

```c#
servicesBuilder.AddSingleton<IHostLifetime, AppConsoleLifetime>();
```

It worked like a charm.
This PR is a proof of concept. Would be happy to work on it, add tests or whatever required.

Usage:

```c#
this.Host =
  Host
  .CreateDefaultBuilder()
  .UseConsoleLifetime(consoleLifetimeOptions => 
    consoleLifetimeOptions.IgnoreCancelKeyPress = true)
  .ConfigureLogging(...)
  .ConfigureServices(...)
  .Build();

Host.Start();
```

Note: My branch is based on `release/3.1`.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,6 +4,7 @@
     <clear />
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
     <add key="darc-pub-dotnet-core-setup-57d5bbb" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-core-setup-57d5bbb5/nuget/v3/index.json" />
+    <add key="darc-pub-dotnet-corefx-8a3ffed" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-corefx-8a3ffed5/nuget/v3/index.json" />
     <add key="darc-pub-dotnet-corefx-282d5b9" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-corefx-282d5b9f/nuget/v3/index.json" />
     <!--End: Package sources managed by Dependency Flow automation. Do not edit the sources above.-->
     <add key="dotnet3" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet3/nuget/v3/index.json" />
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -89,17 +89,17 @@
     </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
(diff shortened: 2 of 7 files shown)


weitzhandler:
My branch is based on `release/3.1`.

ericstj:
I think this should go to dotnet/runtime.  I'm not sure why that's not showing up as a merge conflict as the files you've changed don't exist any more in dotnet/extensions.  /cc @maryamariyan @JunTaoLuo

JunTaoLuo:
Yes this PR needs to be migrated to dotnet/runtime

weitzhandler:
So will it never be part of 3.* ?
Anyway, this is a draft feedback to lemme hear something feedback about this design.

weitzhandler:
I'm happy to migrate it and/or recuse it, it open a new design discussion or whatever needed to get it going forward.

ericstj:
3.* has already been released.  Any changes to 3.* would need to go through servicing and should start with changes in master (which now is in the dotnet/runtime repo per https://github.com/dotnet/extensions#-important-updates-).  I would recommend that you start with an issue in dotnet/runtime to discuss the feature.  IMHO we shouldn't expose API to deal with a PlatformNotSupported exception.  We should just handle it, or even fix the backing Console library to fix it.

weitzhandler:
@ericstj 

> IMHO we shouldn't expose API to deal with a PlatformNotSupported exception. We should just handle it, or even fix the backing Console library to fix it.

Thanks for your response, I agree.
Looks like there is already a similar issue in the runtime repo: https://github.com/dotnet/runtime/issues/35993. Do you suggest opening a new one? I'd like to contribute to this and get it fixed.

weitzhandler:
@ericstj 

> IMHO we shouldn't expose API to deal with a PlatformNotSupported exception. We should just handle it, or even fix the backing Console library to fix it.

Thanks for your response, I agree.

@JunTaoLuo @ericstj 

Looks like there is already a similar issue in the runtime repo: https://github.com/dotnet/runtime/issues/35993. Do you suggest opening a new one? I'd like to contribute to this and get it fixed.

ericstj:
No need to open a new issue if you think that one covers it.  PRs should be opened in dotnet/runtime that contribute towards the issue.  If API needs to be added it should be discussed in issue then go through API approval process.

PR #3211, opened 2020-04-27
Add missing source

ryanbrandenburg:
Hopefully this will fix https://github.com/dotnet/extensions/pull/3209.


dougbu:
I recommend adding this in #3209. Separate PRs to fix dependency updates just doubles the time before things flow.

PR #3239, opened 2020-05-11
Fix signing validation

dougbu:
- work around Arcade eng/Signing.props requirement


wtgodbe:
We shouldn't need this now that https://github.com/dotnet/extensions/pull/3240 is merged. But, looks like internal builds of this repo are now failing during publish with a new error... Will start a thread on First Responders

dougbu:
Have newer Arcade and that avoids this issue.

PR #3244, opened 2020-05-11
[preview5] Disable publishing during Linux builds

dougbu:
- cherry-pick from #3242 in master


dougbu:
/fyi need to rebuild because I targeted the wrong base branch

Issue #2836, opened 2020-01-08
With UseWindowsService, when main thread exits, service is left running and can't be stopped


davidmatson:
### Describe the bug
If a host uses UseWindowsService and its If Program.Main returns without throwing (for example, catches and logs unhandled top-level exceptions), the Windows service is left in the running status and can't be stopped - only killing the processing works.

### To Reproduce
1. Use the code below.
2. Start the service.
3. Try to stop the service.

### Expected behavior
The service is either not started or stops successfully.

### Actual behavior
The service is stuck running and cannot be stopped.

### Additional context
Program.cs:
```c#
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using System;
using System.Threading;
using System.Threading.Tasks;

class Program
{
    public static void Main()
    {
        try
        {
            new HostBuilder()
                .ConfigureServices(s =>
                {
                    s.AddHostedService<TestService>();
                })
                .UseWindowsService()
                .Build()
                .Run();
        }
        catch
        {
            // Log or something similar; anything other than rethrowing/throwing...
        }
    }

    class TestService : IHostedService
    {
        public Task StartAsync(CancellationToken cancellationToken)
        {
            throw new NotImplementedException();
        }

        public Task StopAsync(CancellationToken cancellationToken)
        {
            return Task.CompletedTask;
        }
    }
}
```

davidmatson:
As a more real-world example, here's a Program.Main I'm using in practice that triggers this behavior:
```c#
public static int Main(string[] args)
{
    try
    {
        Run();
        return 0;
    }
    catch (Exception exception)
    {
        if (WindowsServiceHelpers.IsWindowsService())
        {
            throw;
        }
        else
        {
            Console.Error.WriteLine(exception);
            return exception.HResult;
        }
    }
}
```

(Which happens to return a failed exit code from the process without having an unhandled exception set to the .NET runtime.)

anurse:
Triage: We create a new thread in https://github.com/dotnet/extensions/blob/88ba1c877813d015feab3557c4195deab312e063/src/Hosting/WindowsServices/src/WindowsServiceLifetime.cs#L64 and don't make it a background thread. That's probably why it's not shutting down properly. Should be an easy fix: Just make the thread a background thread.

PR #3274, opened 2020-05-29
Setting thread as background

eddynaka:
- Setting thread as background to solve a problem in UseWindowsService

#2836


Tratcher:
Were you able to reproduce #2836 and verify this fixes it?

eddynaka:
Hi @Tratcher ,

with that sample code what were my results:
- without the change:
image

- with the new change:
image

Tratcher:
Thanks

PR #3358, opened 2020-07-14
Use legacy MyGet feeds

garath:
As part of [moving to Azure Artifacts from MyGet](https://github.com/dotnet/arcade/blob/master/Documentation/MigrationPlan/MyGetFeeds.md), point all MyGet feed sources to new Azure Artifacts feed.

@dougbu FYI, from email.

### DIFF
--- a/eng/Sources.props
+++ b/eng/Sources.props
@@ -5,9 +5,8 @@
     <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
     <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true' ">
       $(RestoreSources);
-      https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
-      https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
-      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
+      https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json;
+      https://dotnetmygetlegacy.blob.core.windows.net/dotnet-core/index.json;
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
       https://api.nuget.org/v3/index.json;
     </RestoreSources>


ericstj:
Log has errors like 
```
  Error Message:
   System.Reflection.ReflectionTypeLoadException : Unable to load one or more of the requested types.
  Could not load type 'Microsoft.CodeAnalysis.IMessageSerializable' from assembly 'Microsoft.CodeAnalysis, Version=2.8.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
```

Did this change inadvertently cause a different version of Roslyn to be restored?

garath:
> Did this change inadvertently cause a different version of Roslyn to be restored?

It shouldn't, but there may be some interaction I'm missing. I don't see any errors about package version mismatches. 

Is there someone on that side who can help me identify the cause? I don't see this locally.

ericstj:
You don't see it locally because you've likely got a package cache with exact package version match.

What's most likely happening is nuget is silently picking a newer version of a dependent package because it doesn't find exact match on any feed.

To repro locally, you should be able to clear your package cache (or temporarily rename the packages folder) and build.

To get more detail about the failure you can look at the project.assets.json of the failing test before and after your change.  It should show a different package resolved after your change.

garath:
Looks like build artifacts aren't uploaded for forks. But, I can now repro locally by passing the `-ci` flag. I'll keep looking...

markwilkie:
What's the latest on this?  I see build failures...  Thoughts?  Would love to get this in...

markwilkie:
@dougbu - is this something you have insight into?

ericstj:
@markwilkie, my team owns this repo now.  I've provided a suggestion for @garath to validate his change locally.  I was waiting to hear back if he was able to try that.

markwilkie:
Thanks @ericstj.  Thoughts on next steps @garath ?

garath:
> Thoughts on next steps @garath?

The problem was in Microsoft.CodeAnalysis.CSharp 2.8.0 being restored from the myget-legacy feed instead of NuGet, and this particular nupkg appears to be incorrectly constructed. I've removed this from myget-legacy and am rerunning the build.

dougbu:
@ericstj I missed your comment. Thought we still owned the servicing branches❔

ericstj:
> Thought we still owned the servicing branches❔

OOps, indeed you do.  Sorry I missed this is in servicing.

PR #3393, opened 2020-08-08
Update branding to 3.1.8

dougbu:
- build targeting pack to pick up NuGet/Home#8599 fix
    - #2750 and dotnet/aspnetcore#24200 1/2
    - suppress MSB3243 warnings because ref/ and src/ assemblies have different versions
- fake the baseline update
    - remainder must be done after the 3.1.7 release

!!! Include code merge and baseline generation before merge !!!


dougbu:
This PR is no longer blocked

PR #3408, opened 2020-08-15
[release/3.1] Add dependencies for CPD strict in efcore

dougbu:


eng/Version.Details.xml
```diff
@@ -14,6 +14,10 @@
       <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-corefx</Uri>
       <Sha>059a4a19e602494bfbed473dbbb18f2dbfbd0878</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.DotNet.PlatformAbstractions" Version="3.1.6">
+      <Uri>https://github.com/dotnet/core-setup</Uri>
+      <Sha>d7dd3dd2bad597a566915d6c8065fdffefef4be7</Sha>
+    </Dependency>
     <Dependency Name="Microsoft.Win32.Registry" Version="4.7.0" Pinned="true">
```
dougbu, line 21: While here, should we unpin the dependencies from corefx and use `CoherentParentDependency="Microsoft.NETCore.App.Runtime.win-x64"` for all of them❔

wtgodbe: For these specific packages, we want them to be pinned: https://github.com/dotnet/aspnetcore/issues/18519.

dougbu: That answers my question for `Microsoft.Win32.Registry` because we don't have the `.0` auto-downgrades that dotnet/aspnetcore does. But, what about the following❔
- **`System.IO.Pipelines`**
- **`System.Runtime.CompilerServices.Unsafe`**
- `System.Security.Cryptography.Cng`
- `Microsoft.NETCore.App.Ref`
- `NETStandard.Library.Ref`

The first two aren't pinned at `.0` versions and `System.IO.Pipelines` is out-of-date (there's a 4.7.2 version). These are the ones (after deeper checks) I'm most wondering about unpinning now we have strict coherency to help us.

`System.Security.Cryptography.Cng` should be left where it is, again because we don't have the `.0` auto-downgrades here.

I'm less sure about  `Microsoft.NETCore.App.Ref` and `NETStandard.Library.Ref` because they aren't referenced in this repo or dotnet/aspnetcore. They're probably here to help strict coherency further up the stack. Either way, it's extremely unlikely those ref/ packs will need to be serviced. They're probably fine pinned.

dougbu: Let's continue this discussion in #24937

PR #3473, opened 2020-08-29
Inject a Redis Connection Manager to allow pooling of ConnectionMultiplexers

adamhathcock:
This is similar to https://github.com/dotnet/extensions/pull/3140 but takes it a step further to not cache the connection in the RedisCache object to allow usage of a multiplexer pool as described here: https://imperugo.gitbook.io/stackexchange-redis-extensions/configuration

Addresses #718


Pilchie:
@Tratcher or @JunTaoLuo.

Tratcher:
I've asked @BrennanConroy to take a look.

src/Caching/StackExchangeRedis/src/RedisCache.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.Extensions.Caching.StackExchangeRedis
 {
-    public class RedisCache : IDistributedCache, IDisposable
+    public class RedisCache : IDistributedCache
```
eerhardt, line 14: @Tratcher @BrennanConroy - Are we concerned about API breaking changes in this library?

adamhathcock:
Updated for the CheckDisposed change from upstream.

The RedisCache object shouldn't be using publically so I didn't think removing the IDisposable interface should be breaking.  Strictly, it is breaking.  I can add it back.

src/Caching/StackExchangeRedis/src/RedisCache.cs
Tratcher: Yes, we should avoid any API breaks. In this case it's easier to leave an empty Dispose method.

adamhathcock:
Added IDisposable back to RedisCache

JunTaoLuo:

Issue #3533, opened 2020-09-16
Memory Leak in Microsoft.Extensions.Caching.Memory when handling exceptions


eerhardt:
Duplicating https://github.com/dotnet/runtime/issues/42321 in dotnet/extensions in order to allow this issue to be backported to 3.1.

### Description

We are leaking objects when calling `MemoryCache.GetOrCreate`  and the factory method is throwing an exception.

### Repro

Run the following application, take a memory snapshot, run for a while and take another snapshot

```C#
using Microsoft.Extensions.Caching.Memory;
using System;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            var cache = new MemoryCache(new MemoryCacheOptions());
            while (true)
            {
                try
                {
                    int value = cache.GetOrCreate("hello", entry =>
                    {
                        throw new Exception();
                        return 10;
                    });
                }
                catch
                {
                    Console.WriteLine("Get Failed");
                }
            }
        }
    }
}
```

![image](https://user-images.githubusercontent.com/8291187/93358615-543d8980-f807-11ea-8c82-93b3bc339a1b.png)

### Analysis

It appears the issue is that we aren't Disposing the CacheEntry instances when an exception is being thrown:

https://github.com/dotnet/runtime/blob/33dba9518b4eb7fbc487fadc9718c408f95a826c/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs#L98-L112

The reason we aren't calling Dispose is to fix an issue that we were caching `null` when an Exception was being thrown. See https://github.com/aspnet/Caching/pull/216.

Disposing the CacheEntry is important because every time you create a CacheEntry, it gets stuck in an AsyncLocal "stack":

https://github.com/dotnet/runtime/blob/33dba9518b4eb7fbc487fadc9718c408f95a826c/src/libraries/Microsoft.Extensions.Caching.Memory/src/CacheEntryHelper.cs#L28-L36

and disposing it is what "pops" it off the stack:

https://github.com/dotnet/runtime/blob/33dba9518b4eb7fbc487fadc9718c408f95a826c/src/libraries/Microsoft.Extensions.Caching.Memory/src/CacheEntryHelper.cs#L50-L63

We should always be `Disposing` the entry. That way the `ScopeLease` object is always disposed, and the stack is cleared.

[...]

PR #3535, opened 2020-09-17
[release/3.1] Memory Leak in Microsoft.Extensions.Caching.Memory when handling exceptions

wtgodbe:
Resolves https://github.com/dotnet/extensions/issues/3533

Backport of https://github.com/dotnet/runtime/pull/42355

When an exception is thrown inside MemoryCache.GetOrCreate, we are leaking CacheEntry objects. This is because they are not being Disposed properly, and the async local CacheEntryStack is growing indefinitely.

The fix is to ensure the CacheEntry objects are disposed correctly. In order to do this, I set a flag to indicate whether the CacheEntry.Value has been set. If it hasn't, Disposing the CacheEntry won't add it to the cache.

### DIFF
--- a/src/Caching/Memory/src/MemoryCache.cs
+++ b/src/Caching/Memory/src/MemoryCache.cs
@@ -115,13 +115,6 @@ namespace Microsoft.Extensions.Caching.Memory
                 return;
             }
 
-            if (!entry.ValueHasBeenSet)
-            {
-                // No-op if the CacheEntry Value was never set. We assume an exception occurred and the caller
-                // never set the Value successfully, so don't use this entry.
-                return;
-            }
-
             if (_options.SizeLimit.HasValue && !entry.Size.HasValue)
             {
                 throw new InvalidOperationException($"Cache entry must specify a value for {nameof(entry.Size)} when {nameof(_options.SizeLimit)} is set.");
--- a/src/Caching/Memory/src/MemoryCache.cs
+++ b/src/Caching/Memory/src/MemoryCache.cs
@@ -115,6 +115,13 @@ namespace Microsoft.Extensions.Caching.Memory
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


Tratcher:
Note dotnet/runtime#42355 still has feedback that needs to be applied.

wtgodbe:
> Note dotnet/runtime#42355 still has feedback that needs to be applied.

@Tratcher the compiler doesn't like the shorter `using` syntax here, might not be on a new enough version of C#

> /home/vsts/work/1/s/src/Caching/Abstractions/src/MemoryCacheExtensions.cs(38,19): error CS1003: Syntax error, '(' expected [/home/vsts/work/1/s/src/Caching/Abstractions/src/Microsoft.Extensions.Caching.Abstractions.csproj]

Will apply the rest

wtgodbe:
Feedback has been applied

src/Caching/Memory/src/MemoryCache.cs
```diff
@@ -115,6 +115,13 @@ private void SetEntry(CacheEntry entry)
                 return;
             }
 
+            if (!entry.ValueHasBeenSet)
```
Tratcher, line 118: Revert

Tratcher (approved):

JunTaoLuo:
There has been some changes in the original PR, please make sure this PR is updated with those changes.

wtgodbe:
> There has been some changes in the original PR, please make sure this PR is updated with those changes.

Believe I now have everything except the shorter syntax for `using`, which the compiler in 3.1/2.1 doesn't know about

src/Caching/Memory/src/MemoryCache.cs
wtgodbe: Whoops, missed that, thanks

stephentoub (approved):

PR #3553, opened 2020-09-22
Reset database object during the Dispose method

### LINKED ISSUE #3531: Potential issue while disposing ConnectionMultiplexer in RedisCache?
### Describe the bug
The `RedisCache` implements the `IDisposable` interface. Although this method is not exposed through the `IDistributedCache`, there is still a possibility of this breaking in the future, if exposed. The `Dispose()` method in the class only closes the connection, while retaining the object, and does not dispose/nullify the `IDatabase` object. However, if the calling method uses this Dispose() method and then tries to call any other methods like `Get<T>`, since this database object is not null, the `Connect` method won't actually make a connection, and then later on when this database object is used to make the actual call, there will be an exception since the connection is actually closed.

### To Reproduce
No reproduction as such, since the method is currently not exposed. Just a vulnerability.

### Expected behavior
The database object should be reset during the `Dispose` method, since it will get populated again once the connection is established.

### Additional context
Currently this is present in the latest release [v3.1.8](https://github.com/dotnet/extensions/releases/tag/v3.1.8)

Marusyk:
Reset database object during the Dispose method

Fixes #3531

Please review,
Thank you in advance

### DIFF
--- a/src/Caching/StackExchangeRedis/src/RedisCache.cs
+++ b/src/Caching/StackExchangeRedis/src/RedisCache.cs
@@ -181,7 +181,7 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
                     }
                     else
                     {
-                        _connection = ConnectionMultiplexer.Connect(_options.Configuration);                        
+                        _connection = ConnectionMultiplexer.Connect(_options.Configuration);
                     }
                     _cache = _connection.GetDatabase();
                 }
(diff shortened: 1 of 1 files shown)


Tratcher:
This doesn't seem like an appropriate solution. In general nulling out fields in Dispose leads to race conditions and null reference exceptions.

The assumption in #3531 that you should be able to call Dispose to disconnect and then call other methods and have it reconnect does not match the proper use of the Dispose pattern.

Proposal: Dispose should set an internal _disposed bool field and any APIs called after should check that field and throw ObjectDisposedException.

Marusyk:
What do you think about this https://github.com/dotnet/runtime/issues/42183

Tratcher:
Unnecessary.

Tratcher (approved):
Much better.

PR #3569, opened 2020-09-25
Fix some issues with background thread compilation

### LINKED ISSUE #3566: Background thread can execute user code with captured ExecutionContext when compiling service IL
### Describe the bug
Today we compile an optimized delegate on a background after 2 service resolutions. The assumption is that this background thread should be compiling only after resolving the service twice but it's possible for the second call to outrun the first call resulting in the background thread capturing the execution context of the original thread and calling back into user code. This in itself isn't a bug but can result in bugs if there's strange interactions with async locals that hold onto an `IServiceProvider` and resolve services during this background thread compilation.

### To Reproduce
Hard, extremely hard 😄.

### Expected behavior
The background thread should discard the current execution context and not call back into user code.

### Scenario
Here's where this showed up.

Thread1: Singleton (takes lock) -> Scoped (waiting for scope lock)
Thread2: Scoped (takes scope lock) -> Singleton (waiting for singleton lock)

The DI container makes it illegal for singleton instances to depend on scoped instances so how did this happen? Well in the repro scenario, the background thread inside of the container kicked off and started to resolve singletons before the main thread could. Since we use `Task.Run`, it captures all async locals and that in turn resulted in having access to a scoped service provider via an async local (IHttpContextAccessor). This combination of events results in the above dependency graph that causes a deadlock.

davidfowl:
- Resolve dependencies before counting to avoid the race
where its possible for the background thread to run before
the main thread resulting in singletons being resolved during
compilation (it's meant to be side effect free).
- We also avoid capturing the ExecutionContext on the background thread
to avoid capture of async locals.

Fixes #3566

cc @ericstj @eerhardt  @maryamariyan


src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
```diff
@@ -19,12 +19,30 @@ public DynamicServiceProviderEngine(IEnumerable<ServiceDescriptor> serviceDescri
...
+                // Resolve the result before we increment the call count, this ensures that singletons
+                // won't cause any side effects during the compilation of the resolve function.
+                var result = RuntimeResolver.Resolve(callSite, scope);
+
                 if (Interlocked.Increment(ref callCount) == 2)
                 {
-                    Task.Run(() => base.RealizeService(callSite));
+                    // Don't capture the ExecutionContext when forking to build the compiled version of the
+                    // resolve function
+                    ThreadPool.UnsafeQueueUserWorkItem(state =>
```
ericstj, line 30: @stephentoub is this the right way to suppress ExecutionContext?  Should this instead just use `ExecutionContext.SuppressFlow()`?

davidfowl: Yah, this is best 😄

stephentoub: Ah, this is why David pinged me earlier :-)

Yeah, this is good

src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
```diff
@@ -19,12 +19,30 @@ public DynamicServiceProviderEngine(IEnumerable<ServiceDescriptor> serviceDescri
...
+                        try
+                        {
+                            base.RealizeService(callSite);
+                        }
+                        catch
+                        {
+                            // Swallow the exception, we should log this via the event source in a non-patched release
+                        }
+                    },
+                    null);
```
stephentoub, line 41: Allocation here not a big deal I assume?

src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
ericstj: I see, I wasn't sure if ThreadPool would be supported in all the same cases that ExecutionContext+Task would.  I remember back in the windows phone / early UWP days we had to avoid ThreadPool.

stephentoub: It's all good now. We've even been extending it, e.g. new overloads of this method.

eerhardt: For the education of someone like me, what is the difference between

`ThreadPool.UnsafeQueueUserWorkItem`

vs.

`ExecutionContext.SuppressFlow + Task.Run`?

When would I use one over the other?

src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
davidfowl: Naw, well we could remove it now but it isn't a regression and this is once per service resolution ever.

maryamariyan (approved):

src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
stephentoub: > When would I use one over the other?

Basically, for fire-and-forget, use the former; if you want to later join with the operation, use the latter.

src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
```diff
@@ -19,12 +19,30 @@ public DynamicServiceProviderEngine(IEnumerable<ServiceDescriptor> serviceDescri
...
+                    // resolve function
+                    ThreadPool.UnsafeQueueUserWorkItem(state =>
+                    {
+                        try
+                        {
+                            base.RealizeService(callSite);
+                        }
+                        catch
+                        {
+                            // Swallow the exception, we should log this via the event source in a non-patched release
```
eerhardt, line 38: Do we need an issue for this TODO? or will we take care of it when we port this change to 5.0+?

eerhardt (approved):
Looks good. I really wish we could add a test for this, but race conditions/deadlocks are hard to test in isolation.

src/DependencyInjection/DI/src/ServiceLookup/DynamicServiceProviderEngine.cs
davidfowl: I would take care of it in the port.

danmosemsft:
Is thre an issue/PR for 5.0?

stephentoub (approved):

maryamariyan:
> Is thre an issue/PR for 5.0? 

I can push up the PR for 5.0

ericstj (approved):

maryamariyan:
> Is thre an issue/PR for 5.0?

This is the 5.0 PR, with the identical diff. I am not authorized to merge this PR, curious if we are waiting on something before merging this?

davidfowl:
I think @Pilchie needs to take it to tactics

Pilchie:
I took it to Tactics for 3.1.  I'm assuming that approval for that implies approval for 5.0 GA.  Can we get a PR ready for that?  Branches aren't open yet for 3.1.10, but it would be good to go ahead and get it merged for 5.0 so we don't forget.

Pilchie:
Nevermind, I can read.  I'll ping Steve about the 5.0 PR now.

davidfowl:
@Pilchie the 5.0 version of this was merged, when do we merge this?

Pilchie:
When branches open for 3.1.10 - typically that happens the afternoon of the previous release, which would mean later today.  @wtgodbe - are you coordinating 3.1.10? Can you make sure this gets merged as part of that?

wtgodbe:
> When branches open for 3.1.10 - typically that happens the afternoon of the previous release, which would mean later today. @wtgodbe - are you coordinating 3.1.10? Can you make sure this gets merged as part of that?

Yep! I'll go through and merge servicing-approved PRs once the branding PRs are in later today.

Issue #3632, opened 2020-10-16
 Performance spikes when using Microsoft.Extensions.Caching.StackExchangeRedis


alexandraprisecarutrimble:
Hello,
We have a cloud solution composed of multiple microservices (developed using .Net Core 3.1). We are encountering some performance spikes that occur in every service that uses the redis caching mechanism. 

We have investigated this issue and it seems like the spikes are caused by the action of reading data from the AWS ElastiCache Redis cluster. We are consuming Redis through the “**_Microsoft.Extensions.Caching.StackExchangeRedis_**” nuget package, version 3.1.8 and the faulty method is **_RedisCache.GetAsync(string key)_**. This method sometimes gives response times that take ~1s, generally, we have response times of ~5ms. These spikes appear pretty often, and they don’t seem to follow a particular pattern. 

We used redis-cli to check the cluster to see if we have latency problems on it and all the benchmarks have shown good results.
Then we were thinking it is an infrastructure problem and the DevOps department and the AWS support assured us that our infrastructure is behaving well, we don't have heavy load and there are no issues on that part.

Now, we don’t have any more ideas how to further investigate and how to solve this issue. Therefore, we are approaching you to tell us if this is a known code issue or if it is something you have heard of before? Do you think if it is something Redis specific? Does the NuGet client need some special settings we are not aware of? Do you have any suggestions on how to check if it is an infrastructure problem or not?

### Steps to reproduce the behavior
1. Using version 3.1.8 of package Microsoft.Extensions.Caching.StackExchangeRedis
2. Initialize Redis Cache:
var cacheOptions = new RedisCacheOptions
{
      Configuration = "host:port"
};
var redisCache = new RedisCache(new RedisOptions(cacheOptions));
3. Read from redis: 
var result = await redisCache.GetAsync(key);
4. Create a .net core 3.1 web service and include the logic to read from redis as part of every http request.
5. Make several web requests (delays can be added between the requests)
6. Check the response times of the requests

### Expected behavior
Get good response times at all times, with no major performance spikes. 
Note: We are seeing performance spikes even when there is no heavy load.

### Screenshots
This is an example of a request in which we see a performance spike. It took ~2.55 seconds to read from redis. Usually this wouldn't take more than 5ms.
![performanceSpikeReadFromRedisCrop](https://user-images.githubusercontent.com/72964108/96228806-b7554400-0f9e-11eb-8608-d07bea48b7a0.png)

### Additional context
[...]

PR #3643, opened 2020-10-21
Replaced eval script with native command to read from redis. Native c…

alexandraprisecarutrimble:
Replaced eval script with native command to read from redis. Native commands have better performance compared to EVAL/EVALSHA

Summary of the changes:
 - In order to read from redis cache, a script was evaluated which got data for a key along with expiration information
 - This can be done using methods provided by the StackExchange.Redis nuget package which uses native commands
-  EVAL/EVALSHA have a greater impact on performance compared to native commands and it lead to performance spikes on real use case scenarios

Sample of redis monitoring before applying the changes:
evalscriptsforreading
Sample of redis monitoring after applying the changes:
nativecommandsforreading

Addresses #3632


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/extensions?pullRequest=3643) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/extensions?pullRequest=3643) before we can accept your contribution.<br/><br/>:x: alexandraprisecarutrimble [sign now](https://cla.dotnetfoundation.org/dotnet/extensions?pullRequest=3643)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/extensions?pullRequest=3643) it.</sub>

src/Caching/StackExchangeRedis/src/RedisExtensions.cs
```diff
@@ -8,31 +8,19 @@ namespace Microsoft.Extensions.Caching.StackExchangeRedis
...
             params string[] members)
         {
-            var result = await cache.ScriptEvaluateAsync(
-                HmGetScript,
-                new RedisKey[] { key },
-                GetRedisMembers(members));
-
             // TODO: Error checking?
-            return (RedisValue[])result;
+            return await cache.HashGetAsync(key, GetRedisMembers(members)).ConfigureAwait(false);
```
BrennanConroy, line 23: Nice, a missing `ConfigureAwait(false)` :+1:

BrennanConroy (approved):

Tratcher:
Note this has only been approved for merging into the master (6.0) branch so far. We're going to need more information to justify patching this in 3.1. Just how bad of an issue is this for you?

alexandraprisecarutrimble:
Hello, @Tratcher  
At the moment, we believe the current implementation of the redis cache reading operation is the root cause of the performance spikes in our services. Generally we have very good performance, ~2-5ms response times, but there are performance spikes which can take up to a few seconds. 
We are about to release one of our products next month and we cannot afford having these performance spikes, nor can we wait until version 6.0 to have this implementation available. After this update, we didn't see any major spikes, no matter how much we stressed the services, we even saw a performance boost, the average time of reading from an AWS Elasticache Redis cluster being now <1ms.
I believe this change can be beneficial for everyone and it won't disrupt the current workflows. 
If this change won't be integrated in 3.1, we will most likely stop using this nuget and use a temporary nuget that includes this change. We would like to avoid that and use this official nuget as the single source of truth.
Please let me know what you think and whether or not this change can be patched to 3.1 or not, so we know what steps to make.
Thanks!

Tratcher:
Note even if this is approved for 3.1, those patches only happen once per month which may not meet your schedule.

Tratcher:
@BrennanConroy can you fill out the shiproom template and get this into the next 3.x milestone?

BrennanConroy:
#### Description

The redis cache implementation uses a lua script when getting keys, this is slower than using a native redis function and because we aren't doing anything custom with the lua script we should just be using the native function. Customer noticed spikes that increased latency in their app and found that it was related to the redis cache and after investigation found that switching to the native function fixed the issue.

#### Customer Impact

Issue reported by customer in https://github.com/dotnet/extensions/issues/3632. Causes spikes that increase latency.

Potential workaround would be to implement another caching implementation on top of redis caching that does in-memory lookups and falls back to redis after X time or every X calls. This could lead to invalid cache values though and is almost the same as just forking the code and writing your own caching.

#### Regression?

No

#### Risk

Low

Pilchie:
This is approved, but need to wait to merge until the branch is open for the next release.  @BrennanConroy - we'll also need to make sure we get this into 5.0.1 servicing as well.

PR #3658, opened 2020-10-28
Removed unnecessary assignment and allocation

njy:
Just a small micro optimization: in case options.AbsoluteExpirationRelativeToNow was there, there was a useless assignment to the variable absoluteExpiration from options.AbsoluteExpiration.

After having removed that, the variable itself became useless.


Tratcher (approved):

Tratcher:
Thanks

PR #3788, opened 2021-01-13
Update Branding to 5.0.2 and disable building of packages

Anipik:
- Update branding to 5.0.2 as we didnt ship any of these packages after 5.0.1
- disabling the building of packages by default. we can enable the them when we have something to ship

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -330,16 +330,17 @@ stages:
         artifactName: Source_Build_Logs
         artifactType: Container
         parallel: true
-    - task: PublishBuildArtifacts@1
-      displayName: Upload package artifacts
-      # Only capture source build artifacts in PRs for the sake of inspecting
-      # changes that impact source-build. The artifacts from this build pipeline are never actually used.
-      condition: and(succeeded(), in(variables['Build.Reason'], 'PullRequest'))
-      inputs:
-        pathtoPublish: artifacts/packages/
-        artifactName: Source_Build_Packages
-        artifactType: Container
-        parallel: true
+    - ${{ if eq(variables['_extensionsBuildProducedPackages'], true) }}:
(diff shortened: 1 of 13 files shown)


azure-pipelines.yml
```diff
@@ -330,16 +330,17 @@ stages:
...
-      displayName: Upload package artifacts
-      # Only capture source build artifacts in PRs for the sake of inspecting
-      # changes that impact source-build. The artifacts from this build pipeline are never actually used.
-      condition: and(succeeded(), in(variables['Build.Reason'], 'PullRequest'))
-      inputs:
-        pathtoPublish: artifacts/packages/
-        artifactName: Source_Build_Packages
-        artifactType: Container
-        parallel: true
+    - ${{ if eq(variables['_extensionsBuildProducedPackages'], true) }}:
```
safern, line 333: This won't work as this is evaluated at compile time and the variable will be defined at runtime, so this will evaluate to: 

`eq(Null, true)` and this step will never be executed. You need to use a runtime condition. 

```
condition: and(succeeded(), in(variables['Build.Reason'], 'PullRequest', eq(variables['_extensionsBuildProducedPackages'], true)))
```

Anipik:
```
Evaluating: and(succeeded(), in(variables['Build.Reason'], 'PullRequest'), eq(variables['_extensionsBuildProducedPackages'], True))
Expanded: and(True, in('PullRequest', 'PullRequest'), eq(Null, True))
Result: False
```

its still getting evaluated to Null

mmitche (approved):

azure-pipelines.yml
safern: Yeah, which is what we want, for this to run only when the variable is set. However in the other case, even if the variable was set, it will always evaluate to `Null`.

Issue #3816, opened 2021-02-05
Dead end the Config.NewtonsoftJson package


Anipik:
There hasnt been any functional change After ```May 10, 2019```  in this package, so i think we should dead end the package.

We are moving the rest of hosting packages to dotnet/runtime https://github.com/dotnet/runtime/pull/47686.

cc @JunTaoLuo @dougbu @wtgodbe @ericstj @eerhardt @mmitche

ericstj:
To do this we should add it to the readme that talks about the state of all the extensions packages.

When we do it, we should make sure all downstream repos in the product know to roll back to the 5.0.x version of the package rather than expect any updates to the 6.0.x prerelease package.

PR #3818, opened 2021-02-05
Disable producing  Micorosft.Extensions.Hosting packages in extensions repo

Anipik:
We moved these packages to runtime repo https://github.com/dotnet/runtime/pull/47686

We are in process of moving the tests to the runtime repo, we can delete the source code after that.
In the mean time disabling the building of packages so that we dont have duplicate packages when we decide to publish preview2


ericstj:
> we can delete the source code after that.

Presumably these are the last things that publish out of here?  If so, why wouldn't we just disable the builds instead.  Then we can archive the repo when we are done.

Anipik:
>Presumably these are the last things that publish out of here?

I started the thread to make a decision on this https://github.com/dotnet/extensions/issues/3816, we will still ship this package.

ViktorHofer (approved):

PR #3839, opened 2021-02-25
Update branding to 3.1.14

wtgodbe:
Update branding to 3.1.14 and disable baseline validation. On patch tuesday we'll update the SDK & Baseline, and re-enable baseline validation

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -8,7 +8,9 @@
   <PropertyGroup Label="Version settings">
     <MajorVersion>3</MajorVersion>
     <MinorVersion>1</MinorVersion>
-    <PatchVersion>13</PatchVersion>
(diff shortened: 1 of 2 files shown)


eng/Versions.props
```diff
@@ -8,7 +8,9 @@
   <PropertyGroup Label="Version settings">
     <MajorVersion>3</MajorVersion>
     <MinorVersion>1</MinorVersion>
-    <PatchVersion>13</PatchVersion>
+    <PatchVersion>14</PatchVersion>
+    <ValidateBasline>false</ValidateBasline>
```
dougbu, line 12: 👍 like the placement 😀

dougbu (approved):

PR #3872, opened 2021-05-10
Stop trying to estimate IL sizes

davidfowl:
Backport of https://github.com/dotnet/runtime/pull/52516 to release/3.1

- We originally did this to reduce the amount of resizes when generating IL for huge method bodies but now that we don't inline the transitive closure of scoped dependencies, we can just estimate a modest size that avoids *most* resizes (512).
- The current ILEmitCallsiteAnalyzer doesn't keep track of visited callsites so can over estimate the size. Instead of fixing it, just delete it.

## Customer Impact

Customers with certain shaped dependency graphs will experience massive memory blowouts that are currently unavoidable (exponential memory allocation). This bug initially slipped through the cracks in the repository move to dotnet/runtime.

## Testing

The original issue has a test that was used to verify the behavior. The sample will allocate gigabytes of `byte[]` instead of a 500B `byte[]` because of the size calculation.

## Risk

Very little, the biggest risk is small amounts of `byte[]` re-allocation to resize buffers to where the amount of IL fits.


eerhardt (approved):

davidfowl:
cc @wtgodbe @Pilchie approve?

Pilchie:
This is for July - going to have to sit for a bit.

davidfowl:
There should be a bot for this.

PR #3900, opened 2021-09-21
[release/3.1] Migrate to 1ES hosted pools

lpatalas:
We need to migrate pools to 1ES hosted pools before 9/30. Only the pool/image names have to be changed. The supported functionality stays the same.

Tracking issue: https://github.com/dotnet/core-eng/issues/14276


dougbu (approved):
This looks fine but **do not** merge until servicing branches next open. Most likely approach (because branding will occur first but not until early October) will be to `cherry-pick` this commit into our branding updates.

PR #3930, opened 2022-04-01
Mark Configuration.AzureKeyVault as non-shipping

dougbu:
- see aspnet/Announcements#408


wtgodbe (approved):

Pilchie:
I suspect we should do the same for the other two packages in that announcement.

dougbu:
> I suspect we should do the same for the other two packages in that announcement.

See dotnet/aspnetcore#41007

Pilchie (approved):
Yeah, I hadn't made my way all the way through notifications yet.

PR #3941, opened 2022-05-10
Update branding to 2.1.35 & merge internal commits

wtgodbe:


eng/PatchConfig.props
```diff
@@ -22,7 +22,7 @@
       Microsoft.Extensions.Configuration.AzureKeyVault;
     </PackagesInPatch>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(VersionPrefix)' == '2.1.34' ">
+  <PropertyGroup Condition=" '$(VersionPrefix)' == '2.1.35' ">
```
dougbu, line 25: Why was the 2.1.34 entry empty❔ That doesn't seem right.

dougbu (commented):
I suspect the wrong commit is getting brought in here

dougbu (approved):

PR #3981, opened 2023-05-23
Bunch of analyzer updates/cleanup

geeknoid:


src/Analyzers/Microsoft.Extensions.ExtraAnalyzers/Common/DiagDescriptors.cs
```diff
@@ -46,32 +46,10 @@ internal static class DiagDescriptors
         category: Performance,
         description: Resources.LegacyLoggingDescription,
         defaultSeverity: DiagnosticSeverity.Warning,
-        helpLinkUri: "https://eng.ms/docs/experiences-devices/r9-sdk/docs/static-analysis/analyzers/r9a000",
+        helpLinkUri: "https://TODO/r9a000",
```
RussKie, line 49: In many dotnet/* repos we use `aka.ms/{area}-warnings`, e.g.: https://aka.ms/dotnet-warnings or https://aka.ms/winforms-warnings.

RussKie (approved):

PR #3983, opened 2023-05-23
Sign NuGet packages

RussKie:


RussKie:
The test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2186419&view=results

PR #3984, opened 2023-05-23
Add @msftbot automation

RussKie:
For the most part this is a stripped down copy from https://github.com/dotnet/winforms/blob/main/.github/fabricbot.json which I authored over the years. I aligned the labels in this repo to match.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,1137 @@
+{
+  "version": "1.0",
+  "tasks": [
+    {
+      "taskType": "trigger",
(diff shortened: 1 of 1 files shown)


danmoseley (approved):
seems reasonable. do we need to do anything to make fabricbot start reading this?

later we'll probably want to look at runtime/aspnetcore ones for any other goodness.

RussKie:
> do we need to do anything to make fabricbot start reading this?

No, once the config is in, the bot starts using it instead of a config stored in its database somewhere.

PR #4019, opened 2023-05-30
API Snapshot - resilience

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4019)


joperezr:
@geeknoid @martintmk @davidfowl I've now adjusted the PR to show the state of the API after building on top of Polly v8 so that we can actually take a look at the latest version which we intend to ship. @geeknoid please let me know if I did something wrong, as this is the first time I run APIChief 🤭

PR #4022, opened 2023-05-30
Timeprovider advance multiple invocations

### LINKED ISSUE #3995: FakeTimeProvider.Advance/SetUtcNow does not behave as expected
The current implementation of `FakeTimeProvider`s `Advance(TimeSpan)` and `SetUtcNow(DateTimeOffset)` does not behave as expected.

There are a few different scenarios, so let me take them one at a time:

**Scenario 1: A timer with `dueTime = 5 second` and `period = 5 seconds`.**

Calling `Advance(TimeSpan.FromSeconds(5))` at time 0 should invoke the timer callback 1 time (works).
Calling `Advance(TimeSpan.FromSeconds(10))` at time 0 should invoke the timer callback 2 time (does not work).

Here are two tests that cover this:

```c#
[Fact]
public void Timer_callback_invoked_multiple_times_single_advance()
{
   var sut = new FakeTimeProvider();
   var callbackCount = 0;
   var dueTime = TimeSpan.FromSeconds(3);
   var period = TimeSpan.FromSeconds(5);
   using var timer = sut.CreateTimer(_ => callbackCount++, null, dueTime, period);

   sut.Advance(TimeSpan.FromSeconds(13));

   callbackCount.Should().Be(3);
}

[Fact]
public void GetUtcNow_matches_time_at_callback_time()
{
   var sut = new FakeTimeProvider();
   var startTime = sut.GetUtcNow();
   var callbackTimes = new List<DateTimeOffset>();
   var interval = TimeSpan.FromSeconds(3);
   using var timer = sut.CreateTimer(_ => callbackTimes.Add(sut.GetUtcNow()), null, interval, interval);

   sut.Advance(interval + interval + interval);

   callbackTimes.Should().ContainInOrder(
       startTime + interval,
       startTime + interval + interval,
       startTime + interval + interval + interval);
}
```

**Scenario 2: A periodic timer with a `period = 10 seconds` and a delay of 3 seconds inserted after each `WaitForNextTickAsync` completes.**

The point of this scenario is that during a timer callback, another thing can be scheduled.

Calling `Advance(TimeSpan.FromSeconds(23))` at time 0 should result in callback invocations at times 10, 13, 20, and 23.

Here is a test that covers this:

```c#
[Fact]
public async Task Callbacks_happens_in_schedule_order()
{
    var sut = new FakeTimeProvider();
    var periodicTimer = sut.CreatePeriodicTimer(TimeSpan.FromSeconds(10));
    var startTime = sut.GetUtcNow();
    var callbacks = new List<DateTimeOffset>();
    var callbacksTask = AsyncCallbacks(periodicTimer);

    sut.Advance(TimeSpan.FromSeconds(23));

    callbacks.Should().ContainInOrder(
        startTime + TimeSpan.FromSeconds(10),
        startTime + TimeSpan.FromSeconds(13),
        startTime + TimeSpan.FromSeconds(20),
        startTime + TimeSpan.FromSeconds(23));

    periodicTimer.Dispose();
    await callbacksTask;

    async Task AsyncCallbacks(PeriodicTimer periodicTimer)
    {
        while (await periodicTimer.WaitForNextTickAsync().ConfigureAwait(false))
        {
            callbacks.Add(sut.GetUtcNow());
            await sut.Delay(TimeSpan.FromSeconds(3));
            callbacks.Add(sut.GetUtcNow());
        }
    }
}
```

I could see the benefit in a `SkipAhead`/`BendSpaceAndTime` or similar method, that allows the test user to skip forward in time from `t0` to `tN` and explicitly not invoke any callbacks between `t0` and `tN`, and only invoke callbacks set to trigger at `tN`. However, while using [my own version of FakeTimeProvider](https://github.com/egil/TimeProviderExtensions) in production for the last six months I have yet to have a need. I have however needed both of the scenarios described above in my testing.

Ps. If you want to compare notes, here is my implementation, [`ManualTimeProvider`](https://github.com/egil/TimeProviderExtensions/blob/main/src/TimeProviderExtensions/ManualTimeProvider.cs).

egil:
This fixes #3995. These are the changes I have included:

1. The `FakeTimeProviderTimer.Waiter` now tracks when a `WakeupTime` was scheduled. It does this to ensure that the callbacks are invoked in the order they are scheduled in.
2. `FakeTimeProviderTimer` now triggers callbacks one at a time, and allows their callbacks to add reschedule themselves or influence other timers/create new timers.
3. `FakeTimeProviderTimer` exposes an `Epoch` property, which represents the initial start time when the time provider was instantiated. It is a useful value to have available during testing, e.g. if you have to call `SetUtcNow` multiple times during a test, and the value you are setting is changed in relation to the start time/epoch. It tends to lead to easier-to-understand test code in my experience.

There are some build warnings that I am not sure how to solve, i.e. `warning R9
A049: Newly added symbol 'Microsoft.Extensions.Time.Testing.FakeTimeProvider.Epoch' must be marked as experimental (htt
ps://TODO/r9a049)`, but otherwise I think this is ready for review.


egil:
I didn't want to change too much of the existing implementation, but an alternative approach to implementing this I've used in my [ManualTimeProvider](https://github.com/egil/TimeProviderExtensions/blob/main/src/TimeProviderExtensions/ManualTimeProvider.cs) is to only hold waiters in the `FakeTimeProvider.Waiters` collection in while their timer has a scheduled callback, and keep the waiters sorted by `WakeupTime`. That would eliminate the need for the `Waiter.ScheduledOn` property but would be a bigger change to the existing implementation.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -14,13 +15,16 @@ namespace Microsoft.Extensions.Time.Testing;
...
     internal readonly List<FakeTimeProviderTimer.Waiter> Waiters = new();
 
-    private DateTimeOffset _now = Epoch;
+    private DateTimeOffset _now;
     private TimeZoneInfo _localTimeZone;
 
+    /// <summary>
+    /// Gets the time at which the clock was initialized.
+    /// </summary>
+    public DateTimeOffset Epoch { get; } = new DateTimeOffset(2000, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
```
sebastienros, line 26: I don't think `Epoch` should be changed or made `public` or replace `startTime`. I don't think it's related to the PR, and I don't understand the reason why it would, can you explain the reasons?

sebastienros: Saw the reasons in the comments sorry. I'd suggest to handle that separately or this could slow down the review. And not use the word `epoch` instead of start time, it means different things for me. "Epoch" should be a constant by definition.

egil: I can separate it out if you prefer, but I also used it in some of the tests for the central part of this feature, which was why I included it in this PR.

As for the name, I'm not a native English speaker, and my initial name for the property was also "StartTime". 

However, similar to us not using method name "AdvanceTime()" or "ForwardTime()", where the word "Time" is redundant/implicit due to the context it is used in, a "time provider" object, I started looking for alternatives and saw the word epoch in the original code. 

From the (oxford) dictionary, I see that epoch means _"the beginning of a period in the history of someone or something"_, and in that sense it makes sense to me that the `Epoch` property represents the beginning of time for the instance of the time provider.

All that said, I will leave it up to you and the team to decide on the final name. `StartTime`, `InitialUtcNow`, `StartUtcNow` are other alternatives I have been considering.

RussKie: > And not use the word `epoch` instead of start time, it means different things for me. "Epoch" should be a constant by definition.

I second that. I think it should be "start time" or something obvious like that.

> From the (oxford) dictionary, I see that epoch means _"the beginning of a period in the history of someone or something"_, and in that sense it makes sense to me that the `Epoch` property represents the beginning of time for the instance of the time provider.

I read it as the key being "_period in the history_", to me "epoch" is the same as an "era".

egil: I'm fine with that. I'll change it back to `StartTime`.

Two questions:

1. What about the internal field `epoch` that I removed. Do you want it back as it was or?

2. Do you want me to include the `StartTime` property in this PR or not?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -14,13 +15,16 @@ namespace Microsoft.Extensions.Time.Testing;
...
-    internal static readonly DateTimeOffset Epoch = new(2000, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
-
     internal readonly List<FakeTimeProviderTimer.Waiter> Waiters = new();
 
-    private DateTimeOffset _now = Epoch;
+    private DateTimeOffset _now;
     private TimeZoneInfo _localTimeZone;
 
+    /// <summary>
+    /// Gets the time at which the clock was initialized.
```
RussKie, line 24:
```suggestion
    /// Gets the time at which the instance of the <see cref="FakeTimeProvider"/> was initialized.
```

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -55,31 +61,26 @@ public override DateTimeOffset GetUtcNow()
...
-        lock (Waiters)
+        if (value < _now)
         {
-            _now = value;
-            waiters = GetWaitersToWake();
+            Throw.ArgumentOutOfRangeException(nameof(value), $"Cannot go back in time. Current time is {GetUtcNow()}.");
         }
 
-        WakeWaiters(waiters);
+        while (_now <= value && TryGetWaiterToWake(value) is FakeTimeProviderTimer.Waiter waiter)
```
RussKie, line 69: I'm finding it mentally challenging to parse `_now <= value` - `_now` is essentially a const in this context, and `if (10 <= value)` is harder to parse than `if (value >= 10)`.

```suggestion
        while (value >= _now && TryGetWaiterToWake(value) is FakeTimeProviderTimer.Waiter waiter)
```

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -158,28 +159,43 @@ internal void RemoveWaiter(FakeTimeProviderTimer.Waiter waiter)
...
-        return l;
-    }
-
-    private void WakeWaiters(List<FakeTimeProviderTimer.Waiter> waiters)
-    {
-        foreach (var w in waiters)
-        {
-            if (_now >= w.WakeupTime)
+            // Search through Waiters and find the waiter with the minimum
+            // WakeupTime which is less than or equal to targetNow.
```
RussKie, line 174: This comment appears to reiterate what the code below does. Such are generally not that useful. Useful comments explain "why" something is done.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -158,28 +159,43 @@ internal void RemoveWaiter(FakeTimeProviderTimer.Waiter waiter)
...
+                if (candidate is null)
+                {
+                    candidate = waiter;
+                    continue;
+                }
+
+                // This finds the waiter with the minimum WakeupTime
+                // and also ensures that if multiple waiters have the same
+                // the one that is picked is also the one that was scheduled
+                // first.
```
RussKie, line 191: [nit] Developers are generally constrained in the vertical plane more than in the horizontal plane. We found in few other repos that a line ~120-150 char long facilitates a good developer experience.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProviderTimer.cs
```diff
@@ -72,7 +72,15 @@ internal sealed class Waiter : IDisposable
 
         private long _periodMs;
         private long _dueTimeMs;
-        public DateTimeOffset WakeupTime { get; set; }
+
+        /// <summary>
+        /// Gets the timestamp for when the <see cref="Waiter"/>s
+        /// <see cref="WakeupTime"/> was last set. This is used to ensure
+        /// timer callbacks are invoked in the order they were scheduled.
+        /// </summary>
```
RussKie, line 80:
```suggestion
        /// <summary>
        ///  Gets the timestamp for when the <see cref="Waiter.WakeupTime"/> was last set.
        /// </summary>
        /// <remarks>
        ///  This property ensures timer callbacks are invoked in the order they were scheduled.
        /// </remarks>
        
```

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProviderTimer.cs
```diff
@@ -114,10 +121,9 @@ public void TriggerAndSchedule(bool restart)
                 }
                 else
                 {
-                    // Schedule next event on dueTime
-
-                    WakeupTime = _fakeTimeProvider.GetUtcNow() + TimeSpan.FromMilliseconds(_dueTimeMs);
-
+                    // Schedule next event on dueTime.
+                    // Update the current time to ensure scheduling order is preserved.
```
RussKie, line 125: [nit] This comment is largely unnecessary.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: I agree, good point.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: Fair enough, I have no strong opening here. 

In my head, `_now` is supposed to be before (we are moving forward in time), so I find it easier to read when it is first, i.e. the lower value comes before the larger value.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: Renamed Epoch to StartTime.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: My interpretation of your suggestion here was that I should not break the comment into so many lines. Ive changed that. Let me now if I misunderstood your intentions.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -14,13 +15,16 @@ namespace Microsoft.Extensions.Time.Testing;
...
     internal readonly List<FakeTimeProviderTimer.Waiter> Waiters = new();
 
-    private DateTimeOffset _now = Epoch;
+    private DateTimeOffset _now;
     private TimeZoneInfo _localTimeZone;
 
+    /// <summary>
+    /// Gets the time at which the instance of the <see cref="FakeTimeProvider"/> was initialized.
+    /// </summary>
+    public DateTimeOffset StartTime { get; } = new DateTimeOffset(2000, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
```
egil, line 26: I am not sure what to do about the `error R9A049: (NETCORE_ENGINEERING_TELEMETRY=Build) Newly added symbol 'Microsoft.Extensions.Time.Testing.FakeTimeProvider.StartTime' must be marked as experimental (https://TODO/r9a049)`. 

Please advice.

RussKie: I think any new APIs must be marked as experimental, @rafal-mz and @geeknoid can provide mor clarity on this.

geeknoid: Correct, since this is changing the API signature, it needs to be marked as [Experimental] until the API has been reviewed and approved.

egil: OK. How do I do that? An `[Experimental]` on the property?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -14,13 +15,16 @@ namespace Microsoft.Extensions.Time.Testing;
...
+
```
geeknoid, line 27: I think the comment and the property name are misleading. This is not related to the time that this object was created, it is in fact a completely arbitrary value. I would rename the property back to Epoch and change the parameter name in the constructor to "epoch" to match.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: Ahh it seems a bunch of libraries have their own copy of [ExperimentalAttribute.cs](https://github.com/dotnet/extensions/blob/main/src/LegacySupport/ExperimentalAttribute/ExperimentalAttribute.cs). Do I just copy this file into this project and add the attribute?

Also, I am not sure why this is needed in this case. `FakeTimeProvider` has never been released before, so everything related to it is experimental, or?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: I'll leave it up to you, @sebastienros, and @RussKie to decide where to go with this. Here is the initial input I received for reference: https://github.com/dotnet/extensions/pull/4022#discussion_r1211974488

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: Correct. The attribute doesn't yet exist in the lower-levels of the runtime, but it exists in the src/LegacySupport folder. To get the definition, the .csproj file needs to have a <InjectExperimentalAttributeOnLegacy>true</InjectExperimentalAttributeOnLegacy>

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: I agree about the code docs being misleading. I've updated it. Let me know if you think this is better.

RussKie:
I had a brief chat with @sebastienros, and it looks like I may have misinterpreted his comments. I am removing myself from this PR and letting @sebastienros and @geeknoid take over the reviews.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -158,28 +166,39 @@ internal void RemoveWaiter(FakeTimeProviderTimer.Waiter waiter)
         }
     }
 
-    private List<FakeTimeProviderTimer.Waiter> GetWaitersToWake()
+    private FakeTimeProviderTimer.Waiter? TryGetWaiterToWake(DateTimeOffset targetNow)
```
sebastienros, line 169: Could it use the standard `Try...` pattern by being `bool TryGetWaiterToWake(DateTimeOffset targetNow, out FakeTimeProviderTimer.Waiter?)`

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
sebastienros: Ok to rename this Epoch like @geeknoid is suggesting and keep it public/get. I can see now why this could be important to be able to set it.

egil: Should the argument passed to the ctor also be named `epoch` to align the two?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: Yes. But please indulge me, why? 

I would use the common `bool TryGetX(Foo foo, out Foo bar)` in a public API surface since that is a very recognizable pattern in the .NET ecosystem. For private/internal stuff, I have changed to this pattern since we now have nullable reference types and pattern matching, I feel this version is more direct, clean, and doesn't involve `out` arguments. I am fairly certain that had these language features been around in the early days, the TryGetX pattern would not be the norm. 

I do respect that you may have a coding standard in this repo you like to follow (I admit I didn't look at other projects in the repo, just this one) and I would change it regardless if you prefer the other style.

egil:
@sebastienros / @geeknoid I've added you to my fork of the repo so you can push changes to this PR as you see fit.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -14,13 +16,22 @@ namespace Microsoft.Extensions.Time.Testing;
...
-    private DateTimeOffset _now = Epoch;
+    private DateTimeOffset _now;
     private TimeZoneInfo _localTimeZone;
 
+    /// <summary>
+    /// Gets the time which was used as the starting point for the clock in this <see cref="FakeTimeProvider"/>.
+    /// </summary>
+    /// <remarks>
+    /// This can be set by passing in a <see cref="DateTimeOffset"/> to the constructor
+    /// which takes the <c>startTime</c> argument. If the default constructor is used,
```
geeknoid, line 29:
```suggestion
    /// which takes the <c>epoch</c> argument. If the default constructor is used,
```

egil: Doh, nice catch!

geeknoid (approved):

egil:
Let me add the missing experimental attribute to the ctor and push a rewrite of the git history that bundles up the feedback into clean commits.

PR #4024, opened 2023-05-31
Remove unclear Advance() method from FakeTimeProvider

### LINKED ISSUE #4007: Remove unclear FakeTimeProvider.Advance() overload
I am not a fan of the `Advance()` overload method on `FakeTimeProvider` that does not take a `TimeSpan` and just increments time with 1ms. It is not obvious to the dev/user and is likely going to cause confusion to some.

I suggest removing it. If specific teams like the overload, they can easily add it back in via an extension method.

egil:
The `Advance()` method requires the reader to know implicitly that it is 1ms that time is being advanced with (or read docs/use an editor that shows the info), and it is an overload that can be easily added via an extension method by teams that have the need. So since the goal is a generally applicable fake, I would recommend keeping the API surface clean and not including Advance() out of the box.

Closes #4007

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4024)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
+++ b/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
@@ -81,11 +81,6 @@ public class FakeTimeProvider : TimeProvider
         WakeWaiters(waiters);
     }
 
-    /// <summary>
-    /// Advances the clock's time by one millisecond.
-    /// </summary>
-    public void Advance() => Advance(TimeSpan.FromMilliseconds(1));
-
     /// <inheritdoc />
     public override long GetTimestamp()
     {
--- a/test/Libraries/Microsoft.Extensions.Telemetry.Testing.Tests/Logging/FakeLogCollectorTests.cs
+++ b/test/Libraries/Microsoft.Extensions.Telemetry.Testing.Tests/Logging/FakeLogCollectorTests.cs
@@ -32,7 +32,7 @@ public class FakeLogCollectorTests
         var output = new Output();
 
         var timeProvider = new FakeTimeProvider();
-        timeProvider.Advance();
+        timeProvider.Advance(TimeSpan.FromMilliseconds(1));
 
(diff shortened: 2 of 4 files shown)


rafal-mz (approved):

geeknoid:
I don't agree with this change, I find the Advance method convenient. For many tests, all you need is for time to move forward and this overload just encourages consolidating around millisecond boundaries in tests, which is clean and simple.

rafal-mz:
I agree with the author of the PR. I did like the advance method, when we could set the advance amount in the constructor of FakeTimeProvider. I think 1 millisecond is too specific value to be universally useful, yet as author pointed out it forces to read the library code to understand what is happening. I don't think this overload hurts though, I would not use it myself (without setting its advance value at ctor which I liked a lot).

egil:
> For many tests, all you need is for time to move forward ...

That may be the case for tests in dotnet/extensions, but I have not had that experience when using (something like) this in business logic yet, so I don't think this statement is universally correct.

With business logic, you often have rules like "after 5 minutes, the order status changes to ...".

> ... AdvanceMilisecond would be IMO better. But then question, why only milisecond, and not AdvanceSecond, AdvanceMinute and so on.

I don't mind these convenience methods, at least they make it clear what's going on, which helps with test readability.

RussKie (approved):

danmoseley:
> We've gone through API reviews already for this type

Doh -- I did not think of API review process as it's a test only fake, but have not gotten used to the exciting world where we're shipping fakes 😺 I misled you @egil .

I'm not sure what our API review process is for this repo, but meantime, the commenting folks above effectively decide here. I have no opinion myself.

joperezr:
Thanks for the contribution @egil. I think I agree with @rafal-mz as well in that we should either remove this method, or make it more self-explanatory by renaming it appropriately. No super strong opinions there, but seems like removing it in favor of the use of the one with the parameter seems like the better option. @geeknoid Thoughts?

PR #4035, opened 2023-06-06
Make DiagConfig runnable

RussKie:
The original script had too many hardcoded dependencies on naming and folder-location conventions, which don't work in our ever-changing environment. Another deficiency of the original implementation was that dependency were scattered across multiple locations - e.g., NuGet references along with their versions were hardcoded and had to be kept in sync with eng\Version.props and eng\Packages\*.props.

The bulk of the logic is moved into DiagConfig.csproj in the form of MSBuild targets, properties and items because most of the required information, such as the SDK version, TFM, folder locations, etc., are known to the MSBuild.
The remains of the script executing the tool was also moved into the project folder, so that the tool and its script live together.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4035)


eng/Packages/BuildOnly.props
```diff
@@ -9,13 +9,15 @@
     <PackageVersion Include="Grpc.Tools" Version="2.45.0" />
     <PackageVersion Include="LibGit2Sharp" Version="0.27.0-preview-0175" />
     <PackageVersion Include="Microsoft.Build.Framework" Version="17.4.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.ExternalAccess.AspNetCore" Version="4.4.0-6.23101.13" />
```
RussKie, line 12: The inbox .NET 8 ASP.NET Core analysers (i.e., packs\Microsoft.AspNetCore.App.Ref\8.0.0-preview.5.23260.10\analyzers\dotnet\cs) depend on Microsoft.CodeAnalysis.ExternalAccess.AspNetCore, Version=4.4.0.0. However, this assembly is nowhere to be found publicly. So I had to reference the closest version available on https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-tools/ feed.
This prompted the version updates for other Microsoft.CodeAnalysis.* libraries.

I raised the issue in https://teams.microsoft.com/l/message/19:ab93d6b5dae043a2bef1fb0be802b685@thread.skype/1686021636857?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&parentMessageId=1686021636857&teamName=.NET%20Developer%20Experience&channelName=Analyzers%20and%20Source%20Generators&createdTime=1686021636857.

eng/Tools/DiagConfig/MergeAnalyze.ps1
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env pwsh
```
geeknoid, line 1: Can we please leave this script in the top-level `scripts` folder? This is where maintenance for the repo are located. Note that this another script in there that uses DiagConfig too for instance.

eng/Tools/DiagConfig/DiagConfig.csproj
```diff
@@ -14,8 +14,13 @@
     <PackageReference Include="System.CommandLine" />
     <PackageReference Include="System.CommandLine.NamingConventionBinder" />
     <PackageReference Include="Microsoft.CodeAnalysis.Common" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
+    <PackageReference Include="Microsoft.CodeAnalysis.ExternalAccess.AspNetCore" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Features" />
     <PackageReference Include="YamlDotNet" />
   </ItemGroup>
```
geeknoid, line 22: Why are these new references needed? DiagConfig has been building fine without these...

eng/Tools/DiagConfig/DiagConfig.targets
```diff
@@ -0,0 +1,89 @@
+﻿<Project>
```
geeknoid, line 1: Feels as though it might be simpler and cleaner to have a separate .csproj file for running this logic, distinct from the project file to build DiagConfig itself.

We load all the analyzers anytime we build a project, so there wouldn't need to be any special logic for restore things, msbuild would just take care of that, I think.

src/Analyzers/Microsoft.Extensions.ExtraAnalyzers/Common/CallAnalysis/Fixers/LegacyLoggingFixer.cs
```diff
@@ -367,16 +367,21 @@ private static string GetNamespace(ClassDeclarationSyntax cl)
...
-            parameters.Add(gen.ParameterDeclaration("exception", gen.TypeExpression(invocationOp.Arguments[details.ExceptionParamIndex].Value.Type)));
+            parameters.Add(gen.ParameterDeclaration("exception", gen.TypeExpression(symbol)));
         }
 
         var index = 0;
         if (details.InterpolationArgs != null)
         {
             foreach (var o in details.InterpolationArgs)
             {
+                if (o.Type is null)
```
geeknoid, line 380: What's this about?

src/Generators/Microsoft.Gen.ComplianceReports/Roslyn4.0/Microsoft.Gen.ComplianceReports.Roslyn4.0.csproj
```diff
@@ -11,10 +11,4 @@
     <MinCodeCoverage>99</MinCodeCoverage>
     <MinMutationScore>85</MinMutationScore>
   </PropertyGroup>
-
-  <ItemGroup>
```
geeknoid, line 15: Doesn't removing this effectively defeat the reason to have Roslyn3.8 and Roslyn4.0 builds of the generators (which is to support older toolchains)

src/Analyzers/Microsoft.Extensions.ExtraAnalyzers/Common/CallAnalysis/Fixers/LegacyLoggingFixer.cs
RussKie: Newer packages have nullable annotations which trigger build warnings.

eng/Tools/DiagConfig/DiagConfig.csproj
RussKie: The inbox ASP.NET Core analyzers now depend on this.

geeknoid: But Diagconfig doesn't need to reference any analyzers itself in order to build DiagConfig, right? DiagConfig explicitly loads analyzers at runtime.

joperezr:
@RussKie do we know what might be going on with the tests that are failing? Is it related to the ifdefs that you noted above?

eng/Packages/BuildOnly.props
joperezr: Should we add a comment above these lines that links to an issue that tracks this so we remove once it is no longer needed?

eng/Packages/General.props
```diff
@@ -15,10 +15,10 @@
     <PackageVersion Include="Microsoft.Azure.KeyVault.AzureServiceDeploy" Version="3.0.0" />
     <PackageVersion Include="Microsoft.Azure.Services.AppAuthentication" Version="1.6.1" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.4.0-6.23101.13" />
```
joperezr, line 18: Do note that doing this will have an impact on the supported VS versions for our source generators, so we should be very sure about making these changes. You can find matching versions support in [this table](https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/extensibility/roslyn-version-support.md), but essentially before this change our source generators would have run fine on VS 17.0, but after this they will require 17.4.

src/Generators/Microsoft.Gen.ComplianceReports/Roslyn4.0/Microsoft.Gen.ComplianceReports.Roslyn4.0.csproj
joperezr: It does, we should roll this back.

src/Generators/Microsoft.Gen.AutoClient/Roslyn4.0/Microsoft.Gen.AutoClient.Roslyn4.0.csproj
```diff
@@ -20,10 +20,4 @@
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
   </ItemGroup>
-
-  <ItemGroup>
-    <PackageVersion Update="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.1" />
```
joperezr, line 25: All of these generators are specifically trying to build against 4.0, so we shouldn't bump them to 4.4 as that makes them not be supported any longer on older toolset versions.

joperezr:
@RussKie what is the expectation of this PR? Was it mostly a proof-of-concept or do we actually intend to merge it at some point?

RussKie:
We’d want to have it merged at some point when we figure out the correct way to update the tool. This isn’t high on my priority list though.

geeknoid:
I think we can close this PR, we have a working solution now.

PR #4036, opened 2023-06-06
Stop building for .NET 6

RussKie:
Build the shippable artifacts with `net462,netcoreapp3.1,net8.0`, and run tests for `net462,netcoreapp3.1,net6.0,net7.0,net8.0`

Resolves dotnet/r9#423

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4036)


geeknoid:
Yes, I think the tests need to change as suggested.

RussKie:
I have been thinking about this, and this feels broken. Right now, `LogMethodHelper.SkipEnabledCheckOptions` is only available in .NET 8 since this is the highest TFM we build (`net462,netcoreapp3.1,net8.0`). Next year we won't be building against `net8.0`, instead we'll be building against `net9.0` (`net462,netcoreapp3.1,net9.0`), and the test will fail again... Am I missing something here?

/cc: @joperezr

joperezr:
> I have been thinking about this, and this feels broken. Right now, `LogMethodHelper.SkipEnabledCheckOptions` is only available in .NET 8 since this is the highest TFM we build (`net462,netcoreapp3.1,net8.0`). Next year we won't be building against `net8.0`, instead we'll be building against `net9.0` (`net462,netcoreapp3.1,net9.0`), and the test will fail again... Am I missing something here? Are we planning to add new TFMs to the list, e.g., `net462,netcoreapp3.1,net8.0,net9.0`?

If newer TFMs are adding API like in this case, that TFM will need to be kept in the future in order to not introduce a breaking change. That means that at least for this particular case, we would cross compile for both net8.0 and net9.0 as dropping net8.0 asset would be a breaking change for that API. Package Validation against a baseline package would catch these breaks.

joperezr:
@RussKie do we know what might be going on with the tests that are failing? Is it related to the ifdefs that you noted above?

eng/Packages/TestOnly-net6.0.props
```diff
@@ -3,6 +3,9 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.1" />
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(MicrosoftExtensionsDiagnosticsHealthChecksVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Http" Version="6.0.0" />
```
joperezr, line 7: Shouldn't all of these versions come from versions.props using dependency flow? Also, why are we pinning down the dependency to 6.0 version, we should be always using latest version of the extension packages.

test/Analyzers/Microsoft.Extensions.ExtraAnalyzers/Common/ConditionalAccessAnalyzerTests.cs
```diff
@@ -106,7 +106,9 @@ public Arg(TParameter arg1)
             null,
             new[] { Source }).ConfigureAwait(false);
 
-#if NET6_0_OR_GREATER
+#if NET8_0_OR_GREATER
+        // Whilst these API are marked as NET6_0_OR_GREATER we don't build .NET 6.0,
```
joperezr, line 110: I'm not understanding here what you mean here by "marked as NET6.0 or greater"

RussKie: The actual API are guarded with `#if NET6_0_OR_GREATER` meaning those are technically available in .NET 6.

eng/Packages/TestOnly-net6.0.props
geeknoid: I think once Eric's compatibility PR goes in, we can cleanup the files in eng/Packages to only ever depend on the .NET 8 extensions. At that point, we can also cleanup `#if` throughout the code base to remove dead branches.

test/Analyzers/Microsoft.Extensions.ExtraAnalyzers/Common/ConditionalAccessAnalyzerTests.cs
joperezr: But the actual API isn't building for net6.0, so no matter what the ifdef says, the availability of an API is decided by whether or not the API can be called in 6.0, not if the API could theoretically be available but is only compiled for .net 8.

RussKie: Until we removed `#if` defs, it can create a confusion - e.g., a developer looking at the code and seeing an API guarded with `#if NET6_0_OR_GREATER` then running a test under .NET 6 and seeing a failure - I've gone through that experience and it was confusing.
When we remove the pragmas (if ever) then we can remove the comments too.

RussKie:
This is now ready.

geeknoid (approved):

joperezr (approved):
LGTM. Not sure what is wrong with the CLA check, as you have obviously already signed it. cc @jeffwilcox

jeffwilcox:
@dotnet-policy-service rerun

joperezr:
Didn't seem to have worked. @jeffwilcox should we try closing and reopening? Or perhaps amending the last commit?

jeffwilcox:
I'm told the `rerun` command only works for the person who opened the pull request, I'm asking if that can be expanded to include those with write access to the repo.

RussKie:
Doesn't `/azp run` work for anyone with write access?

PR #4038, opened 2023-06-07
Re-introduce Microsoft.Extensions.ObjectPool.DependencyInjection

sebastienros:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4038)


sebastienros:
This is accounting for the aspnetcore api-review (method names, namespaces, api)

RussKie:
Once this is merged please make an update in https://github.com/dotnet/api-catalog-infra/ - either update https://github.com/dotnet/api-catalog-infra/pull/59 (if it's still unmerged) or apply the renaming separately.

test/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection.Tests/Microsoft.Extensions.ObjectPool.DependencyInjection.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyName>Microsoft.Extensions.ObjectPool.DependencyInjection.Tests</AssemblyName>
+    <RootNamespace>Microsoft.Extensions.ObjectPool.Test</RootNamespace>
+    <Description>Unit tests for Microsoft.Extensions.ObjectPool.DependencyInjection.</Description>
```
RussKie, line 5: You could do this:
```suggestion
    <Description>Tests for $(MSBuildProjectName.Replace('.Tests', ''))</Description>
```

RussKie (approved):

test/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection.Tests/Microsoft.Extensions.ObjectPool.DependencyInjection.Tests.csproj
geeknoid: I'd rather keep it human-readable, like we have in the other project files in this repo.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/PoolOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Extensions.ObjectPool;
+
+/// <summary>
+/// Contains configuration for pools.
+/// </summary>
+[Experimental]
+public sealed class PoolOptions
+{
+    internal static readonly int DefaultCapacity = Environment.ProcessorCount * 2;
```
geeknoid, line 15: Do we need this field? Seems to just be wasting 4 bytes of memory.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,94 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.DependencyInjection;
```
geeknoid, line 10: Should this be Microsoft.Extensions.ObjectPool?

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+
+/// <summary>
+/// Extension methods for adding <see cref="ObjectPool{T}"/> to DI container.
+/// </summary>
+[Experimental]
+public static class ObjectPoolServiceCollectionExtensions
```
geeknoid, line 16: All the code in here is missing null argument checks.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+{
```
geeknoid, line 17: Can we get back the overload that consumes a configuration section? We want to systematically enable things to be configured in this way and it's the convention in this repo.

Can we call all the configuration delegate just "configure" which is the convention we use in this repo?

Thanks.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
sebastienros: I believe all extension methods for `ServiceCollection` are in this namespace (at least in aspnetcore) such that you don't need to add a using statement to get it in intellisense. By your comment I assume it's not the case in dotnet/extensions, intentionally?

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Routing/RoutingStrategyBuilderExtensions.cs
```diff
@@ -6,7 +6,7 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.DependencyInjection.Pools;
+
```
joperezr, line 9: NIT: extra line.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
sebastienros: Currently looking into why the section was removed after aspnetcore api-review, they had some arguments that might be worth checking.

`configureOptions`: there are a few occurrences in this repos in case you think it's worth going through the rest of the code for consistency.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
rafal-mz: It isn't. We use one namespace for feature area usually.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+        if (configure != null)
+        {
+            // Register a PoolOption instance specific to the type
+            _ = services.ConfigurePool<TService>(configure);
+        }
+
+        return services
+            .AddSingleton<ObjectPool<TService>>(provider =>
+            {
+                var options = provider.GetService<IOptionsFactory<PoolOptions>>()?.Create(typeof(TService).FullName!) ?? new PoolOptions();
```
rafal-mz, line 91: How does this differ from `provider.GetRequiredService<IOptions<PoolOptions>>`?

rafal-mz (approved):

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
geeknoid: So yes, we should change "configureOptions" to "configure" for general consistency.

Within this repo, we generally use a triplet of functions for configuring components. No options (you get the defaults), an Action delegate which lets you override options, and a ConfigurationSection to make it easy to consume the configuration state from an external JSON. This model has been working well for us.

@tekian Jan helped establish this pattern 3 years ago or so, maybe he has additional insights to provide.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
sebastienros: `Required` would fail if none is registered. But more importantly it has to be a different value per name (the pool type) which `GetRequiredService` doesn't provide.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/PoolOptions.cs
```diff
@@ -1,19 +1,23 @@
...
 using System.Diagnostics.CodeAnalysis;
 
-namespace Microsoft.Extensions.DependencyInjection.Pools;
+namespace Microsoft.Extensions.ObjectPool;
 
 /// <summary>
 /// Contains configuration for pools.
 /// </summary>
 [Experimental]
 public sealed class PoolOptions
```
geeknoid, line 12: Given that this is now in the general-purpose Microsoft.Extensions.ObjectPool namespace, perhaps we should give this a more specialized name? ServiceCollectionPoolOptions maybe? Any other ideas?

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/ObjectPoolServiceCollectionExtensions.cs
davidfowl: Thats fine but 90% of our code takes IConfiguration, not IConfigurationSection and it usually parses a known schema (more than just an int). The pattern should be consistent with the rest of the APIs in the stack.

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/PoolOptions.cs
sebastienros: This one is hard, `DependencyInjectionPoolOptions`? But the current name loks fine to me, there is no conflict and users will probably not use the class name directly.

davidfowl: Why didn't this get called ObjectPoolOptions?

rafal-mz: I agree with Martin. It looks strange, that this type (or ObjectPoolOptions) is not used by `Microsoft.Extensions.ObjectPool` || `Microsoft.Extensions.ObjectPool<T>`.

sebastienros:
@RussKie flaky tests? also the Tests tab is empty on azdo, the one listing all failed tests and artifacts.

RussKie:
> @RussKie Igor Velikorossov FTE flaky tests? also the Tests tab is empty on azdo, the one listing all failed tests and artifacts.

I believe this is caused by https://github.com/dotnet/extensions/issues/4078. I have a hunch Microsoft.Extensions.Diagnostics.ResourceMonitoring.Test could be unsettling the build agent - I saw a few build runs where tests from that project failed.

RussKie:
@sebastienros please don't forget about https://github.com/dotnet/extensions/pull/4038#issuecomment-1581773426.

PR #4039, opened 2023-06-07
Reference .NET 8 Preview 5 packages

joperezr:
Changing dependency versions to use Preview 5 as opposed to preview 6. This is basically taking all of the changes done in main at this point so all changes till now will be included.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4039)


RussKie (approved):

RussKie:
Whenever we branch, we need to create new subscriptions as well. E.g.:
image

eng/Version.Details.xml
```diff
@@ -1,168 +1,168 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Bcl.TimeProvider" Version="8.0.0-preview.6.23280.5" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
```
RussKie, line 3: Why are we dropping the coherency? We don't pull the runtime packages directly, and without the coherency we won't get updates, and it'd likely create version conflicts.

RussKie:
Replaced by #4043.

PR #4054, opened 2023-06-08
Try out more surgical removal of HttpClient logging handler

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4054)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/HttpClientLoggingExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/HttpClientLoggingExtensions.cs
@@ -8,6 +8,7 @@ using System.Net.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Http.Logging;
 using Microsoft.Extensions.Http.Telemetry.Logging.Internal;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/HttpClientLoggingExtensions.cs
```diff
@@ -236,6 +240,19 @@ public static IServiceCollection AddHttpClientLogEnricher<T>(this IServiceCollec
         return services;
     }
 
+    private static void RemoveDefaultLogging(HttpMessageHandlerBuilder b)
+    {
+        // Remove the logger handlers added by the filter. Fortunately, they're both public, so it is a simple test on the type.
+        for (var i = b.AdditionalHandlers.Count - 1; i >= 0; i--)
+        {
+            var handlerType = b.AdditionalHandlers[i].GetType();
+            if (handlerType == typeof(LoggingScopeHttpMessageHandler) || handlerType == typeof(LoggingHttpMessageHandler))
```
xakep139, line 249: Can we cache both `typeof`s and store their values in variables?

geeknoid: Why? typeof costs nothing, it compiles down to a constant value load.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/HttpClientLoggingExtensions.cs
```diff
@@ -211,13 +214,14 @@ public static IHttpClientBuilder AddHttpClientLogging(this IHttpClientBuilder bu
         _ = builder.Services
             .AddHttpRouteProcessor()
             .AddHttpHeadersRedactor()
-            .AddOutgoingRequestContext()
-            .RemoveAll<IHttpMessageHandlerBuilderFilter>();
```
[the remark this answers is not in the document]
geeknoid: Yes, that's the approach the code was using. But @JamesNK was suggesting a better approach which I tried to do here. But it doesn't work, so I must've done something wrong.

PR #4056, opened 2023-06-08
Layer enrichment & redaction on top of HttpLogging

Tratcher:
This is a proof of concept that re-layers onto the new extensibility in https://github.com/dotnet/aspnetcore/pull/48679.

In theory this replaces the local logging middleware.

*Doesn't compile until the aspnetcore PR gets merged*

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4056)

### DIFF
--- a/src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
+++ b/src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
@@ -22,14 +22,17 @@ using Microsoft.Extensions.Telemetry.Logging;
 
 namespace Microsoft.AspNetCore.Telemetry.Http.Logging;
 
-internal sealed class HttpRedactionHandler : IHttpLoggingInterceptor
+internal sealed class HttpLoggingRedactionInterceptor : IHttpLoggingInterceptor
 {
+    private const string RequestStartTimestamp = "RequestStartTimestamp";
+    private const string Durration = "duration";
+
     // These three fields are "internal" solely for testing purposes:
     internal TimeProvider TimeProvider = TimeProvider.System;
 
     private readonly IncomingPathLoggingMode _requestPathLogMode;
     private readonly HttpRouteParameterRedactionMode _parameterRedactionMode;
-    private readonly ILogger<HttpRedactionHandler> _logger;
+    private readonly ILogger<HttpLoggingRedactionInterceptor> _logger;
     private readonly IHttpRouteParser _httpRouteParser;
     private readonly IHttpRouteFormatter _httpRouteFormatter;
     private readonly IIncomingHttpRouteUtility _httpRouteUtility;
@@ -39,9 +42,9 @@ internal sealed class HttpRedactionHandler : IHttpLoggingInterceptor
     private readonly IHttpLogEnricher[] _enrichers;
     private readonly FrozenDictionary<string, DataClassification> _parametersToRedactMap;
 
-    public HttpRedactionHandler(
+    public HttpLoggingRedactionInterceptor(
         IOptions<LoggingRedactionOptions> options,
-        ILogger<HttpRedactionHandler> logger,
+        ILogger<HttpLoggingRedactionInterceptor> logger,
         IEnumerable<IHttpLogEnricher> httpLogEnrichers,
         IHttpRouteParser httpRouteParser,
         IHttpRouteFormatter httpRouteFormatter,
@@ -75,17 +78,20 @@ internal sealed class HttpRedactionHandler : IHttpLoggingInterceptor
         {
             logContext.LoggingFields = HttpLoggingFields.None;
         }
+        else if (logContext.IsAnyEnabled(HttpLoggingFields.Response))
+        {
+            // We'll need this for the response.
+            // TODO: Should we put a state filed on logContext?
+            context.Items[RequestStartTimestamp] = TimeProvider.GetTimestamp();
+        }
 
         // Don't enrich if we're not going to log any part of the request
-        if ((HttpLoggingFields.Request & logContext.LoggingFields) == HttpLoggingFields.None)
+        if (!logContext.IsAnyEnabled(HttpLoggingFields.Request))
         {
             return;
         }
 
-        // TODO: Should we put a state filed on logContext?
-        context.Items["RequestStartTimestamp"] = TimeProvider.GetTimestamp();
-
-        if (logContext.LoggingFields.HasFlag(HttpLoggingFields.RequestPath))
+        if (logContext.TryOverride(HttpLoggingFields.RequestPath))
         {
             string path = TelemetryConstants.Unknown;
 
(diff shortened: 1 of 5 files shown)


geeknoid (commented):
Looks pretty clean.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+    ///     "/probe/live",
+    ///     "/probe/ready"
+    /// };
+    /// </code>
+    /// </example>
+    [Experimental]
+    [Required]
+    [SuppressMessage("Usage", "CA2227:Collection properties should be read only",
+        Justification = "Options pattern.")]
+    public ISet<string> ExcludePathStartsWith { get; set; } = new HashSet<string>();
```
noahfalk, line 101: This functionality doesn't seem like it belongs on RedactionOptions in an ideal world. Filtering (in whatever form we think is most appropriate) feels like something that would be available in the base layer without having opted into any redaction. Should I think of this as refactoring is still in progress or we believe this needs to stay here because of various time or design constraints?

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,206 @@
...
+        var context = logContext.HttpContext;
+        var request = context.Request;
+        if (ShouldExcludePath(context.Request.Path))
+        {
+            logContext.LoggingFields = HttpLoggingFields.None;
+        }
+        else if (logContext.IsAnyEnabled(HttpLoggingFields.Response))
+        {
+            // We'll need this for the response.
+            // TODO: Should we put a state filed on logContext?
```
noahfalk, line 84: Using HttpContext for state storage seems fine to me. Any particular problem you are aiming to improve upon with an alternative solution?

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,206 @@
...
+                path = request.Path.Value!;
+            }
+
+            logContext.Add(nameof(request.Path), path);
+        }
+
+        if (logContext.TryOverride(HttpLoggingFields.RequestHeaders))
+        {
+            // TODO: HttpLoggingOptions.Request/ResponseHeaders are ignored which could be confusing.
+            // Do we try to reconcile that with LoggingRedactionOptions.RequestHeadersDataClasses?
```
noahfalk, line 141: I think this behavior as-is seems pretty reasonable and we'd document something like "Invoking AddHttpLoggingRedaction() offers more sophisticated redaction customization options for path and headers. The basic header configuration settings in HttpLoggingOptions will no longer be used when this is enabled."

noahfalk (commented):
Looks nice to me 👍

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/LoggingRedactionOptions.cs
Tratcher: It's certainly a design point that needs to be resolved. For now it's here because it's an option that the extensions middleware has and the aspnetcore one doesn't. I wouldn't move it to aspnetcore as is, this is an anti-pattern that ends up duplicating routing. I do understand it was done this way because it's easy and config compatible.

We will revisit this detail once we agree on the larger design.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,206 @@
...
+
+    public void OnRequest(HttpLoggingContext logContext)
+    {
+        var context = logContext.HttpContext;
+        var request = context.Request;
+        if (ShouldExcludePath(context.Request.Path))
+        {
+            logContext.LoggingFields = HttpLoggingFields.None;
+        }
+        else if (logContext.IsAnyEnabled(HttpLoggingFields.Response))
```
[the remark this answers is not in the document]
Tratcher: Turning bits on is fine, so long as an interceptor is internally consistent. E.g. this interceptor only stores the start time if it expects a response. It may need a little hardening in case the key isn't found later for any reason. I don't know that I expect many competing interceptors.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
Tratcher: Just trying to avoid using the Items collection for perf reasons if we already have an object that's going to flow between the two callbacks. However, that state field would have to have similar loose typing so it might not be any better of an experience, and could conflict if there's multiple interceptors.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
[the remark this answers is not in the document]
noahfalk: > I don't know that I expect many competing interceptors.

I don't expect many. Mostly I think we are reliant people only using one, or taking some care when selecting/ordering multiple to avoid conflicts. If it ever does scale up further it can be a future problem.

> It may need a little hardening in case the key isn't found later for any reason

Yeah I would, or unconditionally capture the data.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,206 @@
...
+                logContext.Parameters.Add(pair);
+            }
+            LogMethodHelper.ReturnHelper(enrichmentBag);
+        }
+
+        // Catching duration at the end:
+        // Note this does not include the time spent writing the response body.
+        var startTime = (long)context.Items[RequestStartTimestamp]!;
+        var duration = (long)TimeProvider.GetElapsedTime(startTime, TimeProvider.GetTimestamp()).TotalMilliseconds;
+        logContext.Add(Durration, duration);
```
noahfalk, line 180: This is another one that ideally seems like it could be a base-layer optional field. Duration has no clear connection to redaction functionality.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            _responseHeadersReader.Read(context.Response.Headers, logContext);
+        }
+
+        if (_enrichers.Length > 0)
+        {
+            var enrichmentBag =  LogMethodHelper.GetHelper();
+            foreach (var enricher in _enrichers)
+            {
+                enricher.Enrich(enrichmentBag, context.Request, context.Response);
```
noahfalk, line 166: Perhaps this should be in a separate interceptor that is registered in [AddHttpLogEnricher()](https://github.com/dotnet/extensions/blob/f4952b69a04e9bef266089cbc3059675693fbaa0/src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/HttpLoggingServiceExtensions.cs#L85)? I don't think users would expect enrichment to require redaction in order to work.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
JamesNK, line 180: Durration -> Duration

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,187 @@
...
+using Microsoft.Extensions.Telemetry.Logging;
+
+namespace Microsoft.AspNetCore.Telemetry.Http.Logging;
+
+internal sealed class HttpLoggingRedactionInterceptor : IHttpLoggingInterceptor
+{
+    private const string RequestStartTimestamp = "RequestStartTimestamp";
+    private const string Duration = "duration";
+
+    // These three fields are "internal" solely for testing purposes:
```
rafal-mz, line 30: Only one internal field.

rafal-mz (approved):

geeknoid:
I presume a bunch of stuff would still need to be deleted when enabling this stuff right? Some extension methods, the option type, etc?

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// </remarks>
+    public IncomingPathLoggingMode RequestPathLoggingMode { get; set; } = DefaultRequestPathLoggingMode;
+
+    /// <summary>
+    /// Gets or sets a value indicating how request path parameter should be redacted.
+    /// </summary>
+    /// <remarks>
+    /// Default set to <see cref="HttpRouteParameterRedactionMode.Strict"/>.
+    /// </remarks>
+    [Experimental]
```
geeknoid, line 39: This whole type needs to be marked [Experimental]

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/LoggingRedactionOptions.cs
geeknoid: What would be a config-friendly way of doing this which is not an anti-pattern?

Tratcher:
> I presume a bunch of stuff would still need to be deleted when enabling this stuff right? Some extension methods, the option type, etc?

Yeah, I haven't purged the old code yet, I'll do that before undrafting. For now I want to keep it focused on the new design.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/LoggingRedactionOptions.cs
Tratcher: https://github.com/dotnet/aspnetcore/issues/43642 would allow you to apply metadata (like the logging attribute) to arbitrary route paths and be more config friendly. I'll ping them to see how that's coming along.

Tratcher:
Making a new PR for release/8.0

PR #4079, opened 2023-06-15
Remove dependency on dotnet7 feed

RussKie:
Follow up on https://github.com/dotnet/extensions/pull/4061, which introduced dependency on dotnet7 NuGet feed. This change removes the feed, supressing the warning instead.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4079)

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -13,6 +13,15 @@
     <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);NETSDK1138;MSB3270</MSBuildWarningsAsMessages>
     <WarningLevel>5</WarningLevel>
 
+    <!--
+      Related to https://github.com/dotnet/extensions/pull/4061
+
+      warning NU1603: Microsoft.AspNetCore.Telemetry depends on Microsoft.NET.ILLink.Tasks (>= 7.0.100-1.23211.1)
+                      but Microsoft.NET.ILLink.Tasks 7.0.100-1.23211.1 was not found.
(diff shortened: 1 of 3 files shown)


Directory.Build.targets
```diff
@@ -13,6 +13,15 @@
...
     <WarningLevel>5</WarningLevel>
 
+    <!--
+      Related to https://github.com/dotnet/extensions/pull/4061
+
+      warning NU1603: Microsoft.AspNetCore.Telemetry depends on Microsoft.NET.ILLink.Tasks (>= 7.0.100-1.23211.1)
+                      but Microsoft.NET.ILLink.Tasks 7.0.100-1.23211.1 was not found.
+                      An approximate best match of Microsoft.NET.ILLink.Tasks 8.0.0-alpha.1.23060.9 was resolved.
+    -->
+    <NoWarn>$(NoWarn);NU1603</NoWarn>
```
joperezr, line 23: I don't think we should do this for several reasons:
- Makes the builds less reprodusable in the future. By having the build just falling back to grabbing whatever the latest version available at the time of the build, makes the build flaky, as well as not reprodusable in the future.
- It is potentially silencing many other similar problems. NoWarn approach is a hammer, and applies to everything. This means that if there are any other packages with a similar problem, now we would be silencing those warnings.
- SDK is binding to that version for a reason, so we shouldn't try to overwrite it here.

If we really wanted to remove the the feed sooner rather than later, then my suggestion would be that the right fix is to push the missing versions to the dotnet8 feed of this package, given that technically the .NET 8 SDK and build requires that package, so it should be in that feed.

PR #4087, opened 2023-06-19
MetricCollector revamp.

geeknoid:
MetricCollector revamp.

- This now has a considerably simpler API surface allowing you to collect
data for a single instrument at a time.

- You can now snapshot all recorded measurements and apply various
filters over lists of measurements.

- General shape of the API looks similar in style to the FakeLogger.
You can now take snapshots of existing state, as a client you never get
your hands on "live" state, which helps with racy code.

- There is now a WaitForMeasurementsAsync function which is helpful to
simplify testing async logic.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4087)


src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -2,446 +2,540 @@
...
-    /// Gets a measurement value recorded by a counter instrument.
+    /// Gets the latest metric value that was updated.
     /// </summary>
-    /// <typeparam name="T">The type of metric measurement value.</typeparam>
-    /// <param name="counterName">The metric name.</param>
-    /// <param name="tags">The dimensions collection describing the measurement value.</param>
-    /// <returns>The measurement value or <see langword="null"/> if the value was not recorded.</returns>
-    public T? GetCounterValue<T>(string counterName, params KeyValuePair<string, object?>[] tags)
-        where T : struct => GetCounterValues<T>(counterName)?.GetValue(tags);
+    public MetricValue<T>? LatestMetricValueUpdated
```
noahfalk, line 239: How important do you feel it is to have the collector producing both measurements and metrics? 

The API defines the metric value of counters as the sum of measurements and for everything else it is identical to the last measurement value. Claiming the last measurement value of a histogram is the metric value feels like a very strange definition that is picked solely to force it to use the same MetricValue\<T> type as other instrument types.

Could we eliminate the MetricValue concept on this API and instead work with lists of measurements, last measurements, and if needed sums of measurements? I believe for all the tests modified by this PR other than tests of MetricCollector itself, they would all pass if LatestMetricValueUpdated was substituted with LatestMeasurementTaken. There were a couple tests that tested LatestMetricValueUpdated.Value on a Counter instrument but the expected value was always 1. I assume it would be most common when testing a counter to do some action in the unit test that increments the counter once in which case the sum is also equal to the last measurement.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -2,446 +2,540 @@
...
 
     /// <summary>
-    /// Collects all observable instruments and records their measurements.
+    /// Scan all registered recorded instruments.
     /// </summary>
-    public void CollectObservableInstruments()
+    public void RecordObservableInstruments()
     {
-        _listener.RecordObservableInstruments();
+        if (_instrument is ObservableInstrument<T>)
```
noahfalk, line 314: Nit: I don't think this conditional is necessary, MeterListener.RecordObservableInstruments() should already be a no-op in this case.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -2,446 +2,540 @@
...
     }
 
-    /// <inheritdoc />
-    public void Dispose()
+    private static AggregationType GetAggregationType(Instrument instrument)
     {
-        Dispose(true);
-        GC.SuppressFinalize(this);
+        Type genericDefinedType = instrument.GetType().GetGenericTypeDefinition();
+        AggregationType aggregationType = AggregationType.Save;
```
noahfalk, line 323: Nit: I think we now have 3 enum values for only 2 possible behaviors. These are names that made the most sense to me if you wanted to use them:
Aggregate => Sum
Save => LastValue
SaveOrUpdate => LastValue

test/Libraries/Microsoft.AspNetCore.HeaderParsing.Tests/HeaderParsingFeatureTests.cs
```diff
@@ -139,10 +139,11 @@ public void TryParse_returns_false_on_error()
 
         Assert.Equal("Can't parse header 'Date' due to 'Unable to parse date time offset value.'.", _logger.Collector.LatestRecord.Message);
 
-        var latest = metricCollector.GetCounterValues<long>(@"R9.HeaderParsing.ParsingErrors")!.LatestWritten!;
+        var latest = metricCollector.LatestMetricValueUpdated!;
         latest.Value.Should().Be(1);
-        latest.GetDimension("HeaderName").Should().Be("Date");
-        latest.GetDimension("Kind").Should().Be("Unable to parse date time offset value.");
+        Assert.True(latest.ContainsDimensions(
```
noahfalk, line 144: Its totally up to you how you want to do these tests, but I'd expect this pattern makes the assert failures less informative. Instead of seeing something like "Expected value 'Date' but received actual value 'OtherDate'" now the assert will be a more vague "Assert failed, line 144".

test/Libraries/Microsoft.AspNetCore.Telemetry.Middleware.Tests/Metering/HttpMeteringTests.cs
```diff
@@ -49,7 +49,7 @@ public HttpMeteringTests()
     [Fact]
     public async Task CanLogIncomingRequestMetric()
     {
-        using var metricCollector = new MetricCollector(new List<string> { typeof(HttpMeteringMiddleware).FullName! });
+        using var metricCollector = new MetricCollector<long>(typeof(HttpMeteringMiddleware).FullName!, null, Metric.IncomingRequestMetricName);
```
noahfalk, line 52: You probably want to start using IMeterFactory and specifying a scope for these tests. Binding by name in global scope makes the tests non-deterministic when run in parallel with other unit tests manipulating a Meter of the same name.

noahfalk (commented):
I like it a lot better than the previous API. Added some suggestions inline.

test/Libraries/Microsoft.AspNetCore.Telemetry.Middleware.Tests/Metering/HttpMeteringTests.cs
geeknoid: Can you show me an example?

The test is trying to bind to a meter that's embedded in the implementation of the middleware. So do you mean using a meter factory in the middleware itself, or merely in the test?

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
geeknoid: All right, I nuked the MetricValue type and associated logic, and introduced instead a few extension methods to simplify manipulating lists of meaaurements.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
geeknoid: Gone.

geeknoid:
@noahfalk @tarekgh Addressed Noah's feedback. It's looking even more like InstrumentRecorder, but with time support, wait support, and a few additional conveniences.

tarekgh:
@geeknoid I looked at MetricCollector and it looks good to me. I don't have any more feedback. 

@JamesNK do you have any feedback from aspnet side replacing InstrumentRecorder with MetricCollector?

tarekgh:
do you need this empty file? [Microsoft.Extensions.Telemetry.Console.json](https://github.com/dotnet/extensions/pull/4087/files#diff-899229afe9c796b4ce1be2c3e501723c768b821241248169ee7fb8a1939699ff)

tarekgh (approved):

geeknoid:
@tarekgh Where do you see this .json file? That's an old file that should have been deleted last week when I retired the console package. I don't see that file in my PR.

tarekgh:
> Where do you see this .json file? That's an old file that should have been deleted last week when I retired the console package. I don't see that file in my PR.

image

geeknoid:
@noahfalk Now at 100% code coverage, so ready to check in. Any more feedback?

noahfalk:
> @noahfalk Now at 100% code coverage, so ready to check in. Any more feedback?

You are speedy :) I will take a look later this evening at your updated version.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/CollectedMeasurement.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// <summary>
+    /// Gets the measurement's dimensions.
+    /// </summary>
+    public IReadOnlyDictionary<string, object?> Tags { get; }
+
+    /// <summary>
+    /// Gets a string representing the measurement.
+    /// </summary>
+    /// <returns>The string representation of this measurement.</returns>
+    public override string ToString() => $"[{Timestamp.ToString("HH:mm:ss.ffff", CultureInfo.InvariantCulture)}] {Value}";
```
JamesNK, line 58: nit: I think this would be better in a DebuggerDisplay attribute. I don't imagine anyone would use this string at runtime.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/CollectedMeasurement.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// </summary>
+    /// <param name="value">The measurement's value.</param>
+    /// <param name="tags">The dimensions of this measurement.</param>
+    /// <param name="timestamp">The time that the measurement occurred at.</param>
+    internal CollectedMeasurement(T value, ReadOnlySpan<KeyValuePair<string, object?>> tags, DateTimeOffset timestamp)
+    {
+        var d = new Dictionary<string, object?>();
+        foreach (var tag in tags)
+        {
+            d[tag.Key] = tag.Value;
```
JamesNK, line 31: This will lose duplicate tag values. Last tag with a name wins. Do we care?

JamesNK:
I had a lot of feedback for the wait for measurement methods. I created a PR with the changes that merges into this branch: https://github.com/dotnet/extensions/pull/4091

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/CollectedMeasurement.cs
geeknoid: I asked Noah about this yesterday, and there doesn't seem to be a consensus on first-one-wins or last-one-wins. This is definitely interpreted as a dictionary at the end of the day. So I opted for last-one-wins.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/CollectedMeasurement.cs
geeknoid: Can I express this as an attribute? As it is, the debugger does call ToString so its shows up in the UI. I'd be happy with it as a display attribute instead if that can be expressed.

JamesNK: Yes. See https://github.com/dotnet/runtime/pull/87754 for some examples.

Add:

```csharp
internal string DebuggerToString()
{
    return $"[{Timestamp.ToString("HH:mm:ss.ffff", CultureInfo.InvariantCulture)}] {Value}";
}
```

and

```csharp
[DebuggerDisplay("{DebuggerToString(),nq}")]
```

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/CollectedMeasurement.cs
noahfalk: > This is definitely interpreted as a dictionary at the end of the day

Some scenarios treat it as a list and don't remove duplicates at all :) I have no qualms with this though.

test/Libraries/Microsoft.AspNetCore.Telemetry.Middleware.Tests/Metering/HttpMeteringTests.cs
```diff
@@ -49,7 +49,7 @@ public HttpMeteringTests()
     [Fact]
     public async Task CanLogIncomingRequestMetric()
     {
-        using var metricCollector = new MetricCollector(new List<string> { typeof(HttpMeteringMiddleware).FullName! });
+        using var metricCollector = new MetricCollector<long>(null, typeof(HttpMeteringMiddleware).FullName!, Metric.IncomingRequestMetricName);
```
noahfalk, line 52: I lost track of that other comment that was talking about using scopes but the basic pattern is this:
1. When production code that is DI-aware creates a Meter, it should do so by taking an IMeterFactory as an injected constructor parameter and then creating the Meter from the factory. That factory now serves as the scope for the Meter.
2. When you want to test later, get the factory out of the DI container and use it as the scope parameter to the collector. This ensures that the Meter the test binds to is the one created by the factory in your test's container and not one that was created in a different DI container from a parallel test.

(If you have a way to get a reference to the Meter or a reference to the Instrument that works fine too, but often the components which create Meters and instruments don't expose those references)

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -2,446 +2,336 @@
...
 /// </summary>
-/// <remarks>
-/// This type has been designed to be used only for testing purposes.
-/// </remarks>
+/// <typeparam name="T">The type of metric data being recorded.</typeparam>
 [Experimental]
-[DebuggerDisplay("Count = {Count}")]
-public partial class MetricCollector : IDisposable
+[DebuggerDisplay("{_measurements.Count} measurements")]
+public sealed class MetricCollector<T> : IDisposable
```
noahfalk, line 22: Now that this helper only deals in measurements rather than metrics, perhaps it should be called 'InstrumentRecorder' or 'MeasurementRecorder'. I suspect many devs won't notice the difference but we've tried to use the verb "Record" when saving or reporting raw measurements. For example it is `Instrument.Record()` and `MeterListener.RecordObservableInstruments()`. Whereas collection is described in [our metrics docs](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-collection): "The process of aggregating, transmitting, and storing data is called collection". `CollectedMeasurement` would be similarly updated to `RecordedMeasurement`.

noahfalk (approved):
A few last suggestions, but this is looking pretty good to me. I think it is capturing all the things I liked about the original InstrumentRecorder and has removed portions of the MetricCollector API that felt confusing. Hopefully it still felt pretty ergonomic to use?

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
geeknoid: I'm going to keep collector as the moniker as it matches the terminology we use in other fakes, like the logging fakes and the redaction fakes.

PR #4091, opened 2023-06-20
Improve waiting for metrics

JamesNK:
Builds on top of https://github.com/dotnet/extensions/pull/4087. I thought making these changes myself would be easier than explaining them in code review.

* Add cancellation token to `WaitForMeasurementsAsync`.
* Simplify timeout method to pass cancellation token to `WaitForMeasurementsAsync`.
* Previously, waiters weren't removed from the collector if they timed out/were canceled. They are now removed.
* The TCS is now created with `RunContinuationsAsynchronously` to ensure awaiting user code isn't run when holding the `_measurements` lock.

cc @geeknoid @noahfalk @tarekgh 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4091)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
@@ -33,6 +33,7 @@ public sealed class MetricCollector<T> : IDisposable
         typeof(decimal),
     };
 
+    internal int WaitersCount => _waiters.Count; // Internal for testing
     private readonly MeterListener _meterListener = new();
     private readonly List<CollectedMeasurement<T>> _measurements = new();
     private readonly List<Waiter> _waiters = new();
@@ -136,8 +137,9 @@ public sealed class MetricCollector<T> : IDisposable
             // wake up anybody still waiting and inform them of the bad news: their horse is dead...
             foreach (var w in _waiters)
             {
-                w.TaskSource.SetException(MakeObjectDisposedException());
+                _ = w.TaskSource.TrySetException(MakeObjectDisposedException());
             }
+            _waiters.Clear();
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -136,8 +137,9 @@ public void Dispose()
             // wake up anybody still waiting and inform them of the bad news: their horse is dead...
             foreach (var w in _waiters)
             {
-                w.TaskSource.SetException(MakeObjectDisposedException());
+                _ = w.TaskSource.TrySetException(MakeObjectDisposedException());
```
JamesNK, line 140: I'm not sure if `ObjectDisposedException` is the right thing to do here. Setting the TCS to canceled might be more idiomatic. I'd need to look at what other APIs do for pending async tasks on dispose.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -200,57 +202,58 @@ public IReadOnlyList<CollectedMeasurement<T>> GetMeasurementSnapshot(bool clear
...
+    /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>A task that completes when the collector has collected the requisite number of measurements.</returns>
-    [SuppressMessage("Resilience", "R9A061:The async method doesn't support cancellation", Justification = "Not relevant in this case")]
-    public Task WaitForMeasurementsAsync(int numMeasurements)
+    public Task WaitForMeasurementsAsync(int minCount, CancellationToken cancellationToken = default)
     {
-        _ = Throw.IfLessThan(numMeasurements, 1);
+        _ = Throw.IfLessThan(minCount, 1);
+
+        ThrowIfDisposed();
```
geeknoid, line 212: This should be inside the lock block, otherwise it is not atomic.

geeknoid:
So what's the scenario where a cancellation token is useful?

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
geeknoid: I decided not to use cancel semantics here since I think that would be surprising. The caller didn't cancel anything, they disposed an object.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -200,57 +202,58 @@ public IReadOnlyList<CollectedMeasurement<T>> GetMeasurementSnapshot(bool clear
...
+            {
+                lock (_measurements)
+                {
+#if NET6_0_OR_GREATER
+                    w.TaskSource.TrySetCanceled(cancellationToken);
+#else
+                    w.TaskSource.TrySetCanceled();
+#endif
+
+                    _waiters.Remove(w);
```
geeknoid, line 238: Maybe _waiters should be changed to a HashSet instead of a List?

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -200,57 +202,58 @@ public IReadOnlyList<CollectedMeasurement<T>> GetMeasurementSnapshot(bool clear
...
-                return Task.CompletedTask;
-            }
-
-            var w = new Waiter(numMeasurements);
-            _waiters.Add(w);
-
-            return w.TaskSource.Task.WaitAsync(timeout, timeProvider ?? _timeProvider, CancellationToken.None);
-        }
+        using var cancellationTokenSource = new CancellationTokenSource(timeout);
+        await WaitForMeasurementsAsync(minCount, cancellationTokenSource.Token).ConfigureAwait(false);
```
geeknoid, line 256: Why the await here? Can't this just be tail-optimized to just return a task without the await?

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
JamesNK: A list is fine. There will almost always be 0 or 1 waiters. I can't imagine why there would be enough to make a hashset worth using over a list.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
JamesNK: Because the CTS is disposed with the using:

```csharp
using var cancellationTokenSource = new CancellationTokenSource(timeout);
```

Without the await then the CTS would immediately be disposed when the method exits.

JamesNK:
> So what's the scenario where a cancellation token is useful?

The developer doesn't want to wait for measurements anymore for whatever reason. Maybe they have a test that waits for results from two collectors. When one successfully returns a result they cancel waiting for the other.

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
JamesNK: I looked up what another type does here: `HttpClient` cancels any outstanding requests instead of throwing ODE - https://github.com/dotnet/runtime/blob/05e25ff8c3a623df74720f430c270eba0f1d41b0/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L706-L710

I don't care much, but I'm pretty sure most .NET APIs clean up pending tasks on dispose by canceling the tasks, not throwing ODE.

JamesNK:
@geeknoid I've rebased on main and resolved conflicts. All outstanding feedback has been addressed. Ready for review and merge?

geeknoid (approved):

Issue #3996, opened 2023-05-25
Improve FakeTimeProvider into a canonical fake


danmoseley:
The one in this repo has some issues/limitations, 

1. Consider allowing an arbitrary TimespanFrequency.

Currently it is set to 10^7 == TimeSpan.TicksPerSecond. The actual TimeProvider.System uses TimespanFrequency  == Stopwatch.Frequency == typically 10^6 or 10^9 (and does not change once the system is started).  A different TimespanFrequency could help spot bugs where code assumes TimeSpan.TicksPerSecond. 

Note that while GetElapsedTime() abstracts Frequency, Kestrel generally does math like `Expiration = currentTimestamp + timeoutTimeSpan.TotalSeconds * timeProvider.TimespanFrequency`, and then later `if (currentTimestamp > Expiration) ... `. See [here for some examples](https://github.com/dotnet/aspnetcore/pull/48081/files)

If TimespanFrequency changes we would need to take care about truncation. We [don't do math here](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs#L93-L94) in order to keep stable test results - essentially long has more digits than double. GetElapsedTime() is not virtual, so can't be fixed; one option is possibly to ensure that the base ticks are rounded up front to the nearest 1ms or so.

2. Consider allowing an arbitrary base offset. Eg., one of the Kestrel ones [randomizes the base offset](https://github.com/dotnet/aspnetcore/pull/48081/files#diff-eb56b63e16641b4b01665470700a08c48b687ec6fc9eb6157fbf3906d68c187aR19). Many of the tests in this repo use UtcNow, which actually adds non reproducibility, [example](https://github.com/dotnet/extensions/blob/main/test/Libraries/Microsoft.Extensions.Http.Telemetry.Tests/Logging/HttpLoggingHandlerTest.cs#L341).

3. Look at the TimeProvider's in dotnet/aspnetcore and dotnet/runtime possibly elsewhere, and consider whether we can make one with the combined features that we can share via dotnet/arcade instead of each having our own. Ideally we have a single test TimeProvider that can be configured however needed. Right now we have numerous ones.

danmoseley:
Here's some extension methods they used in Kestrel
```c#
    public static long ToTicks(this TimeSpan timeSpan, TimeProvider timeProvider)
        => timeSpan.ToTicks(timeProvider.TimestampFrequency);

[...]

PR #4092, opened 2023-06-20
FakeTimeProvider revamp.

geeknoid:
- Address a number of race conditions. Several paths could lead to time
potentially moving backwards, wake events being missed, etc.
All read-modify-write updates to _now are protected by lock to ensure
time only ever moves forward.

- Reintroduce the notion of automatically advancing time. This is
essential in scenarios where time is read in some code not under direct
control of the test.

- Remove the Epoch property and rename the epoch constructor parameter
to startTime instead. An epoch defines what time it is when a clock's
counters is at 0, the base point for the clock. This is now what the
value is used for here, it is used here merely as a starting date and
time value for the provider's idea of 'current time'.

- Added some missing argument checking in a few places.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4092)


danmoseley:
@egil

egil:
> * Reintroduce the notion of automatically advancing time. This is essential in scenarios where time is read in some code not under direct control of the test.

@geeknoid, this is super interesting. Can you share a few examples where "time is read in some code not under direct control of the test"?

geeknoid:
> > * Reintroduce the notion of automatically advancing time. This is essential in scenarios where time is read in some code not under direct control of the test.
> 
> @geeknoid, this is super interesting. Can you share a (few) examples where "time is read in some code not under direct control of the test"?

Simple example:

```
TimeProvider provider = new FakeTimeProvider()
var x = myObject.MyFunctionUnderTest(timeProvider);
```

And MyFunctionUnderTest reads the clock N times and expects time to move forward.

geeknoid:
And in fact, I was wondering if auto advance should be the default, maybe by one tick? What do you think @egil ?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -12,83 +12,109 @@
...
+
+        List<FakeTimeProviderTimer.Waiter>? waitersToWake;
+
+        lock (Waiters)
+        {
+            _now += delta;
+            waitersToWake = GetWaitersToWake();
+        }
+
+        WakeWaiters(waitersToWake);
```
egil, line 116:
```suggestion
        SetUtcNow(_now + delta);
```

My instinct is to do this instead of duplicating the logic in both Advance and SetUtcNow. What is your thought process around not doing this?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -12,83 +12,109 @@
...
+            {
+                Throw.ArgumentOutOfRangeException(nameof(value), $"Cannot go back in time. Current time is {_now}.");
+            }
+
+            _now = value;
+            waitersToWake = GetWaitersToWake();
         }
 
-        _now = value;
+        WakeWaiters(waitersToWake);
```
egil, line 97: I think this may break the promise that callbacks happen in the correct order.

E.g., we have two timers:

1. `timer1` has callback `duetime = 1 second` and `period = 1 second`
2. `timer2` has callback `duetime = 3 seconds` and `period = infinite`

If call advance time by 3 seconds, then I expect the callback order to be `[timer1, timer1, timer2]`, but as far as I can tell, the code in the PR will invoke the timers callbacks `[timer1, timer2]` only.

egil:
> Simple example:
> 
> ```
> TimeProvider provider = new FakeTimeProvider()
> var x = myObject.MyFunctionUnderTest(timeProvider);
> ```
> 
> And MyFunctionUnderTest reads the clock N times and expects time to move forward.

I am not sure I understand your example. It looks to me like `MyFunctionUnderTest` is explicitly under the control of the test.

Are you perhaps referring to 3rd party code that you do not own and thus must treat like a "black box", e.g. an external library that also uses TimeProvider, that expects a new DateTimeOffset each time it calls `GetUtcNow()`?

If it's the latter, here are a few points for your consideration:

**First**, is there a guarantee that calling `DateTimeOffset.UtcNow` twice in a row yields different results? I suspect not, and if the code depends `DateTimeOffset.UtcNow < DateTimeOffset.UtcNow` to always be true I'd probably raise some concerns with the developers :)

**Second**, the point of using a `TimeProvider` and `FakeTimeProvider` when testing 1st party code is to create fast and deterministic tests. 

With that in mind, let's say our system-under-test (SUT) looks like this:

```c#
public class MySUT
{
  private readonly Dependency dependency;

  public MySUT(TimeProvider timeProvider)
  {
    dependency = new Dependency(timeProvider);
  }

  public void DoStuff() 
  {
    // uses methods on `dependency`
    // does other stuff with timeProvider etc
  }
}
```

Let's suppose the "auto advance feature" is enabled and `Dependency` is 3rd party code/library, I probably want to pass `TimeProvider.System` to the 3rd party code (`new Dependency(TimeProvider.System)`) instead of passing down `timeProvider` as in the code above. Otherwise, my test could easily break if the 3rd party code changes, e.g. if it calls `GetUtcNow()` more or fewer times than at the time I was writing my tests.

`Dependency` may also be 1st party code that is not directly under test but is being used by my SUT during testing, and again, if the "auto advance feature" is enabled, my SUT's test could easily break if the other parts of the code change how many times it calls `GetUtcNow()`.

In the latter case, where `Dependency` is 1st party code, I will know what the code expects and can decide if I need to pass in `TimeProvider.System` or share the same `timeProvider` instance that my SUT is using.

In short, I am not convinced having the "auto advanced feature" is a good idea and I suspect there may be alternative solutions to the problem you are trying to solve. 

> And in fact, I was wondering if auto advance should be the default, maybe by one tick? What do you think @egil ?

Hard no (if we do keep the "auto advanced feature"). I do not want one test to break because an implementation detail in a dependency was changed.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -17,20 +17,9 @@ namespace Microsoft.Extensions.Time.Testing;
...
-    /// <summary>
-    /// Gets the time which was used as the starting point for the clock in this <see cref="FakeTimeProvider"/>.
-    /// </summary>
-    /// <remarks>
-    /// This can be set by passing in a <see cref="DateTimeOffset"/> to the constructor
-    /// which takes the <c>epoch</c> argument. If the default constructor is used,
-    /// the clock's start time defaults to midnight January 1st 2000.
-    /// </remarks>
-    [Experimental]
-    public DateTimeOffset Epoch { get; } = new DateTimeOffset(2000, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
```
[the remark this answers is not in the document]
geeknoid: I don't think this is a desirable thing, the start time is of extremely limited value since typically it will be out of date instantly. I'd rather not expose a property whose sole use is to read it 2 lines after you have given the value to the constructor.

So I'm going to wait on that until a compelling use case surfaces.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: Indeed, it doesn't guarantee order. And I think that's fine, there is no promise here. This is fundamentally racy in nature and ordering here is an illusion anyway for those edge cases, as the processor scheduler can readily change the effective order that the waiters are woken up.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: This would break atomicity and could lead to unexpected exceptions. If _now gets updated asynchronously after it is read here, the call to SetUtcNow could throw an exception because time is being set backward.

geeknoid:
Let me explain my simple example in more detail:

```
var timeProvider = new FakeTimeProvider();
Assert.True(myClassUnderTest.TimeMarchesOn(timeProvider));
```

```
public bool TimeMarchesOn(TimeProvider timeProvider)
{
    var start = timeProvider.GetUtcNow();
    Thread.Sleep(1000);
    var end = timeProvider.GetUtcNow();

    return end != start;
}
```

This is obviously a silly example, but it's easy to construct code where you assume that time does move forward and there is no way for this behavior to be triggered from the test case itself. There are situations in fact where you should override the FakeTimeProvider.GetUtcNow function in order to provide customized time advancing semantics for tricky cases to test.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: If two callbacks have the same scheduled callback time, then yes, I agree that the ordering is not likely to be guaranteed in production either, that said, for the sake of deterministic tests, it would be nice if they were.

However, in my example above, timer1's callback should be called twice before timer2 callback. That was one of the points of https://github.com/dotnet/extensions/pull/4022, but I realize I did not include a test with multiple timers.

This test should ensure the correct behaviour (just typed out here, not actually compiled/tested):

```c#
[Fact]
public void Timers_callbacks_are_ordered()
{
    var callbackTimes = new List<(int TimerId, DateTimeOffset CallbackTime)>();
    var timeProvider = new FakeTimeProvider();
    var start = timeProvider.GetUtcNow();
    var oneSecond = TimeSpan.FromSeconds(1);
    var threeSeconds = TimeSpan.FromSeconds(3);
    var timer1 = timeProvider.CreateTimer(_ => { callbackTimes.Add((1, timeProvider.GetUtcNow())); }, null, oneSecond, oneSecond);
    var timer2 = timeProvider.CreateTimer(_ => { callbackTimes.Add((2, timeProvider.GetUtcNow())); }, null, threeSeconds, threeSeconds);

    timeProvider.Advance(threeSeconds);

    Assert.Equal(new[]
    {
      (1, start + oneSecond),
      (1, start + oneSecond + oneSecond),
      (2, start + threeSeconds),
      (1, start + threeSeconds),
    },
    callbackTimes);
}
```

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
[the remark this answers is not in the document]
egil: Two points:

1. We do not always pass a startTime to the constructor. A factory method or library like AutoFixture could be used to create the FakeTimeProvider.

2. In more complex test scenarios you may want to pass the FakeTimeProvider to methods that move the SUT into a state before the main part of the test method starts executing. Then it is useful if the timeProvider carries around its "epoch"/"start time" since that info will be readily available in these methods, without having to pass that in as an additional argument.

Even in simpler test scenarios, as in tests for this library, having to either specify a `var start = DateTimeOffset.UtcNow` before instantiating `var timeProvider = FakeTimeProvider(start)` or calling `var start = timeProvider.GetUtcNew()` seems like unnecessary work compared to just being able to reference the start time directly from the object that owns it, e.g. `timeProvider.StartTime`.

geeknoid:
The broader point I'm trying to make here is that a test case calls a single function to test and this test function samples time repeatedly as part of its implementation. Heck, replace the Thread.Sleep I had with Task.Delay(timeProvider). Without auto update, any code that you are testing which consults the TimeProvider and expects time to move forward will not work.

egil:
> The broader point I'm trying to make here is that a test case calls a single function to test and this test function samples time repeatedly as part of its implementation. Heck, replace the Thread.Sleep I had with Task.Delay(timeProvider). Without auto update, any code that you are testing which consults the TimeProvider and expects time to move forward will not work.

The point of `FakeTimeProvider` for me is to control time from the test code, *not* from the production code. 

But I agree there are multiple scenarios in play here where a TimeProvider is used by production code (I do not consider scenarios that do not use a TimeProvider like `Thead.Sleep` or other APIs that uses the system clock directly).

NOTE: My assumption here is that a good test should not be tied to implementation-details of a SUT. Tests should validate the externally visible behavior of the SUT via its public API surface/contract. 

So let's try to list all scenarios. Here are the ones that come to mind atm:

### 1. Time passing is an implementation in the SUT.

E.g. a SUT may have a Task.Delay of 1 second or 10 seconds or multiple delays inserted and read `GetUtcNow()` zero or more times, but it is not something that is externally visible and should not affect the assertion in a test. 

Since the passing of time is internal to the SUT, we cannot set an `autoAdvanceAmount` in our test without breaking the "do not tie your test to implementation details of the SUT", since setting `autoAdvanceAmount` to a meaningful value that will be useful to the SUT means leveraging the knowledge of the implementation-details of the SUT in our tests.

That does not mean a FakeTimeProvider is not useful in this test to speed up execution. I guess you could do a `timeProvider.SetUtcNow(DateTimeOffset.MaxValue)` to move to the end of time.

However, there are caveats with `timeProvider.SetUtcNow(DateTimeOffset.MaxValue)` and auto-advance. For example, 

* if our SUT has a timer with 1 second period and 1 second due time and a callback that calls `GetUtcNow`, 
* and the FakeTimeProvider has auto-advance set to 1 second

Then we are going to have a callback loop that never completes. Each time the callback read GetUtcNow, the timer advanced time by 1 second, causing the callback to be triggered again, and so on. The same is true for `timeProvider.SetUtcNow(DateTimeOffset.MaxValue)`.

### 2. Time passing is externally visible/observable (business/domain rule).

E.g. a test of a circuit breaker (Polly) algorithm that has to wait a certain amount of time before retrying or a benchmark tool that measures time/timestamps from start to finish of an operation.

In this case, we want to explicitly control the progress of time in our tests and would not rely on auto-advance, since the amount of time passing is an important part of the scenario being tested.

### 3. SUT uses dependent code that depends on TimeProvider.

There are two sub-options here. 

**3.1:** In some cases, your SUT is depending on something whose public contracts include insight into how it deals with time passing, i.e. it is configurable. An example here is `ITimer`, `PeriodicTimer`, or `Task.Delay(timespan, timeprovider)`, where you control when callbacks are supposed to happen. In these cases, the FakeTimeProvider instance should be passed down to the dependencies.

**3.2:** If time passing is an implementation detail in the dependency, it is probably safest to pass `TimeProvider.System` to the dependency since our test SUT test should not be broken if an implementation detail in the dependency code changes. 

To gain the advantage of a fast-running time-independent test, we probably need to wrap the 3rd party code in an abstraction and mock/fake/stub that abstraction during testing to make its behavior deterministic. See also the general testing rule "don't mock code you don't own".

--------

It is entirely possible that you and others have different testing strategies than me and/or different problems you are trying to solve, so at the end of the day, it's you to you and the team to decide where you want to go with this, so please do not take my input here as more than me providing input and trying to learn in the process, inspired by the goal about this being a canonical fake outlined here https://github.com/dotnet/extensions/issues/3996.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: I revamped things and ordering promises are back.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: This has been revamped.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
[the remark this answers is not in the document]
geeknoid: OK, I added StartDateTime. I don't think it's actually useful, but it's trivial and doesn't hurt (hint: kinda like optional auto-advancing time 😜)

danmoseley:
As good a place as any to ask... are we discovering any feedback on the TimeProvider abstraction? We have a couple months before we can never change it. Eg i think there was some method we thought could be virtual (that change would not be breaking of course)

rafal-mz:
I think 0 is nice default for auto advance.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -109,92 +146,122 @@ public override long GetTimestamp()
...
+                    {
+                        // our first candidate
+                        candidate = waiter;
+                    }
+                    else if (waiter.WakeupTime < candidate.WakeupTime)
+                    {
+                        // found a waiter with an earlier wake time, it's our new candidate
+                        candidate = waiter;
+                    }
+                    else if (waiter.WakeupTime > candidate.WakeupTime)
```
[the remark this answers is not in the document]
geeknoid: They don't hurt anything and I like the clarity they bring. The code reads cleanly.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -7,88 +7,125 @@
...
-    /// The clock is set to not automatically advance every time it is read.
+    /// The provider is set to not automatically advance time each time it is read.
     /// </remarks>
-    public FakeTimeProvider()
+    [Experimental]
+    public FakeTimeProvider(DateTimeOffset startDateTime)
     {
-        _localTimeZone = TimeZoneInfo.Utc;
-        _now = Epoch;
+        _ = Throw.IfLessThanOrEqual(startDateTime.Ticks, 0);
```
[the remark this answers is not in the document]
geeknoid: I updated to < 0 instead of <= 0.

geeknoid:
> As good a place as any to ask... are we discovering any feedback on the TimeProvider abstraction? We have a couple months before we can never change it. Eg i think there was some method we thought could be virtual (that change would not be breaking of course)

Well, nearly everybody asks "Why isn't this called IClock" :-) Even Nick Chapsas was commenting in a video earlier this week on that very thing.

On the usability front, it seems to work.

danmoseley:
@egil did you have any feedback on TimeProvider so far?

egil:
> @egil did you have any feedback on TimeProvider so far?

Yes, I joined the ongoing issue on github towards the end of the design phase.

> > As good a place as any to ask... are we discovering any feedback on the TimeProvider abstraction? We have a couple months before we can never change it. Eg i think there was some method we thought could be virtual (that change would not be breaking of course)
> 
> Well, nearly everybody asks "Why isn't this called IClock" :-) Even Nick Chapsas was commenting in a video earlier this week on that very thing.
> 
> On the usability front, it seems to work.

The big problem most people have that want an `IClock`, including Nick, is that they just look at the `GetUtcNow` method and do not think about it in a larger context. TimeProvider is not just an abstraction over `UtcNow` it is also an abstraction over a timer (`ITimer`), and then, by extensions, other APIs like `PeriodicTimer`, `Task.Delay`, `CancellationTokenSource`, and others in the future, which can leverage TimeProvider.

I am honestly thinking about submitting a talk to a few conferences on how to use TimeProvider and FakeTimeProvider, because it is not immediately obvious to all :)

My biggest concern with the current implementation is related to the back-port to .NET 6 which does not behave like the .NET 8 version. I have highlighted the issue in https://github.com/dotnet/runtime/issues/85326 but it is yet to be fixed.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -7,88 +7,125 @@
...
+        _now = startDateTime;
+        _autoAdvanceAmount = autoAdvanceAmount;
+        StartDateTime = _now;
     }
 
+    /// <summary>
+    /// Gets the starting date and time for this provider.
+    /// </summary>
+    [Experimental]
+    public DateTimeOffset StartDateTime { get; }
```
egil, line 76: Isn't it redundant to include the word "DateTime" in the property name?

It's more succinct to call `Start`, or as discussed at length in the previous PR, `Epoch`.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -7,88 +7,125 @@
...
 namespace Microsoft.Extensions.Time.Testing;
 
 /// <summary>
-/// A synthetic clock used to provide deterministic behavior in tests.
+/// A synthetic time provider used to enable deterministic behavior in tests.
 /// </summary>
 public class FakeTimeProvider : TimeProvider
 {
-    internal readonly List<FakeTimeProviderTimer.Waiter> Waiters = new();
+    internal readonly HashSet<Waiter> Waiters = new();
```
egil, line 20: Should this not be private and named `waiters`?

egil:
Speaking of confusing users, I do **NOT** think a canonical `FakeTimeProvider` should have the proposed auto-advance feature.

As outlined in https://github.com/dotnet/extensions/pull/4092#issuecomment-1599250724, there are caveats and it is not clear that it is a commonly needed feature.

Additional points not highlighted in the linked comment: 

- If I receive a `FakeTimeProvider` in a test method that was created elsewhere, I have no idea whether or not it uses auto-advance or not. There is no public property on the object that tells me this.
- If I call `GetUtcNow` in my test code, e.g. to save the current time for later use, then I probably do not want to move forward.
- I think this falls into the same category as the `Advance()` method discussed previously which implicitly moved time forward by 1ms.

However, by not sealing `FakeTimeProvider`, anybody can implement their own `AutoIncrementFakeTimeProvider` if they have a need in a specific test scenario:

```c#
public class AutoIncrementFakeTimeProvider : FakeTimeProvider
{
  private readonly TimeSpan _autoAdvanceAmount;

  public AutoIncrementFakeTimeProvider(TimeSpan autoAdvanceAmount)
  {
    _autoAdvanceAmount = autoAdvanceAmount;
  }

  public void GetUtcNow()
  {
    var result = base.GetUtcNow();
    SetUtcNow(result + _autoAdvanceAmount);
    return result;
  }
}
```

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
egil, line 1: Should this file be renamed to `TimerTests`?

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
```diff
@@ -261,30 +264,32 @@ public void WaiterRemovedWhenCollectedWithoutDispose()
     [Fact]
     public void UtcNowUpdatedBeforeTimerCallback()
     {
-        var timeProvider = new FakeTimeProvider();
+        var start = DateTimeOffset.UtcNow;
+        var timeProvider = new FakeTimeProvider(start);
```
egil, line 268:
```suggestion
        var timeProvider = new FakeTimeProvider();
        var start = timeProvider.StartDateTime;
```

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
```diff
@@ -261,30 +264,32 @@ public void WaiterRemovedWhenCollectedWithoutDispose()
...
+        Assert.Equal(start + TimeSpan.FromMilliseconds(1000), value3);
     }
 
     [Fact]
     public void LongPausesTriggerMultipleCallbacks()
     {
+        var start = DateTimeOffset.UtcNow;
         var callbackTimes = new List<DateTimeOffset>();
-        var timeProvider = new FakeTimeProvider();
+        var timeProvider = new FakeTimeProvider(start);
```
egil, line 292:
```suggestion                
                var callbackTimes = new List<DateTimeOffset>();
                var timeProvider = new FakeTimeProvider();
                var start = timeProvider.StartDateTime;
```

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
```diff
@@ -319,13 +324,13 @@ public void MultipleTimersCallbackInvokedInScheduledOrder()
...
-            (1, TimeSpan.FromMilliseconds(9)),
-            (2, TimeSpan.FromMilliseconds(9)),
+            (1, TimeSpan.FromMilliseconds(11)),
+            (2, TimeSpan.FromMilliseconds(11)),
+            (3, TimeSpan.FromMilliseconds(11)),
+            (1, TimeSpan.FromMilliseconds(11)),
+            (2, TimeSpan.FromMilliseconds(11)),
+            (1, TimeSpan.FromMilliseconds(11)),
+            (2, TimeSpan.FromMilliseconds(11)),
             (3, TimeSpan.FromMilliseconds(11)),
```
egil, line 334: Why is the assertion changed here? The point of the test is to verify that invocations happen in order and at the expected time.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
egil: Actually, probably not, maybe just `FakeTimeProviderTests`?

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
geeknoid: The clock advanced by 11 milliseconds, why should the callbacks be annotated with intermediate time stamps? If the timer advanced in smaller increments, then the behavior would be reasonable.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: It's used by tests, so it needs to be internal.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
egil: This particular scenario is described in https://github.com/dotnet/extensions/issues/3995 and was fixed in https://github.com/dotnet/extensions/pull/4022.

In essence, this test verifies that the result of `GetUtcNow` was as expected each time the timer's callbacks were invoked, as well as the order the timers were invoked in.

Granted, this test should be split up in two, one for each condition, to make it clear to all what is going on, that's on me.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
egil: Ahh right. I do remember that. Not a big fan of depending on internal state in a test though. Would suggest changing that but that's for another PR.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
geeknoid: The test has an incorrect assumption that the time at which a timer callback is invoked corresponds to its requested wake time. This was an implementation artifact of the previous design and is not a contractual guarantee of the ITimer interface. The guarantee is that the callbacks will not be invoked before the due time, but invariable are invoked later than the due time.

I've updated the test to increment the clock gradually, which makes it so it satisfies the previously enforced conditions since the callbacks are invoked gradually. But this is still assuming that the timer behaves in a way that is not actually guaranteed by the API contract. I'm OK leaving the test this way however, it's not a big deal.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
geeknoid: Done.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
geeknoid: Done.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: Done.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
geeknoid: Renamed to TimerTests

geeknoid:
@egil I'm proceeding with the auto advance feature. I've added a property to return the instance's auto-advance amount. It is an optional feature that we've had internally for over 2 years without hiccups, so although I appreciate the concerns, I don't share them. It's the only option available for specific scenarios and has no impact on the behavior if not used. The behavior would not reliably be achieved through a derived class due to the lack of atomicity in updates.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/TimerTests.cs
```diff
@@ -294,13 +297,13 @@ public void LongPausesTriggerMultipleCallbacks()
...
         Assert.Equal(new[]
         {
-            timeProvider.Epoch,
-            timeProvider.Epoch + period,
-            timeProvider.Epoch + period + period,
-            timeProvider.Epoch + period + period + period,
+            timeProvider.Start,
+            timeProvider.Start + period + period + period,
+            timeProvider.Start + period + period + period,
+            timeProvider.Start + period + period + period,
```
egil, line 306: This is also wrong and breaks with the promise that the result returned from `GetUtcNow()` is correct at the time the callback happens.

If a timer has a period of 3 seconds, and you advance time by 9 seconds, the callback of the timer should be called three times, where the time is 3, 6, and 9 seconds from the start time. That is how things behave in production and is absolutely essential for the fake to be useful.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/TimerTests.cs
```diff
@@ -315,7 +318,10 @@ public void MultipleTimersCallbackInvokedInScheduledOrder()
         using var timer2 = timeProvider.CreateTimer(_ => callbacks.Add((2, timeProvider.GetElapsedTime(startTime))), null, TimeSpan.FromMilliseconds(3), TimeSpan.FromMilliseconds(3));
         using var timer3 = timeProvider.CreateTimer(_ => callbacks.Add((3, timeProvider.GetElapsedTime(startTime))), null, TimeSpan.FromMilliseconds(6), TimeSpan.FromMilliseconds(5));
 
-        timeProvider.Advance(TimeSpan.FromMilliseconds(11));
+        timeProvider.Advance(TimeSpan.FromMilliseconds(3));
+        timeProvider.Advance(TimeSpan.FromMilliseconds(3));
+        timeProvider.Advance(TimeSpan.FromMilliseconds(3));
+        timeProvider.Advance(TimeSpan.FromMilliseconds(2));
```
egil, line 324: We want to move time forward by 11ms in one step. That should still make the assertion pass below. The fake time provider should call the timer callbacks in order, with the internal `_now` set correctly at each callback time, independent of whether we move past one or more timer "period" callbacks.

geeknoid: Sorry, but this is fundamentally incorrect. There is no guarantee that the perceived wake time of a timer callback matches the specified due time. That's not how timers work.

egil:
> @egil I'm proceeding with the auto advance feature. I've added a property to return the instance's auto-advance amount. It is an optional feature that we've had internally for over 2 years without hiccups, so although I appreciate the concerns, I don't share them. It's the only option available for specific scenarios and has no impact on the behavior if not used. The behavior would not reliably be achieved through a derived class due to the lack of atomicity in updates.

Appreciate that. I'd love to see some examples if you can share them to probably understand the use case.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/TimerTests.cs
geeknoid: No, that's not the way it behaves.

The timer will in fact be invoked 3 times, and for all 3 times the value returned by DateTimeOffset.UtcNow will be greater than the expected due time. The design here will not drop any invocations, the right total count of callbacks invocations will occur. But the time at which the callbacks are invoked will not be the due time.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTimerTests.cs
egil: > The test has an incorrect assumption that the time at which a timer callback is invoked corresponds to its requested wake time. This was an implementation artifact of the previous design and is not a contractual guarantee of the ITimer interface. The guarantee is that the callbacks will not be invoked before the due time, but invariable are invoked later than the due time.

Sure, I know the production timer only gives us this guarantee, however, the FakeTimeProvider is deterministic (should be) and I see no reason it should not behave like this. 

This is the strength of the TimeProvider abstraction, that I can move time forward and have both a timer callback and a "UTC now" be in sync, yielding predictable tests. If the timer callback and the current time are not in sync, we may as well have an `IClock` interface that only exposes `GetUtcNow` and a `ITimer` interface that does its thing.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -7,88 +7,129 @@
...
+    /// Gets the starting date and time for this provider.
+    /// </summary>
+    [Experimental]
+    public DateTimeOffset Start { get; }
+
+    /// <summary>
+    /// Gets the amount of time by which time advances whenever the clock is read.
+    /// </summary>
+    [Experimental]
+    public TimeSpan AutoAdvanceAmount { get; } = TimeSpan.Zero;
```
egil, line 80: Sound this have a public setter?

egil (commented):
I dislike that `GetUtcNow()` doesn't return a predictable result when called from a timer callback, independent of the amount of time passed to `SetUtcNow()` or `Advance()` and I consider it a step back to before #3995 was closed. 

It has been an essential feature in the tests I've been writing using my own fake TimeProvider implementation at work, and my hope is that I will be able to replace that implementation with an "official Microsoft" one.

Yes, `ITimer` only guarantees that a callback is not invoked before "due time" or "period", but we are not breaking any design/object-oriented principles by strengthening this promise and invoking timer callbacks exactly on  "due time" or "period", and the benefit is deterministic behavior in tests.

Anyway, I'll tap out now and leave you folks to decide what you think is best 😊

Appreciate the patience!

danmoseley:
>  doesn't return a predictable result

Just to make sure I understand, as written it is predictable in the sense that by looking at the inputs, you can predict the output, right? It is both deterministic and predictable (unless perhaps there are multiple threads working on the fake). I agree  that a unit test should not expect this time provider to fire a timer exactly after its interval. That's something that doesn't happen in the real world. We want the fake to be deterministic, correct, and realistic.

egil:
> LMK if I misread.

@danmoseley I may not have been clear.

geeknoid:
@danmoseley Single-threaded use of FakeTimeProvider is fully deterministic. It will behave with rock-solid consistency, respecting the ITimer contract. When the timer callback reads the clock, it will always see a value that is >= to the timer's due time, and this value will be consistent across test runs.

The property we're losing here is the fact the prior implementation tried to ensure that the callback would always be triggered and be able to get the current time and it would always equal the timer's due time.

danmoseley:
OK I acknowledge your point @egil but I agree with @geeknoid here. The consistency we want is it's the same each time you run the test.

egil:
> OK I acknowledge your point @egil but I agree with @geeknoid here. The consistency we want is it's the same each time you run the test.

That's fair if you want to go in that direction. But I am pretty sure that there are going to be others that will be surprised about this.

Even though the behavior of the test example below is _valid_ according to the specification of `ITimer`, I do not expect anybody would expect this to be the outcome. I expect most will expect what I am advocating, that when a callback is triggered, the `GetUtcNow()` returns the same independent of what the end time the fake is advancing to.

```c#
    [Fact]
    public void MultipleTimersCallbackInvokedInScheduledOrder()
    {
        var callbacks = new List<(int timerId, TimeSpan callbackTime)>();
        var timeProvider = new FakeTimeProvider();
        var startTime = timeProvider.GetTimestamp();
        using var timer1 = timeProvider.CreateTimer(_ => callbacks.Add((1, timeProvider.GetElapsedTime(startTime))), null, TimeSpan.FromMilliseconds(3), TimeSpan.FromMilliseconds(3));
        using var timer2 = timeProvider.CreateTimer(_ => callbacks.Add((2, timeProvider.GetElapsedTime(startTime))), null, TimeSpan.FromMilliseconds(3), TimeSpan.FromMilliseconds(3));
        using var timer3 = timeProvider.CreateTimer(_ => callbacks.Add((3, timeProvider.GetElapsedTime(startTime))), null, TimeSpan.FromMilliseconds(6), TimeSpan.FromMilliseconds(5));

        timeProvider.Advance(TimeSpan.FromMilliseconds(11));

        Assert.Equal(new[]
        {
            (1, TimeSpan.FromMilliseconds(11)),
            (2, TimeSpan.FromMilliseconds(11)),
            (3, TimeSpan.FromMilliseconds(11)),
            (1, TimeSpan.FromMilliseconds(11)),
            (2, TimeSpan.FromMilliseconds(11)),
            (1, TimeSpan.FromMilliseconds(11)),
            (2, TimeSpan.FromMilliseconds(11)),
            (3, TimeSpan.FromMilliseconds(11)),
        },
        callbacks);
    }
```

Unless users are aware of this, they will spend a lot of time debugging, experimenting, and seeing inconsistent behavior between moving in small increments vs. large.

I wonder if there is no way to get both auto-advance and consistent results from `GetUtcNow()`?

egil:
OK, I promised I was done, but let me just share this ridiculous example.

In the test below, with the latest version of the code in this PR at the time of writing, we have a timer with a `dueTime` of 1 millisecond and a period of 1 hour.

What would you expect should be in the `callbacks` collection after we advance time by 1 hour and 1 millisecond?

```c#
[Fact]
public void GetUtcNowFromTimerCallback()
{
    var callbacks = new List<TimeSpan>();
    var timeProvider = new FakeTimeProvider();
    var startTime = timeProvider.GetTimestamp();
    using var timer1 = timeProvider.CreateTimer(
        callback: _ => callbacks.Add(timeProvider.GetElapsedTime(startTime)),
        state: null,
        dueTime: TimeSpan.FromMilliseconds(1),
        period: TimeSpan.FromHours(1));

    timeProvider.Advance(TimeSpan.FromHours(1) + TimeSpan.FromMilliseconds(1));

    // Try to guess what callbacks[0], aka. the 1ms dueTime callback, should hold
    // and what callbacks[1], aka. the 1 hour period callback, should hold.
}
```

<details>
<summary>Expand this to see the actual assertion that will pass</summary>

This is the assertion that will pass currently:

<pre><code>Assert.Equal(2, callbacks.Count);
Assert.Equal(TimeSpan.FromHours(1) + TimeSpan.FromMilliseconds(1), callbacks[0]);
Assert.Equal(TimeSpan.FromHours(1) + TimeSpan.FromMilliseconds(1), callbacks[1]);
</pre></code>

So yes, `ITimer` does not promise a callback at the exact dueTime of 1ms for the first callback, but the clock is 1 hour off the target. That cannot be considered realistic in my opinion, even if it is technically within specs.

This is what I expect, for the record:

<pre><code>Assert.Equal(2, callbacks.Count);
Assert.Equal(TimeSpan.FromMilliseconds(1), callbacks[0]);
Assert.Equal(TimeSpan.FromHours(1) + TimeSpan.FromMilliseconds(1), callbacks[1]);
</pre></code>
</detail>

danmoseley:
If product code was explicitly asserting the 1ms timer had fired after an hour, it would be buggy. The process could have been suspended for example. Isn't that true for a test as well?

egil:
> If product code was explicitly asserting the 1ms timer had fired after an hour, it would be buggy. The process could have been suspended for example. Isn't that true for a test as well?

Yes, production code should not expect that a timer fires exactly at a specific due time or period and that's not the point I am trying to make. I am trying to illustrate that this PRs implementation of FakeTimeProvider is behaving widely differently from what happens in the production in some cases, like the one above.

The point of my example is that two callbacks that are supposed to be 1 hour between each other should not happen at the same time according to the TimeProviders clock (virtual time or otherwise). I am sure your colleagues responsible for the production `TimeProvider` and `ITimer` would be flooded with issues if that happened.

I have a bunch of tests at work where we record the time during a callback, and use the recorded time later in comparison with other recorded timestamps and `GetUtcNow()`. The current behavior would make writing these tests much more complicated.

Another example (not from work): Suppose we have a 1-hour long-running business process that records its progress every 10 minutes for auditing. In my test of said business process, I may want to advance time to the end of that process, but my 10-minute interval records are all going to show the same timestamp, which will look like they were all recorded at the same time.

Sure, I could call `Advance(TimeSpan.FromMinutes(10)` six times, but why should that be needed and why not have consistent behavior with `Advance(TimeSpan.FromHour(1))` in this particular example?

It is not incorrect if an `ITimer`s callback is invoked exactly at its dueTime/period.

danmoseley:
> I am sure your colleagues responsible for the production TimeProvider and ITimer would be flooded with issues if that happened.

Not trying to be argumentative, and I may be wrong, but I think it could, because of the OS. Eg if the app is suspended. I'm not aware that all OS we run on (or any - but I simply don't know) guarantee to fire the events in the order of the time between the point of firing and the earlier point they could legitimately fire (@kouvel can tell us) I could be wrong, but if not, it seems to be that it's valid behavior, and a test should tolerate it. I agree that it's not necessarily typical.
I defer to @geeknoid though, I was just interested in that point.

Does this mock have enough flexibility that you could derive from it to get what your tests expect? If not perhaps we could add that.

egil:
> Not trying to be argumentative, and I may be wrong, but I think it could, because of the OS....

Likewise. I don't expect I have all the answers either. Apologies if I come off as such. 

> Eg if the app is suspended. I'm not aware that all OS we run on (or any - but I simply don't know) guarantee to fire the events in the order of the time between the point of firing and the earlier point they could legitimately fire (@kouvel can tell us) I could be wrong, but if not, it seems to be that it's valid behavior, and a test should tolerate it. I agree that it's not necessarily typical. I defer to @geeknoid though, I was just interested in that point.

I suspect the answer is that yes, it could be that dueTime callback after 1ms and period callback after 1 hour could happen at more or less the same time in very specific circumstances. But then again, in most unit tests, I am assuming normal OS operations and behavior. 

The weirdness where 1ms callbacks happen at the same time as 1 hour callbacks are left for edge case tests.

> Does this mock have enough flexibility that you could derive from it to get what your tests expect? If not perhaps we could add that.

No, not really. That was my motivation for jumping in with https://github.com/dotnet/extensions/issues/3995 in the first place. That said, it is not clear to me why the functionality currently on main cannot be kept while also adding auto-advance.

egil:
I should add that it is also completely fine if there are special needs in this team which means you want to lean towards a certain behavior. There can be multiple fake implementations of TimeProvider in the .NET ecosystem if need be.

However, if you want to make https://github.com/dotnet/extensions/issues/3996 happen, I'd suggest you ask more widely about this particular point. I am fairly certain that I'm not the only one that would expect that calling `GetUtcNow()` from a timer callback yields the time in which the callback was supposed to be invoked, especially with the promise of a fake TimeProvider that supposedly gives the (test) user control over time in a deterministic way.

geeknoid:
With a real timer, the current time when a calback is executed will *never* be the actual due time. For one thing, the due time is specified as a TimeSpan, so you don't actually know what the actual expected due time will actually be, and by the time you read the clock, the time will have moved forward. So any expectation in code of a specific exact due time is broken by definition.

The fake time provider lets you precisely control the behavior of how time proceeds and it intends to mirror the behavior of a real time for the sake of application behavior testing. In a real world scenario, if your process is suspended, when the timer system eventually regains control it will sample the current time and do catch up work based on the current time. This might involve completing N callbacks in quick succession even though they had vastly different due times. 

The fake timer works the same way. If time jumps by a large amount, the timer logic does catchup work that is needs, including triggering multiple callbacks in quick succession if needed. The simulation of the timer's behavior is in fact fairly precise. The difference between the real timer and the fake timer is that time moves forward programmatically rather than via hardware. It's up to the test code to decide how fast to move the clock forward, and the timer logic deals with this.

The change in behavior here has nothing to do with auto-advancing time. I started only wanting to introduce auto-advancing time in this code, but then I found many race conditions through the code, making the timer abstraction unusable in a multi-threaded environment. So I proceeded to fix these problems. As part of fixing these problems, I studied the code that was fiddling with the wake time of the callbacks and decided explicitly that this was not the right semantics.

Again: the current implementation behaves like the real timer does, except that the advancement of time is controlled through software rather than hardware, and thus tests have full control over how this is used and the resulting effects.

egil:
@geeknoid, I appreciate you taking the time to write out a long explanation. 

For the record, I am not arguing that the fake behaves in a wrong way, my argument is that it behaves in an unexpected way, and I expect users will fall into the trap of assuming the behavior is as I suggest, and waste much time trying to debug. The users are explicitly using a fake time provider, not a real one. The alternative I am proposing would still be a correct implementation of an `ITimer`, I believe.

If you decide to keep the implementation as it is now, add appropriate code docs (e.g. `<remarks>` ) to both `SetUtcNow` and `Advance` where you explain in detail the behavior users should expect (i.e. `GetUtcNow()` does not always return the time a timer callback is supposed to be invoked). This implicit knowledge needs to be as visible as possible to save developers like me from wasting time.

> Does this mock have enough flexibility that you could derive from it to get what your tests expect? If not perhaps we could add that.

@danmoseley, let me update my answer to this. Since @geeknoid says there is nothing preventing the behavior I am advocating for, it would be great if it was something that you could actively toggle like the auto-increment, either directly on the type or via a sub-type that overrides behavior. Otherwise, I will most likely keep my own version alive and keep using that.

PR #4106, opened 2023-06-23
Fix bug in MetricCollector.WaitForMeasurementsAsync

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4106)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
@@ -253,7 +253,11 @@ public sealed class MetricCollector<T> : IDisposable
     [SuppressMessage("Resilience", "R9A061:The async method doesn't support cancellation", Justification = "Not relevant in this case")]
     public async Task WaitForMeasurementsAsync(int minCount, TimeSpan timeout)
     {
-        using var cancellationTokenSource = new CancellationTokenSource(timeout);
+#if NET8_0_OR_GREATER
+        using var cancellationTokenSource = new CancellationTokenSource(timeout, _timeProvider);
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -253,7 +253,11 @@ public Task WaitForMeasurementsAsync(int minCount, CancellationToken cancellatio
     [SuppressMessage("Resilience", "R9A061:The async method doesn't support cancellation", Justification = "Not relevant in this case")]
     public async Task WaitForMeasurementsAsync(int minCount, TimeSpan timeout)
     {
-        using var cancellationTokenSource = new CancellationTokenSource(timeout);
+#if NET8_0_OR_GREATER
+        using var cancellationTokenSource = new CancellationTokenSource(timeout, _timeProvider);
+#else
+        using var cancellationTokenSource = _timeProvider.CreateCancellationTokenSource(timeout);
+#endif
```
tarekgh, line 260: I am wondering can we just use the extension method to avoid the ifdef? or Microsoft.Bcl.TimeProvider is not referenced at all when building for .NET 8.0 and up?

tarekgh (approved):

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
geeknoid: Correct, the symbol doesn't exist when building for .NET 8.

PR #4100, opened 2023-06-22
Improve HttpClient resilience benchmarks

martintmk:
Addiing a new benchmark that compares no-resilience, standard resilience and standard hedging resilience.

```
BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.1848), VM=Hyper-V
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.100-preview.6.23309.5
  [Host] : .NET 8.0.0 (8.0.23.30704), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10

|                    Method |       Mean |    Error |   StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|-------------------------- |-----------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
|             DefaultClient |   236.4 ns |  2.24 ns |  3.28 ns |  1.00 |    0.00 | 0.0191 |     488 B |        1.00 |
| StandardResilienceHandler | 4,101.4 ns | 21.96 ns | 32.19 ns | 17.35 |    0.22 | 0.1602 |    4152 B |        8.51 |
|    StandardHedgingHandler | 8,060.2 ns | 58.09 ns | 85.15 ns | 34.10 |    0.35 | 0.2289 |    5880 B |       12.05 |
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4100)

PR #4084, opened 2023-06-16
Adopt Polly V8 in Microsoft.Extensions.Resilience

martintmk:
This PR adopts the new [Polly.Extensions](https://www.nuget.org/packages/Polly.Extensions/8.0.0-alpha.1) package. 

All the features introduced in `Microsoft.Extensions.Resilience` are now built into Polly V8. This makes most of the code in this library redundant. 

The library will now only expose resilience enrichment and some extensions for `ResilienceContext`.

``` csharp
var services = new ServiceCollection();
services.AddResilienceEnrichment();
```

API changes:

Old API
``` csharp
services
    .AddResiliencePipeline<string>("my-pipeline")
    .AddRetryPolicy("my-retry", options =>
    {
        options.RetryCount = 3;
        options.BaseDelay = TimeSpan.FromSeconds(5);
    })
    .AddTimeoutPolicy("my-timeout", options =>
    {
        options.TimeoutInterval = TimeSpan.FromSeconds(30);
    });
```

New Polly V8 API:
``` csharp
new ServiceCollection()
    .AddResilienceStrategy("my-strategy", builder => builder
    .AddRetry(new RetryStrategyOptions
    {
        StrategyName = "my-retry",
        RetryCount = 3,
        BaseDelay = TimeSpan.FromSeconds(1)
    })
    .AddTimeout(new TimeoutStrategyOptions
    {
        Timeout = TimeSpan.FromSeconds(30),
        StrategyName = "my-timeout"
    }));
```

The PR won't be mergeable until we adopt V8 in `Microsoft.Extensions.Http.Resilience`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4084)

PR #4108, opened 2023-06-23
Adopt Polly V8 in  Microsoft.Extensions.Http.Resilience

martintmk:
Adopting the V8 version for HTTP clients. 

Changes:
- Adopting V8 API and dropping some internals that are no longer needed
- Reorganizing some internals that are shared into `/Internal` folder.
- Http-based options based now derive from Polly V8 options
- Cleanup, simplification of unit tests
- Replacing the token `Pipeline` with `Strategy` to align with Polly V8
- Making some pub-internals internal again

The tests are still WIP, I'll work on them until we have 100% code coverage. I'll run performance tests once #4100 is merged.

Unblocks #4084 

API Changes:

**Standard Handlers**

Mostly cosmetical changes, high-level API stays the same. V8 now also supports asynchronous predicates with richer arguments. 

``` csharp
// Old
services
    .AddHttpClient("my-client")
    .AddStandardResilienceHandler()
    .Configure(options =>
    {
        options.TotalRequestTimeoutOptions.TimeoutInterval = TimeSpan.FromSeconds(40);
        options.RetryOptions.ShouldHandleException = e => e is HttpRequestException;
        options.RetryOptions.ShouldHandleResultAsError = r => !r.IsSuccessStatusCode;
    });

// New
services
    .AddHttpClient("my-client")
    .AddStandardResilienceHandler()
    .Configure(options =>
    {
        options.TotalRequestTimeoutOptions.Timeout = TimeSpan.FromSeconds(40);
        options.RetryOptions.ShouldHandle = args => args switch
        {
            { Exception: HttpRequestException } => PredicateResult.True,
            { Result: { IsSuccessStatusCode: false } } => PredicateResult.True,
            _ => PredicateResult.False,
        };
    });
```

**Resilience Handler**

Conceptually the same, but now the strategy is configured using a callback.

``` csharp
// Old
services
    .AddHttpClient("my-client")
    .AddResilienceHandler("my-handler")
    .AddRetryPolicy("my-retry", options => options.RetryCount = 3)
    .AddTimeoutPolicy("my-timeout", options => options.TimeoutInterval = TimeSpan.FromSeconds(4));

// New
services
    .AddHttpClient("my-client")
    .AddResilienceHandler("my-handler", builder => builder
      .AddRetry(new HttpRetryStrategyOptions
      {
          StrategyName = "my-retry",
          RetryCount = 3,
      })
      .AddTimeout(new HttpTimeoutStrategyOptions
      {
          StrategyName = "my-timeout",
          Timeout = TimeSpan.FromSeconds(4),
      }));
```

**Benchmarks:**

Before:

|                    Method |       Mean |    Error |   StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|-------------------------- |-----------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
|             DefaultClient |   236.4 ns |  2.24 ns |  3.28 ns |  1.00 |    0.00 | 0.0191 |     488 B |        1.00 |
| StandardResilienceHandler | 4,101.4 ns | 21.96 ns | 32.19 ns | 17.35 |    0.22 | 0.1602 |    4152 B |        8.51 |
|    StandardHedgingHandler | 8,060.2 ns | 58.09 ns | 85.15 ns | 34.10 |    0.35 | 0.2289 |    5880 B |       12.05 |

After:

|                    Method |       Mean |     Error |    StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|-------------------------- |-----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
|             DefaultClient |   237.4 ns |   1.56 ns |   2.69 ns |  1.00 |    0.00 | 0.0191 |     488 B |        1.00 |
| StandardResilienceHandler | 2,774.0 ns |  44.61 ns |  74.53 ns | 11.69 |    0.32 | 0.0229 |     648 B |        1.33 |
|    StandardHedgingHandler | 7,132.6 ns | 114.23 ns | 197.03 ns | 30.05 |    1.02 | 0.0458 |    1336 B |        2.74 |

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4108)


martintmk:
cc @martincostello

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/Validators/HttpStandardResilienceOptionsValidator.cs
```diff
@@ -2,11 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Microsoft.Extensions.Options;
-using Microsoft.Extensions.Options.Validation;
 
 namespace Microsoft.Extensions.Http.Resilience.Internal.Validators;
 
-[OptionsValidator]
-internal sealed partial class HttpStandardResilienceOptionsValidator : IValidateOptions<HttpStandardResilienceOptions>
+internal sealed class HttpStandardResilienceOptionsValidator : IValidateOptions<HttpStandardResilienceOptions>
```
martintmk, line 8: @geeknoid I had to disable this validator because it materializes even the internal validation attributes we have in Polly and the PR won't compile. 

I believe the generator should just skip those.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.Resilience.cs
```diff
@@ -4,82 +4,164 @@
...
+            .AddExceptionSummarizer(b => b.AddHttpProvider())
+            .ConfigureFailureResultContext<HttpResponseMessage>((response) =>
+            {
+                if (response != null)
+                {
+                    var statusCodeName = ((int)response.StatusCode).ToInvariantString();
+                    if (string.IsNullOrEmpty(statusCodeName) || char.IsDigit(statusCodeName[0]))
+                    {
+                        statusCodeName = TelemetryConstants.Unknown;
+                    }
```
adamnova, line 152: What does this do? I don't understand. Convert enum to int, then to string and check if its not null or empty? Which I don't think it ever can be or if it's a digit? Can int as string ever not be a digit? I am very confused. Maybe I am missing something.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpRequestMessageExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    /// Gets the <see cref="ResilienceContext"/> from the request message.
+    /// </summary>
+    /// <param name="requestMessage">The request.</param>
+    /// <returns>An instance of <see cref="ResilienceContext"/> or <see langword="null"/>.</returns>
+    public static ResilienceContext? GetResilienceContext(this HttpRequestMessage requestMessage)
+    {
+        _ = Throw.IfNull(requestMessage);
+
+#if !NET8_0_OR_GREATER
```
adamnova, line 30: I think it would be nicer to invert this condition

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpRequestMessageExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    /// Sets the <see cref="ResilienceContext"/> on the request message.
+    /// </summary>
+    /// <param name="requestMessage">The request.</param>
+    /// <param name="resilienceContext">An instance of <see cref="ResilienceContext"/>.</param>
+    public static void SetResilienceContext(this HttpRequestMessage requestMessage, ResilienceContext resilienceContext)
+    {
+        _ = Throw.IfNull(requestMessage);
+        _ = Throw.IfNull(resilienceContext);
+#if !NET8_0_OR_GREATER
```
adamnova, line 53: I think it would be nicer to invert this condition

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/ValidationHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+                    // return zero delay, so no waiting
+                    return new ValueTask<TimeSpan>(TimeSpan.Zero);
+                }
+            })
+            .Build()
+            .Execute(() => { }); // this executes all retries and we aggregate the delays immediately
+
+            return aggregatedDelay;
+        }
+        catch (OverflowException)
```
martincostello, line 43: What scenario causes the overflow? Can it be avoided?

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/ResilienceHandler.cs
```diff
@@ -5,54 +5,86 @@
...
+        _strategyProvider = strategyProvider;
     }
 
     /// <inheritdoc/>
     protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
     {
+        var strategy = _strategyProvider(request);
         var created = false;
-        if (request.GetPolicyExecutionContext() is not Context context)
+        if (request.GetResilienceContext() is not ResilienceContext context)
```
adamnova, line 31: I think it would be better to just check if its null here, instead of hiding null check behind type check. Same below on line 38 `if (request.GetRequestMetadata() is RequestMetadata requestMetadata)`.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/ValidationHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+                {
+                    // the delay hint is calculated for this attempt by the retry strategy
+                    aggregatedDelay += args.Arguments.DelayHint;
+
+                    // return zero delay, so no waiting
+                    return new ValueTask<TimeSpan>(TimeSpan.Zero);
+                }
+            })
+            .Build()
+            .Execute(() => { }); // this executes all retries and we aggregate the delays immediately
```
martincostello, line 39: Would making this `static` do anything meaningful or would it just be noise?

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </remarks>
+    public HttpRetryStrategyOptions()
+    {
+        ShouldHandle = args => new ValueTask<bool>(HttpClientResiliencePredicates.IsTransientHttpOutcome(args.Outcome));
+        BackoffType = RetryBackoffType.ExponentialWithJitter;
+        ShouldRetryAfterHeader = true;
+    }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to retry after header.
```
martincostello, line 34: This sentence doesn't really make any sense. Did you mean _"...whether to use the Retry-After header"_?

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </summary>
+    /// <value>
+    /// Defaults to <see langword="true"/>.
+    /// </value>
+    /// <remarks>
+    /// If the property is set to <c>true</c>, then the generator will resolve the delay
+    /// based on the RetryAfter header rules, otherwise it will return <see cref="RetryDelayArguments.DelayHint"/>
```
martincostello, line 41:
```suggestion
    /// based on the HTTP <c>Retry-After</c> header, otherwise it will return <see cref="RetryDelayArguments.DelayHint"/>
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.Resilience.cs
```diff
@@ -4,82 +4,164 @@
...
 
-        return new HttpResiliencePipelineBuilder(pipelineBuilder);
+        _ = services
+            .AddExceptionSummarizer(b => b.AddHttpProvider())
+            .ConfigureFailureResultContext<HttpResponseMessage>((response) =>
+            {
+                if (response != null)
+                {
+                    var statusCodeName = ((int)response.StatusCode).ToInvariantString();
+                    if (string.IsNullOrEmpty(statusCodeName) || char.IsDigit(statusCodeName[0]))
```
martincostello, line 149: I'm not fully following this. Won't turning an HTTP status integer to a string always result in the first character being a digit and thus the second condition always evaluate to `true`? Or is it to catch the status somehow being a negative number? A comment explaining might be useful.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/ByAuthorityStrategyKeyProvider.cs
```diff
@@ -10,40 +10,40 @@
...
-        pipelineKey = _redactor.Redact(pipelineKey);
+        strategyKey = url.GetLeftPart(UriPartial.Authority);
+        strategyKey = _redactor.Redact(strategyKey);
 
-        if (string.IsNullOrEmpty(pipelineKey))
+        if (string.IsNullOrEmpty(strategyKey))
         {
             Throw.InvalidOperationException(
-                $"The redacted pipeline is an empty string and cannot be used for pipeline selection. Is redaction correctly configured?");
+                $"The redacted strategy key is an empty string and cannot be used for the strategy selection. Is redaction correctly configured?");
```
martincostello, line 41: There's no interpolation happening.

```suggestion
                "The redacted strategy key is an empty string and cannot be used for the strategy selection. Is redaction correctly configured?");
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Routing/Internal/RoutingStrategy.cs
```diff
@@ -41,16 +34,19 @@ public RoutingPolicy(string pipelineName, IRequestRoutingStrategyFactory factory
             Throw.InvalidOperationException("The routing strategy did not provide any route URL on the first attempt.");
         }
 
-        _routingStrategySetter(context, strategy);
+        if (!context.Properties.TryGetValue(ResilienceKeys.RequestMessage, out var request))
+        {
+            Throw.InvalidOperationException("The request message was not found in the resilience context.");
```
martincostello, line 39: Maybe mention it's HTTP?
```suggestion
            Throw.InvalidOperationException("The HTTP request message was not found in the resilience context.");
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.Resilience.cs
martintmk: It's indeed strange. I have just ported this code but it makes little sense.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/ValidationHelper.cs
martintmk: Too many retries, high base delay. This piece is only ever executed on startup, so it's not worth doing any complications anyway.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/ValidationHelper.cs
martintmk: sure, why not.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.Resilience.cs
martintmk: Indeed, does not make any sense. (just ported the code)

I made changes:

- `failureReason`: status code as number
- `additionalInformation`: status code as enum string

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.Resilience.cs
martintmk: fixed ... hopefully

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RetryAfterHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Parses Retry-After value from the relevant HTTP response header.
+    /// If not found then it will return <see cref="TimeSpan.Zero" />.
+    /// </summary>
+    /// <see href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After" />.
+    internal static bool TryParse(HttpResponseMessage response, TimeProvider timeProvider, out TimeSpan retryAfter)
+    {
+        retryAfter = response.Headers.RetryAfter switch
+        {
+            { Date: { } date } => date - timeProvider.GetUtcNow(),
```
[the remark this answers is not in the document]
martintmk: Indeed if the result evaluates to a point in time that already happened the retry attempt should happen immediately.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/Validators/HttpStandardResilienceOptionsValidator.cs
geeknoid: Does the option type in this package derive from the option type in the Polly package?

@tarekgh This looks like a case we hadn't seen before and since you know own the option validation generator, well, here you go :-) I don't know if there's a way to compose this together such that the derived option type gets full validation even for the properties inherited from the base class.

martintmk: > Does the option type in this package derive from the option type in the Polly package?

Yes, this is the internal attribute we are using:

https://github.com/App-vNext/Polly/blob/a9bf3e95a37ff9c98ca02c8ca395f91645ffe5d8/src/Polly.Core/Timeout/TimeoutStrategyOptions.cs#L20

geeknoid: I don't think this is going to work without you publishing something more in the polly code. Otherwise, there is no way for the validation code generated in this assembly to validate the state in the base class.

If you make TimeoutAttribute public, then this will just work as the generated code will have access to the atribute definition.

In order to keep the attribute hidden, then you'd need to make public instead the generated options validation type from the Polly library so that it could be called from this assembly.

Perhaps there would be a way for Polly to register an option validator for its part of the state, and then this assembly only validates its part of the state?

In any case, @tarekgh, we should fix the option generator to ignore attributes that aren't reachable from the generated code.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/Validators/HttpStandardHedgingResilienceOptionsValidator.cs
```diff
@@ -2,11 +2,10 @@
...
 using Microsoft.Extensions.Options;
-using Microsoft.Extensions.Options.Validation;
 
 namespace Microsoft.Extensions.Http.Resilience.Internal.Validators;
 
-[OptionsValidator]
-internal sealed partial class HttpStandardHedgingResilienceOptionsValidator : IValidateOptions<HttpStandardHedgingResilienceOptions>
+internal sealed class HttpStandardHedgingResilienceOptionsValidator : IValidateOptions<HttpStandardHedgingResilienceOptions>
 {
+    public ValidateOptionsResult Validate(string? name, HttpStandardHedgingResilienceOptions options) => ValidateOptionsResult.Success;
```
geeknoid, line 10: Why this change? The prior pattern would achieve the same effect.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/RequestMessageSnapshotStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+        ResilienceContext context,
+        TState state)
+    {
+        var request = context.Properties.GetValue(ResilienceKeys.RequestMessage, null!);
+
+        using var snapshot = _requestCloner.CreateSnapshot(request);
+
+        context.Properties.Set(ResilienceKeys.RequestSnapshot, snapshot);
+
+        return await callback(context, state).ConfigureAwait(context.ContinueOnCapturedContext);
```
geeknoid, line 35: You can probably remove the "async" in the method declaration and remove the await/ConfigureAwait here.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/RequestMessageSnapshotStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        _requestCloner = requestCloner;
+    }
+
+    protected override async ValueTask<Outcome<TResult>> ExecuteCoreAsync<TResult, TState>(
+        Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+        ResilienceContext context,
+        TState state)
+    {
+        var request = context.Properties.GetValue(ResilienceKeys.RequestMessage, null!);
```
geeknoid, line 29: That ! makes me nervous. Why do we need to break nullability rules here?

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/HedgingEndpointOptions.cs
```diff
@@ -3,13 +3,14 @@
...
+using Microsoft.Extensions.Http.Resilience.Internal.Validators;
 using Microsoft.Extensions.Options.Validation;
-using Microsoft.Extensions.Resilience.Options;
+using Polly.Timeout;
 
 namespace Microsoft.Extensions.Http.Resilience;
 
 /// <summary>
-/// Options for resilient pipeline of policies assigned to a particular endpoint. It is using three chained layers in this order (from the outermost to the innermost):
+/// Options for the pipeline of resilience strategies assigned to a particular endpoint. It is using three chained layers in this order (from the outermost to the innermost):
```
geeknoid, line 13: Only one sentence in the summary, the rest can go in remarks.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/HedgingEndpointOptions.cs
```diff
@@ -20,33 +21,40 @@ public class HedgingEndpointOptions
...
+    /// By default it is initialized with a unique instance of <see cref="HttpTimeoutStrategyOptions"/>
+    /// using a custom <see cref="TimeoutStrategyOptions.Timeout"/> of 10 seconds.
     /// </remarks>
     [Required]
     [ValidateObjectMembers]
-    public HttpTimeoutPolicyOptions TimeoutOptions { get; set; } = new()
+    public HttpTimeoutStrategyOptions TimeoutOptions { get; set; } = new()
     {
-        TimeoutInterval = _timeoutInterval,
+        Timeout = _timeoutInterval,
```
geeknoid, line 57:
```suggestion
        Timeout = TimeSpan.FromSeconds(10),
```

which means you don't need _timeoutInterval field, saving 8 bytes :-)

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/Validators/HttpStandardResilienceOptionsValidator.cs
martintmk: Ideally, it would be nice to get rid of it and use built-in attributes.

I need to validate that timespan is in range of -1 milliseconds to max. Also, zero or default value must be excluded.

tarekgh: Could you provide isolated reproduce for this problem so I can look at it?

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RetryAfterHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Net.Http;
+
+namespace Microsoft.Extensions.Http.Resilience.Internal;
+
+internal static class RetryAfterHelper
+{
+    /// <summary>
+    /// Parses Retry-After value from the relevant HTTP response header.
+    /// If not found then it will return <see cref="TimeSpan.Zero" />.
```
geeknoid, line 13: One sentence in the summary, rest in remarks please.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/ByAuthorityStrategyKeyProvider.cs
```diff
@@ -10,40 +10,40 @@
...
 namespace Microsoft.Extensions.Http.Resilience.Internal;
 
-internal sealed class ByAuthorityPipelineKeyProvider : IPipelineKeyProvider
+internal sealed class ByAuthorityStrategyKeyProvider : IStrategyKeyProvider
 {
     private readonly Redactor _redactor;
     private readonly ConcurrentDictionary<(string scheme, string host, int port), string> _cache = new();
 
-    public ByAuthorityPipelineKeyProvider(Redactor redactor, DataClassification _)
+    public ByAuthorityStrategyKeyProvider(Redactor redactor, DataClassification _)
```
geeknoid, line 18: Who allocates this redactor?

If there is already a redactor, then this constructor doesn't need a DataClassification instance, so can we just remove it?

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.StandardResilience.cs
```diff
@@ -37,16 +36,16 @@ public static IHttpStandardResiliencePipelineBuilder AddStandardResilienceHandle
     }
 
     /// <summary>
-    /// Adds a <see cref="PolicyHttpMessageHandler" /> that uses a standard resilience pipeline with default options to send the requests and handle any transient errors.
-    /// The pipeline combines multiple policies that are configured based on HTTP-specific <see cref="HttpStandardResilienceOptions"/> options with recommended defaults.
+    /// Adds a standard resilience handler that uses a multiple resilience strategies with default options to send the requests and handle any transient errors.
+    /// The resilience strategy combines multiple strategies that are configured based on HTTP-specific <see cref="HttpStandardResilienceOptions"/> options with recommended defaults.
```
geeknoid, line 40: Only one sentence in the summary please. Other functions in this file also have this problem.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpStandardResilienceOptions.cs
```diff
@@ -28,54 +29,68 @@ public class HttpStandardResilienceOptions
...
+    /// By default it is initialized with a unique instance of <see cref="HttpTimeoutStrategyOptions"/>
+    /// using custom <see cref="TimeoutStrategyOptions.Timeout"/> of 10 seconds.
     /// </remarks>
     [Required]
     [ValidateObjectMembers]
-    public HttpTimeoutPolicyOptions AttemptTimeoutOptions { get; set; } = new()
+    public HttpTimeoutStrategyOptions AttemptTimeoutOptions { get; set; } = new()
     {
-        TimeoutInterval = _attemptTimeoutInterval,
+        Timeout = _attemptTimeoutInterval,
```
geeknoid, line 93:
```suggestion
        Timeout = TimeSpan.FromSeconds(10),
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/StrategyKeySelector.cs
```diff
@@ -6,12 +6,12 @@
...
-/// <param name="requestMessage">The request message that the pipeline key is extracted from.</param>
-/// <returns>A pipeline key.</returns>
+/// <param name="requestMessage">The request message that the strategy key is extracted from.</param>
+/// <returns>A strategy key.</returns>
 /// <remarks>
-/// The pipeline key is used by metrics and telemetry. Make sure it does not contain any sensitive data.
+/// The strategy key is used by metrics and telemetry. Make sure it does not contain any sensitive data.
 /// </remarks>
-public delegate string PipelineKeySelector(HttpRequestMessage requestMessage);
+public delegate string StrategyKeySelector(HttpRequestMessage requestMessage);
```
geeknoid, line 16: Can we get rid of this type in favor of just using Func<>?

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/Validators/HttpStandardResilienceOptionsValidator.cs
geeknoid: Here a repro: https://github.com/dotnet/extensions/commit/1bdd9251a72090b321cd742c08466de127d8413d

@martintmk I think your best bet for now is to make the attribute public in Polly and then it'll work correctly.

martintmk: @geeknoid Thanks Martin for the repro!

I have actually took a second look at our custom attributes and realized we can actually drop them:
https://github.com/App-vNext/Polly/pull/1351

Still, the generator behavior should properly handle internal attributes.

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptions.cs
martintmk: Done.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/RequestMessageSnapshotStrategy.cs
martintmk: It needs to stay async as we need to dispose the snapshot after the execution.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/ResilienceHandler.cs
martintmk: The construct above also defines the `context` variable, saving additional line.

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptions.cs
martintmk: True, fixed.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/Validators/HttpStandardHedgingResilienceOptionsValidator.cs
martintmk: I'll restore the validation when alpha.4 is ready.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/RequestMessageSnapshotStrategy.cs
martintmk: True, checking for null and throwing exception.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/ByAuthorityStrategyKeyProvider.cs
martintmk: True, removed.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpRequestMessageExtensions.cs
martintmk: Done.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpClientBuilderExtensions.StandardResilience.cs
martintmk: Done.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RetryAfterHelper.cs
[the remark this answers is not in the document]
martintmk: Fixed.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/Validators/HttpStandardHedgingResilienceOptionsValidator.cs
martintmk: Fixed.

martintmk:
@RussKie 

The code coverage reports is really strange. I am pretty sure this call is covered (many tests that the `AddResilienceHandler` API).

image

https://dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab


Any idea on how to fix this?

src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/Internal/Log.cs
```diff
@@ -1,11 +1,13 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Telemetry.Logging;
 
 namespace Microsoft.Extensions.Http.Resilience.FaultInjection.Internal;
 
+[ExcludeFromCodeCoverage]
```
geeknoid, line 10: Shouldn't have this, the logging code should be tested too.

src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/PolicyContextExtensions.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.Extensions.Http.Resilience.FaultInjection;
 
 /// <summary>
-/// Provides extension methods for <see cref="Polly.Context"/>.
+/// Provides extension methods for <see cref="Context"/>.
 /// </summary>
 [Experimental]
 public static class PolicyContextExtensions
```
geeknoid, line 15: Rename to ContextExtensions?

geeknoid (commented):
Looks very nice. I still have ~20 files to go...

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHandlerContext.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using Microsoft.Extensions.Http.Resilience.Internal;
+using Polly.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Http.Resilience;
+
+/// <summary>
+/// The context used when building a resilience strategy HTTP handler.
+/// </summary>
+public class ResilienceHandlerContext
```
geeknoid, line 13: Since there isn't a public constructor and so the type can't be extended, make it sealed.

geeknoid (commented):
Could RequestRoutingStrategyFactory.ReturnRoutingStrategy be removed and just replaced with the TryReset function to implement the IResettable interface?

And related, the name CreateRoutingStrategy is misleading, that should be GetRoutingStrategy.

src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/Internal/Log.cs
martincostello: Doesn't that mean adding exhaustive test coverage for everything the generated logging code does, which is effectively external and an implementation detail?

Seems wasteful to me to 100% cover someone else's code just because it's source-generated as long as it's being _exercised_ by a test. I don't think you'd need it if the requirement wasn't 100% coverage as you could just tweak the threshold, but if it's all or nothing then this seems to be the sensible alternative to exhaustively re-testing the generated code to do logging.

By the same logic I exclude Request Delegate Generator code, otherwise you get fun behaviour like this: https://github.com/dotnet/aspnetcore/issues/48376. I certainly wouldn't spend my time testing ASP.NET Code's error handling scenarios for my user-supplied delegate chasing a coverage metric.

martintmk: What Martin says, we had this problem in Polly too. We had to artificially add test cases for code that was auto-generated. 

https://github.com/App-vNext/Polly/blob/faaf2ee95b9b92386f8e091277604240f9ab05c0/test/Polly.Extensions.Tests/Telemetry/ResilienceTelemetryDiagnosticSourceTests.cs#L173

Not a big fan of that.

martintmk:
> Could RequestRoutingStrategyFactory.ReturnRoutingStrategy be removed and just replaced with the TryReset function to implement the IResettable interface?
> 
> And related, the name CreateRoutingStrategy is misleading, that should be GetRoutingStrategy.

Is it OK to do it in follow-up? I want to turn it to internal abstract class too.

src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/Internal/Log.cs
geeknoid: @martincostello That's not what this implies.

This is about ensuring the telemetry produced by this component matches expectations. Within the tests, you use a FakeLogger that captures the telemetry and the ensure that the right thing is coming out of the logging system at the time you expect it to.

As a general rule, the only things that we consider acceptable to exclude from code coverage involve some legacy design patterns that preclude it and some racy code where its impossible to ensure all paths are visited in a given test run. Both are very rare things.

martincostello: But isn't that code custom telemetry logging code called _through_ the compiler generated code? There's no "real" code here in this class, it's just a partial method stub:

https://github.com/dotnet/extensions/blob/b31e661e4165d3607ad116366c5a593b0e988289/src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/Internal/Log.cs#L13-L23

It not being covered doesn't preclude the assertions in the fake logger.

martincostello: Unless of course `[LogMethod]` doesn't work the same as `[LoggerMessage]` which I initially thought this was.

Looking at some code of my own I have this:

```csharp
[ExcludeFromCodeCoverage]
private static partial class Log
{
    [LoggerMessage(
       EventId = 1,
       Level = LogLevel.Information,
       Message = "Deployments have been frozen by {UserName}.")]
    public static partial void DeploymentsFrozen(ILogger logger, string? userName);
}
```

Which gets turned into this:

```csharp
partial class Log
{
    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "8.0.8.28008")]
    private static readonly global::System.Action<global::Microsoft.Extensions.Logging.ILogger, global::System.String?, global::System.Exception?> __DeploymentsFrozenCallback =
        global::Microsoft.Extensions.Logging.LoggerMessage.Define<global::System.String?>(global::Microsoft.Extensions.Logging.LogLevel.Information, new global::Microsoft.Extensions.Logging.EventId(1, nameof(DeploymentsFrozen)), "Deployments have been frozen by {UserName}.", new global::Microsoft.Extensions.Logging.LogDefineOptions() { SkipEnabledCheck = true }); 

    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "8.0.8.28008")]
    public static partial void DeploymentsFrozen(global::Microsoft.Extensions.Logging.ILogger logger, global::System.String? userName)
    {
        if (logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Information))
        {
            __DeploymentsFrozenCallback(logger, userName, null);
        }
    }
}
```

In that case it's not useful to add tests to test the coverage of the branching for `IsEnabled()` on all the logging we have, hence the coverage suppression in my case.

If it's a different use case here because that just looks _similar_ to the above then disregard my comment 😃

martintmk: In Polly we are correctly checking that message is logged:

https://github.com/App-vNext/Polly/blob/faaf2ee95b9b92386f8e091277604240f9ab05c0/test/Polly.Extensions.Tests/Telemetry/ResilienceTelemetryDiagnosticSourceTests.cs#L155

But for some strange reason, it want's us to cover the code-generated code too.  For example, the auto-generated struct that backs the entry with all it's properties and methods.

RussKie:
> @RussKie Igor Velikorossov FTE
> 
> The code coverage reports is really strange. I am pretty sure this call is covered (many tests that the `AddResilienceHandler` API).
> 
> image
> 
> [dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab](https://dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab)
> 
> Any idea on how to fix this?
> 
> edit: also the threshold check is strange. The code coverage reports 99.8% coverage while the check fails complaining the coverage is <99%.

@jakubch1 are you able to help here?

src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/PolicyContextExtensions.cs
martintmk: Done.

jakubch1:
> > @RussKie Igor Velikorossov FTE
> > The code coverage reports is really strange. I am pretty sure this call is covered (many tests that the `AddResilienceHandler` API).
> > image
> > [dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab](https://dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab)
> > Any idea on how to fix this?
> > edit: also the threshold check is strange. The code coverage reports 99.8% coverage while the check fails complaining the coverage is <99%.
> 
> @jakubch1 are you able to help here?

Issue here can be that Azure DevOps are using old version of Report Generator inside Publish Code Coverage task. I recommend upgrade to V2 version of the task: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-code-coverage-results-v2?view=azure-pipelines

src/Libraries/Microsoft.Extensions.Http.Resilience/FaultInjection/Internal/Log.cs
martintmk: Created a bug #4124. In the meantime, I had to manually decrease code-coverage to 98. Once that bug is fixed, we can put it back to 100.

martintmk:
> > > @RussKie Igor Velikorossov FTE
> > > The code coverage reports is really strange. I am pretty sure this call is covered (many tests that the `AddResilienceHandler` API).
> > > image
> > > [dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab](https://dev.azure.com/dnceng-public/public/_build/results?buildId=320898&view=codecoverage-tab)
> > > Any idea on how to fix this?
> > > edit: also the threshold check is strange. The code coverage reports 99.8% coverage while the check fails complaining the coverage is <99%.
> > 
> > 
> > @jakubch1 are you able to help here?
> 
> Issue here can be that Azure DevOps are using old version of Report Generator inside Publish Code Coverage task. I recommend upgrade to V2 version of the task: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-code-coverage-results-v2?view=azure-pipelines

Created an issue #4124.

geeknoid (approved):

PR #4110, opened 2023-06-23
Introduce new extended logging model.

geeknoid:
Introduce new extended logging model.

- Replace the OpenTelemetry-specific logger design with a
LoggerFactory-based approach independent of OpenTelemetry. This
delivers enrichment and redaction to all ILogger users.

The basic idea is that the code generator will no longer be responsible
for doing redaction. Instead, the generated code will accumulate
normal properties in one collection and will accumulate classified
properties in a different collection.

Within the Logger type, the classified properties are run through
redaction and added to the normal property list. This list is then
used to enrich into. And then the final thing is given to the set of
currently registered ILogger instances in the system. So all these
loggers get redacted and enriched state.

This separates static vs. dynamic log enrichment. The idea is to 
reduce overhead for stuff that never changes. Thus, we have the new
IStaticLogEnricher type.

NOTES:

- This PR doesn't currently include an updated logging code generator
that will take advantage of this new functionality.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4110)


src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -3,177 +3,70 @@
...
-            return;
-        }
-
-        LogMethodHelper propertyBag;
-        LogMethodHelper? rentedHelper = null;
-
         try
         {
-            if (state is LogMethodHelper helper && _provider.CanUsePropertyBagPool)
+            if (typeof(TState) == typeof(LoggerMessageState))
```
xakep139, line 51: Can we do a pattern matching here? The structure will look much cleaner

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -219,26 +112,115 @@ private static void GetInnerExceptionTrace(Exception exception, StringBuilder st
...
+        {
+            state.Add(ExceptionStackTrace, GetExceptionStackTrace(exception, _maxStackTraceLength));
+        }
+
+        _nextLogger.Log(logLevel, eventId, state, exception, formatter);
+    }
+
+    private void R9Path(LogLevel logLevel, EventId eventId, LogMethodHelper? state, Exception? exception, Func<LogMethodHelper, Exception?, string> formatter)
+    {
+        if (!IsEnabled(logLevel) || state == null)
```
xakep139, line 157: Even if the `state` is `null`, shouldn't we pass logging to the `_nextLogger`?

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -219,26 +112,115 @@ private static void GetInnerExceptionTrace(Exception exception, StringBuilder st
...
+
+        // one last dedicated bit of enrichment
+        if (exception != null && _includeStackTraces)
+        {
+            lms.Add(ExceptionStackTrace, GetExceptionStackTrace(exception, _maxStackTraceLength));
+        }
+
+        lms.Formatter = formatter;
+        lms.State = state;
+        _nextLogger.Log(logLevel, eventId, lms, exception, (S, e) =>
```
xakep139, line 217: I guess lambda can be `static`

geeknoid:
Second round:

* This now introduced static log enrichment to reduce overhead of copying the same stuff for every request.

* This now includes a clean customer-facing LoggerMessageProperties class that is designed to be used by the code generator and by hand-written code.

* This now includes the StructuredLog extension method which uses LoggerMessageProperties to deliver a first class structured logging experience, with full support for enrichment and redaction.

davidfowl:
This looks much cleaner than what's currently there and its marked experimental. Are you going to measure the overhead?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// Emits a structured log entry.
+    /// </summary>
+    /// <param name="logger">The logger to emit the log entry to.</param>
+    /// <param name="logLevel">Entry will be written on this level.</param>
+    /// <param name="eventId">Id of the event.</param>
+    /// <param name="properties">The set of name/value pairs to log with this entry.</param>
+    public static void StructuredLog(this ILogger logger, LogLevel logLevel, EventId eventId, LoggerMessageProperties properties)
+    {
+        Throw.IfNull(logger).Log<LoggerMessageProperties>(logLevel, eventId, properties, null, (_, _) => string.Empty);
```
davidfowl, line 26: There's no formatted message? What does the console logger do with this?

geeknoid:
> This looks much cleaner than what's currently there and its marked experimental. Are you going to measure the overhead?

This is marked experimental, but it will definitely not ship in .NET 8 as experimental. This needs to be a fully supported shipping feature. Otherwise, we will have nothing for 1P customers.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
geeknoid: I'm using a specific layout here to allow the JIT to eliminate the code paths and conditional checks that aren't needed at runtime when the TState is known statically.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerExtensions.cs
geeknoid: We could do two things:

* Add a formatter than cranks up a string builder and concats all the properties together in some legible form.

* Extend the console logger to support structured logging by outputting dimensions in a first class way. This was on the todo list but was punted. I suggest we bring this back since it's a fairly easy drop-in which enables more efficient logging use, and supports enrichment and redaction cleanly.

geeknoid:
Third round:

* Further enhance performance, in particular for classic logging that's not using the new code generator.

* Made ExtendedLoggerFactory public. It is not a clean superset of what LoggerFactory exposes so it can be used as a complete substitute with no regrets.

* Replaced LoggingOptions with ExtendedLoggerFilterOptions, which extends the existing LoggerFilterOptions type.

* Made AddExtendedLogging exist on IServiceCollection instead of ILoggingBuilder, to match AddLogging.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -3,179 +3,80 @@
...
 
-                // This parameter needs to be null for OpenTelemetry.Exporter.Geneva to pick up LogRecord.StateValues (the last parameter).
-                // This is equivalent to using OpenTelemetryLogger with ParseStateValues option set to true.
-                null,
-                exception,
-                propertyBag);
+    public bool IsEnabled(LogLevel logLevel)
+    {
+        // TODO: is this the best we can do here? Should/could the enabled state be tracked in this instance so we can avoid the interface call?
+        return _nextLogger.IsEnabled(logLevel);
```
geeknoid, line 60: @davidfowl @noahfalk Is this the best I can do here? I'd love to avoid the overhead of this interface call.

Similarly, the check at the top of the ModernPath/R9Path/LegacyPath methods are also inducing these interface calls.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerExtensions.cs
davidfowl: > Add a formatter than cranks up a string builder and concats all the properties together in some legible form.

I think this is what we should do for now since it'll work for most logging providers (they all call the formatter by default).

davidfowl: > Extend the console logger to support structured logging by outputting dimensions in a first class way. This was on the todo list but was punted. I suggest we bring this back since it's a fairly easy drop-in which enables more efficient logging use, and supports enrichment and redaction cleanly.

Do we have this issue filed?

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+[Experimental]
+public static class ExtendedLoggingExtensions
+{
+    /// <summary>
+    /// Configure extended logging, enabling redaction and enrichment.
+    /// </summary>
+    /// <param name="services">The dependency injection container to add logging to.</param>
+    /// <param name="configure">Logging configuration options.</param>
+    /// <returns>The value of <paramref name="services"/>.</returns>
+    public static IServiceCollection AddExtendedLogging(this IServiceCollection services, Action<ILoggingBuilder> configure)
```
davidfowl, line 26: This is experimental so I don't mind that much but I think this should be exposed on ILoggingBuilder instead of IServiceCollection.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -3,179 +3,80 @@
...
+        {
+            ModernPath(logLevel, eventId, (LoggerMessageProperties?)(object?)state, exception, (Func<LoggerMessageProperties, Exception?, string>)(object)formatter);
         }
-        catch (Exception ex)
+        else if (typeof(TState) == typeof(LogMethodHelper))
         {
-            LoggingEventSource.Log.LogException(ex);
-            throw;
+            // temporary support for legacy generated code, until generator is updated
+            R9Path(logLevel, eventId, (LogMethodHelper?)(object?)state, exception, (Func<LogMethodHelper, Exception?, string>)(object)formatter);
```
davidfowl, line 72: Change the name to legacy source generated path or something like that.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
geeknoid: This is experimental only for the next few weeks. We will review the API and make it non-experimental before .NET 8 ships.

Why on logging builder instead of service collection? AddLogging is on service collection and this is a replacement for AddLogging, so it seemed reasonable to have this on the same interface.

davidfowl: Because I would expect this to disappear when we merge these changes into the logger factory in .NET 9.

geeknoid:
@noahfalk @dpk83 The PR is now in pretty good shape. I have some basic tests that show the machinery is generally working. I'll get to 100% coverage Monday. Please see the few outstanding notes in the description above.

Relative to the original R9 solution, I expect this to be faster as we now have zero-copy static log enrichment, and I've replaced pools with faster thread-local storage instead for working state. All calls to ILogger.Log now involve going through an additional ILogger instance than before, so we pay for a new interface call.

A quick way to eliminate the overhead is to take the logic that's in the default logger from Microsoft.Extensions.Logging and embed directly into the new Logger. We wouldn't need the indirection that way, at the cost of cut & paste of logic.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Enrichment/EnricherExtensions.cs
```diff
@@ -42,6 +43,38 @@ public static IServiceCollection AddLogEnricher(this IServiceCollection services
...
+    /// <returns>The value of <paramref name="services"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null"/>.</exception>
+    [Experimental]
+    public static IServiceCollection AddStaticLogEnricher<T>(this IServiceCollection services)
+        where T : class, IStaticLogEnricher
+    {
+        _ = Throw.IfNull(services);
+
+        return services.AddSingleton<IStaticLogEnricher, T>();
+    }
```
davidfowl, line 60: This should be on ILoggerBuilder

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Enrichment/EnricherExtensions.cs
```diff
@@ -42,6 +43,38 @@ public static IServiceCollection AddLogEnricher(this IServiceCollection services
...
+    /// <returns>The value of <paramref name="services"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="services"/> or <paramref name="enricher"/> are <see langword="null"/>.</exception>
+    [Experimental]
+    public static IServiceCollection AddStaticLogEnricher(this IServiceCollection services, IStaticLogEnricher enricher)
+    {
+        _ = Throw.IfNull(services);
+        _ = Throw.IfNull(enricher);
+
+        return services.AddSingleton(enricher);
+    }
```
davidfowl, line 76: Same as above.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
```diff
@@ -0,0 +1,263 @@
...
+using Microsoft.Extensions.Telemetry.Enrichment;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.Telemetry.Logging;
+
+/// <summary>
+/// Produces instances of <see cref="ILogger"/> classes based on the given providers, with support for redaction and enrichment.
+/// </summary>
+[Experimental]
+public sealed class ExtendedLoggerFactory : ILoggerFactory
```
davidfowl, line 22: This should be internal.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
```diff
@@ -0,0 +1,263 @@
...
+    private readonly IStaticLogEnricher[] _staticEnrichers;
+    private readonly IRedactorProvider? _redactorProvider;
+    private readonly ConcurrentDictionary<string, ILogger> _cache = new();
+    private readonly IDisposable? _changeTokenRegistration;
+    private ExtendedLoggerFilterOptions _filterOptions;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ExtendedLoggerFactory"/> class.
+    /// </summary>
+    public ExtendedLoggerFactory()
```
davidfowl, line 35: Why are there so many constructors.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -1,181 +1,100 @@
...
+// OPEN TOPICS
+//
+// * When logging stack traces, we include the exception message string. Is this safe from a compliance standpoint?
+//   Should we use the exception summarizer instead?
+//
+// * When using the structured logging extension methods, there is no formatter given. This breaks simple loggers
+//   like the console logger. We could fix the console logger to handle dimensions and/or we could change the
+//   extension method to produce a string from all the dimensions.
+//
+// * Should AddExtendedLogging live on IServiceCollection or ILoggingBuilder?
```
davidfowl, line 13: 👍🏾

src/Libraries/Microsoft.Extensions.Telemetry/Logging/Logger.cs
```diff
@@ -1,181 +1,100 @@
...
+//
+// * How can we mitigate the issues around AddLogging & AddExtendedLogging being present in the application?
```
davidfowl, line 15: 👍🏾

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
xakep139: So, before .NET 9 this will looke like this?
```csharp
services.AddLogging(x => x.AddExtendedLogging(/* ... */));
```

xakep139: Can you describe please how it will disappear in .NET 9? Will it be a breaking change?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerExtensions.cs
geeknoid: I did have an issue filed to move the R9 console logging functionality into .NET 8. But we ran out of time on that (API review overhead, etc) so decided for now to try and move it to the Otel console logging so at least the logic would be out in the wild for folks to adopt.

However, I think we should try to pursue something more modest. Rather than try to get all of the R9 features into the existing console logger, we should try to just get support for dimensions. It ends up just being a for loop somewhere that iterates through the name/value pairs and maybe an option to control whether this happens or not.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
davidfowl: This method will be obsolete. The logger factory changed here will be integrated into the logger factory. These methods exists to support redaction and enrichment and shouldn’t be exposed

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
davidfowl: There’s no need to copy all of the constructors as this isn’t a primary public API for consumers to use. It’s just copying the legacy from LoggerFactory and complicating this design

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerExtensions.cs
davidfowl: We should revive the issue on dotnet/runtime regardless.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// Enables redaction functionality within the logging infrastructure.
+    /// </summary>
+    /// <param name="builder">The dependency injection container to add logging to.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    public static ILoggingBuilder EnableRedaction(this ILoggingBuilder builder)
+    {
+        _ = Throw.IfNull(builder);
+
+        builder.Services.TryAdd(ServiceDescriptor.Singleton<ILoggerFactory, ExtendedLoggerFactory>());
+        builder.Services.TryAdd(ServiceDescriptor.Singleton(typeof(ILogger<>), typeof(Logger<>)));
```
davidfowl, line 31: You don't need this.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        _ = builder.Services.AddValidatedOptions<ExtendedLoggerFilterOptions, ExtendedLoggerFilterOptionsValidator>();
+        builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<ExtendedLoggerFilterOptions>>(new DefaultLoggerLevelConfigureOptions(LogLevel.Information)));
```
davidfowl, line 34: Why do you need this?

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -0,0 +1,396 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// OPEN TOPICS
+//
+// * When logging stack traces, we include the exception message string. Is this safe from a compliance standpoint?
+//   Should we use the exception summarizer instead?
+//
+// * When using the structured logging extension methods, the formatter currently creates a string of all the
+//   dimension values. Is this what we want? Should we instead improve the console logger to understand dimensions
```
noahfalk, line 10: > When using the structured logging extension methods

Who is expected to use these? I was thinking these were helpers called by the source generator and not intended for developers to invoke directly.

What creates the formatter in the source generator case when the user doesn't specify a message format in their LoggerMessage attribute, or they explicitly specify an empty string?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerMessageState.ClassifiedProperty.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.Compliance.Classification;
+
+namespace Microsoft.Extensions.Telemetry.Logging;
+
+public partial class LoggerMessageState
```
noahfalk, line 10: Even though loggers are recommended to use the formatter delegate, I believe some of them just call ToString() instead. All of the built-in TStates that .NET generates implement ToString() so that produces the same string you would have gotten if you called formatter(state, null);

noahfalk (commented):
My generalized worries about the future aside, this looks pretty good to me :) Some comments on specific things inline.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Enrichment/EnricherExtensions.cs
```diff
@@ -42,6 +43,38 @@ public static IServiceCollection AddLogEnricher(this IServiceCollection services
...
     }
 
+    /// <summary>
+    /// Registers a static log enricher type.
+    /// </summary>
+    /// <param name="services">The dependency injection container to add the enricher type to.</param>
+    /// <typeparam name="T">Enricher type.</typeparam>
+    /// <returns>The value of <paramref name="services"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null"/>.</exception>
+    [Experimental]
```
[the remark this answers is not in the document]
geeknoid: This is how we drive our API review process. It stays experimental until approved.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerMessageState.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// Gets the property collector instance.
+    /// </summary>
+    /// <param name="parameterName">The name of the parameter to prefix in front of all property names inserted into the collector.</param>
+    /// <returns>The collector instance.</returns>
+    /// <remarks>
+    /// This method is used by the logger message code generator to get an instance of a collector to
+    /// use when invoking a custom property collector method.
+    /// </remarks>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public ILogPropertyCollector GetPropertyCollector(string parameterName)
```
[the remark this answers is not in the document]
geeknoid: The way you suggest is how it was previously (see LogMethodHelper as a reference). While looking at it, I realized this pattern was wasting cycles in the 99% case.

ILogPropertyCollector is rarely used, say in < 1% of log statements. It's necessarily an interface since it's implemented by customers and making a base class instead would be constraining. In the case of ILogPropertyCollector, we need to inject the parameter name in front of all the properties. Rather than paying the cost of an if check any time a property is added to the property bag, this cost is now borne only by the ILogPropertyCollector case, instead of burdening the other 99% of cases.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -0,0 +1,396 @@
...
+                enricher.Enrich(properties.EnrichmentPropertyBag);
+            }
+
+            // one last dedicated bit of enrichment
+            if (exception != null && _includeStackTraces)
+            {
+                properties.AddProperty(ExceptionStackTrace, GetExceptionStackTrace(exception));
+            }
+
+            var joiner = Joiner;
```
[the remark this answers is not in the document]
geeknoid: Hmmm. Someone calling a log method from within the formatter function seems to be intentionally asking for trouble. But I can imagine it happening purely through composition in some error case for example.

So how about I use CompareExchange to get the stuff from TLS and set it to null during execution, and then put it back when I'm done. If some other code executes reentrantly, it wil see null values in the TLS and will allocate new objects in that case.

Either that, or I go back to object pools.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /// <summary>
+    /// Enables enrichment functionality within the logging infrastructure.
+    /// </summary>
+    /// <param name="builder">The dependency injection container to add logging to.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    public static ILoggingBuilder EnableEnrichment(this ILoggingBuilder builder)
+    {
+        // for now, both redaction and enrichment are enabled simultaneously.
```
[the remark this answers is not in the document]
geeknoid: How about I just provide a single EnableEnrichmentAndRedaction call?

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFilterOptions.cs
```diff
@@ -3,39 +3,20 @@
...
 
 namespace Microsoft.Extensions.Telemetry.Logging;
 
 /// <summary>
-/// Options for logger.
+/// Options for extended logging.
 /// </summary>
-public class LoggingOptions
+[Experimental]
+public class ExtendedLoggerFilterOptions : LoggerFilterOptions
```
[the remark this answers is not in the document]
geeknoid: Rename, and not make it derive from the existing option type?

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
geeknoid: In the source generator right now, if the customer doesn't supply a log message then the formatter function return an empty string.

I'm going to remove these functions I think, they don't add enough value and I want folks to use the code generator anyway.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFilterOptions.cs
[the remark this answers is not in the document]
geeknoid: Done.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerMessageState.ClassifiedProperty.cs
geeknoid: Added ToString implementation to the LoggerMessageState.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    public bool IsEnabled(LogLevel logLevel)
+    {
+        return _nextLogger.IsEnabled(logLevel);
+    }
+
+    public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+    {
+        if (typeof(TState) == typeof(LoggerMessageState))
```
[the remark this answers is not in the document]
geeknoid: Noted.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerMessageState.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.Extensions.Telemetry.Enrichment;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.Telemetry.Logging;
+
+/// <summary>
+/// Additional state to use with <see cref="ILogger.Log"/>.
+/// </summary>
+[Experimental]
+public sealed partial class LoggerMessageState : IResettable
```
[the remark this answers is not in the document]
geeknoid: Noted.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
```diff
@@ -0,0 +1,94 @@
...
+    public void AddProvider(ILoggerProvider provider) => _loggerFactory.AddProvider(provider);
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        _loggerFactory.Dispose();
+        _changeTokenRegistration?.Dispose();
+    }
+
+    private void UpdatedFilterOptions(ExtendedLoggerFilterOptions filterOptions) => _filterOptions = filterOptions;
```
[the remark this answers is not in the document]
geeknoid: Fixed by having the loggers point back to the factory and get the options dynamically, thus picking up changes as they occur.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Enrichment/EnricherExtensions.cs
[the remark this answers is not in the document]
noahfalk: Does that mean experimental APIs get shipped with no review or they never stay experimental after review? It feels awkward to me to couple in that way, but I better understand what to expect now.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerMessageState.cs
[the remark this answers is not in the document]
noahfalk: > It's necessarily an interface since it's implemented by customers

Is there an example of where a customer would implement ILogPropertyCollector? I see where customers would call it when using the LogProperties attribute, but I didn't see there was a scenario where customers would implement a different implementation. This is just for my understanding, I'm not worried about it.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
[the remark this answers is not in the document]
noahfalk: > So how about I use CompareExchange to get the stuff from TLS and set it to null during execution

The CompareExchange should be unneeded because the value is already threadlocal. A non-atomic null write should work fine. TLS reads/writes have more overhead than a standard read/write (I think 5-10ns vs. fraction of ns) so if you are trying to squeeze out performance one thing you can do is:

```C#
class ThreadLocalPool
{
    //whatever objects you want TLS singletons for
    Type1 Object1 { get; set; } = new Object1();
    Type2 Object2 { get; set; } = new Object2();
    ...
}


[ThreadStatic]
static ThreadLocalPool? _pool;

void DoSomethingThatNeedsObj1()
{
    ThreadLocalPool pool = _pool; // a single TLS read in the normal case
    if(pool == null)
    {
        pool = _pool = new ThreadLocalPool();
    }
    // these reads and writes are standard fast field accesses
    Type1 obj1 = poolObject1 ?? new Object1();
    pool.Object1 = null;
    // use obj1
    pool.Object1 = obj1;
}
```

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggingExtensions.cs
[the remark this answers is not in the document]
noahfalk: My understanding is that Fowler was trying to achieve decomposition, and to pick names that would be plausibly acceptable if this had gone through API review. I think `EnableEnrichmentAndRedaction` wouldn't achieve either of those goals.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
noahfalk: That sounds fine to me as source generator behavior, its compatible with existing behavior.

What is it that we expect calls the structured logging methods?

noahfalk:
Heads up @tarekgh, not sure if you already got mentioned here so double checking.

davidfowl:
@geeknoid to make review easier, can you do multiple commits and then squash at the end?

geeknoid:
@davidfowl Sorry, bad habit from ADO. Even if you combine your commits, ADO shows them individually so you can still look at 'em. I'll keep 'em separate from now on.

bench/Libraries/Microsoft.Extensions.Telemetry.PerformanceTests/ModernCodeGen.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Microsoft.Extensions.Telemetry.Bench;
+
+internal static partial class ModernCodeGen
+{
+    /// <summary>
+    /// Logs "Connection id '{connectionId}' received {type} frame for stream ID {streamId} with length {length} and flags {flags} and {other}" at "Error" level.
+    /// </summary>
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Gen.Logging", "8.0.0.0")]
+    public static void RefTypes(global::Microsoft.Extensions.Logging.ILogger logger, string connectionId, string type, string streamId, string length, string flags, string other)
```
xakep139, line 16: Why no use a `[LogMethod]` here?

bench/Libraries/Microsoft.Extensions.Telemetry.PerformanceTests/Program.cs
```diff
@@ -6,7 +6,7 @@
 using BenchmarkDotNet.Running;
 using BenchmarkDotNet.Toolchains.InProcess.Emit;
 
-namespace Microsoft.Gen.Logging.Bench;
+namespace Microsoft.Extensions.Telemetry.Bench;
```
xakep139, line 9: `Microsoft.Extensions.Telemetry.Logging.Bench`?

bench/Libraries/Microsoft.Extensions.Telemetry.PerformanceTests/ModernCodeGen.cs
geeknoid: This code shows new code generation pattern, but the code generator doesn't yet produce this output. So I hand-wrote it.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -0,0 +1,350 @@
...
+    public MessageLogger[] MessageLoggers { get; set; } = Array.Empty<MessageLogger>();
+    public ScopeLogger[] ScopeLoggers { get; set; } = Array.Empty<ScopeLogger>();
+
+    public ExtendedLogger(ExtendedLoggerFactory factory, LoggerInformation[] loggers)
+    {
+        _factory = factory;
+        Loggers = loggers;
+    }
+
+    public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
```
dpk83, line 38: How does it call into the OpenTelemetry processors here? It will require us to somehow create the LogRecord object and then pass it to the OpenTelemetry processors to further process.

geeknoid: This is the top-level logger implementation, it delegates to all low-level loggers. In a normal scenario, there will be an Otel-specific logger register which will get called with a TState that has all the enriched properties and redacted state.

geeknoid: So basically:

* Application layer calls ILogger.Log using a LoggerMessageState object as the TState.
* ExtendedLogger.Log is invoked and recognizes the LoggerMessageState type. It performs enrichment and redaction, resulting in a ModernPropertyJoiner object.
* ExtendedLogger.Log then calls all lower-level loggers, passing the ModernPropertyJoiner object as TState. Thus, all lower-level loggers receives enriched and redacted state.

dpk83: Got it. So, we have already done everything and then one can add OpenTelemetry logger and that logger will directly receive the logs which are alreayd redacted and enriched.

src/Libraries/Microsoft.Extensions.Telemetry/Enrichment.Process/StaticProcessLogEnricher.cs
```diff
@@ -0,0 +1,42 @@
...
+    public StaticProcessLogEnricher(IOptions<ProcessLogEnricherOptions> options)
+    {
+        var enricherOptions = Throw.IfMemberNull(options, options.Value);
+
+        if (enricherOptions.ProcessId)
+        {
+#if NET5_0_OR_GREATER
+            var pid = Environment.ProcessId;
+#else
+            var pid = System.Diagnostics.Process.GetCurrentProcess().Id;
```
xakep139, line 28: Shouldn't `System.Diagnostics.Process.GetCurrentProcess()` be disposed?

xakep139: IIRC it allocates some unmanaged resources

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -0,0 +1,351 @@
...
+        joiner.StaticProperties = config.StaticProperties;
+        joiner.Formatter = formatter;
+        joiner.SetIncomingProperties(msgState);
+
+        // enrich
+        foreach (var enricher in config.Enrichers)
+        {
+            try
+            {
+                enricher(joiner.PropertyBag);
```
xakep139, line 228: @geeknoid what if we provide here an additional information/context to an enricher? Otherwise `IStaticLogEnricher` is not that different from `ILogEnricher`, apart from using some fancy things like `AsyncContext`.
cc @andrey-noskov

geeknoid: What would the enricher do with the tstate? I'm fine with the idea of doing this, so long as we have a use case to justify it.

PR #4111, opened 2023-06-23
Resolve a "FIXME" in Options.Validation

xakep139:
This PR is a follow-up after https://github.com/dotnet/runtime/pull/84280 was merged

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4111)

### DIFF
--- a/src/ToBeRemoved/Options.ValidateOnStart/OptionsBuilderExtensions.cs
+++ b/src/ToBeRemoved/Options.ValidateOnStart/OptionsBuilderExtensions.cs
@@ -33,7 +33,7 @@ public static class OptionsBuilderExtensions
 
         _ = services.AddOptions();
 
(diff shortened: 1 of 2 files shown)


xakep139:
@RussKie FYI

geeknoid:
Why do we need this change? We should be deleting the stuff in ToBeRemoved if possible, rather than patching it...

xakep139:
🤦‍♂️ my bad, haven't noticed the folder name. I was viewing the code via VS in solution view only. Abandoning th PR.

RussKie:
> We should be deleting the stuff in ToBeRemoved if possible,

@geeknoid @joperezr do we have a plan for Options.ValidateOnStart project? It's the only one in the "ToBeRemoved" folder.

RussKie:
I guess this is https://github.com/dotnet/runtime/issues/84347 and https://github.com/dotnet/runtime/issues/86511?

PR #4121, opened 2023-06-27
Simplify task handling in MetricCollector to avoid racy edge case

geeknoid:


src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -130,18 +130,18 @@ public void Dispose()
...
-            // wake up anybody still waiting and inform them of the bad news: their horse is dead...
-            foreach (var w in _waiters)
-            {
-                _ = w.TaskSource.TrySetException(MakeObjectDisposedException());
-            }
+        _meterListener.Dispose();
+        _measurements.Clear();
 
-            _waiters.Clear();
+        // wake up anybody still waiting and inform them of the bad news: their horse is dead...
```
tarekgh, line 138: could you please add a comment in all places we complete the waiter.TaskSource telling need to be called outside the lock.

tarekgh (approved):

RussKie:
Test failure: https://dev.azure.com/dnceng-public/public/_build/results?buildId=321638&view=ms.vss-test-web.build-test-results-tab&runId=6599882&resultId=109460&paneView=debug

geeknoid:
@tarekgh I added the comments about usage of the task needing to be outside of the critical section.

Unfortunately, I needed to leave the Thread.Sleep in the code since the test case is still failing. There is still something asynchronous happening which we don't understand. Can I impose on you again to figure out what's happening please?

PR #4136, opened 2023-06-28
Tame package versioning

### LINKED ISSUE #4098: Reference latest libraries packages
This repo is already doing a pretty good job of referencing the latest packages, but I noticed a few places that this could be improved.  @ViktorHofer might be able to spot more.

I'd like to understand why you're referencing older versions of the following packages.  I think the latest version from 8.0 should work everywhere.
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-net462.props#L9
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-net462.props#L14
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-net462.props#L17
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-net462.props#L19

This package reference should be avoided, so long as nothing else brings it in. 
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-net462.props#L20
The following can be removed if you target at least .NET 4.7 and nothing else brings it in -- ~~4.6.2 is out of support.~~
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-net462.props#L23

This doesn't appear to be used anywhere and should be removed.
https://github.com/dotnet/extensions/blob/e72de4cfeb1a2917694b96d0eaaabdc4e8b81478/eng/Packages/General-netcoreapp3.1.props#L15

geeknoid:
Fixes #4098

* Converge on .NET 8 package references when possible.

* Update other external dependencies to latest versions.

* Remove mentions of net472 and netstandard2.0 from various places,
since we don't build for those TFMs anymoe.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4136)


rafal-mz (approved):

eng/Packages/TestOnly.props
```diff
@@ -17,31 +17,40 @@
...
-  <Import Project="$(MSBuildThisFileDirectory)\TestOnly-net7.0.props" />
-  <Import Project="$(MSBuildThisFileDirectory)\TestOnly-latest.props" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc" Version="2.2.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.2.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.1.22" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="3.1.22" />
```
xakep139, line 39: Should we target latest patch versions? E.g. https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Testing/3.1.32

geeknoid: I'll take care of this in my next PR.

PR #4138, opened 2023-06-28
Remove a few #if that are no longer needed

### LINKED ISSUE #4001: cleanup #if NETxxx use
Clean up #if NET6_0_OR_GREATER that are now no longer needed, eg., because the package/dependencies now depend on .NET 8.

geeknoid:
Fixes #4001 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4138)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
@@ -231,15 +231,11 @@ public sealed class MetricCollector<T> : IDisposable
             {
                 lock (_measurements)
                 {
-                    _waiters.Remove(w);
+                    _ = _waiters.Remove(w);
                 }
 
                 // trigger the task from outside the lock
-#if NET6_0_OR_GREATER
-                w.TaskSource.TrySetCanceled(cancellationToken);
-#else
-                w.TaskSource.TrySetCanceled();
-#endif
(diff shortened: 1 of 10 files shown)


src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
```diff
@@ -231,15 +231,11 @@ public Task WaitForMeasurementsAsync(int minCount, CancellationToken cancellatio
...
+                    _ = _waiters.Remove(w);
                 }
 
                 // trigger the task from outside the lock
-#if NET6_0_OR_GREATER
-                w.TaskSource.TrySetCanceled(cancellationToken);
-#else
-                w.TaskSource.TrySetCanceled();
-#endif
+                _ = w.TaskSource.TrySetCanceled(cancellationToken);
```
rafal-mz, line 238: Shouldn't we await this? This is testing code, and discarding a task will ignore potential exceptions in TrySetCanceled.

rafal-mz (approved):

src/Libraries/Microsoft.Extensions.Telemetry.Testing/Metering/MetricCollector.cs
geeknoid: Wait for what? TrySetCanceled returns a boolean. And this is intentionally fire-and-forget so we don't care about the return value.

PR #4141, opened 2023-06-29
Cleanup Microsoft.Extensions.Http.Resilience internals

martintmk:
Simplifying and dropping some unnecessary internal abstractions. 

Changes:

- Dropping some internal interfaces or turning them into abstract classes or just simple options.
- Dropping additional dependencies (Polling, NamedSingletons) making this library leaner.
- Enabling pooling of `RequestMessageSnapshot`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4141)


src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/StrategyKeyOptions.cs
```diff
@@ -1,19 +1,15 @@
...
+using System;
 using System.Net.Http;
 
 namespace Microsoft.Extensions.Http.Resilience.Internal;
 
 /// <summary>
 /// The provider that returns the strategy key from the request message.
 /// </summary>
-internal interface IStrategyKeyProvider
+internal sealed class StrategyKeyOptions
```
geeknoid, line 12: Is this interface used in DI? If not, you could just be using Func<HttpRequestMessage, string> directly

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/StrategyKeyProviderHelper.cs
```diff
@@ -17,25 +18,22 @@ public static void SelectStrategyByAuthority(IServiceCollection services, string
         {
             var redactor = serviceProvider.GetRequiredService<IRedactorProvider>().GetRedactor(classification);
 
-            return new ByAuthorityStrategyKeyProvider(redactor);
+            return new ByAuthorityStrategyKeyProvider(redactor).GetStrategyKey;
```
geeknoid, line 21: Perhaps rename GetStrategyKey to just StrategyKey?

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/StrategyKeyOptions.cs
martintmk: Yes, it comes from DI. This options class is just a holder for that delegate.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/StrategyKeyProviderHelper.cs
martintmk: `GetStrategyKey` is just a function that gets converted to a delegate. I think the name is ok here

PR #4181, opened 2023-07-17
Revive M.E.DependencyInjection.AutoActivation

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4181)


RussKie:
FYI the new package must be recorded in [dotnet/api-catalog-infra](https://github.com/dotnet/api-catalog-infra/) so that the Docs team has the right content for the API docs. You can use https://github.com/dotnet/api-catalog-infra/pull/63 as a reference.

src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/AutoActivationExtensions.cs
```diff
@@ -276,6 +276,8 @@ private static IServiceCollection Activate(this IServiceCollection services, [Dy
         return services;
     }
 
+    // exclude from coverage until https://github.com/microsoft/codecoverage/issues/38 is addressed
+    [ExcludeFromCodeCoverage]
     private static IServiceCollection AddHostedServiceIfNotExist(this IServiceCollection services)
     {
 #if NETFRAMEWORK
```
xakep139, line 283: Do we build for .NET FX?

RussKie: Yes, all libraries are built for `.net8;.netcore3.1;net462` unless explicitly opted-out
https://github.com/dotnet/extensions/blob/5352825bd11925eff0cc0dcd3283e1481fa75cb6/src/Libraries/Directory.Build.props#L5
https://github.com/dotnet/extensions/blob/5352825bd11925eff0cc0dcd3283e1481fa75cb6/Directory.Build.props#L12-L31

joperezr (approved):
LGTM. As @RussKie mentioned above, please make sure we also add this to the catalog in order to get the package's API Docs published.

PR #4183, opened 2023-07-18
Resolve RS1024 suppression

### LINKED ISSUE #4127: Resolve RS1024 suppression
https://github.com/dotnet/extensions/blob/5a8029df19146b9beec1983cf269445e36c52d49/Directory.Build.targets#L44-L47

catcherwong:
fixes #4127 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4183)

### DIFF
--- a/eng/packages/General.props
+++ b/eng/packages/General.props
@@ -5,10 +5,10 @@
     <PackageVersion Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
     <PackageVersion Include="Microsoft.Bcl.TimeProvider" Version="$(MicrosoftBclTimeProviderVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.1.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.1.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.1.0" />
     <PackageVersion Include="Microsoft.Extensions.Caching.Abstractions" Version="$(MicrosoftExtensionsConfigurationAbstractionsVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Caching.Memory" Version="$(MicrosoftExtensionsConfigurationAbstractionsVersion)" />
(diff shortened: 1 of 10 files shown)


RussKie:
Thanks for the PR. Looks like the change introduced more warnings.

catcherwong:
> Looks like the change introduced more warnings.

Fixed the warnings, all checks have passed.

eng/packages/General.props
```diff
@@ -5,10 +5,10 @@
     <PackageVersion Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
     <PackageVersion Include="Microsoft.Bcl.TimeProvider" Version="$(MicrosoftBclTimeProviderVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.1.0" />
```
geeknoid, line 8: We need to discuss this change, as it impacts the SDK that can be used to invoke the generators. If nothing else, all those Roslyn4.0 names should be changed to Roslyn4.1

catcherwong: Following https://github.com/dotnet/roslyn-analyzers/issues/5804, to fix RS1024, we need to update the packages.

> If nothing else, all those Roslyn4.0 names should be changed to Roslyn4.1

Do you mean we should rename the folder and csproj from 4.0 to 4.1?

geeknoid: > Do you mean we should rename the folder and csproj from 4.0 to 4.1?

Yes.

But before doing that, we're debating whether all the analyzers and generators in this repo should be revised to only depend on Roslyn 4.4 (part of the .NET 7 SDK), rather than supporting Roslyn 3.8 (from the .NET 5 SDK) and Roslyn 4.0 (from the .NET 6 SDK).

If we end up snapping to Roslyn 4.4, then we can get simplify the source tree, removing some #if in the code, build the affected components and tests just once instead of twice. But customers would have a higher minimum bar in terms of SDK.

What's your opinion?

catcherwong: > Roslyn 4.0 (from the .NET 6 SDK).

I found the version of **Microsoft.CodeAnalysis.dll** is 
 4.3.1 in the newest .NET 6.0 SDK (v6.0.412), and 4.0.0 in the .NET 6.0 SDK (v6.0.100).

It seems that we can update the version from 4.0.1 to 4.1.0 or 4.3.1 without changed to Roslyn4.1 or Roslyn4.3?

> But customers would have a higher minimum bar in terms of SDK.

This may not be particularly good in terms of user experience. I think we can move to Roslyn 4.4 after .NET 6 SDK out of support.

> If we end up snapping to Roslyn 4.4, then we can get simplify the source tree, removing some #if in the code, build the affected components and tests just once instead of twice. 

Agree.

geeknoid: We decided that to align with other new generators being introduced in .NET 8, we'll adopt Roslyn 4.4 for all the generators and analyzers in this repo. This gives us a significant simplification in the source tree overall. I'll hopefully check that upgrade in shortly. As such, I'm going to close this PR since it will no longer be useful.

Thanks for your work.

PR #4194, opened 2023-07-20
Snap to Roslyn 4.4

geeknoid:
This means folks trying to use the generators and analyzers will need to be using at minimum the .NET 7 SDK.

Resolves https://github.com/dotnet/extensions/issues/4127
Closes https://github.com/dotnet/extensions/pull/4183 as superseded.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4194)


xakep139:
This change essentially means that we will force folks to use .NET 8 toolchain even if they are targeting `netcoreapp3.1`

geeknoid:
No, this forces the .NET 7 toolchain, not .NET 8.

The option code generator, which is now in dotnet/runtime also requires the .NET 7 toolchain, so we're being consistent.

ScottThurlow (approved):

danmoseley:
As noted offline this also forces newer Visual Studio as well. @marcpopMSFT  exactly which VS version would be the minimum after this change?

eng/MSBuild/Analyzers.props
```diff
@@ -26,8 +26,8 @@
...
-    <ProjectReference Condition="'$(SkipLocalAnalyzers)' != 'true'" Include="$(MSBuildThisFileDirectory)\..\..\src\Analyzers\Microsoft.Extensions.LocalAnalyzers\Microsoft.Extensions.LocalAnalyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" Private="false" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
+    <ProjectReference Condition="'$(SkipExtraAnalyzers)' != 'true'" Include="$(MSBuildThisFileDirectory)\..\..\src\Analyzers\Microsoft.Analyzers.Extra\Microsoft.Analyzers.Extra.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" Private="false" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
+    <ProjectReference Condition="'$(SkipLocalAnalyzers)' != 'true'" Include="$(MSBuildThisFileDirectory)\..\..\src\Analyzers\Microsoft.Analyzers.Local\Microsoft.Analyzers.Local.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" Private="false" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
```
RussKie, line 30: ❤️ the change of the names

We should rename the other analyzers and generators too. /ducks

eng/packages/General.props
```diff
@@ -5,10 +5,10 @@
...
     <PackageVersion Include="Microsoft.Bcl.TimeProvider" Version="$(MicrosoftBclTimeProviderVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.0.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.4.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.4.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.4.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.4.0" />
```
RussKie, line 11: If all of these packages should be on the same version then it's worth considering introducing a variable (put it in /eng/Versions.props) and use the variable here.
https://github.com/dotnet/extensions/blob/131ada7622ab51ceb79021c7a027db3bf104e9e1/eng/Versions.props#L73-L81

src/Analyzers/Microsoft.Analyzers.Local/Microsoft.Analyzers.Local.csproj
```diff
@@ -8,9 +8,7 @@
 
   <PropertyGroup>
     <SkipLocalAnalyzers>true</SkipLocalAnalyzers>
-      <AnalyzerLanguage>cs</AnalyzerLanguage>
-    <AnalyzerRoslynVersion>4.0</AnalyzerRoslynVersion>
-    <DefineConstants>$(DefineConstants);ROSLYN_4_0_OR_GREATER</DefineConstants>
+    <AnalyzerLanguage>cs</AnalyzerLanguage>
```
RussKie, line 11: There's an established naming convention in the SDK to suffix assemblies with the language. E.g.,
* Microsoft.Analyzers.Local.dll implies a language agnostic analyzers,
* Microsoft.Analyzers.Local.CSharp.dll implies C#-specific analyzers.

It'd worth considering adapting to the same convention.

src/Generators/Microsoft.Gen.AutoClient/Microsoft.Gen.AutoClient.csproj
```diff
@@ -0,0 +1,32 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <RootNamespace>Microsoft.Gen.AutoClient</RootNamespace>
+    <Description>Code generator to support Microsoft.Extensions.Http.AutoClient.</Description>
+    <Workstream>Fundamentals</Workstream>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AnalyzerLanguage>cs</AnalyzerLanguage>
```
RussKie, line 9: Are we applies this property explicitly to all the analyzer projects? If so, consider declaring it centrally in one of the Directory.Build.props.

marcpopMSFT:
> As noted offline this also forces newer Visual Studio as well. @marcpopMSFT exactly which VS version would be the minimum after this change?

The 4.4 Roslyn matches with the 17.4 VS which ships with the 7.0.1xx SDK if that's what you're asking.

danmoseley:
Thanks @marcpopMSFT that was what I was asking. Essentially they need to have VS 2022 because if they're <17.4 they can presumably update.

@ScottThurlow do we have any data on the versions of VS that internal folks use when consuming these bits? Put another way, are more than a trivial proportion using VS2019?

btw I'm assuming this PR only concerns analyzers and generators that folks consuming our components need, I don't think we care about anything that's only relevant to building them themselves.

danmoseley:
eh, we can move this conversation to the original chat. I should have asked there.

PR #4219, opened 2023-07-27
Improve DependencyInjection.AutoActivationExtensions

geeknoid:
- Avoid calling MakeGenericType when possible.

- Delete the now-useless Hosting.StartupINitialization component.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4219)


martintmk:
@geeknoid, just wondering. 

What is the replacement for `IStartupInitialization`?

geeknoid:
@martintmk @ericstj What's the actual initialization API that was added to dotnet/runtime?

geeknoid:
@steveharter Note that I'll add support for keyed services in a follow-on PR. I've got it all coded, but I need to wait for https://github.com/dotnet/runtime/issues/89447 to be fixed, otherwise none of it works.

ericstj:
> Ah, I think this is it: https://github.com/dotnet/runtime/pull/87335

Yes - the decision was made to not create a new initializer concept but instead incorporate the ask into hosted services themselves.

rafal-mz (approved):

RussKie:
@geeknoid @ericstj  so how do we fix these?
https://github.com/Azure/dotnet-extensions-experimental/blob/0be630b7abb70e9316fcd7868aa64e088a9bbcfe/src/Microsoft.Azure.Extensions.DocumentDb.Cosmos/Extensions/DatabaseBuilder.cs#L199-L200
https://github.com/Azure/dotnet-extensions-experimental/blob/0be630b7abb70e9316fcd7868aa64e088a9bbcfe/src/Microsoft.Azure.Extensions.DocumentDb.Cosmos/Extensions/TableConfigurer.cs#L44-L45

ericstj:
@steveharter can help with these - I believe you just need an `IHostLifecycle` implementation instead of delegate.  You might even be able to use an existing `IHostedService` in the same library to do the job if one exists.

steveharter:
Yes `IHostedLifecycleService.Starting` should be used since that occurs before `IHostedService.Start`.

Here's a pattern from an earlier prototype of mine that has a fallback in case the host doesn't support the new interface:
```cs
  internal sealed class StartupActivator : IHostedLifecycleService
  {
      private bool _done;

      Task IHostedLifecycleService.StartingAsync(CancellationToken cancellationToken) => Activate();

      // For backwards compat with hosts that don't support IHostedLifecycleService, use StartAsync().
      Task IHostedService.StartAsync(CancellationToken cancellationToken) => Activate();

      Task IHostedLifecycleService.StartedAsync(CancellationToken cancellationToken) => Task.CompletedTask;
      Task IHostedService.StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
      Task IHostedLifecycleService.StoppedAsync(CancellationToken cancellationToken) => Task.CompletedTask;
      Task IHostedLifecycleService.StoppingAsync(CancellationToken cancellationToken) => Task.CompletedTask;

      private Task Activate()
      {
          if (!_done)
          {
              // activate the service(s) here
              _done = true;
          }

          return Task.CompletedTask;
      }
}
```

PR #4221, opened 2023-07-27
Nuke SocketHandling since it's now in dotnet/runtime

geeknoid:
https://github.com/dotnet/runtime/issues/84075

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4221)


joperezr (approved):
LGTM, thanks. What about the API that wasn't moved down? Don't we still wanted a thin layer for the certificate handling in here?

geeknoid:
@joperezr Yes, we probably do want this layer. But that work can be handled separately when we get there. At this point, we have nothing in this repo needing this functionality, so we can defer for now.

PR #4224, opened 2023-07-28
HttpClient logging based on new .NET 8 runtime APIs

### LINKED ISSUE #4211: HttpClient logging based on .net8
HttpClient logging with support of:
 - sensitive data redaction (outgoing request metadata)
 - enrichment (global + httpclient)

xakep139:
Fixes #4211

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4224)


src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/IHttpClientLogEnricher.cs
```diff
@@ -17,5 +18,8 @@ public interface IHttpClientLogEnricher
     /// <param name="enrichmentBag">Property bag to add enriched properties to.</param>
     /// <param name="request"><see cref="HttpRequestMessage"/> object associated with the outgoing HTTP request.</param>
     /// <param name="response"><see cref="HttpResponseMessage"/> object associated with the outgoing HTTP request.</param>
-    void Enrich(IEnrichmentPropertyBag enrichmentBag, HttpRequestMessage? request = null, HttpResponseMessage? response = null);
+    /// <param name="exception">An optional <see cref="Exception"/> that was thrown within the outgoing HTTP request processing.</param>
+    // TODO: adding a new parameter to the interface method is a breaking change. Pay additional attention to this during review.
+    // Alternatively, we can decide to add another overload instead.
+    void Enrich(IEnrichmentPropertyBag enrichmentBag, HttpRequestMessage request, HttpResponseMessage? response = null, Exception? exception = null);
```
xakep139, line 24: We need to discuss this one

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
```diff
@@ -0,0 +1,234 @@
...
+            // Return back pooled objects (since the logRecord wasn't fully prepared):
+            _logRecordPool.Return(logRecord);
+
+            if (requestHeadersBuffer is not null)
+            {
+                _headersPool.Return(requestHeadersBuffer);
+            }
+
+            // Recommendation is to swallow the exception (logger shouldn't throw), so we don't re-throw here:
+            return null;
```
xakep139, line 95: As @CarnaViire pointed, a logger shouldn't re-throw (but it will change our current behavior)

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
```diff
@@ -0,0 +1,234 @@
...
+        HttpRequestMessage request,
+        HttpResponseMessage? response,
+        Exception exception,
+        TimeSpan elapsed,
+        CancellationToken cancellationToken = default)
+            => await LogResponseAsync(context, request, response, exception, elapsed, cancellationToken).ConfigureAwait(false);
+
+    public object? LogRequestStart(HttpRequestMessage request)
+    {
+        throw new NotSupportedException();
```
xakep139, line 118: As of now the best we can offer for sync overloads - is to call their async counterparts with `.GetAwaiter().GetResult()`, that's because there's no sync API of reading request/response message body

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
```diff
@@ -0,0 +1,234 @@
...
+                if (_logResponseHeaders)
+                {
+                    responseHeadersBuffer = _headersPool.Get();
+                }
+
+                await _httpRequestReader.ReadResponseAsync(logRecord, response, responseHeadersBuffer, cancellationToken).ConfigureAwait(false);
+            }
+
+            propertyBag = LogMethodHelper.GetHelper();
+            FillLogRecord(logRecord, propertyBag, in elapsed, request, response, exception);
```
xakep139, line 175: We don't use `TimeProvider` here since `elapsed` is provided out-of-the-box, but we can consider ignoring it and using the old logic

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
```diff
@@ -74,4 +95,30 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
...
         }
     }
+
+    private static string OriginalFormatValueFMT(LogMethodHelper request, Exception? _)
+    {
+        int startIndex = FindStartIndex(request);
+        var httpMethod = request[startIndex].Value;
+        var httpHost = request[startIndex + 1].Value;
+        var httpPath = request[startIndex + 2].Value;
+        return FormattableString.Invariant($"{httpMethod} {httpHost}/{httpPath}");
```
xakep139, line 105: This is to align with current behavior in internal library

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
```diff
@@ -40,22 +61,22 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
...
             }
 
-            if (!string.IsNullOrEmpty(record.RequestBody))
+            if (record.RequestBody is not null)
             {
                 collector.Add(HttpClientLoggingDimensions.RequestBody, record.RequestBody);
             }
 
-            if (!string.IsNullOrEmpty(record.ResponseBody))
+            if (record.ResponseBody is not null)
```
xakep139, line 79: Checking just for `null` will be much faster

test/Libraries/Microsoft.Extensions.Http.Telemetry.Tests/Logging/Internal/TestLoggingHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "Base class disposes the handler")]
+    public TestLoggingHandler(IHttpClientLogger logger, HttpMessageHandler? innerHandler)
+        : base(CreateInternalBclLoggingHandler(logger, innerHandler))
+    {
+    }
+
+    private static DelegatingHandler CreateInternalBclLoggingHandler(IHttpClientLogger logger, HttpMessageHandler? innerHandler)
+    {
+        var handlerType = typeof(IHttpClientFactory).Assembly.GetType("Microsoft.Extensions.Http.Logging.HttpClientLoggerHandler");
```
xakep139, line 22: This was taken from Natalia's PoC branch https://github.com/dotnet/extensions/compare/main...CarnaViire:extensions:poc-hcf-custom-logging

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
geeknoid: Trivially faster I think.

So what's the behavior when the body is empty?

xakep139: We'll log it as a separate dimension (but that would mean that its logging was enabled via options, but its content was empty). Previous logic was dropping it from being emitted either way.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/LogRecordPooledObjectPolicy.cs
```diff
@@ -16,8 +16,8 @@ public override bool Return(LogRecord record)
         record.Path = string.Empty;
         record.Duration = 0;
         record.StatusCode = null;
-        record.RequestBody = string.Empty;
-        record.ResponseBody = string.Empty;
+        record.RequestBody = null;
```
geeknoid, line 19: How about getting rid of this LogRecordPooledObjectPolicy object and instead having LogRecord implement IResettable? Less code overall.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
```diff
@@ -26,11 +42,16 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
...
+
+    [LogMethod(LogLevel.Error, ResponseReadErrorMessage)]
+    public static partial void ResponseReadError(ILogger logger, Exception exception, HttpMethod httpMethod, string httpHost, string httpPath);
 
-    // Using the code below to avoid every item in ILogger's logRecord State being prefixed with parameter name.
-    // To be fixed in R9 later.
+    [LogMethod(LogLevel.Error, EnrichmentErrorMessage)]
+    public static partial void EnrichmentError(ILogger logger, Exception exception, HttpMethod httpMethod, string httpHost, string httpPath);
+
+    // Using the code below instead of generated logging method because we have a custom formatter and custom tag keys for headers.
```
geeknoid, line 54: This should be rewritten to use a LoggerMessageState object instead of the LogMethodHelper stuff (which I'm going to delete soon

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
```diff
@@ -40,22 +61,22 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
 
             collector.AddRequestHeaders(record.RequestHeaders);
             collector.AddResponseHeaders(record.ResponseHeaders);
-            collector.Add(HttpClientLoggingDimensions.Host, record.Host);
             collector.Add(HttpClientLoggingDimensions.Method, record.Method);
+            collector.Add(HttpClientLoggingDimensions.Host, record.Host);
             collector.Add(HttpClientLoggingDimensions.Path, record.Path);
```
geeknoid, line 66: I'm surprised we don't do any redaction in this code. Can't the path contain sensitive info for example?

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
xakep139: Good catch, will update these methods

geeknoid: I don't think it makes sense here, but I want to bring it up anyway.

With the new logger model, redaction is handled inside the logger. Basically, when you have the LoggerMessageState object, you can add normal properties to it (name/value pairs) or you can add classified properties to it (name/value/data class). Based on the data class, the ILogger implementation performs redaction.

In this code, we are redacting the path explicitly within this component, and then calling the logger. Would/could it be simpler to delegate redaction to the logger implementation?

xakep139: Not the whole path segment will be redacted, each path parameter is redacted separately

test/Libraries/Microsoft.Extensions.Http.Telemetry.Tests/Logging/Internal/TestLoggingHandler.cs
CarnaViire: Are you sure you need this? The intent on my side was just to show that our HttpClientLoggerHandler passes all the tests you had for your logging handler. Like, "you can trust our handler" 😄

So I believe the set of tests that were verifying the handler itself can be removed from dotnet/extensions repo (I haven't looked at the PR yet, I don't know how exactly you've transformed the case).

Or if you do need a handler for some other tests, create it "normal way" via IHttpMessageHandlerFactory.

RussKie:
MC

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/IHttpClientLogEnricher.cs
xakep139: No, the breaking change here is that I added `Exception? exception` - which means that existing users of this API would need to add the new parameter into their implementation of the interface

src/Libraries/Microsoft.Extensions.Http.Telemetry/Latency/Internal/HttpClientLatencyLogEnricher.cs
```diff
@@ -30,7 +31,7 @@ public HttpClientLatencyLogEnricher(HttpClientLatencyContext latencyContext, ILa
         _enricherInvoked = tokenIssuer.GetCheckpointToken(HttpCheckpoints.EnricherInvoked);
```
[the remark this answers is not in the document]
xakep139: I didn't look into latency measurement source code

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/LogRecordPooledObjectPolicy.cs
xakep139: Done

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/Log.cs
xakep139: Rewritten to `LoggerMessageState`

src/Libraries/Microsoft.Extensions.Http.Telemetry/Latency/Internal/HttpClientLatencyLogEnricher.cs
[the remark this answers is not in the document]
xakep139: General guidance is that we don't add null checks in `internal` types

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -60,6 +60,7 @@ public static class HttpClientLoggingTagNames
     /// Gets a list of all dimension names.
     /// </summary>
     /// <returns>A read-only <see cref="IReadOnlyList{String}"/> of all dimension names.</returns>
+    // TODO: should we rename Dimensions to Tags here?
     public static IReadOnlyList<string> DimensionNames { get; } =
```
xakep139, line 64: @geeknoid should I rename the property to be `TagNames`?

xakep139: Or it's okay to use `Dimensions`?

geeknoid: TagNames please

xakep139: Done

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/IHttpClientLogEnricher.cs
```diff
@@ -17,5 +18,8 @@ public interface IHttpClientLogEnricher
     /// <param name="collector">Tag collector to add tags to.</param>
     /// <param name="request"><see cref="HttpRequestMessage"/> object associated with the outgoing HTTP request.</param>
     /// <param name="response"><see cref="HttpResponseMessage"/> object associated with the outgoing HTTP request.</param>
-    void Enrich(IEnrichmentTagCollector collector, HttpRequestMessage? request = null, HttpResponseMessage? response = null);
```
xakep139, line 20: @geeknoid @RussKie @dariusclay @dpk83 what do you think folks?
- is it better to keep the existing method `Enrich(IEnrichmentTagCollector, HttpRequestMessage, HttpResponseMessage)` and add `Enrich(IEnrichmentTagCollector collector, Exception, HttpRequestMessage, HttpResponseMessage)`; or
- extend the existing method to have `void Enrich(IEnrichmentTagCollector, HttpRequestMessage, HttpResponseMessage?, Exception?)` only?
What will be better for users (and better for backward-compat with 1P services)?

geeknoid: I think I'd rather stick to one method and add the optional exception.

RussKie: The main question is whether the changes are breaking or not. Once we GA, we can't have API breaking changes. Now we're at the RC stage, which means we may still be able to take such changes, but they must have a significant value.

The .NET runtime provides a doc on what's considered a breaking change, what's allowed and what's not - https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md. According to the doc, adding new parameters is considered a breaking change and is not allowed, meaning we need to provide an overload:
image

I'm not sure how the defaultness affects this rule. @ericstj @bartonjs can you help?

iliar-turdushev: As I understand, both of the approaches are already not compatible with 1P services, since the methods are using IEnrichmentTagCollector instead of IEnrichmentPropertyBag, right? If so then I think we can stick to what is more convenient for users. IMO having one method is more convenient.

ericstj: You can still make breaking changes during RC.  @RussKie is correct that this type of change should not be done once the package is made stable.  Note that an additional method here will break interface implementers just as bad as changing the signature.  You should do your best to make sure your interface is complete in its first release.  Focus on what's best here to avoid future breaking changes and allow yourself some breaks now.

xakep139: > I think I'd rather stick to one method and add the optional exception.

In that case the `response` will be optional as well.

RussKie:
Please re-target to release/8.0, if this is necessary for .NET 8.

xakep139:
> Please re-target to release/8.0, if this is necessary for .NET 8.

Will we merge `release/8.0` into `main` afterwards?

RussKie:
MC

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/IHttpClientLogEnricher.cs
xakep139: Okay folks, I'm going to merge this as-is since Martin and Iliar both voted for having just one method.

xakep139: `request` is not optional, you're looking at the old version of the code. `response` and `exception` are both optional.

xakep139: image

xakep139: @iliar-turdushev your comment was addressed in #4323 4323

PR #4230, opened 2023-08-01
Enable CS1591 errors for undocumented API

ViktorHofer:
Arcade by default NoWarns CS1591 errors. While that default is questionable, our team added a condition to not suppress that compiler error. Use that to enable CS1591 errors.

... and document the HttpClientBuilderExtensions class that was missing a class summary.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4230)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -12,6 +12,8 @@
         When StabilizePackageVersion is set to 'true', this branch will produce stable outputs for 'Shipping' packages
     -->
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
+    <!-- Enabling this rule will cause build failures on undocumented public APIs. -->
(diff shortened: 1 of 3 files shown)


carlossanlop (approved):

joperezr:
FYI: @RussKie

joperezr (approved):
Thanks for taking care of this @ViktorHofer

joperezr:
FWIW I believe we have in extensions our own analyzers and rules that were checking for this, which leads me to question how come there seemed to be one violation we weren't catching.

ViktorHofer:
Just as an FYI, @carlossanlop will eventually send a PR to dotnet/extensions to add a transport package for the dotnet-api-docs team. More details in https://github.com/dotnet/runtime/issues/996#issuecomment-1632899743.

That transport package will include both all reference assemblies and the compiler generated XML docs.

PR #4233, opened 2023-08-01
Revamp the logging generator to use the new redaction model.

geeknoid:
- This new code leverages LoggerMessageState to take advantage of the new ExtendedLogger design.

- The tests were updated for three reasons:

    * Stopped using IRedactorProvider in here, this is handled by
      the ExtendedLogger

    * Fixed brittle test logic assuming the order of the name/value pairs
      produced by the logging code generator. The tests are now order-
      independent.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4233)

### DIFF
--- a/src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
+++ b/src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
@@ -20,7 +20,7 @@ internal sealed partial class Emitter : EmitterBase
         string level = GetLoggerMethodLogLevel(lm);
         string extension = lm.IsExtensionMethod ? "this " : string.Empty;
         string eventName = string.IsNullOrWhiteSpace(lm.EventName) ? $"nameof({lm.Name})" : $"\"{lm.EventName}\"";
-        string exceptionArg = GetException(lm);
+        (string exceptionArg, string exceptionLambdaName) = GetException(lm);
         (string logger, bool isNullableLogger) = GetLogger(lm);
 
         if (!lm.HasXmlDocumentation)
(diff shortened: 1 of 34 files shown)


geeknoid:
Note that Emitter.Method contains a bunch of TODO items, I'll tackle these in separate PRs.

test/Generators/Microsoft.Gen.Logging/Generated/Utils.cs
```diff
@@ -0,0 +1,68 @@
...
+            builder.AddProvider(provider);
+            builder.EnableRedaction();
+        });
+
+        serviceCollection.AddRedaction(builder =>
+        {
+            builder.SetFallbackRedactor<StarRedactor>();
+        });
+
+        return (serviceCollection.BuildServiceProvider().GetRequiredService<ILoggerFactory>().CreateLogger(nameof(Utils)), fakeLogger.Collector);
```
xakep139, line 53: I don't really get this method. `FakeLogger` is an `ILogger`, right?
Do we really need to return `FakeLogCollector` as a second element in the tuple?
Can't we just return `fakeLogger`?

xakep139:
> Stopped using LogLevel.Debug since it ends up getting filtered
> away by default, and switching the default is a pain.

Why is that? You can add `builder.SetMinimumLevel(LogLevel.Trace)` in `Utils.GetLogger()`

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
```diff
@@ -619,36 +618,12 @@ class MyClass
             internal static partial class C
             {
                 [LogMethod(0, LogLevel.Debug, ""Only {A}"")]
-                static partial void M(ILogger logger, IRedactorProvider provider, [FeedbackData] string a, [LogProperties] MyClass param);
```
xakep139, line 622: As @klauco pointed, we should still support this use-case. Maybe not in 3P though 🤔

test/Generators/Microsoft.Gen.Logging/Generated/Utils.cs
```diff
@@ -0,0 +1,68 @@
...
+            return _logger;
+        }
+
+        public void Dispose()
+        {
+            // nothing
+        }
+    }
+
+    public static (ILogger logger, FakeLogCollector collector) GetLogger(out FakeLogger fakeLogger)
```
xakep139, line 36: Can we return a wrapper of `ILogger`, something like `IDsisposableLogger`?
I would prefer `serviceCollection.BuildServiceProvider()` to be disposed properly.

geeknoid:
> > Stopped using LogLevel.Debug since it ends up getting filtered
> > away by default, and switching the default is a pain.
> 
> Why is that? You can add `builder.SetMinimumLevel(LogLevel.Trace)` in `Utils.GetLogger()`

Oh, let me try that. I was trying to figure out how to change the default minimum level and just didn't land on a solution.

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
geeknoid: Removing redactor provider support was a huge simplification. I'd really need to be convinced that supporting it is a useful thing within the redaction infra.

Note that there's nothing preventing the customer from grabbing an IRedactorProvider themselves and redacting stuff on their own before calling the lower-level log method.

test/Generators/Microsoft.Gen.Logging/Generated/Utils.cs
geeknoid: Cleaned up GetLogger.

test/Generators/Microsoft.Gen.Logging/Generated/Utils.cs
geeknoid: Done.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -276,132 +165,317 @@ static string ConvertPropertyToString(LoggingProperty lp, string arg)
...
+
             foreach (var p in lm.Parameters)
             {
                 if (p.IsException)
                 {
                     exceptionArg = p.Name;
+
+                    if (p.UsedAsTemplate)
+                    {
+                        exceptionLambdaArg = lm.GetParameterNameInTemplate(p);
```
xakep139, line 181: I thought that we emit a warning when the exception is listed in message template

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -276,132 +165,317 @@ static string ConvertPropertyToString(LoggingProperty lp, string arg)
...
+
+                            OutLn($"{stateName}.PropertyArray[{--count}] = new({key}, {value});");
+                        }
+                        else
+                        {
+                            var value = ShouldStringify(p.Type)
+                                ? ConvertToString(p, p.NameWithAt)
+                                : p.NameWithAt;
+
+                            OutLn($"{stateName}.PropertyArray[{--count}] = new({key}, {value});");
```
xakep139, line 269: Can be extracted outside of `if ... else` block

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -276,132 +165,317 @@ static string ConvertPropertyToString(LoggingProperty lp, string arg)
...
-                if (p.IsLogger)
+                OutLn();
+                OutLn($"_ = {stateName}.ReserveClassifiedPropertySpace({numReservedClassifiedProps});");
+                int count = numReservedClassifiedProps;
+                foreach (var p in lm.Parameters)
                 {
-                    logger = p.Name;
-                    isNullable = p.IsNullable;
-                    break;
+                    if (NeedsASlot(p) && p.ClassificationAttributeType != null)
```
xakep139, line 312: Can the check `ClassificationAttributeType != null` be extracted into a get-only property on the type itself?

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -276,132 +165,317 @@ static string ConvertPropertyToString(LoggingProperty lp, string arg)
...
                 }
-            }
 
-            return (logger, isNullable);
-        }
+                foreach (var p in lm.Parameters)
+                {
+                    if (p.HasProperties && !p.SkipNullProperties)
+                    {
+                        p.TraverseParameterPropertiesTransitively((propertyChain, member) =>
```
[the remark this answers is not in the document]
xakep139: I also noticed that these two traverses use array-based assignments, whereas the traverse below uses `AddProperty()` and `AddClassifiedProperty()`. Is there a specific reason for that?

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
geeknoid: Done.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
geeknoid: Done.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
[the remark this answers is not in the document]
xakep139: Oh, I see. Never mind then

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
klauco: can we provide analyzer and fixer for existing customers, to move to the signature which does not use IRedactorProvider?

PR #4227, opened 2023-07-29
Generator cleanup before updating code gen.

geeknoid:
- Remove constraints about not starting log method names or log method parameters with _, the code now dynamically picks non-conflicting symbol names. There's still a potential conflict with the name of redactor instances, but I'll be removing all those in my follow-on PR.

- Simplified logic around handling of template parameters. I couldn't figure out what the code was doing, so it needed a bit of cleanup.

- Remove some dead code I stumbled on.

- Fix broken .resx file handling in the generators, I messed that up when simplifying the Roslyn dependencies last week.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4227)

PR #4235, opened 2023-08-02
Update Emitter.cs

IEvangelist:
The source generator should generate variable names using standard naming conventions that match the rest of the source, for example, only use `_` prefix for fields, not local variables.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4235)

### DIFF
--- a/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
+++ b/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
@@ -219,7 +219,7 @@ internal sealed class Emitter : EmitterBase
         var definePath = restApiMethod.FormatParameters.Count > 0 || !firstQuery;
         if (definePath)
         {
-            OutLn(@$"var _path = {Invariant}($""{pathSb}"");");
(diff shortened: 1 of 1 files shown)


geeknoid:
The use of _ was intentional, this is to avoid conflicts with the user-defined method parameter names. If you look in Parser.cs, you'll see the logic around `Diag(DiagDescriptors.ErrorInvalidParameterName, paramSymbol.Locations[0]);` which disallows parameter names to start with _.

We had the same constraint in the logging generator, but a few days ago I made a change such that we removed the constraint on parameter names starting with _ and instead made the generator smart enough to pick a non-conflicting name relative to the parameters. We should do the same in this generator too. See #4227 for those changes, if you'd like to apply them here too :-)

IEvangelist:
Ah, that makes sense. Let's just close this as it seems like more work that I would have anticipated.

geeknoid:
FYI: https://github.com/dotnet/extensions/pull/4237

PR #4237, opened 2023-08-02
Remove naming constraints in auto-client.

geeknoid:
- The auto-client generator no longer disallows method names and parameter names that start with _. This was previously done to avoid naming conflicts with identifiers created by the generator. Now, the generator dynamically picks new names if there are conflicts with user-defined symbols.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4237)

### DIFF
--- a/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
+++ b/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
@@ -88,6 +88,22 @@ internal sealed class Emitter : EmitterBase
         return pathTemplateSb.ToString();
     }
 
+    private static string PickUniqueName(string baseName, HashSet<string> potentialConflicts)
+    {
+        var name = baseName;
+        while (true)
+        {
+            if (!potentialConflicts.Contains(name))
+            {
+                return name;
+            }
+
+#pragma warning disable S1643 // Strings should not be concatenated using '+' in a loop
+            name += "_";
+#pragma warning restore S1643 // Strings should not be concatenated using '+' in a loop
+        }
+    }
+
     private void GenTypeByNamespace(string nspace, IEnumerable<RestApiType> restApiTypes, CancellationToken cancellationToken)
     {
         OutLn();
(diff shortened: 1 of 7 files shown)


src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -412,10 +435,10 @@ private void EmitClassVariablesAndConstructor(RestApiType restApiType)
 
             if (restApiMethod.FormatParameters.Count == 0 && !restApiMethod.AllParameters.Any(p => p.IsQuery))
             {
-                OutLn(@$"private static readonly {Uri} _uri{requestName} = new(""{restApiMethod.Path}"", {UriKind}.Relative);");
+                OutLn(@$"public static readonly {Uri} Uri{requestName} = new(""{restApiMethod.Path}"", {UriKind}.Relative);");
             }
 
-            OutLn(@$"private static readonly {RequestMetadata} _requestMetadata{requestName} = new()");
+            OutLn(@$"public static readonly {RequestMetadata} RequestMetadata{requestName} = new()");
```
dbarbosapn, line 441: Why is request metadata and Uri "public" now? You want users to be able to change those statically?

dbarbosapn (changes requested):
LGTM, just curious about one thing. Left a comment :)

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
geeknoid: All of the statics are not in a nested static class which is private, so customers still can't get to that stuff.

The reason to put all the statics into a class was to reduce "naming pollution" of the main class. The names in the nested types don't have any risk of conflicting with names of things in the main type being generated.

dbarbosapn: Ohh... Got it. It's so hard to read code written by other code...
I think we should keep on the repo the source-generated file somehow... It would help a ton to review.

Thanks!

dbarbosapn (approved):
👍

Issue #4207, opened 2023-07-25
Provide ability to replace the default logging code-gen


klauco:
Implement MSBuild magic such that when the new logging code generator is used, it automatically disables the dotnet/runtime generator.

joperezr:
Closed via #4238

PR #4238, opened 2023-08-02
Replace platform's Logging source generator

joperezr:
cc: @geeknoid @RussKie @ViktorHofer @tarekgh 

This PR will make sure that when Microsoft.Extensions.Telemetry.Abstractions package gets referenced, the import of the Microsoft.Extensions.Logging.Abstractions package is disabled as the new referenced package's source generator should take precedence.

For additional context, the package Microsoft.Extensions.Logging.Abstractions has a built-in targets file which will check if the property `DisableMicrosoftExtensionsLoggingSourceGenerator` is set to true, and if so, it will disable the automatic reference of the logging generator included in that package. The changes in this PR are just setting that value when the extensions Telemetry package gets referenced, in order to replace the logging generator with the extensions version.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4238)


joperezr:
@ViktorHofer this currently doesn't work when targeting netcoreapp3.1, because currently the logging.abstractions package has a different buildTransitive .targets file for  that TFM which just shows the not supported warning instead of having the mechanism that the rest of the TFMs have for disabling the source generator:

image

You would need a workaround similar than the one I'm introducing here which allows for extra content on that generated file in order for this to work.

geeknoid (commented):
Terrific!

geeknoid (approved):

ViktorHofer:
> @ViktorHofer this currently doesn't work when targeting netcoreapp3.1, because currently the logging.abstractions package has a different buildTransitive .targets file for that TFM which just shows the not supported warning instead of having the mechanism that the rest of the TFMs have for disabling the source generator

This is by design as packages produced by dotnet/runtime don't support netcoreapp3.1. The runtime has been out-of-support since December 2022. We communicate this via the Warning task that we emit. cc @ericstj

xakep139:
@joperezr does this PR resolves #4207?
If so, could you please link these two?

joperezr:
Yes, it does. 🙂

PR #4247, opened 2023-08-07
Dynamically calculating the paths to the analyzers to use in MergeAnalyzerMetadata script

joperezr:
cc @geeknoid please make sure this does what you wanted as I haven't fully tested this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4247)

### DIFF
--- a/scripts/MergeAnalyzerMetadata.ps1
+++ b/scripts/MergeAnalyzerMetadata.ps1
@@ -48,20 +48,20 @@ try {
 
     Copy-Item -Path $DiagToolPath -Destination $tempDir
 
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.Analyzers.Extra.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.Analyzers.Local.dll
-    & .\DiagConfig.exe $Diags analyzer merge StyleCop.Analyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge SonarAnalyzer.CSharp.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.VisualStudio.Threading.Analyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.VisualStudio.Threading.Analyzers.CSharp.dll
-    & .\DiagConfig.exe $Diags analyzer merge xunit.analyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.CodeStyle.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.CSharp.CodeStyle.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.NetAnalyzers.dll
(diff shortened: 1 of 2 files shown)


geeknoid (approved):

scripts/MergeAnalyzerMetadata.ps1
```diff
@@ -75,20 +48,20 @@ try {
...
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.VisualStudio.Threading.Analyzers.CSharp.dll
-    & .\DiagConfig.exe $Diags analyzer merge xunit.analyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.CodeStyle.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.CSharp.CodeStyle.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.NetAnalyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.CodeAnalysis.CSharp.NetAnalyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge ILlink.RoslynAnalyzer.dll
-#    & .\DiagConfig.exe $Diags analyzer merge Microsoft.AspNetCore.App.Analyzers.dll
-    & .\DiagConfig.exe $Diags analyzer merge Microsoft.AspNetCore.Components.Analyzers.dll
+    & dotnet exec .\DiagConfig.dll $Diags analyzer merge Microsoft.Analyzers.Extra.dll
```
geeknoid, line 51: Why change this stuff here?

joperezr: When calling it via `.\DiagConfig.exe` the project will always try to use the globally installed dotnet (meaning the one in `C:\Program Files\dotnet\`, but that won't always be the same version of the SDK that was used to build the tool. Since we are building the tool a few lines above using `dotnet` (whichever path wins for that), here I'm changing it so that we also just use `dotnet` for the execution so we can guarantee that it uses the same SDK for running than the one it used for building.

I hit this locally because I didn't have the version listed in global.json, so I temporarily added the local path of the restored dotnet tool to my path by running `$env:PATH="C:\Users\joperezr\git\extensions\.dotnet;" + $env:PATH`. After doing this, running the script fails locally as I'm building the tool using the locally restored SDK, but it is trying to run with the globally installed one which doesn't match. After my changes things work as expected.

geeknoid: I guess I'll need to do the same thing to the MakeEditorConfigs.ps1 and MakeApiBaselines.ps1...

joperezr: It would surely help making things consistent 😃

RussKie: This should really be:
```suggestion
    & ./.dotnet/dotnet exec ./DiagConfig.dll $Diags analyzer merge Microsoft.Analyzers.Extra.dll
```

Note, all paths must be in POSIX so the script is runnable on Linux as well.

PR #4253, opened 2023-08-08
Leverage KeyedServices feature in HttpClient logging

xakep139:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4253)


test/Libraries/Microsoft.Extensions.Http.Telemetry.Tests/Logging/HttpClientLoggerTest.cs
```diff
@@ -30,6 +30,7 @@
 
 namespace Microsoft.Extensions.Http.Telemetry.Logging.Test;
 
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "R9A061:The async method doesn't support cancellation", Justification = "Temp")]
```
xakep139, line 33:
```suggestion
```

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
```diff
@@ -36,20 +36,32 @@ internal sealed class HttpClientLogger : IHttpClientAsyncLogger
     private IHttpClientLogEnricher[] _enrichers;
 
     public HttpClientLogger(
+        ILogger<HttpClientLogger> logger,
+        IEnumerable<IHttpClientLogEnricher> enrichers,
+        IOptionsMonitor<LoggingOptions> optionsMonitor,
+        IServiceProvider serviceProvider,
```
xakep139, line 42: Even though I'm not a fan of service locator antipattern, .NET folks didn't suggest anything better: https://github.com/dotnet/runtime/issues/64427#issuecomment-1665940109

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Http.Telemetry.Logging.Internal;
+
+internal static class ServiceProviderExtensions
+{
+    public static T GetRequiredKeyedServiceOrDefault<T>(this IServiceProvider serviceProvider, string? serviceKey)
```
RussKie, line 11: "Default" usually implies "default value of T" (which is typically `null` or `0`). This API appears to be retrieving a _required keyed service if the key is available_, alternatively the return value is a _required service_. `GetRequiredService` should never return defaults, it should return a service or fail....

I can't really of a good name, but I think `GetRequiredKeyedService` is a better name than this because the "OrDefault" suffix is misleading.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
RussKie: I don't agree that injecting `IServiceProvider` is the "service locator antipattern", this is the normal DI. The "service locator" pattern was defined by Martin Fowler in https://martinfowler.com/articles/injection.html, and then debunked many times since (e.g., by Mark Seemann [in his blog](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/)).

My past experience tells me that it is kind of customary to put `IServiceProvider serviceProvider` as the first parameter. But I don't have any specific guidelines to back that up.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/ServiceProviderExtensions.cs
xakep139: `GetRequiredKeyedService` method is already defined in runtime, so I can't use it.
Please also bear in mind that it's not a public API.

xakep139: In fact, the method semantic is `GetRequiredKeyedServiceOrNonKeyedOtherwise` which I personally found a bit too long and clumsy.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
xakep139: I'm referring to this technique as a service locator pattern mainly because of https://learn.microsoft.com/en-us/previous-versions/msp-n-p/hh860410(v=pandp.51).
It's also considered a good practice to have all dependencies listed explicitly - in a constructor in that particular case.
Strictly speaking - yes, it's not a static service locator described by Martin Fowler

xakep139: I will put `serviceProvider` as a first argument, as you suggested

xakep139: You can also check [Dependency Injection Principles, Practices, and Patterns](https://learning.oreilly.com/library/view/dependency-injection-principles/9781617294730/), specifically chapter 4.1 where it's described that using DI container outside of the composition root is essentially a service locator pattern anyway.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/ServiceProviderExtensions.cs
xakep139: I'm all for finding a good name for the method, so if you have any other ideas - please share.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
RussKie: > It's also considered a good practice to have all dependencies listed explicitly - in a constructor in that particular case.

Yes, however there's a price to pay when passing multiple parameters in a constructor (any method).

In a general case I strongly prefer for all dependencies to be resolved in a constructor, but whether those dependencies are provided explicitly or come from a IServiceProvider I don't have a strong preference.

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/ServiceProviderExtensions.cs
RussKie: I find the more verbose name (e.g., `GetRequiredKeyedServiceOrNonKeyedOtherwise`) better than the "...OrDefault". Especially, since it's a non-public method the length is not a big issue.
What do you think of `GetRequiredOrKeyedService`?

xakep139: Sounds good!

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/HttpClientLogger.cs
xakep139: Done

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/ServiceProviderExtensions.cs
xakep139: Renamed

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/OptionsExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.Http.Telemetry.Logging.Internal;
+
+internal static class OptionsExtensions
+{
+    public static TOptions GetKeyedOrCurrent<TOptions>(this IOptionsMonitor<TOptions> options, string? serviceKey)
```
RussKie, line 10: [nit] It'd be good to have xml-docs here too.

RussKie (approved):

src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/Internal/OptionsExtensions.cs
xakep139: Will add

xakep139: Done

PR #4254, opened 2023-08-08
Documentation transport package

carlossanlop:
This will be based on the runtime version: https://github.com/dotnet/runtime/pull/89312

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4254)


Directory.Build.props
```diff
@@ -114,6 +114,10 @@
     <UseVSTestRunner>true</UseVSTestRunner>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <LibrariesProjectRoot>$([MSBuild]::NormalizeDirectory('$(MSBuildThisFileDirectory)', 'src', 'Libraries'))</LibrariesProjectRoot>
```
carlossanlop, line 118: I didn't find a property that defined the Libraries folder. I figured I'll need it at least once, so I created it here. Was this the right location?

eng/MSBuild/Packaging.props
```diff
@@ -4,7 +4,7 @@
     <SignArtifacts>$(SignAssembly)</SignArtifacts>
 
     <!-- Arcade SDK marks all but *.pkgproj as non-packable, we need to override. -->
-    <IsPackable>true</IsPackable>
+    <IsPackable Condition="'$(IsPackable)' == ''">true</IsPackable>
```
carlossanlop, line 7: I'm noticing in the msbuild.binlog that IsPackable is showing up as false for the transport.proj. I am unsure why, and would like to get some pointers. Not sure if this is the correct way to address it.

eng/MSBuild/Packaging.targets
```diff
@@ -9,7 +9,7 @@
 
   <!-- packaging control -->
   <PropertyGroup>
-    <IsPackable Condition="'$(Stage)' == ''">false</IsPackable>
+    <IsPackable Condition="'$(Stage)' == '' and '$(IsPackable)' == ''">false</IsPackable>
```
carlossanlop, line 12: Same as above, I was trying to only change its value if it was previously unset, but I am unsure if this is the right choice.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+
+    <PropertyGroup>
+        <TargetFramework>$(LatestTargetFramework)</TargetFramework>
```
carlossanlop, line 4: LatestTargetFramework seems to be the value to use here. In runtime, we also add the target OS at the end, but it does not seem to apply here.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
...
+        <TargetFrameworks>$(TargetFrameworks)</TargetFrameworks>
```
carlossanlop, line 5: Adding the `TargetFrameworks` property was my failed attempt to fix the error I'm getting:

```
C:\Program Files\dotnet\sdk\8.0.100-preview.7.23376.3\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDe
pendencyResolution.targets(266,5): error NETSDK1005: Assets file 'C:\Users\calope\source\repos\extension 
s\artifacts\obj\Microsoft.Internal.Extensions.DotNetApiDocs.Transport\project.assets.json' doesn't have  
a target for 'net8.0'. Ensure that restore has run and that you have included 'net8.0' in the TargetFram 
eworks for your project. [C:\Users\calope\source\repos\extensions\src\Packages\Microsoft.Internal.Extens 
ions.DotNetApiDocs.Transport\Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj::TargetFramework 
=net8.0]
```

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
...
+        <IsShipping>false</IsShipping>
+        <!-- Reference the outputs to have them available as build outputs. -->
+        <NoTargetsDoNotReferenceOutputAssemblies>false</NoTargetsDoNotReferenceOutputAssemblies>
+        <IsPackable>true</IsPackable>
```
carlossanlop, line 9: This is being set to false, not sure why.

carlossanlop (commented):
@ViktorHofer @ericstj would you mind looking at my questions?

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
RussKie: We multi-target (currently: net462,netcoreapp3.1,net8.0), so we don't use `TargetFramework` and use `TargetFrameworks` instead.

`LatestTargetFramework` points currently to the `net8.0`, soon it'll be pointing to net9.0. Though, the projects themselves will likely retain 'net462,netcoreapp3.1,net8.0', and only projects that require .NET 9-specific API will be targeting `LatestTargetFramework` (or `net9.0` explicitly). This is something we haven't thought through yet, but we'll have to do so very soon :) /cc: @joperezr

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
RussKie: It's the Arcade's default: https://github.com/dotnet/arcade/blob/5cd8dc258e8177b8542adf7abe45e7cfc1c1b1c0/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.props#L25-L26

Directory.Build.props
ericstj: Seems reasonable.  I see a few projects that might benefit from using such a property: https://github.com/search?q=repo%3Adotnet%2Fextensions+Libraries+language%3AXML&type=code&l=XML

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
ericstj: It looks to me like this repo doesn't have projects define `IsPackable` but instead has them define `Stage` : https://github.com/search?q=repo%3Adotnet%2Fextensions+Stage+language%3AXML&type=code&l=XML and the packing state is inferred from that.

Probably you want to decide if you should participate in that by setting some value for `Stage` or if you wanted to create a way to opt-out of it.  Today it doesn't look like there's a great way to opt-out since the targets will assume that anything that doesn't set `Stage` is not packable.

My suggestion would be to remove the default value of True from `Packaging.props`, allow projects to specify either `IsPackable` or `Stage` then in `Packaging.targets` set `IsPackable` to `true` if `Stage` is set.

eng/MSBuild/Packaging.targets
ericstj: Probably needs a discussion with repo owners about how IsPackable/Stage interact.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
ericstj: You shouldn't set both `TargetFramework` and `TargetFrameworks`.  Since this project is using P2P references to gather output from other projects you should make a choice that's based on what you want to include.

If you want to include 1 set of files for single framework, then set TargetFramework.

If you want to include multiple sets of files for all frameworks, then set TargetFrameworks and make sure it's a superset of all the frameworks targeted by projects referenced.

ViktorHofer: We just need the LatestTargetFramework (= net8.0) for the docs transport package.

carlossanlop:
After a long paired coding session with @ericstj, this is now working (Eric pushed a commit fixing various things), so I marked it ready for review. @ViktorHofer @RussKie would you mind taking a look again?

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
RussKie: I defer this to @rafal-mz @geeknoid for decision and guidance.

Directory.Build.props
RussKie: At some point there was a desire to keep MSBuild-related stuff in https://github.com/dotnet/extensions/tree/main/eng/MSBuild, but something's gone wrong :)
This looks similar to the [Arcade's doing](https://github.com/dotnet/arcade/blob/bebd2de1988036b47ecefc7e72b66ce2f403e43d/src/Microsoft.DotNet.Arcade.Sdk/tools/RepoLayout.props#L48) and I would align the naming with the Arcade:
```suggestion
    <!-- Similar to https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/RepoLayout.props#L48 -->
    <SrcLibrariesDir>$([MSBuild]::NormalizeDirectory('$(MSBuildThisFileDirectory)', 'src', 'Libraries'))</SrcLibrariesDir>
```

RussKie:
@geeknoid @rafal-mz could you please help with the failure? This PR isn't touching Microsoft.Extensions.Compliance.Abstractions, yet it's failing because of it...

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+
+    <PropertyGroup>
+        <TargetFramework>$(LatestTargetFramework)</TargetFramework>
+        <!-- Workaround https://github.com/microsoft/MSBuildSdks/issues/469 -->
+        <TargetFrameworks />
```
ViktorHofer, line 6: While you link to a workaround, I find unsetting the `TargetFrameworks` property the right to do anyway. It makes the build slightly faster as the outer build doesn't need to be evaluated and ran.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
...
+        <PackageDescription>Internal transport package to provide dotnet-api-docs with the reference assemblies and compiler generated documentation files from dotnet/extensions.</PackageDescription>
+        <NoWarn>$(NoWarn);NU5128;NU5131</NoWarn>
+        <BuildProjectReferences Condition="'$(NoBuild)' == 'true'">false</BuildProjectReferences>
+        <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludeProjectReferenceCompilerGeneratedSecondaryOutputInPackage</TargetsForTfmSpecificContentInPackage>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <!-- Set PrivateAssets to avoid dependencies in the nupkg
+             Set Private=false to avoid copying the referenced projects to output -->
+        <ProjectReference Include="$(LibrariesProjectRoot)\*\*.*proj;$(LibrariesProjectRoot)\..\ToBeMoved\*\*.*proj;$(LibrariesProjectRoot)\..\ToBeRemoved\*\*.*proj" PrivateAssets="all" Private="false" />
```
ViktorHofer, line 24: Please add a new line after each `;` charater. That makes it easier to read the items.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
...
+        <EnablePackageValidation>false</EnablePackageValidation>
+        <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+        <PackageDescription>Internal transport package to provide dotnet-api-docs with the reference assemblies and compiler generated documentation files from dotnet/extensions.</PackageDescription>
+        <NoWarn>$(NoWarn);NU5128;NU5131</NoWarn>
+        <BuildProjectReferences Condition="'$(NoBuild)' == 'true'">false</BuildProjectReferences>
+        <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludeProjectReferenceCompilerGeneratedSecondaryOutputInPackage</TargetsForTfmSpecificContentInPackage>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <!-- Set PrivateAssets to avoid dependencies in the nupkg
```
ViktorHofer, line 22: Strictly speaking you don't need PrivateAssets here as you already set `SuppressDependenciesWhenPacking` which emptires the package dependency nodes.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,41 @@
...
+        DependsOnTargets="BuildOnlySettings;ResolveReferences">
+        <ItemGroup>
+            <_projectReferencesToInclude Include="@(ReferencePath->WithMetadataValue('ReferenceSourceTarget', 'ProjectReference'))" />
+            <_referenceAssemblyFile Include="@(_projectReferencesToInclude->Metadata('ReferenceAssembly'))" />
+            <!-- Locate the XML next to the ReferencePath -->
+            <_referenceXmlDocFile Include="@(_projectReferencesToInclude->'%(RootDir)%(Directory)%(Filename).xml')" />
+            <TfmSpecificPackageFile Include="@(_referenceAssemblyFile);@(_referenceXmlDocFile)"
+                PackagePath="ref\$(LatestTargetFramework)\" />
+        </ItemGroup>
+
```
ViktorHofer, line 38: nit:

```suggestion
```

eng/MSBuild/Packaging.props
```diff
@@ -4,7 +4,7 @@
     <SignArtifacts>$(SignAssembly)</SignArtifacts>
 
     <!-- Arcade SDK marks all but *.pkgproj as non-packable, we need to override. -->
```
ViktorHofer, line 6: This comment should be updated as dotnet/extensions doesn't use pkgprojs. cc @RussKie

eng/MSBuild/Packaging.props
ViktorHofer: I think you should undo this change. It's not necessary for your change.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
RussKie: I totally missed this before. I believe we only need to include `src\**\*.*proj`. As the names suggest `ToBeMoved` and `ToBeRemoved` aren't here for long :)

eng/MSBuild/Packaging.props
RussKie: Could you please help me understand why it needs an update? The comment states that all projects except *.pkgproj are non-packable by default.

ViktorHofer: This comment comes from dotnet/runtime which used pkgprojs. This repositories doesn't use pkgprojs so it doesn't make sense to mention them in this repository at all.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
ericstj: When Carlos and I were trying this we noticed that the projects in ToBeMoved and ToBeRemoved would be included transitively by others, but would be missing the `Private="false"` marking and be copied into the output folder unnecessarily.  This was more of just a placeholder to spark discussion -- should transitive projects be included in this transport package if they weren't directly referenced?  Do folks care about copying the references into the output folder of Microsoft.Internal.Extensions.DotNetApiDocs.Transport?

eng/MSBuild/Packaging.props
ericstj: Per my other comment I think you should just keep arcade's default of `false` here and have packages opt-in by setting it themselves, or setting `Stage`.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
ViktorHofer: We don't care about the output folder of the transport package, it's more or less just an intermediate. We care about the package layout though. We require all shipping reference assemblies (or the implementation assembly, if dotnet/extensions doesn't ship reference assemblies) and their documentation files to be present.

Btw, @carlossanlop please make sure that the transport package doesn't include non-shipping intermediate reference assemblies that never reach customers. You might be doing that accidentally by using the `ReferenceAssembly` item metadata from the ProjectReferences.

ericstj:
> @geeknoid @rafal-mz could you please help with the failure? This PR isn't touching Microsoft.Extensions.Compliance.Abstractions, yet it's failing because of it...

Something about the way this project builds the references is causing the warnings to show up.  We didn't see them when building from the root but did when building project references.  Probably the thing to look at would be how these warnings don't happen during the regular build then find out why that wouldn't happen in this case.

joperezr:
While I'm not against this change, I'd like to understand a bit more why is it necessary. If I understand correctly, the usual process for api docs (especially for the ones that get built via NuGet packages) is just to submit a work item to the docs team with the collection of packages, and they have a system that can extract the xmls from there and produce the content of the website. Since all we ship out of this repo are packages, why does that not  work?

Again, I have absolutely nothing against this change (and thanks for taking care of it @carlossanlop! 😃 ), I'm only trying to understand why we need it for educational purposes.

carlossanlop:
Forgot to mention that we will need a similar change for wcf, wpf, winforms. @ViktorHofer is working on the wcf one.

@dreddy-work @JeremyKuhne who can help on the wpf and winforms side with a similar task?

eng/MSBuild/Packaging.props
joperezr: Yup, the change you are making is basically only respecting the original value in case it was already set, but if you are seeing false in the binlog, then that only means it is getting overridden later, so we should just keep this as it was.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
joperezr: Maybe not now, but in the future I do envision having APIs which are not present in latest target framework but are present in down level TFMs, so ideally the transport package would have docs per TFM. That is kind of why I asked the overall question before of why can't we just use the process that the docs team already supports where you give them a NuGet package per library, and they are able to extract the docs per TFM for each of them.

ViktorHofer: > why can't we just use the process that the docs team already supports where you give them a NuGet package per library, and they are able to extract the docs per TFM for each of them.

Please elaborate, I'm not sure which process you are referring to. The current model, as @carlossanlop described below, is manual and AFAIK only considers the latest TFM. There is nothing that prevents the transport package from including other TFMs. We could even do that as part of this PR today but the docs team would need to agree on indexing more than just the current TFM.

In general, the docs team wants a single transport package per repository, including the reference assemblies and the compiler generated XML files. We aim towards creating these packages before .NET 8 RC2. In the future, we could/should consider adding centralized infrastructure (presumably in Arcade) to automatically create transport packages based on repository information.

joperezr: Sure. So after chatting with @gewarren a few months ago, we were instructed that the process for pushing api Reference docs to the docs.microsoft.com, we would simply need to create a new workitem here: https://aka.ms/publish-on-docs/reference outlining the packages that need to be updated, versions and the rest of the info that is asked in that template. From there, the Docs team would restore those packages locally, extract the xmls from each of our packages, and then generate the content in the format then need it in in order to publish it to the docs site. We were also told that we would have to repeat this process each time we wanted to push updates for the docs whenever there was a new version of the packages.

All of this process is documented [here](https://review.learn.microsoft.com/en-us/help/onboard/dotnet-requirements?branch=main&branchFallbackFrom=master). The nice thing about this process is that it doesn't really require for us to have to generate anything special, and instead we could just share the NuGet packages and that would be sufficient on their side. Worth noting that this is also the process that we have been following from the dotnet/iot repository for the past few years and it has worked without any issues. Again, I understand why dotnet/runtime's process is more convoluted and hard, and why it requires a special transport package to be created, but what I don't fully understand is why is this required for the extensions repo which is much simpler and basically only ships individual nuget packages.

ViktorHofer: How does this correlate with the api-catalog-infra that today needs to index the extensions packages by manually specifying them here? https://github.com/dotnet/api-catalog-infra/blob/20eadd453545ab028052662436288f4bd2a20f3c/eng/droplayout.props#L149-L172

@carlossanlop

joperezr: I'm not 100% sure (@gewarren or @RussKie  can probably guide us here) but I believe that is a separate step, so in general you submit a workitem for the docs team specifying the packages, and then if the package is new you just add a line there so that the package is also included.

gewarren: I didn't know you had to add an entry to api-catalog-infra if you're generating docs via https://aka.ms/publish-on-docs/reference requests. But anything that ends up in the "extensions" drop when api-catalog-infra is run will be ingested for API doc generation to the https://github.com/dotnet/dotnet-api-docs repo as part of each .NET release.

joperezr: I see, so I guess the uber question is: Is there benefit to having these transport packages being produced from the extensions repo too, or would it be the same thing if we just continue to push our NuGet packages through aka.ms/publish-on-docs/reference requests? @gewarren can you help identify the pros and cons so we can decide what the best approach is for dotnet-extensions?

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
RussKie: > When Carlos and I were trying this we noticed that the projects in ToBeMoved and ToBeRemoved would be included transitively by others, but would be missing the `Private="false"` marking and be copied into the output folder unnecessarily. This was more of just a placeholder to spark discussion -- should transitive projects be included in this transport package if they weren't directly referenced? Do folks care about copying the references into the output folder of Microsoft.Internal.Extensions.DotNetApiDocs.Transport?

That makes sense. I suppose the build will fail once we removed both of these folders, and we'll just adjust the config.
/resolved

eng/MSBuild/Packaging.props
RussKie: > This comment comes from dotnet/runtime which used pkgprojs. This repositories doesn't use pkgprojs so it doesn't make sense to mention them in this repository at all.

Nope :) This comment reflects the [Arcade's configuration](https://github.com/dotnet/arcade/blob/5cd8dc258e8177b8542adf7abe45e7cfc1c1b1c0/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.props#L25-L26). 
image

I added it on purpose to help with questions like https://github.com/dotnet/extensions/pull/4254#discussion_r1287757994.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,49 @@
...
+  <!-- Include the reference assembly and put the documentation file next to it. -->
+  <Target Name="IncludeProjectReferenceCompilerGeneratedSecondaryOutputInPackage"
+      DependsOnTargets="BuildOnlySettings;ResolveReferences">
+    <ItemGroup>
+      <_projectReferencesToInclude Include="@(ReferencePath->WithMetadataValue('ReferenceSourceTarget', 'ProjectReference'))" />
+      <_referenceAssemblyFile Include="@(_projectReferencesToInclude->Metadata('ReferenceAssembly'))" />
+
+      <!-- Locate the XML next to the ReferencePath -->
+      <_referenceXmlDocFile Include="@(_projectReferencesToInclude->'%(RootDir)%(Directory)%(Filename).xml')" />
+      <TfmSpecificPackageFile Include="@(_referenceAssemblyFile);@(_referenceXmlDocFile)" PackagePath="ref\$(LatestTargetFramework)\" />
```
RussKie, line 44: We need to collect refs for each TFM we build because we may have API differences between TFMs.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
RussKie: I added a new stage called "transport" and marked this project with this stage.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
gewarren: I think the reason we went with api-catalog-infra is because we wanted the docs in the same repo as the other .NET extensions. There's an issue if you have them in a different repo and there are overlapping namespaces (although not sure if that's the case). Then there's also monikers to consider. @v-alje knows more about the pros/cons than me.

v-alje: @gewarren @joperezr Looking at the code in the linked pull request <https://github.com/dotnet/runtime/pull/89312>, it looks like these extensions share namespaces with .NET API reference content currently in the <https://github.com/dotnet/dotnet-api-docs> repo. You can look that up by entering the namespace in the Search box on this page: <https://learn.microsoft.com/en-us/dotnet/api/?branch=main>.  

Libraries and packages which share a namespace need to be processed in the same repo and under the same CI job otherwise one namespace page will take precedence over all the others and the types from the latter will not get linked to in the namespace page.  

So, these extensions need to be processed in the <https://github.com/dotnet/dotnet-api-docs> repo.

joperezr: Gotcha, thanks @v-alje  so much for the explanation and apologies for the questioning, just wanted to make sure that the changes in this PR were indeed required which now I understand they are.

geeknoid:
What's happening with this PR? Can we get it in?

joperezr:
We totally should. IIRC the last remaining issue was that this PR makes it so just the .NET 8 surface area is used for grabbing documentaiton, as opposed to having different docs for different TFMs, but that said, I think we should still get this in as a starting point. @carlossanlop is there anything remaining work here? If not, Can we go ahead and merge this please?

carlossanlop:
> is there anything remaining work here?

It's not ready. I had to focus on other higher pri things in the last weeks. Let me retake this effort this week and I'll request your review.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
carlossanlop: We talked about this with @gewarren and we only need the latest tfm, for now.

ViktorHofer (approved):

carlossanlop:
I had a chat with @ViktorHofer today and I added two changes:
- Made sure the transport *.proj file is explicitly included in _ProjectsToBuild in eng/build.proj.
- Moved the PackageDescription a few lines below to ensure TargetFramework is at the top always.

I also noticed that ProjectReference in the transport package was including all *.*proj, so I explicitly excluded itself.

When building this proj from its folder using `dotnet build`, I noticed there are two roslyn warnings showing up:
- `IDE0028: Collection initialization can be simplified`.
- `NETSDK1212: IsTrimmable is not supported for the target framework. Consider multi-targeting to a supported framework to enable trimming, and set IsTrimmable only for the supported frameworks`.

Should I ignore them in the transport proj using `<NoWarn>` or do we not care?

Thanks @ViktorHofer for the sign-off.

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,49 @@
...
+
+  <ItemGroup>
+    <!-- Set PrivateAssets to avoid dependencies in the nupkg
+         Set Private=false to avoid copying the referenced projects to output -->
+    <ProjectReference PrivateAssets="all"
+      Private="false"
+      Include="$(SrcLibrariesDir)\*\*.*proj;
+               $(SrcLibrariesDir)\..\ToBeMoved\*\*.*proj;
+               $(SrcLibrariesDir)\..\ToBeRemoved\*\*.*proj"
+      Exclude="$(SrcLibrariesDir)\Packages\Microsoft.Internal.Extensions.DotNetApiDocs.Transport\Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj"
```
ViktorHofer, line 31:
```suggestion
      Exclude="$(MSBuildProjectFullPath)"
```

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,49 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Set PrivateAssets to avoid dependencies in the nupkg
+         Set Private=false to avoid copying the referenced projects to output -->
+    <ProjectReference PrivateAssets="all"
+      Private="false"
+      Include="$(SrcLibrariesDir)\*\*.*proj;
+               $(SrcLibrariesDir)\..\ToBeMoved\*\*.*proj;
+               $(SrcLibrariesDir)\..\ToBeRemoved\*\*.*proj"
```
RussKie, line 30: [nit] this should have been removed

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
```diff
@@ -0,0 +1,49 @@
...
+      Exclude="$(MSBuildProjectFullPath)"
+                      />
```
RussKie, line 32: [nit] misindented

RussKie (commented):
Thank you!

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
carlossanlop: Addressed in https://github.com/dotnet/extensions/pull/4562

src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj
carlossanlop: Addressed in https://github.com/dotnet/extensions/pull/4562

PR #4258, opened 2023-08-09
API Review for M.E.TimeProvider.Testing

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4258)

### DIFF
--- a/API.Microsoft.Extensions.TimeProvider.Testing/WithDocs/Microsoft.Extensions.Time.Testing/FakeTimeProvider.cs
+++ b/API.Microsoft.Extensions.TimeProvider.Testing/WithDocs/Microsoft.Extensions.Time.Testing/FakeTimeProvider.cs
@@ -1,0 +1,97 @@
+// Assembly 'Microsoft.Extensions.TimeProvider.Testing'
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Microsoft.Extensions.Time.Testing;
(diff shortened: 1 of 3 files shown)


API.Microsoft.Extensions.TimeProvider.Testing/WithDocs/Microsoft.Extensions.Time.Testing/FakeTimeProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// Advances time by a specific amount.
+    /// </summary>
+    /// <param name="delta">The amount of time to advance the clock by.</param>
+    /// <remarks>
+    /// Advancing time affects the timers created from this provider, and all other operations that are directly or
+    /// indirectly using this provider as a time source. Whereas when using <see cref="P:System.TimeProvider.System" />, time
+    /// marches forward automatically in hardware, for the fake time provider the application is responsible for
+    /// doing this explicitly by calling this method.
+    /// </remarks>
+    public void Advance(TimeSpan delta);
```
mddddb, line 78: Would it make sense to return the new value of Now here?

geeknoid: I looked through all existing uses of Advance, and I didn't see anywhere that was reading the time after advancing the clock. So I'd say no, I don't think this would be that useful.

PR #4260, opened 2023-08-09
Add missing option validation

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4260)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Latency/Registration/LatencyRegistryExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Latency/Registration/LatencyRegistryExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options.Validation;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.Telemetry.Latency;
@@ -79,7 +80,7 @@ public static class LatencyRegistryExtensions
 
     private static void ConfigureOption(this IServiceCollection services, Action<LatencyContextRegistrationOptions> action)
     {
-        _ = services.AddOptions<LatencyContextRegistrationOptions>();
+        _ = services.AddValidatedOptions<LatencyContextRegistrationOptions, LatencyContextRegistrationOptionsValidator>();
         _ = services.Configure(action);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Latency/Registration/LatencyRegistryExtensions.cs
```diff
@@ -79,7 +80,7 @@ private static void CheckNames(string[] names)
 
     private static void ConfigureOption(this IServiceCollection services, Action<LatencyContextRegistrationOptions> action)
     {
-        _ = services.AddOptions<LatencyContextRegistrationOptions>();
+        _ = services.AddValidatedOptions<LatencyContextRegistrationOptions, LatencyContextRegistrationOptionsValidator>();
```
xakep139, line 83: Do we need the options type to be public at all?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Microsoft.Extensions.Telemetry.Abstractions.csproj
```diff
@@ -19,6 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\ToBeRemoved\Options.ValidateOnStart\Options.ValidateOnStart.csproj" />
```
joperezr, line 22: I was under the impression that we were able to consume the .NET 8 version of this, is that not the case?

xakep139: Good point

geeknoid: Not yet, it's the last bit of ToBeRemoved functionality left. I think the PR went into dotnet/runtime yesterday or this morning, so this should be removed soon....

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Latency/Registration/LatencyRegistryExtensions.cs
geeknoid: I asked Ananth for feedback on that. Maybe we should hide it, we'll see what he ssys.

geeknoid: He's OOF, so it'll take a while. We can address this in a separate PR.

joperezr (approved):

PR #4277, opened 2023-08-11
Fix potential infinite loop in FakeTimeProvider when a timer callback changes the current time

### LINKED ISSUE #4275: Changing FakeTimeProvider's current time during a timer callback can lead to lack of forward progress
If a user has a timer with a duetime/period that is less than or equal to whatever `AutoAdvanceAmount` is set to, the fake time provider will go into an endless loop. 

We discussed the pros/cons of having the feature in there already during the PR where `AutoAdvanceAmount`, but since the feature is going to stay this should be highlighted in the documentation.

For example, the following test will never reach the assertion statement.

```c#
[Fact]
public void Timer_callback_GetUtcNow_AutoAdvance()
{
    var oneSecond = TimeSpan.FromSeconds(1);
    var timeProvider = new FakeTimeProvider
    {
        AutoAdvanceAmount = oneSecond
    };

    var timer = timeProvider.CreateTimer(_ =>
    {
        timeProvider.GetUtcNow();
    }, null, TimeSpan.Zero, oneSecond);

    Assert.Fail("This is never reached.");
}
```

Ps. would still very much like to see examples where `AutoAdvanceAmount` is used to understand when it is useful.

geeknoid:
Fixes #4275

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4277)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
+++ b/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
@@ -18,7 +18,7 @@ public class FakeTimeProvider : TimeProvider
     internal readonly HashSet<Waiter> Waiters = new();
     private DateTimeOffset _now = new(2000, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
     private TimeZoneInfo _localTimeZone = TimeZoneInfo.Utc;
-    private int _wakeWaitersGate;
+    private volatile int _wakeWaitersGate;
     private TimeSpan _autoAdvanceAmount;
 
     /// <summary>
@@ -59,6 +59,7 @@ public class FakeTimeProvider : TimeProvider
     /// <remarks>
     /// This defaults to <see cref="TimeSpan.Zero"/>.
     /// </remarks>
+    /// <exception cref="ArgumentOutOfRangeException">if the time value is set to less than <see cref="TimeSpan.Zero"/>.</exception>
     public TimeSpan AutoAdvanceAmount
     {
         get => _autoAdvanceAmount;
@@ -88,6 +89,7 @@ public class FakeTimeProvider : TimeProvider
     /// Sets the date and time in the UTC time zone.
     /// </summary>
     /// <param name="value">The date and time in the UTC time zone.</param>
+    /// <exception cref="ArgumentOutOfRangeException">if the supplied time value is before the curent time.</exception>
     public void SetUtcNow(DateTimeOffset value)
     {
         lock (Waiters)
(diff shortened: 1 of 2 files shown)


sebastienros (approved):

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -240,15 +243,29 @@ private void WakeWaiters()
...
+                if (oldTicks != newTicks)
+                {
+                    // time changed while in the callback, readjust the wake time accordingly
+                    candidate.WakeupTime = newTicks + candidate.Period;
+                }
+                else
+                {
+                    // move on to the next period
+                    candidate.WakeupTime += candidate.Period;
+                }
```
egil, line 268: Good solution. This is safe because the current thread that is invoking the waiters/callback is the only one that can change _ now, since it is protected by the lock, and it is only that thread that can reenter the lock. So a change to _now could only come from the same thread.

Is my understanding correct?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -18,7 +18,7 @@ public class FakeTimeProvider : TimeProvider
     internal readonly HashSet<Waiter> Waiters = new();
     private DateTimeOffset _now = new(2000, 1, 1, 0, 0, 0, 0, TimeSpan.Zero);
     private TimeZoneInfo _localTimeZone = TimeZoneInfo.Utc;
-    private int _wakeWaitersGate;
+    private volatile int _wakeWaitersGate;
```
egil, line 21: Didn't think volatile was needed when you read the value using Interlocked.CompareExchange?

Is it because it is being set by direct assignment later?

egil (approved):
Looks good to me.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: > Didn't think volatile was needed when you read the value using Interlocked.CompareExchange?
> 
> Is it because it is being set by direct assignment later?

Yes, just as an added precaution against funny business in the compiler.

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
geeknoid: Exactly :-)

PR #4278, opened 2023-08-11
Remove the last ToBeRemoved component!

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4278)


src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheckExtensions.cs
```diff
@@ -30,7 +29,7 @@ public static IHealthChecksBuilder AddResourceUtilizationHealthCheck(this IHealt
         _ = Throw.IfNull(builder);
         _ = Throw.IfNull(tags);
 
-        _ = builder.Services.AddValidatedOptions<ResourceUtilizationHealthCheckOptions, ResourceUtilizationHealthCheckOptionsValidator>();
+        _ = builder.Services.AddOptionsWithValidateOnStart<ResourceUtilizationHealthCheckOptions, ResourceUtilizationHealthCheckOptionsValidator>();
```
joperezr, line 32: Assuming customers will also have to adapt for these sorts of changes, I wonder if this is something that UpgradeAssistant supports as well (meaning mapping static methods, since what we've done so far is mapping NuGet packages and renamed types) cc @RussKie

joperezr (approved):
LGTM. Thanks for thaking care of this @geeknoid! 😃

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheckExtensions.cs
RussKie: This is not the only API rename :)
My understanding is that the .NET UA supports such conversions.

PR #4279, opened 2023-08-11
Get coverage numbers up

geeknoid:


src/Generators/Microsoft.Gen.AutoClient/Parser.cs
```diff
@@ -653,16 +654,19 @@ private void Diag(DiagnosticDescriptor desc, Location? location, params object?[
         _reportDiagnostic(Diagnostic.Create(desc, location, messageArgs));
     }
 
+    [ExcludeFromCodeCoverage]
```
sebastienros, line 657: Is it because the compiler generates some code that is never accessed?

If so I would assume that compiler generated code could be excluded and the tool should have an option for that.

geeknoid: I filed a bug about this against Roslyn. Not all the generated code in a record type is properly annotated. The properties for example don't have annotations. So it messes up coverage.

RussKie:
I suggest pushing this change in to the internal repo as well and kick off an internal build to verify it builds successfully.

RussKie (approved):
If it's working - :shipit:

PR #4280, opened 2023-08-11
Forgot to escape JSON string output

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4280)

### DIFF
--- a/src/Generators/Microsoft.Gen.ComplianceReports/Emitter.cs
+++ b/src/Generators/Microsoft.Gen.ComplianceReports/Emitter.cs
@@ -174,6 +174,10 @@ internal sealed class Emitter : EmitterBase
 
     private void OutNameValue(string name, string value)
     {
+        value = value
+            .Replace("\\", "\\\\")
+            .Replace("\"", "\\\"");
+
         NewItem();
         OutIndent();
         Out($"\"{name}\": \"{value}\"");


src/Generators/Microsoft.Gen.ComplianceReports/Emitter.cs
```diff
@@ -174,6 +174,10 @@ private void OutArray(string name, Action action)
 
     private void OutNameValue(string name, string value)
     {
+        value = value
```
sebastienros, line 177: Why not use `JsonSerializer.Serialize(value)` ?

xakep139: IIRC Generators can't reference packages (e.g. `System.Text.Json`). @geeknoid, is that correct?

sebastienros: And they target NS2.0, so that wouldn't be possible, ok.

sebastienros (approved):

src/Generators/Microsoft.Gen.ComplianceReports/Emitter.cs
geeknoid: That's what I thought until a few minutes ago. Someone just sent me https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#use-functionality-from-nuget-packages which shows the MSBuild magic to make this happen.

I'll create an issue to track upgrading to that approach.

Issue #4249, opened 2023-08-08
Remove OTel-specific bits needed in tracing


klauco:
We want to provide enrichment and redaction capabilities for tracing, but at the same time we don't want to depend directly on OpenTelemetry-dotnet. Therefore, we should keep abstractions and core helper methods in dotnet/extensions and provide the glue code, connecting this functionality with OTel processors etc in opentelemetry-dotnet-contrib.

xakep139:
Already implemented abstractions are in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Extensions.Enrichment

RussKie:
Is there anything left outstanding?

xakep139:
@dariusclay needs to check whether sampling has to be moved to abstractions or not

dariusclay:
We don't have any real high level abstractions for tracing or sampling yet. The options and validator, which was used in our implementation of `DynamicSampler` could move in `Microsoft.Extensions.Telemetry.Abstractions`, but by themselves do not provide much value and not sure how useful it would be. What do you think folks?

dariusclay:
We'll keep ITraceEnricher, all other bits will go into Otel.

xakep139:
@dariusclay, I remember @evgenyfedorov2 initially added telemetry abstractions in OTel as abstract classes for easier future extension. Should we follow the same strategy here?

dariusclay:
@xakep139 I started this conversation with the team, and yes we feel it's best to have the same here. I'll open a new issue for it.

PR #4304, opened 2023-08-17
Remove OTel-specific bits

### LINKED ISSUE #4052: Distinguish OpenTelemetry dependent functionality
We should be clear what functionality works only with OpenTelemetry and either document it clearly or physically separate it in unique namespaces and assemblies. Today I see references in these assemblies:

https://github.com/dotnet/extensions/blob/7b7deb411dd0ceadcff22106ada65874a3b9adbc/src/Libraries/Microsoft.Extensions.Http.Telemetry/Microsoft.Extensions.Http.Telemetry.csproj#L42
https://github.com/dotnet/extensions/blob/7b7deb411dd0ceadcff22106ada65874a3b9adbc/src/Libraries/Microsoft.AspNetCore.Telemetry/Microsoft.AspNetCore.Telemetry.csproj#L35
https://github.com/dotnet/extensions/blob/7b7deb411dd0ceadcff22106ada65874a3b9adbc/src/Libraries/Microsoft.Extensions.Telemetry/Microsoft.Extensions.Telemetry.csproj#L37

I assume documentation will be less disruptive than refactoring, and I wouldn't object to it, but I do think it comes with some consequences:
1. Some dependency sensitive developers may avoid all functionality in these assemblies, even if the specific type or scenario they are interested never runs any OTel code.
2. Because of (1) there is now a tradeoff that needs to be considered when adding any new non-OTel specific functionality to the assembly. Implicitly in my head, these assemblies are currently 'OTel extension assemblies' and not 'general purpose telemetry assemblies'.
3. There is always a little extra risk of misuse/confusion if naming doesn't make the dependency obvious.

xakep139:
This resolves #4052
Related to #4249

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4304)


THIRD-PARTY-NOTICES.TXT
```diff
@@ -9,6 +9,8 @@ bring it to our attention. Post an issue or email us:
 The attached notices are provided for information only.
 
 
+TODO: Should we keep this notice?
```
xakep139, line 12: @joperezr @RussKie what do you think folks? There's references to OTel left at all.

src/Libraries/Microsoft.Extensions.Telemetry/Tracing.Sampling/SamplingOptions.cs
```diff
@@ -11,22 +11,28 @@ namespace Microsoft.Extensions.Telemetry.Tracing;
 public class SamplingOptions
```
xakep139, line 11: @dariusclay FYI, I decided to keep sampling for now, since I wasn't sure about its future.

THIRD-PARTY-NOTICES.TXT
RussKie: Good catch. No, it's no longer required if we don't have OTel code.

xakep139: I'll remove it then

geeknoid (approved):

THIRD-PARTY-NOTICES.TXT
xakep139: Done

RussKie (approved):

PR #4309, opened 2023-08-18
Correct redaction of Array.Empty<char>() and new char[0]

### LINKED ISSUE #4308: Breaking change in Redactor implementation
Calling `ToString()` here looks like a breaking change from the previous behaviour:
https://github.com/dotnet/extensions/blob/c8e34c57199a2e5e3864649f3bd16258b88ff67e/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs#L184

Possible here as well?
https://github.com/dotnet/extensions/blob/c8e34c57199a2e5e3864649f3bd16258b88ff67e/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs#L143

The change was introduced in https://github.com/dotnet/r9/pull/384.

The change manifests itself in Achs_Encrypting_Redactor_Does_Nothing_When_Source_Is_Empty tests. The callsite is:
```cs
var written = redactor.Redact(Array.Empty<char>(), buffer);
```
The current implementation doesn't receive an empty string (as one would expect) but a string `System.Char[]`:
image

image

The original implementation would receive the empty string. This change is very subtle and impossible to guard against in the consumer code.

Here's a repro:
```cs
    [Fact]
    public static void ArrayEmptyOfchar_Redacted_correctly()
    {
        var buffer = new char[5];

        var r = new PassthroughRedactor();
        var written = r.Redact(Array.Empty<char>(), buffer);

        Assert.Equal(0, written);

        foreach (var item in buffer)
        {
            Assert.Equal('\0', item);
        }
    }
```

RussKie:
Fixes #4308

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4309)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs
+++ b/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs
@@ -32,7 +32,7 @@ public abstract class Redactor
             return string.Empty;
         }
 
-        var length = GetRedactedLength(source);
+        int length = GetRedactedLength(source);
 
 #if NETCOREAPP3_1_OR_GREATER
         unsafe
@@ -115,12 +115,12 @@ public abstract class Redactor
 
             // Stryker disable all : Cannot kill the mutant because the only difference is allocating buffer on stack or renting it.
             // Null forgiving operator: The null case is checked with default equality comparer, but compiler doesn't understand it.
-            if (((ISpanFormattable)value).TryFormat(buffer, out var written, format.AsSpan(), provider))
+            if (((ISpanFormattable)value).TryFormat(buffer, out int written, format.AsSpan(), provider))
             {
                 // Stryker enable all : Cannot kill the mutant because the only difference is allocating buffer on stack or renting it.
 
                 var formatted = buffer.Slice(0, written);
-                var length = GetRedactedLength(formatted);
+                int length = GetRedactedLength(formatted);
 
                 unsafe
                 {
@@ -138,6 +138,16 @@ public abstract class Redactor
         if (value is IFormattable)
         {
             return Redact(((IFormattable)value).ToString(format, provider));
+        }
+
+        if (value is char[])
+        {
+            // An attempt to call value.ToString() on a char[] will produce a string "System.Char[]" and all redaction will be attempted on it,
+            // instead of the provided array. This will lead to incorrectly allocated buffers.
+            //
+            // NB: not using pattern matching as it is recognized by the JIT and since this is a generic type, the JIT ends up generating code
+            // without any of those conditional statements being present. But this only happens when not using pattern matching.
+            return Redact(((char[])(object)value).AsSpan());
         }
 
         return Redact(value?.ToString());
@@ -166,7 +176,7 @@ public abstract class Redactor
             Span<char> buffer = stackalloc char[MaximumStackAllocation];
 
             // Stryker disable all : Cannot kill the mutant because the only difference is allocating buffer on stack or renting it.
-            if (((ISpanFormattable)value).TryFormat(buffer, out var written, format.AsSpan(), provider))
+            if (((ISpanFormattable)value).TryFormat(buffer, out int written, format.AsSpan(), provider))
             {
                 // Stryker enable all : Cannot kill the mutant because the only difference is allocating buffer on stack or renting it.
                 var formatted = buffer.Slice(0, written);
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs
```diff
@@ -181,6 +186,11 @@ public int Redact<T>(T value, Span<char> destination, string? format = null, IFo
             return Redact(((IFormattable)value).ToString(format, provider), destination);
         }
 
+        if (value is char[] arrayOfChar)
```
geeknoid, line 189: I'm not sure whether this is semantically correct behavior. Why should a char array have special treatment over any other kind of array?

I think the test passing an empty array of char may need to be fixed instead.

RussKie: Because the redactor works with sequence of chars or strings represented as `ReadOnlySpan<char>`:
https://github.com/dotnet/extensions/blob/c8e34c57199a2e5e3864649f3bd16258b88ff67e/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs#L28C26-L28C44

Converting an array of char to a string is a wasteful allocation because at the next steps the new string is converted back to a span. 
https://github.com/dotnet/extensions/blob/c8e34c57199a2e5e3864649f3bd16258b88ff67e/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs#L94

Besides, when a user passes `Array.Empty<char>` or `new char[0]`, the user expects the work to be performed on essentially the empty sequence, and not on the string `System.Char[]`.

geeknoid: The real issue is the overload resolution. The test is passing a char[] and that's binding to the generic version of the call instead of binding to the overload that takes a ReadOnlySpan. We didn't use to have a generic version of the call, so the test would bind to the ReadOnlySpan overload and things worked.

geeknoid (commented):
We may want to rename Redact<T> to RedactObject<T> to avoid the overload resolution problem.

Looking at this code, I'm also wondering why we don't have other TryRedact overloads to mirror the Redact overloads. Seems like an oversight...

@rafal-mz thought?

test/Libraries/Microsoft.Extensions.Compliance.Abstractions.Tests/Redaction/RedactorTest.cs
```diff
@@ -31,38 +31,38 @@ public static void Redaction_Extensions_Return_Zero_On_Null_Input_Value()
     [InlineData(10000)]
     public static void User_Can_Get_String_From_IRedactor_Using_Extension_Method_With_Different_Input_Length(int length)
     {
-        var data = new string('*', length);
+        string data = new string('*', length);
```
rafal-mz, line 34: why this change?

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs
rafal-mz: Probably this usage is niche, I am not sure if we need this optimization. I don't think it is wrong though, we already have typechecks for iformattable, and it looks the same IMO.

rafal-mz (approved):

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Redaction/Redactor.cs
geeknoid: OK, then let's keep it.

@RussKie, please update the code to not use pattern matching and use the same pattern as the if checks for IFormattable and ISpanFormattable above. This pattern is recognized by the JIT and since this is a generic type, the JIT ends up generating code without any of those conditional statements being present. But this only happens when not using pattern matching.

Also, the same change needs to be applied to TryRedact<T>.

RussKie: We can't directly cast, and have to do it via an `object` cast
image
image

test/Libraries/Microsoft.Extensions.Compliance.Abstractions.Tests/Redaction/RedactorTest.cs
RussKie: In many .NET repos we don't use `var` for integral types and when the type isn't obvious, see https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md points 10 and 11. 
I rolled back this specific one.

RussKie:
Ready for another round of review.

RussKie:
Thank you

PR #4317, opened 2023-08-20
Simplify redaction logic a bit.

geeknoid:
- In LoggerMessageState, the redacted tag array is now grown in concert with the classified tag array. This makes it so if the new generated code is called using a legacy logger, the generated formatting function will never hit an "array out of bounds" exception.

- Remove some useless exception handling logic in the extended logger. This was a holdover from when the redaction logic was more complicated.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4317)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
@@ -188,36 +188,24 @@ internal sealed partial class ExtendedLogger : ILogger
         var loggers = MessageLoggers;
         var config = _factory.Config;
 
-        List<Exception>? exceptions = null;
-
         // redact
         JustInTimeRedactor? jitRedactors = null;
         for (int i = 0; i < msgState.NumClassifiedTags; i++)
         {
-            try
+            ref var cp = ref msgState.ClassifiedTagArray[i];
+            if (cp.Value != null)
             {
-                ref var cp = ref msgState.ClassifiedTagArray[i];
-                if (cp.Value != null)
-                {
-                    var jr = JustInTimeRedactor.Get();
-                    jr.Value = cp.Value;
-                    jr.Redactor = config.GetRedactor(cp.Classification);
-                    jr.Next = jitRedactors;
-                    jitRedactors = jr;
+                var jr = JustInTimeRedactor.Get();
+                jr.Value = cp.Value;
+                jr.Redactor = config.GetRedactor(cp.Classification);
+                jr.Next = jitRedactors;
+                jitRedactors = jr;
 
-                    var index = msgState.ReserveRedactedTagSpace(1);
-                    msgState.RedactedTagArray[index] = new(cp.Name, jr);
-                }
-                else
-                {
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -188,13 +188,11 @@ private void ModernPath(LogLevel logLevel, EventId eventId, LoggerMessageState m
         var loggers = MessageLoggers;
         var config = _factory.Config;
 
-        List<Exception>? exceptions = null;
-
         // redact
         JustInTimeRedactor? jitRedactors = null;
-        for (int i = 0; i < msgState.NumClassifiedTags; i++)
+        if (msgState.NumClassifiedTags > 0)
```
xakep139, line 193: Is it faster to have an `if` statement before the `for` loop? If not, then we don't really need this `if` - it will be covered by `i < msgState.NumClassifiedTags` in the `for` loop

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -205,20 +203,13 @@ private void ModernPath(LogLevel logLevel, EventId eventId, LoggerMessageState m
...
+                    msgState.RedactedTagArray[i] = new(cp.Name, jr);
                 }
                 else
                 {
-                    var index = msgState.ReserveRedactedTagSpace(1);
-                    msgState.RedactedTagArray[index] = new(cp.Name, null);
+                    msgState.RedactedTagArray[i] = new(cp.Name, null);
                 }
             }
-            catch (Exception ex)
```
xakep139, line 217: Does that mean that we can;t throw here anymore?

xakep139 (approved):
LGTM

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
geeknoid: Whoops, that was a leftover from when I had something explicitly in the if block, but not in the loop. I removed the if.

Thanks.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
geeknoid: Correct, this code is no longer calling outside of this module and so isn't vulnerable to exceptions.

joperezr:
@dotnet/dnceng Any clues why @geeknoid is not allowed to backport PRs somehow? He is part of the dotnet github team and he is an admin on this repo.

dougbu:
@joperezr, I seem to remember something about doing the 2FA dance and making myself a public member of one of the organizations. might as well confirm your settings in all of Microsoft, .NET Platform, and ASP.NET @geeknoid

Issue #4210, opened 2023-07-25
Http logging based on .NET8


klauco:
Http logging with support of the following:
 - sensitive data redaction of URL params and headers
 - formatted vs structured request path logging mode
 - exclude paths
 - logging of request/response body of requested content types
 - enrichment

xakep139:
PoC is in #4056, it's based on https://github.com/dotnet/aspnetcore/pull/48679

xakep139:
@Tratcher any chance you can complete your work in ASP.NET and in this repo?

Tratcher:
I'm planning to do the work in both repos next week.

klauco:
@Tratcher what is the current state?

Tratcher:
The AspNetCore changes were merged, I'm wrapping up the extensions changes in https://github.com/dotnet/extensions/pull/4330 now.

joperezr:
This is just waiting on a review for #4330

PR #4330, opened 2023-08-23
Layer enrichment & redaction on top of HttpLogging

Tratcher:
Consumes the new APIs from https://github.com/dotnet/aspnetcore/pull/50163 to consolidate functionality. The build won't pass until we consume that PR through an updated dependency, but everything builds and tests pass locally.

Fixes https://github.com/dotnet/extensions/issues/4210 #4492 #4440 #4451

Major changes:
- Removed redundant types like middleware, request body handling, response body handling, media types, etc.
- Exception logging is delegated to the ExceptionHandling middleware, but we still log any request data we've accumulated so far.

Now that it's functional, parts of the original design should be revisited like the manual path filtering. We don't want to re-invent routing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4330)


src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/HttpLoggingServiceExtensions.cs
```diff
@@ -22,56 +20,45 @@ namespace Microsoft.AspNetCore.Telemetry;
...
-    /// </param>
-    /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
-    /// <exception cref="ArgumentNullException">
-    /// Either <paramref name="services"/> or <paramref name="configure"/> is <see langword="null" />.
-    /// </exception>
-    public static IServiceCollection AddHttpLogging(this IServiceCollection services, Action<LoggingOptions> configure)
-    {
-        _ = Throw.IfNull(services);
-        _ = Throw.IfNull(configure);
+        _ = services.Configure(configureRedaction ?? (_ => { }));
```
xakep139, line 35:
```suggestion
        _ = services.Configure(configureRedaction ?? (static _ => { }));
```

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/HttpLoggingServiceExtensions.cs
```diff
@@ -22,56 +20,45 @@ namespace Microsoft.AspNetCore.Telemetry;
...
     /// </summary>
-    /// <param name="services">The <see cref="IServiceCollection"/> to add the service to.</param>
-    /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
-    /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null" />.</exception>
-    public static IServiceCollection AddHttpLogging(this IServiceCollection services)
+    /// <param name="services">The service collection.</param>
+    /// <param name="configureRedaction">Configures the redaction options.</param>
+    /// <param name="configureLogging">Configures the logging options.</param>
+    /// <returns>The original service collection.</returns>
+    [Experimental("ID")]
```
xakep139, line 29:
```suggestion
    /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null" />.</exception>
    [Experimental("ID")]
```

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/IHttpLogEnricher.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#if NET8_0_OR_GREATER
```
xakep139, line 4: I don't think we need these, the package targets `net8.0` only

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingEnrichmentInterceptor.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Microsoft.AspNetCore.Telemetry.Http.Logging;
+
+internal sealed class HttpLoggingEnrichmentInterceptor : IHttpLoggingInterceptor
+{
+    private readonly IHttpLogEnricher[] _enrichers;
+
+    public HttpLoggingEnrichmentInterceptor(IEnumerable<IHttpLogEnricher> httpLogEnrichers)
+    {
+        _enrichers = httpLogEnrichers.ToArray();
```
xakep139, line 22: I would log a warning/info here if there's no enrichers. It would mean that there's no point of calling `AddHttpLoggingEnrichment()`

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/HttpLoggingServiceExtensions.cs
```diff
@@ -22,56 +20,45 @@ namespace Microsoft.AspNetCore.Telemetry;
...
     {
         _ = Throw.IfNull(services);
-        _ = Throw.IfNull(section);
 
-        return AddHttpLoggingInternal(services, x => x.Bind(section));
+        services.TryAddEnumerable(ServiceDescriptor.Singleton<IHttpLoggingInterceptor, HttpLoggingEnrichmentInterceptor>());
+
+        // Internal stuff for route processing:
+        _ = services.AddHttpRouteProcessor();
+        _ = services.AddHttpRouteUtilities();
```
xakep139, line 60: These two aren't needed for enrichment

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingEnrichmentInterceptor.cs
```diff
@@ -0,0 +1,58 @@
...
+            || (!logContext.IsAnyEnabled(HttpLoggingFields.Response) && logContext.Parameters.Count == 0))
+        {
+            return default;
+        }
+
+        var context = logContext.HttpContext;
+        var enrichmentBag = LogMethodHelper.GetHelper();
+        foreach (var enricher in _enrichers)
+        {
+            enricher.Enrich(enrichmentBag, context.Request, context.Response);
```
xakep139, line 44: Enrichers should be called in a `try .. catch` block

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,176 @@
...
+    public ValueTask OnRequestAsync(HttpLoggingInterceptorContext logContext)
+    {
+        var context = logContext.HttpContext;
+        var request = context.Request;
+        if (ShouldExcludePath(context.Request.Path))
+        {
+            logContext.LoggingFields = HttpLoggingFields.None;
+        }
+
+        // Don't enrich if we're not going to log any part of the request
```
xakep139, line 72:
```suggestion
        // Don't redact if we're not going to log any part of the request
```

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        var context = logContext.HttpContext;
+
+        if (logContext.TryOverride(HttpLoggingFields.ResponseHeaders))
+        {
+            _responseHeadersReader.Read(context.Response.Headers, logContext.Parameters, HttpLoggingTagNames.ResponseHeaderPrefix);
+        }
+
+        // TODO: What about the exception case?
```
xakep139, line 150: Asked the same in https://github.com/dotnet/aspnetcore/pull/50163#discussion_r1305574882

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/HttpLoggingTagNames.cs
```diff
@@ -14,47 +16,47 @@ public static class HttpLoggingTagNames
...
     /// HTTP Request Body.
     /// </summary>
-    public const string RequestBody = "httpRequestBody";
+    public const string RequestBody = "Body";
 
     /// <summary>
     /// HTTP Response Body.
     /// </summary>
-    public const string ResponseBody = "httpResponseBody";
+    public const string ResponseBody = "Body";
```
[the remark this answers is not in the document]
Tratcher: Removed, this component is no longer involved in logging the body.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/Internal/HttpLoggingEnrichmentInterceptor.cs
Tratcher: I'll just make it throw.

xakep139: I don't think that we need to fail the HTTP request because of a one failed logging enricher.
@CarnaViire suggested the same in HttpClient logging component.

Tratcher:
Updated with the real AspNetCore dependency. Now it only fails due to complaints about APIs marked as Experimental.

src/Libraries/Microsoft.AspNetCore.Telemetry.Middleware/Logging/HttpLoggingServiceExtensions.cs
```diff
@@ -1,77 +1,123 @@
...
-    /// <param name="services">The <see cref="IServiceCollection"/> to add the service to.</param>
-    /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+    /// <param name="services">The service collection.</param>
+    /// <param name="configureRedaction">Configures the redaction options.</param>
+    /// <param name="configureLogging">Configures the logging options.</param>
+    /// <returns>The original service collection.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="services"/> is <see langword="null" />.</exception>
-    public static IServiceCollection AddHttpLogging(this IServiceCollection services)
+    public static IServiceCollection AddHttpLoggingRedaction(this IServiceCollection services, Action<LoggingRedactionOptions>? configureRedaction = null,
+        Action<HttpLoggingOptions>? configureLogging = null)
```
Tratcher, line 38: Having two callbacks here is awkward, but slightly more discoverable than not having them.

Ideas:
- Don't combine, doc that AddHttpLogging controls half of the settings.
- Duplicate the options from HttpLoggingOptions and then copy them over? That's especially awkward with the header collections that come pre-populated. (Need to consolidate the those settings too).

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// Any path added to the set will not be logged.
+    /// Paths are case insensitive.
+    /// Default set to an empty <see cref="HashSet{T}"/>.
+    /// </remarks>
+    /// <example>
+    /// A typical set of HTTP paths would be:
+    /// - "/probe/live".
+    /// - "/probe/ready".
+    /// </example>
+    public ISet<string> ExcludePathStartsWith { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
Tratcher, line 85: Is this redundant with the new features like short circuit routes, logging attributes, etc.?

xakep139: Not really, some folks implement their service's readiness/liveness probes in a controller and there's no point of logging these requests.
Are there logging attributes they can apply on a controller/action to exclude it from being logged?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HeaderReader.cs
```diff
@@ -19,17 +21,10 @@ public HeaderReader(IDictionary<string, DataClassification> headersToLog, IRedac
     {
         _redactorProvider = redactorProvider;
 
-        _headers = headersToLog.Count == 0
-            ? Array.Empty<KeyValuePair<string, DataClassification>>()
-            : headersToLog.ToArray();
+        _headers = headersToLog.Count == 0 ? [] : headersToLog.ToArray();
```
xakep139, line 24: Is `[]` a new syntax for `Array.Empty<T>`?

Tratcher: Yes

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+    public IDictionary<string, DataClassification> RequestHeadersDataClasses { get; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// Gets a map between response headers to be logged and their data classification.
+    /// </summary>
+    /// <remarks>
+    /// Default set to an empty dictionary.
+    /// That means that no response header will be logged by default.
+    /// </remarks>
+    public IDictionary<string, DataClassification> ResponseHeadersDataClasses { get; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
```
xakep139, line 70: @geeknoid maybe `StringComparer.InvariantCultureIgnoreCase` would fit better here?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/RequestHeadersLogEnricherOptions.cs
```diff
@@ -1,28 +1,27 @@
...
 public class RequestHeadersLogEnricherOptions
 {
     /// <summary>
-    /// Gets or sets a dictionary of header names that logs should be enriched with and their data classification.
+    /// Gets a dictionary of header names that logs should be enriched with and their data classification.
     /// </summary>
     /// <remarks>
     /// Default value is an empty dictionary.
     /// </remarks>
     [Required]
```
xakep139, line 24: Shouldn't there be an options validator for this class?

xakep139, line 24: I think this can be removed since you deleted its setter

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingRedactionInterceptor.cs
```diff
@@ -0,0 +1,203 @@
...
+        _responseHeadersReader = new(optionsValue.ResponseHeadersDataClasses, redactorProvider);
+
+        _excludePathStartsWith = optionsValue.ExcludePathStartsWith.ToArray();
+    }
+
+    public ValueTask OnRequestAsync(HttpLoggingInterceptorContext logContext)
+    {
+        var context = logContext.HttpContext;
+        var request = context.Request;
+        if (ShouldExcludePath(context.Request.Path.Value!))
```
xakep139, line 68: Since this is a hot path, we can have a field like `isExclusionListEmpty` to not call the `ShouldExcludePath()` method if the `_excludePathStartsWith` is empty

Tratcher: I'd expect the foreach to short circuit almost as fast without special handling.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/RequestHeadersLogEnricherOptions.cs
Tratcher:
```suggestion
```

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
Tratcher: Rational? I'd expect these to be opaque names.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingRedactionInterceptor.cs
xakep139: Method invocation will eat some CPU cycles

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
xakep139: We can still decide to keep the option as it provides the ability to change the behavior without recompiling the code. And some folks might not have an access to the source code of some controllers.
@geeknoid @noahfalk thoughts?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HeaderReader.cs
```diff
@@ -19,17 +21,10 @@ public HeaderReader(IDictionary<string, DataClassification> headersToLog, IRedac
...
     }
 
-    /// <summary>
-    /// Reads headers and applies filtering (if required).
-    /// </summary>
-    /// <param name="headers">A collection of headers to be read from.</param>
-    /// <param name="listToFill">A list to be filled with headers.</param>
-    public void Read(IHeaderDictionary headers, List<KeyValuePair<string, string>> listToFill)
+    public void Read(IHeaderDictionary headers, IList<KeyValuePair<string, object?>> logContext, string prefix)
```
geeknoid, line 27: Why make this an IList? That makes the call to Add() below an interface dispatch instead of an inlined call.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingServiceCollectionExtensions.cs
```diff
@@ -1,72 +1,110 @@
...
 
 /// <summary>
 /// Extension methods to register the HTTP logging feature within the service.
 /// </summary>
+[Experimental(diagnosticId: Experiments.HttpLogging, UrlFormat = Experiments.UrlFormat)]
 public static class HttpLoggingServiceCollectionExtensions
 {
     /// <summary>
-    /// Adds components for incoming HTTP requests logging into <see cref="IServiceCollection"/>.
+    /// Enables redaction of HTTP logging.
```
geeknoid, line 27:
```suggestion
    /// Enables redaction of HTTP request logging output.
```

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingServiceCollectionExtensions.cs
```diff
@@ -1,72 +1,110 @@
...
     /// </summary>
-    /// <param name="services">The <see cref="IServiceCollection"/> to add the service to.</param>
-    /// <returns>The value of <paramref name="services"/>.</returns>
+    /// <param name="services">The service collection.</param>
+    /// <param name="configureRedaction">Configures the redaction options.</param>
```
geeknoid, line 30: We systematically call this parameter "configure" throughout dotnet/extensions, can we do the same here?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingServiceCollectionExtensions.cs
```diff
@@ -1,72 +1,110 @@
...
+    /// <returns>The original service collection.</returns>
```
geeknoid, line 31: I've updated this text for everything in dotnet/extensions to say:

```
<returns>The value of <paramref name="services".</returns>
```

Can we do the same here?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingServiceCollectionExtensions.cs
```diff
@@ -1,72 +1,110 @@
...
         _ = Throw.IfNull(services);
-        _ = Throw.IfNull(configure);
 
-        return AddHttpLoggingInternal(services, builder => builder.Configure(configure));
+        _ = services.Configure(configureRedaction ?? (static _ => { }));
+        services.TryAddEnumerable(ServiceDescriptor.Singleton<IHttpLoggingInterceptor, HttpLoggingRedactionInterceptor>());
+
+        _ = services.AddHttpLogging(o =>
+        {
+            o.CombineLogs = true;
```
geeknoid, line 42: Should we document this behavior?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingServiceCollectionExtensions.cs
```diff
@@ -1,72 +1,110 @@
...
+                foreach (var path in paths)
+                {
+                    _ = o.ExcludePathStartsWith.Add(path);
+                }
+            }
+
+            var routeParams = section.GetSection(nameof(LoggingRedactionOptions.RouteParameterDataClasses));
+            foreach (var entry in routeParams.GetChildren())
+            {
+                var taxonomy = entry.GetValue<string>(nameof(DataClassification.TaxonomyName));
```
geeknoid, line 78: I don't really have a better answer here, but I'm pretty sure this isn't the right approach. I don't want developer using magic numbers in their config files to represent data classes, that's just a bug farm.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
```diff
@@ -1,60 +1,65 @@
...
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.DiagnosticIds;
 
 namespace Microsoft.AspNetCore.Diagnostics.Logging;
 
 /// <summary>
 /// Constants used for incoming HTTP request logging tags.
 /// </summary>
+[Experimental(diagnosticId: Experiments.HttpLogging, UrlFormat = Experiments.UrlFormat)]
```
geeknoid, line 16: Why add this?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.Extensions.Http.Diagnostics;
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.AspNetCore.Diagnostics.Logging;
+
+/// <summary>
+/// Top-level model for redacting incoming HTTP requests and their corresponding responses.
+/// </summary>
+[Experimental(diagnosticId: Experiments.HttpLogging, UrlFormat = Experiments.UrlFormat)]
+public class LoggingRedactionOptions
```
geeknoid, line 19: For all the properties, the default value should be documented in a `<value>` block, rather than in `<remarks>`

geeknoid: This type shoudl be using option validation code gen to ensure non-nullable fields are not forced to null, and enum values are in range, etc.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <summary>
+    /// Gets a map between HTTP path parameters and their data classification.
+    /// </summary>
+    /// <remarks>
+    /// Default set to an empty dictionary.
+    /// If a parameter within a controller's action is not annotated with a data classification attribute and
+    /// it's not found in this map, it will be redacted as if it was <see cref="DataClassification.Unknown"/>.
+    /// If you don't want a parameter to be redacted, mark it as <see cref="DataClassification.None"/>.
+    /// </remarks>
+    public IDictionary<string, DataClassification> RouteParameterDataClasses { get; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
```
geeknoid, line 52: [Required]

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// <summary>
+    /// Gets a map between request headers to be logged and their data classification.
+    /// </summary>
+    /// <remarks>
+    /// Default set to an empty dictionary.
+    /// That means that no request header will be logged by default.
+    /// </remarks>
+    public IDictionary<string, DataClassification> RequestHeadersDataClasses { get; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
```
geeknoid, line 61: [Required]

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <remarks>
+    /// Any path added to the set will not be logged.
+    /// Paths are case insensitive.
+    /// Default set to an empty <see cref="HashSet{T}"/>.
+    /// </remarks>
+    /// <example>
+    /// A typical set of HTTP paths would be:
+    /// - "/probe/live".
+    /// - "/probe/ready".
+    /// </example>
```
geeknoid, line 84: [Required]

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// </remarks>
+    public IDictionary<string, DataClassification> RequestHeadersDataClasses { get; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// Gets a map between response headers to be logged and their data classification.
+    /// </summary>
+    /// <remarks>
+    /// Default set to an empty dictionary.
+    /// That means that no response header will be logged by default.
+    /// </remarks>
```
geeknoid, line 69: [Required]

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/RequestHeadersLogEnricherOptions.cs
```diff
@@ -1,28 +1,25 @@
...
     /// </summary>
     /// <remarks>
     /// Default value is an empty dictionary.
     /// </remarks>
-    [Required]
     [Experimental(diagnosticId: Experiments.Telemetry, UrlFormat = Experiments.UrlFormat)]
-#pragma warning disable CA2227 // Collection properties should be read only
-    public IDictionary<string, DataClassification> HeadersDataClasses { get; set; } = new Dictionary<string, DataClassification>();
-#pragma warning restore CA2227 // Collection properties should be read only
+    public IDictionary<string, DataClassification> HeadersDataClasses { get; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
```
geeknoid, line 24:
```suggestion
    [Required]
    public IDictionary<string, DataClassification> HeadersDataClasses { get; set; } = new Dictionary<string, DataClassification>(StringComparer.OrdinalIgnoreCase);
```

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/IncomingPathLoggingMode.cs
```diff
@@ -1,11 +1,17 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.DiagnosticIds;
+
 namespace Microsoft.AspNetCore.Diagnostics.Logging;
 
 /// <summary>
 /// Strategy to decide how request path is logged.
 /// </summary>
+[Experimental(diagnosticId: Experiments.HttpLogging, UrlFormat = Experiments.UrlFormat)]
```
geeknoid, line 14: Why add this?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
Tratcher: It's not settable.

xakep139: There's no setter here, Martin

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HeaderReader.cs
Tratcher: It's coming directly from AspNetCore, we don't have a choice here. https://github.com/dotnet/aspnetcore/blob/933bc8447e10e3e4a327e8ac1c598935ddd294ae/src/Middleware/HttpLogging/src/HttpLoggingInterceptorContext.cs#L81

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
Tratcher: I marked all public API for this feature as experimental.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs
Tratcher: There aren't any fields that can be forced to null. How does the enum validation work? I don't see any attributes for that.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/IncomingPathLoggingMode.cs
Tratcher: I marked all public API for this feature as experimental.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
Tratcher: Removed

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/IncomingPathLoggingMode.cs
Tratcher: Removed

geeknoid (approved):

Tratcher:
@geeknoid FYI I've pushed changes to also fix https://github.com/dotnet/extensions/issues/4451.

PR #4352, opened 2023-09-04
Major update to the logging generator.

geeknoid:
- Now has full support for modern data classifications, including combined data classifications (where two annotations apply to the same piece of state).

- Fix some bugs around putting spurious @ in logging property names in certain cases.

- Improved many error paths by eliminating cascading errors. You now tend to just get one clean error.

- Simplify [LogProperties] and create the [TagProvider] attribute. Both features are orthogonal, they should never have been provided by a single attribute.

- Dynamically adjust the name of the generated static variables in order to avoid conflicts with user-created symbols.

- Harmonize transitive behavior. When you apply [LogProperties] on a parameter, it results in a one level deep iteration of the parameter's properties. If you want to go deeper, you need to apply [LogProperties] on each property you want to go deeper on. At any point, you can apply [TagProvider] to invoke user-controlled code instead of using the automatic way.

- Rename LogPropertiesAttribute.OmitParameterName to OmitReferenceName since the attribute can now be used on properties as well as parameters.

- Fully support the LoggerMessageAttribute constructors. Prior logic was still based on the similar-but-different LogMethodAttribute and was therefore missing a few corner cases.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4352)


tekian (approved):

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -348,43 +389,68 @@ void GenPropertyLoads(LoggingMethod lm, string stateName, out int numReservedUnc
...
-                                : string.Empty;
+                            var value = member.IsEnumerable
+                                ? $"{LoggerMessageHelperType}.Stringify({accessExpression})"
+                                : ts;
 
-                            OutLn($"{skipNull}{stateName}.AddClassifiedTag(\"{propName}\", {value}, {classification});");
+                            if (member.PotentiallyNull && !member.IsEnumerable)
+                            {
+                                OutOpenBrace();
+                                OutLn($"var {tmpVarName} = {value};");
```
xakep139, line 435: Shouldn't we check whether an `IEnumerable` member is `null` before strigifying it?

src/Generators/Microsoft.Gen.Logging/Model/LoggingMethodParameterExtensions.cs
```diff
@@ -38,10 +36,10 @@ internal static class LoggingMethodParameterExtensions
     {
         foreach (var propertyToLog in propertiesToLog)
         {
-            if (propertyToLog.TransitiveMembers.Count > 0)
+            if (propertyToLog.Properties.Count > 0)
```
xakep139, line 39: nit: Use `HasProperties` here?

src/Generators/Microsoft.Gen.Logging/Parsing/DiagDescriptors.cs
```diff
@@ -212,4 +212,22 @@ internal sealed class DiagDescriptors : DiagDescriptorsBase
...
+        category: Category);
+
+    public static DiagnosticDescriptor TagProviderInvalidUsage { get; } = Make(
+        id: "LOGGEN033",
+        title: Resources.TagProviderInvalidUsageTitle,
+        messageFormat: Resources.TagProviderInvalidUsageMessage,
+        category: Category);
+
+    public static DiagnosticDescriptor InvalidAttributeUsage { get; } = Make(
+        id: "LOGGEN033",
```
xakep139, line 229: Should be `LOGGEN034`

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
```diff
@@ -309,4 +303,28 @@
   <data name="TemplateStartsWithAtSymbolTitle" xml:space="preserve">
     <value>A logging message template starts with "@"</value>
   </data>
+  <data name="CantUseDataClassificationWithLogPropertiesOrTagProviderMessage" xml:space="preserve">
+    <value>Cant combine the [LogProperties] or [TagProvider] attributes with data classification attributes</value>
+  </data>
+  <data name="CantUseDataClassificationWithLogPropertiesOrTagProviderTitle" xml:space="preserve">
+    <value>Cant combine the [LogProperties] or [TagProvider] attributes with data classification attributes</value>
```
xakep139, line 310: `Can't`?

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
```diff
@@ -309,4 +303,28 @@
   <data name="TemplateStartsWithAtSymbolTitle" xml:space="preserve">
     <value>A logging message template starts with "@"</value>
   </data>
+  <data name="CantUseDataClassificationWithLogPropertiesOrTagProviderMessage" xml:space="preserve">
+    <value>Cant combine the [LogProperties] or [TagProvider] attributes with data classification attributes</value>
```
xakep139, line 307: `Can't`?

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
```diff
@@ -309,4 +303,28 @@
...
+    <value>Can only have one of [LogProperties], [TagProvider], and [LogPropertyIgnore]</value>
+  </data>
+  <data name="CantMixAttributesTitle" xml:space="preserve">
+    <value>Can only have one of [LogProperties], [TagProvider], and [LogPropertyIgnore]</value>
+  </data>
+  <data name="TagProviderInvalidUsageMessage" xml:space="preserve">
+    <value>Parameter "{0}" is annotated to use a tag provider but it has special semantics (ILogger, LogLevel, Exception, etc.)</value>
+  </data>
+  <data name="TagProviderInvalidUsageTitle" xml:space="preserve">
+    <value>Method parameter can't be used for with a tag provider</value>
```
xakep139, line 322: `Method parameter can't be used with a tag provider`? (without `for`)

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
```diff
@@ -309,4 +303,28 @@
...
     <value>A logging message template starts with "@"</value>
   </data>
+  <data name="CantUseDataClassificationWithLogPropertiesOrTagProviderMessage" xml:space="preserve">
+    <value>Cant combine the [LogProperties] or [TagProvider] attributes with data classification attributes</value>
+  </data>
+  <data name="CantUseDataClassificationWithLogPropertiesOrTagProviderTitle" xml:space="preserve">
+    <value>Cant combine the [LogProperties] or [TagProvider] attributes with data classification attributes</value>
+  </data>
+  <data name="CantMixAttributesMessage" xml:space="preserve">
+    <value>Can only have one of [LogProperties], [TagProvider], and [LogPropertyIgnore]</value>
```
xakep139, line 313: Does it make sense to have `LogPropertyIgnore` since we don't do transitive traversal by default?

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
geeknoid: Indeed, nice catch!

src/Generators/Microsoft.Gen.Logging/Parsing/DiagDescriptors.cs
geeknoid: Another good catch!

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
geeknoid: For a given property, you can ask:

* Output the property's string (no attribute)
* Ignore this property and don't output it (LogPropertyIgnore)
* Output the properties of the property's object (LogProperties)
* Invoke the tag collector to get properties for the property's object (TagProvider)

So I think they're all mutually-exclusive.

xakep139: Ah, I see

PR #4355, opened 2023-09-05
Rename Metering to Metrics

### LINKED ISSUE #4048: Rename Metering->Metrics
.NET uses the terms "Metrics" in our namespaces. Dotnet/extensions currently uses the term "Metering". We should converge to "Metrics" both to match existing namespaces and the industry more generally.

geeknoid:
Fixes #4048 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4355)


joperezr (approved):
LGTM from a very quick glance. Seems like this also changed properties inside targets which are used for various things like generating reports and such, which I'm assuming won't be caught by CI if broken. Were you able to test those targets to ensure things still work as expected?

geeknoid:
Yeah, everything seems to be working.

PR #4356, opened 2023-09-05
Fix NaN handling

### LINKED ISSUE #4329: Throw.IfOutOfRange(double, ...) and related functions do not handle NaN correctly
The `Throw.IfOutOfRange(double, ...)` function is implemented as so:

```cs
if (argument < min || argument > max) { /* fail */ }
```

However, this does not account for NaN correctly, as NaN inputs will cause this to evaluate to `(false || false)`, which means that the function will _not_ fail. If the input is provided by an untrusted agent (e.g., being deserialized from the wire into a double), this could cause the `Throw.IfOutOfRange` logic to be bypassed and the application to attempt to treat NaN as if it were a legitimate value. This could result in unwanted app behavior.

Consider instead rewriting this expression as:

```cs
if (!(min <= argument && argument <= max)) { /* fail */ }
```

That is, instead of writing "if any of the preconditions is violated, fail" - rewrite the statement as "if not all of the preconditions succeed, fail." This will ensure that weird floating point values like NaN do not unexpectedly bypass these checks.

geeknoid:
Fixes #4329 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4356)

### DIFF
--- a/src/Shared/Throw/Throw.cs
+++ b/src/Shared/Throw/Throw.cs
@@ -870,7 +870,10 @@ internal static class Throw
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static double IfLessThan(double argument, double min, [CallerArgumentExpression(nameof(argument))] string paramName = "")
     {
-        if (argument < min)
+        // strange conditional needed in order to handle NaN values correctly
+#pragma warning disable S1940 // Boolean checks should not be inverted
+        if (!(argument >= min))
+#pragma warning restore S1940 // Boolean checks should not be inverted
         {
             ArgumentOutOfRangeException(paramName, argument, $"Argument less than minimum value {min}");
         }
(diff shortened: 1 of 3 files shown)


danmoseley:
@tannergooding is this the right/best way?

src/Shared/Throw/Throw.cs
```diff
@@ -943,7 +943,7 @@ public static double IfGreaterThanOrEqual(double argument, double max, [CallerAr
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static double IfOutOfRange(double argument, double min, double max, [CallerArgumentExpression(nameof(argument))] string paramName = "")
     {
-        if (argument < min || argument > max)
+        if (!(min <= argument && argument <= max))
```
RussKie, line 946: I suggest adding a comment providing the background to avoid someone inverting the check later.
Also add a test? :)

geeknoid: Done.

geeknoid:
@tannergooding Thanks for your feedback. I think I addressed it, please review again.

tannergooding (approved):

danmoseley:
core libraries felt IComparable was fast enough
https://github.com/dotnet/runtime/blob/17b60e35528ec0f1eba8a6606168412738bade59/src/libraries/System.Private.CoreLib/src/System/ArgumentOutOfRangeException.cs#L205

should we just copy that code / conditional compile to use it?

PR #4387, opened 2023-09-11
Converge resource monitoring interface

geeknoid:
* Revamp the resource monitoring API to eliminate the split between Windows and Linux. There are still some options which are Windows-specific and Linux-specific, but I expect those to be harmonized in a subsequent PR.

* This component has received a lot of changes from a lot of different people over the years, so it has become a bit messy. This PR renames various things to try and bring some sanity.

Note that aside from revamping the extensions to add the components and combining the option types into a single class, there are no functional changes in this PR. It's all lipstick and no pig.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4387)


eng/Diags/SonarAnalyzer.CSharp.yml
```diff
@@ -463,7 +463,7 @@ Diagnostics:
     Attributes:
       general:
         Severity: Warning
-  S1144:
+  S1144:
```
RussKie, line 466: This can be reverted

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringExtensions.cs
```diff
@@ -88,24 +84,114 @@ public static class ResourceMonitoringExtensions
...
+            _ = builder.AddWindowsProvider();
+        }
+        else if (os.Platform == PlatformID.Unix)
+        {
+            _ = builder.AddLinuxProvider();
+        }
+        else
+        {
+            throw new NotSupportedException("Resource monitoring is not supported on this operating system.");
+        }
```
RussKie, line 130: Is there a reason we can't use this for .NET8+ too?

geeknoid: I'm not sure what you mean. Can you restate please?

PR #4392, opened 2023-09-12
Converge resource monitoring interface

geeknoid:
* Revamp the resource monitoring API to eliminate the split between Windows and Linux. There are still some options which are Windows-specific and Linux-specific, but I expect those to be harmonized in a subsequent PR.

* This component has received a lot of changes from a lot of different people over the years, so it has become a bit messy. This PR renames various things to try and bring some sanity.

Note that aside from revamping the extensions to add the components and combining the option types into a single class, there are no functional changes in this PR. It's all lipstick and no pig.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4392)


RussKie:
There are few styling issues. Other than that - 👍

RussKie (approved):

tekian (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Internal/ISnapshotProvider.cs
```diff
@@ -18,6 +18,6 @@ internal interface ISnapshotProvider
     /// </summary>
     /// <returns>An appropriate sample.</returns>
 #pragma warning disable S4049 // Properties should be preferred
-    ResourceUtilizationSnapshot GetSnapshot();
+    Snapshot GetSnapshot();
```
rafal-mz, line 21: The second struct was renamed from `Utilization` to `ResourceUtilization`, while this one was shortened. Why don't we have symmetry?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Internal/Linux/LinuxUtilizationProvider.cs
```diff
@@ -56,8 +51,8 @@ internal sealed class LinuxUtilizationProvider : ISnapshotProvider
         _scale = hostCpus * Hundred / availableCpus;
         _scaleForTrackerApi = hostCpus / availableCpus;
 
-        _ = meter.CreateObservableGauge<double>(name: LinuxResourceUtilizationCounters.CpuConsumptionPercentage, observeValue: CpuPercentage);
-        _ = meter.CreateObservableGauge<double>(name: LinuxResourceUtilizationCounters.MemoryConsumptionPercentage, observeValue: MemoryPercentage);
+        _ = meter.CreateObservableGauge<double>(name: ResourceUtilizationCounters.CpuConsumptionPercentage, observeValue: CpuPercentage);
```
rafal-mz, line 54: I would have rename it to 'ResourceUtilizationMetrics'. Then we could have counters, gauges and up/down metric names in one class (better discoverability).

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringOptions.cs
```diff
@@ -29,6 +24,10 @@ public class ResourceMonitoringOptions
     /// <value>
     /// The default value is 5 seconds.
     /// </value>
+    /// <remarks>
+    /// This value represents the total amount of time for which the resource monitor tracks utilization
+    /// information for the system.
+    /// </remarks>
     [TimeSpan(MinimumSamplingWindow, MaximumSamplingWindow)]
     public TimeSpan CollectionWindow { get; set; } = DefaultCollectionWindow;
```
rafal-mz, line 32: From my experience those three time values are often misunderstood by users; from what I've seen they often get invlaid utilization value (NaN), while changing them. I would consider providing less control or extend our docs on how to set it up and what to expect by changing each of the time span value here.

PR #4394, opened 2023-09-13
Update TFMs

RussKie:
* Drop support for .NET Core 3.1
* Add support for .NET 6.0
* Run tests for only supported TFMs - i.e., net462.net6.0,net8.0

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4394)


Directory.Build.targets
```diff
@@ -86,7 +81,7 @@
     </PropertyGroup>
   </Target>
 
-  <!-- This target will make sure that projects targeting netcoreapp3.1 will also have the Microsoft.Extensions.Logging.Abstractions analyzer removed. -->
+  <!-- This target will make sure that projects targeting net462 will also have the Microsoft.Extensions.Logging.Abstractions analyzer removed. -->
```
RussKie, line 84: @joperezr I kept this target because the comment in [Microsoft.Extensions.Telemetry.Abstractions.csproj](https://github.com/dotnet/extensions/pull/4394/files#diff-09674220f6f8ad9ca75c179401eeca831f746b20d7f1c11f099b46c7c72d7411) refers to both net462 and netcoreapp3.1:
image


Please let me know, if I misunderstood it, and these should be removed.

RussKie:
@geeknoid I got hundreds of IDE0028 warnings:
```
warning IDE0028: Collection initialization can be simplified (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0028)
```

...which addressed will look like this:
image


Should we suppress the warning? Or update the whole codebase?

geeknoid:
@RussKie Where do the warnings come from? This did this just enabled as a result of an analyzer upgrade?

RussKie:
There are also hundreds of TBD warnings:
```
error TBD: (NETCORE_ENGINEERING_TELEMETRY=Build) 'Microsoft.Extensions.Options.Contextual.IOptionsContext' is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed. (https://aka.ms/dotnet-extensions-warnings/TBD)
```

These are the result of (likely because of the newer SDK?):
https://github.com/dotnet/extensions/blob/80f8c709c81ed9975d6b2994ed6ef4232aaee963/Directory.Build.targets#L56-L62

As an immediate workaround I'm going to suppress `TBD` warning, but we need to generalise how the assembly-level attribute is applied based on the workstream? We'll need a mapping between workstreams and the experiment areas listed in https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md. I'll raise a tracking issue.

RussKie:
> @RussKie Igor Velikorossov FTE Where do the warnings come from? This did these just appear as a result of an analyzer upgrade?

The newer SDK: 
```diff
   "sdk": {
-    "version": "8.0.100-rc.1.23381.2"
+    "version": "8.0.100-rc.2.23462.28"
   },
```

geeknoid:
I think we should fix the IDE0028 warnings (I'm guessing VS can do this over the whole solution).

And please suppress the TBD warning for now, I'll find a suitable workaround.

RussKie:
> I think we should fix the IDE0028 warnings (I'm guessing VS can do this over the whole solution).
> 

Sure.
Personally, I am not too fond of the new style, but I guess I may get used to it :)

RussKie:
The build is green, good to merge 😉

joperezr:
Thanks for fixing this @RussKie 

FYI @ericstj

Directory.Build.props
```diff
@@ -27,12 +27,12 @@
...
     <NetCoreTargetFrameworks Condition="'$(CustomTargetFrameworks)' != ''">$(SupportedNetCoreTargetFrameworks)</NetCoreTargetFrameworks>
-    <NetCoreTargetFrameworks Condition="'$(CustomTargetFrameworks)' != '' AND $(CustomTargetFrameworks.Contains('netcoreapp3.1'))">$(NetCoreTargetFrameworks);netcoreapp3.1</NetCoreTargetFrameworks>
+    <NetCoreTargetFrameworks Condition="'$(CustomTargetFrameworks)' != '' AND $(CustomTargetFrameworks.Contains('net6.0'))">$(NetCoreTargetFrameworks);net6.0</NetCoreTargetFrameworks>
     <ConditionalNet462 Condition="'$(CustomTargetFrameworks)' == '' OR $(CustomTargetFrameworks.Contains('net462'))">;net462</ConditionalNet462>
 
     <!-- All the .NET TFMs we're testing against -->
     <TestNetCoreTargetFrameworks>$(NetCoreTargetFrameworks)</TestNetCoreTargetFrameworks>
-    <TestNetCoreTargetFrameworks Condition="'$(CustomTargetFrameworks)' == ''">$(TestNetCoreTargetFrameworks);net7.0;net6.0</TestNetCoreTargetFrameworks>
+    <TestNetCoreTargetFrameworks Condition="'$(CustomTargetFrameworks)' == ''">$(TestNetCoreTargetFrameworks);net6.0</TestNetCoreTargetFrameworks>
```
xakep139, line 35: @RussKie now this emits a warning that looks like:
```
Error occurred while restoring NuGet packages: Invalid restore input. The original target frameworks value must match the aliases. Original target frameworks: net8.0;net6.0;net6.0;net462, aliases: net8.0;net6.0;net462. Input files: C:\Users\...\source\dotnet-ext\test\Generators\Microsoft.Gen.AutoClient\Generated\Microsoft.Gen.AutoClient
```

I.e., the `net6.0` TFM is listed twice for test projects that have `<TargetFrameworks>$(TestNetCoreTargetFrameworks)</TargetFrameworks>`

joperezr: Can either you or @RussKie log an issue to fix this so that we make sure it is fixed? How come we aren't seeing this error in CI?

RussKie: https://github.com/dotnet/extensions/pull/4449

PR #4405, opened 2023-09-13
Start shuffling namespaces per API feedback

geeknoid:


geeknoid:
I'm going to force this in since it's just a bunch of mind-numbing namespace renames.

noahfalk (commented):
I skimmed through and it seems fine (agreed, mind-numbing!)

PR #4407, opened 2023-09-14
Escape C# strings using helper

sebastienros:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4407)

### DIFF
--- a/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
+++ b/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Gen.AutoClient.Model;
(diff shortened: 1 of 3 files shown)


RussKie (approved):

xakep139:
@dbarbosapn FYI

xakep139:
@sebastienros thank you for the fix! Do you think we need to extend the same approach to other source-gens?

geeknoid:
Oh, I didn't know this existed. I have a custom version of the conversion function in the logging generator, I'll check in the others.

@tarekgh The existing options validator generator has an EscapeString function that should be replaced by the Roslyn function.

geeknoid (approved):

tarekgh:
> The existing options validator generator has an EscapeString function that should be replaced by the Roslyn function.

I can track this for .NET 9.0.

sebastienros:
@xakep139 I think so, I only changed this component as I was reviewing it and wanted to keep my focus.

PR #4443, opened 2023-09-19
Update SDK

halter73:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4443)

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "microsoft.visualstudio.slngen.tool": {
-      "version": "10.3.4",
+      "version": "10.4.0",
       "commands": [
         "slngen"
       ]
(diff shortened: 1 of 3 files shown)


RussKie:
O.O we need to update SlnGen tool to make it aware of .NET 9... @jeffkl is there way forcing the tool to use the latest SDK?

SteveSandersonMS:
I updated to the latest slngen but the issue remains.

Have filed https://github.com/microsoft/slngen/issues/527, which I think we're now blocked on.

.config/dotnet-tools.json
```diff
@@ -15,7 +15,7 @@
       ]
     },
     "microsoft.visualstudio.slngen.tool": {
-      "version": "9.5.3",
+      "version": "10.3.4",
```
RussKie, line 18:
```suggestion
      "version": "10.4.0",
```

RussKie (approved):

PR #4469, opened 2023-09-26
Update HttpClient logging APIs per Noah's comments

geeknoid:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4469)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingServiceCollectionExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingServiceCollectionExtensions.cs
@@ -15,7 +15,7 @@ using Microsoft.Shared.Diagnostics;
 namespace Microsoft.Extensions.DependencyInjection;
 
 /// <summary>
-/// Extension methods to register HTTP client logging features.
+/// Extensions to register extended HTTP client logging features.
 /// </summary>
 public static class HttpClientLoggingServiceCollectionExtensions
 {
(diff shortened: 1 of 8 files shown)


src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingServiceCollectionExtensions.cs
```diff
@@ -28,7 +28,7 @@ public static class HttpClientLoggingServiceCollectionExtensions
     /// All other loggers are removed - including the default one, registered via <see cref="HttpClientBuilderExtensions.AddDefaultLogger(IHttpClientBuilder)"/>.
     /// </remarks>
     /// <exception cref="ArgumentNullException">Argument <paramref name="services"/> is <see langword="null"/>.</exception>
-    public static IServiceCollection AddDefaultHttpClientLogging(this IServiceCollection services)
+    public static IServiceCollection AddExtendedHttpClientLogging(this IServiceCollection services)
```
xakep139, line 31: So, now both names are the same?
`IServiceCollection.AddExtendedHttpClientLogging` and `IHttpClientBuilder.AddExtendedHttpClientLogging`
Wouldn't it confuse users?

xakep139 (approved):
Left only one comment, otherwise looks good

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingServiceCollectionExtensions.cs
noahfalk: IMO as long as `serviceCollection.AddExtendedHttpClientLogging()` has the same behavior as `services.ConfigureHttpClientDefaults(httpBuilder => httpBuilder.AddExtendedHttpClientLogging())` then I wouldn't expect significant confusion. Worst case if a customer was aware of both and confused by it there is still no wrong answer, whichever one they try will work.

@davidfowl I think has stronger opinions that we should avoid offering IServiceCollection extension methods when a sub-builder extension method also exists. Personally I'm neutral about this pattern, but assuming I didn't misunderstand his position, he'd probably want us not to include the IServiceCollection variant.

noahfalk (approved):
I'm happy. @davidfowl take a look since it seemed like you wanted to avoid extension methods simultaneously on IServiceCollection and sub-builders.

PR #4471, opened 2023-09-26
Improve AutoClient generator's resilience against null/empty values

### LINKED ISSUE #4463: Enhance tests and docs for AutoClient
See https://github.com/dotnet/extensions/pull/4373#discussion_r1320603340

dbarbosapn:
Changes:
- Handles null/empty REST method paths
- Handles null/empty static header names and values
- Adds testing for all new scenarios

fixes #4463 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4471)


src/Libraries/Microsoft.Extensions.Http.AutoClient/StaticHeaderAttribute.cs
```diff
@@ -33,8 +35,8 @@ public sealed class StaticHeaderAttribute : Attribute
     /// <summary>
     /// Initializes a new instance of the <see cref="StaticHeaderAttribute"/> class.
     /// </summary>
-    /// <param name="header">The name of the header.</param>
-    /// <param name="value">The value of the header.</param>
+    /// <param name="header">The name of the header. Cannot be empty or null.</param>
+    /// <param name="value">The value of the header. If null, the header will not be sent.</param>
     public StaticHeaderAttribute(string header, string value)
```
geeknoid, line 40: Should this be `string? value`, given the comment you just added?

geeknoid: Or maybe `string? value = null`

dbarbosapn: Sounds good. 

I went for this solution on the "Value" specifically because it's what happens with dynamic ones as well. If you pass null value to the method the header/query will not be sent.

But for static we can definitely validate and just not allow it. I'll do that :)

dbarbosapn: Fixed!

geeknoid (approved):

Issue #4477, opened 2023-09-27
Support metrics in ResourceMonitoring for Windows


xakep139:
Currently only Linux part supports .NET metrics to be emitted via `ObservableGauge`'s. This should be extended to cover Windows counterpart as well.

PR #4373, opened 2023-09-07
Current dotnet/extensions API Surface

geeknoid:
Please review to make sure you're happy with:

* Package names
* Namespace names
* Type names
* Method names

Last chance to speak up before November is upon us.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4373)

PR #4482, opened 2023-09-27
Aligning metrics with OTel semantic conventions

### LINKED ISSUE #4432: dotnet/extensions metric names to follow some convention
Currently, the metric names are using backslashes. OTel uses dots. We need to go through the metrics and ensure we stick to the same naming convention (and semantic grouping as part of the metric name).

**Proposal:**
1. use dots instead of backslashes
2. use singular, not plural

xakep139:
Fixes #4432 

The final names are:
1. Meter `Microsoft.Extensions.Diagnostics.ResourceMonitoring`:
- Observable Gauge `process.cpu.utilization`, unit `1`, values: `0.0` - `1.0`
- Observable Gauge `dotnet.process.memory.virtual.utilization`, unit `1`, values: `0.0` - `1.0`
- Observable UpDownCounter `system.network.connections`, unit `{connection}`, tags/attributes:
  - `network.transport` = `tcp`
  - `network.type` = [`ipv4`, `ipv6`]
  - `system.network.state` = [`close`, `close_wait`, `closing`, `delete`, `established`, `fin_wait_1`, `fin_wait_2`, `last_ack`, `listen`, `syn_recv`, `syn_sent`, `time_wait`]
  
2. Meter `Microsoft.AspNetCore.HeaderParsing`:
- Counter `aspnetcore.header_parsing.parse_errors`, tags/attributes:
  - `error.type`
  - `aspnetcore.header_parsing.header.name`
- Counter `aspnetcore.header_parsing.cache_accesses`, tags/attributes:
  - `aspnetcore.header_parsing.header.name`
  - `aspnetcore.header_parsing.cache_access.type`

3. Meter `Microsoft.Extensions.Diagnostics.HealthChecks`:
- Counter `dotnet.health_check.reports`, tags/attributes:
  - `dotnet.health_check.status` of type [`HealthStatus`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.diagnostics.healthchecks.healthstatus), [`Degraded`, `Healthy`, `Unhealthy`]
- Counter `dotnet.health_check.unhealthy_checks`, tags/attributes:
  - `dotnet.health_check.name`
  - `dotnet.health_check.status` of type [`HealthStatus`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.diagnostics.healthchecks.healthstatus), [`Degraded`, `Healthy`, `Unhealthy`]
  
4. Resilience tags for enrichment:
- `error.type`
- `request.dependency.name` - to correlate with HttpClient metrics
- `request.name` - to correlate with HttpClient metrics

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4482)


martintmk:
Can you also align the resilience tags?

https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs

xakep139:
> Can you also align the resilience tags?
> 
> https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs

Surely I will, thanks for pointing that out

noahfalk:
@lmolkova - are you interested in taking a look at these?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
```diff
@@ -7,24 +7,24 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
...
 /// <remarks>
-/// These counters are currently only published on Linux.
+/// These metrics are currently only published on Linux.
 /// </remarks>
 /// <seealso cref="System.Diagnostics.Metrics.Instrument"/>
 public static class ResourceUtilizationCounters
 {
     /// <summary>
-    /// Gets the CPU consumption of the running application in percentages.
+    /// Gets the CPU consumption of the running application in range <c>[0, 1]</c>.
```
reyang, line 16: I have a stepping back question regarding all utilizations measured by %.

Related discussions and PRs:

* https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/625#discussion_r967319029
* https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/612#discussion_r956469274

FYI @noahfalk IIRC for garbage collector metrics we decided to use absolute value (e.g. total time spent in GC since the beginning of the process lifecycle), and encourage users to derive utilization based on their actual need (so the user can choose whatever sliding window and sampling frequency).

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -91,7 +92,7 @@ public double CpuPercentage()
 
                     if (deltaHost > 0 && deltaCgroup > 0)
                     {
-                        var percentage = Math.Min(Hundred, deltaCgroup / deltaHost * _scale);
+                        var percentage = Math.Min(1.0, deltaCgroup / deltaHost * _scale);
```
reyang, line 95: Curious, why do we need the `Min` logic here? (sounds like a sequencing issue where different variables are fetched at different time, or is this about clock skew)

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
```diff
@@ -7,24 +7,24 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
...
     /// </summary>
     /// <remarks>
     /// The type of an instrument is <see cref="System.Diagnostics.Metrics.ObservableGauge{T}"/>.
     /// </remarks>
-    public static string CpuConsumptionPercentage => "cpu_consumption_percentage";
+    public static string CpuUtilization => "process.cpu.utilization";
```
noahfalk, line 21: Be wary that the process semantic conventions aren't stable and might change after you ship this. https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/process-metrics.md#semantic-conventions-for-os-process-metrics. You either need to work with OTel folks to try agreeing on a stable convention up-front, potentially take breaking changes in the future, or have metrics that are unaligned with the standard.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
noahfalk: As I recall this component has a sampling mechanism and sliding window built into it and the app developer can configure some of those parameters or accept defaults. So when this component reports an instantaneous utilization measurement of 73%, it really means that over the last X minutes some aggregation of Y CPU usage samples had an average CPU utilization of 73% for the app developer's chosen X and Y. Its not the pattern I'd expect most metrics to follow as there is extra implementation complexity and it can't accurately be re-aggregated to compute other sliding windows later. However because the app developer configures a sampling window of known fixed size it doesn't suffer from the issues that those other metrics had where they inferred a sample window from the polling frequency or from some unpredictable process behavior.

A while back I encouraged the engineers who were working on this to simplify and not maintain their own sliding window sampling reservoir, but I feel like what is there now is still well defined and likely useful as long as the sampling window is configured well up-front. I didn't object to this even though I expect metrics of this style will be an outlier in the long run. Having something like this shouldn't block us from later creating an alternate metric that emits absolute measures of CPU and memory usage with more aggregation flexibility.

@reyang do you feel like that addresses your concerns or you are still worried here?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
xakep139: IIRC sometimes you can exceed your service's CPU limits in a container. @rafal-mz is it the case?

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
```diff
@@ -12,10 +12,12 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 
 internal static partial class Metric
 {
-    [Counter("healthy", "status", Name = @"R9\\HealthCheck\\Report")]
+    // TODO: should we rename "healthy" to "is_healthy"? Or maybe remove the attribute at all?
+    // To klauco: do we really need this true/false dimension? We already have "health.status" dimension.
+    [Counter("healthy", "health.status", Name = "health_check.reports")]
```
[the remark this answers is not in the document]
xakep139: @klauco, can we remove the dimension?

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Metric.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.AspNetCore.HeaderParsing;
 
 internal static partial class Metric
 {
-    [Counter("HeaderName", "Kind", Name = @"HeaderParsing.ParsingErrors")]
+    [Counter("header.name", "error.kind", Name = "aspnetcore.header_parsing.parse_errors")]
```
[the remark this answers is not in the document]
xakep139: Thanks, applied

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
```diff
@@ -7,24 +7,24 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
 /// Represents the names of instruments published by this package.
 /// </summary>
 /// <remarks>
-/// These counters are currently only published on Linux.
+/// These metrics are currently only published on Linux.
 /// </remarks>
 /// <seealso cref="System.Diagnostics.Metrics.Instrument"/>
 public static class ResourceUtilizationCounters
```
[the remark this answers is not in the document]
xakep139: So, do you suggest having these instruments' names somewhere in documentation instead?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
```diff
@@ -7,24 +7,24 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
...
 
     /// <summary>
-    /// Gets the memory consumption of the running application in percentages.
+    /// Gets the memory consumption of the running application in range <c>[0, 1]</c>.
     /// </summary>
     /// <remarks>
     /// The type of an instrument is <see cref="System.Diagnostics.Metrics.ObservableGauge{T}"/>.
     /// </remarks>
-    public static string MemoryConsumptionPercentage => "memory_consumption_percentage";
+    public static string MemoryUtilization => "process.memory.utilization";
```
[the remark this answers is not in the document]
xakep139: Good point, I was following https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/process-metrics.md, but there's no such thing we have here. Will add `virtual` then since it's the right "type" of memory we measure here.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
reyang: I'm not too concerned about having alternative metrics that customers can conveniently use - as long as there is clarity:
1) What exactly does this metric mean.
2) What's the recommendation if there are multiple ways to get similar things - e.g. pros/cons, pitfalls, best practices.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -58,13 +60,13 @@ internal sealed class LinuxUtilizationProvider : ISnapshotProvider
         var meter = meterFactory.Create("Microsoft.Extensions.Diagnostics.ResourceMonitoring");
 #pragma warning restore CA2000 // Dispose objects before losing scope
 
-        _ = meter.CreateObservableGauge(name: ResourceUtilizationCounters.CpuConsumptionPercentage, observeValue: CpuPercentage);
-        _ = meter.CreateObservableGauge(name: ResourceUtilizationCounters.MemoryConsumptionPercentage, observeValue: MemoryPercentage);
+        _ = meter.CreateObservableGauge(name: ResourceUtilizationCounters.CpuUtilization, observeValue: CpuUtilization, unit: "1");
+        _ = meter.CreateObservableGauge(name: ResourceUtilizationCounters.MemoryUtilization, observeValue: MemoryUtilization, unit: "1");
```
reyang, line 64: Doesn't blocking this PR, we should add clear description to all instruments.
Here are some examples on how we hammered out the description:
* https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/455
* https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Runtime#metrics

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
reyang: > 2. What's the recommendation if there are multiple ways to get similar things - e.g. pros/cons, pitfalls, best practices. For example, https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Process.

@noahfalk given OpenTelemetry.Instrumentation.Process is still in Preview, I think we should make a decision about the direction - e.g. all CPU/Memory metrics will be exposed by `Microsoft.Extensions.Diagnostics.ResourceMonitoring`, and we will deprecate `OpenTelemetry.Instrumentation.Process` and point customers to `Microsoft.Extensions.Diagnostics.ResourceMonitoring`. Meanwhile we'll make sure `Microsoft.Extensions.Diagnostics.ResourceMonitoring` works for all supported version of the runtime + operating systems. Does this make sense and do you feel we can make a call before .NET 8 GA?

@Yun-Ting FYI since you've worked on `OpenTelemetry.Instrumentation.Process`.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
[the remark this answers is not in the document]
noahfalk: Yeah. I expect most developers will consuming from out-of-process so they will be discovering these via docs, other external sources of info (samples, blogs, Q&A sites), or exploratory telemetry tooling. This is some doc work I started but need to get back to and finish up:
https://github.com/dotnet/docs/pull/37213
Once that reference existed I'd encourage you or anyone else adding metrics in the .NET platform to add to it.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
noahfalk: My thought would be:
In .NET 8 - Point devs to these ResourceMonitoring metrics. Either deprecate OTel.Instrumentation.Process or adjust guidance so that we only suggest it only as a fallback for devs that need the metric in a specific form not already handled by ResourceMonitoring.

In the future (.NET 9?) - Implement the core CPU + memory metrics currently in OTel.Instrumentation.Process inside System.Diagnostic.DiagnosticSource. I think metrics like `process.cpu.time` or `process.memory.virtual` are so commonly desired that we shouldn't be asking devs to add extra library dependencies or new background polling services (`serviceCollection.AddResourceMonitoring()`) in order to enable them. At this point ResourceMonitoring would serve two roles going forward: (1) support for networking metrics which are a little more involved to configure (2) back-compat for projects already using the other process metrics defined here and wanting to continue doing so.

What does everyone think?

reyang: > My thought would be: In .NET 8 - Point devs to these ResourceMonitoring metrics.

Just confirming - point devs that are using any supported version of .NET (core/framework), not just limited to folks who use .NET 8?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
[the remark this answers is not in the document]
noahfalk: From a little web searching, usage_in_bytes appears to be RSS + swap cache + swap, which approximately corresponds to committed virtual memory. Looking on Windows the container case appears aligned with that (https://github.com/dotnet/extensions/blob/2967ec96d4d3a46cf43e54f8ff5ef22e4e5d69e1/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs#L151), but the non-container case is very different:
https://github.com/dotnet/extensions/blob/2967ec96d4d3a46cf43e54f8ff5ef22e4e5d69e1/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsSnapshotProvider.cs#L36

I think we need all cases to agree on the same memory abstraction or it will be very confusing trying to name this metric or describe it to users. I don't think of this as a naming problem, its an inconsistency in the underlying abstraction. I'd suggest we either get it resolved or avoid making this memory metric public.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
noahfalk: Yeah, I meant as of November 2023 (or whenever this assembly releases as stable), all .NET devs should be able to use ResourceMonitoring to satisfy their need for CPU/Virtual Memory metrics.

NOTE: After I wrote that I saw that this library has a weirdly inconsistent implementation of how it measures memory. I'm presuming that gets resolved. If it didn't then I wouldn't feel comfortable recommending it.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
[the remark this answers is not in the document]
xakep139: Created #4522 to track documentation thing. Will make the class internal.

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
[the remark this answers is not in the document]
klauco: this indeed looks redundant. This dimension just reduces the set of possible health states from 3 to 2, so not much. I think it makes sense to remove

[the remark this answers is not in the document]
xakep139: Removed, thanks

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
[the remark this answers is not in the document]
xakep139: I see. Renamed the instrument to be `process.memory.virtual.utilization`.
Regarding Windows - we don't emit anything via `Meter` or `IMeterFactory` there yet, thus it will be only pushed via `IResourceUtilizationPublisher` defined by a user. But I agree to your point, now it's `VirtualMemorySize64` for non-container case. Does it seem correct now?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
```diff
@@ -10,203 +11,310 @@ internal sealed class WindowsCounters
...
 
+        var tcpTag = new KeyValuePair<string, object?>("network.transport", "tcp");
+
+        // These are covered in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-metrics.md#attributes:
+        var tcpVersionFourTag = new KeyValuePair<string, object?>("network.type", "ipv4");
+        var tcpVersionSixTag = new KeyValuePair<string, object?>("network.type", "ipv6");
+
+        // TODO: these are heavily rewritten to align with
+        // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemnetworkconnections
+        // All these should be of UpDownCounter type
```
xakep139, line 31: @noahfalk @reyang I'm mostly worried about these gauges. Should I make them `UpDownCounter` instead or use `process.network.connections.count` name (to continue keeping them as gauges). What do you think?

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Metrics/MetricCollector.cs
```diff
@@ -250,7 +250,7 @@ public Task WaitForMeasurementsAsync(int minCount, CancellationToken cancellatio
     [SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Not relevant in this case")]
     public async Task WaitForMeasurementsAsync(int minCount, TimeSpan timeout)
     {
-#if NET8_0_OR_GREATER
+#if NET8_0
```
xakep139, line 253:
```suggestion
#if NET8_0_OR_GREATER
```

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
xakep139: Good point. I created #4523 to track that.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
xakep139: @klauco FYI

klauco: @noahfalk can we have all this functionality marked as experimental + have a clear notice here and in docs that the semantic conventions are not yet stable and we might change this?

src/Generators/Microsoft.Gen.Metrics/Emitter.cs
```diff
@@ -297,7 +297,7 @@ private void GenTagList(MetricMethod metricMethod)
             foreach (var tagName in metricMethod.TagKeys)
             {
                 var paramName = GetSanitizedParamName(tagName);
-                OutLn($"new global::System.Collections.Generic.KeyValuePair<string, object?>(\"{paramName}\", {paramName}),");
+                OutLn($"new global::System.Collections.Generic.KeyValuePair<string, object?>(\"{tagName}\", {paramName}),");
```
xakep139, line 300: FYI @klauco: the generator now emits dots, colons and dashes as-is in dimension names

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
reyang: > @noahfalk can we have all this functionality marked as experimental + have a clear notice here and in docs that the semantic conventions are not yet stable and we might change this?

+1

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
reyang: +1 These should be `UpDownCounter`s. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#instrument-selection

The current name (`process.network.connections`) seems _better_ than `process.network.connections.count` in my personal opinion - although OTel semantic convention has the final saying. (this is another example: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserveractive_requests, note that I didn't use a permanent link, since the main branch version is very likely to be the final stable version)

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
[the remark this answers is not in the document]
noahfalk: No unfortunately VirtualMemorySize64 measures reserved virtual memory. That number can be (and usually is) vastly different than committed virtual memory (OTel's instrumentation appears to have the same issue and we missed it!). Pretty sure you'll need to pinvoke to the OS to get this value.

Here is an example app you can test with:
```C#
using System.Diagnostics;
using System.Runtime.InteropServices;

// You can experiment with manually allocating reserved or committed memory
//uint reserveSize = 2_000_000_000;
//uint commitSize = 100_000_000;
//IntPtr reservedMem = VirtualAlloc(IntPtr.Zero, reserveSize, AllocationType.Reserve, MemoryProtection.ReadWrite);
//IntPtr commitMem = VirtualAlloc(IntPtr.Zero, commitSize, AllocationType.Commit, MemoryProtection.ReadWrite);
//Console.WriteLine($"Reserved {reserveSize/1_000_000_000.0} GB of memory");
//Console.WriteLine($"Committed {commitSize / 1_000_000_000.0} GB of memory");

Console.WriteLine($"Process.VirtualMemorySize64= {Process.GetCurrentProcess().VirtualMemorySize64 / 1_000_000_000.0} GB");
Console.ReadLine();


[DllImport("kernel32.dll", SetLastError = true, ExactSpelling = true)]
static extern IntPtr VirtualAlloc(IntPtr lpAddress,
   uint dwSize, AllocationType flAllocationType, MemoryProtection flProtect);

[Flags]
public enum AllocationType : uint
{
    Commit = 0x1000,
    Reserve = 0x2000,
}

[Flags]
public enum MemoryProtection : uint
{
    ReadWrite = 0x04,
}
```

When I run this locally it reports:
```
Process.VirtualMemorySize64= 2204.04000768 GB
```

Also I was looking more at the code and I see in some cases it is using a commit limit and in other cases it is using physical memory as the denominator of the ratio. These mismatched measurement abstractions seem weird and ratios larger than 1 would certainly be possible, maybe even desirable/healthy. I think reporting raw numbers and naming them accurately rather than inventing novel ratios would make this a lot more comprehensible.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
noahfalk: We are free to write whatever text we want in the docs and its np from my perspective if you write that kind of disclaimer. In terms of programatically identifying it as experimental the only thing I am aware of is the ExperimentalAttribute on APIs. There isn't anything that specifically attaches to a metric.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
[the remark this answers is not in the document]
xakep139: I created #4538 to track that. The behavior doesn't affect metrics (as I mentioned we don't push Windows measurements to metrics yet - #4477) and I don't want to extend scope of this PR more. Feel free to unresolve the discussion.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsSnapshotProvider.cs
```diff
@@ -33,6 +33,6 @@ public Snapshot GetSnapshot()
             TimeSpan.FromTicks(TimeProvider.GetUtcNow().Ticks),
             TimeSpan.FromTicks(process.PrivilegedProcessorTime.Ticks),
             TimeSpan.FromTicks(process.UserProcessorTime.Ticks),
-            (ulong)process.WorkingSet64);
+            (ulong)process.VirtualMemorySize64);
```
xakep139, line 36:
```suggestion
            (ulong)process.WorkingSet64);
```

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
xakep139: I can add this into the instrument's description (e.g. `"This instrument is in Experimental stage and can be changed/removed in future versions."`). What do you think folks?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
xakep139: Changed to `UpDownCounter`s, thanks

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
rafal-mz: That is correct that those values are read sequentially (one after another), so how we read them might affect the precision. We might also lose a bit of precision while multiplying by fraction of cpu (cgroups allow to have 1.5 cpu for instance).  Also, in calculations we are assuming that all cpus produce the same number of cycles through the given time slice, which is probably not the case.

I didn't look for the root cause of the issue, since I've seen that other tools (like docker stats) have the same problem. As, it is impossible to spend more resources than we have and this is some approxximation error, I've cut it not to trick the customers.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
noahfalk: I don't think it would be a substitute for docs, but it wouldn't hurt to do that.

danmoseley:
@JamesNK  could you have a look?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
```diff
@@ -10,203 +11,289 @@ internal sealed class WindowsCounters
...
+        // These are covered in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/rpc-metrics.md#attributes:
+        var tcpVersionFourTag = new KeyValuePair<string, object?>("network.type", "ipv4");
+        var tcpVersionSixTag = new KeyValuePair<string, object?>("network.type", "ipv6");
+
+        // These metrics are aligned with
+        // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemnetworkconnections
+
+        // IPv4:
+        _ = meter.CreateObservableUpDownCounter(
+            InstrumentName,
```
JamesNK, line 35: Are multiple observable counters with the same name and different tags supported? I haven't used this pattern at all.

Has someone verified this produces expected data in something like the Prometheus exporter scrape exprter?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
```diff
@@ -10,203 +11,289 @@ internal sealed class WindowsCounters
...
             () =>
             {
                 var snapshot = tcpTableInfo.GetIPv4CachingSnapshot();
                 return snapshot.ClosedCount;
-            });
+            },
+            unit: "{connection}",
```
JamesNK, line 41: nit: InstrumentName is a constant. Might as well do the same with unit to reduce duplication and to be consistent.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Metric.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.AspNetCore.HeaderParsing;
 
 internal static partial class Metric
 {
-    [Counter("HeaderName", "Kind", Name = @"HeaderParsing.ParsingErrors")]
+    [Counter("header.name", "error.type", Name = "aspnetcore.header_parsing.parse_errors")]
```
JamesNK, line 11: I don't know if we're still at the stage to discuss API, but making tag names the param array here feels kind of weird. I could imagine someone entering `[Counter("my.cool.counter.name")] ` and expecting a counter with that name but instead getting a counter with no name a tag that name instead.

IMO `CounterAttribute` should have a ctor that takes the name. It's the most significant and required piece of information for a counter. Then specify a type or array of strings as tags as optional parameter, e.g.

```cs
[Counter("my.cool.counter.name", Tags = [ "one", "two", "etc" ])]
```

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
```diff
@@ -5,13 +5,13 @@ namespace Microsoft.Extensions.Resilience.Internal;
...
+    public const string FailureReason = "failure.reason";
 
-    public const string FailureSummary = "failure-summary";
+    public const string FailureSummary = "failure.summary";
 
-    public const string DependencyName = "dep-name";
+    public const string DependencyName = "dependency.name";
 
-    public const string RequestName = "req-name";
+    public const string RequestName = "request.name";
```
JamesNK, line 16: Are these tags reused from semantic conventions or specific to resilience? If they are specific, then maybe they should be prefixed to avoid clashing with other tag names, e.g. `resilience.failure.source`.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Metric.cs
JamesNK: Added a comment to the API review PR: https://github.com/dotnet/extensions/pull/4373/files#r1351364773

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
noahfalk: OTel may tolerate it, but it isn't recommended and it will probably generate warnings at minimum. https://opentelemetry.io/docs/specs/otel/metrics/sdk/#duplicate-instrument-registration

The intended pattern is to create a single ObservableUpDownCounter and then use the constructor overload where you provide a callback that returns an IEnumerable<Measurement> instead of a single Measurement. This will let you enumerate multiple tag combinations and values with a single instrument.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Metric.cs
xakep139: Thanks for raising that. It's out of scope of this PR, so I'm resolving this thread. Feel free to re-open it.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationCounters.cs
xakep139: Documentation is tracked under #4522, we'll highlight that these metrics are experimental.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
xakep139: Thanks for the explanations!

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
xakep139: Will update these to have just one instrument

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
xakep139: Not applicable anymore since we have just one instrument

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
xakep139: Done, thanks!
@martintmk FYI (again 😄)

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
xakep139: Done

joperezr:
@xakep139 is this ready to go? @JamesNK any additional feedback?

xakep139:
> @xakep139 is this ready to go? @JamesNK any additional feedback?

I think I addressed all comments. If anybody has any other thoughts, I look forward to resolve them as well 🙂

joperezr:
@JamesNK shared he is going to help take another look, and also suggested @lmolkova to take a look as well if possible 😃

JamesNK (approved):
LGTM. @lmolkova might have some thoughts on naming.

noahfalk (approved):
I'm good with it, though you might want to consider removing the memory.utilization gauge. I think its at pretty high risk of needing the name and/or calculations re-worked and I'd be reticent to advertise it to anyone as-is.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -58,13 +60,13 @@ internal sealed class LinuxUtilizationProvider : ISnapshotProvider
         var meter = meterFactory.Create("Microsoft.Extensions.Diagnostics.ResourceMonitoring");
 #pragma warning restore CA2000 // Dispose objects before losing scope
 
-        _ = meter.CreateObservableGauge(name: ResourceUtilizationCounters.CpuConsumptionPercentage, observeValue: CpuPercentage);
-        _ = meter.CreateObservableGauge(name: ResourceUtilizationCounters.MemoryConsumptionPercentage, observeValue: MemoryPercentage);
+        _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.CpuUtilization, observeValue: CpuUtilization, unit: "1");
+        _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.MemoryUtilization, observeValue: MemoryUtilization, unit: "1");
```
lmolkova, line 64: ITel defines different metrics in the [process namespace](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/process-metrics.md) - `process.memory.usage` and `process.memory.virtual`. If we have strong reasons not to report those and define something for .NET, I'd suggest following [JVM pattern](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/jvm-metrics.md) and put dotnet-only metrics into `dotnet` namespace.

test/Libraries/Microsoft.AspNetCore.HeaderParsing.Tests/HeaderParsingFeatureTests.cs
```diff
@@ -147,10 +147,11 @@ public void TryParse_returns_false_on_error()
...
         Assert.Equal("Can't parse header 'Date' due to 'Unable to parse date time offset value.'.", _logger.Collector.LatestRecord.Message);
 
-        var latest = metricCollector.LastMeasurement!;
+        var latest = metricCollector.LastMeasurement;
+        Assert.NotNull(latest);
         latest.Value.Should().Be(1);
-        latest.Tags["HeaderName"].Should().Be("Date");
-        latest.Tags["Kind"].Should().Be("Unable to parse date time offset value.");
+        latest.Tags["header.name"].Should().Be("Date");
+        latest.Tags["error.type"].Should().Be("Unable to parse date time offset value.");
```
lmolkova, line 154: We use a closed list of errors (or a full exception name) is other places where `error.type` is set and should follow the same pattern consistently.
The exception message has high-cardinality and should not be used on metrics dimensions.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Metric.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.AspNetCore.HeaderParsing;
...
     public static partial ParsingErrorCounter CreateParsingErrorCounter(Meter meter);
 
-    [Counter("HeaderName", "Type", Name = @"HeaderParsing.CacheAccess")]
+    [Counter("header.name", "access.type", Name = "aspnetcore.header_parsing.cache_accesses")]
```
lmolkova, line 14: we prefix ASP.NET Core-specific  attributes with `aspnetcore` https://github.com/open-telemetry/semantic-conventions/blob/4c4945569908b30349ab2a133880a67ac48c79ce/docs/dotnet/dotnet-aspnetcore-metrics.md.

I'd suggest to follow the same pattern here

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
```diff
@@ -12,14 +11,14 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 
 internal static partial class Metric
 {
-    [Counter("healthy", "status", Name = @"R9\\HealthCheck\\Report")]
+    [Counter("health.status", Name = "health_check.reports")]
     public static partial HealthCheckReportCounter CreateHealthCheckReportCounter(Meter meter);
 
-    [Counter("name", "status", Name = @"R9\\HealthCheck\\UnhealthyHealthCheck")]
+    [Counter("health_check.name", "health.status", Name = "health_check.unhealthy_checks")]
```
lmolkova, line 17: please consider adding `aspnetcore` prefix to metric and attributes

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
```diff
@@ -1,212 +1,85 @@
...
+        var commonTags = new TagList
+        {
+            tcpTag
+        };
+
+        // The metric is aligned with
+        // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemnetworkconnections
+
+        _ = meter.CreateObservableUpDownCounter(
+            "process.network.connections",
```
lmolkova, line 36: OTel metric is 
```suggestion
            "system.network.connections",
```

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
```diff
@@ -5,13 +5,13 @@ namespace Microsoft.Extensions.Resilience.Internal;
 
 internal static class ResilienceTagNames
 {
-    public const string FailureSource = "failure-source";
+    public const string FailureSource = "resilience.failure.source";
```
lmolkova, line 8: `resilience` is a bit vague term and can be used for many different things beyond `Microsoft.Extensions.Resilience`. Can we come up with a more specific prefix? E.g. `ms.resilience` or `aspnetcore.resilience` ?

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
```diff
@@ -5,13 +5,13 @@ namespace Microsoft.Extensions.Resilience.Internal;
...
 
-    public const string FailureReason = "failure-reason";
+    public const string FailureReason = "resilience.failure.reason";
```
lmolkova, line 10: based on what I see here https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceMetricsEnricher.cs, we can use `error.type` attribute here (and full exception name)

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
```diff
@@ -5,13 +5,13 @@ namespace Microsoft.Extensions.Resilience.Internal;
...
 
-    public const string FailureSummary = "failure-summary";
+    public const string FailureSummary = "resilience.failure.summary";
```
lmolkova, line 12: what's the expected cardinality of this attribute? It looks a bit scary - `_exceptionSummarizer.Summarize(e).ToString()`

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
martintmk: One note ere. These are just additional tags on top of the ones built-in into Polly:
https://www.pollydocs.org/advanced/telemetry.html#metrics

One of the tags is `exception.type` one as defined by:
https://opentelemetry.io/docs/specs/otel/trace/exceptions/


There is an intersection here between `exception.type` and `error.type` and I am wondering how should we handle this.

Just for context:

This is how `HttpResponseMessage` is translated into `FailureResultContext`:
https://github.com/dotnet/extensions/blob/124736421d4e44ce1c5a8d9120edef4621a9c87b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs#L156

And this is how exceptions are translated:
https://github.com/dotnet/extensions/blob/124736421d4e44ce1c5a8d9120edef4621a9c87b/src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceMetricsEnricher.cs#L37



Can we trim the `FailureResultContext`

geeknoid:
@xakep139 Is this ready to go in?

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
xakep139: This package might be used in non-web applications, thus I'm not sure we need to prefix metrics here with `aspnetcore`.
The package's name and its namespace don't contain `AspNetCore` either, do you strongly feel we need to do the prefix?

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Metric.cs
xakep139: Done

xakep139:
> @xakep139 Is this ready to go in?

Not yet, I need to address latest feedback that @lmolkova left

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsCounters.cs
xakep139: Applied, thanks!

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
xakep139: We agreed with @klauco to schedule a meeting to discuss memory metrics `ResourceMonitoring` component

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
JamesNK: This one shouldn't have `aspnetcore` if it's not specific to aspnetcore

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
xakep139: Switched to `dotnet.resilience.*`

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
xakep139: Resolving this for now, feel free to re-open the thread

test/Libraries/Microsoft.AspNetCore.HeaderParsing.Tests/HeaderParsingFeatureTests.cs
xakep139: Header parsers should provide a closed set of parsing errors. @geeknoid can you confirm that?

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
lmolkova: thank you for the context!

To the original comment: if polly already reports `exception.type`, is there any reason to report `resilience.failure.reason` at all since it's always an exception type?

> Can we trim the FailureResultContext or maybe just expose add the following tags?

error.summary: will be status code such as 500 for HttpResponseMessage or exception summary for exceptions.
exception.source: applicable only for exceptions (or even don't add anything here)

I'd avoid adding new tags to OTel-defined namespaces - this is against OTel recommendations https://github.com/open-telemetry/opentelemetry-specification/blob/563958cb2bd8529990f19fdce7a5f3643bf63091/specification/common/attribute-naming.md?plain=1#L128

But I believe that if the exception summary has low cardinality, you can put it into `error.type` attribute which is intended to capture all kinds of error codes and fallback to (full) exception type.

We can also introduce `dotnet.exception.source`  attribute and (hopefully) reuse in other parts of .NET and 3rd party libraries.

lmolkova (commented):
The changes look good to me, but I believe we still need to confirm a few things:
- do we expect exception summarize to produce low-cardinality summaries
- do we still need failure reason attribute
- what exactly CPU utilization metric reports and if it needs `state` attributes (as OTel one)

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
```diff
@@ -5,13 +5,13 @@ namespace Microsoft.Extensions.Resilience.Internal;
...
+    public const string FailureSource = "dotnet.resilience.failure.source";
 
-    public const string FailureReason = "failure-reason";
+    public const string FailureReason = "dotnet.resilience.failure.reason";
 
-    public const string FailureSummary = "failure-summary";
+    public const string FailureSummary = "dotnet.resilience.failure.summary";
 
-    public const string DependencyName = "dep-name";
+    public const string DependencyName = "dotnet.resilience.dependency.name";
```
martintmk, line 14:
```suggestion
    public const string DependencyName = "dotnet.dependency.name";
```

The idea behind adding these tags to resilience metrics was that we can correlate resilience events with the http metrics. In that sense the name of this tag should be the same as the tag used to track `RequestName` and `DependencyName` for HTTP scenarios.

Do we know how the `RequestMetadata.DependencyName` and `RequestMetadata.RequestName` is represented in Otel world?

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
martintmk: I agree with having just:

- `error.type` - to represent both exception summary and the failure reason for results
- `dotnet.exception.source`

And at initial release we won't add tags for the following properties?

- `FailureResultContext.Source`
- `FailureResultContext.AdditionalInformation`

https://github.com/dotnet/extensions/blob/261b3546ab8280dbecd77efcf063771649397fc3/src/Libraries/Microsoft.Extensions.Resilience/Resilience/FailureResultContext.cs#L13

Maybe we can slim the `FailureResultContext` down then?

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
xakep139: In the current HttpClient metrics there's no such dimensions: https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-system-net#instrument-httpclientrequestduration
Neither OTel specifies this: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpclientrequestduration
I agree we need to unify these dimensions/attributes if they will represent same things in HttpClient metering (hopefully we'll add these bits eventually) and resilience libraries.

xakep139:
> The changes look good to me, but I believe we still need to confirm a few things:
> 
> * do we expect exception summarizer to produce low-cardinality summaries
> * do we still need the failure reason attribute
> * what exactly does CPU utilization metric reports and if it needs `state` attributes (as OTel one)

@geeknoid can you please clarify the expected cardinality that comes from exception summarizer?
@martintmk, looking at your comment https://github.com/dotnet/extensions/pull/4482#discussion_r1367685592, are you saying we can replace `dotnet.resilience.failure.source`, `dotnet.resilience.failure.reason` and `dotnet.resilience.failure.summary` with `error.type` and `dotnet.exception.source`? Did I get it right?

martintmk:
> > The changes look good to me, but I believe we still need to confirm a few things:
> > 
> > * do we expect exception summarizer to produce low-cardinality summaries
> > * do we still need the failure reason attribute
> > * what exactly does CPU utilization metric reports and if it needs `state` attributes (as OTel one)
> 
> @geeknoid can you please clarify the expected cardinality that comes from exception summarizer? @martintmk, looking at your comment [#4482 (comment)](https://github.com/dotnet/extensions/pull/4482#discussion_r1367685592), are you saying we can replace `dotnet.resilience.failure.source`, `dotnet.resilience.failure.reason` and `dotnet.resilience.failure.summary` with `error.type` and `dotnet.exception.source`? Did I get it right?

Correct, this is also based on recommendation by @lmolkova

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
martintmk: Yup correct, these are not strictly tied to resilience so it would be better to not include it in the name to avoid doing breaking changes in the future.

Ideally, this would be the same name as for HTTP metrics.

src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/ResilienceTagNames.cs
martintmk: PR with the changes is up:

#4600

geeknoid:
@xakep139 Yes, the point of the summarizer is specifically to produce low cardinality values devoid of privacy-sensitive information.

xakep139:
We agreed with @martintmk that I will apply naming changes he made in #4600 to this PR, specifically we keep these attributes:
- `request.dependency.name` - neither Resilience nor .NET-specific
- `request.name` - neither Resilience nor .NET-specific
- `error.type`

Everything else is removed from Resilience tag names.

martintmk (approved):

lmolkova:
> request.dependency.name - neither Resilience nor .NET-specific
> request.name - neither Resilience nor .NET-specific

I believe you take request and dependency terms from application Insights, correct?
Request or dependency names in this case are span name and `request.dependency.name` is an oxymoron (request is a server span and dependency is a client span). 

There is no direct translation for span name to a metric tag, I'd say you'd have a metric with the same as span.

If this is not an option, then consider `dotnet.resilience.operation.name` or something that would be specific to this library.
If you have a good suggestion for a genric attribute name, let's bring it up to the otel spec.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringOptions.Linux.cs
```diff
@@ -26,7 +26,7 @@ public partial class ResourceMonitoringOptions
     public TimeSpan CpuConsumptionRefreshInterval { get; set; } = DefaultRefreshInterval;
 
     /// <summary>
-    /// Gets or sets the default interval used for refreshing values reported by <see cref="ResourceUtilizationCounters.MemoryConsumptionPercentage"/>.
+    /// Gets or sets the default interval used for refreshing values reported by <c>"process.memory.virtual.utilization"</c> metrics.
```
xakep139, line 29:
```suggestion
    /// Gets or sets the default interval used for refreshing values reported by <c>"dotnet.process.memory.virtual.utilization"</c> metrics.
```

davidfowl:
Excellent work!

PR #4495, opened 2023-09-29
Enhance path validation logic

sebastienros:
Adding query arguments tests for the chars `?`, `/`, `%`, `#`, ` `

Adding these rules for path segment arguments:

- If the string contains '/' or '\' anywhere, fail.
- If after calling string.Trim():
  - The string is empty, fail.
  - The string contains only '.' characters, fail.
  
This is because we don't think it's legitimate for users to be able to use these as valid path arguments since it could be provided by user data (tenant name, username, ...) and escaping could be tricky as servers might un-encode them.

I am not sure `ArgumentException` is the best here so please send suggestions.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4495)

### DIFF
--- a/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
+++ b/src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
@@ -43,6 +43,8 @@ internal sealed class Emitter : EmitterBase
     private const string AutoClientHttpError = "global::Microsoft.Extensions.Http.AutoClient.AutoClientHttpError";
     private const string Invariant = "global::System.FormattableString.Invariant";
     private const string UriKind = "global::System.UriKind";
+    private const string ArgumentException = "global::System.ArgumentException";
+    private const string AutoClientUtilities = "global::Microsoft.Extensions.Http.AutoClient.AutoClientUtilities";
 
     public string EmitRestApis(IReadOnlyList<RestApiType> restApiTypes, CancellationToken cancellationToken)
     {
@@ -66,6 +68,8 @@ internal sealed class Emitter : EmitterBase
 
         return Capture();
     }
+
+    private static string GetEscapedParamName(string name) => $"{name}Escaped";
 
     private static string GetPathTemplate(RestApiMethod restApiMethod)
     {
@@ -229,10 +233,26 @@ internal sealed class Emitter : EmitterBase
 
         var pathSb = new StringBuilder(restApiMethod.Path);
 
+        // Validate path values
         foreach (var param in restApiMethod.FormatParameters)
         {
-            var escapedParamName = $"{param}Escaped";
-            OutLn($"var {escapedParamName} = {Uri}.EscapeDataString($\"{{{param}}}\");");
+            var escapedParamName = GetEscapedParamName(param);
+
+            // Use interpolated string to handle any null values from any type like nullable value types or reference types.
+            OutLn($"var {escapedParamName} = $\"{{{param}}}\";");
+
+            // Values should not contain '/', '\', be empty, whitespace or made of dots (.) only
+            OutLn($"if (!{AutoClientUtilities}.IsPathParameterValid({escapedParamName})) throw new {ArgumentException}({staticsName}.ValidationExceptionMessage, nameof({param}));");
+            OutLn();
+
+            _ = pathSb.Replace($"{{{param}}}", $"{{{escapedParamName}}}");
+        }
+
+        // Encode path values
+        foreach (var param in restApiMethod.FormatParameters)
+        {
+            var escapedParamName = GetEscapedParamName(param);
+            OutLn($"{escapedParamName} = {Uri}.EscapeDataString({escapedParamName});");
 
             _ = pathSb.Replace($"{{{param}}}", $"{{{escapedParamName}}}");
         }
@@ -242,7 +262,7 @@ internal sealed class Emitter : EmitterBase
         var firstQuery = true;
         foreach (var param in restApiMethod.AllParameters.Where(m => m.IsQuery))
         {
-            var escapedParamName = $"{param.Name}Escaped";
+            var escapedParamName = GetEscapedParamName(param.Name);
 
             // Use interpolated string to handle any null values from any type like nullable value types or reference types.
             OutLn($"var {escapedParamName} = {Uri}.EscapeDataString($\"{{{param.Name}}}\");");
@@ -357,6 +377,7 @@ internal sealed class Emitter : EmitterBase
         OutCloseBrace();
 
         OutCloseBrace();
+
     }
 
     private void EmitSendRequestMethod(string httpClientName, string optionsName)
@@ -439,6 +460,9 @@ internal sealed class Emitter : EmitterBase
     {
         OutLn($"private static class {staticsName}");
         OutOpenBrace();
+
+        OutLn("public const string ValidationExceptionMessage = \"The value can't contain '\\\\', '/', be empty, null or contain only dots (.).\";");
+        OutLn();
 
         OutLn(@$"public static readonly {MediaTypeHeaderValue} ApplicationJsonHeader = new(""application/json"")");
         OutOpenBrace();
--- a/src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
@@ -1,0 +1,68 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.ComponentModel;
+
+namespace Microsoft.Extensions.Http.AutoClient;
+
+/// <summary>
+/// Utilities for AutoClient feature.
+/// </summary>
+[EditorBrowsable(EditorBrowsableState.Never)]
+public static class AutoClientUtilities
+{
+    private static readonly char[] _slashes = { '/', '\\' };
+    private static readonly char[] _slashesOrDot = { '/', '\\', '.' };
+
+    /// <summary>
+    /// Returns whether a value is a valid path argument.
+    /// </summary>
+    /// <param name="value">The value to validate.</param>
+    /// <returns>Whether the value is a valid path argument.</returns>
+    public static bool IsPathParameterValid(string value)
+    {
+        var trimmed = value.AsSpan().Trim();
+
+        if (trimmed.Length == 0)
+        {
+            return false;
+        }
+
+        // Fast path for most common cases
+        var index = trimmed.IndexOfAny(_slashesOrDot);
+        if (index == -1)
+        {
+            return true;
+        }
+
+        // Slashes can't be used
+        if (trimmed.Slice(index).IndexOfAny(_slashes) >= 0)
+        {
+            return false;
+        }
+
+        // The trimmed string can't be made of dots only
+
+#if NET7_0_OR_GREATER
+
+        if (trimmed.IndexOfAnyExcept('.') >= 0)
+        {
+            return true;
+        }
+
+#else
+
+        for (var i = 0; i < trimmed.Length; i++)
+        {
+            if (trimmed[i] != '.')
+            {
+                return true;
+            }
+        }
+
+#endif
+
+        return false;
+    }
+}
--- a/src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.json
+++ b/src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.json
@@ -116,6 +116,16 @@
       ]
     },
     {
+      "Type": "static class Microsoft.Extensions.Http.AutoClient.AutoClientUtilities",
+      "Stage": "Stable",
+      "Methods": [
+        {
+          "Member": "static bool IsPathParameterValid(string value);",
+          "Stage": "Stable"
+        }
+      ]
+    },
+    {
       "Type": "sealed class Microsoft.Extensions.Http.AutoClient.BodyAttribute : System.Attribute",
       "Stage": "Stable",
       "Methods": [
--- a/src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.json
+++ b/src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.json
@@ -120,7 +120,7 @@
       "Stage": "Stable",
       "Methods": [
         {
-          "Member": "static bool IsPathParameterValid(string value);",
+          "Member": "static bool Microsoft.Extensions.Http.AutoClient.AutoClientUtilities.IsPathParameterValid(string value);",
           "Stage": "Stable"
         }
       ]
--- a/test/Generators/Microsoft.Gen.AutoClient/Generated/QueryTests.cs
+++ b/test/Generators/Microsoft.Gen.AutoClient/Generated/QueryTests.cs
@@ -80,6 +80,37 @@ public class QueryTests : IDisposable
         Assert.Equal("Success!", response);
     }
 
+    [Theory]
+    [InlineData("?")]
+    [InlineData("=")]
+    [InlineData("&")]
+    [InlineData("%")]
+    [InlineData("+")]
+    [InlineData("#")]
+    [InlineData(" ")]
+    [InlineData("/")] // / isn't required to be escaped in query string values but there is no harm in doing so
+    public async Task SensitiveCharsAreEscaped(string value)
+    {
+        var encodedValue = Uri.EscapeDataString(value);
+
+        _handlerMock
+            .Protected()
+            .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.Is<HttpRequestMessage>(message =>
+                message.Method == HttpMethod.Get &&
+                message.RequestUri != null &&
+                message.RequestUri.PathAndQuery == $"/api/users?paramQuery={encodedValue}"),
+                ItExpr.IsAny<CancellationToken>())
+            .ReturnsAsync(new HttpResponseMessage
+            {
+                StatusCode = HttpStatusCode.OK,
+                Content = new StringContent("Success!")
+            });
+
+        var response = await _sut.GetUsers(value);
+
+        Assert.Equal("Success!", response);
+    }
+
     [Fact]
     public async Task QueryFromParameterCustom()
     {
--- a/test/Generators/Microsoft.Gen.AutoClient/Generated/PathTests.cs
+++ b/test/Generators/Microsoft.Gen.AutoClient/Generated/PathTests.cs
@@ -83,12 +83,14 @@ public class PathTests : IDisposable
     [Fact]
     public async Task EncodedPathParameters()
     {
+        // "+=&" aren't required to be escaped in path segments but there is no harm in doing so
+
         _handlerMock
             .Protected()
             .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.Is<HttpRequestMessage>(message =>
                 message.Method == HttpMethod.Get &&
                 message.RequestUri != null &&
-                message.RequestUri.PathAndQuery == "/api/users/some%2Fvalue/3"),
+                message.RequestUri.PathAndQuery == "/api/users/some%2B%3D%26value/3"),
                 ItExpr.IsAny<CancellationToken>())
             .ReturnsAsync(new HttpResponseMessage
             {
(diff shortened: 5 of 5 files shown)


src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -43,6 +43,8 @@ internal sealed class Emitter : EmitterBase
     private const string AutoClientHttpError = "global::Microsoft.Extensions.Http.AutoClient.AutoClientHttpError";
     private const string Invariant = "global::System.FormattableString.Invariant";
     private const string UriKind = "global::System.UriKind";
+    private const string ArgumentException = "global::System.ArgumentException";
+    private const string String = "global::System.String";
```
geeknoid, line 47: Should just use 'string' in the generated code.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -66,6 +68,7 @@ public string EmitRestApis(IReadOnlyList<RestApiType> restApiTypes, Cancellation
 
         return Capture();
     }
+    private static string GetEscapedParamName(string name) => $"{name}Escaped";
```
geeknoid, line 71: Add blank line before.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -489,4 +514,58 @@ private void EmitClassVariablesAndConstructor(RestApiType restApiType, string ht
         OutLn($"{optionsName} = autoClientOptions;");
         OutCloseBrace();
     }
+
+    private void EmitPathParameterValidation()
```
geeknoid, line 518: Instead of generating this, could this be a static method in the M.E.EnumStrings package?

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -489,4 +514,58 @@ private void EmitClassVariablesAndConstructor(RestApiType restApiType, string ht
...
+        OutLn($$"""
+            private static bool IsPathParameterValid(string value)
+            {
+                if ({{String}}.IsNullOrWhiteSpace(value))
+                {
+                    return false;
+                }
+
+                #if NET6_0_OR_GREATER
+                if (value.Contains('/'))
```
geeknoid, line 529: Is this the best we can do here? This is doing three full iteration of the string. Aren't there some overloads that take an array of chars to find so we only iterate once?

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -489,4 +514,58 @@ private void EmitClassVariablesAndConstructor(RestApiType restApiType, string ht
...
+            
+                if (value.Contains("\\"))
+                {
+                    return false;
+                }
+            
+                if (value.Contains("."))
+                #endif
+                {
+                    value = value.Trim();
```
geeknoid, line 554: Can we avoid this potential allocation?

geeknoid (commented):
I'm not sure what we're trying to achieve here. This generator is producing a client library that calls an endpoint. Are we trying to protect the endpoint by preventing the client from sending specific strings? Is this really appropriate? The endpoint is the one that should defend against this, right?

What if the client legitimately needs to send those funny strings?

sebastienros:
> Are we trying to protect the endpoint by preventing the client from sending specific strings?

It's not to protect then endpoint, but to protect the application using the generated code from accessing endpoints that it's not supposed to. The code path it's changing uses method arguments to create the destination endpoint. If this argument is not sanitized then it could be formed in a way that the application would call other endpoints that could potentially be abused, leaking other information from the request, or altering another part of the system.

Simple example: your application intends to send a request to delete a product using `/products/{productId}/delete`, but the user of the app manages to pass a `productId` value of `../users/admin` (this could come from a query string in your own app), which could then delete a user account instead on the destination system.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
danmoseley: IndexOfAny.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
danmoseley: ROSpan?

test/Generators/Microsoft.Gen.AutoClient/Generated/QueryTests.cs
```diff
@@ -80,6 +80,37 @@ public async Task QueryIsEscaped()
...
 
+    [Theory]
+    [InlineData("?")]
+    [InlineData("=")]
+    [InlineData("&")]
+    [InlineData("%")]
+    [InlineData("+")]
+    [InlineData("#")]
+    [InlineData(" ")]
+    [InlineData("/")] // / isn't required to be escaped in query string values but there is no harm in doing so
```
danmoseley, line 91: is there any value in a test that puts every other ASCII char in and checks none are escaped?

GrabYourPitchforks: &#x55;&#x6E;&#x6C;&#x69;&#x6B;&#x65;&#x6C;&#x79;&#x2E;&#x20;&#x49;&#x74;&#x27;&#x73;&#x20;&#x70;&#x65;&#x72;&#x66;&#x65;&#x63;&#x74;&#x6C;&#x79;&#x20;&#x76;&#x61;&#x6C;&#x69;&#x64;&#x20;&#x74;&#x6F;&#x20;&#x65;&#x73;&#x63;&#x61;&#x70;&#x65;&#x20;&#x65;&#x76;&#x65;&#x72;&#x79;&#x20;&#x63;&#x68;&#x61;&#x72;&#x61;&#x63;&#x74;&#x65;&#x72;&#x3B;&#x20;&#x74;&#x68;&#x65;&#x20;&#x6F;&#x74;&#x68;&#x65;&#x72;&#x20;&#x65;&#x6E;&#x64;&#x70;&#x6F;&#x69;&#x6E;&#x74;&#x20;&#x77;&#x69;&#x6C;&#x6C;&#x20;&#x64;&#x6F;&#x20;&#x74;&#x68;&#x65;&#x20;&#x72;&#x69;&#x67;&#x68;&#x74;&#x20;&#x74;&#x68;&#x69;&#x6E;&#x67;&#x2E;&#x20;&#x49;&#x6E;&#x20;&#x66;&#x61;&#x63;&#x74;&#x2C;&#x20;&#x65;&#x76;&#x65;&#x72;&#x79;&#x20;&#x73;&#x69;&#x6E;&#x67;&#x6C;&#x65;&#x20;&#x63;&#x68;&#x61;&#x72;&#x61;&#x63;&#x74;&#x65;&#x72;&#x20;&#x6F;&#x66;&#x20;&#x74;&#x68;&#x69;&#x73;&#x20;&#x63;&#x6F;&#x6D;&#x6D;&#x65;&#x6E;&#x74;&#x20;&#x69;&#x73;&#x20;&#x65;&#x73;&#x63;&#x61;&#x70;&#x65;&#x64;&#x20;&#x28;&#x75;&#x73;&#x65;&#x20;&#x74;&#x68;&#x65;&#x20;&#x74;&#x68;&#x72;&#x65;&#x65;&#x20;&#x64;&#x6F;&#x74;&#x73;&#x20;&#x74;&#x6F;&#x20;&#x73;&#x65;&#x65;&#x20;&#x69;&#x74;&#x21;&#x29;&#x2C;&#x20;&#x61;&#x6E;&#x64;&#x20;&#x69;&#x74;&#x20;&#x73;&#x68;&#x6F;&#x75;&#x6C;&#x64;&#x20;&#x72;&#x65;&#x6E;&#x64;&#x65;&#x72;&#x20;&#x63;&#x6F;&#x72;&#x72;&#x65;&#x63;&#x74;&#x6C;&#x79;&#x20;&#x69;&#x6E;&#x20;&#x79;&#x6F;&#x75;&#x72;&#x20;&#x62;&#x72;&#x6F;&#x77;&#x73;&#x65;&#x72;&#x2E;

GrabYourPitchforks:
For context, this is something that came up while threat modeling the AutoClient functionality. When it was presented in the threat modeling session, the scenario provided was that it was trying to simplify the client's ability to invoke these endpoints by allowing the client to perform the equivalent of format string building. Threat modeling asked if it was reasonable to expect that these inputs might come directly from untrusted user input (e.g., provided directly by a remote caller) and thus might be hostile, and the answer was yes.

So we end up with a path traversal-style attack. A malicious client tricks an innocent frontend into making unintended privileged requests to an access-controlled backend. Sebastien talked about this earlier at https://github.com/dotnet/extensions/pull/4495#issuecomment-1743421451. There are two mitigations recommended for this risk.

First, the arguments need to be properly escaped when the URI string is being built. (Sebastien already got to this in some previous PRs.)

Second, certain patterns - even after escaping - may still be dangerous. It's fairly common for web servers to collapse `.` and `..` segments, or to collapse empty segments, or to unescaped `%2F` and treat it as a normal path separator, etc. So if a malicious caller gets one of these patterns into a path segment, it could still lead to path traversal. How you want to approach this depends on the philosophy behind how you're presenting this library to your consumers. If you're positioning this library as something "friendly" that callers don't need much experience to use correctly, then it's prudent to put some defensive checks in place to thwart these types of attacks. If you're positioning this library as something meant for power users and that it's really just a glorified string builder, then you're implicitly expecting your callers to know what values are \[un\]safe to put in any of these segments, and you should document this expectation so that callers _explicitly_ know it.

The security review team was fine with either of the two approaches: defense-in-depth or document warnings for consumers. It's primarily a function of what audience you see consuming this library. Looks like this PR opts for the former.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -489,4 +517,39 @@ private void EmitClassVariablesAndConstructor(RestApiType restApiType, string ht
...
+                while (i < value.Length && value[i] == '.') i++;
+
+                while (i < value.Length && char.IsWhiteSpace(value[i])) i++;
+
+                if (i == value.Length)
+                {
+                    return false;
+                }
+            
+                return true;
```
danmoseley, line 551:
```suggestion
var sp = value.AsSpan().Trim();

return (sp.Length != 0 && sp.IndexOf("\\") == -1 && (sp.IndexOf('.') == -1 || sp.IndexOfAnyExcept('.') == -1))
```
would that work? untested

> // If it contains a dot it should only be made of dots if not ending with whitespaces

I read this as, if it contains a dot the dots must be contiguous and be followed by nothing but possibly whitespace?

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: I can do that, the package is quite empty right now, and I thought it was only for enums manipulation (or that it would go away). Does it still make sense and should I then just add a new static class with this helper?

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: > I read this as, if it contains a dot the dots must be contiguous and be followed by nothing but possibly whitespace?

Correct

> would that work? untested

There is no such `ROS<string>.Trim()`, only one that takes a `char/char[]` to be trimmed. But that would make the whole thing much simpler since the complexity is to prevent using `string.Trim()`.

@GrabYourPitchforks would that be ok to check specifically for `new char[] { ' ', '\t', '\n', '\r' };` or a subset of this even?

GrabYourPitchforks: There are other whitespace chars beyond what you listed. For simplicity, if you want to go this route, you should just call the `Trim` API. It's available across all runtimes `ROS<char>` is available on: https://apisof.net/catalog/65462c87-fbb3-e410-a5b0-e57bcdae22f1

GrabYourPitchforks: FWIW, since you're writing a generator, you may need to invoke the extension method by its fully qualified name rather than as an instance method.

```cs
System.ReadOnlySpan<char> trimmed = System.MemoryExtensions.Trim(value.AsSpan());
```

danmoseley: > There is no such ROS<string>.Trim(),

As Mr G. Y. Pitchforks suggested it’s an expansion method (what I pasted does compile)

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
geeknoid: Sorry, I meant a utility method in M.AspNetCore.AutoClient. This has nothing to do with enum strings.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -440,6 +462,12 @@ private void EmitStatics(RestApiType restApiType, string staticsName)
         OutLn($"private static class {staticsName}");
         OutOpenBrace();
 
+        OutLn("public static readonly char[] Slashes = { '/', '\\\\' };");
```
geeknoid, line 465: Why is this public?

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -489,4 +517,39 @@ private void EmitClassVariablesAndConstructor(RestApiType restApiType, string ht
...
+    {
+        OutLn($$"""
+            private static bool IsPathParameterValid(string value)
+            {
+                if (string.IsNullOrWhiteSpace(value))
+                {
+                    return false;
+                }
+
+                if (value.IndexOfAny({{staticsName}}.Slashes) > -1)
```
geeknoid, line 531: If we assume that 99% of the times, the strings will not contain slashes or dots, can we make a fast trivial exit paths to reduce computation?

if (value.IndexOfAny(SlahesOrDots) < 0)
   return true;

Or something like it.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
```diff
@@ -440,6 +462,12 @@ private void EmitStatics(RestApiType restApiType, string staticsName)
         OutLn($"private static class {staticsName}");
         OutOpenBrace();
 
+        OutLn("public static readonly char[] Slashes = { '/', '\\\\' };");
+        OutLn();
+
+        OutLn("public const string ValidationExceptionMessage = \"The value can't contain '\\\\', '/', be empty, null or contain only dots (.).\";");
```
dbarbosapn, line 468: I'm curious if it's a common pattern across dotnet to make the exception messages as a public constant? Struggling to understand the value...

If it is to allow for testing this exception is thrown, perhaps creating a "AutoClient.RuntimePathValidationException" or something similar would make sense...

dbarbosapn (changes requested):

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: It's part of `private static class {staticsName}` so if it's private the generated code won't be able to access it.

sebastienros: This could be made internal, though the other properties from this private class are public.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: Moved to `AutoClientUtilities.cs` (new file)

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: `IndexOfAnyExcept` is only .NET7+, so substituted it in other cases.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: These have moved to the utilities class now.

src/Generators/Microsoft.Gen.AutoClient/Emitter.cs
sebastienros: Done

sebastienros:
Waiting for approvals or feedback

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Extensions.Http.AutoClient;
+
+
+/// <summary>
```
joperezr, line 9:
```suggestion
namespace Microsoft.Extensions.Http.AutoClient;

/// <summary>
```
NIT: Extra new line causing PR build failure.

sebastienros: Feel free to edit the files directly in GH, I won't mind

joperezr (approved):
Changes look good to me. Thanks for following up on this @sebastienros

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Extensions.Http.AutoClient;
+
+/// <summary>
+/// Utilities for AutoClient feature.
+/// </summary>
+public static class AutoClientUtilities
```
geeknoid, line 11: Let's mark this with EditorBrowsable.Never since it is not intended for use by humans, and only by the code generator.

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            return true;
+        }
+
+        // Slashes can't be used
+        if (trimmed.IndexOfAny(_slashes) >= 0)
+        {
+            return false;
+        }
+
```
geeknoid, line 41:
```suggestion
        // Fast path for most common cases
        var index = trimmed.IndexOfAny(_slashesOrDot);
        if (index == -1)
        {
            return true;
        }

        // Slashes can't be used
        trimmed = trimmed.Slice(index);
        if (trimmed.IndexOfAny(_slashes) >= 0)
        {
            return false;
        }

```

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
sebastienros: Done

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
sebastienros: Had to adapt the suggestion a little so that `trimmed` is not changed or it would have side effects with the last check on dots

src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.json
```diff
@@ -115,6 +115,16 @@
         }
       ]
     },
+    {
```
sebastienros, line 118: @geeknoid is that the correct way to change this file (manually)? or should I use a tool for correctness.

geeknoid: You can change it manually, that's fine.

Normally, I use ./scripts/MakeApiBaselines.ps1 to do this for me.

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+        // Fast path for most common cases
+        var index = trimmed.IndexOfAny(_slashesOrDot);
+        if (index == -1)
+        {
+            return true;
+        }
+
+        // Slashes can't be used
+        if (trimmed.Slice(index).IndexOfAny(_slashes) >= 0)
```
geeknoid, line 40: I think you want:

trimmed = trimmed.Slice();

as a separate statement, such that it applies to both this use of IndexOfAny as well as for the following code below

sebastienros: Did that on purpose or a string like `"a.."` would return `false` while it's fine.

src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.json
sebastienros: I tried again with ApiChief directly (this ps1 failed with some invalid paths) and it worked. I was previously using it on a code generation dll while it should have been use on the main assembly.
Also added the required `-o` argument to generate the json file.

Example:

```
PS C:\code\extensions\eng\Tools\ApiChief> dotnet run -- "C:\code\extensions\artifacts\bin\Microsoft.Extensions.Http.AutoClient\Debug\net8.0\Microsoft.Extensions.Http.AutoClient.dll" emit baseline -o C:\code\extensions\src\Libraries\Microsoft.Extensions.Http.AutoClient\Microsoft.Extensions.Http.AutoClient.json
```

FYI @joperezr since we were not sure it could output this json file too.

src/Libraries/Microsoft.Extensions.Http.AutoClient/AutoClientUtilities.cs
geeknoid: Ah, I see.

geeknoid (approved):

dbarbosapn (approved):

PR #4505, opened 2023-10-04
Reimplemented data classification to support > 64 data classes

mariamgerges:
- In Microsoft.Extensions.Compliance.Redaction, the IRedactionBuilder interface lets you configure the redactor to use for any data classification. Right now, this is populating a FrozenDictionary<DataClassification, Redactor>. Instead, this would need to populate a List<KeyValuePair<FrozenSet<DataClassification>, Redactor>>.
- We'll want to add a new overload to SetRedactor which takes an IReadOnlySet<DataClassification> as input.
- We'll need to add a new overload for RedactorProvider.GetRedactor which takes an IReadOnlySet<DataClassification>
- The logging code generator (here [extensions/src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs at main · dotnet/extensions (github.com)](https://github.com/dotnet/extensions/blob/main/src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs)) combines multiple data classification attributes into a single thing using the OR semantics. This would need to produce a list instead.
- The generated code is filling out a LoggerMessageState struct to pass to the logger. This struct has a Classification property which would need to change from type DataClassification to be instead a FrozenSet<DataClassification>
- The logger code in [extensions/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs at release/8.0 · dotnet/extensions (github.com)](https://github.com/dotnet/extensions/blob/release/8.0/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs) would need to be slightly updated to deal with the change in the LoggerMessageState object in order to call the new overload of GetRedactor.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4505)


RussKie:
@mariamgerges is this needed for .NET 8? If yes, please retarget to release/8.0.

PR #4512, opened 2023-10-04
Adding README for AutoClient package

joperezr:
Marking AutoClient as experimental for November, and adding README.md file so that it contains basic documentation for consumption.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4512)


joperezr:
@sebastienros we may want to add to the readme somewhere that we are now also escaping special characters on parameters in some cases. We probably want to wait for your PR to go in first, and then I can update the README as a follow up PR

geeknoid:
This is against the agreement we reached with Steve and David, we should not be marking anything in dotnet/extensions experimental at this point.

joperezr:
I think there is a misunderstanding here. During last week's tactical sync we agreed that AutoClient was the only assembly which was "okay" to make experimental given the fact that it is in its own assembly and nothing depends on it, and we also chatted with @davidfowl and said we would mark this as stable sometime next year, and you said that would be okay. Am I missing something here?

davidfowl:
I think we said we would leave it as non-experimental and do our best to find any API gaps in the short term, potentially making breaking changes if we need to (which we hopefully don't).

joperezr:
I just checked the recording and we did say we would mark it as experimental but that's okay, maybe you guys had a follow up chat where the decision was reversed 😝. I'll make the required changes here then.

joperezr:
API is no longer experimental. @geeknoid and @dbarbosapn I'd appreciate if we can validate that we are okay with the contents of the README file 😃

src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.csproj
```diff
@@ -3,10 +3,12 @@
     <RootNamespace>Microsoft.Extensions.Http.AutoClient</RootNamespace>
     <Description>Makes it easy to automatically create efficient and easy to use client code to invoke REST HTTP APIs.</Description>
     <Workstream>Fundamentals</Workstream>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
   </PropertyGroup>
 
   <PropertyGroup>
     <InjectExperimentalAttributeOnLegacy>true</InjectExperimentalAttributeOnLegacy>
+    <InjectSharedDiagnosticIds>true</InjectSharedDiagnosticIds>
```
geeknoid, line 11: No longer needed.

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -0,0 +1,217 @@
+# Microsoft.Extensions.Http.AutoClient
+
+This package allows creating efficient HTTP auto clients with minimal code. The client code is generated at compile time, or while editing if analyzers are enabled.
```
geeknoid, line 3:
```suggestion
This package allows creating efficient HTTP clients for existing REST APIs with minimal code. The client code is generated at compile time, is highly efficient, and integrates with modern telemetry patterns.
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -0,0 +1,217 @@
...
+using Microsoft.Extensions.Http.AutoClient;
+
+[AutoClient("MyHttpClient")]
+public interface IMyDependencyClient
+{
+    // ... client methods
+}
+```
+
+By default, the generated client would be of type `MyDependencyClient` (Interface name without the `I`). If you need to use a different name for the dependency name, then you can use the different AutoClient constructor:
```
geeknoid, line 29:
```suggestion
By default, the generated client would be of type `MyDependencyClient` (Interface name without the `I`). If you need to use a different name for the dependency name, then you can use the different AutoClient constructors:
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.csproj
joperezr:
```suggestion
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -0,0 +1,217 @@
+# Microsoft.Extensions.Http.AutoClient
+
+This package allows creating efficient HTTP clients for existing REST APIs with minimal code. The client code is generated at compile time, is highly efficient, and integrates with modern telemetry patterns.
+
+## Getting started
+
+### Install the package
```
Tratcher, line 7: Seems redundant?
```suggestion
## Install the package
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+```dotnetcli
+dotnet add package Microsoft.Extensions.Http.AutoClient
+```
+
+## Usage Example
+
+### Defining a client
```
Tratcher, line 15: Same here:
```suggestion
## Defining a client
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.csproj
```diff
@@ -3,6 +3,7 @@
     <RootNamespace>Microsoft.Extensions.Http.AutoClient</RootNamespace>
     <Description>Makes it easy to automatically create efficient and easy to use client code to invoke REST HTTP APIs.</Description>
     <Workstream>Fundamentals</Workstream>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
```
Tratcher, line 6: No longer needed
```suggestion
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/Microsoft.Extensions.Http.AutoClient.csproj
```diff
@@ -17,6 +18,7 @@
 
   <ItemGroup>
     <None Include="buildTransitive\*" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath="buildTransitive\" />
+    <None Include="README.md" Pack="true" PackagePath="\" />
```
Tratcher, line 21:
```suggestion
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+<ItemGroup>
+  <PackageReference Include="Microsoft.Extensions.Http.AutoClient" Version="[CURRENTVERSION]" />
+</ItemGroup>
+```
+
+## Usage Example
+
+### Defining a client
+
+The first step for generating an auto client, is to define an interface[^1] that describes the client. The interface must be marked with the `[AutoClient]` attribute.
```
danmoseley, line 23:
```suggestion
The first step to generate a client is to define an interface[^1] that describes the client. The interface must be marked with the `[AutoClient]` attribute.
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+Or using minimal APIs:
+
+```csharp
+app.MapGet("/ProcessUsers", async ([FromServices] IMyDependencyClient client) =>
+{
+    return await client.GetUsers(CancellationToken.None);
+});
+```
+
+[^1]: The attribute must be used in interfaces, not classes.
```
danmoseley, line 219:
```suggestion
[^1]: The attribute is only supported on interfaces.
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+<ItemGroup>
+  <PackageReference Include="Microsoft.Extensions.Http.AutoClient" Version="[CURRENTVERSION]" />
+</ItemGroup>
+```
+
+## Usage Example
+
+### Defining a client
+
+The first step to generate a client is to define an interface that describes the client. The interface must be marked with the `[AutoClient]` attribute[^1].
```
dbarbosapn, line 23: I'm wondering if the brackets should be removed or add ("MyHttpClient") here as well. I know there's an example below but quick readers might just see "[AutoClient]" and go with that. No strong opinion though :)

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+public interface IMyDependencyClient
+{
+    [Get("/api/users/{userId}")]
+    public Task<User> GetUser(string userId, CancellationToken cancellationToken = default);
+}
+```
+
+### HTTP Payloads
+
+To send a payload with your request, you mus tuse the `BodyAttribute` on a method parameter. If you don't pass any parameter to it, it will consider the content type is JSON, serializing your parameter before sending. Otherwise, you can choose a `BodyContentType` and use it within the `BodyAttribute`.
```
dbarbosapn, line 102: typo here, "mus**t** use"

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+{
+    [Get("/api/users")]
+    [StaticHeader("X-MethodHeader", "HeaderValue")]
+    public Task<List<User>> GetUsers(CancellationToken cancellationToken = default);
+}
+```
+
+#### Parameter Headers
+
+With the `HeaderAttribute`, you can receive the value for a header from the attributes of your Auto Client method. You must pass the header name to its constructor.
```
dbarbosapn, line 144: I think it's worth mentioning that the value is not serialized! Any complex structures passed as an header value will just have its "ToString()" method called.

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+public interface IMyDependencyClient
+{
+    [Get("/api/users")]
+    public Task<List<User>> GetUsers([Header("X-MyHeader")] string myHeader, CancellationToken cancellationToken = default);
+}
+```
+
+### Query Parameters
+
+Query parameters must be defined using the `QueryAttribute`. If not specified in the attribute's constructor, the parameter name will be used as the query key.
```
dbarbosapn, line 161: Same as for the header values.

dbarbosapn (changes requested):
Just couple comments 👍  Thanks a lot for this contribution 🙇

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+Along with the interface's implementation, extension methods will also be generated to help with registering the Auto Client in the DI container. The extension method will be named `Add{DependencyName}`. For example, if your interface is named `IMyDependencyClient`, the extension method will be `AddMyDependencyClient`.
+
+```csharp
+builder.Services
+    .AddHttpClient("MyHttpClient", client =>    // <-- Adds a named HttpClient which will be matched with the parameter passed to the AutoClientAttribute.
+    {
+        client.BaseAddress = new Uri("https://myapi.com");
+    });
+
+builder.Services.AddMyDependencyClient();               // <-- Registers the Auto Client in the DI container.
```
davidfowl, line 187: ?
```suggestion
builder.Services.AddMyHttpClient();               // <-- Registers the Auto Client in the DI container.
```

src/Libraries/Microsoft.Extensions.Http.AutoClient/README.md
```diff
@@ -1 +1,229 @@
...
+        var users = await _myClient.GetUsers();
+        ...
+    }
+}
+```
+
+Or using minimal APIs:
+
+```csharp
+app.MapGet("/ProcessUsers", async ([FromServices] IMyDependencyClient client) =>
```
davidfowl, line 213:
```suggestion
app.MapGet("/ProcessUsers", async (IMyDependencyClient client) =>
```

geeknoid:
AutoClient is being removed.

PR #4527, opened 2023-10-06
Seed readmes

Tratcher:
I've added props so that a readme is required in every package (it will fail to pack without one) and I've placed the empty files as placeholders.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4527)


sebastienros (approved):

src/Libraries/Microsoft.Extensions.ObjectPool.DependencyInjection/Microsoft.Extensions.ObjectPool.DependencyInjection.csproj
```diff
@@ -20,4 +20,5 @@
   <ItemGroup>
     <InternalsVisibleToTest Include="$(AssemblyName).Tests" />
   </ItemGroup>
+
```
Tratcher, line 23:
```suggestion
```

joperezr:
This is great, thanks @Tratcher! I believe we also want it for the three projects under src/Packages?

Tratcher:
> This is great, thanks @Tratcher! I believe we also want it for the three projects under src/Packages?

Added

src/Directory.Build.props
```diff
@@ -9,6 +9,14 @@
...
   </PropertyGroup>
 
+  <PropertyGroup>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="README.md" Pack="true" PackagePath="\" />
+  </ItemGroup>
+
```
joperezr, line 19:
```suggestion
  <PropertyGroup>
    <ShouldPackReadme Condition="'$(ShouldPackReadme)' == '' And '$(IsPackable)' == 'true' And Exists('$(MSBuildProjectDirectory)\README.md')">true</ShouldPackReadme>
    <PackageReadmeFile Condition="'$(ShouldPackReadme)' == 'true'>README.md</PackageReadmeFile>
  </PropertyGroup>

  <ItemGroup >
    <None Condition="'$(ShouldPackReadme)' == 'true' Include="README.md" Pack="true" PackagePath="\" />
  </ItemGroup>

```
NIT

joperezr: didn't test the above though so if you do take the suggestion it would be good to validate that it works 😄

joperezr (approved):

src/Directory.Build.props
Tratcher: But we want it to fail if there isn't a readme file. Non-packable projects are already not affected.

joperezr: I see. If we want it to fail, usually it's best to have a target that runs post Pack that validates that a readme was added, and uses `<Error` task instead so that we can show a more custom message than just something around `some x file wasn't found`, but I don't want to block the PR on this so whatever you decide is fine.

Tratcher: This seems clear enough?
```
D:\git\extensions\.dotnet\sdk\8.0.100-rc.2.23462.28\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.P
ack.targets(221,5): error NU5019: File not found: 'D:\git\extensions\src\Libraries\Microsoft.AspNetCore.AsyncState\READ
ME.md'. [D:\git\extensions\src\Libraries\Microsoft.AspNetCore.AsyncState\Microsoft.AspNetCore.AsyncState.csproj]
```

joperezr: Yeah, I think that's okay.

PR #4532, opened 2023-10-06
Remove project names

sebastienros:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4532)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ExtraAbstractions/Http/TelemetryConstants.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ExtraAbstractions/Http/TelemetryConstants.cs
@@ -11,7 +11,7 @@ public static class TelemetryConstants
     /// <summary>
     /// Request metadata key that is used when storing request metadata object.
     /// </summary>
-    public const string RequestMetadataKey = "R9-RequestMetadata";
+    public const string RequestMetadataKey = "Extensions-RequestMetadata";
 
     /// <summary>
     /// Represents the placeholder text for an unknown request name or dependency name in telemetry.
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
@@ -12,10 +12,10 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 
 internal static partial class Metric
 {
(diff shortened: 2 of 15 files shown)


src/Libraries/Microsoft.Extensions.Diagnostics.ExtraAbstractions/Http/TelemetryConstants.cs
```diff
@@ -11,7 +11,7 @@ public static class TelemetryConstants
     /// <summary>
     /// Request metadata key that is used when storing request metadata object.
     /// </summary>
-    public const string RequestMetadataKey = "R9-RequestMetadata";
+    public const string RequestMetadataKey = "Extensions-RequestMetadata";
```
joperezr, line 14: I'm assuming we do want to make this change, yet we do expect this to cause breaks for teams currently depending on this. @geeknoid can you confirm?

sebastienros:
I am assuming that some of these suggestions are not good, but I can then add comments to clarify the reasons for later.

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Metric.cs
```diff
@@ -12,10 +12,10 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 
 internal static partial class Metric
 {
-    [Counter("healthy", "status", Name = @"R9\\HealthCheck\\Report")]
```
joperezr, line 15: I know that @xakep139 is also making changes around this on PR https://github.com/dotnet/extensions/pull/4482, so let's just make sure that these changes won't conflict with his.

sebastienros: I see, reverting this one then

joperezr:
Changes look good to me, but it would be good to also get a review from @xakep139 on this especially due to some of the telemetry changes being done here.

xakep139 (approved):

RussKie (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.ExtraAbstractions/Http/TelemetryConstants.cs
geeknoid: This should be fine, it's something teams will have to deal with when upgrading.

geeknoid (approved):

PR #4544, opened 2023-10-09
Fix FakeTimeProvider namespace, add readme

Tratcher:
The `Microsoft.Extensions.Time.Testing` namespace was mismatched to the package name `Microsoft.Extensions.TimeProvider.Testing;`.

Added a readme.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4544)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/README.md
+++ b/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/README.md
@@ -22,12 +22,12 @@ Or directly in the C# project file:
 
 `FakeTimeProvider` can be used to manually adjust time to test time dependent components in a deterministic way.
 
-`FakeTimeProvider` derives from TimeProvider and adds the following APIs:
+`FakeTimeProvider` derives from [TimeProvider](https://learn.microsoft.com/dotnet/api/system.timeprovider) and adds the following APIs:
 
 ```csharp
 public FakeTimeProvider(DateTimeOffset startDateTime)
 public DateTimeOffset Start { get; }
-public TimeSpan AutoAdvanceAmount
+public TimeSpan AutoAdvanceAmount { get; set; }
 public void SetUtcNow(DateTimeOffset value)
 public void Advance(TimeSpan delta)
 public void SetLocalTimeZone(TimeZoneInfo localTimeZone)
--- a/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/README.md
+++ b/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/README.md
@@ -18,7 +18,7 @@ Or directly in the C# project file:
 </ItemGroup>
 ```
 
-## FakeTimeProvider
+## Usage Example
 
 `FakeTimeProvider` can be used to manually adjust time to test time dependent components in a deterministic way.
 
@@ -33,7 +33,7 @@ public void Advance(TimeSpan delta)
 public void SetLocalTimeZone(TimeZoneInfo localTimeZone)
 ```
 
-For example:
+These can be used as follows:
 
 ```csharp
 var timeProvider = new FakeTimeProvider();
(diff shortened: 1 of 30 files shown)


test/Libraries/Microsoft.Extensions.Diagnostics.Probes.Tests/TcpEndpointHealthCheckServiceTests.cs
```diff
@@ -7,7 +7,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging.Testing;
-using Microsoft.Extensions.Time.Testing;
+using Microsoft.Extensions.TimeProvider.Testing;
```
RussKie, line 10: The issue with this namespace that it will clash with the actual type name
image

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/README.md
```diff
@@ -1 +1,47 @@
...
+## FakeTimeProvider
+
+`FakeTimeProvider` can be used to manually adjust time to test time dependent components in a deterministic way.
+
+`FakeTimeProvider` derives from TimeProvider and adds the following APIs:
+
+```csharp
+public FakeTimeProvider(DateTimeOffset startDateTime)
+public DateTimeOffset Start { get; }
+public TimeSpan AutoAdvanceAmount
```
danmoseley, line 30: `{ get; ..}` ?

src/Libraries/Microsoft.Extensions.TimeProvider.Testing/README.md
```diff
@@ -1 +1,47 @@
...
+```csharp
+var timeProvider = new FakeTimeProvider();
+var myComponent = new MyComponent(timeProvider);
+timeProvider.Advance(TimeSpan.FromSeconds(5));
+myComponent.CheckState();
+```
+
+## Feedback & Contributing
+
+For any feedback or contributions, please visit us in [our GitHub repo](https://github.com/dotnet/extensions).
```
danmoseley, line 47:
```suggestion
We welcome feedback and contributions in [our GitHub repo](https://github.com/dotnet/extensions).
```
Maybe? I assume we will copy into all readmes

Tratcher: Yeah, we're still working out some standard format and content.

geeknoid:
Note that I intentionally used the Time namespace instead of TimeProvider, since having a type as a namespace component leads to all sorts of weird/unfortunate confusion with the C# compiler.

Tratcher:
> Note that I intentionally used the Time namespace instead of TimeProvider, since having a type as a namespace component leads to all sorts of weird/unfortunate confusion with the C# compiler.

@geeknoid ok, I can appreciate that. We should still re-consider if A) the package and namespace should align, and B) if there would be some other namespace that would be more appropriate, this is the only type in this namespace and it doesn't correlate with the type being tested/extended (System.TimeProvider). I'm not concerned about putting the type in a more common namespace because the type name is unique and the package will only be added to test projects that need it.

Alternatives:
1. Change the namespace to:
   1. System (Matches TimeProvider)
   2. System.Testing (Adjacent to TimeProvider)
   3. System.Time.Testing (Adjacent to TimeProvider, more narrow scope)
1. Rename the package to Microsoft.Extensions.Time.Testing to match the namespace.

Tratcher:
I've reverted the namespace change for now.

joperezr (approved):

joperezr:
> I've reverted the namespace change for now.

Do you plan on tracking that in the same fit & finish workitem or want to log a new one in favor of closing the fit & fnish one? I'm fine either way, as long as we make sure we revisit this in the next couple of weeks 😃

Tratcher:
Using the same work item is easier.

Issue #4451, opened 2023-09-20
Ensure that the logged dimensions in HTTP logging follow the same naming conventions as defined for spans/metrics


klauco:
Currently, OTel defines semantic conventions for HTTP for spans and metrics and there are none for logs. However, it makes sense to not reinvent the wheel and stick to the naming which is already available even though it's meant for a different pillar. 

**What needs to be done:**
- dimension names used for HTTP and HTTP client log autocollection should be aligned with dimension names used in span/metric semantic conventions

cc @noahfalk @xakep139 @geeknoid

xakep139:
Todo: stick with the .NET dimension names in https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Telemetry/Logging/HttpClientLoggingTagNames.cs

xakep139:
Decide what to do with complex object logging in SG (separator is `_`)

xakep139:
OTel conventions are here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

klauco:
we should check and update remaining enrichers as well (remove env_<partA>_ and use pascalCase)

klauco:
@noahfalk what would you recommend as a delimiter for nested objects when doing complex object logging? "_" (right now) or "." or sth else?

noahfalk:
IMO dots seem marginally more aligned with the conventions. It also aligns better with typical syntax used in many programing languages when doing field access. So barring some particular technical problem I'm unaware of, I'd pick dot.

iliar-turdushev:
The following will be implemented:

* Use `PascalCase` for the names of dimensions and LoggerMessage parameters.
* Use dot "." instead of underscore "_" for HTTP header names and for complex object logging.
* Remove "env_" suffix from the names of enrichment dimensions.

FYI @xakep139 @klauco

joperezr:
@iliar-turdushev are you also planning on looking at logging categories for this and making sure they match the rest of dotnet?

geeknoid:
@Tratcher  FYI

iliar-turdushev:
Reopened the issue, because here we'll address naming of dimension not only in HTTP Logging, but also in HTTP Client logging, Complex Object Logging (aka `LogPropertiesAttribute`), enrichers, and etc. See the following [comment](https://github.com/dotnet/extensions/issues/4451#issuecomment-1748840036).

[...]

PR #4545, opened 2023-10-10
Use the same naming rules for the names of logging dimensions

iliar-turdushev:
The following has been implemented:

* Use [OTel semantic conventions](https://github.com/open-telemetry/semantic-conventions/tree/b127e121b9b458e83a0fee8a32f7eb6721912951) for the names of logging dimensions where possible. Since [semantic conventions for logs](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/general/logs.md) are **Experimental**, and are missed for the most of areas/objects we need to agree on the dimensions names, we decided to use [semantic conventions for traces](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/general/trace.md). Semantic conventions for traces are also **Experimental**, but they are defined for a wide variety of areas/objects.
* If a semantic convention for a dimensions or a log message parameter is missing, then we use `PascalCase` for the name of the dimension.
* Use dot "." instead of underscore "_" as a separator for the names of dimension of `LogPropertiesAttribute` and `TagProviderAttribute`.
* Remove "env_" preffix from the names of enrichment dimensions.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4545)


iliar-turdushev:
@ananthsr If we [rename](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpClientLatencyLogEnricher.cs#L53C13-L54C1) enrichment property `latencyInfo` to `LatencyInfo` (the first letter is Capital), won't it affect latency measurement? Thank you.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
```diff
@@ -15,7 +15,8 @@ namespace Microsoft.Extensions.Http.Logging.Internal;
 internal static partial class Log
 {
     internal const string OriginalFormat = "{OriginalFormat}";
-    internal const string OriginalFormatValue = "{httpMethod} {httpHost}/{httpPath}";
+    internal const string OriginalFormatValue =
+        $"{HttpClientLoggingTagNames.Method} {HttpClientLoggingTagNames.Host}/{HttpClientLoggingTagNames.Path}";
```
xakep139, line 19:
```suggestion
        $"{{HttpClientLoggingTagNames.Method}} {{HttpClientLoggingTagNames.Host}}/{{HttpClientLoggingTagNames.Path}}";
```

klauco:
> @ananthsr If we [rename](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpClientLatencyLogEnricher.cs#L53C13-L54C1) enrichment property `latencyInfo` to `LatencyInfo` (the first letter is Capital), won't it affect latency measurement? Thank you.

let's ensure we align the naming here regardless of the impact.

Tratcher (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.Extra/Enrichment/ApplicationEnricherTags.cs
```diff
@@ -14,22 +14,22 @@ public static class ApplicationEnricherTags
     /// <summary>
     /// Application name.
     /// </summary>
-    public const string ApplicationName = "AppName";
+    public const string ApplicationName = "service.name";
```
xakep139, line 17: Is this lowercased because it's used in metrics only?

src/Libraries/Microsoft.Extensions.Diagnostics.Extra/Enrichment/ProcessEnricherTagNames.cs
```diff
@@ -14,12 +14,12 @@ public static class ProcessEnricherTagNames
...
     /// Process ID.
     /// </summary>
-    public const string ProcessId = "pid";
+    public const string ProcessId = "process.pid";
 
     /// <summary>
     /// Thread ID.
     /// </summary>
-    public const string ThreadId = "tid";
+    public const string ThreadId = "thread.id";
```
xakep139, line 22: Same question applies here

xakep139:
Is the PR still draft?

xakep139:
@iliar-turdushev any chance you checked logging categories we use?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
```diff
@@ -19,7 +19,7 @@ public static class HttpLoggingTagNames
     /// <summary>
     /// HTTP Host.
     /// </summary>
-    public const string Host = "Host";
+    public const string Host = "server.address";
```
iliar-turdushev, line 22: New name https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/http/http-spans.md#http-server-semantic-conventions

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
```diff
@@ -34,12 +34,12 @@ public static class HttpLoggingTagNames
     /// <summary>
     /// HTTP Request Headers prefix.
     /// </summary>
-    public const string RequestHeaderPrefix = "RequestHeader.";
+    public const string RequestHeaderPrefix = "http.request.header.";
```
iliar-turdushev, line 37: New name [http.request.header.](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/http/http-spans.md#common-attributes)

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
```diff
@@ -34,12 +34,12 @@ public static class HttpLoggingTagNames
...
 
     /// <summary>
     /// HTTP Response Headers prefix.
     /// </summary>
-    public const string ResponseHeaderPrefix = "ResponseHeader.";
+    public const string ResponseHeaderPrefix = "http.response.header.";
```
iliar-turdushev, line 42: New name [http.request.header.](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/http/http-spans.md#common-attributes)

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -14,47 +14,47 @@ public static class HttpClientLoggingTagNames
...
     /// HTTP Request duration.
     /// </summary>
-    public const string Duration = "duration";
+    public const string Duration = "Duration";
 
     /// <summary>
     /// HTTP Host.
     /// </summary>
-    public const string Host = "httpHost";
+    public const string Host = "Host";
```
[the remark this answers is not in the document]
xakep139: You meant removal of `http`?

src/Libraries/Microsoft.Extensions.Diagnostics.Extra/Enrichment/ApplicationEnricherTags.cs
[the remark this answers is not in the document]
xakep139: What do you think about putting the link into the code (as a comment)?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
xakep139: We also need to normalize header names:
image

xakep139: It's defined in https://github.com/dotnet/extensions/blob/61a28bead089f12396158b52bca00df48847a732/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HeaderReader.cs#L40

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
xakep139: See the comment above

xakep139: Wait, the link you shared is related to spans. Is it applicable to logs?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
iliar-turdushev, line 1: Some dimensions (e.g. `Host` and `Method`) were not renamed, though they have corresponding names in [OTel's semantic conventions for HTTP spans](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/http/http-spans.md#semantic-conventions-for-http-spans). The reason is that they are coming from ASP.NET Core 8's HttpLogging, therefore we cannot rename them.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
iliar-turdushev: Yeah, at first I just wanted to agree if we are going to follow this conventions, and then make such changes.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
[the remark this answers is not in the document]
iliar-turdushev: Correct, `http`.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
iliar-turdushev, line 1: We were using [Semantic convetions for HTTP Spans](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/http/http-spans.md#semantic-conventions-for-http-spans) for choosing names for the dimensions.

`Duration`, `RequestBody` and `ResponseBody` doesn't have matching semantic conventions. There is a metric for duration of Http Client requests though. But it's a [metric name](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/http/http-metrics.md#metric-httpclientrequestduration) rather than a dimension name, therefore we didn't apply it.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
```diff
@@ -46,17 +46,114 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
         OutgoingRequest(logger, LogLevel.Error, 2, nameof(OutgoingRequestError), record, exception);
     }
 
-    [LoggerMessage(LogLevel.Error, RequestReadErrorMessage)]
-    public static partial void RequestReadError(ILogger logger, Exception exception, HttpMethod httpMethod, string? httpHost, string? httpPath);
+    public static void RequestReadError(ILogger logger, Exception exception, HttpMethod method, string? host, string? path)
```
iliar-turdushev, line 49: New names of HttpClientLogging dimensions contain dots, e.g. `http.request.method`, but LoggerMessage currently don't support such parameters, because the name of a parameter must match the name of an argument of the logging method. Therefore, in order to create a log message with dimensions containing dots, we have to create the message manually.

The code below is a copy for code generated by `LoggerMessage`, I only made changes to the parameter names.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
```diff
@@ -198,12 +198,12 @@ private void CopyTo(Box<T> to)
         }
     }
 
-    [LoggerMessage(LogLevel.Debug, "Can't parse header '{headerName}' due to '{error}'.")]
+    [LoggerMessage(LogLevel.Debug, "Can't parse header '{HeaderName}' due to '{Error}'.")]
```
iliar-turdushev, line 201: I think we shouldn't apply a name [`http.request.header.<key>`](https://github.com/open-telemetry/semantic-conventions/blob/b127e121b9b458e83a0fee8a32f7eb6721912951/docs/attributes-registry/http.md) here, because we are not logging the value of the header here. The aim of this and other logger messages in this file is to log the name of the header, not its value, i.e. the data we should store in a log record should be `[HeaderName] = "name_of_the_header` rather than `["name_of_the_header"] = ???`.

What do you think?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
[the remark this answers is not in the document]
iliar-turdushev: Closing this discussion, since it is not relevant after the latest changes.

src/Libraries/Microsoft.Extensions.Diagnostics.Extra/Enrichment/ApplicationEnricherTags.cs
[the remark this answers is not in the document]
iliar-turdushev: That's a good idea. When we agree on the names, I'll put appropriate links into our code.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
iliar-turdushev: Probably, we'll also need to apply the request header prefix and normalization in the RequestHeaderLogEnricher:
https://github.com/dotnet/extensions/blob/6fda2c8b8f8bacf74eaafff243a10ea20f849aaa/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/RequestHeadersLogEnricher.cs#L68

iliar-turdushev:
> @iliar-turdushev any chance you checked logging categories we use?

@joperezr You asked to check logging categories https://github.com/dotnet/extensions/issues/4451#issuecomment-1749231823.

I checked them, and they look good. What exactly I should pay attention to? Who can give guidance?

Thank you.

iliar-turdushev:
@reyang @noahfalk Please, review this PR. Thank you.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -14,6 +14,9 @@ namespace Microsoft.Gen.Logging.Emission;
 
 internal sealed partial class Emitter : EmitterBase
 {
+    private const string PropertySeparator = ".";
+    private const string NullablePropertySeparator = "?.";
```
geeknoid, line 18: I'm not thrilled with this NUllablePropertySeparator. PropertySeparator is used when we're producing a name string to be used in telemetry, so the string could in fact be anything we want and not break anything. However, NullablePropertySeparator is used to produce source code that references a particular member in the generated code. So that can't readily be changed as it would break the generated code.

Maybe just add some comments to deacribe what these are for...

src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
geeknoid: Agreed.

Why change the logging messages from headerName to HeaderName?

geeknoid (commented):
Why do we need to change all the logging templates from {foo} to {Foo}?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
geeknoid: Let's file a feature request to introduce a [TagName] attribute that the code generator will take the final name from.

iliar-turdushev:
> Why do we need to change all the logging templates from {foo} to {Foo}?

One of the goals of this PR is to be aligned with naming rules/conventions being used in the rest of dotnet, in case there is no a corresponding OTel semantic-convention. In dotnet a `PascalCase` is being used for the names of logging dimensions and parameters of logger messages. For example: https://github.com/dotnet/aspnetcore/blob/main/src/Middleware/HttpLogging/src/HttpLoggingExtensions.cs#L23C1-L24C93.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
iliar-turdushev: I didn't know that the string constants were inlined intentionally, therefore when making changes I extracted them into constant fields following "best practices". I removed the constant fields and inlined the constant strings.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
xakep139: +1

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
xakep139: If we agree on the change, it should be also applied to HttpClient logging of headers

xakep139:
@iliar-turdushev the PR LGTM, the only missing bit is header names normalization (if we agree on the naming)

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
iliar-turdushev: Create an issue https://github.com/dotnet/extensions/issues/4576.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -14,47 +14,47 @@ public static class HttpClientLoggingTagNames
...
     /// HTTP Request Body.
     /// </summary>
-    public const string RequestBody = "httpRequestBody";
+    public const string RequestBody = "RequestBody";
 
     /// <summary>
     /// HTTP Response Body.
     /// </summary>
-    public const string ResponseBody = "httpResponseBody";
+    public const string ResponseBody = "ResponseBody";
```
dariusclay, line 42: even though it's not part of semantic conventions, why aren't we following the same naming rules? ie: `http.response.body`

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
iliar-turdushev: That's what we agreed on:
* if a dimension is a part of the semantic conventions, then we use it;
* otherwise - we use dotnet naming rules for logging dimensions, i.e. `PascalCase`.

joperezr:
@xakep139 @iliar-turdushev Other than fixing code coverage, is this ready to go in?

iliar-turdushev:
> @xakep139 @iliar-turdushev Other than fixing code coverage, is this ready to go in?

@joperezr We are waiting for review from @reyang and @noahfalk. Also, if we agree to follow OTel's HTTP Request/Response header name semantics, then we'll also need to implement header name normalization.

noahfalk (approved):
This seems fine to me, but for anything nuanced about the conventions I probably wouldn't catch it. If you want a higher level of scrutiny, someone like Liudmila is probably a better reviewer than I.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Microsoft.Extensions.Http.Diagnostics.csproj
```diff
@@ -27,7 +27,7 @@
 
   <PropertyGroup>
     <Stage>normal</Stage>
-    <MinCodeCoverage>94</MinCodeCoverage>
+    <MinCodeCoverage>91</MinCodeCoverage>
```
iliar-turdushev, line 30: Code coverage has dropped because we had to replace code generated logger messages with their actual implementations (i.e. see [here](https://github.com/dotnet/extensions/pull/4545/files#diff-01befa772fdd79453ebf661760fa2509b4cfc9f0cbe25c7644cc1680c8678ef1R129)) because of the following limitation of the `LoggerMessage` #4576. As soon as we implement #4576, we increase the code coverage again.

iliar-turdushev:
> This seems fine to me, but for anything nuanced about the conventions I probably wouldn't catch it. If you want a higher level of scrutiny, someone like Liudmila is probably a better reviewer than I.

@lmolkova Please, review the PR. Thank you.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -14,47 +14,47 @@ public static class HttpClientLoggingTagNames
...
     /// HTTP Request duration.
     /// </summary>
-    public const string Duration = "duration";
+    public const string Duration = "Duration";
 
     /// <summary>
     /// HTTP Host.
     /// </summary>
-    public const string Host = "httpHost";
+    public const string Host = "server.address";
```
lmolkova, line 22: note that HTTP Host header is a combination of host and port. If you want to capture the host header as it, it's best to use `http.request.header.host` attribute instead.

Also, `server.address` is not always derived from the host header. It may be derived from `Forwareded` or `X-Forwarded-Host`.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
lmolkova: It looks inconsistent and it'd be great to have the same naming conventions across different attributes. This way user at least won't need to think about casing when writing their queries against logs.

As a side-note, OTel is now incorporating a lot of attributes from ECS - https://www.elastic.co/guide/en/ecs/current/ecs-http.html. There is no promise they will go as is, but they have a lot of relevant attribute such as [http.request.body.bytes](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-request-body-bytes) or [http.request.body.content](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-request-body-content)

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
```diff
@@ -46,17 +46,114 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
...
+    {
+        var state = LoggerMessageHelper.ThreadLocalState;
+
+        _ = state.ReserveTagSpace(4);
+        state.TagArray[3] = new(HttpClientLoggingTagNames.Method, method);
+        state.TagArray[2] = new(HttpClientLoggingTagNames.Host, host);
+        state.TagArray[1] = new(HttpClientLoggingTagNames.Path, path);
+        state.TagArray[0] = new(OriginalFormat, RequestReadErrorMessage);
```
lmolkova, line 57: would be nice to add [exception.type](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-logs.md)  and other `exception` attributes

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
```diff
@@ -46,17 +46,114 @@ public static void OutgoingRequestError(ILogger logger, LogRecord record, Except
...
+    {
+        var state = LoggerMessageHelper.ThreadLocalState;
 
-    [LoggerMessage(LogLevel.Error, LoggerContextMissingMessage)]
-    public static partial void LoggerContextMissing(ILogger logger, Exception? exception, string requestState, HttpMethod httpMethod, string? httpHost);
+        _ = state.ReserveTagSpace(4);
+        state.TagArray[3] = new(HttpClientLoggingTagNames.Method, method);
+        state.TagArray[2] = new(HttpClientLoggingTagNames.Host, host);
+        state.TagArray[1] = new(HttpClientLoggingTagNames.Path, path);
+        state.TagArray[0] = new(OriginalFormat, ResponseReadErrorMessage);
```
lmolkova, line 84: same here and in other places where exception is reported - `exception.*` attributes would be awesome

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
```diff
@@ -45,10 +46,17 @@ public static void AddResponseHeaders(this LoggerMessageState state, List<KeyVal
...
 
             state.TagArray[index++] = new(key, items[i].Value);
         }
     }
+
+    [SuppressMessage("Globalization", "CA1308:Normalize strings to uppercase",
+        Justification = "Normalization to lower case is required by OTel's semantic conventions")]
+    private static string Normalize(string header)
+    {
+        return header.ToLowerInvariant().Replace('-', '_');
```
lmolkova, line 60: same here - no need for `Replace('-', '_')` anymore

lmolkova (commented):
Left a few minor comments on naming.

I'm a bit concerned about the inconsistent naming patterns. otel.attributes and PascalCaseAttributes in the same log record would look messy.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
lmolkova, line 42: It looks inconsistent and it'd be great to have the same naming conventions across different attributes. This way user at least won't need to think about casing when writing their queries against logs.

As a side-note, OTel is now incorporating a lot of attributes from ECS - https://www.elastic.co/guide/en/ecs/current/ecs-http.html. There is no promise they will go as is, but they have a lot of relevant attribute such as [http.request.body.bytes](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-request-body-bytes) or [http.request.body.content](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-request-body-content)

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
iliar-turdushev: This was implemented in another PR #4584.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
iliar-turdushev: This was implemented in another PR #4584.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
iliar-turdushev: We are reading the host from a request's Host property which is the name of the host without port information:

https://github.com/dotnet/extensions/blob/d74d9f4915cf6a1bdd38f784fcbda2861312bd8e/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs#L93

So, as I understand, in our case `Host` is exactly `server.address` dimensions from the semantic-convetions.

iliar-turdushev:
> Left a few minor comments on naming.
> 
> I'm a bit concerned about the inconsistent naming patterns. otel.attributes and PascalCaseAttributes in the same log record would look messy.

@lmolkova Thank you very much for reviewing the PR.

Yeah, we agree, but, unfortunately, there is no perfect solution here. In our opinion, the best we can do is to follow OTel's semantic-conventions where possible. The rest, which is not yet present in OTel or cannot be aligned with OTel due to implementation details in .NET, will be aligned with .NET's naming rules. This, at least, will allow us to leverage some benefits of following OTel's schema, for example, integration with tools built on top of it.

cc @klauco

lmolkova (approved):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
iliar-turdushev: > As a side-note, OTel is now incorporating a lot of attributes from ECS - https://www.elastic.co/guide/en/ecs/current/ecs-http.html. There is no promise they will go as is, but they have a lot of relevant attribute such as [http.request.body.bytes](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-request-body-bytes) or [http.request.body.content](https://www.elastic.co/guide/en/ecs/current/ecs-http.html#field-http-request-body-content)

Thank you for the information. For now we want to follow what OTel conventions provide, and don't want not to bring confusion by following both OTel and ECS, since if we bring some names from ECS then someone can reasonably ask why we are following ECS here, and not there.

PR #4553, opened 2023-10-11
[release/8.0] Update baselines

sebastienros:
There is a discrepancy between release/8.0 and main for these files, this is updating the one in release/8.0 to conform with the actual code in the same branch.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4553)


RussKie (approved):

joperezr:
Do we understand why these got into this state? Is it just that we've updated these in main and haven't backported to release/8.0?

PR #4556, opened 2023-10-11
Hosting.Testing readme

Tratcher:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4556)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
+++ b/src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
@@ -2,11 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Hosting;
-using Microsoft.Shared.DiagnosticIds;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.Hosting.Testing;
@@ -14,7 +11,6 @@ namespace Microsoft.Extensions.Hosting.Testing;
 /// <summary>
 /// Unit testing friendly configured host.
 /// </summary>
-[Experimental(diagnosticId: Experiments.Hosting, UrlFormat = Experiments.UrlFormat)]
 public sealed class FakeHost : IHost
 {
     /// <summary>
--- a/src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
+++ b/src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
@@ -2,9 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Shared.DiagnosticIds;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.Hosting.Testing;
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
```diff
@@ -2,16 +2,19 @@
...
 using Microsoft.Extensions.Hosting;
+using Microsoft.Shared.DiagnosticIds;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.Hosting.Testing;
 
 /// <summary>
 /// Unit testing friendly configured host.
 /// </summary>
+[Experimental(diagnosticId: Experiments.Hosting, UrlFormat = Experiments.UrlFormat)]
```
joperezr, line 17: @geeknoid can you confirm that this is correct?

src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHostOptions.cs
```diff
@@ -2,13 +2,16 @@
...
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Shared.DiagnosticIds;
 
 namespace Microsoft.Extensions.Hosting.Testing;
 
 /// <summary>
 /// Options to configure <see cref="FakeHost"/>.
 /// </summary>
+[Experimental(diagnosticId: Experiments.Hosting, UrlFormat = Experiments.UrlFormat)]
```
joperezr, line 14: @geeknoid same here.

joperezr (commented):
Other than making sure the Experimental attributes are right, this LGTM.

src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
Tratcher: Note these were added to match the manifest. If it's wrong, we'll need to update the manifest.
https://github.com/dotnet/extensions/blob/6fda2c8b8f8bacf74eaafff243a10ea20f849aaa/src/Libraries/Microsoft.Extensions.Hosting.Testing/Microsoft.Extensions.Hosting.Testing.json#L5-L6

joperezr: Got it, I just want to make sure we understand which one is correct, manifest or code. Given this is a Fake, it wasn't immediately obvious to me why this wouldn't be stable.

geeknoid: Oh actually, this is all working at expected.

The whole package is experimental as we have Stage=dev in the .csproj. Thus, the experimental attribute is being applied at the level of the assembly and doesn't need to be applied to each type/member. So you can safely remove this experimental annotation.

geeknoid (approved):
Just remove the changes to the .cs files please, those aren't needed.

geeknoid (approved):
Just remove the changes to the .cs files please, those aren't needed.

src/Libraries/Microsoft.Extensions.Hosting.Testing/FakeHost.cs
joperezr: I see, that makes sense. I didn't know that we had infrastructure setup for injecting the attribute to the assembly but it does seem to be working then.

src/Libraries/Microsoft.Extensions.Hosting.Testing/README.md
```diff
@@ -1 +1,52 @@
-README
+# Microsoft.Extensions.Hosting.Testing
+
+Tools for integration test hosting and related test oriented helpers.
```
danmoseley, line 3: .. of ASP.NET Core apps?

also is it worth mentioning the Microsoft.AspNetCore.Testing package here, and vice versa in that readme?

danmoseley: actually not necessarily of ASP.NET Core apps,  but of apps built with Microsoft.Extensions.Hosting. Anyway -- it should say it's not totally generic.

PR #4561, opened 2023-10-12
AspNetCore.Testing readme, namespaces

Tratcher:
- Add ReadMe
- Fix extension method namespaces (and split the class by namespace).
- ~~Marked APIs as experimental to match the manifest~~ Updated the manifests.
- Removed a bad dependency reference.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4561)


src/Libraries/Microsoft.AspNetCore.Testing/ServiceFakesHostExtensions.cs
```diff
@@ -5,71 +5,26 @@
...
 using Microsoft.Shared.Diagnostics;
 
-namespace Microsoft.AspNetCore.Testing;
+namespace Microsoft.Extensions.Hosting;
 
 /// <summary>
 /// Extension methods supporting Kestrel server unit testing scenarios.
 /// </summary>
-public static class ServiceFakesExtensions
+[Experimental(diagnosticId: Experiments.AspNetTesting, UrlFormat = Experiments.UrlFormat)]
```
joperezr, line 23: CC: @geeknoid to confirm this.

src/Libraries/Microsoft.Extensions.Hosting.Testing/Microsoft.Extensions.Hosting.Testing.csproj
```diff
@@ -26,7 +26,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Bcl.TimeProvider" />
-    <xPackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
```
joperezr, line 29: Good catch.

joperezr (commented):
Changes LGTM and I like the separation as it allows to make sure we get the right namespaces. Just waiting on @geeknoid to confirm on the addition of the experimental attributes.

src/Libraries/Microsoft.Extensions.Hosting.Testing/Microsoft.Extensions.Hosting.Testing.csproj
geeknoid: whops, that's a leftover when I was trimming the dependency set across the repo.

src/Libraries/Microsoft.AspNetCore.Testing/ServiceFakesHostExtensions.cs
geeknoid: No need for [Experimental] here.

For renames, until November 2023, you can just do the rename and then run scripts/MakeApiBaselines.ps1 to update the manifest file.

After November, this would be a breaking change so then the process will be different.

src/Libraries/Microsoft.AspNetCore.Testing/README.md
```diff
@@ -1 +1,50 @@
-README
+# Microsoft.AspNetCore.Testing
+
+This package provides test fakes for integration testing AspNetCore applications.
```
danmoseley, line 3:
```suggestion
This package provides test fakes for integration testing ASP.NET Core applications.
```

activate.ps1
```diff
@@ -0,0 +1,66 @@
+#
+# This file must be used by invoking ". .\activate.ps1" from the command line.
```
Tratcher, line 2: Copied from AspNetCore, needed to run the baseline update scripts.

joperezr: Will this still be required after https://github.com/dotnet/extensions/pull/4587? It seems that the changes going there will fix the issues you were hitting, so I would rather not add this script if not required any longer.

Tratcher: Ah, that would help. It's a tradeoff though, do you write one script for activation, or do you code it into every other script? It's also useful if you want to call any dotnet commands directly.

joperezr (approved):
:shipit:

geeknoid (approved):

activate.ps1
RussKie: If you use `start-code.cmd` then you can run `dotnet` directly from the VS Code's terminal as the script bootstraps all the necessary variables.

PR #4593, opened 2023-10-20
Revamp data classification model.

### LINKED ISSUE #4493: Reimplement data classification to support > 64 data classes
The existing classification model only supports up to 64 claasifications, but alas GDPR has a taxonomy with hundreds of things in it. So we need to extend the model accordingly.

Right now, individual classifications hold a 64-bit bitmask to represent classes. We can OR classifications together by just oring the bit masks. For the new model to work, we'll need to removing this OR semantics and instead lug around sets of classifications instead.

geeknoid:
Fixes #4493

- Introduce the DataClassificationSet type that represents a set of combined classifications. This replaces the previous ability to OR together multiple classifications, which was too constraining.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4593)


src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassification.cs
```diff
@@ -2,25 +2,24 @@
...
 
 using System;
-using System.Diagnostics.CodeAnalysis;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.Compliance.Classification;
 
 /// <summary>
-/// Represents a set of data classes as a part of a data taxonomy.
+/// Represents a single data classes which is part of a data taxonomy.
```
danmoseley, line 10:
```suggestion
/// Represents a single data class which is part of a data taxonomy.
```

geeknoid: Done.

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassificationSet.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// </summary>
+    /// <param name="classification">The classification to include in the set.</param>
+    /// <returns>The resulting data classification set.</returns>
+    public static DataClassificationSet FromDataClassification(DataClassification classification) => new(classification);
+
+    /// <summary>
+    /// Creates a new data classification that combines the current classifications with another set.
+    /// </summary>
+    /// <param name="other">The other set.</param>
+    /// <returns>The resulting set which combines the current instance's classifications and the other ones.</returns>
```
xakep139, line 68: Maybe put a remark that this function doesn't change the current set and returns a new copy with the result?

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassificationSet.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// <summary>
+    /// Returns a string representation of this object.
+    /// </summary>
+    /// <returns>The string representation of this object.</returns>
+    public override string ToString()
+    {
+        var sb = PoolFactory.SharedStringBuilderPool.Get();
+
+        var a = _classifications.ToArray();
+        Array.Sort(a, (x, y) =>
```
xakep139, line 116: Is sorting strongly needed here?

geeknoid: I think so, otherwise ToString would produce non-deterministic results.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -559,23 +559,25 @@ bool GenVariableAssignments(LoggingMethod lm, string lambdaStateName, int numRes
...
+
+        int i = 0;
+
         foreach (var ct in classificationTypes)
         {
-            if (sb.Length > 0)
+            if (i > 0)
             {
-                _ = sb.Append(" | ");
+                _ = sb.Append(" , ");
```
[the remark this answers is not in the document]
geeknoid: Done.

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassification.cs
```diff
@@ -110,48 +93,9 @@ public override int GetHashCode()
...
-    {
-        return Or(left, right);
-    }
-
     /// <summary>
     /// Gets a string representation of this object.
     /// </summary>
     /// <returns>A string representing the object.</returns>
-    public override string ToString() => $"{TaxonomyName}:{Value:x}";
+    public override string ToString() => $"{TaxonomyName}:{Value}";
```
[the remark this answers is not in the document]
geeknoid: Done.

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassificationSet.cs
```diff
@@ -0,0 +1,144 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Shared.Diagnostics;
+using Microsoft.Shared.Pools;
+
+namespace Microsoft.Extensions.Compliance.Classification;
+
+/// <summary>
+/// Represents a set of data classes.
```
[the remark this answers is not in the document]
geeknoid: Should be fixed, please let me know if you spot additional inconsistencies.

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassificationSet.cs
geeknoid: Done.

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassificationSet.cs
xakep139: There's no mention or requirement in the guidelines for `ToString()` to be deterministic or not: https://learn.microsoft.com/en-us/dotnet/api/system.object.tostring?view=net-7.0#notes-to-inheritors

IMO the `ToString()` shouldn't allocate and do sorting every time. Can we maybe cache the resulted string once it's constructed and return it in subsequent calls? There's no way to modify the `_classifications`, right?

geeknoid: We could cache the string, but I'm not sure it's worth it. I expect this ToString to be effectively never be used, except via the debugger. I could just remove ToString completely...

geeknoid: Do note that many ToString calls allocate in order to form a string. And note that the vast vast majority of cases of DataClassificationSet will have a single element in it, so sorting will be a noop.

dariusclay (approved):
LGTM

src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Classification/DataClassificationSet.cs
xakep139: Okay, thanks

PR #4600, opened 2023-10-23
Drop FailureResultContext

martintmk:
When going over #4482 I realized that `FailureResultContext` is something that is not required with Polly v8. 

The idea behind `FailureResultContext` was to allow to customize the metering dimensions based on the result type. However, Polly v8 supports [telemetry enrichment](https://www.pollydocs.org/advanced/telemetry.html#usage) natively and this API feels unnecessary now.

Instead of using `FailureResultContext`, the consumer of Polly can just register their own `MeteringEnricher` and use it to extract the result and add whatever dimensions they want.
 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4600)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
@@ -19,7 +19,7 @@
 
   <PropertyGroup>
     <Stage>normal</Stage>
-    <MinCodeCoverage>98</MinCodeCoverage>
+    <MinCodeCoverage>97</MinCodeCoverage>
     <MinMutationScore>100</MinMutationScore>
   </PropertyGroup>
 
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
@@ -39,9 +39,7 @@ public static partial class ResilienceHttpClientBuilderExtensions
         _ = Throw.IfNullOrEmpty(pipelineName);
         _ = Throw.IfNull(configure);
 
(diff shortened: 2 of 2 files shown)


src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/HttpResilienceTagNames.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.Http.Resilience.Internal;
+
+internal static class HttpResilienceTagNames
+{
+    public const string FailureReason = "failure-reason";
```
xakep139, line 8: `dotnet,resilience.*`?

martintmk: This will become `error.type` right?

martintmk:
cc @joperezr

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -20,7 +19,7 @@
 
   <PropertyGroup>
     <Stage>normal</Stage>
-    <MinCodeCoverage>98</MinCodeCoverage>
+    <MinCodeCoverage>97</MinCodeCoverage>
```
martintmk, line 22: Should be at 100, but code coverage is reporting lower number :/

RussKie: It's been reported as
image

Because of this:
image

RussKie: ...though the validation script reports it as 97.38%...
image

PR #4613, opened 2023-10-24
Add comments pointing to the source of logging dimension names

iliar-turdushev:
The PR adds links to OTel semantic conventions that are used for naming of logging dimensions.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4613)

### DIFF
--- a/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
+++ b/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
@@ -9,6 +9,12 @@ namespace Microsoft.AspNetCore.Diagnostics.Logging;
 /// <summary>
 /// Constants used for incoming HTTP request logging tags.
 /// </summary>
+// There is no OTel semantic convention for HTTP logs, therefore we are using
+// semantic conventions for HTTP spans:
+// https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md.
+// Some dimensions (e.g. Host and Method) were not renamed, though they had corresponding
+// names in the semantic conventions. The reason is that they are coming from ASP.NET Core 8
+// HttpLogging component, therefore we cannot rename them.
 public static class HttpLoggingTagNames
 {
     /// <summary>
@@ -19,6 +25,7 @@ public static class HttpLoggingTagNames
     /// <summary>
     /// HTTP Host.
     /// </summary>
+    // See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-server-semantic-conventions.
     public const string Host = "server.address";
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


iliar-turdushev:
The PR is ready to be merged.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
```diff
@@ -9,6 +9,12 @@ namespace Microsoft.AspNetCore.Diagnostics.Logging;
 /// <summary>
 /// Constants used for incoming HTTP request logging tags.
 /// </summary>
+// There is no OTel semantic convention for HTTP logs, therefore we are using
+// semantic conventions for HTTP spans:
+// https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md.
+// Some dimensions (e.g. Host and Method) were not renamed, though they had corresponding
+// names in the semantic conventions. The reason is that they are coming from ASP.NET Core 8
+// HttpLogging component, therefore we cannot rename them.
```
RussKie, line 17: Is this something we'd want end-users to see in the docs (e.g., at learn.microsoft), or is this for the maintainers only?
Ditto for the other comments.

geeknoid:
The PR mingles /// and // comments together, which isn't great. I think most of these added comments can go into /// as `<remarks>` sections.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/HttpLoggingTagNames.cs
iliar-turdushev: I didn't mean to include this comment as public documentation, and my initial intent was to include it as a reference and justification "WHY" only for maintainers. It is part of public API though, and such a comment/remark could be useful for users.

iliar-turdushev: Regarding web doc, e.g. learn.microsoft, I think that we'll include into the doc a list of log dimensions and remark that some of those dimensions are following OTel semantic-conventions.

iliar-turdushev:
> The PR mingles /// and // comments together, which isn't great. I think most of these added comments can go into /// as `<remarks>` sections.

@geeknoid 

My initial intent was to include this kind of information only for maintainers, hence not as part of public API doc. We are currently following OTel conventions for **spans**, because there are no conventions for **logs**. It could be confusing to include links to span conventions into API documentation of Http and HttpClient logging.

We'll include information that the names of logging tags were inspired by OTel semantic conventions for spans into web documentation, e.g. learn.microsoft, and add there a justification why we are using span conventions.

I'm going to abandon the PR.

What do you think?

geeknoid:
Yeah, I think abandoning is reasonable.

PR #4619, opened 2023-10-25
Add readme for Ambient Metadata

Tratcher:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4619)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AmbientMetadata.Application/README.md
+++ b/src/Libraries/Microsoft.Extensions.AmbientMetadata.Application/README.md
@@ -1,1 +1,67 @@
-README
+# Microsoft.Extensions.AmbientMetadata.Application
+
+This provides runtime information for application-level ambient metadata such as the version, deployment ring, environment, and name.
+
+## Install the package
+
+From the command-line:
+
+```dotnetcli
+dotnet add package Microsoft.Extensions.AmbientMetadata.Application
+```
+
+Or directly in the C# project file:
+
+```xml
+<ItemGroup>
+  <PackageReference Include="Microsoft.Extensions.AmbientMetadata.Application" Version="[CURRENTVERSION]" />
+</ItemGroup>
+```
+
+## Usage Example
+
+### Registering Services
+
+The services can be registered using any of the following methods:
+
+```csharp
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AmbientMetadata.Application/README.md
```diff
@@ -1 +1,67 @@
-README
+# Microsoft.Extensions.AmbientMetadata.Application
+
+This provides runtime information for application-level ambient metadata such as the version, deployment ring, environment, and name.
```
geeknoid, line 3: Should we motivate this a bit? The design intent is that this supplies state that is used by telemetry enrichers down the line.

joperezr: > Should we motivate this a bit?

@geeknoid can you explain what you mean by this?

geeknoid: I mean this is kind of a strange component, I know David asked about it multiple times.

This component exists to capture data from the environment and populate it in DI in a convenient form where it can be consumed by enrichers. The design makes it possible to easily manually supply data to enable effective testing.

joperezr: I see, so are you saying we don't expect folks to depend on this directly, and instead depend on some of the Diagnostic or Telemetry packages that provide enrichment and for this to be a transitive dependency?

geeknoid: @tekian How is this used in practice by our customers? Do they depend on it explicitly, or does it come in for free? I think customers still need to explicitly initialize the component themselves, rigth?

Tratcher:
```suggestion
This flows runtime information for application-level ambient metadata such as the version, deployment ring, environment, and name. This information can be useful to enrich telemetry and logs.
```

geeknoid (approved):

PR #4621, opened 2023-10-26
Adding README for Microsoft.Extensions.EnumStrings

joperezr:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4621)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.EnumStrings/README.md
+++ b/src/Libraries/Microsoft.Extensions.EnumStrings/README.md
@@ -58,7 +58,7 @@ Which you can later use as follows:
 
 Both calls will return the same string, but the second one will be more efficient.
 
-### Using the generated with an enum from a different assembly
+### Using the generator with an enum from a different assembly
 
 You can also use this feature to generate efficient `ToInvariantString` extension methods on enums that are not part of your code. You can do this by applying the `EnumStrings` attribute at the assembly level, and passing in the type of the enum you want to generate the extension method for. For example:
 
--- a/src/Libraries/Microsoft.Extensions.EnumStrings/README.md
+++ b/src/Libraries/Microsoft.Extensions.EnumStrings/README.md
@@ -1,6 +1,6 @@
 # Microsoft.Extensions.EnumStrings
 
-It's often necessary to convert a C# enum value to a string. The standard ToString() is normally used for this, but that can not be the best in terms of allocations or performance. This package introduces a source generator that is triggered by an attribute which produces an extension method for each enum type annotated with the attribute which can efficiently convert an enum value to a string.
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.EnumStrings/README.md
```diff
@@ -1 +1,80 @@
-README
+# Microsoft.Extensions.EnumStrings
+
+It's often necessary to convert a C# enum value to a string. The standard ToString() is normally used for this, but that can not be the best in terms of allocations or performance. This package introduces a source generator that is triggered by an attribute which produces an extension method for each enum type annotated with the attribute which can efficiently convert an enum value to a string.
```
sebastienros, line 3:
```suggestion
It's often necessary to convert a C# enum value to a string. The standard `ToString()` method is normally used for this, but that can not be the best in terms of allocations or performance. This package introduces a source generator that is triggered by an attribute which produces an extension method for each `enum` type annotated with the attribute which can efficiently convert an enumeration value to a string.
```

Pick whatever suggestion

src/Libraries/Microsoft.Extensions.EnumStrings/README.md
```diff
@@ -1 +1,80 @@
...
+```
+
+## Usage Example
+
+### Adding the attribute to an enum
+
+To use this feature, you can annotate an enum value with the `EnumStringAttribute` attribute. For example:
+
+```csharp
+using Microsoft.Extensions.EnumStrings;
```
joperezr, line 28: @geeknoid should we consider moving the EnumStrings attribute to the usual `System.Runtime.CompilerServices` namespace where we put the rest of the attributes in .NET in favor of a future move of this down to dotnet/runtime?

joperezr: cc: @terrajobst @stephentoub

geeknoid: We certainly can.

geeknoid (approved):

src/Libraries/Microsoft.Extensions.EnumStrings/README.md
```diff
@@ -1 +1,80 @@
...
+    var color = Color.Red;
+
+    var s1 = color.ToString(); // <- This line will cause an allocation and won't cache the result.
+    var s2 = color.ToInvariantString(); // <- This call is more performant and won't cause a new allocation.
+}
+```
+
+Both calls will return the same string, but the second one will be more efficient.
+
+### Using the generated with an enum from a different assembly
```
Tratcher, line 61:
```suggestion
### Using the generator with an enum from a different assembly
```

Tratcher (approved):

Issue #4622, opened 2023-10-27
Make sure extensions libraries are marked <IsAotCompatible>true</IsAotCompatible>


davidfowl:
While trying to use the Http resilisnce library auth PublishAot I got warnings:

**EnableReloads**

```
ILC : Trim analysis warning IL2091: Microsoft.Extensions.Http.Resilience.ResilienceHandlerContext.EnableReloads<TOptions>(String): 'TOptions' generic argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in 'Polly.DependencyInjection
.AddResiliencePipelineContext`1.EnableReloads<TOptions>(String)'. The generic parameter 'TOptions' of 'Microsoft.Extensions.Http.Resilience.ResilienceHandlerContext.EnableReloads<TOptions>(String)' does not have matching annotations. The source value must declare at le
ast the same requirements as those declared on the target location it is assigned to. [C:\dev\git\eShop\src\Ordering.BackgroundTasks\Ordering.BackgroundTasks.csproj]
ILC : Trim analysis warning IL2026: Polly.Timeout.TimeoutStrategyOptions.Timeout: Using member 'System.ComponentModel.DataAnnotations.RangeAttribute.RangeAttribute(Type,String,String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming
application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All. [C:\dev\git\eShop\src\Ordering.BackgroundTasks\Ordering.BackgroundTasks.csproj]
ILC : Trim analysis warning IL2026: Polly.Timeout.TimeoutStrategyOptions.Timeout: Using member 'System.ComponentModel.DataAnnotations.RangeAttribute.RangeAttribute(Type,String,String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming
application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All. [C:\dev\git\eShop\src\Ordering.BackgroundTasks\Ordering.BackgroundTasks.csproj]
[...]

PR #4624, opened 2023-10-27
Fix trim warning in ResilienceHandlerContext

eerhardt:
Contributes to #4622

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4624)


joperezr (approved):
Thanks for fixing this one. We will work on enabling the analyzer on the repo so that we can take a look at what other warnings we get.

PR #4633, opened 2023-10-30
Fix standard hedging

martintmk:
Hedging should assign the action resilience context to cloned request message, otherwise the inner handler might use incorrect context.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4633)


joperezr:
Thanks for the PR @martintmk, we have now began our lock down and stabilization period for 8.0. Do you think this PR is something we should consider for 8.0, or is it okay to let this one be part of 8.1 instead?

martintmk:
> Thanks for the PR @martintmk, we have now began our lock down and stabilization period for 8.0. Do you think this PR is something we should consider for 8.0, or is it okay to let this one be part of 8.1 instead?

Without this PR the standard hedging is basically broken, so is a must from my perspective.

geeknoid (approved):

PR #4639, opened 2023-10-30
Removing Microsoft.Extensions.EnumStrings

joperezr:
cc: @geeknoid

As discussed offline, removing EnumStrings for now in favor of adding this functionality in .NET 9 as part of the  platform using new roslyn features.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4639)


geeknoid:
For the record, I strongly disagree with removing this functionality.

geeknoid (approved):

docs/list-of-diagnostics.md
```diff
@@ -58,17 +58,6 @@
...
 
-# EnumStrings
-
-| Diagnostic ID     | Description |
-| :---------------- | :---------- |
-| `ENUMSTRGEN000` | Invalid extension namespace |
-| `ENUMSTRGEN001` | Invalid attribute constructor used |
-| `ENUMSTRGEN002` | Invalid extension class name |
-| `ENUMSTRGEN003` | Invalid extension method name |
-| `ENUMSTRGEN004` | Invalid enum type specified |
```
RussKie, line 69: @joperezr have you deleted these from aka.ms?

src/Analyzers/Microsoft.Analyzers.Extra/DiagDescriptors.cs
```diff
@@ -64,14 +64,6 @@ internal static class DiagDescriptors
         description: Resources.ArraysDescription,
         defaultSeverity: DiagnosticSeverity.Warning);
 
-    public static DiagnosticDescriptor EnumStrings { get; } = Make(
-        id: "EA0006",
```
RussKie, line 68: ditto

The change also needs to be reflected in docs/list-of-diagnostics.md.

docs/list-of-diagnostics.md
joperezr: No. Can you please log an issue for that?

src/Analyzers/Microsoft.Analyzers.Extra/DiagDescriptors.cs
joperezr: Same here, would you mind logging an issue to clean things up if they weren't from this?

PranavSenthilnathan:
@joperezr / @geeknoid Could you share what roslyn features you wanted to use for the new implementation (interceptors?) and where that implementation would go (this repo or some other one?)?

geeknoid:
I don't know, hopefully Jose does...

joperezr:
cc: @stephentoub who I originally heard talking about interceptors. I don't have a lot of info myself, but AFAIU it would be a feature that would allow to "intercept" a call to a method dynamically and then replace it with a different call, which for the particular issue that this library was meant to solve would be very impactful.

stephentoub:
Right. From my perspective, interceptors would be the way to go. The design of EnumStrings was that it emitted a different method and then call sites needed to be changed to call that method. Interceptors avoid the need for such manual changes to code; effectively the source generator can ask the compiler to replace a specific call with a call to a different method that it can generate, such that any use of enum.ToString would be replaceable. I'm still not a huge fan of this overall; I'd rather find ways to just make enum formatting fast enough that folks aren't concerned about it (many scenarios are already pretty darn fast), but if we really believe we need a source generator for it, interceptors would be where I'd want to go.

PR #4647, opened 2023-11-01
Fix inconsistent signature of `GetOptions` method

martintmk:
Noticed this while working on samples. The new signature matches the API of Polly, has null as a default. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4647)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHandlerContext.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHandlerContext.cs
@@ -44,6 +44,7 @@ public sealed class ResilienceHandlerContext
     /// <remarks>
     /// You can decide based on the <paramref name="name"/> to listen for changes in global options or named options.
     /// If <paramref name="name"/> is <see langword="null"/> then the global options are listened to.
+    /// By default, the <paramref name="name"/> parameter is <see langword="null"/>.
     /// <para>
     /// You can listen for changes only for single options. If you call this method multiple times, the preceding calls are ignored and only the last one wins.
     /// </para>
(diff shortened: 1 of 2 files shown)


joperezr (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHandlerContext.cs
```diff
@@ -59,7 +59,7 @@ internal ResilienceHandlerContext(AddResiliencePipelineContext<HttpKey> context)
     /// <remarks>
     /// If <paramref name="name"/> is <see langword="null"/> then the global options are returned.
     /// </remarks>
-    public TOptions GetOptions<TOptions>(string name) => _context.GetOptions<TOptions>(name);
+    public TOptions GetOptions<TOptions>(string? name = null) => _context.GetOptions<TOptions>(name);
```
joperezr, line 62: Can we also add to the triple slash docs remarks section that if nothing gets passed then null will be used by default?

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHandlerContext.cs
martintmk: Isn't that implied by default null value?

joperezr: For people looking at the code, yes, but people looking at docs.microsoft.com where this info is surfaced that won't be the case. You can either add it on the remarks, or on the param section instead.

martintmk: Done.

PR #4651, opened 2023-11-02
Fix system.network.state dimension value

iliar-turdushev:
According to [OTel semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/1eb70c4d74aa904a8c81c5ed10302af8bb3c1b70/docs/system/system-metrics.md#metric-systemnetworkconnections) we should use value `close` instead of `closed`.

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4651)


danmoseley:
Coverage increased?? 😀

PR #4652, opened 2023-11-02
Review log levels

iliar-turdushev:
The PR is intended to discuss log levels of existing LoggerMessages.
Currently, the PR is not intended to be merged.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4652)

### DIFF
--- a/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Log.cs
+++ b/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Log.cs
@@ -8,6 +8,6 @@ namespace Microsoft.AspNetCore.Diagnostics;
 
 internal static partial class Log
 {
-    [LoggerMessage(0, LogLevel.Warning, "Enricher failed: {Enricher}.")]
+    [LoggerMessage(0, LogLevel.Warning, "Enricher failed: {Enricher}.")] 
     internal static partial void EnricherFailed(this ILogger logger, Exception exception, string enricher);
 }
--- a/src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
+++ b/src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
@@ -198,12 +198,12 @@ public sealed partial class HeaderParsingFeature
         }
     }
 
-    [LoggerMessage(LogLevel.Debug, "Can't parse header '{HeaderName}' due to '{Error}'.")]
+    [LoggerMessage(LogLevel.Debug, "Can't parse header '{HeaderName}' due to '{Error}'.")] 
     private partial void LogParsingError(string headerName, string error);
 
-    [LoggerMessage(LogLevel.Debug, "Using a default value for header '{HeaderName}'.")]
+    [LoggerMessage(LogLevel.Debug, "Using a default value for header '{HeaderName}'.")] 
     private partial void LogDefaultUsage(string headerName);
 
-    [LoggerMessage(LogLevel.Debug, "Header '{HeaderName}' not found.")]
+    [LoggerMessage(LogLevel.Debug, "Header '{HeaderName}' not found.")] 
     private partial void LogNotFound(string headerName);
 }
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Log.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Log.cs
@@ -8,13 +8,13 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 
 internal static partial class Log
 {
-    [LoggerMessage(0, LogLevel.Warning, "Process reporting unhealthy: {Status}. Health check entries are {Entries}")]
+    [LoggerMessage(0, LogLevel.Warning, "Process reporting unhealthy: {Status}. Health check entries are {Entries}")] 
     public static partial void Unhealthy(
         ILogger logger,
         HealthStatus status,
         StringBuilder entries);
 
-    [LoggerMessage(1, LogLevel.Debug, "Process reporting healthy: {Status}.")]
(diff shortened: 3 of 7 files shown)


src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
```diff
@@ -198,12 +198,12 @@ private void CopyTo(Box<T> to)
         }
     }
 
-    [LoggerMessage(LogLevel.Debug, "Can't parse header '{HeaderName}' due to '{Error}'.")]
+    [LoggerMessage(LogLevel.Debug, "Can't parse header '{HeaderName}' due to '{Error}'.")]
```
iliar-turdushev, line 201: While `LogParsingError` and `LogNotFound` messages log erroneous situations, making them of level >= Information would result in logging of a bunch of messages. So, I think it is good to leave them as is of Debug level, isn't it?

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/Log.cs
iliar-turdushev, line 1: Log levels here are good.

src/Libraries/Microsoft.Extensions.Diagnostics.Probes/Log.cs
iliar-turdushev, line 1: Log levels here are good:

* `SocketExceptionCaughtTcpEndpoint` (Error) could be logged once per 5 sec at max.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Log.cs
iliar-turdushev, line 1: Log levels here are good. They will be logged once per 1 sec max in case of some erroneous situation.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Log.cs
iliar-turdushev, line 1: Log levels here are good. They'll be logged only once per application's lifetime.

src/Libraries/Microsoft.Extensions.Hosting.Testing/HostTerminatorService.cs
iliar-turdushev, line 1: Log levels here are good. They are supposed to be logged only during Unit or Integration testing.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Log.cs
```diff
@@ -8,6 +8,6 @@ namespace Microsoft.AspNetCore.Diagnostics;
 
 internal static partial class Log
 {
-    [LoggerMessage(0, LogLevel.Warning, "Enricher failed: {Enricher}.")]
+    [LoggerMessage(0, LogLevel.Warning, "Enricher failed: {Enricher}.")]
```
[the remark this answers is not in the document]
iliar-turdushev: Using the following comment as a guidance https://github.com/dotnet/extensions/pull/4604#issuecomment-1790492949, I think that using Warning level is correct, because an enricher that throws is not a typical situation, and it would be good for a user to have visibility of error caused by enrichers.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
iliar-turdushev: Originally `LogParsingError` had level Warning, then it was changed to Debug.

@geeknoid Can you help here, please? Are we good to have log level for `LogParsingError` message as Debug, though it reports erroneous situation? Thank you.

src/Libraries/Microsoft.Extensions.Diagnostics.Probes/Log.cs
davidfowl: That sounds like noise. The health check will fail, do we need to also log? Does something else also log?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Log.cs
davidfowl: These should be warnings not errors

iliar-turdushev:
@noahfalk @geeknoid Could you, please, review the PR and help to agree on log levels? Thank you.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/HeaderParsingFeature.cs
geeknoid: This must be debug. Otherwise, an attacker could set malformed headers and trigger a large amount of log spew from the service, which could potentially lead to an outage.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Log.cs
iliar-turdushev: You meant not information not "not errors", right? Currently, log levels are informations.

src/Libraries/Microsoft.Extensions.Diagnostics.Probes/Log.cs
iliar-turdushev: As I understand, that this log message could be logged even in case of successful health check. The implementation involves opening or closing a TCP port when health checks succeed or fail. I believe, we need this log message to have visibility of what is happening.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Log.cs
iliar-turdushev: I think that Information logs are good here. What do you think @davidfowl?

iliar-turdushev:
We are good with the current log levels.

PR #4659, opened 2023-11-02
Resilience readme, namespace

Tratcher:
I moved the ResilienceContext extensions to match the Polly namespace.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4659)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Resilience/README.md
+++ b/src/Libraries/Microsoft.Extensions.Resilience/README.md
@@ -1,6 +1,6 @@
 # Microsoft.Extensions.Resilience
 
-Mechanisms to harden applications against transient failures.
+Extensions to the Polly libraries to enrich telemetry with metadata and exception summaries.
 
 ## Install the package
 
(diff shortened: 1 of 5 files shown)


joperezr (approved):
:shipit:

src/Libraries/Microsoft.Extensions.Resilience/README.md
```diff
@@ -18,6 +18,22 @@ Or directly in the C# project file:
...
 
+## Usage Examples
+
+The services can be registered using the following method:
+
+```csharp
+public static IServiceCollection AddResilienceEnricher(this IServiceCollection services)
+```
+
+This will optionally consume the `IExceptionSummarizer` service if it has been registered and add that data to Polly's telemetry. It will also include `RequestMetadata` that can be set or retrieved with these extensions:
```
joperezr, line 29: I wonder if the README should have a pointer to the exception summarizer package as well as the Telemetry one. That said, it probably makes more sense to point to conceptual docs describing these, so perhaps it's fine to keep as is and once we have those conceptual docs we just add a pointer to those here.

Tratcher: Yeah, concept doc links would be more helpful.

src/Libraries/Microsoft.Extensions.Resilience/README.md
```diff
@@ -18,6 +18,22 @@ Or directly in the C# project file:
 </ItemGroup>
 ```
 
+## Usage Examples
+
+The services can be registered using the following method:
+
+```csharp
+public static IServiceCollection AddResilienceEnricher(this IServiceCollection services)
```
sebastienros, line 26: I put a semi-colon on these API blocks. We should standardize when all the READMEs are in

Tratcher: How does the semi-colon help? Neither are syntactically correct. This is a method definition which would be followed by `{ ... }`. A semi-colon would make sense if this was a calling example like `serviceCollection.AddResilienceEnricher()`.

Issue #4649, opened 2023-11-01
Clarify experimental package warning


Tratcher:
https://github.com/dotnet/extensions/blob/7d69192a804087d6a4e67d91803072003666b54f/eng/MSBuild/ProjectStaging.targets#L20

This gets prepended to package descriptions. It should instead say something about being Experimental and link to what that means.

joperezr:
@RussKie did you say yesterday you were also going to look into this?

joperezr:
We should just change the text to say "Experimental package". Development Build is not a concept tied to Experimental APIs.

joperezr:
We discussed this in the tactical sync, we shouldn't be making these packages stable.

PR #4662, opened 2023-11-02
Have dev packages produce non-stable package versions.

joperezr:
Fixes https://github.com/dotnet/extensions/issues/4649

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4662)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -12,7 +12,7 @@
     <!--
         When DotNetFinalVersionKind is set to 'release', this branch will produce stable outputs for 'Shipping' packages
     -->
-    <DotNetFinalVersionKind Condition="'$(DotNetFinalVersionKind)' == ''">release</DotNetFinalVersionKind>
+    <DotNetFinalVersionKind Condition="'$(DotNetFinalVersionKind)' == '' And '$(Stage)' != 'dev'">release</DotNetFinalVersionKind>
     <!-- Enabling this rule will cause build failures on undocumented public APIs. -->
     <SkipArcadeNoWarnCS1591>true</SkipArcadeNoWarnCS1591>
   </PropertyGroup>
--- a/src/Generators/Microsoft.Gen.MetricsReports/Directory.Build.props
+++ b/src/Generators/Microsoft.Gen.MetricsReports/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project>
+  <!-- In order to get the right package versions for projects that shouldn't stabilize, we need to set this property before
+  importing the root level Directory.Build.props file. This property should be kept in here, as opposed to moving it to
+  the project itself. -->
+  <PropertyGroup>
+    <Stage>dev</Stage>
+  </PropertyGroup>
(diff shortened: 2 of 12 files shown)


joperezr:
FYI @Tratcher who originally reported this.

joperezr:
Here is a local build of these changes, where you can see how other packages are still produced with stable versions, but dev packages re not stabilized:

image

src/Generators/Microsoft.Gen.MetricsReports/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <!-- In order to get the right package versions for projects that shouldn't stabilize, we need to set this property before
```
Tratcher, line 2: This is unfortunate, it defines the stage property in an inconsistent place and makes it harder to use correctly. Can we do something else like have these projects override the DotNetFinalVersionKind property?

joperezr: Using Directory.Build.props on library-specific projects to work around things like this is not that uncommon, in fact [dotnet/runtime use this all the time](https://github.com/search?q=repo%3Adotnet%2Fruntime+path%3Adirectory.build.props+path%3A%2F%5Esrc%5C%2Flibraries%5C%2F%2F&type=code). One more issue is that the reason why this property (DotnetFinalVersionKind) needs to be defined so early is because it needs to be defined before the import of `Version.BeforeCommonProps.props, which gets imported before the project, meaning that even if we want to override that property in the .csprojs it would still be too late for some other things that are derived from it.

eng/Versions.props
```diff
@@ -12,7 +12,7 @@
     <!--
         When DotNetFinalVersionKind is set to 'release', this branch will produce stable outputs for 'Shipping' packages
     -->
-    <DotNetFinalVersionKind Condition="'$(DotNetFinalVersionKind)' == ''">release</DotNetFinalVersionKind>
+    <DotNetFinalVersionKind Condition="'$(DotNetFinalVersionKind)' == '' And '$(Stage)' != 'dev'">release</DotNetFinalVersionKind>
```
RussKie, line 15: What if we moved this line to the targets file? That is, leave the original line in this file, and then copy the new line to the targets file.
Then, we'd not have to create individual props files... 🤔

joperezr: The problem is not this file, this file needs to be here because there are some props files imported from Arcade infrastructure that depend on this being set, so if we move this to targets files, arcade props that depend on this won't work.

joperezr: It is possible that we could hook this up differently but I think that would be too risky to do this late in the release. I'm fine if we want to explore a different way of doing this for main branch, but for the release branch I want to have the least amount of moving pieces possible, and moving the way we set a property used by the whole rest of arcade is much more risk that what I think we should take.

RussKie (approved):

PR #4686, opened 2023-11-08
Fix hedging for primary callbacks

martintmk:
Found this issue when working on the resilience demo. The origin of this problem comes from the fact that we are attaching `ResilienceContext` to `HttpRequestMessage`. Based on this we need to "refresh" and assign the correct instance of resilience context for each hedged action. 

For secondary hedged actions this was fixed in #4633, however, we omitted this fix for primary actions as these also get their own resilience context instance. This change makes sure that the primary hedged calls gets the correct resilience context instance.

Also switched to Polly 8.1.0 which improves the handling of resilience contexts for hedging (creates clones for primary executions too).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4686)


geeknoid (approved):

martintmk:
Removed the Polly bump and tested that everything works fine for both standard resilience and standard hedging handler. 
This also fixes issues that I had in my demo (last part about standard hedging).

joperezr (approved):

PR #4744, opened 2023-11-20
Optimize resource utilization parser for linux

rafal-mz:
LinuxUtilizationParser class is protected by lock and used internally to get utilization data from Linux OS. Parser allocates a pool of buffers at ctor time. Usage of this class makes pool of buffer writes not needed. We can just allocate one buffer writer and pass it around. 

This change improves both memory that we used to maintain the pool and cycles to pass around buffer writer instance.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParser.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParser.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
-using Microsoft.Extensions.ObjectPool;
 using Microsoft.Shared.Diagnostics;
 using Microsoft.Shared.Pools;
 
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParser.cs
```diff
@@ -4,14 +4,15 @@
...
-using Microsoft.Extensions.ObjectPool;
 using Microsoft.Shared.Diagnostics;
 using Microsoft.Shared.Pools;
 
 namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux;
 
 /// <summary>
 /// Parses Linux files to retrieve resource utilization data.
+/// This class is not thread safe.
+/// When the same instance is called by multiple threads it may return corrupted data.
```
RussKie, line 15: Suggest moving this as `<remarks>`. Intellisense shows those, making the discoverability easier

RussKie (approved):

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParser.cs
rafal-mz: Oh, it should be remarks. Thanks for spotting. :-)

Issue #4759, opened 2023-11-27
[API Proposal]: Expose Resilience HTTP Message Handler


kmcclellan:
### Background and motivation

I'm attempting to migrate from `Microsoft.Extensions.Http.Polly` to `Microsoft.Extensions.Http.Resilience` (see https://github.com/dotnet/extensions/issues/4717).

However, in my code base there are some scenarios where I want to bypass `IHttpClientBuilder` and instantiate an HTTP delegating message handler directly. The former library exposes a public [`PolicyHttpMessageHandler`](https://github.com/dotnet/aspnetcore/blob/v8.0.0/src/HttpClientFactory/Polly/src/PolicyHttpMessageHandler.cs) for this purpose. Can we consider exposing [this library's handler](https://github.com/dotnet/extensions/blob/b31f7e964180e45b9b0855f24b91de024e0a438f/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/Internal/ResilienceHandler.cs)?

To motivate, I'll illustrate two reasons I have found for bypassing the builder.

1) To prevent duplicate handlers from being registered when `AddResilienceHandler` is called more than once. 
2) To configure multiple resilience handlers at once using the client name.

These can both be accomplished with a custom implementation of `IConfigureNamedOptions<HttpClientFactoryOptions>`:

```csharp
class ConfigureHttpResilience : IConfigureNamedOptions<HttpClientFactoryOptions>
{
        public void Configure(string? name, HttpClientFactoryOptions options)
        {
            if (name != null)
            {
                // Create a custom resilience pipeline using the client name.
                var pipeline = GetPipeline(name);

                // Can't do this since ResilienceHandler is not public!
                options.HttpMessageHandlerBuilderActions.Add(
                    builder => builder.AdditionalHandlers.Add(new ResilienceHandler(x => pipeline)));
            }
        }
}
```

```csharp

public static IServiceCollection AddHttpResilience(this IServiceCollection services)
{
    // "Try" prevents duplicates if this method is called more than once.
    services.TryAddEnumerable(
        ServiceDescriptor.Singleton<IConfigureOptions<HttpClientFactoryOptions>, ConfigureHttpResilience>());

    return services;
}
```

### API Proposal

```csharp
public class ResilienceHttpMessageHandler : DelegatingHandler
{
    public ResilienceHttpMessageHandler(ResiliencePipeline<HttpResponseMessage> pipeline)
    {
        throw new NotImplementedException();
    }

[...]

PR #4763, opened 2023-11-28
Expose HTTP resilience handler (#4759)

kmcclellan:
As discussed in #4759, we'd like to expose the resilience HTTP handler. I have left it sealed and given it two public constructors.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4763)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpMessageHandler.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpMessageHandler.cs
@@ -6,18 +6,32 @@ using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Http.Diagnostics;
+using Microsoft.Extensions.Http.Resilience.Internal;
 using Polly;
 
-namespace Microsoft.Extensions.Http.Resilience.Internal;
+namespace Microsoft.Extensions.Http.Resilience;
 
 /// <summary>
-/// Base class for resilience handler, i.e. handlers that use resilience strategies  to send the requests.
+/// An HTTP message handler using resilience strategies to send requests.
 /// </summary>
-internal sealed class ResilienceHandler : DelegatingHandler
+public sealed class ResilienceHttpMessageHandler : DelegatingHandler
 {
     private readonly Func<HttpRequestMessage, ResiliencePipeline<HttpResponseMessage>> _pipelineProvider;
 
-    public ResilienceHandler(Func<HttpRequestMessage, ResiliencePipeline<HttpResponseMessage>> pipelineProvider)
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ResilienceHttpMessageHandler"/> class using a resilience pipeline.
+    /// </summary>
+    /// <param name="pipeline">The pipeline to use for requests.</param>
+    public ResilienceHttpMessageHandler(ResiliencePipeline<HttpResponseMessage> pipeline)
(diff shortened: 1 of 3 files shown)


kmcclellan:
> Newly added symbol 'Microsoft.Extensions.Http.Resilience.ResilienceHttpMessageHandler' must be marked as experimental

Should that apply for types that have had access modifier changes?

joperezr:
@kmcclellan thanks so much for your proposal and for the contribution. I realize that we don't have this covered in our docs on this repo yet (I'll fix that), but in dotnet/extensions we follow the same API review process than the[ one used in aspnetcore repository](https://github.com/dotnet/aspnetcore/blob/main/docs/APIReviewProcess.md). This essentially means that when we want to add a new API, we should just make sure that it is approved by the API Review board before we can take a change that adds it.

It's still great that you already went and created the PR as that can be used during the review process in case there are any implementation questions. @martintmk will work on getting this API proposal reviewed and approved, so that we can take this change. I will mark this PR as No Merge for now while the API process unfolds.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpMessageHandler.cs
```diff
@@ -6,18 +6,32 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ResilienceHttpMessageHandler"/> class using a resilience pipeline delegate.
+    /// </summary>
+    /// <param name="pipelineProvider">The delegate to select a pipeline for a request.</param>
+    public ResilienceHttpMessageHandler(Func<HttpRequestMessage, ResiliencePipeline<HttpResponseMessage>> pipelineProvider)
     {
         _pipelineProvider = pipelineProvider;
```
martintmk, line 36: Since this is now internal, please validate the arguments.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpMessageHandler.cs
```diff
@@ -6,18 +6,32 @@
...
+public sealed class ResilienceHttpMessageHandler : DelegatingHandler
 {
     private readonly Func<HttpRequestMessage, ResiliencePipeline<HttpResponseMessage>> _pipelineProvider;
 
-    public ResilienceHandler(Func<HttpRequestMessage, ResiliencePipeline<HttpResponseMessage>> pipelineProvider)
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ResilienceHttpMessageHandler"/> class using a resilience pipeline.
+    /// </summary>
+    /// <param name="pipeline">The pipeline to use for requests.</param>
+    public ResilienceHttpMessageHandler(ResiliencePipeline<HttpResponseMessage> pipeline)
```
martintmk, line 25: Also validate that pipeline is not null.

martintmk (changes requested):

martintmk:
@joperezr , @kmcclellan 

The exposed API is pretty well aligned with [`PolicyHttpMessageHandler`](https://github.com/dotnet/aspnetcore/blob/5b1ad1ee4934be7d587e50670524876763cd45eb/src/HttpClientFactory/Polly/src/PolicyHttpMessageHandler.cs#L74) that uses Polly v7. This just adds even smaller subset of APIs for Polly v8 scenarios.

I believe this handler can be used for valid scenarios and bridges some feature gaps compared to the older library.

PR #4773, opened 2023-12-01
Correct build transitive targets location

### LINKED ISSUE #4772: Unable to disable Microsoft.Extensions.Logging.SourceGenerator with `DisableMicrosoftExtensionsLoggingSourceGenerator`
### Description

dotnet/extensions have own generators for `LoggerMessageAttribute]` and require the runtime's one to be disabled.
 
I'm having issues building dotnet/extensions-samples:
```
D:\Development\dotnet-r9-samples\src\Telemetry\LatencyMeasurement\LatencyContext\LatencyDataExporter\Log.cs(12,65): error SYSLIB1015: Argument 'level' is not referenced from the logging message (https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib1015.md) [D:\Development\dotnet-r9-samples\src\Telemetry\LatencyMeasurement\LatencyContext\LatencyContext.csproj::TargetFramework=net9.0]
D:\Development\dotnet-r9-samples\src\Telemetry\LatencyMeasurement\LatencyContext\Microsoft.Extensions.Logging.Generators\Microsoft.Extensions.Logging.Generators.LoggerMessageGenerator\LoggerMessage.g.cs(58,36): error CS0757: A partial method may not have multiple implementing declarations [D:\Development\dotnet-r9-samples\src\Telemetry\LatencyMeasurement\LatencyContext\LatencyContext.csproj::TargetFramework=net9.0]
```

The repo-global Directory.Build.targets sets the flag:
```xml
<Project>
  <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />

  <PropertyGroup>
    <DisableMicrosoftExtensionsLoggingSourceGenerator>true</DisableMicrosoftExtensionsLoggingSourceGenerator>
  </PropertyGroup>

</Project>
```

However, the runtime's generator is still being executed:
image

Looking at binlog it looks like `_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzers` target is being executed _before_ another target that re-adds the generator back:
image

I think that `AfterTargets` must include `ResolveReferences` target instead - testing it locally everything "just works" (c):
https://github.com/dotnet/runtime/blob/8cb3c6187f9e1338f22f9e37c4dfe615123eaa59/eng/MultiTargetRoslynComponent.targets.template#L21-L24

```diff
  <Target Name="_{TargetPrefix}RemoveAnalyzers" 
          Condition="'$({DisableSourceGeneratorPropertyName})' == 'true'"
-         AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets"
+         AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets;ResolveReferences"
          DependsOnTargets="_{TargetPrefix}GatherAnalyzers">
```

### Reproduction Steps

* https://github.com/dotnet/extensions-samples/pull/162
* https://dev.azure.com/dnceng-public/public/_build/results?buildId=484506&view=results

### Expected behavior

The generator is removed.

### Actual behavior

The generator is being re-added.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

RussKie:
Resolves #4772

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4773)


RussKie:
I tried avoiding copying the props/targets, and tried automatically build `PackagePath` based on the TFM but I couldn't figure out a way to do that... E.g., this didn't work:
```xml
  <Target Name="asdfsdfasdfsa" BeforeTargets="AssignTargetPaths">
    <ItemGroup>
      <_TargetFramework Include="$(TargetFrameworks)" />
    </ItemGroup>
    <ItemGroup>
      <None Include="buildTransitive\**\*" CopyToOutputDirectory="PreserveNewest" Pack="true" PackagePath="buildTransitive\%(_TargetFramework.Identity)\%(Filename)%(Extension)" />
    </ItemGroup>
  </Target>
```

joperezr:
While this change does fix the issue, I think that the root cause that should be fixed is the logic that is adding the `_._` placeholders to the package. This was originally intended to be used in order to be able to emit an error when users consumed the packages targeting unsupported TFMs (like netcoreapp3.1, and net5.0) but that is no longer required since the support for those TFMs was dropped from the package in the 8.0 timeframe when we stopped cross-compiling for them. At that time, we needed a place holder file on the minimum supported TFM so that the error wouldn't be printed for net6.0+.

That is what was causing this issue, since that minimum supported TFM is now net8.0, meaning in net8.0 we are emitting the placeholder, causing our net6.0 props and targets to no longer get imported. My suggestion is to disable the logic putting the placeholders all together for Core, and only keeping it for net462 which is the only place where we want it. That way, we will solve this issue, but we would also make it so that any future packages that want to add their own props and targets won't come into this pitfall. Does that make sense?

joperezr:
We should do more cleanup in general for that target as it was designed for a very different purpose than what it is doing now, but essentially I'm suggesting that perhaps just deleting

https://github.com/dotnet/extensions/blob/3b57f78e4caf4afb1652c742e4c7087a9a287685/eng/MSBuild/Packaging.targets#L67-L75

And

https://github.com/dotnet/extensions/blob/3b57f78e4caf4afb1652c742e4c7087a9a287685/eng/MSBuild/Packaging.targets#L79-L85

will also fix this issue, and it will also be a step closer to the right direction we want. It would also future proof so that we don't have to go through the same issue each year when we change the minimum supported TFM.

RussKie:
Thanks for the ideas. The whole existing infra to me looks overly complicated and, what's more important, inconsistent.

For `$(MinimumSupportedTfmForPackaging)` (i.e., .NET 8+) we provide two files that disable the generators and remove the analyzers:

```xml
// net8.0/Microsoft.Extensions.Telemetry.Abstractions.props

<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <!-- This package should replace the Microsoft.Extensions.Logging.Abstractions source generator, so we set the property to remove the source generator from the project. -->
  <PropertyGroup>
    <DisableMicrosoftExtensionsLoggingSourceGenerator>true</DisableMicrosoftExtensionsLoggingSourceGenerator>
  </PropertyGroup>
</Project>
```

```xml
// net8.0/Microsoft.Extensions.Telemetry.Abstractions.targets

<Project>
  <!-- This package should replace the Microsoft.Extensions.Logging.Abstractions source generator. -->
  <Target Name="_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzers" 
          Condition="'$(DisableMicrosoftExtensionsLoggingSourceGenerator)' == 'true'"
          AfterTargets="ResolveReferences">
    <ItemGroup>
      <_Microsoft_Extensions_Logging_AbstractionsAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.AssemblyName)' == 'Microsoft.Extensions.Logging.Generators' Or
                                                                                           '%(Analyzer.NuGetPackageId)' == 'Microsoft.Extensions.Logging.Abstractions'" />
    </ItemGroup>

    <!-- Remove Microsoft.Extensions.Logging.Abstractions Analyzer -->
    <ItemGroup>
      <Analyzer Remove="@(_Microsoft_Extensions_Logging_AbstractionsAnalyzer)" />
    </ItemGroup>
  </Target>
</Project>
```

However, for .NET Framework we only provide a single file, which doesn't remove the analyzers:

```xml
// net462/Microsoft.Extensions.Telemetry.Abstractions.targets

<Project InitialTargets="NETStandardCompatError_Microsoft_Extensions_Telemetry_Abstractions_net8_0">
  
  <!-- This package should replace the Microsoft.Extensions.Logging.Abstractions source generator, so we set the property to remove the source generator from the project. -->
  <PropertyGroup>
    <DisableMicrosoftExtensionsLoggingSourceGenerator>true</DisableMicrosoftExtensionsLoggingSourceGenerator>
  </PropertyGroup>

  <Target Name="NETStandardCompatError_Microsoft_Extensions_Telemetry_Abstractions_net8_0"
          Condition="'$(SuppressTfmSupportBuildWarnings)' == ''">
    <Warning Text="Microsoft.Extensions.Telemetry.Abstractions 9.0.0-dev doesn't support $(TargetFramework) and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later. You may also set &lt;SuppressTfmSupportBuildWarnings&gt;true&lt;/SuppressTfmSupportBuildWarnings&gt; in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk." />
  </Target>
</Project>
```

And I maintain that we must continue renaming the folder (unless, of course we can automate it with the MSBuild) whenever `$(MinimumSupportedTfmForPackaging)` changes (this would only happen once in two years). Because having `net6.0` folder in a package that targets .NET 8 and above is plain wrong.

joperezr:
>However, for .NET Framework we only provide a single file, which doesn't remove the analyzers:

It does, it's just not in the package :).  Setting that property `DisableMicrosoftExtensionsLoggingSourceGenerator` to true will remove the generator whenever an application got it via the Logging.Abstractions NuGet package, which for net462, that is always the case. For .NET Core though, you could also get the generator via the aspnetcore ref pack, which would then not have the target specified to disable it which is why for net6.0 (or net8.0 after your change) we have a target that does this.

RussKie:
> We should do more cleanup in general for that target as it was designed for a very different purpose than what it is doing now, but essentially I'm suggesting that perhaps just deleting
> 
> https://github.com/dotnet/extensions/blob/3b57f78e4caf4afb1652c742e4c7087a9a287685/eng/MSBuild/Packaging.targets#L67-L75
> 
> And
> 
> https://github.com/dotnet/extensions/blob/3b57f78e4caf4afb1652c742e4c7087a9a287685/eng/MSBuild/Packaging.targets#L79-L85
> 
> will also fix this issue, and it will also be a step closer to the right direction we want. It would also future proof so that we don't have to go through the same issue each year when we change the minimum supported TFM.

I had a look but I don't think we can remove this infra _without_ making substantial manual work it automates. It essentially generates the "not supported" warnings for all packages that we ship. If we remove this infra we'll have to create all these files manually, and that'd be a significant overhead.
image

I added some comments that would hopefully help removing the confusion I had. The newly added target we'll alert us when we bump the `$(MinimumSupportedTfmForPackaging)`, which will be once in two years event.

RussKie:
I'm merging this to unblock the build of the dotnet/extensions-samples repo

PR #4774, opened 2023-12-01
Migrate to the 'locker' GitHub action for locking closed/stale issues/PRs

jeffhandley:
This migrates away from the FabricBot implementation of a scheduled search over closed issues and adopts the [Locker](https://github.com/microsoft/vscode-github-triage-actions/?tab=readme-ov-file#locker) GitHub action authored by the VS Code team.

The updated fabricbot.json was generated via https://github.com/dotnet/fabricbot-config/pull/83. The workflow file was implemented and tested at https://github.com/dotnet/fabricbot-config/blob/main/.github/workflows/locker.yml.

Since this workflow uses the actions/checkout action, we need to ensure the following GitHub Actions settings are enabled:

1. Allow enterprise, and select non-enterprise, actions and reusable workflows
2. Allow actions created by GitHub

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4774)


.github/workflows/locker.yml
```diff
@@ -0,0 +1,36 @@
...
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions
+        uses: actions/checkout@v4
```
RussKie, line 25: We use `checkout@v3`:
https://github.com/dotnet/extensions/blob/13a6288c9de8d8f83c40ae419803e6a990bf5825/.github/workflows/azure-sync.yml#L19

Can we use the same? Or we specifically need v4?

We can probably update the others to v4, if required...

RussKie (approved):

RussKie:
Thank you

.github/workflows/locker.yml
jeffhandley: I do not know if v4 was _required_ here. It's a straightforward git clone, so I suspect it would work. Since this PR has already been merged, I'll leave it to you to update this or other actions as you see fit though, @RussKie.

PR #4776, opened 2023-12-01
Fix missing readme entry in NuSpecs

### LINKED ISSUE #4722: READMEs are not rendered on NuGet.org
Source for the metadata: https://github.com/dotnet/extensions/blob/release/8.0/src/Directory.Build.props#L12-L14

Using Package Explorer we can confirm that the `<readme>` tag is missing.
Example: https://nuget.info/packages/Microsoft.Extensions.Diagnostics.ExceptionSummarization/8.0.0

Details for npgsql:

```xml
   <icon>postgresql.png</icon>
    <readme>README.md</readme>
    <projectUrl>https://github.com/npgsql/npgsql</projectUrl>
    <description>Npgsql is the open source .NET data provider for PostgreSQL.</description>
```

and for `Microsoft.Extensions.Diagnostics.ExceptionSummarization`

```xml
    <icon>Icon.png</icon>
    <projectUrl>https://dot.net/</projectUrl>
    <description>Lets you retrieve exception summary information.</description>
```

However, the README file is available in the package at the expected location.

RussKie:
Resolves #4722

`$(IsShipping)` is being set in one of the *.targets, which requires the readme setting to be done in a *.targets as well.

While at it, move the definitions to Packaging.targets which is the better suited location.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4776)

### DIFF
--- a/eng/MSBuild/Packaging.targets
+++ b/eng/MSBuild/Packaging.targets
@@ -22,6 +22,14 @@
                                        ($(TargetFrameworks.Contains('$(MinimumSupportedTfmForPackaging)')) and $(TargetFrameworks.Contains('$(ConditionalNet462)')))" />
   </ItemGroup>
 
+  <PropertyGroup Condition=" '$(IsPackable)' == 'true' and '$(IsShipping)' == 'true' ">
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Condition=" '$(IsPackable)' == 'true' and '$(IsShipping)' == 'true' " Include="README.md" Pack="true" PackagePath="\" />
+  </ItemGroup>
+
   <!-- Verify that the minimum supported TFM is actually used. -->
   <Target Name="_VerifyMinimumSupportedTfmForPackagingIsUsed"
           Condition="'$(IsPackable)' == 'true' and '$(DisableNETStandardCompatErrors)' != 'true'">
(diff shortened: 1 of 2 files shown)


geeknoid (approved):

joperezr (commented):
While this approach is fine too, Is there any issue of instead removig the condition entirely for PackageReadmeFile, and just have the one single package that violates this (the transport docs package) to either override the property or generate an empty README? Given property and item ordering, and how these ones being moved control many other things, I think that doing it that way would reduce the risk of this introducing a different hard-to-catch issue.

joperezr:
If you still think that this is the right way to address the issue, I'm fine with it too, I would just ask that before merging we do a full diff of package contents and nuspec before and after the change, to ensure that we are getting the desired effect.

RussKie:
This is the correct way to fix the issue.

### Before ([the latest main](https://dev.azure.com/dnceng/internal/_build/results?buildId=2328137&view=artifacts&pathAsName=false&type=publishedArtifacts)):
image


### After ([private build of this PR](https://dev.azure.com/dnceng/internal/_build/results?buildId=2328369&view=artifacts&pathAsName=false&type=publishedArtifacts)):
image

RussKie:
Btw, if there's a desire to open the readmes in VS upon install, then those have to be `readme.txt` and not `readme.md` ([source](https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/#add-a-readme-to-your-package)).

joperezr:
> This is the correct way to fix the issue.

I didn't mean to infer that your change doesn't work, I do believe that it works (and thanks for validating that 😃). I was just suggesting to consider a less-risky change that would also work which was keep properties where they were, and just removing the condition as it isn't necessary (once we add a dummy readme to the docs package). But anyway, I don't think this feedback is worth blocking this PR, so I'll go ahead and approve, and let you decide if you want to go with the less-risky alternative.

joperezr (approved):

PR #123, opened 2016-06-04
Remove debugType: full from test since it prevents compilation on non…

pranavkm:
…-Windows machines.

PR #4798, opened 2023-12-07
Address issue with legacy logger's calling ToString on the TState arguments.

### LINKED ISSUE #4637: FakeLogRecord should expose a dictionary of its tags
Right now, the FakeLogRecord exposes the StructuredState property that returns the record's set of name/value pairs. This is not particularly convenient for tests that want to probe whether a record contains a particular tag with a particular value.

FakeLogRecord could expose a dictionary to make test code easier to write. Alternatively, instead of expose a dictionary, it could expose a MatchTag(string tagName, object? tagValue) method that test code can assert on.

### LINKED ISSUE #4795: Enabling logging enrichment effects the logged message
### Description

Using `EnableEnrichment` on `ILoggingBuilder` causes the Message of the log events to be some type name.

### Reproduction Steps

Example:
```c#
using Microsoft.Extensions.Diagnostics.Enrichment;
using System.Text.Json;

services.AddLogging(x => x
    .EnableEnrichment()
    .AddJsonConsole(y =>
    {
        y.JsonWriterOptions = new JsonWriterOptions()
        {
            Indented = true
        };
    }));

var serviceProvider = services.BuildServiceProvider();

var logger = serviceProvider.GetRequiredService<ILogger<object>>();

logger.LogInformation("this is a test");
```
This outputs
```json
{
  "EventId": 0,
  "LogLevel": "Information",
  "Category": "object",
  "Message": "this is a test",
  "State": {
    "Message": "Microsoft.Extensions.Logging.ExtendedLogger\u002BLegacyTagJoiner",
    "{OriginalFormat}": "this is a test"
  }
}
```
If you disable / comment out the EnableEnrichment line:
```json
{
  "EventId": 0,
  "LogLevel": "Information",
  "Category": "object",
  "Message": "this is a test",
  "State": {
    "Message": "this is a test",
    "{OriginalFormat}": "this is a test"
  }
}
```
Note the State.Message property value.

### Expected behavior

The State.Message property should be unaffected.

### Actual behavior

The State.Message property is the string "Microsoft.Extensions.Logging.ExtendedLogger\u002BLegacyTagJoiner"

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

geeknoid:
Fixes #4795 

And while I was in the neighborhood, also fixes #4637 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4798)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.LegacyTagJoiner.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.LegacyTagJoiner.cs
@@ -76,5 +76,7 @@ internal sealed partial class ExtendedLogger
         }
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public override string? ToString() => State?.ToString();
     }
 }
--- a/test/Libraries/Microsoft.Extensions.Telemetry.Tests/Logging/ExtendedLoggerTests.cs
+++ b/test/Libraries/Microsoft.Extensions.Telemetry.Tests/Logging/ExtendedLoggerTests.cs
@@ -451,8 +451,6 @@ public static class ExtendedLoggerTests
         Assert.Equal(new EventId(2, "ID2b"), snap[3].Id);
         Assert.Equal("MSG2b", snap[3].Message);
 
-        var state = snap[3].StructuredState;
-
         var exceptionType = snap[3].GetStructuredStateValue("exception.type")!;
         Assert.Equal("System.AggregateException", exceptionType);
 
@@ -474,6 +472,7 @@ public static class ExtendedLoggerTests
         }
         else
         {
+            var state = snap[3].StructuredState;
             Assert.DoesNotContain(state!, x => x.Key == "exception.message");
 
             Assert.DoesNotContain("EM1", stackTrace);
--- a/test/Libraries/Microsoft.Extensions.Telemetry.Tests/Logging/ExtendedLoggerTests.cs
+++ b/test/Libraries/Microsoft.Extensions.Telemetry.Tests/Logging/ExtendedLoggerTests.cs
@@ -74,18 +74,18 @@ public static class ExtendedLoggerTests
         Assert.Null(snap[0].Exception);
         Assert.Equal(new EventId(0), snap[0].Id);
         Assert.Equal("MSG0", snap[0].Message);
-        Assert.Equal("EV1", snap[0].StructuredState!.GetValue("EK1"));
-        Assert.Equal("SEV1", snap[0].StructuredState!.GetValue("SEK1"));
+        Assert.Equal("EV1", snap[0].GetStructuredStateValue("EK1"));
+        Assert.Equal("SEV1", snap[0].GetStructuredStateValue("SEK1"));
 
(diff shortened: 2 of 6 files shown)


RussKie:
@geeknoid could you please update the subject giving a more meaningful message? 

Having "Fix #123" in the git history isn't' very meaningful. At the same that information is great to have in the body of the commit because that makes GitHub create cross-references. 
Please see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

test/Libraries/Microsoft.Extensions.Telemetry.Tests/Logging/ExtendedLoggerTests.cs
```diff
@@ -431,18 +453,18 @@ public static void Exceptions(bool includeExceptionMessage)
 
         var state = snap[3].StructuredState;
```
iliar-turdushev, line 454: nit: This variable is used only once on line 477 and can be inlined there.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.LegacyTagJoiner.cs
```diff
@@ -76,5 +76,7 @@ public void SetIncomingTags(IReadOnlyList<KeyValuePair<string, object?>> value)
         }
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public override string? ToString() => State?.ToString();
```
iliar-turdushev, line 80: As I understand, in case of the modern logging path (State is a LoggerMessageState) we'll output the LoggerMessageState serialized to a string as the Message, right? Is it the desired behavior?

iliar-turdushev, line 80: As I understand, in case of the modern logging path (the tag joiner is a ModernTagJoiner, and the State is a LoggerMessageState) we'll output the LoggerMessageState serialized to a string as the Message, right? Is it the desired behavior?

iliar-turdushev (approved):

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.LegacyTagJoiner.cs
geeknoid: Yes, that is correct.

The LegacyTagJoiner deals with log state generated using the old LogInformation/LogDebug/LogError methods. For those cases, the TState comes in as a string which is meaningful to apply ToString on.

But in the ModernTagJoiner, the TState is a LoggerMessageState instance and there isn't a string available there.

test/Libraries/Microsoft.Extensions.Telemetry.Tests/Logging/ExtendedLoggerTests.cs
geeknoid: Done.

PR #4802, opened 2023-12-10
Add missing [GeneratedCode] attribute

geeknoid:
The logging code generator emits a static lambda. The lambda's containing function is annotated with the [GeneratedCode] attribute, but the lambda itself is not. The result is that code generators can get confused, considering the lambda as being user code.

This PR adds the [GeneratedCode] attribute to the lambda.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4802)

### DIFF
--- a/scripts/Slngen.ps1
+++ b/scripts/Slngen.ps1
@@ -70,7 +70,7 @@ param (
     [Parameter(Mandatory = $false, HelpMessage="Minimizes console output.")]
     [switch]$Quiet = $false,
     [Parameter(Mandatory = $false, HelpMessage="Output file name.")]
-    [string]$OutSln = "SDK.sln",
+    [string]$OutSln = "Extensions.sln",
     [Parameter(Mandatory = $false, HelpMessage="Parameters passed to MSBuild with slngen.")]
     [string[]]$MsBuildParams,
     [Parameter(Mandatory = $false, HelpMessage="Path to the repository.")]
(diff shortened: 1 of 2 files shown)


scripts/Slngen.ps1
```diff
@@ -70,7 +70,7 @@ param (
     [Parameter(Mandatory = $false, HelpMessage="Minimizes console output.")]
     [switch]$Quiet = $false,
     [Parameter(Mandatory = $false, HelpMessage="Output file name.")]
-    [string]$OutSln = "SDK.sln",
+    [string]$OutSln = "Extensions.sln",
```
RussKie, line 73: This change will lead to chaos
image

RussKie (changes requested):

scripts/Slngen.ps1
geeknoid: Undone.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -99,7 +99,7 @@ private void GenLogMethod(LoggingMethod lm)
 
         var lambdaStateName = PickUniqueName("s", lm.TemplateToParameterName.Select(kvp => kvp.Key));
 
-        OutLn($"static ({lambdaStateName}, {exceptionLambdaName}) =>");
+        OutLn($"[{GeneratorUtilities.GeneratedCodeAttribute}] static ({lambdaStateName}, {exceptionLambdaName}) =>");
```
jaredpar, line 102: Consider also adding the lambda return type to the generated lambda as well. Has a positive impact on compiler performance.

RussKie (approved):

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
jaredpar: It's fairly new syntax.

PR #4808, opened 2023-12-12
Add more information and sample to Http.Diagnostics Readme

joperezr:
Adding more documentation to the Microsoft.Extensions.Http.Diagnostics package. We got a few reports stating that there is some confusion on how this package is intended to work, so hopefully this extra docs and samples will make it clearer and will help folks hitting issues with this.

cc: @CarnaViire @stephentoub @geeknoid @xakep139 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4808)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Diagnostics/README.md
+++ b/src/Libraries/Microsoft.Extensions.Http.Diagnostics/README.md
@@ -1,6 +1,6 @@
 # Microsoft.Extensions.Http.Diagnostics
 
-Telemetry support for `HttpClient` that allows tracking latency and enriching and redacting log output.
+Telemetry support for `HttpClient` that allows tracking latency and enriching and redacting log output for structured logs.
 
 ## Install the package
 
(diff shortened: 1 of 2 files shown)


geeknoid (approved):

xakep139:
Thanks @joperezr for improving the docs!

src/Libraries/Microsoft.Extensions.Http.Diagnostics/README.md
```diff
@@ -55,6 +55,47 @@ builder.Services.AddHttpClientLogEnricher<MyHttpClientLogEnricher>();
 var host = builder.Build();
 ```
 
+It is important to note that the `AddExtendedHttpClientLogging` method will add information to the logs using *enrichment*. This means that the information will be added as tags to the structured logs, but will not be visible in the log message that is printed by default in the console. To view the information, you will need to use a logging provider that supports structured logs. One quick and built-in way to do this, is to call `AddJsonConsole()` to your logging builder, which will print out the full structured logs to the console. Here is a quick sample that uses the `ExtendedHttpClientLogging()` method to automatically log all `HttpClient` request and response bodies, and then prints the full structured logs to the console:
```
CarnaViire, line 58: Small nit: the example below would only print the response body, because only `ResponseBodyContentTypes` is set up, right? And the description says "request and response bodies"

joperezr: Good catch! I'll fix that in a follow-up PR.

PR #4838, opened 2023-12-24
Separate EnableRedaction and EnableEnrichment

### LINKED ISSUE #4683: Inconsistent behavior in extended logging
### Description

Current implementation of ExtendedLogger enables both enrichment and redaction regardless of what was enabled.

### Reproduction Steps

**Scenario №1:**
```cs
using var loggerFactory =
    LoggerFactory.Create(builder =>
    {
        builder.EnableRedaction();
        builder.Services.AddLogEnricher<CustomEnricher>();
    });

var logger = loggerFactory.CreateLogger("MyDemoLogger");
```

Even though I haven't enabled enrichment, the `CustomEnricher` will be called.

**Scenario №2:**
```cs
using var loggerFactory =
    LoggerFactory.Create(builder =>
    {
        builder.EnableEnrichment();
        builder.Services.AddRedaction();
    });

var logger = loggerFactory.CreateLogger("MyDemoLogger");
```

Even though I haven't enabled redaction, the `ErasingRedactor` will be called to redact data annotated with data classification.

**Scenario №3:**
One more confusing thing is that when I enable redaction feature, but don't have any redactors registered, I would get sensitive data being emitted as-is. Consider this example:
```cs
using var loggerFactory = LoggerFactory.Create(builder => builder.EnableRedaction());
var logger = loggerFactory.CreateLogger("MyDemoLogger");
Log.LogMethod(logger, "Sensitive user name");

{
    [LoggerMessage(Level = LogLevel.Warning, Message = "User {Name} has now different status")]
    public static partial void LogMethod(ILogger logger, [PrivateData] string name);
}
```

### Expected behavior

Each call to `EnableRedaction()` and `EnableEnrichment()` enables the corresponding feature only.

### Actual behavior

Any of calls `EnableRedaction()` and `EnableEnrichment()` apply both of them.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

geeknoid:
- Prior, calling EnableRedaction or EnableEnrichment would always enable both enrichment and redaction together. Now, each call only affects the specific feature it is intended.

Fixes #4683

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4838)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
@@ -63,7 +63,7 @@ internal sealed class ExtendedLoggerFactory : ILoggerFactory
         _filterOptionsChangeTokenRegistration = filterOptions.OnChange(RefreshFilters);
         RefreshFilters(filterOptions.CurrentValue);
 
-        if (enrichmentOptions == null)
+        if (enrichmentOptions is null)
         {
             // enrichmentOptions is only present if EnableEnrichment was called, so if it's null
             // then ignore all the supplied enrichers, we're not doing enrichment
--- a/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
@@ -63,6 +63,16 @@ internal sealed class ExtendedLoggerFactory : ILoggerFactory
         _filterOptionsChangeTokenRegistration = filterOptions.OnChange(RefreshFilters);
         RefreshFilters(filterOptions.CurrentValue);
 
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
```diff
@@ -63,6 +63,16 @@ internal sealed class ExtendedLoggerFactory : ILoggerFactory
         _filterOptionsChangeTokenRegistration = filterOptions.OnChange(RefreshFilters);
         RefreshFilters(filterOptions.CurrentValue);
 
+        if (enrichmentOptions == null)
```
RussKie, line 66: [nit]
```suggestion
        if (enrichmentOptions is null)
```

RussKie (approved):

xakep139 (approved):

PR #4840, opened 2023-12-25
Work on the logging generator.

### LINKED ISSUE #4576: LoggerMessage: Introduce an attribute TagName allowing to set the name of the tag being logged
## Problem

Currently `LoggerMessage` attribute doesn't support logging a message with a tag name which is not a valid C# variable name, because the name of the corresponding argument must match the name of the `LoggerMessage` tag, i.e. the following is not supported:

```csharp
[LoggerMessage("My message: {tag.with.dots.in.name}")]
public static void Log(ILogger log, string ???/* Currently we cannot provide a variable to match the tag {tag.with.dots.in.name}*/);
```

For example, we need tag names containing dots to be aligned with [OTel's semantic-conventions](https://github.com/open-telemetry/semantic-conventions/tree/b127e121b9b458e83a0fee8a32f7eb6721912951).

## Proposal

Introduce an attribute `TagName` that will guide `LoggerMessage` which tag a log method variable must be matched to:

```csharp
[LoggerMessage("My message: {http.request.host}")]
public static void Log(ILogger log, [TagName("http.request.host")] string host);
```

As a result, this log message must output the following tags:
```csharp
["{OriginalFormat}"] = "My message: {http.request.host}"
["http.request.host"] = "request host value"
```

### LINKED ISSUE #4738: Add support to transitively log complex object graphs.
### Description

Using LogProperties attribute combined with the LoggerMessage to log a complex type doesn't output all properties including those that themselves are complex types. I would expect it to traverse the entire object graph outputting key/value pairs of all properties - flatten all properties so to speak.

### Reproduction Steps

var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();
var logger = app.Services.GetRequiredService<ILogger<Program>>();
logger.ForecastRetrieved(new WeatherForecast());
app.Run();

static partial class Log
{
    [LoggerMessage(LogLevel.Information, "Test")]
    public static partial void ForecastRetrieved(this ILogger logger, [LogProperties(SkipNullProperties = false)] WeatherForecast forecast);
}

public class WeatherForecast
{
    public DateOnly Date { get; set; }
    public int TemperatureC { get; set; }
    public int TemperatureF => 32 + (int)(TemperatureC / 0.5556);
    public string? Summary { get; set; }
    public GeoArea SubArea { get; set; } = new GeoArea();
}

public class GeoArea
{
    public string Name { get; set; } = "California";
}

### Expected behavior

Generated should look like this:

// <auto-generated/>
#nullable enable
#pragma warning disable CS1591 // Compensate for https://github.com/dotnet/roslyn/issues/54103
partial class Log
{
    /// <summary>
    /// Logs "Test" at "Information" level.
    /// </summary>
    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Gen.Logging", "8.0.0.0")]
    public static partial void ForecastRetrieved(this global::Microsoft.Extensions.Logging.ILogger logger, global::WeatherForecast forecast)
    {
        if (!logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Information))
        {
            return;
        }

        var state = global::Microsoft.Extensions.Logging.LoggerMessageHelper.ThreadLocalState;

        _ = state.ReserveTagSpace(6);
        state.TagArray[5] = new("forecast.Date", forecast?.Date);
        state.TagArray[4] = new("forecast.TemperatureC", forecast?.TemperatureC);
        state.TagArray[3] = new("forecast.TemperatureF", forecast?.TemperatureF);
        state.TagArray[2] = new("forecast.Summary", forecast?.Summary);
        state.TagArray[1] = new("forecast.SubArea.Name", forecast?.SubArea?.Name);
        state.TagArray[0] = new("{OriginalFormat}", "Test");

        logger.Log(
            global::Microsoft.Extensions.Logging.LogLevel.Information,
            new(0, nameof(ForecastRetrieved)),
            state,
            null,
            static (s, _) =>
            {
                return "Test";
            });

        state.Clear();
    }
}

### Actual behavior

// <auto-generated/>
#nullable enable
#pragma warning disable CS1591 // Compensate for https://github.com/dotnet/roslyn/issues/54103
partial class Log
{
    /// <summary>
    /// Logs "Test" at "Information" level.
    /// </summary>
    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Gen.Logging", "8.0.0.0")]
    public static partial void ForecastRetrieved(this global::Microsoft.Extensions.Logging.ILogger logger, global::WeatherForecast forecast)
    {
        if (!logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Information))
        {
            return;
        }

        var state = global::Microsoft.Extensions.Logging.LoggerMessageHelper.ThreadLocalState;

        _ = state.ReserveTagSpace(6);
        state.TagArray[5] = new("forecast.Date", forecast?.Date);
        state.TagArray[4] = new("forecast.TemperatureC", forecast?.TemperatureC);
        state.TagArray[3] = new("forecast.TemperatureF", forecast?.TemperatureF);
        state.TagArray[2] = new("forecast.Summary", forecast?.Summary);
        state.TagArray[1] = new("forecast.SubArea", forecast?.SubArea);
        state.TagArray[0] = new("{OriginalFormat}", "Test");

        logger.Log(
            global::Microsoft.Extensions.Logging.LogLevel.Information,
            new(0, nameof(ForecastRetrieved)),
            state,
            null,
            static (s, _) =>
            {
                return "Test";
            });

        state.Clear();
    }
}

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

8.0

### Other information

I would suggest to make the depth of this object graph traversal configurable and also add an option to enable/disable it.

### LINKED ISSUE #4835: Suggest using [LogProperties] in [LoggerMessage] source-gen
Consider this example:

```csharp
namespace MyApp;

class MyOptions
{
    public string AppName { get; set; }
}

static partial class Log
{
    [LoggerMessage(LogLevel.Information, "Configured options: {Options}")]
    static partial void OptionsAreConfigured(ILogger logger, MyOptions options);
}
```

In that case the resulting message would be `Configured options: MyApp.MyOptions` which is not intended.
The logging generator might emit a warning instead:

`The type of a logging parameter doesn't include a custom ToString() implementation, consider using [LogProperties] or [TagProvider]`

geeknoid:
- The generator now produces a warning when asked to log an object which doesn't implement ToString(), IConvertible, or IFormattable. Fixed #4835.

- Added support for the Transitive property in the LogProperties attribute. When set to true, this causes automatic transitive traversal of a complex object, instead of requiring manual annotations of individual properties. Fixes #4738.

- Introduce the [TagName] attribute to make it possible to control the tag name used when logging a parameter or property. Fixes #4576.

- Fixed some situations where unnatural errors were produced as a result of a prior error. The dummy follow-on errors are now avoided.

- Fixed handling of cases where parameters or properties were of type of the non-generic IEnumerable. The specific type wasn't being recognized and treated as an enumerable.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4840)


docs/list-of-diagnostics.md
```diff
@@ -8,56 +8,6 @@
...
-| `EA0005` | Performance  | Consider using an array instead of a collection |
-| `EA0006` | Performance  | Replace uses of 'Enum.GetName' and 'Enum.ToString' for improved performance |
-| `EA0007` | Performance  | Use 'System.ValueTuple' instead of 'System.Tuple' for improved performance |
-| `EA0008` | Performance  | Use generic collections instead of legacy collections for improved performance |
-| `EA0009` | Performance  | Use 'System.MemoryExtensions.Split' for improved performance |
-| `EA0010` | Correctness  | Fire-and-forget async call inside a 'using' block |
-| `EA0011` | Performance  | Consider removing unnecessary conditional access operator (?) |
-| `EA0012` | Performance  | Consider removing unnecessary null coalescing assignment (??=) |
-| `EA0013` | Performance  | Consider removing unnecessary null coalescing operator (??) |
-| `EA0014` | Resilience  | The async method doesn't support cancellation |
```
RussKie, line 58: Should all of these be deleted from https://aka.ms?

docs/list-of-diagnostics.md
```diff
@@ -134,7 +82,7 @@ if desired.
 | `LOGGEN033` | Method parameter can't be used with a tag provider |
 | `LOGGEN034` | Attribute can't be used in this context |
 | `LOGGEN035` | The logging method parameter leaks sensitive data |
-
+| `LOGGEN036` | A value being logged doesn't have an effective way to be converted into a string |
```
RussKie, line 85: This needs to be added to https://aka.ms

RussKie (approved):

src/Generators/Microsoft.Gen.Logging/Parsing/TypeSymbolExtensions.cs
```diff
@@ -56,12 +57,29 @@ internal static bool ImplementsIFormattable(this ITypeSymbol sym, SymbolHolder s
...
 #pragma warning disable RS1024
         symbols.IgnorePropertiesSymbols.Contains(typeSymbol);
 #pragma warning restore RS1024
+
+    internal static bool HasCustomToString(this ITypeSymbol type)
+    {
+        ITypeSymbol? current = type;
+        while (current != null && current.SpecialType != SpecialType.System_Object)
+        {
+            if (current.GetMembers("ToString").Where(m => m.Kind == SymbolKind.Method).Cast<IMethodSymbol>().Any(m => m.Parameters.Length == 0))
```
xakep139, line 74: Should be also `public`, right?

docs/list-of-diagnostics.md
geeknoid: Yes, since we don't actually ship the analyzer package.

test/Generators/Microsoft.Gen.Logging/TestClasses/LogPropertiesSpecialTypesExtensions.cs
```diff
@@ -6,6 +6,8 @@
 using System.Numerics;
 using Microsoft.Extensions.Logging;
 
+#pragma warning disable LOGGEN036
```
xakep139, line 9: We shouldn't emit this warning when a parameter is annotated with `[LogProperties]`

src/Generators/Microsoft.Gen.Logging/Parsing/TypeSymbolExtensions.cs
geeknoid: That's enforced by the fact object.ToString is public.

test/Generators/Microsoft.Gen.Logging/TestClasses/LogPropertiesSpecialTypesExtensions.cs
xakep139: My understanding was that we emit the warning only when a parameter is presented in message template

docs/list-of-diagnostics.md
geeknoid: Can you handle this please?

test/Generators/Microsoft.Gen.Logging/TestClasses/LogPropertiesSpecialTypesExtensions.cs
geeknoid: The warning that gets produced is this:

image

which I think is correct.

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
```diff
@@ -387,4 +393,34 @@ partial class LoggerClass
 
         await RunGenerator(Source, DiagDescriptors.LogPropertiesHiddenPropertyDetected);
     }
+
+    [Fact]
+    public async Task DefaultToString()
```
xakep139, line 398: We should also check that we don't emit the warning for records and record structs

docs/list-of-diagnostics.md
xakep139: +1

test/Generators/Microsoft.Gen.Logging/TestClasses/LogPropertiesSpecialTypesExtensions.cs
xakep139: I see. Can we maybe move `#pragma` closer to the properties which cause this warning?

src/Generators/Microsoft.Gen.Logging/Parsing/TypeSymbolExtensions.cs
xakep139: My point was that you can define something like this:
```cs
	class MyClass
	{
		private string ToString() => "Hello!";
	}
```

And in that case your code will return `true` even though `MyClass.ToString()` won't be called.

docs/list-of-diagnostics.md
xakep139: Done: https://aka.ms/dotnet-extensions-warnings/LOGGEN036

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
geeknoid: Done.

test/Generators/Microsoft.Gen.Logging/TestClasses/LogPropertiesSpecialTypesExtensions.cs
geeknoid: Done.

src/Generators/Microsoft.Gen.Logging/Parsing/TypeSymbolExtensions.cs
geeknoid: Done.

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
```diff
@@ -333,4 +333,10 @@
   <data name="RecordTypeSensitiveArgumentIsInTemplateTitle" xml:space="preserve">
     <value>The logging method parameter leaks sensitive data</value>
   </data>
+  <data name="DefaultToStringMessage" xml:space="preserve">
+    <value>The type "{0}" doesn't implement ToString, IConvertible, or IFormattable (did you forget to apply [LogProperties] or [TagProvider] to "{1}"?)</value>
```
xakep139, line 337:
```suggestion
    <value>The type "{0}" doesn't implement ToString(), IConvertible, or IFormattable (did you forget to apply [LogProperties] or [TagProvider] to "{1}"?)</value>
```

src/Generators/Microsoft.Gen.Logging/Parsing/AttributeProcessors.cs
```diff
@@ -180,4 +189,7 @@ public static (bool omitReferenceName, ITypeSymbol providerType, string provider
 
         return (omitReferenceName, providerType!, providerMethodName!);
     }
+
+    public static string ExtractTagNameAttributeValues(AttributeData attr)
+        => attr.ConstructorArguments[0].Value as string ?? string.Empty;
```
xakep139, line 194: What if it doesn't have ctor's arguments? E.g., when syntax is invalid

src/Generators/Microsoft.Gen.Logging/Model/LoggingMethod.cs
```diff
@@ -28,8 +29,44 @@ internal sealed class LoggingMethod
...
+            if (templateName.Equals(p.ParameterName, StringComparison.OrdinalIgnoreCase))
+            {
+                return p;
+            }
+        }
+
+        return null;
+    }
+
+    public List<string> GetTemplatesForParameter(LoggingMethodParameter lp)
```
xakep139, line 45: This overload can use the overload defined below:
```cs
public List<string> GetTemplatesForParameter(LoggingMethodParameter lp)
    => GetTemplatesForParameter(lp.ParameterName);
```

src/Generators/Microsoft.Gen.Logging/Parsing/Parser.cs
```diff
@@ -378,12 +398,26 @@ public IReadOnlyList<LoggingType> GetLogTypes(IEnumerable<TypeDeclarationSyntax>
                 keepMethod = false;
             }
 
-            if (templatesWithAtSymbol.Count > 0)
+            TemplateExtractor.ExtractTemplates(message, lm.Templates);
+
+#pragma warning disable EA0003 // Use the character-based overloads of 'String.StartsWith' or 'String.EndsWith'
```
xakep139, line 403: Why can't we use it?

xakep139 (approved):
Left a few non-blocking comments, but the PR looks good to me

src/Generators/Microsoft.Gen.Logging/Parsing/Parser.cs
geeknoid: The overload doesn't exist. It's an analyzer bug, it shouldn't be recommending it.

src/Generators/Microsoft.Gen.Logging/Parsing/AttributeProcessors.cs
geeknoid: We never get here in that case. Roslyn simply won't list attribute which it couldn't construct.

src/Generators/Microsoft.Gen.Logging/Parsing/Resources.resx
geeknoid: Done.

src/Generators/Microsoft.Gen.Logging/Model/LoggingMethod.cs
geeknoid: Done.

PR #4842, opened 2023-12-27
Support content headers in HttpClient logging

### LINKED ISSUE #4769: AddExtendedHttpClientLogging() doesn't log Content headers
### Description

The new `HttpClient` logging facility in Microsoft.Extensions.Http.Diagnostics does not log content headers. `HttpHeadersReader` only inspects the request and response `Headers` property, not `Content.Headers`. Unfortunately, many of the interesting headers to log are only available via `Content.Headers`.

https://github.com/dotnet/extensions/blob/13a6288c9de8d8f83c40ae419803e6a990bf5825/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpHeadersReader.cs#L39
https://github.com/dotnet/extensions/blob/13a6288c9de8d8f83c40ae419803e6a990bf5825/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpHeadersReader.cs#L49

### Reproduction Steps

```csharp
using Microsoft.Extensions.Compliance.Classification;
using Microsoft.Extensions.Compliance.Redaction;

using var host = new HostBuilder()
    .ConfigureLogging(builder =>
    {
        builder.AddJsonConsole(options =>
        {
            options.JsonWriterOptions = new() { Indented = true };
            options.IncludeScopes = true;
        });
    })
    {
        services
            .AddRedaction(builder =>
            {
                // docs suggest this should be the default, but AFAICT it's not
                builder.SetRedactor<NullRedactor>(DataClassificationSet.FromDataClassification(DataClassification.None));
            });

        services.AddHttpClient("test")
            .AddExtendedHttpClientLogging(options =>
            {
                options.ResponseHeadersDataClasses.Add("Content-Type", DataClassification.None);
            });
    })
    .Build();

var client = host.Services.GetRequiredService<IHttpClientFactory>().CreateClient("test");

using var response = await client.GetAsync("https://ifconfig.co/json");

System.Console.WriteLine($"Content-Type: {response.Content.Headers.ContentType}");
```

### Expected behavior

The log message includes the property `http.response.headers.content-type` = `application/json`.

### Actual behavior

```json
{
  "EventId": 1,
  "LogLevel": "Information",
  "Category": "Microsoft.Extensions.Http.Logging.HttpClientLogger",
  "Message": "GET ifconfig.co/REDACTED",
  "State": {
    "Message": "http.request.method=GET,server.address=ifconfig.co,url.path=REDACTED,Duration=297,http.response.status_code=200",
    "http.request.method": "GET",
    "server.address": "ifconfig.co",
    "url.path": "REDACTED",
    "Duration": 297,
    "http.response.status_code": 200
  },
  "Scopes": []
}
```

### Regression?

_No response_

### Known Workarounds

I do not think a work-around is possible because the affected service, `IHttpHeadersReader`, is `internal`, so consumers cannot replace it.

### Configuration

- dotnet 8.0.100
- Microsoft.Extensions.Http.Diagnostics 8.0.0

### Other information

_No response_

xakep139:
Fixes #4769

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4842)


bench/Libraries/Microsoft.Extensions.Http.Diagnostics.PerformanceTests/Benchmarks/HeadersReaderBenchmark.cs
```diff
@@ -0,0 +1,107 @@
...
+    [Benchmark]
+    public void HeadersReaderNew()
+    {
+        _headersReader.ReadRequestHeadersNew(Request!, _outputBuffer);
+    }
+}
+
+/*
+
+BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22631.2861), VM=Hyper-V
```
xakep139, line 71: @geeknoid can you please take a look at these results and share your thoughts? The updated logic (`HeadersReaderNew`) works much faster when there are no headers or when headers count is less than the number of headers to log (specified via options). But in other cases it works a bit slower because it adds an overhead of getting headers count (~ 20ns).

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpHeadersReader.cs
```diff
@@ -36,7 +43,25 @@ public void ReadRequestHeaders(HttpRequestMessage request, List<KeyValuePair<str
             return;
         }
 
-        ReadHeaders(request.Headers, _requestHeaders, destination);
+        ReadHeaders(request.Headers, _requestHeadersToLog, destination);
+        if (request.Content is not null)
```
geeknoid, line 47: Since this is a change in behavior that could potentially result in a lot of extra telemetry and perhaps a privacy incident, this feature probably needs to be controlled by an option.

xakep139: Good point, done

bench/Libraries/Microsoft.Extensions.Http.Diagnostics.PerformanceTests/Benchmarks/HeadersReaderBenchmark.cs
geeknoid: Hmmm. I left a comment here earlier, but where did it go?

Anyway, I think this is reasonable. Let's just mention the extra perf impact in the docs.

xakep139: Done

geeknoid (approved):

xakep139:
@RussKie can you please check what's wrong with the code coverage build step?

eng/pipelines/templates/VerifyCoverageReport.yml
```diff
@@ -39,3 +39,4 @@ steps:
         repositoryName: '$(Build.Repository.Name)'
         id: $(System.PullRequest.PullRequestNumber)
       displayName: Report coverage to GitHub
+      continueOnError: 'true' # Temporary setting to avoid failing the build if the GitHub comment fails
```
xakep139, line 42: @RussKie I added this suppression to unblock the build. Please so the proper fix when you're back from vacation.

PR #4843, opened 2023-12-28
Ignore statics in metric source-gen

### LINKED ISSUE #4787: Metric generator generates tags from constant and static fields
### Description

Metrics generator generates tags (dimensions) from constant and static variables. This is not ideal behaviour as constants can be useful for something like this `[Histogram(typeof(TestMetric), Name = TestMetric.Name)]` where the Name is inside the `TestMetric` struct.

### Reproduction Steps

```
internal static partial class Metrics
{
    [Histogram(typeof(TestMetric), Name = "MyCustomMetricName")]
    public static partial SomeMetric CreateMetric(Meter meter);
}

public struct TestMetric()
{
    public const string TestConstant = "a";
    public static string TestStatic = "b";
    public string TestNormal = "c";
}
```

### Expected behavior

```
var tagList = new global::System.Diagnostics.TagList
{
    new global::System.Collections.Generic.KeyValuePair<string, object?>("TestNormal", o.TestNormal!),
};
```

### Actual behavior

```
var tagList = new global::System.Diagnostics.TagList
{
    new global::System.Collections.Generic.KeyValuePair<string, object?>("TestConstant", o.TestConstant!), // cannot be accessed by instance
    new global::System.Collections.Generic.KeyValuePair<string, object?>("TestStatic", o.TestStatic!), // cannot be accessed by instance
    new global::System.Collections.Generic.KeyValuePair<string, object?>("TestNormal", o.TestNormal!),
};
```

### Regression?

_No response_

### Known Workarounds

* Do not use statics or constants inside metric data types.
* Create inner static class `Constants` and put constants there.

### Configuration

_No response_

### Other information

_No response_

xakep139:
Fixes #4787

This PR also enables and aligns all metering tests for generated code.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4843)

### DIFF
--- a/src/Generators/Microsoft.Gen.Metrics/Parser.cs
+++ b/src/Generators/Microsoft.Gen.Metrics/Parser.cs
@@ -215,8 +215,8 @@ internal sealed class Parser
             : symbol.TypeArguments[0];
 
     private static (InstrumentKind instrumentKind, ITypeSymbol? genericType) GetInstrumentType(
-    INamedTypeSymbol? methodAttributeSymbol,
-    SymbolHolder symbols)
+        INamedTypeSymbol? methodAttributeSymbol,
+        SymbolHolder symbols)
     {
         if (methodAttributeSymbol == null)
         {
@@ -492,13 +492,11 @@ internal sealed class Parser
             Diag(DiagDescriptors.ErrorMethodHasBody, methodSymbol.GetLocation());
             keepMethod = false;
         }
-#pragma warning disable S2583 // Conditionally executed code should be reachable
         else if (!isPartial)
         {
             Diag(DiagDescriptors.ErrorNotPartialMethod, methodSymbol.GetLocation());
             keepMethod = false;
         }
-#pragma warning restore S2583 // Conditionally executed code should be reachable
 
         // ensure Metric name is not empty and starts with a Capital letter.
         // ensure there are no duplicate ids.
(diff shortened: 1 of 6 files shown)


test/Generators/Microsoft.Gen.Metrics/Unit/ParserTests.Diagnostics.cs
```diff
@@ -96,27 +96,6 @@ class Transitive
         Assert.Contains("Test.Transitive ⇆ Test.Interim", diag.GetMessage());
     }
 
-    [Fact]
-    public async Task StrongTypeCounter_CyclicReference_Guid()
```
xakep139, line 100: This test is not relevant anymore - we don't consider `Guid.Empty` since it's `static`

RussKie (approved):

src/Generators/Microsoft.Gen.Metrics/Parser.cs
```diff
@@ -688,23 +686,24 @@ private void Diag(DiagnosticDescriptor desc, Location? location, params object?[
         SpecialType specialType;
         ITypeSymbol typeSymbol;
 
-        if (symbol.IsImplicitlyDeclared)
+        if (member.IsImplicitlyDeclared ||
+            member.IsStatic)
```
xakep139, line 690: This is the main change

PR #4845, opened 2023-12-29
Log outgoing HTTP path parameters in Structured mode

### LINKED ISSUE #4698: HttpClient logging doesn't log path parameters in Structured mode
Consider this example:
```cs
var builder = Host.CreateApplicationBuilder(args);

builder.Logging.AddJsonConsole();

builder.Services.AddRedaction();

builder.Services.AddExtendedHttpClientLogging(options =>
{
    options.RequestPathLoggingMode = OutgoingPathLoggingMode.Structured;
    options.RequestPathParameterRedactionMode = HttpRouteParameterRedactionMode.Loose;
});

var host = builder.Build();

var httpClientFactory = host.Services.GetRequiredService<IHttpClientFactory>();
using var httpClient = httpClientFactory.CreateClient();

using var requestMessage = new HttpRequestMessage(HttpMethod.Get, "https://api.github.com/users/aspnet/repos");
var requestMetadata = new RequestMetadata()
{
    RequestRoute = "users/{userId}/repos",
    DependencyName = "GitHub",
    RequestName = "Get user repos"
};

requestMessage.SetRequestMetadata(requestMetadata);
using var response = await httpClient.SendAsync(requestMessage);
```

That way you don't have a chance to see the value of `userId` path parameter in logs. The same `RequestPathLoggingMode` applied in incoming HTTP logging emits path parameters as separate tags.
Is it a bug?

xakep139:
Fixes #4698

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4845)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.Http.Logging.Internal;
 /// Logs <see cref="HttpRequestMessage"/>, <see cref="HttpResponseMessage"/> and the exceptions due to errors of request/response.
 /// </summary>
 [SuppressMessage("Major Code Smell", "S109:Magic numbers should not be used", Justification = "Event ID's.")]
-internal static class Log
+internal static partial class Log
 {
     internal const string OriginalFormat = "{OriginalFormat}";
     private const string NullString = "(null)";
@@ -72,6 +72,14 @@ internal static class Log
 
         state.Clear();
     }
+
+    [LoggerMessage(LogLevel.Error, RequestReadErrorMessage)]
+    public static partial void RequestReadErrorSG(
+        ILogger logger,
+        Exception exception,
+        [TagName(HttpClientLoggingTagNames.Method)] HttpMethod method,
+        [TagName(HttpClientLoggingTagNames.Host)] string? host,
+        [TagName(HttpClientLoggingTagNames.Path)] string? path);
 
     public static void ResponseReadError(ILogger logger, Exception exception, HttpMethod method, string host, string path)
     {
--- a/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.Http.Logging.Internal;
 /// Logs <see cref="HttpRequestMessage"/>, <see cref="HttpResponseMessage"/> and the exceptions due to errors of request/response.
 /// </summary>
 [SuppressMessage("Major Code Smell", "S109:Magic numbers should not be used", Justification = "Event ID's.")]
-internal static partial class Log
+internal static class Log
 {
     internal const string OriginalFormat = "{OriginalFormat}";
     private const string NullString = "(null)";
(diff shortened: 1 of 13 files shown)


src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
```diff
@@ -171,7 +171,10 @@ public static void EnrichmentError(ILogger logger, Exception exception, string?
         var requestHeadersCount = record.RequestHeaders?.Count ?? 0;
         var responseHeadersCount = record.ResponseHeaders?.Count ?? 0;
 
-        var index = loggerMessageState.ReserveTagSpace(MinimalPropertyCount + statusCodePropertyCount + requestHeadersCount + responseHeadersCount);
+        var spaceToReserve = MinimalPropertyCount + statusCodePropertyCount + requestHeadersCount + responseHeadersCount +
+            record.PathParametersCount + (record.RequestBody is null ? 0 : 1) + (record.ResponseBody is null ? 0 : 1);
```
geeknoid, line 175: Now that I've added support for the [TagName] property, can we get rid of this custom code and use the code generator instead?

geeknoid (approved):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
xakep139: I have doubts regarding prefixes in header names, but I will try to convert this logic

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
```diff
@@ -26,8 +27,7 @@ public static void AddRequestHeaders(this LoggerMessageState state, List<KeyValu
         {
             var key = _requestPrefixedNamesCache.GetOrAdd(
                 items[i].Key,
-                static (x, p) => p + Normalize(x),
-                HttpClientLoggingTagNames.RequestHeaderPrefix);
+                static x => HttpClientLoggingTagNames.RequestHeaderPrefix + Normalize(x));
```
iliar-turdushev, line 30: Will this transformation lead to a delegate allocation on each invocation of GetOrAdd?

iliar-turdushev (approved):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
xakep139: No, the compiler is smart enough to create a `static` backing field and assign the delegate instance to it just once:
```cs
string key = LoggerMessageStateExtensions._requestPrefixedNamesCache.GetOrAdd(items[i].Key, LoggerMessageStateExtensions.<>c.<>9__2_0 ?? (LoggerMessageStateExtensions.<>c.<>9__2_0 = new Func<string, string>((object) LoggerMessageStateExtensions.<>c.<>9, __methodptr(<AddRequestHeaders>b__2_0))));

//...
    [CompilerGenerated]
    [Serializable]
    private sealed class <>c
    {
      [Nullable(0)]
      public static readonly LoggerMessageStateExtensions.<>c <>9;
      [Nullable(new byte[] {0, 1, 1})]
      public static Func<string, string> <>9__2_0;
      // ...
    }
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Microsoft.Extensions.Http.Diagnostics.csproj
```diff
@@ -42,6 +42,7 @@
     <PackageReference Include="Microsoft.Extensions.Http" />
     <PackageReference Include="System.Text.Json" />
     <PackageReference Include="System.Collections.Immutable" Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))" />
+    <PackageReference Include="System.Net.Http" Condition="'$(TargetFramework)' == 'net462'" />
```
xakep139, line 45: This is to update the implicit reference of 4.3.0 which is vulnerable

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
xakep139: Done in #4842

PR #4852, opened 2024-01-03
Refactor AsyncState class to use ConcurrentDictionary

mobratil:
The Features class, which was previously implemented with a List, has been replaced with a ConcurrentDictionary.
This should make it finally thread safe and it is in line with async local rules (https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#asynclocalt).

This PR is a follow-up to unsuccessful fix in https://github.com/dotnet/extensions/pull/4833 

From the discussion the issue with the current implementation is:
"Microsoft.Extensions.AsyncState.Features.Set(int index, object? value) now synchronizes when it expands the list, but there is still a race condition if two threads are setting values at different indexes in parallel and only one of them needs to expand the list. In this case, thread 1 can do _items[index] = value; while thread 2 is doing _items.EnsureCapacity(count);. EnsureCapacity creates a new larger array, copies the elements to it, and finally assigns the List<T>._items property. If EnsureCapacity in thread 2 has already copied the element to which thread 1 is assigning, then this assignment goes only to the old array and not to the new array, so it is lost."

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4852)


src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
```diff
@@ -59,7 +60,7 @@ public bool TryGet(AsyncStateToken token, out object? value)
             return false;
         }
 
-        value = _asyncContextCurrent.Value.Features.Get(token.Index);
+        _ = _asyncContextCurrent.Value.Features.TryGetValue(token, out value);
```
xakep139, line 63: I guess we can return the result of `TryGetValue()` call

src/Libraries/Microsoft.Extensions.AsyncState/FeaturesPooledPolicy.cs
```diff
@@ -1,20 +1,21 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Concurrent;
 using Microsoft.Extensions.ObjectPool;
 
 namespace Microsoft.Extensions.AsyncState;
 
-internal sealed class FeaturesPooledPolicy : IPooledObjectPolicy<Features>
+internal sealed class FeaturesPooledPolicy : IPooledObjectPolicy<ConcurrentDictionary<AsyncStateToken, object?>>
```
xakep139, line 9: nit: the name of the class might be updated since we don't have `Features` type anymore

andrey-noskov (commented):
AsyncState was meant to be used as a drop-in replacement for the built-in FeatureCollection and so similar to the latter wasn't designed to provide thread safety.

Using it in a concurrent flow doesn't look right to me because it might lead to non-deterministic behaviour and code races as precense of specific keys in the store might start to depend on context switches, etc.

src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
mobratil: We cannot because false is returned only in if the context is not initialized.

src/Libraries/Microsoft.Extensions.AsyncState/FeaturesPooledPolicy.cs
mobratil: Done

src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
```diff
@@ -52,14 +53,14 @@ public bool TryGet(AsyncStateToken token, out object? value)
...
 #pragma warning disable EA0011
-        if (_asyncContextCurrent.Value?.Features == null)
+        if (_asyncContextCurrent.Value?.Dictionary == null)
 #pragma warning restore EA0011
         {
             value = null;
             return false;
         }
 
-        value = _asyncContextCurrent.Value.Features.Get(token.Index);
```
xakep139, line 62: BTW, do we need `Index` in `AsyncStateToken` now?

geeknoid (approved):

geeknoid:
@mobratil Based on @andrey-noskov 's insights, perhaps we revert the prior change and just document that this is for single-threaded use.

mobratil:
Ok, I'm closing the PR in favor of updating the documentation: https://github.com/dotnet/extensions/pull/4881

PR #4853, opened 2024-01-04
Simplify platform check.

### LINKED ISSUE #4642: Replace outdates platform checks with modern ones
https://github.com/dotnet/extensions/blob/d1d66fc97f0e0eb387983a4fc89914f490026b15/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs#L65

This check (and any other like this) should be using the [OperatingSystem](https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem?view=net-7.0) APIs.

geeknoid:
Fixes #4642

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4853)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
@@ -51,6 +51,8 @@ public static class ResourceMonitoringServiceCollectionExtensions
         return services.AddResourceMonitoringInternal(configure);
     }
 
+    // can't easily test the exception throwing case
+    [ExcludeFromCodeCoverage]
     private static IServiceCollection AddResourceMonitoringInternal(
         this IServiceCollection services,
         Action<IResourceMonitorBuilder> configure)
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
```diff
@@ -62,7 +62,7 @@ public static class ResourceMonitoringServiceCollectionExtensions
 #if NETFRAMEWORK
         _ = builder.AddWindowsProvider();
 #else
-        if (GetPlatform() == PlatformID.Win32NT)
+        if (OperatingSystem.IsWindows())
```
xakep139, line 65: In that case we need to check whether `else` clause falls under `Linux`  - and throw an exception otherwise

geeknoid: Added.

RussKie (approved):

PR #4871, opened 2024-01-09
Enable AOT compatibility for all libraries

eerhardt:
Port #4625 to `main` now that the Configuration.Binder bugs are fixed in 8.0.1.

Enable AOT compatibility for all libraries and fix AOT warnings.

- Enable configuration binder source generator
- The only library that can't use the ConfigBinder SG is the HeaderParsing library.
  - Blocked by https://github.com/dotnet/runtime/issues/94547

Explicitly reference `Microsoft.Extensions.Configuration.Binder` in projects that use the source generator to ensure we get the latest version (8.0.1).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4871)

### DIFF
--- a/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
+++ b/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
@@ -23,7 +23,7 @@
 
   <PropertyGroup>
     <Stage>normal</Stage>
-    <MinCodeCoverage>99</MinCodeCoverage>
+    <MinCodeCoverage>100</MinCodeCoverage>
     <MinMutationScore>85</MinMutationScore>
   </PropertyGroup>
 
--- a/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
+++ b/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
@@ -39,6 +39,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
     <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" />
     <PackageReference Include="System.Collections.Immutable" Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))" />
--- a/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
+++ b/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
@@ -8,6 +8,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
+    <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
     <UseLoggingGenerator>true</UseLoggingGenerator>
     <InjectSharedThrow>false</InjectSharedThrow>
     <InjectTrimAttributesOnLegacy>false</InjectTrimAttributesOnLegacy>
@@ -22,7 +23,7 @@
 
   <PropertyGroup>
     <Stage>normal</Stage>
-    <MinCodeCoverage>100</MinCodeCoverage>
+    <MinCodeCoverage>99</MinCodeCoverage>
     <MinMutationScore>85</MinMutationScore>
   </PropertyGroup>
 
--- a/src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/AutoActivationExtensions.Keyed.cs
+++ b/src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/AutoActivationExtensions.Keyed.cs
@@ -55,13 +55,9 @@ public static partial class AutoActivationExtensions
     /// <param name="serviceType">The type of the service to activate.</param>
     /// <param name="serviceKey">An object used to uniquely identify the specific service.</param>
     /// <returns>The value of <paramref name="services" />.</returns>
-    [UnconditionalSuppressMessage(
-        "Trimming",
-        "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code",
-        Justification = "Addressed with [DynamicallyAccessedMembers]")]
     public static IServiceCollection ActivateKeyedSingleton(
         this IServiceCollection services,
-        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type serviceType,
+        Type serviceType,
         object? serviceKey)
     {
         _ = Throw.IfNull(services);
@@ -72,7 +68,7 @@ public static partial class AutoActivationExtensions
             .AddOptions<AutoActivatorOptions>()
             .Configure(ao =>
             {
-                var constructed = typeof(IEnumerable<>).MakeGenericType(serviceType);
+                var constructed = GetEnumerableServiceType(serviceType);
                 if (ao.KeyedAutoActivators.Contains((constructed, serviceKey)))
                 {
                     return;
@@ -85,6 +81,10 @@ public static partial class AutoActivationExtensions
                 }
 
                 _ = ao.KeyedAutoActivators.Add((serviceType, serviceKey));
+
+                [UnconditionalSuppressMessage("AotAnalysis", "IL3050:RequiresDynamicCode",
+                    Justification = "When IsDynamicCodeSupported is not supported, DependencyInjection ensures IEnumerable service types are not a ValueType.")]
+                static Type GetEnumerableServiceType(Type serviceType) => typeof(IEnumerable<>).MakeGenericType(serviceType);
             });
 
         return services;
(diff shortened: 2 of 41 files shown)


src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/AutoActivationExtensions.Keyed.cs
```diff
@@ -85,6 +81,10 @@ public static partial class AutoActivationExtensions
                 }
 
                 _ = ao.KeyedAutoActivators.Add((serviceType, serviceKey));
+
+                [UnconditionalSuppressMessage("AotAnalysis", "IL3050:RequiresDynamicCode",
+                    Justification = "When IsDynamicCodeSupported is not supported, DependencyInjection ensures IEnumerable service types are not a ValueType.")]
```
joperezr, line 86: Just curious, Is this false positive something that can be fixed in the analyzer at some point?

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
```diff
@@ -8,6 +8,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
+    <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
```
joperezr, line 11: NIT: Does it hurt to set this property to true when a project is not referencing the configuration binder package? If not, then we might want to consider defaulting this to true in directory.build.props to avoid having to add it in most places. And to that same line, should we consider adding the generator by default to all projects too? I'm assuming libraries not using configuration will simply no-op for this generator (which may include a slight perf regresion in build time) but this may be beneficial for new libraries in the future.

joperezr (approved):
Left minor comments, but looks good overall. Thanks Eric!

joperezr:
> 🎉 **Good job! The coverage increased** 🎉 Update `MinCodeCoverage` in the project files.
> 
> Project	Expected	Actual
> Microsoft.Extensions.Http.Diagnostics	91	**94**
> Full code coverage report: https://dev.azure.com/dnceng-public/public/_build/results?buildId=520582&view=codecoverage-tab

This is being addressed via https://github.com/dotnet/extensions/pull/4870

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
```diff
@@ -22,7 +23,7 @@
 
   <PropertyGroup>
     <Stage>normal</Stage>
-    <MinCodeCoverage>100</MinCodeCoverage>
+    <MinCodeCoverage>99</MinCodeCoverage>
```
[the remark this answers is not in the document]
geeknoid: If coverage fell, it might indicate that the code generator is forgetting to annotate something with [GeneratedCode], which was the case with the logging generator.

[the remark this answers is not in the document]
geeknoid: If this is the case, we shoud:

* Fix the code generator.
* Add an exclusion to eng/Common.runsettings to exclude the generated files.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/StandardHedgingHandlerBuilderExtensions.cs
```diff
@@ -25,13 +25,16 @@ public static class StandardHedgingHandlerBuilderExtensions
...
-    [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(HttpStandardHedgingResilienceOptions))]
     public static IStandardHedgingHandlerBuilder Configure(this IStandardHedgingHandlerBuilder builder, IConfigurationSection section)
     {
         _ = Throw.IfNull(builder);
         _ = Throw.IfNull(section);
 
-        var options = Throw.IfNull(section.Get<HttpStandardHedgingResilienceOptions>());
+        if (!section.GetChildren().Any())
+        {
+            Throw.ArgumentNullException(nameof(section));
```
geeknoid, line 35:
```suggestion
            Throw.ArgumentException(nameof(section));
```

"section" is not in fact null...

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/HttpStandardResiliencePipelineBuilderExtensions.cs
```diff
@@ -22,12 +22,15 @@ public static class HttpStandardResiliencePipelineBuilderExtensions
...
-    [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(HttpStandardResilienceOptions))]
     public static IHttpStandardResiliencePipelineBuilder Configure(this IHttpStandardResiliencePipelineBuilder builder, IConfigurationSection section)
     {
         _ = Throw.IfNull(builder);
         _ = Throw.IfNull(section);
-        var options = Throw.IfNull(section.Get<HttpStandardResilienceOptions>());
+
+        if (!section.GetChildren().Any())
+        {
+            Throw.ArgumentNullException(nameof(section));
```
geeknoid, line 32:
```suggestion
            Throw.ArgumentException(nameof(section));
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/StandardHedgingHandlerBuilderExtensions.cs
eerhardt: Are you OK with the (technically) breaking change here? The existing code throws an `ArgumentNullException` in this case, which is why I choose to preserve the behavior.

eerhardt: This is the test that verifies this:

https://github.com/dotnet/extensions/blob/7db3a7edc35ae8008c75631da35c01467912cacf/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Hedging/StandardHedgingTests.cs#L150-L151

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
eerhardt: (Totally my opinion, I'm curious to hear others')

Since this source generator is off by default, I thought it would be better to continue to keep it off by default. It isn't really "most places". It is 12 libraries. By comparison, `UseLoggingGenerator` is used in 13 libraries, and it isn't defaulted to be "on by default".

src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/AutoActivationExtensions.Keyed.cs
eerhardt: Unfortunately, no. The static analysis isn't able to know that this `IEnumerable<T>` is only going to be used in DependencyInjection, and it is only ever going to be ref types. So this is a case where we have to suppress the warning since we have more context than the tools can know.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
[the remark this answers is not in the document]
eerhardt: The config binder source generator does add `[GeneratedCode]` attributes. I'm reverting this change to see what the problem was - this change was ported from the 9.0 branch in #4625.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/StandardHedgingHandlerBuilderExtensions.cs
geeknoid: Yes, I think the code was previously broken in that respect.

geeknoid: Devs continuously use "ArgumentNullException" for non-arguments...

geeknoid: I think the Throws.IfNull pattern makes this worse in that respect, developers just call this for any null value instead of just for arguments being null.

geeknoid (approved):

eerhardt:
> hrm... Looking at the reprot it is complaining about this line:

I dug deeper and it is complaining about these 3 setters:

https://github.com/dotnet/extensions/blob/7db3a7edc35ae8008c75631da35c01467912cacf/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Logging/LoggingRedactionOptions.cs#L58-L82

However, when I look back at other CI runs and dig into their `merged.cobertura.xml` files, I'm seeing some weirdness.

I downloaded the merged xml file from https://dev.azure.com/dnceng-public/public/_build/results?buildId=522132, and that one has 100% for the class/package.

But then I download it from https://dev.azure.com/dnceng-public/public/_build/results?buildId=524014, and that one has
`<package line-rate="0.9916666666666667" branch-rate="1" complexity="168" name="Microsoft.AspNetCore.Diagnostics.Middleware">`, but the code coverage task passed in CI. So I'm at a loss at what is going wrong here, or why my change would cause this.

I verified that the source generator sets these properties, so I have no idea why this change is causing this to pop up now.

eerhardt:
I've traced down the code coverage issue to https://github.com/dotnet/runtime/issues/96873.

It looks like the only test we have in this area is

https://github.com/dotnet/extensions/blob/7db3a7edc35ae8008c75631da35c01467912cacf/test/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware.Tests/Logging/HttpLoggingServiceExtensionsTests.cs#L31-L40

Which isn't setting these properties. I'll add to this test to ensure the properties get set so we can keep code coverage at 100%.

src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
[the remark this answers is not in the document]
eerhardt: I figured out why coverage dropped, see https://github.com/dotnet/extensions/pull/4871#issuecomment-1888094778. I will keep MinCodeCoverage=100 here.

PR #4892, opened 2024-01-19
Expose AddTcpEndpointProbe

### LINKED ISSUE #4879: [API Proposal]: Expose AddTcpEndpointHealthCheck
### Background and motivation

Currently, `Microsoft.Extensions.Diagnostics.Probes` package provides a convenient method `AddKubernetesProbes` for registering three TCP endpoints (Liveness, Startup, and Readiness probes) internally using the `AddTcpEndpointHealthCheck` method.
However, the `AddTcpEndpointHealthCheck` method is marked as internal, limiting its accessibility for users who may not be hosting in Kubernetes (even if I understand that it's compatible outside Kubernetes as it just exposes TCP endpoints in fact) or those who do not require all three probes.

### API Proposal

I propose exposing the `AddTcpEndpointHealthCheck` method publicly to allow users more flexibility in configuring TCP health checks probes based on their specific needs:

```diff
namespace Microsoft.Extensions.Diagnostics.Probes;

- internal static class TcpEndpointHealthCheckExtensions
+ public static class TcpEndpointHealthCheckExtensions
{
-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services)

-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name)

-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, Action<KubernetesProbesOptions.EndpointOptions> configure)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, Action<TcpEndpointOptions> configure)

-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name, Action<KubernetesProbesOptions.EndpointOptions> configure)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name, Action<TcpEndpointOptions> configure)

-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, IConfigurationSection configurationSection)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, IConfigurationSection configurationSection)

-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name, IConfigurationSection configurationSection)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name, IConfigurationSection configurationSection)
}
```

`KubernetesProbesOptions.EndpointOptions` should probably not be a subclass of `KubernetesProbesOptions` and renamed `TcpEndpointOptions` as well:
```diff
namespace Microsoft.Extensions.Diagnostics.Probes;

- public partial class KubernetesProbesOptions
{
-     public class EndpointOptions { }
+     public class TcpEndpointOptions { }
}
```

### API Usage

```csharp
services.AddTcpEndpointHealthCheck(options =>
{
    options.TcpPort = 5000;
    options.MaxPendingConnections = 20;
    options.HealthAssessmentPeriod = TimeSpan.FromMinutes(1);
})
```

### Alternative Designs

_No response_

### Risks

Nothing I can see now. This package is still in "dev" stage so modifications to the exposed APIs (rename of `KubernetesProbesOptions.EndpointOptions` class) can be done.

joegoldman2:
Fixes #4879.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/extensions/pull/4892)


src/Libraries/Microsoft.Extensions.Diagnostics.Probes/TcpEndpointOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Shared.Data.Validation;
+
+namespace Microsoft.Extensions.Diagnostics.Probes;
+
+/// <summary>
+/// Options to control TCP-based health check probes.
+/// </summary>
+[SuppressMessage("Major Code Smell", "S109:Magic numbers should not be used", Justification = "In place numbers make the ranges cleaner")]
+[SuppressMessage("Design", "CA1034:Nested types should not be visible", Justification = "It's fine")]
+public class TcpEndpointOptions
```
joegoldman2, line 17: @geeknoid perhaps a question to answer during your API reviews meeting: Should this class be named `TcpEndpointHealthCheckOptions` to match other file names?

classes

Or as this class is in the namespace `Microsoft.Extensions.Diagnostics.Probes`, it is enough to indicate that these are the options for the health check probe perhaps.

src/Libraries/Microsoft.Extensions.Diagnostics.Probes/TcpEndpointHealthCheckExtensions.cs
```diff
@@ -12,17 +12,17 @@
 namespace Microsoft.Extensions.Diagnostics.Probes;
```
geeknoid, line 12: Change to Microsoft.Extensions.DependencyInjection, that's where extensions on IServiceCollection go.

geeknoid (commented):
Change looks good, but let's rename the types and files please:

```
TcpEndpointProbesExtensions
TcpEndpointProbesService
TcpEndpointProbesOptions
TcpEndpointProbesOptionsValidator
```

The origin of this code had a different naming convention and use case. When we ported the code to dotnet/extensions, we only bothered renaming the public symbols. But now that you're making private stuff public, those should be renamed as well.

joegoldman2:
I renamed all the files and types as requested.

src/Libraries/Microsoft.Extensions.Diagnostics.Probes/TcpEndpointProbesExtensions.cs
```diff
@@ -36,19 +37,19 @@ internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollec
     /// <param name="services">The <see cref="IServiceCollection" /> to add the services to.</param>
     /// <param name="name">Name used to retrieve the options.</param>
     /// <returns>The value of <paramref name="services"/>.</returns>
-    internal static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name)
+    public static IServiceCollection AddTcpEndpointHealthCheck(this IServiceCollection services, string name)
```
geeknoid, line 40: Whoops, this also needs to be renamed. AddTcpEndpointProbe.

Thanks.

joegoldman2: Done.

geeknoid (approved):
Thanks for pitching in!

PR #4945, opened 2024-02-14
Fix Microsoft.Extensions.AuditReports output path

xakep139:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/4945)


xakep139:
@joperezr is my understanding correct, that it will land in 8.3?

joperezr:
Correct, all changes going into main now, up until 1 week before the release, will be included in 8.3

joperezr (approved):
While not directly related to this, but would it be possible to also use this PR to finish up the README for this package? Right now there is basically no docs at all so the only way to consume this package is to know about it, and then to know how to read MSBuild in order to figure out what needs to be done in order to generate the reports.

mobratil (approved):

xakep139:
> While not directly related to this, but would it be possible to also use this PR to finish up the README for this package? Right now there is basically no docs at all so the only way to consume this package is to know about it, and then to know how to read MSBuild in order to figure out what needs to be done in order to generate the reports.

Good point, will fill README

src/Shared/DiagnosticIds/DiagnosticIds.cs
```diff
@@ -111,6 +111,12 @@ internal static class Metrics
         internal const string METGEN018 = nameof(METGEN018);
         internal const string METGEN019 = nameof(METGEN019);
     }
+
+    internal static class AuditReports
+    {
+        internal const string AUDREP000 = nameof(AUDREP000);
+        internal const string AUDREP001 = nameof(AUDREP001);
```
RussKie, line 118: Do we need to document these in https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md

xakep139: Done both: https://aka.ms/dotnet-extensions-warnings/AUDREPGEN001

global.json
```diff
@@ -1,9 +1,9 @@
 {
   "sdk": {
-    "version": "8.0.101"
+    "version": "8.0.200"
```
xakep139, line 3: @RussKie are we allowed to change that manually?

RussKie: yes, any time as required

dariusclay (approved):

PR #4960, opened 2024-02-21
Set ProducesDotNetReleaseShippingAssets property in Publishing.props

MilenaHristova:
Add a boolean property named `ProducesDotNetReleaseShippingAssets` in Publishing.props for repos that produce .NET shipping packages (packages that we ship with the release infra)
Based on this property we will select which packages to ship as part of .NET on release day.

This is a infrastructure only change. It will add extra metadata to the MergedManifest.xml produced during CI build.

Issue: https://github.com/dotnet/release/issues/822

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -2,5 +2,6 @@
 <Project>
   <PropertyGroup>
     <PublishingVersion>3</PublishingVersion>
+    <ProducesDotNetReleaseShippingAssets>true</ProducesDotNetReleaseShippingAssets>
   </PropertyGroup>
 </Project>


RussKie (approved):

RussKie:
@MilenaHristova please note our main branch is 8.x and not 9.x. Our dev branch is 9.x, and the change will be automatically forward ported to that branch.

MilenaHristova:
> @MilenaHristova please note our main branch is 8.x and not 9.x. Our dev branch is 9.x, and the change will be automatically forward ported to that branch.

@RussKie we should get this change to all branches that produce packages we ship - so for 9.x, 8.x, 7.x and 6.x

Issue #4959, opened 2024-02-20
AsyncState breaks when used in multiple service collections


xakep139:
### Description

Current implementation of AsyncState has both its pool and `AsyncLocal` as `static` fields, but the counter isn't `static`:
https://github.com/dotnet/extensions/blob/a8e17517bd5150cc0f992b66aa6591c7c6fbdfc4/src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs#L17

This counter is used to generate indexes of `IAsyncContext`s, thus when used within different service collections, it start to pointing to items that were added previously (and they have different types potentially).

### Reproduction Steps

```csharp
using Microsoft.Extensions.AsyncState;
using Microsoft.Extensions.DependencyInjection;

await using var spOne = PrepareAsyncState(new Tuple<double>(3.14));
await using var spTwo = PrepareAsyncState(new Tuple<int>(42));

// This line throws:
_ = spOne.GetRequiredService<IAsyncContext<Tuple<double>>>().Get();

static ServiceProvider PrepareAsyncState<T>(T value)
	where T : notnull
{
	var services = new ServiceCollection().AddAsyncState().BuildServiceProvider();
	services.GetRequiredService<IAsyncState>().Initialize();
	services.GetRequiredService<IAsyncContext<T>>().Set(value);
	return services;
}

```

### Expected behavior

Code doesn't throw.

### Actual behavior

The code throws `InvalidCastException`.

### Regression?

No

### Other information

Related discussion exists regarding the `IHttpContextAccessor`: https://github.com/aspnet/HttpAbstractions/pull/753

xakep139:
cc @davidfowl @andrey-noskov @geeknoid @galakt

PR #4966, opened 2024-02-23
AsyncState can be used across different service providers

mobratil:
Current implementation of AsyncState has problems. 
This PR fixes https://github.com/dotnet/extensions/issues/4959 by making the field _contextCount static.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/4966)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
+++ b/src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
@@ -14,7 +14,7 @@ internal sealed class AsyncState : IAsyncState
 {
     private static readonly AsyncLocal<AsyncStateHolder> _asyncContextCurrent = new();
     private static readonly ObjectPool<List<object?>> _featuresPool = PoolFactory.CreatePool(new FeaturesPooledPolicy());
-    private int _contextCount;
+    private static int _contextCount;
 
     public void Initialize()
     {
--- a/src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
+++ b/src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
@@ -104,9 +104,9 @@ internal sealed class AsyncState : IAsyncState
         }
     }
 
-#pragma warning disable CA1822
+#pragma warning disable CA1822 // Member 'ContextCount' does not access instance data and can be marked as static.
     internal int ContextCount => Volatile.Read(ref _contextCount);
-#pragma warning restore CA1822
+#pragma warning restore CA1822 // Member 'ContextCount' does not access instance data and can be marked as static.
 
     private sealed class AsyncStateHolder
     {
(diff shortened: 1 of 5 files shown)


test/Libraries/Microsoft.Extensions.AsyncState.Tests/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+using Xunit;
```
RussKie, line 1: error IDE0073: A source file is missing a required header.

src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
```diff
@@ -104,7 +104,9 @@ internal static void EnsureCount(List<object?> features, int count)
         }
     }
 
+#pragma warning disable CA1822
```
RussKie, line 107: It would be good to add a comment explaining what this warning do, e.g.
```suggestion
#pragma warning disable CA1822 // Comments goes here....
```

test/Libraries/Microsoft.Extensions.AsyncState.Tests/AsyncStateTests.cs
```diff
@@ -229,4 +231,23 @@ public void EnsureCount_WhenCountEqualWithExpected()
         AsyncState.EnsureCount(l, 5);
         Assert.Equal(5, l.Count);
     }
+
+    [Fact]
+    public async Task AsyncStateCanBeUsedInDifferentServiceProviders()
```
RussKie, line 236: error S2699: Add at least one assertion to this test case.

RussKie (commented):
Few more issues to tackle

src/Libraries/Microsoft.Extensions.AsyncState/AsyncState.cs
mobratil: I added explanation comment

test/Libraries/Microsoft.Extensions.AsyncState.Tests/AssemblyInfo.cs
mobratil: Resolved

RussKie (approved):

sebastienros:
The counter is always increasing and declared as `int`. Is the counter tied to the pool, meaning that it won't increase more than the number of pooled objects, or is it incremented with each `AsyncState` resolution (multiple times per request) and possibly exceeding the `int` capacity?

mobratil:
@sebastienros Yes, the int is increasing but only when we create a new instance of AsyncContext<>. But we register these types as singletons (https://github.com/dotnet/extensions/blob/88d79f71e9af747fceb087ea9261b697d29ae71d/src/Libraries/Microsoft.Extensions.AsyncState/AsyncStateExtensions.cs#L27) so there's limited number of these instances.

PR #4978, opened 2024-03-04
try dependabot

danmoseley:
This doesn't work in my fork -- "Dependabot requires a .(cs|vb|fs)proj to evaluate your .NET dependencies. It had expected to find one at the path: /<anything>.(cs|vb|fs)proj." .. which is a new one.

I cannot figure out why it's doing that, as it's essentially the same as the one in dotnet/aspire. I wonder whether we should just merge it in here anyway and see if it works here...
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/4978)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,61 @@
+version: 2
+registries:
+  public-nuget:
+    type: nuget-feed
+    url: https://api.nuget.org/v3/index.json
+updates:
+  - package-ecosystem: nuget
+    directory: "/"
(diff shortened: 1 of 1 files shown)


RussKie (approved):

danmoseley:
Hmm it's having the same problem still. How odd. Well I'll try more things in my fork .. maybe this is another dependabot bug.

Issue #4969, opened 2024-02-26
The logging source generator emits `CS0757` errors when referencing `Microsoft.Extensions.Resilience`


ZzZombo:
### Description

See also #4286, [#4565](https://github.com/dotnet/extensions/issues/4565).

Right after I added this package reference to my SDK-style project targeting `net8.0-windows` for my WPF program I started to get these errors.

### Reproduction Steps

A new WPF application targeting .NET 8, using the following packages:

```
 		<PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="8.0.0" />
		<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
		<PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
		<PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.0" />
		<PackageReference Include="Microsoft.Extensions.Options" Version="8.0.2" />
		<PackageReference Include="Microsoft.Extensions.Resilience" Version="8.2.0" />
```

### Expected behavior

No error.

### Actual behavior

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0757	A partial method may not have multiple implementing declarations	Program	G:\home\dev\csharp\temp\src\Microsoft.Extensions.Logging.Generators\Microsoft.Extensions.Logging.Generators.LoggerMessageGenerator\LoggerMessage.g.cs	13	Active
```

### Regression?

See also #4286, [#4565](https://github.com/dotnet/extensions/issues/4565).

### Known Workarounds

None barring not using the package.

### Configuration

.NET SDK:
 Version:           8.0.200
 Commit:            438cab6a9d
 Workload version:  8.0.200-manifests.e575128c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19045
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.200\

.NET workloads installed:
 [maui-windows]
   Installation Source: VS 17.9.34616.47
   Manifest Version:    8.0.6/8.0.100
   Manifest Path:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.maui\8.0.6\WorkloadManifest.json
   Install Type:        FileBased

 [android]
   Installation Source: VS 17.9.34616.47
   Manifest Version:    34.0.52/8.0.100
[...]

PR #5024, opened 2024-03-12
Fix logging source-gens duplication on WPF

xakep139:
This is a naive fix for the #4969

We can also consider fixing `RemoveDuplicateAnalyzers` Target since it basically adds back the analyzer:
https://github.com/dotnet/ResXResourceManager/blob/c7f0d75c55917c5de01f807394c37dd2af11251b/src/Directory.Build.targets#L28
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5024)


RussKie:
@tarekgh is there someone else you could suggest to rope in for a review of this?

tarekgh:
@joperezr can advise about that. I have no idea why the analyzer get injected twice in the WPF scenario (if I am understanding the issue correctly).

xakep139:
> Reading the comment [#4969 (comment)](https://github.com/dotnet/extensions/issues/4969#issuecomment-1991870300) it suggests the issue is not specific to extension telemetry library and it suggest the fix should be somewhere else maybe in SDK targets? I am not sure.

In fact, it is specific for `Microsoft.Extensions.Telemetry.Abstractions` package. I left that comment to highlight that the root cause isn't related to `Microsoft.Extensions.Resilience` package - as the author wrote there.

xakep139:
> @joperezr can advise about that. I have no idea why the analyzer get injected twice in the WPF scenario (if I am understanding the issue correctly).

Correct, the issue doesn't happen on console projects, nor does it happen for WinForms.

RussKie:
If it's WPF specific, perhaps we should get WPF folks roped in? E.g., ping @dotnet/dotnet-wpf alias.

xakep139:
@dotnet/dotnet-wpf folks, can you please suggest a good way of disabling an analyzer/source-generator (or confirm that the approach in this PR is acceptable) for WPF apps?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -13,4 +13,19 @@
       <Analyzer Remove="@(_Microsoft_Extensions_Logging_AbstractionsAnalyzer)" />
     </ItemGroup>
   </Target>
-</Project>
+  <Target Name="_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzersWPF"
```
joperezr, line 16: Do we understand why this only repros for WPF projects? It sounds like maybe we may not be sure yet. I wonder if it has to do with the fact that they use a different TFM (as WPF uses netx.y-windows), and so build targets are not imported correctly. Have we validated that this fix works in a WPF app?

RussKie: > Do we understand why this only repros for WPF projects?

I _think_ that's because WPF generates a temporary project during the build phase to compile xaml, and that temp project isn't importing/executing the targets we expect. @xakep139 please correct me, if I got it wrong.

xakep139: That's my understanding as well. There's an additional `RemoveDuplicateAnalyzers` target that is added when temp project is compiled

ZzZombo:
Is there anything interim for us to implement on our end in order to progress while you figure this out? Thanks in advance.

RussKie:
> Is there anything interim for us to implement on our end in order to progress while you figure this out? Thanks in advance.

Copy the new target into your project, and it should remove the analyzers from the intermediate project.

RussKie:
Setting to draft for now until we agree this is the best way to proceed.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -13,4 +13,19 @@
...
+        Condition="'%(Analyzer.NuGetPackageId)' == 'Microsoft.Extensions.Logging.Abstractions' Or
+                   '%(Analyzer.AssemblyName)' == 'Microsoft.Extensions.Logging.Generators'" />
+    </ItemGroup>
+
+    <!-- Remove Microsoft.Extensions.Logging.Abstractions Analyzer for WPF projects -->
+    <ItemGroup>
+      <Analyzer Remove="@(_Microsoft_Extensions_Logging_AbstractionsAnalyzerWPF)" />
+    </ItemGroup>
+  </Target>
+</oject>
```
ZzZombo, line 31: Typo right here.

ZzZombo:
I can confirm the change works! Thanks!

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
xakep139: Good catch, thanks!

xakep139:
```suggestion
</Project>
```

gdamiaN538:
xakep139/log-source-gen-WPF-bug

gdamiaN538 (approved):
Directory.Build.targets

gdamiaN538 (commented):
Directory.Build.targets

RussKie:
I propose the following change instead:

```diff
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net8.0/Microsoft.Extensions.Telemetry.Abstractions.targets
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net8.0/Microsoft.Extensions.Telemetry.Abstractions.targets
@@ -2,10 +2,10 @@
   <!-- This package should replace the Microsoft.Extensions.Logging.Abstractions source generator. -->
   <Target Name="_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzers" 
           Condition="'$(DisableMicrosoftExtensionsLoggingSourceGenerator)' == 'true'"
-          AfterTargets="ResolveReferences">
+          BeforeTargets="CoreCompile">
     <ItemGroup>
-      <_Microsoft_Extensions_Logging_AbstractionsAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.AssemblyName)' == 'Microsoft.Extensions.Logging.Generators' Or
-                                                                                           '%(Analyzer.NuGetPackageId)' == 'Microsoft.Extensions.Logging.Abstractions'" />
+      <_Microsoft_Extensions_Logging_AbstractionsAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.Filename)' == 'Microsoft.Extensions.Logging.Generators' Or
+                                                                                           '%(Analyzer.Filename)' == 'Microsoft.Extensions.Logging.Abstractions'" />
     </ItemGroup>
 
     <!-- Remove Microsoft.Extensions.Logging.Abstractions Analyzer -->
```

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -13,4 +13,19 @@
       <Analyzer Remove="@(_Microsoft_Extensions_Logging_AbstractionsAnalyzer)" />
     </ItemGroup>
   </Target>
+  <Target Name="_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzersWPF"
+          Condition="'$(DisableMicrosoftExtensionsLoggingSourceGenerator)' == 'true'"
+          AfterTargets="RemoveDuplicateAnalyzers">
```
xakep139, line 18:
```suggestion
          BeforeTargets="CoreCompile">
```

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -13,4 +13,19 @@
...
+    <ItemGroup>
+      <_Microsoft_Extensions_Logging_AbstractionsAnalyzerWPF
+        Include="@(Analyzer)"
+        Condition="'%(Analyzer.NuGetPackageId)' == 'Microsoft.Extensions.Logging.Abstractions' Or
```
xakep139, line 22:
```suggestion
        <_Microsoft_Extensions_Logging_AbstractionsAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.Filename)' == 'Microsoft.Extensions.Logging.Generators' Or
```

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -13,4 +13,19 @@
...
+                   '%(Analyzer.AssemblyName)' == 'Microsoft.Extensions.Logging.Generators'" />
```
xakep139, line 23:
```suggestion
                   '%(Analyzer.Filename)' == 'Microsoft.Extensions.Logging.Abstractions'" />
```

xakep139:
@RussKie, thanks! I applied your change, please review

RussKie (approved):

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -1,11 +1,11 @@
 <Project>
   <!-- This package should replace the Microsoft.Extensions.Logging.Abstractions source generator. -->
-  <Target Name="_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzers" 
+  <Target Name="_Microsoft_Extensions_Logging_AbstractionsRemoveAnalyzers"
           Condition="'$(DisableMicrosoftExtensionsLoggingSourceGenerator)' == 'true'"
-          AfterTargets="ResolveReferences">
+          BeforeTargets="CoreCompile">
```
joperezr, line 5: I haven't followed all of the thread here, but can you explain why this needs to happen before corecompile as opposed to after RAR? Are we expecting any other target to add the analyzer that is not RAR? For things like this, having a comment explaining why we want to run at a specific time is helpful.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -1,11 +1,11 @@
...
     <ItemGroup>
-      <_Microsoft_Extensions_Logging_AbstractionsAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.AssemblyName)' == 'Microsoft.Extensions.Logging.Generators' Or
-                                                                                           '%(Analyzer.NuGetPackageId)' == 'Microsoft.Extensions.Logging.Abstractions'" />
+      <_Microsoft_Extensions_Logging_AbstractionsAnalyzer Include="@(Analyzer)" Condition="'%(Analyzer.Filename)' == 'Microsoft.Extensions.Logging.Generators' Or
```
joperezr, line 7: Not sure I'm following this. The analyzer will always be called Microsoft.Extensions.Logging.Generators, right? When will it be called Microsoft.Extensions.Logging.Abstractions?

xakep139: The analyzer (source-generator) is shipped as part of the `Microsoft.Extensions.Logging.Abstractions` package. You can check the line below to ensure that: https://github.com/dotnet/extensions/blob/75a0edec8499be9c348b708bf3730a81d93d5e15/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Microsoft.Extensions.Telemetry.Abstractions.csproj#L44

xakep139: Good catch. @RussKie, should we remove the second part of the condition?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
xakep139: You can check Igor's message: https://github.com/dotnet/extensions/issues/4969#issuecomment-2116528521
TL;DR: this works universally for both regular projects (e.g. console, web API) and WPF apps.

joperezr: In that case, can we add a comment above explaining this? While we don't have a comment there today explaining why it was after ResolveReferences, that one feels obvious as it is the one that typically add's stuff the Analyzer item. On the other hand, it is not immediately obvious why we'd put it in the before CoreCompile, and especially now that we know there is an edge case we are covering for with WPF, so I'd probably write down a comment explaining that just doing it after RAR is not enough, and share this scenario.

Also, not sure if you tested, but it would be great to validate that this also works for design time builds. (Design time builds don't run all the exact same targets that are run during a regular build, so if you don't hook properly you could end up breaking intellisense. ResolveReferences is one of the ones that does run in both regular builds and design time builds)

RussKie: `ResolveReferences` target doesn't exist in the WPF's temp project, so the removal of the analyzers
image

Intead the temp project has `RemoveDuplicateAnalyzers` target, which is run before `CoreCompile` target:
image


A WPF fix would be to run our target _after_ `RemoveDuplicateAnalyzers` target, however, since that target is Windows Desktop specific, a more generic fix is to run our target before `CoreCompile`.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
RussKie: Tbh, I don't understand why the original condition was written the way it was - that's a question to @joperezr.

I used `%(Analyzer.Filename)` because it is a) more obvious and b) this is what I saw inspecting binlogs.

If there's no real reason/need for the second condition - I'd remove it.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
xakep139: @joperezr I validated that it works both in compile and design time. Both WPF and WinForms apps are working just fine.
The new [WinUI](https://learn.microsoft.com/en-us/windows/apps/winui/) doesn't work though, so I guess I will have to update the targets again.

RussKie: > The new [WinUI](https://learn.microsoft.com/en-us/windows/apps/winui/) doesn't work though, so I guess I will have to update the targets again.

@xakep139 could you share more information? E.g., a binlog.

xakep139: So, for WinUI applications the `CoreCompile` target happens after the `XamlPreCompile` target which causes the error:
image

I would recommend to stick with the initial solution and have the "standard" `BeforeTargets="ResolveReferences"` for non-WPF projects, and a separate `RemoveDuplicateAnalyzers` target for WPF ones.

xakep139: Please review the changes, folks

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
xakep139: I decided to get back to the original approach, as it makes sense to check the NuGet package ID. But please let me know if you have any doubts

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
```diff
@@ -1,16 +1,34 @@
...
+                   '%(Analyzer.NuGetPackageId)' == 'Microsoft.Extensions.Logging.Abstractions'" />
     </ItemGroup>
 
     <!-- Remove Microsoft.Extensions.Logging.Abstractions Analyzer -->
     <ItemGroup>
       <Analyzer Remove="@(_Microsoft_Extensions_Logging_AbstractionsAnalyzer)" />
     </ItemGroup>
   </Target>
+
+  <!-- We have a separate target for WPF projects because there's "GenerateTemporaryTargetAssembly" target for those -->
```
RussKie, line 19: The comment is a good idea, but I feel like it could be expanded a little to make the reason clearer.

RussKie (commented):
:shipit:

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/buildTransitive/net6.0/Microsoft.Extensions.Telemetry.Abstractions.targets
xakep139: Any suggestions?

RussKie: I'd summarise the information from the discussion above so won't need to scratch our heads two months from now wondering why this target is needed. So, I'd make the comment something like this:

_We have a separate target for WPF projects because generates a generates and compiles a temporary project, and that temp project doesn't have "ResolveReferences" we expect. This leads to the runtime's generator not being removed and to a compile-time failure. This target resolves the problem._

xakep139: Thanks for the suggestion, applied it

geeknoid (approved):

joperezr:
Thanks @xakep139!

bench/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware.PerformanceTests/Microsoft.AspNetCore.Diagnostics.Middleware.PerformanceTests.csproj
```diff
@@ -6,6 +6,7 @@
 
   <PropertyGroup>
     <InjectSharedPools>true</InjectSharedPools>
+    <SkipSharedPoolsReferences>true</SkipSharedPoolsReferences>
```
RussKie, line 9: @xakep139 could you please elp me understand the connection between this and the rest of the change?

xakep139: Sure thing. There was a warning complaining about an unused package in this project. I added a condition into the `Shared.props` to avoid touching other projects.

xakep139: You can see the warning here: https://github.com/dotnet/extensions/runs/25307769976

RussKie: Thank you

PR #5068, opened 2024-03-26
Add support for Linux cgrpoup v2, issue-4885

nezdali:
- Introduced ILinuxUtilizationParser interface to be able to choose between parsers during DI
- Introduced new class for cgroup v2 specifics LinuxUtilizationParserCgroupV2
- GetHostCpuUsageInNanoseconds() parses /sys/fs/cgroup/cpu.stat file and calculates cpu usage in microseconds
- TryGetCpuUnitsFromCgroups() parses cpu quota and period from a single file /sys/fs/cgroup/cpu.max
- TryGetCgroupRequestCpu() parses /sys/fs/cgroup/cpu.weight and calculates from cpu weight to shares (1024 share == 1 CPU core) in order to get the Pod's cpu request limit.
- Introduced new method GetMemoryUsageInBytesFromSlices() to calculate memory stats from slices (cgroup v2 specific)
- Introduced an overload in SystemResources struct where new metric is added guaranteedPodCpuUnits, which get cpu request limit (as per k8s terminology)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5068)


mobratil:
@nezdali , I deployed your change into my own cluster that has cgroupsv2 and printed SystemResources object.

The cluster have following resources in yaml:
          resources:
            requests:
              cpu: "300m"
              memory: "1.5Gi"
            limits:
              cpu: "900m"
              memory: "3.5Gi"

The printed values are following.

  System Resources: 
          Requests:
                  CPU = 309m,			(based on GuaranteedPodCpuUnits)
                  Memory = 3.5Gi,		(based on GuaranteedMemoryInBytes)
          Limits:
                  CPU = 900m,			(based on MaximumCpuUnits)
                  Memory: 15.61508560180664Gi	(based on MaximumMemoryInBytes)


While the requests for CPU corresponds to YAML specification, the memory seems to be not accurate.
Memory limit (MaximumMemoryInBytes) seems to be showing the capacity of the host machine instead of the pod limit.
The memory request (GuaranteedMemoryInBytes) contains the pod limit.

nezdali:
After more research it turns out that GuaranteedMemoryInBytes is the Resource Memory Limit (in k8s terms) which is read from /sys/fs/cgroup/memory.max
And currently there is no way to get the Resource Memory Request. It is always 0 in /sys/fs/cgroup/memory.min , for cgroupv2 container, VM (AzureLinux, Ubuntu)
K8s documentation on how memory.min  populated https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2570-memory-qos/#readme

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -63,7 +63,11 @@ internal sealed class LinuxUtilizationProvider : ISnapshotProvider
         _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.CpuUtilization, observeValue: CpuUtilization, unit: "1");
         _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.MemoryUtilization, observeValue: MemoryUtilization, unit: "1");
 
-        Resources = new SystemResources(1, hostCpus, _totalMemoryInBytes, hostMemory);
+        // cpuGuaranteedRequest is a Cpu request for POD, for HOST its 1 Core
+        // availableCpus is a Cpu limit for Pod or for a HOST.
```
RussKie, line 67:
```suggestion
        // cpuGuaranteedRequest is a CPU request for pod, for host its 1 core
        // available CPUs is a CPU limit for a pod or for a host.
```

mobratil (changes requested):

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/IFileSystem.cs
```diff
@@ -12,6 +12,18 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux;
...
 internal interface IFileSystem
 {
+    /// <summary>
+    /// Checks for file existence.
+    /// </summary>
+    /// <returns> True/False.</returns>
+    bool Exists(FileInfo fileInfo);
+
+    /// <summary>
+    /// Get directory name on the filesystem.
```
xakep139, line 22:
```suggestion
    /// Get directory names on the filesystem based on the provided pattern.
```

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/ILinuxUtilizationParser.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    /// <returns>maybeMemory.</returns>
+    ulong GetAvailableMemoryInBytes();
+
+    /// <summary>
+    /// Reads the file /sys/fs/cgroup/cpu.stat, which is part of the cgroup v2 CPU controller.
+    /// It provides statistics about the CPU usage of a cgroup.
+    /// </summary>
+    /// <returns>nanoseconds.</returns>
+    long GetCgroupCpuUsageInNanoseconds();
```
xakep139, line 24: Do we need to use `Cgroup` in method names here?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParser.cs
```diff
@@ -14,8 +14,10 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux;
 /// This class is not thread safe.
 /// When the same instance is called by multiple threads it may return corrupted data.
 /// </remarks>
-internal sealed class LinuxUtilizationParser
+internal sealed class LinuxUtilizationParser : ILinuxUtilizationParser
```
xakep139, line 17: Should this class be renamed to `LinuxUtilizationParserCgroupV1` then?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/ILinuxUtilizationParser.cs
nezdali: This method is specifically used to get metrics created by control group (cgroup). Another method GetHostCpuUsageInNanoseconds() reads /proc/cpuinfo which is for host in general.

xakep139: I see, thanks

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParser.cs
nezdali: Actually both parsers do not parse only cgroup fs at /sys/fs/cgroup , but also /proc , but the main purpose is to parse the cgroup.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -63,7 +63,11 @@ internal sealed class LinuxUtilizationProvider : ISnapshotProvider
...
+        // 1 - Currently value in /sys/fs/cgroup/memory.min per container is 0
```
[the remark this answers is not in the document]
nezdali: no, fixed

mobratil (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/IFileSystem.cs
```diff
@@ -12,6 +12,18 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux;
...
+    /// Checks for file existence.
+    /// </summary>
+    /// <returns> True/False.</returns>
+    bool Exists(FileInfo fileInfo);
+
+    /// <summary>
+    /// Get directory names on the filesystem based on the provided pattern.
+    /// </summary>
+    /// <returns>string.</returns>
+    string[] GetDirectoryNames(string directory, string pattern);
```
RussKie, line 25: Even though this is an internal class I'd caution against using mutable arrays as return values. This could lead to some gnarly hard to track bugs.
Could this be changed to something like `IList<string>` or `IEnumerable<string>`?

nezdali: Makes sense, will do the changes in another PR

PR #5112, opened 2024-04-16
Add support for processing requests with StreamContent to AddStandardHedgingHandler()

### LINKED ISSUE #5105: [New Feature Proposal] Add support for HttpRequestMessage objects containing StreamContent to AddStandardHedgingHandler()
### Background and motivation
Many clients trying to use the `AddStandardHedgingHandler()` resilience API based on top of Polly v8 have requirements which force them send `HttpRequestMessage` objects that contains `StreamContent`. Today, if a client built with an `IHttpClientBuilder` that was configured with a resilience handler via the `AddStandardHedgingHandler()` API attempts to send an `HttpRequestMessage` object that contains `StreamContent` to a downstream service, then an `InvalidOperationException` will be thrown. This exception is thrown by the `Initialize()` method in `RequestMessageSnapshot`: https://github.com/dotnet/extensions/blob/5fc05f874f74374749b01e284321168c7b92476a/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs#L74-L80

There is no reason to limit users by only supporting a subset of `HttpRequestMessage` `Content` types. Further, StreamContent support can be added while still defaulting to a shallow Content copy if the Content type is not `StreamContent` (i.e. how it's currently working in `RequestMessageSnapshot.Initialize()` today).

#### Proposed Changes to [RequestMessageSnapshot.cs](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs):
```csharp

using System;
using System.IO;
using System.Net.Http;
using System.Threading.Tasks;
using Microsoft.Extensions.ObjectPool;
using Microsoft.Shared.Pools;

internal sealed class RequestMessageSnapshot : IResettable, IDisposable
{
    private static readonly ObjectPool<RequestMessageSnapshot> _snapshots = PoolFactory.CreateResettingPool<RequestMessageSnapshot>();

    private readonly List<KeyValuePair<string, IEnumerable<string>>> _headers = [];
    private readonly List<KeyValuePair<string, object?>> _properties = [];

    private HttpMethod? _method;
    private Uri? _requestUri;
    private Version? _version;
    private HttpContent? _content;

    public static async Task<RequestMessageSnapshot> Create(HttpRequestMessage request)
    {
        var snapshot = _snapshots.Get();
        snapshot.Initialize(request);
        return snapshot;
    }

    public async Task<HttpRequestMessage> CreateRequestMessage()
    {
        var clone = new HttpRequestMessage(_method!, _requestUri)
        {
            Version = _version!
        };

        if (_content is StreamContent)
        {
            (HttpContent content, HttpContent clonedContent) = await CloneContent(_content).ConfigureAwait(false);
            _content = content;
            clone.Content = clonedContent;
        }
        else
        {
            clone.Content = _content;
        }

#if NET5_0_OR_GREATER
        foreach (var prop in _properties)
        {
            _ = clone.Options.TryAdd(prop.Key, prop.Value);
        }
#else
        foreach (var prop in _properties)
        {
            clone.Properties.Add(prop);
        }
#endif
        foreach (KeyValuePair<string, IEnumerable<string>> header in _headers)
        {
            _ = clone.Headers.TryAddWithoutValidation(header.Key, header.Value);
        }

        return clone;
    }

    bool IResettable.TryReset()
    {
        _properties.Clear();
        _headers.Clear();

        _method = null;
        _version = null;
        _requestUri = null;
        if (_content is StreamContent)
        {
            // a snapshot's StreamContent is always a unique copy (deep clone)
            // therefore, it is safe to dispose when snapshot is no longer needed
            _content.Dispose();
        }

        _content = null;

        return true;
    }

    void IDisposable.Dispose() => _snapshots.Return(this);

    private async Task Initialize(HttpRequestMessage request)
    { 
        _method = request.Method;
        _version = request.Version;
        _requestUri = request.RequestUri;
        if (request.Content is StreamContent)
        {
            (HttpContent requestContent, HttpContent clonedRequestContent) = await CloneContent(requestMessage.Content).ConfigureAwait(false);
            _content = clonedRequestContent;
            request.Content = requestContent;
        }
        else
        {
            _content = request.Content;
        }

        // headers
        _headers.AddRange(request.Headers);

        // props
#if NET5_0_OR_GREATER
        _properties.AddRange(request.Options);
#else
        _properties.AddRange(request.Properties);
#endif
    }

    private static async Task<(HttpContent content, HttpContent clonedContent)> CloneContent(HttpContent content)
    {
        HttpContent clonedContent = null;
        if (content != null)
        {
            HttpContent originalContent = content;
            Stream originalRequestBody = await content.ReadAsStreamAsync().ConfigureAwait(false);
            MemoryStream clonedRequestBody = new MemoryStream();
            await originalRequestBody.CopyToAsync(clonedRequestBody).ConfigureAwait(false);
            clonedRequestBody.Position = 0;
            if (originalRequestBody.CanSeek)
            {
                originalRequestBody.Position = 0;
            }
            else
            {
                originalRequestBody = new MemoryStream();
                await clonedRequestBody.CopyToAsync(originalRequestBody).ConfigureAwait(false);
                originalRequestBody.Position = 0;
                clonedRequestBody.Position = 0;
            }

            clonedContent = new StreamContent(clonedRequestBody);
            content = new StreamContent(originalRequestBody);
            foreach (KeyValuePair<string, IEnumerable<string>> header in originalContent.Headers)
            {
                clonedContent.Headers.TryAddWithoutValidation(header.Key, header.Value);
                content.Headers.TryAddWithoutValidation(header.Key, header.Value);
            }
        }

        return (content, clonedContent);
    }
}
```
#### Proposed Changes to [ResilienceHttpClientBuilderExtensions.Hedging.cs](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs):
```csharp
public static IStandardHedgingHandlerBuilder AddStandardHedgingHandler(this IHttpClientBuilder builder)
{
    _ = Throw.IfNull(builder);

    var optionsName = builder.Name;
    var routingBuilder = new RoutingStrategyBuilder(builder.Name, builder.Services);

    builder.Services.TryAddSingleton<Randomizer>();

    _ = builder.Services.AddOptionsWithValidateOnStart<HttpStandardHedgingResilienceOptions, HttpStandardHedgingResilienceOptionsValidator>(optionsName);
    _ = builder.Services.AddOptionsWithValidateOnStart<HttpStandardHedgingResilienceOptions, HttpStandardHedgingResilienceOptionsCustomValidator>(optionsName);
    _ = builder.Services.PostConfigure<HttpStandardHedgingResilienceOptions>(optionsName, options =>
    {
        options.Hedging.ActionGenerator = args =>
        {
            if (!args.PrimaryContext.Properties.TryGetValue(ResilienceKeys.RequestSnapshot, out var snapshot))
            {
                Throw.InvalidOperationException("Request message snapshot is not attached to the resilience context.");
            }

            return async () =>
            {
                try
                {
                    var requestMessage = await snapshot.CreateRequestMessage().ConfigureAwait(false);

                    // The secondary request message should use the action resilience context
                    requestMessage.SetResilienceContext(args.ActionContext);

                    // replace the request message
                    args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);

                    if (args.PrimaryContext.Properties.TryGetValue(ResilienceKeys.RoutingStrategy, out var routingPipeline))
                    {
                        if (!routingPipeline.TryGetNextRoute(out var route))
                        {
                            // no routes left, stop hedging
                            return null;
                        }

                        requestMessage.RequestUri = requestMessage.RequestUri!.ReplaceHost(route);
                    }

                    return await args.Callback(args.ActionContext).AsTask().ConfigureAwait(false);
                }
                catch (Exception e)
                {
                    return Outcome.FromException<HttpResponseMessage>(e);
                }
            };
        };
    });
    ...
```

#### Proposed Changes to [RequestMessageSnapshotStrategy.cs](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/Internals/RequestMessageSnapshotStrategy.cs):
```csharp
protected override async ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
    Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
    ResilienceContext context,
    TState state)
{
    if (!context.Properties.TryGetValue(ResilienceKeys.RequestMessage, out var request) || request is null)
    {
        Throw.InvalidOperationException("The HTTP request message was not found in the resilience context.");
    }

    using var snapshot = await RequestMessageSnapshot.Create(request).ConfigureAwait(false);

    context.Properties.Set(ResilienceKeys.RequestSnapshot, snapshot);

    return await callback(context, state).ConfigureAwait(context.ContinueOnCapturedContext);
}
...
```

adamhammond:
Fixes #5105 

Add support for processing HttpRequestMessage objects containing StreamContent to the AddStandardHedgingHandler() resilience API.

This change does not update any public API contracts. It updates internal and private API contracts only.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5112)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
@@ -90,11 +90,12 @@ public static partial class ResilienceHttpClientBuilderExtensions
                     Throw.InvalidOperationException("Request message snapshot is not attached to the resilience context.");
                 }
 
-                // if a routing strategy has been configured but it does not return the next route, then no more routes
-                // are availabe, stop hedging
+                // if a routing strategy has been configured, get the next route from the routing strategy
                 Uri? route;
                 if (args.PrimaryContext.Properties.TryGetValue(ResilienceKeys.RoutingStrategy, out var routingPipeline))
                 {
+                    // if a routing strategy has been configured but it does not return the next route, then no more routes
+                    // are availabe, stop hedging
                     if (!routingPipeline.TryGetNextRoute(out route))
                     {
                         return null;
@@ -121,7 +122,7 @@ public static partial class ResilienceHttpClientBuilderExtensions
 
                         if (route != null)
                         {
-                            // replace the RequestUri of the request per the routing strategy
+                            // replace the Host on the RequestUri of the request per the routing strategy
                             requestMessage.RequestUri = requestMessage.RequestUri!.ReplaceHost(route);
                         }
                     }
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
@@ -120,7 +120,7 @@ public static partial class ResilienceHttpClientBuilderExtensions
                         // replace the request message
                         args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);
 
-                        if (route != null)
+                        if (route is not null)
                         {
                             // replace the Host on the RequestUri of the request per the routing strategy
                             requestMessage.RequestUri = requestMessage.RequestUri!.ReplaceHost(route);
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
 using System.Net.Http;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Http.Resilience;
@@ -88,26 +90,48 @@ public static partial class ResilienceHttpClientBuilderExtensions
                     Throw.InvalidOperationException("Request message snapshot is not attached to the resilience context.");
                 }
 
-                var requestMessage = snapshot.CreateRequestMessage();
-
-                // The secondary request message should use the action resilience context
-                requestMessage.SetResilienceContext(args.ActionContext);
-
-                // replace the request message
-                args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);
-
+                // if a routing strategy has been configured but it does not return the next route, then no more routes
+                // are availabe, stop hedging
+                Uri? route;
                 if (args.PrimaryContext.Properties.TryGetValue(ResilienceKeys.RoutingStrategy, out var routingPipeline))
                 {
-                    if (!routingPipeline.TryGetNextRoute(out var route))
+                    if (!routingPipeline.TryGetNextRoute(out route))
                     {
-                        // no routes left, stop hedging
                         return null;
                     }
-
-                    requestMessage.RequestUri = requestMessage.RequestUri!.ReplaceHost(route);
+                }
+                else
+                {
+                    route = null;
                 }
 
-                return () => args.Callback(args.ActionContext);
+                return async () =>
+                {
+                    Outcome<HttpResponseMessage>? actionResult = null;
+
+                    try
+                    {
+                        var requestMessage = await snapshot.CreateRequestMessageAsync().ConfigureAwait(false);
+
+                        // The secondary request message should use the action resilience context
+                        requestMessage.SetResilienceContext(args.ActionContext);
+
+                        // replace the request message
+                        args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);
+
+                        if (route != null)
+                        {
+                            // replace the RequestUri of the request per the routing strategy
+                            requestMessage.RequestUri = requestMessage.RequestUri!.ReplaceHost(route);
+                        }
+                    }
+                    catch (IOException e)
+                    {
+                        actionResult = Outcome.FromException<HttpResponseMessage>(e);
+                    }
+
+                    return actionResult ?? await args.Callback(args.ActionContext).ConfigureAwait(args.ActionContext.ContinueOnCapturedContext);
+                };
             };
         });
 
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
@@ -111,7 +111,7 @@ public static partial class ResilienceHttpClientBuilderExtensions
 
                     try
                     {
-                        var requestMessage = await snapshot.CreateRequestMessageAsync().ConfigureAwait(false);
+                        var requestMessage = await snapshot.CreateRequestMessageAsync().ConfigureAwait(args.ActionContext.ContinueOnCapturedContext);
 
                         // The secondary request message should use the action resilience context
                         requestMessage.SetResilienceContext(args.ActionContext);
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Net.Http;
+using System.Threading.Tasks;
 using Microsoft.Extensions.ObjectPool;
 using Microsoft.Shared.Diagnostics;
 using Microsoft.Shared.Pools;
@@ -22,20 +24,39 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
     private Version? _version;
     private HttpContent? _content;
 
-    public static RequestMessageSnapshot Create(HttpRequestMessage request)
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
+    public static async Task<RequestMessageSnapshot> CreateAsync(HttpRequestMessage request)
     {
+        _ = Throw.IfNull(request);
+
         var snapshot = _snapshots.Get();
-        snapshot.Initialize(request);
+        await snapshot.InitializeAsync(request).ConfigureAwait(false);
         return snapshot;
     }
 
-    public HttpRequestMessage CreateRequestMessage()
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
+    public async Task<HttpRequestMessage> CreateRequestMessageAsync()
     {
+        if (IsReset())
+        {
+            throw new InvalidOperationException($"{nameof(CreateRequestMessageAsync)}() cannot be called on a snapshot object that has been reset and has not been initialized");
+        }
+
         var clone = new HttpRequestMessage(_method!, _requestUri)
         {
-            Content = _content,
             Version = _version!
         };
+
+        if (_content is StreamContent)
+        {
+            (HttpContent? content, HttpContent? clonedContent) = await CloneContentAsync(_content).ConfigureAwait(false);
+            _content = content;
+            clone.Content = clonedContent;
+        }
+        else
+        {
+            clone.Content = _content;
+        }
 
 #if NET5_0_OR_GREATER
         foreach (var prop in _properties)
@@ -56,6 +77,7 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
         return clone;
     }
 
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Critical Bug", "S2952:Classes should \"Dispose\" of members from the classes' own \"Dispose\" methods", Justification = "Handled by ObjectPool")]
     bool IResettable.TryReset()
     {
         _properties.Clear();
@@ -64,6 +86,13 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
         _method = null;
         _version = null;
         _requestUri = null;
+        if (_content is StreamContent)
+        {
+            // a snapshot's StreamContent is always a unique copy (deep clone)
+            // therefore, it is safe to dispose when snapshot is no longer needed
+            _content.Dispose();
+        }
+
         _content = null;
 
         return true;
@@ -71,17 +100,62 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
 
     void IDisposable.Dispose() => _snapshots.Return(this);
 
-    private void Initialize(HttpRequestMessage request)
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
+    private static async Task<(HttpContent? content, HttpContent? clonedContent)> CloneContentAsync(HttpContent? content)
     {
-        if (request.Content is StreamContent)
+        HttpContent? clonedContent = null;
+        if (content != null)
         {
-            Throw.InvalidOperationException($"{nameof(StreamContent)} content cannot by cloned.");
+            HttpContent originalContent = content;
+            Stream originalRequestBody = await content.ReadAsStreamAsync().ConfigureAwait(false);
+            MemoryStream clonedRequestBody = new MemoryStream();
+            await originalRequestBody.CopyToAsync(clonedRequestBody).ConfigureAwait(false);
+            clonedRequestBody.Position = 0;
+            if (originalRequestBody.CanSeek)
+            {
+                originalRequestBody.Position = 0;
+            }
+            else
+            {
+                originalRequestBody = new MemoryStream();
+                await clonedRequestBody.CopyToAsync(originalRequestBody).ConfigureAwait(false);
+                originalRequestBody.Position = 0;
+                clonedRequestBody.Position = 0;
+            }
+
+            clonedContent = new StreamContent(clonedRequestBody);
+            content = new StreamContent(originalRequestBody);
+            foreach (KeyValuePair<string, IEnumerable<string>> header in originalContent.Headers)
+            {
+                _ = clonedContent.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                _ = content.Headers.TryAddWithoutValidation(header.Key, header.Value);
+            }
         }
 
+        return (content, clonedContent);
+    }
+
+    private bool IsReset()
+    {
+        return _method == null;
+    }
+
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
+    private async Task InitializeAsync(HttpRequestMessage request)
+    {
         _method = request.Method;
         _version = request.Version;
         _requestUri = request.RequestUri;
-        _content = request.Content;
+        if (request.Content is StreamContent)
+        {
+            (HttpContent? requestContent, HttpContent? clonedRequestContent) = await CloneContentAsync(request.Content).ConfigureAwait(false);
+            _content = clonedRequestContent;
+            request.Content = requestContent;
+        }
+        else
+        {
+            _content = request.Content;
+        }
 
         // headers
         _headers.AddRange(request.Headers);
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
@@ -25,7 +25,7 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
     private HttpContent? _content;
 
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
-    public static async Task<RequestMessageSnapshot> CreateAsync(HttpRequestMessage request)
+    public static async ValueTask<RequestMessageSnapshot> CreateAsync(HttpRequestMessage request)
     {
         _ = Throw.IfNull(request);
 
@@ -35,7 +35,7 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
     }
 
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
-    public async Task<HttpRequestMessage> CreateRequestMessageAsync()
+    public async ValueTask<HttpRequestMessage> CreateRequestMessageAsync()
     {
         if (!IsInitialized())
         {
@@ -101,7 +101,7 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
     void IDisposable.Dispose() => _snapshots.Return(this);
 
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
-    private static async Task<(HttpContent? content, HttpContent? clonedContent)> CloneContentAsync(HttpContent? content)
+    private static async ValueTask<(HttpContent? content, HttpContent? clonedContent)> CloneContentAsync(HttpContent? content)
     {
         HttpContent? clonedContent = null;
         if (content is not null)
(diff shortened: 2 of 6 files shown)


amcasey:
Apparently, the code coverage for M.E.Http.Resilience has dropped from 97% to 96.75% and that's why CI failed.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
```diff
@@ -88,26 +90,48 @@ public static IStandardHedgingHandlerBuilder AddStandardHedgingHandler(this IHtt
...
+                    {
+                        var requestMessage = await snapshot.CreateRequestMessageAsync().ConfigureAwait(args.ActionContext.ContinueOnCapturedContext);
+
+                        // The secondary request message should use the action resilience context
+                        requestMessage.SetResilienceContext(args.ActionContext);
+
+                        // replace the request message
+                        args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);
+
+                        if (route != null)
```
RussKie, line 122: The equality operator can be overloaded.
```suggestion
                        if (route is not null)
```
Ditto other places.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
```diff
@@ -64,24 +86,76 @@ bool IResettable.TryReset()
...
+            }
         }
 
+        return (content, clonedContent);
+    }
+
+    private bool IsReset()
+    {
+        return _method == null;
+    }
```
RussKie, line 141: Should be this be a property?

joperezr:
cc: @iliar-turdushev can you take a peek?

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: It's private, so no; however, I did refactor this to `IsInitialized()`. It eliminates ambiguity since technically an instance can be uninitialized without ever having been reset.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
adamhammond: Done 👍

adamhammond:
> Apparently, the code coverage for M.E.Http.Resilience has dropped from 97% to 96.75% and that's why CI failed.

Is code coverage calculated for the entire package or just for the lines I touched? I have added full test coverage for this change with the exception of the catch-block in two of the files where I catch `IOException`. Test coverage can't be added for this without doing something awkward like extending the `StreamContent` class with virtual methods so that they can be mocked. That would also require adding a certain flag on the assembly to permit this AFAIK. It doesn't seem like this is worth the trouble and added complexity. Thoughts?

RussKie:
> > Apparently, the code coverage for M.E.Http.Resilience has dropped from 97% to 96.75% and that's why CI failed.
> 
> is code coverage calculated for the entire package or just for the lines I touched? 

The coverage is calculated for the whole package.

You can see the build here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=649752&view=logs&j=8cf0092d-1f6c-522b-3c27-9f632d33f4f3&t=cbf2093c-af29-552b-05db-cced36272f1c
...and the report here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=649752&view=codecoverage-tab

IIUIC, the missing coverage is here:
image

You could also collect the coverage locally with `-testcoverage` switch, e.g.: `.\build.cmd -build -testcoverage`. See `.\build.cmd -help` for more info.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
```diff
@@ -64,24 +86,76 @@ bool IResettable.TryReset()
...
 
     void IDisposable.Dispose() => _snapshots.Return(this);
 
-    private void Initialize(HttpRequestMessage request)
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
+    private static async Task<(HttpContent? content, HttpContent? clonedContent)> CloneContentAsync(HttpContent? content)
     {
-        if (request.Content is StreamContent)
+        HttpContent? clonedContent = null;
+        if (content is not null)
```
martintmk, line 107: This introduces performance regression for non-streamed content. Create an extra branch for `StreamContent` where this logic belongs.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
```diff
@@ -64,24 +86,76 @@ bool IResettable.TryReset()
...
         {
-            Throw.InvalidOperationException($"{nameof(StreamContent)} content cannot by cloned.");
+            HttpContent originalContent = content;
+            Stream originalRequestBody = await content.ReadAsStreamAsync().ConfigureAwait(false);
+            MemoryStream clonedRequestBody = new MemoryStream();
+            await originalRequestBody.CopyToAsync(clonedRequestBody).ConfigureAwait(false);
```
martintmk, line 112: The streamed content can point to a large file which can have > GBs in size. Since we are copying to memory we should constraint it. Let's say < 10MB

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
```diff
@@ -22,21 +24,40 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
...
+
         var snapshot = _snapshots.Get();
-        snapshot.Initialize(request);
+        await snapshot.InitializeAsync(request).ConfigureAwait(false);
         return snapshot;
     }
 
-    public HttpRequestMessage CreateRequestMessage()
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Resilience", "EA0014:The async method doesn't support cancellation", Justification = "Past the point of no cancellation.")]
+    public async Task<HttpRequestMessage> CreateRequestMessageAsync()
```
martintmk, line 38:
```suggestion
    public async ValueTask<HttpRequestMessage> CreateRequestMessageAsync()
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
```diff
@@ -22,21 +24,40 @@ internal sealed class RequestMessageSnapshot : IResettable, IDisposable
...
+            throw new InvalidOperationException($"{nameof(CreateRequestMessageAsync)}() cannot be called on a snapshot object that has been reset and/or has not been initialized");
+        }
+
+        var clone = new HttpRequestMessage(_method!, _requestUri?.OriginalString)
         {
-            Content = _content,
             Version = _version!
         };
 
+        if (_content is StreamContent)
```
martintmk, line 50: Not a fan of extra branching, can we keep it to `CloneContentAsync`?

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: @martintmk can you expand on why you believe this will cause a performance regression?

The cloning of the `HttpContent` (i.e. this method) only executes if the `request.Content` is `StreamContent`. See the other places in this class where this `CloneContentAsync` method is called for more context. You will see that each call is wrapped in a `if (request.Content is StreamContent)` check. I did this specifically to prevent any performance regression for existing non-Streamed content users of the `AddStandardHedgingHandler` API.

Also, if you are referring to the `if (request.Content is StreamContent)` check as being the culprit of a new performance regression, please observe that this check already existed before my change. The only difference is that, currently, if `request.Content is StreamContent` is true, then the `RequestMessageSnapshot` methods throw an `InvalidOperationException`. Therefore, perf should remain the same for existing, non-Streamed content users.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: I'm not sure if it's a good idea to impose a size limit. For example, I don't think the framework imposes any such limit when cloning request content for redirected requests. IMO, we should allow this to be limitless, and require users to enforce their own request content size limits, if they have them, via server configurations and/or their own custom handlers, filters, middleware, etc. Thoughts?

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: If we don't branch, then there could be a performance degradation for existing, non-Streamed content users of the `AddStandardHedgingApi` (i.e. the only current users of this `RequestMessageSnapshot` class).  In fact, I believe you told me to create this extra branch in [your other comment](https://github.com/dotnet/extensions/pull/5112#discussion_r1573985985) on this PR:

> This introduces performance regression for non-streamed content. Create an extra branch for `StreamContent` where this logic belongs.

@martintmk apologies if I am misunderstanding what you mean here in relation to your other comment. Can you please clarify?

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: @martintmk done 👍

adamhammond:
> > > Apparently, the code coverage for M.E.Http.Resilience has dropped from 97% to 96.75% and that's why CI failed.
> > 
> > 
> > is code coverage calculated for the entire package or just for the lines I touched?
> 
> The coverage is calculated for the whole package.

@RussKie I was able to add code coverage for the catch-block. I had to create a helper that extended `MemoryStream` with overloaded stream operation-methods to throw `IOException` in order to emulate the needed behavior. After adding the additional coverage, all pipeline checks passed 👍

RussKie:
> > > > Apparently, the code coverage for M.E.Http.Resilience has dropped from 97% to 96.75% and that's why CI failed.
> > > 
> > > 
> > > is code coverage calculated for the entire package or just for the lines I touched?
> > 
> > 
> > The coverage is calculated for the whole package.
> 
> @RussKie I was able to add code coverage for the catch-block. I had to create a helper that extended `MemoryStream` with overloaded stream operation-methods to throw `IOException` in order to emulate the needed behavior. After adding the additional coverage, all pipeline checks passed 👍

Awesome, thank you!
I was about to check that our CI reports the coverage failure back to GitHub but you beat me to it with your fix :)

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
martintmk: I mean instead of creating all three branches, we can keep the logic in one. It would keep the code simpler.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
martintmk: This change can buffer potentially endless stream into memory and crash the process. I think it's something we should guard against.

This thing should be used for relatively small streamed payloads.

Btw, you can try to call `LoadIntoBufferAsync` on the content to see how it behaves.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
martintmk: I see, thanks for detailed explanation. Can you still try to avoid the extra branches by condensing the code into a single `PrepareContentAsync` method
It would also reduce the number of mutants you have to kill.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: @martintmk apologies, I'm still not following what you mean by "keep the logic in one [branch]". Can you expand further? Perhaps provide a code example of what you are suggesting?

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: I do understand the concern; however, the max size that a given stream can grow before it crashes the process is bound only by the memory resources available on the given server that it's executing on. Therefore, by choosing a static max limit, we are creating a cap that might unnecessarily hinder users of the API that have both requirements and the necessary server resources available to process requests with exceptionally large stream content. Further, `RequestMessageSnapshot` is only cloning content from _existing_ request objects that contain _existing_ `StreamContent`. Therefore, a `HttpRequestMessage` object has already been created with `StreamContent` of the given size before any of the `RequestMessageSnapshot` operations are executed. If the size of the `StreamContent` was in fact too large, it would have already crashed the process.

I feel strongly that we should not impose a size limit when cloning the StreamContent.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: Ahh I see what you mean. I understand where you are coming from; however, I don't think encapsulating the content operations behind a single method will achieve the additional clarity or eliminate the required mutation tests as you are suggesting:
1. The mutation tests still need to be accounted for regardless.
2. I don't believe it would add any additional clarity. In fact, I believe it might do the opposite and here's why: the method would need to operate in both directions (i.e. cloning a passed-in request object into the given `RequestMessageSnapshot` instance fields and cloning the given `RequestMessageSnapshot` instance into a new request object). I don't see any clear way to represent that behind a single method without it creating more confusion. It even seems like it would be a [code smell](https://en.wikipedia.org/wiki/Code_smell).

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: @martintmk ahh nevermind. I believe what you are suggesting here is the same as you are suggesting in [your other comment](https://github.com/dotnet/extensions/pull/5112#discussion_r1576731694):
> ... Can you still try to avoid the extra branches by condensing the code into a single `PrepareContentAsync` method
> It would also reduce the number of mutants you have to kill.

I'm resolving this comment to dedupe the threads. If I'm incorrect and this comment is proposing an entirely separate change, then please feel free to reactivate 👍

iliar-turdushev (approved):
The PR looks good to me.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
martintmk: > If the size of the StreamContent was in fact too large, it would have already crashed the process.

This is not true, the `StreamContent` can point to file stream and it consumes minimal memory even as you read the whole file. What you are doing here is materializing the whole stream into the memory. This is what concerns me.

martintmk:
@adamhammond

I have alternative proposal that should enable your scenarios and won't cause potential issues.

In your code-base you add the following handler:

``` csharp
public class BufferingHandler : DelegatingHandler
{
    {
        await request.Content.LoadIntoBufferAsync();
        return await base.SendAsync(request, cancellationToken);
    }
}
```

Now hedging will support `StreamContent` if the stream is seekable. It also gets the buffer and replicates it across hedged requests, so there are no multiple duplications. Something like:


``` csharp
        if (request.Content is StreamContent streamContent && await streamContent.ReadAsStreamAsync(cancellationToken) is Stream { CanSeek: true } stream)
        {
            byte[] buffer;

            if (stream is MemoryStream memoryStream)
            {
                // but buffer asside
                buffer = memoryStream.GetBuffer();
            }
            else
            {
                var memory = new MemoryStream();
                await stream.CopyToAsync(memory);
                buffer = memory.GetBuffer();
            }

            request.Content = new ByteArrayContent(buffer);
        }
```
This delegates the responsibility of buffering to the caller. In other words it's an explicit actions that tells us: "I have buffered the streamed content and I know what I am doing"

The only extra code on your part is that minimal `BufferingHandler` that you need to add before standard hedging.

WDYT?

martintmk:
Just to reiterate why I am pushing back on on this. The standard hedging is/will be used broadly across Microsoft most important services. My intuition tells me that allowing unlimited or default buffering of large streams into memory by default with eventually cause problems or outages.

So we need to do it opt-in with the option to limit the size that will be buffered.

adamhammond:
> Now hedging will support `StreamContent` if the stream is seekable. It also gets the buffer and replicates it across hedged requests, so there are no multiple duplications. Something like:
> 
> ```cs
>         if (request.Content is StreamContent streamContent && await streamContent.ReadAsStreamAsync(cancellationToken) is Stream { CanSeek: true } stream)
>         {
>             byte[] buffer;
> 
>             if (stream is MemoryStream memoryStream)
>             {
>                 // but buffer asside
>                 buffer = memoryStream.GetBuffer();
>             }
>             else
>             {
>                 var memory = new MemoryStream();
>                 await stream.CopyToAsync(memory);
>                 buffer = memory.GetBuffer();
>             }
> 
>             request.Content = new ByteArrayContent(buffer);
>         }
> ```
> 
> WDYT?

@martintmk I like your proposal in that it adds support for utilizing a buffer. I have a couple of questions/concerns:

1. What are you proposing happens if the given StreamContent is not seekable? Currently, I have the PR set to convert any non-seekable StreamContent into a seekable StreamContent by way of copying it to a new MemoryStream and using that to instantiate a new StreamContent. That is the only way to add hedging support for requests containing non-seekable StreamContent since the content must obviously be cloned in order to support request hedging. I feel an extension method like this that is provided by the framework itself should support doing this.
2. I see in your proposal that you are calling `memoryStream.GetBuffer` if the given stream is a `MemoryStream`; however, this isn't safe since you don't know if the given `MemoryStream` was instantiated with the [`publiclyVisible` flag set to `true`](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.-ctor?view=net-8.0#system-io-memorystream-ctor(system-byte()-system-int32-system-int32-system-boolean-system-boolean)), which enables the use of `GetBuffer`; otherwise, [an UnauthorizedAccessException will be thrown](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.getbuffer?view=net-8.0#exceptions) when you attempt to call the method. Unfortunately, there is no property or method that the `MemoryStream` class exposes to check whether or not use of the `GetBuffer` method is permitted on a given `MemoryStream` instance. So the only way to know would be to attempt to call it as you have done, but wrap the call in the try-catch block to catch the `UnauthorizedAccessException` and then execute some other flow to clone the content accordingly.

src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/RequestMessageSnapshot.cs
adamhammond: The only way what you are saying is true is if the content is a `FileStream` and is read into a buffer, correct? This is somewhat of a remarkable edge case since we are talking about streams in the context of HTTP request content. It would help me understand this better if you could provide an example scenario where a developer would want to do this.

That said, the idea of supporting buffered content is certainly not outside the realm of possibilities given that this will be an extension method exposed by the framework itself and we should do our best to cover all possible use cases. I still don't agree that we should enforce any kind of size limit; however, I do believe that we should add support for utilizing a buffer if the given request's `StreamContent` supports it. To this end, I responded to your other comment proposing support for utilizing a given `StreamContent`'s buffer if said `StreamContent` permits it to be read from.

martintmk:
> For hedging to work properly, the RequestMessageSnapshot must be a true snapshot.

I agree here, let's find a way to make this work opt-in. How about adding a `MaxRequestBodyBufferLength` to `HttpHedgingStrategyOptions` or similar if the name is too much. Set it to some sensible value, 5_000_000 for example or -1 for disabled. 

Then the snapshot can read this value a use it for buffering, throwing exception of buffer is exceeded.


Another alternative is to add `ConfigureRequestBodyBuffering` extension for `IStandardHedgingBuilder`. The letter being more appropriate IMHO.

src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs
```diff
@@ -88,26 +90,49 @@ public static IStandardHedgingHandlerBuilder AddStandardHedgingHandler(this IHtt
...
+
+                    try
+                    {
+                        var requestMessage = await snapshot.CreateRequestMessageAsync().ConfigureAwait(args.ActionContext.ContinueOnCapturedContext);
+
+                        // The secondary request message should use the action resilience context
+                        requestMessage.SetResilienceContext(args.ActionContext);
+
+                        // replace the request message
+                        args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);
```
martintmk, line 121:
```suggestion
                        args.ActionContext.Properties.Set(ResilienceKeys.RequestMessage, requestMessage);
```

There is a breaking change here :/

The `OnHedging` callback does not have access to request message anymore. This is because OnHedging is called after the action is created and before it is invoked.

https://github.com/App-vNext/Polly/blob/f85029c6d14ad20fd36e4fcdde7a32f33409137a/src/Polly.Core/Hedging/Controller/TaskExecution.cs#L127

RussKie:
Hi @adamhammond I'm closing this PR as it looks like this work has stalled. If/when you decide to resume, feel free to re-open.

PR #5167, opened 2024-05-20
Remove incorrect System.Net.Http dependency

### LINKED ISSUE #5164: System.Net.Http reference
In `R9` repository there are only `PackageReference Include="System.Net.Http"`
In `dotnet/extensions` repository there is only `Reference Include="System.Net.Http"`
 
I have tried to add `Microsoft.Extensions.Http.Resilience` package reference to internal R9 project and got compilation error in sample project (project consumes R9 nuget packages and demonstrates common use cases): 
```
error MSB3277: Found conflicts between different versions of "System.Net.Http" that could not be resolved.
error MSB3277: There was a conflict between "System.Net.Http, Version=4.1.1.3, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"
error MSB3277: "System.Net.Http, Version=4.1.1.3, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" was chosen because it was primary and "System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" was not..
```

After that I had replaced reference with obsolete resilience package from R9 (`..\Obsolete\HttpClient.Resilience\HttpClient.Resilience.csproj`) and everything worked as expected.
 
Questions:
1. What is intention behind different choices (`PackageReference` vs `Reference`) for `System.Net.Http`?
2. I believe sample project can be fixed with binding redirects or any other way, but it means every partner on net462 should perform the same change. Any idea how to add `Microsoft.Extensions.Http.Resilience` package reference without compilation error and any additional change on partners side?

RussKie:
Resolves #5164
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5167)

### DIFF
--- a/eng/packages/General.props
+++ b/eng/packages/General.props
@@ -51,7 +51,6 @@
     <PackageVersion Include="System.IO.Hashing" Version="$(SystemIOHashingVersion)" />
     <PackageVersion Include="System.Memory" Version="4.5.5" />
     <PackageVersion Include="System.Net.Http.Json" Version="$(SystemNetHttpJsonVersion)" />
-    <PackageVersion Include="System.Net.Http" Version="4.3.4" />
     <PackageVersion Include="System.Private.Uri" Version="4.3.2" />
     <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
     <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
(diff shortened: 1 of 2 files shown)


eng/packages/General.props
```diff
@@ -51,7 +51,6 @@
     <PackageVersion Include="System.IO.Hashing" Version="$(SystemIOHashingVersion)" />
     <PackageVersion Include="System.Memory" Version="4.5.5" />
     <PackageVersion Include="System.Net.Http.Json" Version="$(SystemNetHttpJsonVersion)" />
-    <PackageVersion Include="System.Net.Http" Version="4.3.4" />
```
joperezr, line 54: While I'm fine with the change, why did you singled out System.Net.Http here? For instance, while validating this change I noticed that System.Net.Http.Json is also present on this file and also not used apparently. One final thing to check, is I think that CPM will also affect transitive dependencies, so it would be good to double check and see if these aren't needed there.

xakep139 (approved):

geeknoid (approved):

eng/packages/General.props
RussKie: The issue is about System.Net.Http, so I only looked at System.Net.Http. I didn't look at anything else as it wasn't in the scope of this change.

joperezr: I see. Can we either cleanup the rest of unused ones here or if not then log an issue to cleanup to track the follow up?

RussKie: I'm going to send a separate PR for the cleanup.

PR #5169, opened 2024-05-20
FakeTimeProvider usage update

### LINKED ISSUE #4912: FakeTimeProvider.Advance fails to advance delays (at least in the case of Polly)
### Description

When using FakeTimerProvider in unit tests, retries configured in a ResiliencyPipeline are not executed.

The library maintainer believes the issue lies in FakeTimerProvider
https://github.com/App-vNext/Polly/issues/1932

### Reproduction Steps

Please see the minimal unit test repo
https://github.com/rcollette/PollyRetryFakeTimeProviderIssue

### Expected behavior

Delays used in Poly should be advanced.

### Actual behavior

They don't appear to be triggered.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

.NET 8.0.101
Polly 8.2.1

### Other information

I tried to debug down into the WakeWaiters/InvokeCallback/TimerCallback code but was not seeing why this was not triggering.

dariusclay:
Closes #4912 

Adds additional test and information regarding usage of FakeTimeProvider

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5169)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTests.cs
+++ b/test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTests.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Time.Testing;
-using Moq;
 using Xunit;
 
 namespace Microsoft.Extensions.Time.Testing.Test;
--- a/test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTests.cs
+++ b/test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTests.cs
@@ -393,24 +393,28 @@ public class FakeTimeProviderTests
         {
             async Task callback()
             {
-                tries++;
                 await provider.Delay(TimeSpan.FromSeconds(taskDelay));
-                if (tries < 2)
+
+                if (tries >= 2)
                 {
-                    throw new InvalidOperationException();
+                    return;
                 }
+
+                throw new InvalidOperationException();
             }
 
-            for (int attempt = 0; attempt < 3; attempt++)
+            for (int attempt = 1; attempt <= 3; attempt++)
             {
                 try
                 {
-                    await callback();
+                    tries = attempt;
+                    await callback().ConfigureAwait(false);
+                    break;
                 }
                 catch (InvalidOperationException)
                 {
                     // Trigger retry and sleep before new attempt
-                    await provider.Delay(TimeSpan.FromSeconds(1));
+                    await provider.Delay(TimeSpan.FromSeconds(1)).ConfigureAwait(false);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


dariusclay:
In the original thread #4912, the minimal repro overlooked the need to also advance due to the simulated task delay.

```cs
    [Fact]
    public void PollyRetry_WhenTaskDelayLessThanCancellationAndAboveRetryDelay_ShouldHave2Tries()
    {
        // Act
        var result = _someService.PollyRetry(taskDelay: 1, cancellationSeconds: 6);

        // Advance for simulated task delay
        _timeProvider.Advance(TimeSpan.FromMilliseconds(1000));

        // Advancing the time should resolve the first execution delay.
        _timeProvider.Advance(TimeSpan.FromMilliseconds(1000));

        // Advance for simulated task delay
        _timeProvider.Advance(TimeSpan.FromMilliseconds(1000));

        // Advancing the time should resolve the second execution delay.
        _timeProvider.Advance(TimeSpan.FromMilliseconds(1000));

        // Assert
        result.IsCompleted.Should().BeFalse();
        _someService.Tries.Should().Be(2);
    }
```

dariusclay:
The root issue here is with the way `Task.Delay(..., timeProvider)` works under the hood. In the case for non-system time providers, it calls `provider.CreateTimer()` to schedule a timer which FakeTimeProvider adds to it's list of Waiters. These get scheduled to run outside of the current context and may not all get a chance to be fired when Advance is called.

One thing of interest to note here, is that using `Delay(...).ConfigureAwait(true)` works in unit tests even without sleeping in the Advance method, as it ensures the continuation occurs on the same execution context. This would put the burden on any library or user code to know that it is using the fake provider instead of the TimeProvider.System instance.

geeknoid (approved):

xakep139 (commented):
LGTM

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTests.cs
```diff
@@ -381,4 +381,57 @@ public void ShouldResetGateUnderLock_PreventingContextSwitching_AffectionOnTimer
...
     }
+
+    [Fact]
+    public void SimulateRetryPolicy()
+    {
+        // Arrange
+        var retries = 42;
+        var tries = 0;
+        var taskDelay = 0.5;
+        var delay = 1;
```
RussKie, line 392: Please don't use `var` for intrinsic types.

test/Libraries/Microsoft.Extensions.TimeProvider.Testing.Tests/FakeTimeProviderTests.cs
```diff
@@ -381,4 +381,57 @@ public void ShouldResetGateUnderLock_PreventingContextSwitching_AffectionOnTimer
...
+                catch (InvalidOperationException)
+                {
+                    // ConfigureAwait(true) is required to ensure that tasks continue on the captured context
+                    await provider.Delay(TimeSpan.FromSeconds(delay)).ConfigureAwait(true);
+                }
+            }
+        }
+
+        // Act
+        var result = simulatedPollyRetry();
```
RussKie, line 422: Please don't use `var` unless the type is obvious.
https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md #10

Issue #5072, opened 2024-03-29
[API Proposal]: FakeTimeProvider, option to allow going backwards


jjxtra:
### Background and motivation

I'm using reflection to get around FakeTimeProvider inability to go backwards. Would be nice if a bool could be added to allow setting time earlier.

Thanks for considering.

### API Proposal

```csharp

public sealed class FakeTimeProvider : TimeProvider
{
    public bool AllowBackwardsTimeProgression { get; set; }
}
```

### API Usage

```csharp
// Fancy the value
FakeTimeProvider t = ...
t.AllowBackwardsTimeProgression = true;
t.SetUtcNow(new DateTimeOffset(1990, 1, 1, 1, 1, 1));
t.SetUtcNow(new DateTimeOffset(1984, 1, 1, 1, 1, 1));
```

### Risks

None I can think of. This new property will default to false.

geeknoid:
@stephentoub I presume it's possible for the normal time provider's idea of current time to move backwards? And this presumably doesn't affect any outstanding timers (since they are created with relative time), and presumably wouldn't affect the value returned by GetTimestamp?

jjxtra:
The internal property storing the time for FakeTimeProvider is called _now and is simply a DateTimeOffset value.

evgenyfedorov2:
@geeknoid -can this issue be closed as fixed by https://github.com/dotnet/extensions/pull/5192 ?

geeknoid:
@evgenyfedorov2 We need an API review to approve the new API, so we can remove the experimental attribute, and then we can close this issue.

amadeuszl:
I've sent a request to API Review Board for a review, so we can go out of experimental stage for that method.

bartonjs:
[Video](https://www.youtube.com/watch?v=eFQFAQnyLxc&t=1h4m56s)

* We feel the new property is unnecessary.  Rather, SetTimeUtc should just cease throwing the exception when time goes backwards.
  * When the new time is older, SetTimeUtc should defer to AdjustTime.
* Given no proposals exist to modify the existing AdjustTime method, we feel that it should cease being experimental (and, if there's disagreement, then the owners should come back to us).


```diff
namespace Microsoft.Extensions.Time.Testing;

public partial class FakeTimeProvider
{
-   [Experimental(diagnosticId: DiagnosticIds.Experiments.TimeProvider, UrlFormat = DiagnosticIds.UrlFormat)]
    public void AdjustTime(DateTimeOffset value)
[...]

PR #5192, opened 2024-05-31
Introduce FakeTimeProvider.AdjustTime

geeknoid:
This addresses #5072 by allowing time to be
adjusted forwards and backwards in order to simulate the system's clock being adjusted. Messing around
with the time in this way doesn't affect outstanding timers.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5192)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
+++ b/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


test/Generators/Microsoft.Gen.Logging/TestClasses/Bug5188.cs
dariusclay, line 1: Was this meant to be included or was it from working on #5191 ?

dariusclay (approved):

test/Generators/Microsoft.Gen.Logging/TestClasses/Bug5188.cs
geeknoid: Yep, makes sure we don't regress...

geeknoid: Oh, sorry, that file shouldn't be there.

Issue #5241, opened 2024-06-21
LinuxUtilizationProvider can't handle 'max' in /sys/fs/cgroup/memory.max


morti12:
### Description

We are using the newest OPC Publisher docker image which introduced in version 2.9.8 the newest version of the `.NET Extensions 8.6.0`. It actually should include a bug fix for #5161 from the version `8.5.0`.

On the startup we get the following exception because the `/sys/fs/cgroup/memory.max` contains `max` as value which can't be parsed as integer.

When checking it over the console, `cat /sys/fs/cgroup/memory.max` returns `max` from this path.

The default value of the `/sys/fs/cgroup/memory.max` can actually be max. Please have a look into the [documentation](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#memory-interface-files). 

```text
  memory.max
	A read-write single value file which exists on non-root
	cgroups.  The default is "max".
```

### Reproduction Steps

I tried it with different distributions but all of them have a different structure of the `cgroup` folder. Sry, I'm not a Linux expert...

So it's only not working on this RHEL machine. At least on every start of the docker container we get this message.

It's really hard to reproduce.

### Expected behavior

* The exception `System.InvalidOperationException: Could not parse '/sys/fs/cgroup/memory.max' content. Expected to find available memory in bytes but got 'max' instead.` should not be thrown.
* Max should be mapped correctly

### Actual behavior

Because of the `max` in the `memory.max` we get the following error:

```log
Unhandled exception. Autofac.Core.DependencyResolutionException: An exception was thrown while activating Azure.IIoT.OpcUa.Publisher.Services.PublisherDiagnosticCollector -> λ:Microsoft.Extensions.Diagnostics.ResourceMonitoring.IResourceMonitor -> Microsoft.Extensions.Diagnostics.ResourceMonitoring.ResourceMonitorService -> Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux.LinuxUtilizationProvider.
---> Autofac.Core.DependencyResolutionException: An exception was thrown while invoking the constructor 'Void .ctor(Microsoft.Extensions.Options.IOptions`1[Microsoft.Extensions.Diagnostics.ResourceMonitoring.ResourceMonitoringOptions], Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux.ILinuxUtilizationParser, System.Diagnostics.Metrics.IMeterFactory, System.TimeProvider)' on type 'LinuxUtilizationProvider'.
[...]

Issue #5161, opened 2024-05-16
System.InvalidOperationException : Could not parse '/sys/fs/cgroup/memory.max' content. Expected to find available memory in bytes but got 'max ' instead


tai-yi:
### Description

Run web API integration test(in container) error caused by `Microsoft.Extensions.Diagnostics.ResourceMonitoring`
```
System.InvalidOperationException : Could not parse '/sys/fs/cgroup/memory.max' content. Expected to find available memory in bytes but got 'max' instead
```

`Microsoft.Extensions.Diagnostics.ResourceMonitoring` 8.5.0
docker Image : mcr.microsoft.com/dotnet/sdk:8.0

### Reproduction Steps

run `dotnet test`

### Expected behavior

should not exception

### Actual behavior

```
Error Message:
   System.InvalidOperationException : Could not parse '/sys/fs/cgroup/memory.max' content. Expected to find available memory in bytes but got 'max
' instead.
  Stack Trace:
     at Microsoft.Shared.Diagnostics.Throw.InvalidOperationException(String message)
   at Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux.LinuxUtilizationParserCgroupV2.GetAvailableMemoryInBytes()
   at Microsoft.Extensions.Diagnostics.ResourceMonitoring.Linux.LinuxUtilizationProvider..ctor(IOptions`1 options, ILinuxUtilizationParser parser, IMeterFactory meterFactory, TimeProvider timeProvider)
   at InvokeStub_LinuxUtilizationProvider..ctor(Object, Span`1)
   at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
```

### Known Workarounds

Fallback to 8.4.0

### Configuration

Image : mcr.microsoft.com/dotnet/sdk:8.0
Host OS: Ubuntu 22.04.4 LTS
docker: 26.1.2

RussKie:
Duplicate of #5114

PR #5267, opened 2024-07-04
Fix resource detection for unset CPU and memory limits in cgroup v2

mobratil:
Address the issue where `/sys/fs/cgroup/cpu.max` and `/sys/fs/cgroup/memory.max` report "max", indicating unset limits in cgroup v2.

**Changes Made:**
- When `/sys/fs/cgroup/cpu.max` reads "max", the number of CPU cores is now derived from the host's `/proc/stat` file.
- Similarly, available memory is determined from the host's `/proc/meminfo` file if `/sys/fs/cgroup/memory.max` contains "max".

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5267)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParserCgroupV2.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationParserCgroupV2.cs
@@ -218,7 +218,7 @@ internal sealed class LinuxUtilizationParserCgroupV2 : ILinuxUtilizationParser
 
             ReadOnlySpan<char> memoryBuffer = bufferWriter.Buffer.WrittenSpan;
 
-            if (memoryBuffer.Equals("max", StringComparison.InvariantCulture))
+            if (memoryBuffer.Equals("max\n", StringComparison.InvariantCulture))
             {
                 return GetHostAvailableMemory();
             }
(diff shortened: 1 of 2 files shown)


gitmln:
Unfortunately, this pull request does not fix either of the two issues.
1. CPU limits
https://github.com/dotnet/extensions/issues/5241
_Could not parse '/sys/fs/cgroup/cpu.max'. Expected an integer but got: 'max 100000'._
` if (quotaBuffer.Equals("max", StringComparison.InvariantCulture))`
quotaBuffer in the above case contains "max 100000", not "max" string,
2. Memory limits
This one is more tricky.
https://github.com/dotnet/extensions/issues/5161
 _Could not parse '/sys/fs/cgroup/memory.max' content. Expected to find available memory in bytes but got 'max
' instead._
`if (memoryBuffer.Equals("max", StringComparison.InvariantCulture))`
 Again, you are comparing the wrong strings. The correct one is "max\n" (note the new line at the end!).

mobratil:
@gitmln thanks for pointing that out. The PR was just a draft because i didn't have a real installation of Linux with cgroup v2 on which I could test it.
Now I finally tested the PR with proper modifications.

evgenyfedorov2 (approved):

RussKie (approved):

RussKie:
@gitmln how does the current implementation look to you?

gitmln:
It looks perfect to me. This is the code that I've already thought about and tested successfully using a generic web app with mcr.microsoft.com/dotnet/runtime-deps:8.0 image on my AKS cluster.

mobratil:
@gitmln thanks for checking. Do not hesitate to create a PR next time 😊

RussKie:
Thank you @mobratil

PR #5305, opened 2024-07-23
Suppress IL2026 with attributes

martincostello:
Use attributes to suppress IL2026 so suppressions are visible to the linker.

I originally noticed this here when trying to root the assembly for the trimmer: https://github.com/dotnet/runtime/issues/105034#issuecomment-2233608643

I couldn't find any specific justification for the four suppressions. Two came from #4410 and the rest were present when r9 was originally imported to the repo. I assume it's related to the validation source generator? I can fill in the justification as appropriate.

If these shouldn't be suppressed, then instead the properties should instead pass along some variant of the original warning on `[Length]` by adding `[RequiresUnreferencedCode]` (which is `[RequiresUnreferencedCode("Uses reflection to get the 'Count' property on types that don't implement ICollection. This 'Count' property may be trimmed. Ensure it is preserved.")]`).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5305)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Routing/OrderedGroupsRoutingOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Routing/OrderedGroupsRoutingOptions.cs
@@ -19,15 +19,14 @@ public class OrderedGroupsRoutingOptions
     /// Gets or sets the collection of ordered endpoints groups.
     /// </summary>
 #pragma warning disable CA2227 // Collection properties should be read only
-#pragma warning disable IL2026 // Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code
     [Required]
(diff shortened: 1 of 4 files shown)


MichalStrehovsky:
I think this should stay as a pragma. If I understand this correctly, the warning comes from the attribute LengthAttribute itself - creating an instance of the attribute at runtime is not safe. Accessing the property annotated with it is completely fine.

Simply accessing the property won't generate warnings, since accessing a property is not something that deals with attributes, it's just an accessor method call. Reflecting on the property is a problem because it allows activating the problematic attribute. You see a new warning when reflection-rooting the whole assembly because it means "please consider the property reflection accessed". This is by design, it should warn, because the attribute is not trim safe. Using an attribute would remove that warning, and actually activating the attribute would result in a broken app without publish-time warnings; that's a bug.

PR #5309, opened 2024-07-25
Rename metrics for Resource Monitoring

### LINKED ISSUE #5113: Incorrect metric names in ResourceUtilizationInstruments
Metric names in ResourceUtilizationInstruments have incorrect keys https://github.com/dotnet/extensions/blob/10681a1cdb1e044b05341150203b94d5eec41557/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationInstruments.cs

1. `public const string CpuUtilization = "process.cpu.utilization";` - in reality, the CPU utilization is measured for all processes in a container, not only for current.
3. `public const string MemoryUtilization = "dotnet.process.memory.virtual.utilization";`- in reality, the Memory utilization is measured for all processes in a container, not only for current dotnet process.

evgenyfedorov2:
Fixes #5113 

1. Renamed metrics
2. Introduced more specific metrics for container related measurements to distinguish process vs. container (cgroup).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5309)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationInstruments.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationInstruments.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
+
 namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
 
 /// <summary>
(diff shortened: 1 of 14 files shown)


RussKie (changes requested):
This change looks like it potentially may be a breaking change. Metrics are kind part of the public surface - i.e., documentations, dashboards and other monitoring systems depend on existing/published metrics.

Unfortunately the original issue doesn't provide great level of detailed information on the underlying issue(s), scenarios where and how it manifests, how it affects customers, etc.

Could you please provide more information and assessment of the change in the following form:
- Customer Imparct - how would this change affect existing customers?
- Is this a regression? If yes, when was it introduced?
- How risky is this change?

Thank you.

I'm blocking the merge to understand the risks and impacts and stop an accidental merge.

evgenyfedorov2:
@RussKie, here you are:
- Customer Impact - Monitoring dashboards will break, it requires customers to fix the dashboards.
- Is this a regression? - No
- How risky is this change? - Not sure how many customers we have for those two `ObservableGauge`metrics (I understand those are pretty new and not widely adopted), but I suggest making the change ASAP to avoid onboarding even more customers to incorrect metric names.

evgenyfedorov2:
> Unfortunately the original issue doesn't provide great level of detailed information on the underlying issue(s), scenarios where and how it manifests, how it affects customers, etc.

It does not affect customers as such, they can live with it forever and be happy :) However, it is just semantically incorrect to have `process.*` names for metrics measuring CPU/Memory utilization of the whole container or control group.

RussKie:
@JamesNK @joperezr any thoughts on this?

JamesNK:
It's definitely a breaking change. It needs to be documented as a breaking change and the mitigations required to work around it be in release notes.

You could require people to update whatever consumes the metrics to make a fix. If there are people who can't do that then you could add an app context switch that someone can set in their app to keep old behavior.

joperezr:
@evgenyfedorov2 can you please go into a bit more detail as to what is the motivation behind this change? Is it mostly just that the naming is inaccurate so we want to fix it? We should get the full motivation so we can evaluate if this is worth the breaking change.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceUtilizationInstruments.cs
```diff
@@ -13,18 +15,50 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
...
+    /// </remarks>
+    public const string ProcessCpuUtilization = "process.cpu.utilization";
+
+    /// <summary>
+    /// Gets the memory consumption share of the running process in range <c>[0, 1]</c>.
+    /// </summary>
+    /// <remarks>
+    /// The type of an instrument is <see cref="System.Diagnostics.Metrics.ObservableGauge{T}"/>.
+    /// </remarks>
+    public const string ProcessMemoryUtilization = "dotnet.process.memory.virtual.utilization";
```
RussKie, line 31: Should we decorate these as `[Obsolete]` to indicate they shouldn't be used?
image

PR #5352, opened 2024-08-12
Logging generator: Handle multiple type definitions

### LINKED ISSUE #5173: `Microsoft.Extensions.Http.Resilience` breaks `LoggerMessage` code generation.
### Description

We have a .NET 8 project where we add a reference to a package that includes a dependency on `Microsoft.Extensions.Http.Resilience`. That project also uses `[LoggerMessage]` to take advantage of the source generator.

That project also includes a reference to a package (.NET Standard 2.0) that is using [il-repack,](https://github.com/gluck/il-repack) which results in all the types under `Microsoft.Extensions.Logging` (including `Microsoft.Extensions.Logging.LoggerMessageAttribute`) being included in the assembly as internal types.

In this scenario, the source generator does not work.

```
[LoggerMessage(Level = LogLevel.Information, Message = "Something")]
private static partial void LogSomething(ILogger logger);
```

The error is:

```
CS8795	Partial method 'Program.LogSomething(ILogger)' must have an implementation part because it has accessibility modifiers.
```

Investigating this, I believe that the problem is related with how `Microsoft.Gen.Logging.LoggingGenerator.LoggingGenerator` is looking up for `Microsoft.Extensions.Logging.LoggerMessageAttribute`.

I think that `Microsoft.Extensions.Logging.Generators.LoggerMessageGenerator.Parser` did that lookup considering the visibility of the types:

```
INamedTypeSymbol bestTypeByMetadataName = _compilation.GetBestTypeByMetadataName("Microsoft.Extensions.Logging.LoggerMessageAttribute");
```

I'm not 100% sure, but it looks like `Microsoft.Gen.Logging.LoggingGenerator.LoggingGenerator` isn't considering visibility. Which results in the code not being generated.

### Reproduction Steps

Here's the csproj definition:

```

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <Nullable>enable</Nullable>
    </PropertyGroup>

    <ItemGroup>
        <PackageReference Include="Cegid.Hydrogen.Hosting" Version="8.0.5" />
        <PackageReference Include="Cegid.Lithium.Settings.Client.Rest.SingleAssembly" Version="6.0.0" />
    </ItemGroup>

</Project>
```

`Cegid.Hydrogen.Hosting` is the package that depends on `Microsoft.Extensions.Http.Resilience`.
`Cegid.Lithium.Settings.Client.Rest.SingleAssembly` is the package that uses il-repack and has `LoggerMessageAttribute` as an internal type.

capture

### Expected behavior

The source generator should work as expected, because there is only one public type accessible named `Microsoft.Extensions.Logging.LoggerMessageAttribute`.

### Actual behavior

The source generator does not kick-in.

### Regression?

No.

### Known Workarounds

Adding this in the csproj, after the package reference, makes the old source generator work:

```
<PropertyGroup>
   <DisableMicrosoftExtensionsLoggingSourceGenerator>false</DisableMicrosoftExtensionsLoggingSourceGenerator>
</PropertyGroup>
```

Obviously, this does not work if we remove the reference to the single assembly, as BOTH source generators kick-in and we end up with 2 implementations of the logging method.

### Configuration

We're using .NET 8.0.4 and Microsoft.Extensions.Http.Resilience 8.4.0.
Nothing else seams relevant for this issue.

### Other information

_No response_

dariusclay:
Fixes #5173 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5352)


src/Generators/Microsoft.Gen.Logging/Parsing/SymbolLoader.cs
```diff
@@ -52,16 +53,16 @@ internal static class SymbolLoader
...
-        var logLevelSymbol = compilation.GetTypeByMetadataName(LogLevelType);
-        var loggerMessageAttributeSymbol = compilation.GetTypeByMetadataName(LoggerMessageAttribute);
-        var logPropertiesAttributeSymbol = compilation.GetTypeByMetadataName(LogPropertiesAttribute);
-        var tagProviderAttributeSymbol = compilation.GetTypeByMetadataName(TagProviderAttribute);
-        var tagNameAttributeSymbol = compilation.GetTypeByMetadataName(TagNameAttribute);
-        var tagCollectorSymbol = compilation.GetTypeByMetadataName(ITagCollectorType);
-        var logPropertyIgnoreAttributeSymbol = compilation.GetTypeByMetadataName(LogPropertyIgnoreAttribute);
-        var dataClassificationAttribute = compilation.GetTypeByMetadataName(DataClassificationAttribute);
-        var noDataClassificationAttribute = compilation.GetTypeByMetadataName(NoDataClassificationAttribute);
+        var loggerSymbol = compilation.GetBestTypeByMetadataName(ILoggerType);
```
dariusclay, line 56: This aligns with the symbol lookup behavior from dotnet runtime logging generator.

src/Generators/Shared/RoslynExtensions.cs
```diff
@@ -0,0 +1,203 @@
+using System;
```
RussKie, line 1: Missing header

src/Generators/Shared/RoslynExtensions.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+#pragma warning disable CA1716
```
RussKie, line 6: What's this suppression?

src/Generators/Shared/RoslynExtensions.cs
```diff
@@ -0,0 +1,203 @@
...
+namespace Microsoft.Gen.Shared;
+#pragma warning restore CA1716
+
+internal static class RoslynExtensions
+{
+    // Copied from: https://github.com/dotnet/roslyn/blob/main/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/CompilationExtensions.cs
```
RussKie, line 12: I recommend using permalinks because there's no guarantee the file or its content will always exist in the main
```suggestion
    // Copied from: https://github.com/dotnet/roslyn/blob/dd13c33ad54aa8afed4442b029bb14cb5893bc7e/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/CompilationExtensions.cs
```

src/Generators/Shared/RoslynExtensions.cs
```diff
@@ -0,0 +1,203 @@
...
+    public static SimpleNameSyntax GetUnqualifiedName(this NameSyntax name)
+        => name switch
+        {
+            AliasQualifiedNameSyntax alias => alias.Name,
+            QualifiedNameSyntax qualified => qualified.Right,
+            SimpleNameSyntax simple => simple,
+            _ => throw new InvalidOperationException("Unreachable"),
+        };
+
+    // copied from https://github.com/dotnet/roslyn/blob/main/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/ISymbolExtensions.cs
```
RussKie, line 137:
```suggestion
    // Copied from https://github.com/dotnet/roslyn/blob/dd13c33ad54aa8afed4442b029bb14cb5893bc7e/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/ISymbolExtensions.cs
```

src/Generators/Shared/RoslynExtensions.cs
```diff
@@ -0,0 +1,203 @@
...
+                    // symbol stays at the level we've gotten so far.
+            }
+
+            symbol = symbol.ContainingSymbol;
+        }
+
+        return visibility;
+    }
+
+    // Copied from: https://github.com/dotnet/roslyn/blob/main/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Utilities/SymbolVisibility.cs
```
RussKie, line 185: ditto

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
```diff
@@ -966,6 +968,56 @@ static partial class C
...
+        Compilation compilation = CompilationHelper.CreateCompilation(source, additionalReferences);
+        LoggingGenerator generator = new LoggingGenerator();
+
+        (IReadOnlyList<Diagnostic> diagnostics, ImmutableArray<GeneratedSourceResult> generatedSources) =
+            RoslynTestUtils.RunGenerator(compilation, generator);
+
+        // Make sure compilation was successful.
+        Assert.Empty(diagnostics);
+        Assert.Single(generatedSources);
+        Assert.Equal(40, generatedSources[0].SourceText.Lines.Count);
```
RussKie, line 1018: For blob assertions we should use VerifyXunit
https://github.com/dotnet/extensions/blob/0cd4e2e66a3b453c15943c3f9c242545775f706c/test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/LinuxUtilizationParserCgroupV2Tests.cs#L164

src/Generators/Shared/RoslynExtensions.cs
dariusclay: Identifiers should not match keywords. Added for consistency, as it's also found in other types in the M.Gen.Shared namespace.

src/Generators/Shared/RoslynExtensions.cs
dariusclay: Thanks! Added

src/Generators/Shared/RoslynExtensions.cs
dariusclay: changed all to use permalinks

iliar-turdushev (approved):

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
dariusclay: Thanks for the tip. Had to include a scrubber for the assembly version.

test/Generators/Microsoft.Gen.Logging/Unit/CompilationHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+        refs.Add(MetadataReference.CreateFromFile(Path.Combine(runtimeDir, "netstandard.dll")));
+        refs.Add(MetadataReference.CreateFromFile(Path.Combine(runtimeDir, "System.Runtime.dll")));
+        refs.Add(MetadataReference.CreateFromFile(typeof(ILogger).Assembly.Location));
+        refs.Add(MetadataReference.CreateFromFile(typeof(LoggerMessageAttribute).Assembly.Location));
+        refs.Add(MetadataReference.CreateFromFile(typeof(IEnrichmentTagCollector).Assembly.Location));
+        refs.Add(MetadataReference.CreateFromFile(typeof(DataClassification).Assembly.Location));
+        refs.Add(MetadataReference.CreateFromFile(typeof(PrivateDataAttribute).Assembly.Location));
+        refs.Add(MetadataReference.CreateFromFile(typeof(BigInteger).Assembly.Location));
+
+        if (additionalReferences != null)
```
RussKie, line 39:
```suggestion
        if (additionalReferences is not null)
```

src/Generators/Shared/RoslynExtensions.cs
```diff
@@ -0,0 +1,208 @@
...
+            case SymbolKind.Parameter:
+                // Parameters are only as visible as their containing symbol
+                return GetResultantVisibility(symbol.ContainingSymbol);
+
+            case SymbolKind.TypeParameter:
+                // Type Parameters are private.
+                return SymbolVisibility.Private;
+        }
+
+        while (symbol != null && symbol.Kind != SymbolKind.Namespace)
```
RussKie, line 164:
```suggestion
        while (symbol is not null && symbol.Kind != SymbolKind.Namespace)
```

PR #5363, opened 2024-08-14
Disable HttpClient's timeout for Standard Resilience and Hedging

### LINKED ISSUE #4770: HttpClient.Timeout overrides the TotalRequestTimeout when using AddStandardResilienceHandler
### Description

The `HttpClient.Timeout` interferes with the `TotalRequestTimeout` when using HttpClient with standard pipeline. If `HttpClient.Timeout` is less than `TotalRequestTimeout`, then the `TotalRequestTimeout` is ignored.

This causes mis-alignment and unexpected behavior for anyone setting the `TotalRequestTimeout` to a value greater than 100 seconds, which is the default for `HttpClient.Timeout`.

Calling `AddStandardResilienceHandler` should disable the `HttpClient.Timeout` by setting the value to `Timeout.InfiniteTimeSpan`, because the overall timeout is already enforced by standard pipeline. This also saves some allocations due to HttpClient not needing to create `CancellationTokenSource`. 

### Reproduction Steps

Execute this code:

``` csharp
var client = new ServiceCollection()
    .AddHttpClient("my-client", c => c.BaseAddress = new Uri("https://jsonplaceholder.typicode.com"))
    .AddStandardResilienceHandler()
    .Configure(options =>
    {
        options.AttemptTimeout.Timeout = TimeSpan.FromSeconds(1);
        options.TotalRequestTimeout.Timeout = TimeSpan.FromSeconds(150);
        options.Retry.ShouldHandle = _ => new ValueTask<bool>(watch.Elapsed < TimeSpan.FromSeconds(100));
        options.Retry.MaxRetryAttempts = int.MaxValue;
    })
    .Services.BuildServiceProvider()
    .GetRequiredService<IHttpClientFactory>()
    .CreateClient("my-client");

await client.GetStringAsync("/", default);
```

This should be successful. Instead, the `TaskCancelledException` is thrown.

### Expected behavior

Above example should not throw exception.

### Actual behavior

Above example throws `TaskCancelledException`.

### Regression?

_No response_

### Known Workarounds

Manually disable HttpClient timeout:

``` csharp
var client = new ServiceCollection()
    .AddHttpClient("my-client", c =>
    {
        c.BaseAddress = new Uri("https://jsonplaceholder.typicode.com");
        c.Timeout = Timeout.InfiniteTimeSpan;
    })
    .AddStandardResilienceHandler();
```

### Configuration

.NET SDK:
 Version:           8.0.100
 Commit:            57efcf1350
 Workload version:  8.0.100-manifests.6a1e483a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.100\

### Other information

_No response_

### LINKED ISSUE #4924: The ConfigureHttpClient method is not supported when creating gRPC clients.
### Description

https://github.com/dotnet/extensions/commit/3654748e14445fa794bcdb0d2ebffa46e758603c is breaking using gRPC clients with Http.Resilience. This is a breaking change from 8.1.

### Reproduction Steps

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.2.0-preview.24105.1" />

    <PackageReference Include="Google.Protobuf" Version="3.25.2" />
    <PackageReference Include="Grpc.Net.ClientFactory" Version="2.59.0" />
    <PackageReference Include="Grpc.Tools" Version="2.59.0">
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
  </ItemGroup>

  <ItemGroup>
    <Protobuf Include="Protos\greet.proto" GrpcServices="Client" />
  </ItemGroup>
  
</Project>
```

```C#
using GrpcGreeter;

var builder = WebApplication.CreateBuilder(args);

// Add services to the container.

builder.Services.AddGrpcClient<Greeter.GreeterClient>(o =>
{
    o.Address = new Uri("https://localhost:5001");
});

builder.Services.ConfigureHttpClientDefaults(http =>
{
    // Turn on resilience by default
    http.AddStandardResilienceHandler();
});

var app = builder.Build();

// Configure the HTTP request pipeline.

var summaries = new[]
{
    "Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"
};

app.MapGet("/hello", (Greeter.GreeterClient client) =>
{
    return "hello";
});

app.Run();
```

greet.proto:
```proto
syntax = "proto3";

option csharp_namespace = "GrpcGreeter";

package greet;

// The greeting service definition.
service Greeter {
  // Sends a greeting
  rpc SayHello (HelloRequest) returns (HelloReply);
}

// The request message containing the user's name.
message HelloRequest {
  string name = 1;
}

// The response message containing the greetings.
message HelloReply {
  string message = 1;
}
```

### Expected behavior

The app should work, like it did when I use `<PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.2.0-preview.24105.1" />`

### Actual behavior

The app fails with:

```
An unhandled exception occurred while processing the request.
InvalidOperationException: The ConfigureHttpClient method is not supported when creating gRPC clients. Unable to create client with name 'GreeterClient'.
Grpc.Net.ClientFactory.Internal.GrpcCallInvokerFactory.CreateInvoker(EntryKey key)

Stack Query Cookies Headers Routing
InvalidOperationException: The ConfigureHttpClient method is not supported when creating gRPC clients. Unable to create client with name 'GreeterClient'.
Grpc.Net.ClientFactory.Internal.GrpcCallInvokerFactory.CreateInvoker(EntryKey key)
System.Collections.Concurrent.ConcurrentDictionary<TKey, TValue>.GetOrAdd(TKey key, Func<TKey, TValue> valueFactory)
Grpc.Net.ClientFactory.Internal.GrpcCallInvokerFactory.CreateInvoker(string name, Type type)
Grpc.Net.ClientFactory.Internal.DefaultGrpcClientFactory.CreateClient<TClient>(string name)
Microsoft.Extensions.DependencyInjection.GrpcClientServiceExtensions+<>c__DisplayClass7_0<TClient>.<AddGrpcHttpClient>b__2(IServiceProvider s)
Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitDisposeCache(ServiceCallSite transientCallSite, RuntimeResolverContext context)
Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor<TArgument, TResult>.VisitCallSite(ServiceCallSite callSite, TArgument argument)
Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
Microsoft.Extensions.DependencyInjection.ServiceLookup.DynamicServiceProviderEngine+<>c__DisplayClass2_0.<RealizeService>b__0(ServiceProviderEngineScope scope)
Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(ServiceIdentifier serviceIdentifier, ServiceProviderEngineScope serviceProviderEngineScope)
Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
lambda_method2(Closure , object , HttpContext )
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddlewareImpl.Invoke(HttpContext context)
```

### Regression?

Yes, from 8.1

### Known Workarounds

Don't call `AddStandardResilienceHandler()` on HttpClients that use gRPC.

### Configuration

net8.0

### Other information

cc @martintmk @geeknoid @joperezr @JamesNK

iliar-turdushev:
Resolves #4770
Resolves #4924

Reintroduces changes made in the following PR #4862. Originally that PR was reverted because of the following bug #4924.
Adds a note to the documentation mentioning the minimal grpc-dotnet version to be used with M.E.Http.Resilience package.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5363)


src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -4,6 +4,11 @@ Resilience mechanisms for `HttpClient` built on the [Polly framework](https://ww
 
 ## Install the package
 
+> [!IMPORTANT]
```
iliar-turdushev, line 7: @RussKie This is a note helping our users to avoid issues when using `grpc-dotnet + Http.Resilience`.
Is it good to put here a link to the original issue we want to avoid? How do you think? Thank you.

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -4,6 +4,11 @@ Resilience mechanisms for `HttpClient` built on the [Polly framework](https://ww
...
+> If you're using the `Grpc.Net.ClientFactory` package then update it to the version `>=2.64.0`
+> before installing the `Microsoft.Extensions.Http.Resilience` package. This will help you to
+> avoid issues like the [following](https://github.com/dotnet/extensions/issues/4924).
```
RussKie, line 10: Having a second thought now... What do you think of adding a "Known Issues" section before "Feedback & Contributing" section?

And in there say something along the lines of:
````
Ensure `Grpc.Net.ClientFactory` v2.64.0 or later is installed, if you receive the following exception:
```
InvalidOperationException: The ConfigureHttpClient method is not supported when creating gRPC clients.
```
````

I don't know if there's any good to link to #4924. What would a developer gain from that thread?

Comment (RussKie) [reply src/Libraries/Microsoft.Extensions.Http.Resilience/README.md:10]: @gewarren would you have any suggestions here? We'll need to add a similar note to learn.ms as well.
Thanks

Comment (iliar-turdushev) [reply src/Libraries/Microsoft.Extensions.Http.Resilience/README.md:10]: > Having a second thought now... What do you think of adding a "Known Issues" section before "Feedback & Contributing" section?

I think that is better what we currently have. Having "Known Issues" section explaining why users need to update `Grpc.Net.ClientFactory` makes perfect sense.

Comment (IEvangelist) [reply src/Libraries/Microsoft.Extensions.Http.Resilience/README.md:10]: Sure, if it's a _known issue_ - call it out as such. I support that idea!

Comment (joperezr) [reply src/Libraries/Microsoft.Extensions.Http.Resilience/README.md:10]: I wonder if we should just add some targets file in the package that checks for this before  build and if this is the case we log a warning. If we think this is an important thing to call out, then doing that will bring this error to devs faces as opposed to keeping it on docs only.

Comment (RussKie) [reply src/Libraries/Microsoft.Extensions.Http.Resilience/README.md:10]: @joperezr it's an interesting idea.

Comment (RussKie) [reply src/Libraries/Microsoft.Extensions.Http.Resilience/README.md:10]: @iliar-turdushev the target would look something like this:
```xml
  <!-- 
    Check whether the project is referencing Grpc.Net.ClientFactory/2.64.0 or older.
    If the vesion is newer warn the user to update.
    -->
  <Target Name="_CheckGrpcNetClientFactoryVersion"
          AfterTargets="ResolveReferences"
          Condition=" '$(SuppressCheckGrpcNetClientFactoryVersion)' != 'true' ">
    <ItemGroup>
      <_GrpcNetClientFactory Include="@(PackageReference)"
                             Condition=" '%(PackageReference.Identity)' == 'Grpc.Net.ClientFactory' AND $([MSBuild]::VersionLessThan('%(PackageReference.Version)', '2.64.0')) " />
    </ItemGroup>

    <Error Condition="@(_GrpcNetClientFactory->Count()) &gt; 0"
           Text="Grpc.Net.ClientFactory v2.64.0 or later is required (see https://github.com/dotnet/extensions/issues/4924 for more details). The check can be suppressed with SuppressCheckGrpcNetClientFactoryVersion=true." />
  </Target>
```

I performed some light tests on the repro provided in #4924. Please test other scenarios (including a project using [CPM](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management)).

iliar-turdushev: @RussKie Thank you very much for help. I extended your code sample a little bit and added the check. Please, review it and share your feedback. Also I added "Known issues" section describing the issue and how to fix or suppress it.

src/Libraries/Microsoft.Extensions.Http.Resilience/buildTransitive/Microsoft.Extensions.Http.Resilience.props
```diff
@@ -0,0 +1,8 @@
+﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <SuppressCheckGrpcNetClientFactoryVersion>false</SuppressCheckGrpcNetClientFactoryVersion>
+    <_GrpcNetClientFactory>Grpc.Net.ClientFactory</_GrpcNetClientFactory>
+    <_CompatibleGrpcNetClientFactoryVersion>2.64.0</_CompatibleGrpcNetClientFactoryVersion>
+    <_GrpcNetClientFactoryVersionIsIncorrect>Grpc.Net.ClientFactory 2.63.0 or earlier could cause issues when used together with Microsoft.Extensions.Http.Resilience. For more details, see https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience#known-issues. Consider using Grpc.Net.ClientFactory $(_CompatibleGrpcNetClientFactoryVersion) or later. To suppress the error set SuppressCheckGrpcNetClientFactoryVersion=true.</_GrpcNetClientFactoryVersionIsIncorrect>
```
iliar-turdushev, line 6: Alternatively, we can add a link to the learn.microsoft.com: https://learn.microsoft.com/en-us/dotnet/core/resilience/http-resilience. @RussKie @joperezr How do you think what would work better? Thanks.

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -73,6 +73,37 @@ clientBuilder.AddResilienceHandler("myHandler", b =>
 });
 ```
 
+## Known issues
```
iliar-turdushev, line 76: @IEvangelist Could you, please, review this section? How do you think if I can add it as-is to the following page https://learn.microsoft.com/en-us/dotnet/core/resilience/http-resilience? If not, what should I adjust? Thank you.

src/Libraries/Microsoft.Extensions.Http.Resilience/buildTransitive/Microsoft.Extensions.Http.Resilience.props
RussKie: I would point to learn.microsoft.com.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -38,4 +38,73 @@
     <InternalsVisibleToTest Include="$(AssemblyName).Tests" />
     <InternalsVisibleToDynamicProxyGenAssembly2 Include="*" />
   </ItemGroup>
+
+  <ItemGroup>
```
RussKie, line 42: Our implementation of rewriting targets files for net462 makes us duplicate the content, and in this case it's a significant maintenance overhead.

I think we can be a little smarter and preserve the existing targets, though just rename it during the packaging.
This should work:
```diff
diff --git a/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj b/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
index f7b7fa6d85..0861d4261c 100644
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
@@ -40,10 +40,17 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="buildTransitive\$(MSBuildProjectName).*"
+    <None Include="buildTransitive\**\*"
           CopyToOutputDirectory="PreserveNewest"
           Pack="true"
           PackagePath="buildTransitive\$(MinimumSupportedTfmForPackaging)\" />
+    <None Include="buildTransitive\Microsoft.Extensions.Http.Resilience.targets"
+          Condition="'$(MinimumSupportedTfmForPackaging)' != '' and
+                     '$(ConditionalNet462)' != '' and
+                     ($(TargetFrameworks.Contains('$(MinimumSupportedTfmForPackaging)')) and $(TargetFrameworks.Contains('$(ConditionalNet462)')))"
+          CopyToOutputDirectory="PreserveNewest"
+          Pack="true"
+          PackagePath="buildTransitive\net462\Microsoft.Extensions.Http.Resilience.net462.targets" />
     
     <!-- For net462 we copy only .props file. The .targets file is added automatically by the build process, and the code
          that checks the version of the Grpc.Net.ClientFactory package will be added to the .targets file as a value
@@ -59,51 +66,7 @@
   <PropertyGroup>
     <_AdditionalNETStandardCompatErrorFileContents>
       <![CDATA[
-  <!--
-    Check whether the project is referencing Grpc.Net.ClientFactory 2.64.0 or later.
-    If the vesion is earlier warn the user to update.
-    See https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience#known-issues for more details.
-    -->
-  <Target Name="_CheckGrpcNetClientFactoryVersion"
-          BeforeTargets="ResolveReferences"
-          Condition=" '%24(SuppressCheckGrpcNetClientFactoryVersion)' != 'true' ">
-    <ItemGroup>
-      <!-- Find the package in the .csproj file. -->
-      <_GrpcNetClientFactoryPackageReference Include="%40(PackageReference)" Condition=" '%(PackageReference.Identity)' == '%24(_GrpcNetClientFactory)' " />
-
-      <!-- Find the version of the package in the Central Package Source. The solution uses the Central Package Management. -->
-      <_GrpcNetClientFactoryPackageVersion Include="%40(PackageVersion)" Condition=" '%(PackageVersion.Identity)' == '%24(_GrpcNetClientFactory)' " />
-      
-      <!-- The package is added to the project as a transitive dependency. -->
-      <_GrpcNetClientFactoryTransitiveDependency Include="%40(ReferencePath)" Condition=" '%(ReferencePath.NuGetPackageId)' == '%24(_GrpcNetClientFactory)' " />
-    </ItemGroup>
-
-    <!-- The version of the package is included in the .csproj file. -->
-    <Error Condition=" %40(_GrpcNetClientFactoryPackageReference->Count()) &gt%3B 0
-                       AND '%(_GrpcNetClientFactoryPackageReference.Version)' != ''
-                       AND %24([MSBuild]::VersionLessThan('%(_GrpcNetClientFactoryPackageReference.Version)', '%24(_CompatibleGrpcNetClientFactoryVersion)')) "
-           Text="%24(_GrpcNetClientFactoryVersionIsIncorrect)" />
-
-    <!-- The solution uses the Central Package Management and the version of the package is overridden in the .csproj file using the VersionOverride property. -->
-    <Error Condition=" '%24(ManagePackageVersionsCentrally)' == 'true'
-                       AND %40(_GrpcNetClientFactoryPackageReference->Count()) &gt%3B 0
-                       AND '%(_GrpcNetClientFactoryPackageReference.VersionOverride)' != ''
-                       AND %24([MSBuild]::VersionLessThan('%(_GrpcNetClientFactoryPackageReference.VersionOverride)', '%24(_CompatibleGrpcNetClientFactoryVersion)')) "
-           Text="%24(_GrpcNetClientFactoryVersionIsIncorrect)" />
-
-    <!-- The solution uses the Central Package Management and the version of the package is included in the Central Package Source. -->
-    <Error Condition=" '%24(ManagePackageVersionsCentrally)' == 'true'
-                       AND %40(_GrpcNetClientFactoryPackageReference->Count()) &gt%3B 0
-                       AND '%(_GrpcNetClientFactoryPackageVersion.Version)' != ''
-                       AND %24([MSBuild]::VersionLessThan('%(_GrpcNetClientFactoryPackageVersion.Version)', '%24(_CompatibleGrpcNetClientFactoryVersion)')) "
-           Text="%24(_GrpcNetClientFactoryVersionIsIncorrect)" />
-
-    <!-- This condition handles a case when the package is added to the project as a transitive dependency. -->
-    <Error Condition=" %40(_GrpcNetClientFactoryTransitiveDependency->Count()) &gt%3B 0
-                       AND '%(_GrpcNetClientFactoryTransitiveDependency.NuGetPackageVersion)' != ''
-                       AND %24([MSBuild]::VersionLessThan('%(_GrpcNetClientFactoryTransitiveDependency.NuGetPackageVersion)', '%24(_CompatibleGrpcNetClientFactoryVersion)')) "
-           Text="%24(_GrpcNetClientFactoryVersionIsIncorrect)" />
-  </Target>
+  <Import Project="%24(MSBuildThisFileDirectory)\Microsoft.Extensions.Http.Resilience.net462.targets" />
 ]]>
     </_AdditionalNETStandardCompatErrorFileContents>
   </PropertyGroup>
```

image


Though, please test it.

RussKie (commented):
👍

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -40,70 +40,26 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="buildTransitive\$(MSBuildProjectName).*"
+    <None Include="buildTransitive\$(MSBuildProjectName).targets"
```
RussKie, line 43: This doesn't look correct. What about the props?

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -38,4 +38,29 @@
...
+          Pack="true"
+          PackagePath="buildTransitive\$(MinimumSupportedTfmForPackaging)\" />
+    
+    <!-- For net462 we have to rename the .targets file, because our build process automatically adds a .targets file
+         with the same name as the project file. Then we'll import the renamed .targets file in the automatically
+         added .targets file. For that we need to set the _AdditionalNETStandardCompatErrorFileContents variable. -->
+    <None Include="buildTransitive\$(MSBuildProjectName).targets"
+          CopyToOutputDirectory="PreserveNewest"
+          Pack="true"
+          PackagePath="buildTransitive\net462\Microsoft.Extensions.Http.Resilience.net462.targets" />
```
RussKie, line 54: Shouldn't this be conditional?

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: You suggested above (https://github.com/dotnet/extensions/pull/5363#discussion_r1741274520) to move the whole PropertyGroup to the .targets file. I moved the group and removed the .props file. We now have only the .targets file, therefore we copy only it. Right?

RussKie: Ah yes, indeed.
Though, it we add a props file, we may forget to change this. Suggest keeping it as `*`.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: I believe it shouldn't. We just want to copy the .targets file into the buildTransitive folder. I tested it and it worked. Do you have some kind of condition in mind?

We have similar tag in another file and it doesn't use any condition:
https://github.com/dotnet/extensions/blob/3131920ec0aa445af3d8bc615a088b5caf81fac4/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Microsoft.Extensions.Telemetry.Abstractions.csproj#L27

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: Thank you. Done. Tested and it worked.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
RussKie: In the telemetry we aren't doing anything fancy. And if you exclude `net462` from TFMs, the telemetry package will only package `buildTransitive\net6.0`. But in this package, it will package `buildTransitive\net462` regardless, won't it?

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -73,6 +73,37 @@ clientBuilder.AddResilienceHandler("myHandler", b =>
...
+
+We've implemented a build time check that verifies if your using `Grpc.Net.ClientFactory` `2.63.0`
+or earlier, and if you are the check produces a compilation error. You can suppress the check and the
+error by setting the following property in the project's file:
+
+```xml
+<PropertyGroup>
+  <SuppressCheckGrpcNetClientFactoryVersion>true</SuppressCheckGrpcNetClientFactoryVersion>
+</PropertyGroup>
+```
```
dariusclay, line 105: My suggestion:

    If you are using `Grpc.Net.ClientFactory` version `2.63.0` or earlier, enabling
    the standard resilience or hedging handlers for a gRPC client may result in
    a runtime exception. For example, this code:

    ```csharp
    services
        .AddGrpcClient<Greeter.GreeterClient>()
        .AddStandardResilienceHandler();
    ```

    will trigger the following exception:

    ```
    System.InvalidOperationException: The ConfigureHttpClient method is not supported when creating gRPC clients. Unable to create client with name 'GreeterClient'.
    ```

    To resolve this issue, we recommend upgrading to `Grpc.Net.ClientFactory` version `2.64.0` or later.

    A build-time check produces a compilation error if you are if you
    are using `Grpc.Net.ClientFactory` version `2.63.0` or earlier.
    To suppress this check, set the following property in your project file:

    ```xml
    <PropertyGroup>
        <SuppressCheckGrpcNetClientFactoryVersion>true</SuppressCheckGrpcNetClientFactoryVersion>
    </PropertyGroup>
    ```

dariusclay (approved):
For doc changes I leave it up to @IEvangelist to confirm, otherwise everything looks good

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -73,6 +73,37 @@ clientBuilder.AddResilienceHandler("myHandler", b =>
 });
 ```
 
+## Known issues
+
+### Compatibility with the `Grpc.Net.ClientFactory` package
+
+If you're using `Grpc.Net.ClientFactory` `2.63.0` or earlier then enabling the standard resilience or
+hedging handlers for a gRPC client could cause a runtime exception. Specifically, the following code sample
```
IEvangelist, line 81:
```suggestion
If you're using `Grpc.Net.ClientFactory` version 2.63.0 or earlier, then enabling the standard resilience or hedging handlers for a gRPC client could cause a runtime exception. Specifically, consider the following code sample:
```

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -73,6 +73,37 @@ clientBuilder.AddResilienceHandler("myHandler", b =>
...
+
+```csharp
+services
+    .AddGrpcClient<Greeter.GreeterClient>()
+    .AddStandardResilienceHandler();
+```
+
+will cause the following exception
```
IEvangelist, line 89:
```suggestion
The preceding code results in the following exception:
```

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
```diff
@@ -73,6 +73,37 @@ clientBuilder.AddResilienceHandler("myHandler", b =>
...
+
+```
+System.InvalidOperationException: The ConfigureHttpClient method is not supported when creating gRPC clients. Unable to create client with name 'GreeterClient'.
+```
+
+If you're getting this exception consider using `Grpc.Net.ClientFactory` `2.64.0.` or later.
+
+We've implemented a build time check that verifies if your using `Grpc.Net.ClientFactory` `2.63.0`
+or earlier, and if you are the check produces a compilation error. You can suppress the check and the
+error by setting the following property in the project's file:
```
IEvangelist, line 99:
```suggestion
We've implemented a build time check that verifies if your using `Grpc.Net.ClientFactory` version 2.63.0 or earlier, and if you are the check produces a compilation error. You can suppress the check and the
error by setting the following property in the project's file:
```

IEvangelist (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/buildTransitive/Microsoft.Extensions.Http.Resilience.props
iliar-turdushev: Changed the link to learn.ms.com.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: Did as you suggested. Now we copy all files, i.e. use `.*`.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: Added a condition to check whether we need to pack buildTransitive for net462. The condition is the same that is used [here](https://github.com/dotnet/extensions/blob/dc1ba70804189b12ada17bd8b3ac56be93ebdf66/eng/MSBuild/Packaging.targets#L20C29-L22C169).

src/Libraries/Microsoft.Extensions.Http.Resilience/README.md
iliar-turdushev: Thank you for the feedback. Applied some changes.

src/Libraries/Microsoft.Extensions.Http.Resilience/buildTransitive/Microsoft.Extensions.Http.Resilience.targets
```diff
@@ -0,0 +1,53 @@
...
+
+      <!-- Find the version of the package in the Central Package Source. The solution uses the Central Package Management. -->
+      <_GrpcNetClientFactoryPackageVersion Include="@(PackageVersion)" Condition=" '%(PackageVersion.Identity)' == '$(_GrpcNetClientFactory)' " />
+      
+      <!-- The package is added to the project as a transitive dependency. -->
+      <_GrpcNetClientFactoryTransitiveDependency Include="@(ReferencePath)" Condition=" '%(ReferencePath.NuGetPackageId)' == '$(_GrpcNetClientFactory)' " />
+    </ItemGroup>
+
+    <!-- The version of the package is included in the .csproj file. -->
+    <Error Condition=" @(_GrpcNetClientFactoryPackageReference->Count()) &gt; 0
```
iliar-turdushev, line 28: @RussKie @joperezr Are we good with notifying users about potential issues with Grpc.Net.ClientFactory as compilation errors? I think it should an error, since it is much easier to not notice a warning. What are thoughts on that? Thank you.

joperezr: We would typically avoid as much as possible introducing a build breaking change across minor versions unless it is absolutely required. I don't have a lot of background on the issue here, but from the error message it seems to be that while this could cause issues it will not always be the case, so just based on that I would suggest to consider using a Warning instead. A lot of people use Warn as errors anyway, so this would still be caught. I like the fact that this is suppressible. One suggestion would be to update the Warning text depending on whether the version is specified via PackageReference, PackageVersion, or ReferencePath, as that would help the user diagnose and understand how to update the version.

RussKie: I actually think it should be an "error" since using [..., 2.63] would result in a runtime error. We want actively stop the user at the compile-time.

joperezr: Would it always be a runtime error? From the above that didn't seem like the case.

RussKie: My interpretation is that there's a significant risk per the new readme. And we have two independent reports.
I suppose, we could make it a warning and monitor for any similar reports in the future - that could be a signal to up the severity.

joperezr: up to you and @iliar-turdushev as I trust you can assess the risk here, but in general as you know we really try to avoid introducing an intentional breaking change in a minor release

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -38,4 +38,47 @@
...
+          CopyToOutputDirectory="PreserveNewest"
+          Pack="true"
+          PackagePath="buildTransitive\net462\"
+          Condition="'$(PackNet462)' == 'true'" />
+    
+    <None Include="buildTransitive\$(MSBuildProjectName).targets"
+          CopyToOutputDirectory="PreserveNewest"
+          Pack="true"
+          PackagePath="buildTransitive\net462\$(MSBuildProjectName).net462.targets"
+          Condition="'$(PackNet462)' == 'true'" />
```
RussKie, line 72: Whilst this works in this current state, I don't think it's correct in general.
It's conceivable that in the future there could be other *.targets files added, but they won't get packaged.

My original proposal in https://github.com/dotnet/extensions/pull/5363#discussion_r1741290810 was generalised, and will work in any situation, i.e.:
1. Pack everything under "buildTransitive" folder.
2. For net462, if it's used, pack "buildTransitive\\Microsoft.Extensions.Http.Resilience.targets" as "buildTransitive\\Microsoft.Extensions.Http.Resilience.net462.targets" because the original "buildTransitive\\Microsoft.Extensions.Http.Resilience.targets" will get overwritten.

```suggestion
    <!--
        For net462 we have to rename the .targets file because our build process will overwrite the content
        of the original buildTransitive\Microsoft.Extensions.Http.Resilience.targets.
        We import the copy via _AdditionalNETStandardCompatErrorFileContents variable.
      -->
    <None Include="buildTransitive\Microsoft.Extensions.Http.Resilience.targets"
          Condition="'$(MinimumSupportedTfmForPackaging)' != '' and
                     '$(ConditionalNet462)' != '' and
                     ($(TargetFrameworks.Contains('$(MinimumSupportedTfmForPackaging)')) and $(TargetFrameworks.Contains('$(ConditionalNet462)')))"
          CopyToOutputDirectory="PreserveNewest"
          Pack="true"
          PackagePath="buildTransitive\net462\Microsoft.Extensions.Http.Resilience.net462.targets" />
```

RussKie (changes requested):

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -38,4 +38,47 @@
...
     <InternalsVisibleToDynamicProxyGenAssembly2 Include="*" />
   </ItemGroup>
+
+  <PropertyGroup>
+    <!-- Defines whether we need to pack buildTransitive artifacts for net462. -->
+    <PackNet462 Condition=" '$(MinimumSupportedTfmForPackaging)' != '' and
+                            '$(ConditionalNet462)' != '' and
+                            ($(TargetFrameworks.Contains('$(MinimumSupportedTfmForPackaging)')) and $(TargetFrameworks.Contains('$(ConditionalNet462)'))) "
+                >true</PackNet462>
+  </PropertyGroup>
```
RussKie, line 48: It's a good idea to create a variable to avoid duplications. 👍 

Though the variable needs to live in eng/MSBuild/Packaging.props.
This way we can update use it here and in 
https://github.com/dotnet/extensions/blob/f335f2a9928580a361807ba2ced86f1ed27aca0b/eng/MSBuild/Packaging.targets#L20-L22

src/Libraries/Microsoft.Extensions.Http.Resilience/buildTransitive/Microsoft.Extensions.Http.Resilience.targets
iliar-turdushev: Thank you for the feedback. Decided to go with Warning message, because the error occurs when a user enables the standard resilience or hedging handlers for a GrpcClient which is currently unsupported use-case. We think that users will not do that intentionally.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: Thank you. Moved the variable to the Packaging.targets file as you'd suggested.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
RussKie: > I partially applied your suggestion. In particular,
> 
> * We pack everything for `MinimumSupportedTfmForPackaging`.
> * We pack everything except M.E.Http.Resilience.targets for `net462`. We have to explicitly exclude it from copying, because otherwise we get the following error, saying that msbuild cannot overwrite the file:
>   ```
>   C:\Repos\github\dotnet-extensions\.dotnet\sdk\8.0.300\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(221,5): warning NU5118: File 'C:\Repos\github\dotnet-extensions\artifacts\obj\Microsoft.Extensi
>   ons.Http.Resilience\netstandardcompaterror_net462.targets' is not added because the package already contains file 'buildTransitive\net462\Microsoft.Extensions.Http.Resilience.targets' [C:\Repos\github\dotnet-extensions\src\Libra 
>   ries\Microsoft.Extensions.Http.Resilience\Microsoft.Extensions.Http.Resilience.csproj]
>   ```
> * For `net462` we copy M.E.Http.Resilience.targets renaming it to M.E.Http.Resilience.net462.targets.
> 
> What do you think?

The reason you get the error is because there's a duplicated collection. There have to be only two items:
```diff
diff --git a/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj b/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
index f0499dada2..340fdb2abf 100644
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
@@ -41,22 +41,14 @@
 
   <ItemGroup>
     <None Include="buildTransitive\**\*"
-          CopyToOutputDirectory="PreserveNewest"
-          Pack="true"
-          PackagePath="buildTransitive\$(MinimumSupportedTfmForPackaging)\"
-          Condition="'$(MinimumSupportedTfmForPackaging)' != '' and $(TargetFrameworks.Contains('$(MinimumSupportedTfmForPackaging)'))" />
+           CopyToOutputDirectory="PreserveNewest"
+           Pack="true"
+           PackagePath="buildTransitive\$(MinimumSupportedTfmForPackaging)\" />
 
     <!-- For net462 we copy all buildTransitive files except the .targets file. We have to rename the .targets file
          when copying it because our build process automatically adds a .targets file with the same name as the project
          file. Then we'll import the renamed .targets file in the automatically added .targets file. For that we need
          to set the _AdditionalNETStandardCompatErrorFileContents variable. -->
-    <None Include="buildTransitive\**\*"
-          Exclude="buildTransitive\$(MSBuildProjectName).targets"
-          CopyToOutputDirectory="PreserveNewest"
-          Pack="true"
-          PackagePath="buildTransitive\net462\"
-          Condition="'$(IsPackNet462)' == 'true'" />
-    
     <None Include="buildTransitive\$(MSBuildProjectName).targets"
           CopyToOutputDirectory="PreserveNewest"
           Pack="true"
```

With that the package looks as follows:
image

src/Libraries/Microsoft.Extensions.Http.Resilience/buildTransitive/Microsoft.Extensions.Http.Resilience.targets
```diff
@@ -0,0 +1,54 @@
+<Project>
+  <PropertyGroup>
+    <_GrpcNetClientFactory>Grpc.Net.ClientFactory</_GrpcNetClientFactory>
+    <_CompatibleGrpcNetClientFactoryVersion>2.64.0</_CompatibleGrpcNetClientFactoryVersion>
+    <_GrpcNetClientFactoryVersionIsIncorrect>Grpc.Net.ClientFactory 2.63.0 or earlier could cause issues when used together with Microsoft.Extensions.Http.Resilience. For more details, see https://learn.microsoft.com/en-us/dotnet/core/resilience/http-resilience#known-issues. Consider using Grpc.Net.ClientFactory $(_CompatibleGrpcNetClientFactoryVersion) or later. To suppress the warning set SuppressCheckGrpcNetClientFactoryVersion=true.</_GrpcNetClientFactoryVersionIsIncorrect>
```
RussKie, line 5:
```suggestion
    <_GrpcNetClientFactoryVersionIsIncorrect>Grpc.Net.ClientFactory 2.63.0 or earlier could cause issues when used together with Microsoft.Extensions.Http.Resilience. For more details, see https://learn.microsoft.com/dotnet/core/resilience/http-resilience#known-issues. Consider using Grpc.Net.ClientFactory $(_CompatibleGrpcNetClientFactoryVersion) or later. To suppress the warning set SuppressCheckGrpcNetClientFactoryVersion=true.</_GrpcNetClientFactoryVersionIsIncorrect>
```

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: > The reason you get the error is because there's a duplicated collection. There have to be only two items:

Those collections are not duplicates. They copy files to different folders: net462 and net6.0. Please, note the difference:

```xml
    <None Include="buildTransitive\**\*"
          CopyToOutputDirectory="PreserveNewest"
          Pack="true"
          PackagePath="buildTransitive\$(MinimumSupportedTfmForPackaging)\"   <--- Copy to net6.0 folder.
          Condition="'$(MinimumSupportedTfmForPackaging)' != '' and $(TargetFrameworks.Contains('$(MinimumSupportedTfmForPackaging)'))" />

    <None Include="buildTransitive\**\*"
          Exclude="buildTransitive\$(MSBuildProjectName).targets"   <--- We have to exclude this file to avoid the warning.
          CopyToOutputDirectory="PreserveNewest"
          Pack="true"
          PackagePath="buildTransitive\net462\"   <--- Copy to net462 folder.
          Condition="'$(IsPackNet462)' == 'true'" />
```

RussKie (approved):

PR #5368, opened 2024-08-16
Use Configuration Binder in HeaderParsing

### LINKED ISSUE #4914: Set IsAotCompatible=true for Microsoft.AspNetCore.HeaderParsing
With https://github.com/dotnet/extensions/pull/4871 all projects in `src/Libraries` were marked as AOT compatible, so the trim and AOT analyzers run during the build.

The only library that couldn't be marked is Microsoft.AspNetCore.HeaderParsing. This is because it is blocked on https://github.com/dotnet/runtime/issues/94547.

https://github.com/dotnet/extensions/blob/a44a76e708dfbd74ece5f4b79d4c092ef6e14c10/src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj#L11-L12

When the underlying runtime bug is fixed we should enable the config source generator in this library and mark it as AOT compatible.

eerhardt:
https://github.com/dotnet/runtime/issues/94547 has been fixed in 9.0. We can now use the Configuration Binder in HeaderParsing.

This fixes the last AOT warning in the repo.

Fix #4914

NOTE: this targets the `dev` branch, since the underlying issue is only fixed in 9.0.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5368)

### DIFF
--- a/src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj
+++ b/src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj
@@ -8,14 +8,11 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
+    <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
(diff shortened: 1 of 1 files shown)


RussKie (approved):
Thank you

PR #5370, opened 2024-08-17
Bring parity to dotnet extensions logging generator behavior

### LINKED ISSUE #5178: Add support for primary constructors in our Logging generator
For more info, look a this PR: https://github.com/dotnet/runtime/pull/101660

cc: @tarekgh

### LINKED ISSUE #5221: `Microsoft.Extensions.Telemetry.Abstractions` doesn't find protected field `ILogger` for `[LoggerMessage]`
### Description

We have an ASP.NET Core 8 application. There are two classes: `Base` and `Derived`. `Derived` inherits from `Base`. `Base` contains a protected field `ILogger` which is used for logging in both classes.

When we added logging methods with `LoggerMessageAttribute`, they work as expected in both classes using the `ILogger` field from base class. But if we add the `Microsoft.Extensions.Telemetry.Abstractions` package, then we immediately have an error in the derived class:

> LOGGEN015	Couldn't find a field or property of type "Microsoft.Extensions.Logging.ILogger" in type "Derived"	

There is no such error in the base class.

### Reproduction Steps

Here's the `.csproj` definition:
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

	<PropertyGroup>
		<Nullable>enable</Nullable>
	</PropertyGroup>

	<ItemGroup>
<!--		<PackageReference Include="Microsoft.Extensions.Telemetry.Abstractions" Version="8.6.0" />-->
	</ItemGroup>

</Project>
```
`Program.cs`:
```c#
WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
WebApplication app = builder.Build();

app.MapGet("/", () => "Hello World!");

await app.RunAsync();

public abstract partial class Base(ILogger<Base> logger)
{
	protected readonly ILogger<Base> _logger = logger;

	public void BaseMethod()
	{
		LogInformationBase();
	}

	[LoggerMessage(Level = LogLevel.Information, Message = "Base LoggerMessage")]
	private partial void LogInformationBase();
}

public partial class Derived(ILogger<Derived> logger)
	: Base(logger)
{
	public void DerivedMethod()
	{
		_logger.LogInformation("Derived");
		LogInformationDerived();
	}

	[LoggerMessage(Level = LogLevel.Information, Message = "Derived LoggerMessage")]
	private partial void LogInformationDerived();
}
```
There is no error in this configuration.

To get the error, we should uncomment the `PackageReference` item in the `.csproj` file. After this, we will have an error:

image

### Expected behavior

In this situation, the source code generator from the `Microsoft.Extensions.Telemetry.Abstractions` package should behave like the source code generator from the `Microsoft.Extensions.Logging.Abstractions` package. When we use `[LoggerMessage]` in a derived class, it must find a protected field `ILogger` from the base class.

### Actual behavior

When we use the `Microsoft.Extensions.Telemetry.Abstractions` package, the source code generator doesn't find a protected field `ILogger` from the base class.

### Regression?

Regression compared to the `Microsoft.Extensions.Logging.Abstractions` package.

### Known Workarounds

_No response_

### Configuration

<details><summary>dotnet --info</summary>
<p>
ПАКЕТ SDK ДЛЯ .NET:
 Version:           8.0.300
 Commit:            326f6e68b2
 Workload version:  8.0.300-manifests.5273bb1c
 MSBuild version:   17.10.4+10fbfbf2e

Среда выполнения:
 OS Name:     Windows
 OS Version:  10.0.22631
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.300\

Установленные рабочие нагрузки .NET:
 [maccatalyst]
   Источник установки: VS 17.10.35004.147, VS 17.10.35004.147
   Версия манифеста:    17.2.8053/8.0.100
   Путь к манифесту:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.maccatalyst\17.2.8053\WorkloadManifest.json
   Тип установки:        FileBased

 [ios]
   Источник установки: VS 17.10.35004.147, VS 17.10.35004.147
   Версия манифеста:    17.2.8053/8.0.100
   Путь к манифесту:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.ios\17.2.8053\WorkloadManifest.json
   Тип установки:        FileBased

 [android]
   Источник установки: VS 17.10.35004.147, VS 17.10.35004.147
   Версия манифеста:    34.0.95/8.0.100
   Путь к манифесту:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.android\34.0.95\WorkloadManifest.json
   Тип установки:        FileBased

 [maui-windows]
   Источник установки: VS 17.10.35004.147, VS 17.10.35004.147
   Версия манифеста:    8.0.40/8.0.100
   Путь к манифесту:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.maui\8.0.40\WorkloadManifest.json
   Тип установки:        FileBased

 [wasm-tools]
   Источник установки: VS 17.10.35004.147, VS 17.10.35004.147
   Версия манифеста:    8.0.5/8.0.100
   Путь к манифесту:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.workload.mono.toolchain.current\8.0.5\WorkloadManifest.json
   Тип установки:        FileBased

 [aspire]
   Источник установки: VS 17.10.35004.147, VS 17.10.35004.147
   Версия манифеста:    8.0.0/8.0.100
   Путь к манифесту:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.aspire\8.0.0\WorkloadManifest.json
   Тип установки:        FileBased

Host:
  Version:      8.0.5
  Architecture: x64
  Commit:       087e15321b

.NET SDKs installed:
  3.1.426 [C:\Program Files\dotnet\sdk]
  5.0.101 [C:\Program Files\dotnet\sdk]
  5.0.201 [C:\Program Files\dotnet\sdk]
  5.0.408 [C:\Program Files\dotnet\sdk]
  6.0.202 [C:\Program Files\dotnet\sdk]
  6.0.321 [C:\Program Files\dotnet\sdk]
  8.0.300 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.17 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.17 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.19 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.23 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.10 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.26 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.19 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.15 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.19 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.23 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.10 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.15 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.26 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.30 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.31 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.19 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.15 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.19 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.23 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.10 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.15 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.26 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.30 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.31 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.19 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  arm64 [C:\Program Files\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\arm64\InstallLocation]
  x86   [C:\Program Files (x86)\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  Not set

global.json file:
  Not found</p>
</details> 

### Other information

_No response_

### LINKED ISSUE #5332: Behavior changes between .NET logging generator and .NET extensions
To add to this: I'm having a similar issue with the visibility of an inherited `ILogger` property.

We're using the [Azure Mobile Apps SDK](https://github.com/Azure/azure-mobile-apps/) from Microsoft. This SDK has a base class for a controller which you can inherit from, [containing a `public` property of type `ILogger`](https://github.com/Azure/azure-mobile-apps/blob/be660e558ad79ec59035acf73526d92dbbc7fbde/sdk/dotnet/src/Microsoft.AspNetCore.Datasync/Tables/TableController.cs#L94):
```csharp
namespace Microsoft.AspNetCore.Datasync
{
    ...
    public class TableController<TEntity> : ControllerBase where TEntity : class, ITableData
    {
        ...
        public ILogger Logger { get; set; } = null;
        ...
    }
}
```

I inherited from that class as intended by the Azure Mobile Apps SDK and in my class have a more specific `ILogger<T>` property as follows:
```csharp
public partial class InspectionSyncController : TableController<FarmInspectionSyncModel>
{
    private readonly ILogger<InspectionSyncController> logger;

    public InspectionSyncController(
        ...
        ILogger<InspectionSyncController> logger)
        : base(...)
    {
        ...
        this.logger = logger;
    }
    ...
    [LoggerMessage(
    EventId = 0,
    Level = LogLevel.Warning,
    Message = "Not authorized to perform operation `{Operation}` on entity `{EntityId}`")]
    private partial void LogNotAuthorizedToPerformOperationOnEntity(TableOperation operation, string? entityId);
    ...
}
```

This works without `Microsoft.Extensions.Http.Resilience`.

### Actual behavior
After adding `Microsoft.Extensions.Http.Resilience`, I get this warning:
> Found multiple fields or properties of type "Microsoft.Extensions.Logging.ILogger" in type "InspectionSyncController" (https://aka.ms/dotnet-extensions-warnings/LOGGEN016).

### Expected behavior
It still works after adding `Microsoft.Extensions.Http.Resilience`

_Originally posted by @hansmbakker in https://github.com/dotnet/extensions/issues/5173#issuecomment-2260871645_

### LINKED ISSUE #5336: LOGGEN031 Diagnostic breaks Serilog object serialization
### Description

Just simply including `Microsoft.Extensions.Http.Resilience` package in a project breaks source generated logging, because of `LOGGEN031` reporting that
```text
Error LOGGEN031 : Logging method "LogSomeMessage" has template(s) that start with "@": @SomeProperty (https://aka.ms/dotnet-extensions-warnings/LOGGEN031)
```

### Reproduction Steps

https://github.com/alexaka1/repro-dotnet-extensions-LOGGEN031/blob/main/WebApplication1/Program.cs#L16

### Expected behavior

The diagnostic does not error on use `@` symbol in logger templates, since many logging frameworks use this to indicate serialization of the property is required.

### Actual behavior

Cannot build project because of the error.

### Regression?

Don't know.

### Known Workarounds

Don't use the mentioned library.

### Configuration

```text
.NET SDK:
 Version:           8.0.303
 Commit:            29ab8e3268
 Workload version:  8.0.300-manifests.c915c39d
 MSBuild version:   17.10.4+10fbfbf2e

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  14.5
 OS Platform: Darwin
 RID:         osx-arm64
 Base Path:   /usr/local/share/dotnet/sdk/8.0.303/

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.7
  Architecture: arm64
  Commit:       2aade6beb0

.NET SDKs installed:
  8.0.303 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.7 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.7 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

Other architectures found:
  None

Environment variables:
  DOTNET_ROOT       [/opt/homebrew/opt/dotnet/libexec]

global.json file:
  Not found
```

### Other information

_No response_

dariusclay:
Fixes #5221 
Fixes #5332
Fixes #5336 
Resolves #5178 

- [X] Update message template extraction
- [X] Align behavior regarding @ symbol in message templates
- [X] Fix issues looking up ILogger in derived classes
- [X] Adds support for passing ILogger in primary class constructors

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5370)


test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
```diff
@@ -195,7 +195,7 @@ public async Task UnderscoresInMethodName()
                 partial class C
                 {
                     [LoggerMessage(0, LogLevel.Debug, ""M1"")]
-                    static partial void __M1(ILogger logger);
+                    static partial void /*0+*/__M1/*-0*/(ILogger logger);
```
geeknoid, line 198: What's this change about?

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
```diff
@@ -300,7 +300,39 @@ partial class C
                 static partial void M1(ILogger logger, int param);
             }}";
 
-        await RunGenerator(source, DiagDescriptors.TemplateStartsWithAtSymbol);
+        await RunGenerator(source);
+    }
+
+    [Theory]
```
geeknoid, line 306: In general, test cases that validate the happy path should be listed in the TestClasses, and then the generated code for those can be validated in the Generated/* test cases.

Otherwise, this test right now is validating that the generator doesn't fail, but it's not checking that the generated code is actually correct.

test/Generators/Microsoft.Gen.Logging/Generated/Microsoft.Gen.Logging.Generated.Tests.csproj
```diff
@@ -11,6 +11,7 @@
     <InjectIsExternalInitOnLegacy>true</InjectIsExternalInitOnLegacy>
     <InjectSharedNumericExtensions>true</InjectSharedNumericExtensions>
     <NoWarn>$(NoWarn);IDE0161;S1144</NoWarn>
+    <UseLoggingGenerator>true</UseLoggingGenerator>
```
geeknoid, line 14: This should not be needed here.

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
dariusclay: It's unneccessary now. The base generator does not allow underscores, but we can.

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
dariusclay: There's added tests for this under AtSymbolsTestExtensions.cs

test/Generators/Microsoft.Gen.Logging/Generated/Microsoft.Gen.Logging.Generated.Tests.csproj
dariusclay: You're right. I was trying to get the changes from the updated parsing logic to take hold. All I needed to do was clean + reload solution.

src/Generators/Microsoft.Gen.Logging/Parsing/Parser.cs
```diff
@@ -639,67 +617,125 @@ private void CheckTagNamesAreUnique(LoggingMethod lm, Dictionary<LoggingMethodPa
...
+                    {
+                        loggerMember = ms.Name;
+                        isNullable = typeSymbol.NullableAnnotation == NullableAnnotation.Annotated;
+                    }
+                    else if (!onMostDerivedType)
+                    {
+                        // we found a public logger member on a base class, prefer the existing one
+                        // because it's more likely to be the one the user intended
+                        // the dotnet logging generator doesn't support ILogger properties so this matches that behavior
+                        continue;
```
dariusclay, line 682: This is probably the most interesting case -- dotnet loggen doesn't support ILogger properties, but this one does. So our code would throw LOGGEN016: MultipleLoggerMembers in cases where the other wouldn't. This fixes that issue and we just prefer the most derived member if it exists.

dariusclay: For example see #5332

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogMethod.cs
```diff
@@ -0,0 +1,445 @@
...
+            ";
+
+        await RunGenerator(Source, DiagDescriptors.MissingLoggerMember);
+    }
+
+    [Fact]
+    public async Task ValidDerivation()
+    {
+        const string Source = @"
+                partial class C
```
dariusclay, line 76: - [ ] TODO: Add generated tests to validate this happy path

eng/Versions.props
```diff
@@ -82,5 +82,19 @@
     Versions below this comment are not managed by automation and can be changed as needed.
   -->
   <PropertyGroup Label="Manual">
+    <!-- Compatibility with VS 17.4/.NET SDK 7.0.1xx -->
```
geeknoid, line 85: @RussKie Is there a more elegant solution here?

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
geeknoid: Then you can delete these :-)

The idea is that anything that compiles is validated via the GeneratedTests, and the only things that get tested by the parser tests are things that don't compile, where we check the error handling behavior.

test/Generators/Microsoft.Gen.Logging/TestClasses/NonStaticTestClasses.cs
```diff
@@ -43,7 +43,33 @@ public partial class LoggerInNullablePropertyDerivedTestClass : LoggerInNullable
         public partial void M1(string p0);
     }
 
-    public partial class GenericLoggerInPropertyDerivedTestClass : LoggerInNullablePropertyTestClass
+    // The following classes are not used in any tests, but they validate that the generation is working correctly.
```
geeknoid, line 46: Why not? The point of these generated tests is that you can actually test that the code gen worked as you expected.

eng/Versions.props
```diff
@@ -82,5 +82,19 @@
...
+    <!-- Compatibility with the latest Visual Studio Preview release -->
+    <!--
+      The exact version is always a moving target. This version should never go ahead of the version of Roslyn that is included in the most recent
+      public Visual Studio preview version. If it were to go ahead, then any components depending on this version would not work in Visual Studio
+      and would cause a major regression for any local development that depends on those components contributing to the build.
+      This version must also not go ahead of the most recently release .NET SDK version, as that would break the source-build build.
+      Source-build builds the product with the most recent previously source-built release. Thankfully, these two requirements line up nicely
+      such that any version that satisfies the VS version requirement will also satisfy the .NET SDK version requirement because of how we ship.
+    -->
+    <MicrosoftCodeAnalysisVersion_LatestVS>4.8.0</MicrosoftCodeAnalysisVersion_LatestVS>
```
RussKie, line 98: This looks like a copy from https://github.com/dotnet/runtime/blob/59c2ea578bd615a63d56e8ff4b1de0a6b824691f/src/coreclr/tools/Directory.Build.props#L16-L25, though with a different value.

Overall, this property isn't being explicitly used. @dariusclay, is it needed at all? If it does - we need comments explaining why and how it is used. If not - it shouldn't be here.

eng/Versions.props
```diff
@@ -82,5 +82,19 @@
     Versions below this comment are not managed by automation and can be changed as needed.
   -->
   <PropertyGroup Label="Manual">
+    <!-- Compatibility with VS 17.4/.NET SDK 7.0.1xx -->
+    <MicrosoftCodeAnalysisVersion_4_4>4.4.0</MicrosoftCodeAnalysisVersion_4_4>
+    <!-- Compatibility with VS 17.8/.NET SDK 8.0.1xx -->
+    <MicrosoftCodeAnalysisVersion_4_8>4.8.0</MicrosoftCodeAnalysisVersion_4_8>
```
RussKie, line 88: How is this property used? Why do we need it?

eng/packages/General.props
```diff
@@ -5,10 +5,10 @@
...
     <PackageVersion Include="Microsoft.Bcl.TimeProvider" Version="$(MicrosoftBclTimeProviderVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.4.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.4.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.4.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.4.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="$(MicrosoftCodeAnalysisVersion_4_4)" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(MicrosoftCodeAnalysisVersion_4_4)" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="$(MicrosoftCodeAnalysisVersion_4_4)" />
+    <PackageVersion Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersion_4_4)" />
```
RussKie, line 11: In [dotnet/runtime](https://github.com/search?q=repo%3Adotnet%2Fruntime%20MicrosoftCodeAnalysisVersion_LatestVS&type=code) all packages that affect the VS design-time experience use `MicrosoftCodeAnalysisVersion_LatestVS` property
image

There's also a comment stating the reason: https://github.com/dotnet/runtime/blob/59c2ea578bd615a63d56e8ff4b1de0a6b824691f/eng/Versions.props#L40-L44

eng/Versions.props
RussKie: It doesn't look we're doing much different from dotnet/runtime, so I guess this is as elegant as it gets... There's no direct way of knowing what currently supported versions of VS are targeting.

src/Generators/Microsoft.Gen.Logging/Parsing/TemplateProcessor.cs
```diff
@@ -86,56 +129,68 @@ internal static void ExtractTemplates(string? message, List<string> templates)
...
+        return findIndex == -1 ? endIndex : findIndex;
     }
 
-    private static int FindBraceIndex(string message, char brace, int startIndex, int endIndex)
+    /// <summary>
+    /// Searches for the next brace index in the message.
+    /// </summary>
+    /// <remarks> The search skips any sequences of {{ or }}.</remarks>
+    /// <example>{{prefix{{{Argument}}}suffix}}.</example>
+    /// <returns>The zero-based index position of the first occurrence of the searched brace; -1 if the searched brace was not found; -2 if the wrong brace was found.</returns>
```
RussKie, line 140: Perhaps, it makes sense to introduce consts for the "magic" numbers. It'd make the code a little easier to comprehend.

src/Shared/DiagnosticIds/DiagnosticIds.cs
```diff
@@ -86,6 +86,8 @@ internal static class LoggerMessage
         internal const string LOGGEN034 = nameof(LOGGEN034);
         internal const string LOGGEN035 = nameof(LOGGEN035);
         internal const string LOGGEN036 = nameof(LOGGEN036);
+        internal const string LOGGEN037 = nameof(LOGGEN037);
+        internal const string LOGGEN038 = nameof(LOGGEN038);
```
RussKie, line 90: Need to update https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md and add an aka.ms entry.

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
dariusclay: Hmm, with parser tests also providing the GeneratedSourceResult we can additionally test the actual source being generated and use Verify against it.

eng/Versions.props
dariusclay: We only have been targeting Roslyn 4.4 where as dotnet/runtime also has targeting projects for 3.11 and 4.0 and 4.4+. This part is being added to prep in case we want to use the same approach. Right now it's still using only 4.4

eng/Versions.props
dariusclay: It's from https://github.com/dotnet/runtime/blob/dc0432bcb6e7afba54c426f15d03fe8c8fd0d0aa/eng/Versions.props, see comment above

src/Shared/DiagnosticIds/DiagnosticIds.cs
dariusclay: Thanks, missed that!

src/Generators/Microsoft.Gen.Logging/Parsing/TemplateProcessor.cs
dariusclay: I get it for the part of returning the index -1, or -2. But it also complains about `bracesCount % 2` 😄 
Is that really a magic number, it's not like it would be anything else if checking for evens.

eng/packages/General.props
dariusclay: This is to keep this PR focused only on Logging. If we bump to latest, then there's other areas which would needing updated to resolve warnings. We can do another PR which addresses all the warnings in other non-logging related generators (options, meter, etc).

src/Generators/Microsoft.Gen.Logging/Parsing/TemplateProcessor.cs
RussKie: I see a value of introducing consts for -1 and -2 as those are used in few places (including the docs). It'll make the maintenance easier.
`bracesCount % 2` can/should be supressed, as you do.

eng/Versions.props
RussKie: If we don't use it - then we shouldn't have it. We should only add it when/if it's required. Otherwise, it, essentially, becomes a technical debt.

eng/packages/General.props
RussKie: Sounds good.

test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.cs
dariusclay: Thinking about it, the emitter tests could do this instead. I have some things to handle and will return back to the PR in a few days.

eng/Versions.props
RussKie: Since we're not using this, it should be removed.

test/Analyzers/Microsoft.Analyzers.Extra.Tests/Resources/RoslynTestUtils.cs
```diff
@@ -538,7 +538,7 @@ public override async Task<IEnumerable<Diagnostic>> GetProjectDiagnosticsAsync(P
             }
 
             var comp = await _project.GetCompilationAsync(cancellationToken).ConfigureAwait(false);
-            var diags = await comp!.WithAnalyzers(_analyzers, cancellationToken: cancellationToken)
+            var diags = await comp!.WithAnalyzers(_analyzers)
```
dariusclay, line 541: CodeAnalysis 4.8 does not have a WithAnalyzers overload with CancellationToken.

test/Analyzers/Microsoft.Analyzers.Local.Tests/Resources/RoslynTestUtils.cs
```diff
@@ -257,9 +257,10 @@ public static async Task<IReadOnlyList<Diagnostic>> RunAnalyzer(
         var analyzers = ImmutableArray.Create(analyzer);
 
         var comp = await proj!.GetCompilationAsync().ConfigureAwait(false);
-        var diags = await comp!.WithAnalyzers(analyzers, options).GetAllDiagnosticsAsync().ConfigureAwait(false);
+        var result = await comp!.WithAnalyzers(analyzers, options)
+            .GetAnalysisResultAsync(CancellationToken.None).ConfigureAwait(false);
 
-        return Sort(diags);
+        return Sort(result.GetAllDiagnostics());
```
dariusclay, line 263: This correctly returns only diagnostics from the analyzer being ran.

src/Generators/Microsoft.Gen.Logging/Microsoft.Gen.Logging.csproj
```diff
@@ -7,6 +7,7 @@
 
   <PropertyGroup>
     <AnalyzerLanguage>cs</AnalyzerLanguage>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
```
dariusclay, line 10: This will also need to be included in the other source gen libs

eng/Versions.props
dariusclay: Cleaned this part up.

src/Generators/Directory.Build.targets
```diff
@@ -9,5 +9,9 @@
   <PropertyGroup>
     <!-- Stop complaining about enabling /doc for tests -->
     <NoWarn>$(NoWarn);RT0000</NoWarn>
+
+    <!-- Stop warnings about banned symbols in generators -->
+    <!-- https://github.com/dotnet/roslyn/issues/63290 -->
+    <NoWarn>$(NoWarn);RS1035</NoWarn>
```
dariusclay, line 15: The report generators use IO, so we either need to do it differently or keep this suppression.

RussKie: Does it affect all gen projects? -or- Can we suppress only in the affected project?

src/Generators/Microsoft.Gen.Logging/Parsing/Parser.cs
```diff
@@ -639,67 +617,125 @@ private void CheckTagNamesAreUnique(LoggingMethod lm, Dictionary<LoggingMethodPa
...
+                    if (loggerMember == null)
+                    {
+                        loggerMember = ms.Name;
+                        isNullable = typeSymbol.NullableAnnotation == NullableAnnotation.Annotated;
+                    }
+                    else if (!onMostDerivedType)
+                    {
+                        // we found a public logger member on a base class, prefer the existing one
+                        // because it's more likely to be the one the user intended
+                        // the dotnet logging generator doesn't support ILogger properties so this matches that behavior
```
RussKie, line 681:
```suggestion
                        // We found a public logger member on a base class, prefer the existing one
                        // because it's more likely to be the one the user intended.
                        // The .NET runtime logging generator doesn't support ILogger properties, so this matches that behavior.
```

src/Generators/Microsoft.Gen.Logging/Parsing/Parser.cs
```diff
@@ -639,67 +617,125 @@ private void CheckTagNamesAreUnique(LoggingMethod lm, Dictionary<LoggingMethodPa
...
+            }
+
+            onMostDerivedType = false;
+            currentClassType = currentClassType.BaseType;
+        }
+
+        // We prioritize fields over primary constructor parameters and avoid warnings if both exist.
+        if (loggerMember is not null)
+        {
+            return (loggerMember, null, isNullable);
```
RussKie, line 702: [nit/readability] add parameter names for "opaque" arguments (i.e., `true/false/null/""`):
```suggestion
            return (loggerMember, secondMember: null, isNullable);
```

_In some cases_, this may be even better:
```suggestion
            return (member: loggerMember, secondMember: null, isNullable);
```



Ditto other places.

src/Generators/Microsoft.Gen.Logging/Parsing/Parser.cs
```diff
@@ -639,67 +617,125 @@ private void CheckTagNamesAreUnique(LoggingMethod lm, Dictionary<LoggingMethodPa
...
-                            isNullable = f.Type.NullableAnnotation == NullableAnnotation.Annotated;
-                        }
-                        else
-                        {
-                            return (null, f, isNullable);
-                        }
+                        continue;
+                    }
+
+                    if (loggerMember == null)
```
RussKie, line 725:
```suggestion
                    if (loggerMember is null)
```

RussKie (commented):
Few more minor comments in addition to few outstanding comments (not sure if you addressed those and forgot to push... ¯\\_(ツ)\_/¯).

src/Generators/Directory.Build.targets
dariusclay: This would affect all gen projects. The suppression can be moved. 👍

dariusclay:
> Few more minor comments in addition to few outstanding comments (not sure if you addressed those and forgot to push... ¯\_(ツ)_/¯).

The final comments are being addressed while bringing the code coverage back up to meet the threshold.

RussKie:
I'm going to review either tomorrow or early next week.

dariusclay:
> I'm going to review either tomorrow or early next week.

Thanks for everything @RussKie

src/Generators/Microsoft.Gen.Logging/Parsing/TemplateProcessor.cs
```diff
@@ -86,56 +128,69 @@ internal static void ExtractTemplates(string? message, List<string> templates)
...
```
RussKie, line 139:
```suggestion
    /// <returns>The zero-based index position of the first occurrence of the searched brace; <see cref="NoBracesFound"/> if the searched brace was not found; <see cref="WrongBraceFound"/> if the wrong brace was found.</returns>
```

RussKie (approved):

RussKie:
@dariusclay feel free to squash-merge when you're ready.

PR #5391, opened 2024-08-28
Migrate HybridCache from aspnetcore to extensions

mgravell:
Plan agreed in "tactics" to migrate [`HybridCache`](https://learn.microsoft.com/aspnet/core/performance/caching/hybrid) from dotnet/aspnetcore to dotnet/extensions.

This is a direct snapshot of the code from [this point in dotnet/aspnetcore](https://github.com/dotnet/aspnetcore/tree/82b0fc9f43ae2bd50fb95f427116efc2f6f094df/src/Caching/Hybrid), with minor modifications to the library code:

1. to satisfy local repo style rules (member order, comment location, spacing, etc)
2. work with different build setup (csproj delta)
    
Minor global changes:
    
1. support netstandard2.1 on some legacy shim includes (the previous preview builds have included ns2.0 and ns2.1 targets)
2. add some additional required packages (for testing)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5391)

Reasons for move:

1. additional work required before GA (question: should we mark this `<IsPackable>false</IsPackable>` for now?)
2. to utilize higher release cadence and additional change flexibility for OOB NuGet packages


mgravell:
remaining CI is about adding `[Experimental]`; happy to take guidance here on the best approach

mgravell:
test failure looks unrelated?

``` txt
VSTest : error : Tests failed: D:\a\_work\1\s\artifacts\TestResults\Release\Microsoft.Extensions.Diagnostics.Probes.Tests_net8.0_x64.html [net8.0|x64] [D:\a\_work\1\s\test\Libraries\Microsoft.Extensions.Diagnostics.Probes.Tests\Microsoft.Extensions.Diagnostics.Probes.Tests.csproj]
##[error]VSTest(0,0): error : (NETCORE_ENGINEERING_TELEMETRY=Build) Tests failed: D:\a\_work\1\s\artifacts\TestResults\Release\Microsoft.Extensions.Diagnostics.Probes.Tests_net8.0_x64.html [net8.0|x64]
```

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
```diff
@@ -0,0 +1,153 @@
...
+                if (!pendingLegacy.IsCompletedSuccessfully)
+#else
+                if (pendingLegacy.Status != TaskStatus.RanToCompletion)
+#endif
+                {
+                    return new(AwaitedLegacyAsync(pendingLegacy, this));
+                }
+
+                return new(GetValidPayloadSegment(pendingLegacy.Result)); // already complete
+            case CacheFeatures.BackendCache | CacheFeatures.BackendBuffers: // IBufferWriter<byte>-based
```
RussKie, line 35: [styling] Adding a blank line before this one will help with readability.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            return new(payload);
+        }
+
+        return default;
+    }
+
+    [DoesNotReturn]
+    [MethodImpl(MethodImplOptions.NoInlining)]
```
RussKie, line 123: Why no inlining? All throw helpers are marked with aggressive inlining (see https://github.com/dotnet/extensions/blob/main/src/Shared/Throw/Throw.cs).

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.MutableCacheItem.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            _serializer = serializer;
+            var writer = RecyclableArrayBufferWriter<byte>.Create(maxLength);
+            serializer.Serialize(value, writer);
+
+            _buffer = new(writer.DetachCommitted(out var length), length, returnToPool: true);
+            writer.Dispose(); // no buffers left (we just detached them), but just in case of other logic
+        }
+
+        public override bool NeedsEvictionCallback => _buffer.ReturnToPool;
```
RussKie, line 30: [Styling] Properties must be defined before methods.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.Serialization.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.DependencyInjection;
+
+#if !(NETCOREAPP2_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER)
+using System.Runtime.InteropServices;
+using System.Runtime.Serialization;
+#endif
+
+namespace Microsoft.Extensions.Caching.Hybrid.Internal;
+internal partial class DefaultHybridCache
```
RussKie, line 17: [Styling] Missing blank line.
Also, unnecessary blank lines in the usings block.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.Serialization.cs
```diff
@@ -0,0 +1,121 @@
...
+                throw new InvalidOperationException($"No {nameof(IHybridCacheSerializer<T>)} configured for type '{typeof(T).Name}'");
+            }
+
+            // store the result so we don't repeat this in future
+            @this._serializers[typeof(T)] = serializer;
+            return serializer;
+        }
+    }
+
+    internal static class ImmutableTypeCache<T> // lazy memoize; T doesn't change per cache instance
```
RussKie, line 60: [Styling] one type per file

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.Serialization.cs
```diff
@@ -0,0 +1,121 @@
...
+            // the overall nullability; for example, a type could contain a private int[] field,
+            // where the field is mutable and the list is mutable; but if the type is annotated:
+            // we're trusting that the API and use-case is such that the type is immutable
+            return type.GetCustomAttribute<ImmutableObjectAttribute>() is { Immutable: true };
+        }
+
+        // don't trust interfaces and non-sealed types; we might have any concrete
+        // type that has different behaviour
+        return false;
+
```
RussKie, line 119: [Styling] blank line

RussKie:
> test failure looks unrelated?
> 
> ```
> VSTest : error : Tests failed: D:\a\_work\1\s\artifacts\TestResults\Release\Microsoft.Extensions.Diagnostics.Probes.Tests_net8.0_x64.html [net8.0|x64] [D:\a\_work\1\s\test\Libraries\Microsoft.Extensions.Diagnostics.Probes.Tests\Microsoft.Extensions.Diagnostics.Probes.Tests.csproj]
> ##[error]VSTest(0,0): error : (NETCORE_ENGINEERING_TELEMETRY=Build) Tests failed: D:\a\_work\1\s\artifacts\TestResults\Release\Microsoft.Extensions.Diagnostics.Probes.Tests_net8.0_x64.html [net8.0|x64]
> ```
> 
> with
> 
> ```
> Microsoft.Extensions.Diagnostics.Probes.Test.TcpEndpointProbesServiceTests.ExecuteAsync_CheckListenerOpenAndCloseAfterHealthStatusEvents
> Assert.False() Failure
> Expected: False
> Actual:   True
> ```

If it self-heals on a re-run - it can be safely ignored.

RussKie:
> remaining `Correctness WarningCheck` fail is about adding `[Experimental]`; happy to take guidance here on the best approach; I have a spike locally with `[Experimental]` using token `EXTEXP0018` (IIRC), but not sure how we add the aka.ms link doc.

Yes, let's mark it as `[Experimental]`. 
We'll need to add an entry in to https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md. I can show you how to add a redirection link when we merge this.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.cs
```diff
@@ -0,0 +1,170 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.Caching.Hybrid.Internal;
+
+/// <summary>
+/// The inbuilt ASP.NET implementation of <see cref="HybridCache"/>.
```
RussKie, line 21: Should this comment be reworded?

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
```diff
@@ -0,0 +1,41 @@
...
+  <PropertyGroup>
+    <Description>Multi-level caching implementation building on and extending IDistributedCache</Description>
+    <!--
+      relevant TFM feature summary (all "nice to have", with reasonable workarounds):
+        nca 2.0 / ns 2.1: RuntimeHelpers.IsReferenceOrContainsReferences; Task.IsCompletedSuccessfully; HashCode
+        nca 3.0 : IThreadPoolWorkItem
+        net 5.0 : [DynamicallyAccessedMembers], EncodingExtensions.GetString/GetBytes, IsExternalInit
+        net 7.0 : ArgumentNullException.ThrowIfNull
+    -->
+    <TargetFrameworks>$(NetCoreTargetFrameworks)$(ConditionalNet462);net8.0;netstandard2.0;netstandard2.1</TargetFrameworks>
```
RussKie, line 12: I don't believe explicit `net8.0` is required. `$(NetCoreTargetFrameworks)` should be `net9.0;net8.0`, isn't this the case?

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
```diff
@@ -0,0 +1,41 @@
...
+
+  <PropertyGroup>
+    <Description>Multi-level caching implementation building on and extending IDistributedCache</Description>
+    <!--
+      relevant TFM feature summary (all "nice to have", with reasonable workarounds):
+        nca 2.0 / ns 2.1: RuntimeHelpers.IsReferenceOrContainsReferences; Task.IsCompletedSuccessfully; HashCode
+        nca 3.0 : IThreadPoolWorkItem
+        net 5.0 : [DynamicallyAccessedMembers], EncodingExtensions.GetString/GetBytes, IsExternalInit
+        net 7.0 : ArgumentNullException.ThrowIfNull
+    -->
```
RussKie, line 11: Is this comment necessary? I'm not seeing much relevance here.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
```diff
@@ -0,0 +1,41 @@
...
+
+    <!-- ignore "use [Experimental] for new API" -->
+    <NoWarn>$(NoWarn),LA0003</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Options" />
+    <InternalsVisibleToTest Include="Microsoft.Extensions.Caching.Hybrid.Tests" />
```
RussKie, line 31: Please move out to own `ItemGroup`

test/Libraries/Microsoft.Extensions.Caching.Hybrid.Tests/Microsoft.Extensions.Caching.Hybrid.Tests.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
```
RussKie, line 4: Any specific reason not to test net462 version?

test/Libraries/Microsoft.Extensions.Caching.Hybrid.Tests/StampedeTests.cs
```diff
@@ -0,0 +1,473 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Hybrid.Internal;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Caching.Hybrid.Tests;
```
RussKie, line 11: [styling] missing blank line

RussKie (commented):
I started the review before I realised this changeset moves implementations from dotnet/aspnetcore. I left a number of style-related comments, as I'm guessing dotnet/aspnetcore follows a different style to us.

Besides that there are few other comments and questions. I find the use of `!` operator and reliance one Debug.Assert somewhat worrisome, it does look those cases could result either in NREs or incorrect behaviours in release builds.

Overall, it would be good to understand the target state of this code move (since I'm not attending the Tactics) - will dotnet/aspnetcore version be deprecated, and dotnet/aspnetcore will instead depend on our artifacts and implementations? What would this mean for consumers? Etc.

mgravell:
> Yes, let's mark it as `[Experimental]`.
> We'll need to add an entry in to https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md. I can show you how to add a redirection link when we merge this.

Discussed this with Jose separately; this API has already been shipped non-[Experimental] in multiple net9 previews; might be confusing to add it now. That said, I'm happy with any outcome - this is not a hill I need to die on, as long as it is appropriately considered ;p

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
mgravell: Ah, I must have looked at [`main`](https://github.com/dotnet/extensions/blob/main/Directory.Build.props#L16) rather than [`dev`](https://github.com/dotnet/extensions/blob/dev/Directory.Build.props#L17C66-L17C73); yes, you're right, in `dev`, net8.0 is included - will remove (for context: the intent here is "also include LTS, because Aspire targets LTS" - if that happens to match the intent of "extensions", then: great!)

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.Extensions.Caching.Hybrid.Internal;
+
+// used to convey buffer status; like ArraySegment<byte>, but Offset is always
+// zero, and we use the most significant bit (MSB, the sign flag) of the length
+// to track whether or not to recycle this value
+internal readonly struct BufferChunk
+{
+    private const int MSB = (1 << 31);
```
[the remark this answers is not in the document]
mgravell: Already there, see [L12](https://github.com/dotnet/extensions/pull/5391/files/501ec0013f9c2124bcd40c9ae8c94bf79a89c16a#diff-520465981a4a3c668ccffb4dc31b0d6372c820e5ad22e0e816fc6aaef59dd6d6R12) - however, renaming for clarity

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -0,0 +1,84 @@
...
+    public byte[] ToArray()
+    {
+        var length = Length;
+        if (length == 0)
+        {
+            return [];
+        }
+
+        var copy = new byte[length];
+        Buffer.BlockCopy(Array!, 0, copy, 0, length);
```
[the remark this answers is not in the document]
mgravell: It is guaranteed, but rather than explain *here*, I'll add in a comment for future eyeballs.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -0,0 +1,84 @@
...
+        return copy;
+    }
+
+    public BufferChunk(byte[] array)
+    {
+        Debug.Assert(array is not null, "expected valid array input");
```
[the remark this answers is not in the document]
mgravell: This is intentional; this is an `internal` type with fully checked usage and pre-validated inputs (and NRT checks); we genuinely do not expect to ever see this debug exception - it is added for paranoia only, so IMO the assertion is sufficient to catch future fat-finger problems, and if not: the NRE+stacktrace is a reasonable and pragmatic/acceptable output.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -0,0 +1,84 @@
...
+        Array = array;
+        _lengthAndPoolFlag = array!.Length;
+
+        // assume not pooled, if exact-sized
+        // (we don't expect array.Length to be negative; we're really just saying
+        // "we expect the result of assigning array.Length to _lengthAndPoolFlag
+        // to give the expected Length *and* not have the MSB set; we're just
+        // checking that we haven't fat-fingered our MSB logic)
+        Debug.Assert(!ReturnToPool, "do not return right-sized arrays");
+        Debug.Assert(Length == array.Length, "array length not respected");
```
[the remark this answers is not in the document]
mgravell: This is intended as an ongoing paranoid "check we haven't fat-fingered the logic"; if it is contentious here, my preference would be to remove them from here and add them as unit-tests instead.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -0,0 +1,84 @@
...
+        if (ReturnToPool)
+        {
+            ArrayPool<byte>.Shared.Return(Array!);
+        }
+
+        Unsafe.AsRef(in this) = default; // anti foot-shotgun double-return guard; not 100%, but worth doing
+        Debug.Assert(Array is null && !ReturnToPool, "expected clean slate after recycle");
+    }
+
+    internal ReadOnlySequence<byte> AsSequence() => Length == 0 ? default : new ReadOnlySequence<byte>(Array!, 0, Length);
```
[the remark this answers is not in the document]
mgravell: same logic as from `ToArray`; adding cross-reference comment

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.CacheItem.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            return false;
+        }
+
+        public bool TryReserve()
+        {
+            // this is basically interlocked increment, but with a check against:
+            // a) incrementing upwards from zero
+            // b) overflowing *back* to zero
+            var oldValue = Volatile.Read(ref _refCount);
```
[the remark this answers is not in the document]
mgravell: Personally I'd argue that a read on a "count" is fairly obviously an integer, but: changed

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.Debug.cs
```diff
@@ -0,0 +1,83 @@
...
+
+#if DEBUG
+using System.Threading;
+#endif
+
+namespace Microsoft.Extensions.Caching.Hybrid.Internal;
+
+internal partial class DefaultHybridCache
+{
+    internal bool DebugTryGetCacheItem(string key, [NotNullWhen(true)] out CacheItem? value)
```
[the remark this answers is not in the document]
mgravell: comment added

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.ImmutableCacheItem.cs
```diff
@@ -0,0 +1,47 @@
...
+
+internal partial class DefaultHybridCache
+{
+    private sealed class ImmutableCacheItem<T> : CacheItem<T> // used to hold types that do not require defensive copies
+    {
+        private T _value = default!; // deferred until SetValue
+
+        public void SetValue(T value) => _value = value;
+
+        private static ImmutableCacheItem<T>? _sharedDefault;
```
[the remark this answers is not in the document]
mgravell: seems odd that the auto-enforcer didn't catch this one, but: rectified

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.MutableCacheItem.cs
mgravell: fixed; aside: I wonder if the auto-enforcer could catch this

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.Serialization.cs
mgravell: in the usings: is separating a conditional set (TFM-dependent - otherwise get "remove unused using" on some TFM builds); IMO much clearer with the separation

mgravell:
@RussKie  thanks for the detailed feedback; commit with tweaks pushed

mgravell:
> Besides that there are few other comments and questions. I find the use of ! operator and reliance one Debug.Assert somewhat worrisome, it does look those cases could result either in NREs or incorrect behaviours in release builds.

The assertions here are for things that we already absolutely positively already believe to be true and pre-validated by separate logic, but we just want to be paranoid in debug, because of paranoia.

> Overall, it would be good to understand the target state of this code move (since I'm not attending the Tactics) - will dotnet/aspnetcore version be deprecated, and dotnet/aspnetcore will instead depend on our artifacts and implementations? What would this mean for consumers? Etc.

Yes, the plan is to relocate `HybridCache` from dotnet/aspnetcore to dotnet/extensions (but that doesn't mean it becomes "your problem" - it is still 100% my fault on any ticket items here). Basically, this is an OOB NuGet package that has absolutely no need to be tied to dotnet/aspnetcore other than habit. We actively want to make use of the different build train here; however, there should be no change to consumers (other than we have more flexibility to evolve and iterate out-of-step of runtime/aspnetcore).

test/Libraries/Microsoft.Extensions.Caching.Hybrid.Tests/Microsoft.Extensions.Caching.Hybrid.Tests.csproj
mgravell: I've added net462, but I'm having difficulty validating that locally - the runner seems to have a SNK problem - and I expect it to fail the CI build:

``` txt
[xUnit.net 00:00:00.72] Microsoft.Extensions.Caching.Hybrid.Tests: Catastrophic failure: System.IO.FileLoadException: Could not load file or assembly 'Microsoft.Extensions.Caching.Hybrid.Tests, Version=9.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)
File name: 'Microsoft.Extensions.Caching.Hybrid.Tests, Version=9.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
```

If I remove the signing, I get instead:

``` txt
Friend access was granted by 'Microsoft.Extensions.Caching.Hybrid, Version=9.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35', but the public key of the output assembly ('') does not match that specified by the InternalsVisibleTo attribute in the granting assembly.
```

Any advice on making that happier?

RussKie:
> Yes, the plan is to relocate `HybridCache` from dotnet/aspnetcore to dotnet/extensions (but that doesn't mean it becomes "your problem" - it is still 100% my fault on any ticket items here). Basically, this is an OOB NuGet package that has absolutely no need to be tied to dotnet/aspnetcore other than habit. We actively want to make use of the different build train here; however, there should be no change to consumers (other than we have more flexibility to evolve and iterate out-of-step of runtime/aspnetcore).

Sounds good, the more - the merrier :)

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
RussKie: Yep, we're targeting the LTS and the most recent supported STS.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.CacheItem.cs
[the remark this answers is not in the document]
RussKie: It could be an `int` or `short` or `byte` or `long`... :)

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.ImmutableCacheItem.cs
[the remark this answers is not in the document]
RussKie: Aye. But some styling preferences are kind of hard to enforce. 
¯\\_(ツ)\_/¯

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
RussKie: Ah! Thank you for explaining.

RussKie:
> > Yes, let's mark it as `[Experimental]`.
> > We'll need to add an entry in to [`main`/docs/list-of-diagnostics.md](https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md?rgh-link-date=2024-08-29T08%3A54%3A21Z). I can show you how to add a redirection link when we merge this.
> 
> Discussed this with Jose separately; this API has already been shipped non-[Experimental] in multiple net9 previews; might be confusing to add it now. That said, I'm happy with any outcome - this is not a hill I need to die on, as long as it is appropriately considered ;p

I wasn't in the loop.
But since this is an OOB package, and we're just changing the repo and leaving the implementation as-is (bar styling changes), then, it makes sense to ship it as-is too (i.e., non-experimental). In that case, we'll need to set the `Stage=normal` (which has few additional implications - like certain test coverage levels). I'll have to brush up on that.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -0,0 +1,94 @@
...
+        var copy = new byte[length];
+        Buffer.BlockCopy(Array!, 0, copy, 0, length);
+        return copy;
+
+        // Note on nullability of Array; the usage here is that a non-null array
+        // is always provided during construction, so the only null scenario is for default(BufferChunk).
+        // Since the constructor explicitly accesses array.Length, any null array passed to the constructor
+        // will cause an exception, even in release (the Debug.Assert only covers debug) - although in
+        // reality we do not expect this to ever occur (internal type, usage checked, etc). In the case of
+        // default(BufferChunk), we know that Length will be zero, which means we will hit the [] case.
```
RussKie, line 44: I figured what triggered and confused me in this class - the .ctor is interleaved with other methods. It's customary (and per the old .NET code guidelines) to place the .ctor after fields and events, and I'm so used to that - and I totally missed the .ctor and, thus, got confused by the `!` operator use.

RussKie (approved):

test/Libraries/Microsoft.Extensions.Caching.Hybrid.Tests/Microsoft.Extensions.Caching.Hybrid.Tests.csproj
RussKie: It looks like your issue may have gone away. The CI build is green, and I tried your branch locally - al tests are green.
image

RussKie:
> > Yes, let's mark it as `[Experimental]`.
> > We'll need to add an entry in to [`main`/docs/list-of-diagnostics.md](https://github.com/dotnet/extensions/blob/main/docs/list-of-diagnostics.md?rgh-link-date=2024-08-29T08%3A54%3A21Z). I can show you how to add a redirection link when we merge this.
> 
> Discussed this with Jose separately; this API has already been shipped non-[Experimental] in multiple net9 previews; might be confusing to add it now. That said, I'm happy with any outcome - this is not a hill I need to die on, as long as it is appropriately considered ;p

In this repo we only have two states - "**dev**" and "**normal**" or (in a consumer terms) - "**experimental**" and "**prod-ready**". All our packages except for the following three are "prod-ready", these three are "experimental":
- Microsoft.Extensions.Diagnostics.Probes
- Microsoft.Extensions.Hosting.Testing
- Microsoft.Extensions.Options.Contextual

During the "9.0-previewX" releases there are not visible branding difference, however, at and after the official release the "prod-ready" packages have stable branding, e.g.: "_Microsoft.Extensions.Compliance.Testing.8.8.0.nupkg_"; and the "experimental" packages have not, e.g.: "_Microsoft.Extensions.Diagnostics.Probes.8.8.0-**rtm.24412.9**.nupkg_".

The "experimental" packages are also decorated with the `ExperimentalAttribute` at the assembly level with the assigned warning code.
E.g.:
https://github.com/dotnet/extensions/blob/135a94a29b49130a8302f045e20b87e772e5a610/src/Libraries/Microsoft.Extensions.Options.Contextual/Directory.Build.props#L2-L8

So... **_if_** we're not expecting any further API changes in this package, then we can mark the package as "prod-ready", i.e., set `<Stage>normal</Stage>`. Otherwise, we need to mark the package as `<Stage>dev</Stage>`

We also need to do the following:
* ensure the test coverage is above 75% (i.e., `<MinCodeCoverage>XX</MinCodeCoverage>`, where `XX` is an integer in [75, 100]).
* since we're currently not running any mutation tests, `MinMutationScore` can be set to anything above 50 (if necessary),
* run [ApiChief](https://github.com/dotnet/extensions/tree/main/eng/Tools/ApiChief) to generate the baseline
    ```bash
    dotnet run --project .\eng\Tools\ApiChief\ApiChief.csproj .\artifacts\bin\Microsoft.Extensions.Caching.Hybrid\Debug\net9.0\Microsoft.Extensions.Caching.Hybrid.dll emit baseline > .\src\Libraries\Microsoft.Extensions.Caching.Hybrid\Microsoft.Extensions.Caching.Hybrid.json
    ```

/cc: @joperezr

mgravell:
(rebased against dev)

eng/Versions.props
```diff
@@ -30,6 +30,7 @@
     <!-- Packages from dotnet/runtime -->
     <MicrosoftBclTimeProviderVersion>9.0.0-rc.1.24419.2</MicrosoftBclTimeProviderVersion>
     <MicrosoftExtensionsCachingAbstractionsVersion>9.0.0-rc.1.24419.2</MicrosoftExtensionsCachingAbstractionsVersion>
+    <MicrosoftExtensionsCachingMemoryVersion>9.0.0-rc.1.24419.2</MicrosoftExtensionsCachingMemoryVersion>
```
joperezr, line 33: Can we add these to Version.Details.xml as well in order to onboard into dependency flow so that these versions being added here are automatically updated each time there is a new aspnetcore build? Same goes for the new property bellow. If you have questions on how to do it I'm happy to help.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/README.md
```diff
@@ -0,0 +1 @@
+﻿Full `HybridCache` documentation is [here](https://learn.microsoft.com/aspnet/core/performance/caching/hybrid).
```
joperezr, line 1: We do try to enforce that all packages produced out of this repo have a README that more or less follows the same format (brief description of the package intent, Instructions for installing the package, Usage Examples). You can see an example of this here: https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.Common/README.md. They don't have to be very exahustive, but it would be good to have at least a very basic README here with at least some of those details as opposed to just pointing to some other learn page. The main idea of this README is: "What would a consumer **NEED TO KNOW** in the first 5 minutes of consuming the package?" That means it is totally fine to still have more detailed documentation elsewhere and we specifically don't want this README to be big, but it would be good to follow pattern if possible.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
+    <InjectIsExternalInitOnLegacy>true</InjectIsExternalInitOnLegacy>
+    <InjectCallerAttributesOnLegacy>true</InjectCallerAttributesOnLegacy>
+    <InjectTrimAttributesOnLegacy>true</InjectTrimAttributesOnLegacy>
+    <InjectExperimentalAttributeOnLegacy>true</InjectExperimentalAttributeOnLegacy>
+    <InjectSharedDiagnosticIds>true</InjectSharedDiagnosticIds>
+    <IsPackable>true</IsPackable>
+
+    <!-- ignore "use [Experimental] for new API" -->
+    <NoWarn>$(NoWarn),LA0003</NoWarn>
```
joperezr, line 17: Instead of this, can we just add a json file on this same directory with the existing API marked as stable? That would silence the analyzer. Those files can be easily generated using this tool in the repo: https://github.com/dotnet/extensions/tree/main/eng/Tools/ApiChief which you can either run manually, or alternatively you can run the script that invokes it: https://github.com/dotnet/extensions/blob/main/scripts/MakeApiBaselines.ps1

RussKie:
To confirm - is the decision to ship this package and the API within as "stable/prod-ready"?

joperezr:
> To confirm - is the decision to ship this package and the API within as "stable/prod-ready"?

If the intention is whether or not the APIs should be experimental, I don't think it is needed, I would leave that up to @mgravell's discretion. The intent of that analyzer requiring you to mark the APIs as experimental the first time you introduce them is to make sure that you at least release them once and let customers use it before you mark them as stable. In this library's case, it's API has already been shipped as preview multiple times and people has shared feedback, so I don't think that is a concern. Also, this is in the dev branch which means that even if this gets merged now, it would still only ship as preview.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
mgravell: reordering .ctor first

mgravell:
Suggestions integrated; code remains as per vanilla "relocate from aspnetcore main"

RussKie:
> @RussKie @joperezr am I good to merge? (I have permissions - I just don't want to stomp if that's not your preferred policy)

Yes (squash-merge).


But since I'm already here, I'll save you the troubles and merge it now :)

PR #5407, opened 2024-09-07
ResourceUtilizationHealthCheck - Report both CPU and Memory issues

### LINKED ISSUE #4677: ResourceUtilizationHealthCheck only reports CPU _or_ memory issues, not both
### Description

When checking health based on resource utilization, the conditions are checked in series and only the first failure is reported.
https://github.com/dotnet/extensions/blob/68810ca7399be38b8490a114ef52fdae4ce00058/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs#L39-L60

### Reproduction Steps

Test on a system with both high CPU _and_ memory.

### Expected behavior

Both CPU and memory issues should be reported.

### Actual behavior

Only the first issue found is reported.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

8.0

### Other information

Recommendation: ResourceUtilizationHealthCheck should be split into two IHealthCheck implementations, one for CPU and one for memory. That way both states can be reported independently.

willibrandon:
Fixes #4677 

## Proposed changes

- Report both CPU and Memory issues when checking health based on resource utilization.

## Customer Impact

- Fixes an issue where only CPU or Memory issues are reported during a resource utilization health check, but not both.
- This change also includes the additional key-value pairs describing the health of each component in the health check result.

## Test methodology 

- Updated `ResourceHealthCheckTests` to accommodate for this change.
- Ran all tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5407)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Diagnostics.ResourceMonitoring;
@@ -14,7 +15,6 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 /// </summary>
 internal sealed class ResourceUtilizationHealthCheck : IHealthCheck
 {
-    private static readonly Task<HealthCheckResult> _healthy = Task.FromResult(HealthCheckResult.Healthy());
     private readonly ResourceUtilizationHealthCheckOptions _options;
     private readonly IResourceMonitor _dataTracker;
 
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
```diff
@@ -39,26 +39,58 @@ public ResourceUtilizationHealthCheck(IOptions<ResourceUtilizationHealthCheckOpt
...
+            else if (cpuUnhealthy)
+            {
+                message = "CPU";
+            }
+            else
+            {
+                message = "Memory";
+            }
+
+            message += " usage is above the limit";
```
RussKie, line 67: I suggest creating dedicated strings for each case to avoid extra allocations.

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
```diff
@@ -39,26 +39,58 @@ public ResourceUtilizationHealthCheck(IOptions<ResourceUtilizationHealthCheckOpt
...
+        bool memoryUnhealthy = utilization.MemoryUsedPercentage > _options.MemoryThresholds.UnhealthyUtilizationPercentage;
 
-        if (utilization.CpuUsedPercentage > _options.CpuThresholds.DegradedUtilizationPercentage)
+        if (cpuUnhealthy || memoryUnhealthy)
         {
-            return Task.FromResult(HealthCheckResult.Degraded("CPU usage is close to the limit"));
+            string message = string.Empty;
+            if (cpuUnhealthy && memoryUnhealthy)
+            {
+                message = "CPU and Memory";
```
RussKie, line 56:
```suggestion
                message = "CPU and memory";
```

Ditto below

willibrandon:
@RussKie thanks for the feedback; commit with suggestions pushed.

RussKie (approved):

RussKie:
@evgenyfedorov2 @sebastienros

evgenyfedorov2 (approved):

RussKie:
Thank you

PR #5413, opened 2024-09-10
Register internal ResourceMonitor with ResourceUtilizationHealthCheck

### LINKED ISSUE #4560: The health checks resource utilization library assumes `IResourceMonitor` is registered.
### Description

The `Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization` assumes that the `IResourceMonitor` service has been registered, it's my belief that if this package relies on the resource monitoring services:

https://github.com/dotnet/extensions/blob/6fda2c8b8f8bacf74eaafff243a10ea20f849aaa/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization.csproj#L20

It should be responsible for ensuring that the `IResourceMonitor` is added to the `IServiceCollection`.

### Reproduction Steps

Create a simple console app:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
    <LangVersion>preview</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization" Version="8.0.0-rc.2.23510.2" />
    <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0-rc.2.23479.6" />
  </ItemGroup>

</Project>
```

Update the _Program.cs_ with the following:

```cs
using Microsoft.Extensions.Diagnostics.HealthChecks;

var builder = Host.CreateApplicationBuilder(args);

var healthChecksBuilder = builder.Services.AddHealthChecks()
    .AddResourceUtilizationHealthCheck();

var app = builder.Build();

var healthCheckService = app.Services.GetRequiredService<HealthCheckService>();

var result = await healthCheckService.CheckHealthAsync();

_ = result;

app.Run();
```

This results in the following:

image

### Expected behavior

I'd expect this to run without exception, as calling `AddResourceUtilizationHealthCheck` would add the required services.

### Actual behavior

A runtime exception is thrown.

### Regression?

_No response_

### Known Workarounds

Manually call `AddResourceMonitoring`.

### Configuration

_No response_

### Other information

_No response_

evgenyfedorov2:
Fixes #4560 

## What is changing

`ResourceUtilizationHealthCheck` depends on `IResourceMonitor` and requires an implementation of it to be registered first, which is also mentioned in #4560. Typically, users would achieve this by calling the `AddResourceMonitoring()` method which registers the internal [ResourceMonitorService](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitorService.cs) for this purpose:
```
var serviceCollection = new ServiceCollection();
serviceCollection.AddResourceMonitoring();
serviceCollection.AddResourceUtilizationHealthCheck();
```
After this change, users don't have to call `AddResourceMonitoring()` and can just call `AddResourceUtilizationHealthCheck()` immediately:
```
var serviceCollection = new ServiceCollection();
serviceCollection.AddResourceUtilizationHealthCheck();
```
## Impact

- If you use `AddResourceMonitoring()` method call, you may omit it.
- If you use a custom implementation of `IResourceMonitor` and register it using the `AddSingleton<IResourceMonitor, TImplementation>()` method, then no action is required. 
- If you  use a custom implementation of `IResourceMonitor` and register it using the `TryAddSingleton<IResourceMonitor, TImplementation>()` method which is placed before `AddResourceUtilizationHealthCheck()`, then no action is required. For example, assuming the custom implementation `TImplementation` class name is `MyResourceMonitorService`, the following scenarios will continue to work as is:
	
	:white_check_mark: works:
	```
	// 1.
	var serviceCollection = new ServiceCollection();
	serviceCollection.AddSingleton<IResourceMonitor, MyResourceMonitorService>();
	serviceCollection.AddResourceUtilizationHealthCheck();
	```
	
	:white_check_mark: works:
	 ```
	// 2.
	var serviceCollection = new ServiceCollection();
	serviceCollection.AddResourceUtilizationHealthCheck();
	serviceCollection.AddSingleton<IResourceMonitor, MyResourceMonitorService>();
	```
	
	:white_check_mark: works:
	 ```
	// 3.
	var serviceCollection = new ServiceCollection();
	serviceCollection.TryAddSingleton<IResourceMonitor, MyResourceMonitorService>();
	serviceCollection.AddResourceUtilizationHealthCheck();
	```

- If you use `TryAddSingleton<IResourceMonitor, TImplementation>()` which is placed after the `AddResourceUtilizationHealthCheck()` method call, then the following will not work, because the internal [ResourceMonitorService](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitorService.cs) will be registered and used instead of the `MyResourceMonitorService`:

	:x: does not work:
	```
	// 4.
	var serviceCollection = new ServiceCollection();
	serviceCollection.AddResourceUtilizationHealthCheck();
	serviceCollection.TryAddSingleton<IResourceMonitor, MyResourceMonitorService>();
	```
	
	To fix that, please use either `AddSingleton<IResourceMonitor, TImplementation>()` extension methods, or place the registration before the `AddResourceUtilizationHealthCheck()` method call:
	
	:white_check_mark: works:
	```
	// 5.
	var serviceCollection = new ServiceCollection();
	serviceCollection.TryAddSingleton<IResourceMonitor, MyResourceMonitorService>();
	serviceCollection.AddResourceUtilizationHealthCheck();
	```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5413)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheckExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheckExtensions.cs
@@ -28,6 +28,8 @@ public static class ResourceUtilizationHealthCheckExtensions
         _ = Throw.IfNull(builder);
         _ = Throw.IfNull(tags);
 
+        _ = builder.Services.AddResourceMonitoring();
+
         _ = builder.Services.AddOptionsWithValidateOnStart<ResourceUtilizationHealthCheckOptions, ResourceUtilizationHealthCheckOptionsValidator>();
         return builder.AddCheck<ResourceUtilizationHealthCheck>(HealthCheckName, tags: tags);
     }
@@ -43,6 +45,8 @@ public static class ResourceUtilizationHealthCheckExtensions
     {
         _ = Throw.IfNull(builder);
         _ = Throw.IfNull(tags);
+
+        _ = builder.Services.AddResourceMonitoring();
 
         _ = builder.Services.AddOptionsWithValidateOnStart<ResourceUtilizationHealthCheckOptions, ResourceUtilizationHealthCheckOptionsValidator>();
         return builder.AddCheck<ResourceUtilizationHealthCheck>(HealthCheckName, tags: tags);
--- a/test/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization.Tests/ResourceHealthCheckExtensionsTests.cs
+++ b/test/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization.Tests/ResourceHealthCheckExtensionsTests.cs
@@ -16,10 +16,10 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks.Test;
 public class ResourceHealthCheckExtensionsTests
 {
     [Fact]
-    public async Task Extensions_AddResourceHealthCheck()
+    public async Task AddResourceHealthCheck()
     {
         var dataTracker = new Mock<IResourceMonitor>();
-        var samplingWindow = TimeSpan.FromSeconds(1);
+        TimeSpan samplingWindow = TimeSpan.FromSeconds(1);
 
         var serviceCollection = new ServiceCollection();
         serviceCollection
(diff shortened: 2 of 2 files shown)


RussKie:
> * This change breaks customers who use `TryAddSingleton<IResourceMonitor>()` only when it is placed after the `AddResourceUtilizationHealthCheck()` method call. I don't think there is a way to avoid this.

How can we help customers to detect and avoid this? 
What will be the experience if customers do have the "incorrect" code? Will they be getting meaningful error messages? 
What kind of documentation and primers do we need for this?

evgenyfedorov2:
> > * This change breaks customers who use `TryAddSingleton<IResourceMonitor>()` only when it is placed after the `AddResourceUtilizationHealthCheck()` method call. I don't think there is a way to avoid this.
> 
> How can we help customers to detect and avoid this?

We will update our documentation at https://learn.microsoft.com/en-us/dotnet/core/diagnostics/diagnostic-health-checks


>What will be the experience if customers do have the "incorrect" code? Will they be getting meaningful error messages?

The library will silently use the built-in `ResourceMonitoringService` instead of the custom one provided by customers. Also, I don't see any way to avoid this because we cannot hook into the `IServiceProvider` right before the build started and check if a custom implementation was registered or not.

>What kind of documentation and primers do we need for this?

Update the docs as mentioned above. Add a note about the order of registration.

RussKie (approved):

RussKie:
Please link the pull-request updating the docs here as well.

evgenyfedorov2:
> Please link the pull-request updating the docs here as well.

https://github.com/dotnet/docs/pull/42626

evgenyfedorov2:
see also https://github.com/dotnet/docs/issues/42853

Issue #5411, opened 2024-09-09
[LogProperties] Generated Code Incorrectly Calls ToString() on Nullable-Typed Properties (e.g. DateTime?)


rwoll:
### Description

When using the `[LogProperties]` on an object with nullable-typed properties (e.g. `DateTime?`), the nullable-typed properties are being forced to `string` before being added to the `state.TagArray` while their non-nullable type (`DateTime`) remains `DateTime`-typed in the `TagArray`.

### Reproduction Steps

Given the following user-written code:

``` csharp
using Microsoft.Extensions.Logging;

public partial class MyLogger
{
  [LoggerMessage(
    EventId = 1,
    Level = LogLevel.Information,
    Message = "Event triggered!"
  )]
  public static partial void MyLogMessage(ILogger logger, [LogProperties] MyEvent myEvent);

  public class MyEvent
  {
    public DateTime TimeStamp { get; set; }
    public DateTime? TimeStampNullable { get; set; }
  }
}
```

and `csproj`:

``` xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
    <CompilerGeneratedFilesOutputPath>Generated</CompilerGeneratedFilesOutputPath>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.1" />
    <PackageReference Include="Microsoft.Extensions.Telemetry.Abstractions" Version="8.9.0" />
  </ItemGroup>
</Project>

```

### Expected behavior

Both `myEvent.TimeStamp` and `myEvent.TimeStampNullable` get added to the `state.TagArray` as `DateTime`s in the auto-generated logging code.

### Actual behavior

In the autogenerated code, `myEvent.TimeStamp` field gets added to the `TagArray` as a `DateTime` (same type as it was in the event), but `myEvent.TimeStampNullable` is added to the `TagArray` as a `string`:

``` csharp
        state.TagArray[2] = new("myEvent.TimeStamp", myEvent?.TimeStamp);
[...]

PR #5416, opened 2024-09-11
fix: [LogProperties] avoid early ToString of Nullable Properties

rwoll:
Fixes dotnet/extensions#5411.

Prior to this PR, when using the `[LogProperties]` on an object with nullable-typed properties (e.g. `DateTime?`), the nullable-typed properties were being forced to `string` before being added to the `state.TagArray` while their non-nullable type (`DateTime`) remained `DateTime`-typed in the `TagArray`. In our environment, this manifested as timestamp columns having different formats in logs!

Looking at the issue from a code perspective, given the following user-written code:

``` csharp

public partial class MyLogger
{
  [LoggerMessage(
    EventId = 1,
    Level = LogLevel.Information,
    Message = "Event triggered!"
  )]
  public static partial void MyLogMessage(ILogger logger, [LogProperties] MyEvent myEvent);

  public class MyEvent
  {
    public DateTime TimeStamp { get; set; }
    public DateTime? TimeStampNullable { get; set; }
  }
}
```

In the autogenerated code, `myEvent.TimeStamp` field got added to the `TagArray` as a `DateTime` (same type as it was in the event), but `myEvent.TimeStampNullable` is added to the `TagArray` as a `string`:

``` csharp
        state.TagArray[2] = new("myEvent.TimeStamp", myEvent?.TimeStamp);
        state.TagArray[1] = new("myEvent.TimeStampNullable", myEvent?.TimeStampNullable?.ToString()); // <--- LINE WITH BUG
```

This PR:

1. Adds a test that originally repro'd the broken behavior.
2. Introduces a potential fix.

The root of the issue appears to have been in the determination of `ImplementsISpanFormattable`.

In the `ShouldStringify*` methods, `p` for `System.Nullable<DateTime>` was not passing the `p.ImplementsISpanFormattable` check:

```
       static bool ShouldStringifyParameter(LoggingMethodParameter p)
        {
            if (p.IsReference)
            {
                // pass object as-is
                return false;
            }

            if (p.ImplementsISpanFormattable)
            {
                // pass object as-is, it will be formatted directly into the output buffer
                return false;
            }

            if (!p.Type.Contains("."))
            {
                // no . means this is a primitive type, pass as-is 
                return false;
            }

            // should convert the object to a string before calling the Log function
            return true;
        }
```

This was b/c the computation of that property was not being computed on the type wrapped by `Nullable`.

**Suggested Followup**

While the original bug manifested as string formatting discrepancies, a test asserting the string result of the log is insufficient to robustly test. Asserting the generated output is helpful (as we've also added here), but may easily be incorrectly re-baselined.

To help create a more robust test in the future, I recommend:

1. Adding a test against the `Type` of each value in `state.TagArray`
2. Adding tests directly for `TypeSymbolExtensions.{ImplementsIConvertible,ImplementsIFormattable,ImplementsISpanFormattable` and the newly introduced helper, `TypeSymbolExtensions.GetPossiblyNullWrappedType`.


src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogRecord.cs
```diff
@@ -147,4 +153,6 @@ internal static string Formatter(FakeLogRecord record)
 
         return $"[{record.Timestamp:mm:ss.fff}, {level}] {record.Message}";
     }
+
+    public readonly record struct AnnotatedStateValue(Type? type, string? value);
```
RussKie, line 157: Move to own file

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogger.cs
```diff
@@ -66,8 +66,8 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
 
         var l = new List<object?>();
         ScopeProvider.ForEachScope((scopeState, list) => list.Add(scopeState), l);
-
-        var record = new FakeLogRecord(logLevel, eventId, ConsumeTState(state), exception, formatter(state, exception),
+        var (stateTypes, consumedState) = ConsumeTState(state);
```
RussKie, line 69: As per our coding style guide, we don't use `var` unless the type is obvious.
https://github.com/dotnet/extensions/blob/main/CONTRIBUTING.md#dos-and-donts

test/Generators/Microsoft.Gen.Logging/Generated/LogPropertiesTests.cs
```diff
@@ -313,6 +313,27 @@ public void LogPropertiesInTemplateTest()
...
+        Assert.Equal($"Testing nullable property within class here...", _logger.Collector.LatestRecord.Message);
+
+        var expectedState = new Dictionary<string, FakeLogRecord.AnnotatedStateValue>
+        {
+            ["classWithNullablePropertyParam.NullableDateTime"] = new FakeLogRecord.AnnotatedStateValue(typeof(DateTime), now.ToString(CultureInfo.InvariantCulture)),
+            ["classWithNullablePropertyParam.NonNullableDateTime"] = new FakeLogRecord.AnnotatedStateValue(typeof(DateTime), now.ToString(CultureInfo.InvariantCulture)),
+            ["{OriginalFormat}"] = new FakeLogRecord.AnnotatedStateValue(typeof(string), "Testing nullable property within class here...")
+        };
+
+        _logger.Collector.LatestRecord.AnnotatedState.Should().NotBeNull().And.Equal(expectedState);
```
RussKie, line 334: Instead of this, can we use the VerifyXunit snapshot verification?
```suggestion
        await Verifier.Verify(_logger.Collector.LatestRecord.AnnotatedState);
```

RussKie:
> @RussKie - I've addressed a few of the comments and errors. However, I'm a bit stuck:
> 
> * `src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogRecord.cs(70,20): error LA0003: (NETCORE_ENGINEERING_TELEMETRY=Build) Newly added symbol 'Microsoft.Extensions.Logging.Testing.FakeLogRecord.StateTypes' must be marked as experimental` - what's the proper fix for this? The sybmol is in a test file, why does it need to be marked experimental?

You're adding a new public API, which is not allowed without an API proposal/review/etc.
Do these new API need to be public?

> * Re-baselining the snapshots - how do I re-baseline the failing snapshot tests across all platforms?

You need to run the tests locally and accept the new snapshot (via the diff/merge tool or by manually renaming files).
For these tests the snapshots must be identical across all platforms.

> 
> Full disclosure: I'm reaching the end of my allocated time to work on this issue so depending on how much further work is required, I may not be able to finish this PR off. However, I'm happy for your team to take it over when you have time, or — if it's easier — close the PR, and then y'all can create a fresh one when it comes time (either re-using bits here or taking a fresh pass).

Noted. @dariusclay please prioritise this review.

RussKie:
> OH, I didn't realize the `Testing.*` packages are actually shipped, but now see that they are ([learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.testing.fakelogrecord?view=net-8.0](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.testing.fakelogrecord?view=net-8.0)). Neat! And especially neat that the analyzers alert on it!

Yes, this is a public package that helps simplifying telemetry-related testing --> https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Diagnostics.Testing.
Essentially, everything located under https://github.com/dotnet/extensions/tree/main/src/Libraries folder is shipped to public.
Our tests are located under the https://github.com/dotnet/extensions/tree/main/test folder.

> > Do these new API need to be public?
> 
> No, these bits don't need to be public. And if they do, I'm not a fan of the API I've introduced. It was simply a "bolt-on" to repro/test. Not intended to ship to production. If you or @dariusclay think we need to ship to production, then I'd like to brainstorm a cleaner API. Even if they are changed to be internal-implementation only, the implementation would need to be cleaned up from what I have it at this iteration.


So, if those changes aren't required - please revert those and only keep the necessary changes.

dariusclay (approved):

dariusclay:
@RussKie everything looks good here to me. We can follow up after this merge to include additional tests.

RussKie:
Thank you @rwoll.

Thank you @dariusclay. Could you please raise a tracking issue for follow up on the outstanding work?

PR #5437, opened 2024-09-20
Fix test warning in 9.0

eerhardt:
Test methods should not call ConfigureAwait(false), as it may bypass parallelization limits. Omit ConfigureAwait, or use ConfigureAwait(true) to avoid CA2007. (https://xunit.net/xunit.analyzers/rules/xUnit1030)

This is blocking #5435 

See https://github.com/dotnet/extensions/pull/5435/checks?check_run_id=30404388453

cc @rwoll
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5437)

### DIFF
--- a/test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
+++ b/test/Generators/Microsoft.Gen.Logging/Unit/ParserTests.LogProperties.cs
@@ -482,8 +482,7 @@ public partial class ParserTests
                 Assembly.GetAssembly(typeof(DateTime))!,
             },
             [source],
-            symbols)
-            .ConfigureAwait(false);
+            symbols);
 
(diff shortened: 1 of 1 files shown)


dariusclay (approved):
Thanks for the catch! Nowhere else are we using `.ConfigureAwait(false)`

eerhardt:
Looks like that got removed in the `dev` branch with

https://github.com/dotnet/extensions/commit/62c74d39f2510fed97166c9439a4fafff04fcdd7#diff-c8046fc49e1b5a2d71a80ae4566b8a4567b68adef5f6b5a8d0f75026341d15ff

cc @joperezr

Issue #5483, opened 2024-10-08
Address trimmer warning on Microsoft.Extensions.AI due to JSON source generator


stephentoub:
https://github.com/dotnet/extensions/blob/a7e4f3b402620610eaa02b32978efe229bff32e6/src/Libraries/Microsoft.Extensions.AI/Microsoft.Extensions.AI.csproj#L21-L24

It's hitting https://github.com/dotnet/runtime/issues/108682.

cc: @eiriktsarpalis

PR #5492, opened 2024-10-09
Mark the FunctionCall/ResultContent.Exception properties as [JsonIgnore]

stephentoub:
Given implementation details of the JSON source generator today, even with the converter applied to these properties, code is still being generated for Exception, leading to unsuppressable trimmer warnings.

Closes https://github.com/dotnet/extensions/issues/5483
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5492)


test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Contents/FunctionCallContentTests..cs
```diff
@@ -84,48 +81,6 @@ public void Constructor_PropsRoundtrip()
         Assert.Same(e, c.Exception);
     }
 
-    [Fact]
```
eiriktsarpalis, line 87: Should this be replaced with a test checking that the roundtripped value returns null? Or do we not need to specify this?

stephentoub: Sure

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionResultContent.cs
```diff
@@ -20,10 +20,31 @@ public sealed class FunctionResultContent : AIContent
...
+    /// </summary>
+    /// <param name="callId">The function call ID for which this is the result.</param>
+    /// <param name="name">The function name that produced the result.</param>
+    /// <param name="result">
+    /// This may be <see langword="null"/> if the function returned <see langword="null"/>, if the function was void-returning
+    /// and thus had no result, or if the function call failed. Typically, however, in order to provide meaningfully representative
+    /// information to an AI service, a human-readable representation of those conditions should be supplied.
+    /// </param>
+    /// <param name="exception">Any exception that occurred when invoking the function.</param>
+    public FunctionResultContent(string callId, string name, object? result, Exception? exception)
```
eiriktsarpalis, line 47: Why do we need to expose a ctor overload accepting Exception given that the property is settable?

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionResultContent.cs
stephentoub: Trying to avoid breaking body changes right now if they're not necessary.

eiriktsarpalis: Sure, but if we can't make them now it will be even more difficult to make them later.

stephentoub: I'd rather take any such changes in batch when we have a meaningful set, so a more thorough review based on received feedback, etc. Force people to rev fewer times.

eiriktsarpalis: Will we remember/have enough context to remove this constructor when such a review is going to be made? Unless we release new bits to customers on a very frequent basis I don't see how postponing such a decision would impact churn substantially.

stephentoub: Each one of these is a binary breaking change. If every single release makes such a change but is binary breaking, it effectively means that the ecosystem gets reset with every single release.

eiriktsarpalis: I understand, but at the same time this seems fair game or even expected when evolving previews.

stephentoub: Removing this constructor is not a meaningful fix. It doesn't improve what someone is able to do. All it does is force someone to use a different syntax to construct it. Why would we want to reset the ecosystem for such a change? When there are other binary breaking changes we decide to make in preview, then we can make this at the same time, but I don't see the value in just making this one. And shipping will be frequent enough that it very well could be the only one.

PR #5504, opened 2024-10-10
Sanitize structured output schema type name to satisfy restrictions

### LINKED ISSUE #5501: Invalid response_format.json_schema.name when structured output type is generic
### Description

Given a record like `public record Movie(string Title, int Year, ...);`, when invoking the client with `var movies = await client.CompleteAsync<Movie[]>(...)` we get an exception:

```
Unhandled exception. System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: invalid_value)
Parameter: response_format.json_schema.name

Invalid 'response_format.json_schema.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.
   at OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
   at OpenAI.Chat.ChatClient.CompleteChatAsync(BinaryContent content, RequestOptions options)
   at OpenAI.Chat.ChatClient.CompleteChatAsync(IEnumerable`1 messages, ChatCompletionOptions options, CancellationToken cancellationToken)
   at Microsoft.Extensions.AI.OpenAIChatClient.CompleteAsync(IList`1 chatMessages, ChatOptions options, CancellationToken cancellationToken)
   at Microsoft.Extensions.AI.OpenTelemetryChatClient.CompleteAsync(IList`1 chatMessages, ChatOptions options, CancellationToken cancellationToken)
   at Microsoft.Extensions.AI.LoggingChatClient.CompleteAsync(IList`1 chatMessages, ChatOptions options, CancellationToken cancellationToken)
   at Microsoft.Extensions.AI.ChatClientStructuredOutputExtensions.CompleteAsync[T](IChatClient chatClient, IList`1 chatMessages, JsonSerializerOptions serializerOptions, ChatOptions options, Nullable`1 useNativeJsonSchema, CancellationToken cancellationToken)
   at Program.<Main>$(String[] args) in C:\Code\script\TypedChat.cs:line 39
   at Program.<Main>(String[] args)
```

### Reproduction Steps

```csharp
var services = new ServiceCollection()
    .AddChatClient(builder => builder
        .Use(new OpenAI.OpenAIClient(configuration["OpenAI:Key"]!).AsChatClient("gpt-4o-mini")))
    .BuildServiceProvider();

var movies = await client.CompleteAsync<Movie[]>("Top 5 movies from Christopher Nolan", useNativeJsonSchema: true);

Console.WriteLine(movies.Length);

public record Movie(string Title, int Year);
```

If you switch the type to be a wrapper type, things work, such as: `public record MoviesResult(Movie[] Movies);`.

Note that switching to a non-generic type such as `public class Movies : List<Movie> { }` also fails, but this time with:

```
Unhandled exception. System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: )
Parameter: response_format

Invalid schema for response_format 'Movies': schema must be a JSON Schema of 'type: "object"', got 'type: "array"'.
   at OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
   at OpenAI.Chat.ChatClient.CompleteChatAsync(BinaryContent content, RequestOptions options)
   at OpenAI.Chat.ChatClient.CompleteChatAsync(IEnumerable`1 messages, ChatCompletionOptions options, CancellationToken cancellationToken)
```

In this case, it's due to the fact that OpenAI expects the root schema element to be an object, and the JSON exporter infers it to be an array instead. This can be solved by detecting this situation automatically and using a wrapper type automatically and unwrapping before returning, such as 

```csharp
    public class Values<T>
    {
        public required T Data { get; set; }
    }
```

### Expected behavior

Typed chats work regardless of which T you are using: arrays, lists, dictionaries,etc. should all be supported and work as-is.

### Actual behavior

Multiple failure modes depending on what types you are specifying for the result, with many "gotchas" to learn as a consequence. User doesn't fall in the proverbial pit of success :) 

### Regression?

_No response_

### Known Workarounds

Use "Response" wrapper types always. Very annoying.

### Configuration

    <PackageReference Include="Microsoft.Extensions.AI" Version="9.0.0-preview.9.24507.7" />
    <PackageReference Include="Microsoft.Extensions.AI.OpenAI" Version="9.0.0-preview.9.24507.7" />

SDK: 9.0.100-rc.1.24452.12

### Other information

_No response_

kzu:
The underlying native schema support in OpenAI has specific requirements on valid schema names, as shown in the following exception when using either an array or any other generic type:

```
Unhandled exception. System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: invalid_value)
Parameter: response_format.json_schema.name

Invalid 'response_format.json_schema.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.
   at OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
```

This fix follows the approach used to sanitize function names, and sanitizes the schema name the same way.

Fixes #5501
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5504)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -11,6 +11,7 @@ using System.Text.Json.Nodes;
 using System.Text.Json.Schema;
 using System.Text.Json.Serialization;
 using System.Text.Json.Serialization.Metadata;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
@@ -167,7 +168,7 @@ public static partial class ChatClientStructuredOutputExtensions
             // the LLM backend is meant to do whatever's needed to explain the schema to the LLM.
             options.ResponseFormat = ChatResponseFormat.ForJsonSchema(
                 schema,
-                schemaName: typeof(T).Name,
+                schemaName: SanitizeMetadataName(typeof(T).Name),
                 schemaDescription: typeof(T).GetCustomAttribute<DescriptionAttribute>()?.Description);
         }
         else
@@ -224,4 +225,21 @@ public static partial class ChatClientStructuredOutputExtensions
     [JsonSerializable(typeof(JsonNode))]
     [JsonSourceGenerationOptions(WriteIndented = true)]
     private sealed partial class JsonNodeContext : JsonSerializerContext;
+
+    /// <summary>
+    /// Remove characters from type name that are valid in metadata but shouldn't be used in a schema name.
+    /// Removes arrays and generic type parameters, and replaces invalid characters with underscores.
+    /// </summary>
+    private static string SanitizeMetadataName(string typeName) =>
+        InvalidNameCharsRegex().Replace(typeName, "_");
+
+    /// <summary>Regex that flags any character other than ASCII digits or letters or the underscore.</summary>
+#if NET
+    [GeneratedRegex("[^0-9A-Za-z_]")]
+    private static partial Regex InvalidNameCharsRegex();
+#else
+    private static Regex InvalidNameCharsRegex() => _invalidNameCharsRegex;
+    private static readonly Regex _invalidNameCharsRegex = new("[^0-9A-Za-z_]", RegexOptions.Compiled);
+#endif
+
 }
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -11,10 +11,10 @@ using System.Text.Json.Nodes;
 using System.Text.Json.Schema;
 using System.Text.Json.Serialization;
 using System.Text.Json.Serialization.Metadata;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
+using static Microsoft.Extensions.AI.FunctionCallHelpers;
 
 namespace Microsoft.Extensions.AI;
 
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -224,4 +225,21 @@ private static JsonSerializerOptions GetOrCreateDefaultJsonSerializerOptions()
...
+        InvalidNameCharsRegex().Replace(typeName, "_");
+
+    /// <summary>Regex that flags any character other than ASCII digits or letters or the underscore.</summary>
+#if NET
+    [GeneratedRegex("[^0-9A-Za-z_]")]
+    private static partial Regex InvalidNameCharsRegex();
+#else
+    private static Regex InvalidNameCharsRegex() => _invalidNameCharsRegex;
+    private static readonly Regex _invalidNameCharsRegex = new("[^0-9A-Za-z_]", RegexOptions.Compiled);
+#endif
```
stephentoub, line 243: Thanks for the fix.

Could you move the equivalent logic that's currently in AIFunctionFactory to instead be on the internal FunctionCallHelpers class and then use that from both there and here (rather than duplicating the code)?  We may subsequently choose to expose some of the functionality publicly, too.

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/ChatClientStructuredOutputExtensionsTests.cs
```diff
@@ -172,6 +172,40 @@ public async Task CanUseNativeStructuredOutput()
...
+        var expectedResult = new Data<Animal> { Value = new Animal { Id = 1, FullName = "Tigger", Species = Species.Tiger } };
+        var expectedCompletion = new ChatCompletion([new ChatMessage(ChatRole.Assistant, JsonSerializer.Serialize(expectedResult))]);
+
+        using var client = new TestChatClient
+        {
+            CompleteAsyncCallback = (messages, options, cancellationToken) =>
+            {
+                var responseFormat = Assert.IsType<ChatResponseFormatJson>(options!.ResponseFormat);
+
+                Assert.Matches("^[a-zA-Z0-9_-]+$", responseFormat.SchemaName);
```
eiriktsarpalis, line 187: Nit: could you assert the actual name being returned?

kzu: Sure, but I thought it was better to make the test more resilient to .net metadata naming since the point is actually the regex sanitization. Perhaps a theory would make that more obvious? 

But I'll change it if you think that is better, sure.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
kzu: done!

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/ChatClientStructuredOutputExtensionsTests.cs
eiriktsarpalis: I think it would be valuable to verify that the output of the sanitization is something expected rather than any string that happens to be validated by the regex. I don't mind using a theory if you would like to test this against more inputs.

kzu: Done!

PR #5507, opened 2024-10-10
Use the same JsonSerializerOptions default in all locations.

eiriktsarpalis:
And removes all remaining RUC/RDC annotations from the codebase.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5507)


src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -23,17 +23,12 @@ namespace Microsoft.Extensions.AI;
...
-    internal const string UsesReflectionJsonSerializerMessage =
-        "This method uses the reflection-based JsonSerializer which can break in trimmed or AOT applications.";
-
     /// <summary>Lazily-initialized default options instance.</summary>
     private static AIFunctionFactoryCreateOptions? _defaultOptions;
 
     /// <summary>Creates an <see cref="AIFunction"/> instance for a method, specified via a delegate.</summary>
     /// <param name="method">The method to be represented via the created <see cref="AIFunction"/>.</param>
     /// <returns>The created <see cref="AIFunction"/> for invoking <paramref name="method"/>.</returns>
-    [RequiresUnreferencedCode(UsesReflectionJsonSerializerMessage)]
```
eiriktsarpalis, line 35: A number of these overloads are now redundant, but I defer to @stephentoub when and when they should be removed.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -44,16 +35,14 @@ public static partial class ChatClientStructuredOutputExtensions
...
-    [RequiresDynamicCode(UsesReflectionJsonSerializerMessage)]
     public static Task<ChatCompletion<T>> CompleteAsync<T>(
         this IChatClient chatClient,
         IList<ChatMessage> chatMessages,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
         CancellationToken cancellationToken = default)
         where T : class =>
-        CompleteAsync<T>(chatClient, chatMessages, DefaultJsonSerializerOptions, options, useNativeJsonSchema, cancellationToken);
+        CompleteAsync<T>(chatClient, chatMessages, JsonDefaults.Options, options, useNativeJsonSchema, cancellationToken);
```
SteveSandersonMS, line 45: I'm curious how this works without needing any annotations for AOT.

Does `JsonDefaults.Options.GetJsonSchemaAsNode` work with all possible user types in an AOT application?

eiriktsarpalis: No, just the ones defined in the context. The idea being that if a type is missing it will unambiguously throw `NotSupportedException` as opposed to entering the undefined behaviour of the reflection-based serializer.

eiriktsarpalis: > Or is it that in non-AOT app, `JsonDefaults.Options` enables the reflection-based serializer so it will work the same as before this change?

Exactly that. The constructor for `JsonDefault.Options` effectively checks if `PublishTrimmed` has been enabled for the application and either returns a reflection-based or source generated contract resolver:

https://github.com/dotnet/extensions/blob/9ea3e7614ca9bc48215aead3bd2d5cf6169309a0/src/Libraries/Microsoft.Extensions.AI/JsonDefaults.cs#L28-L44

This strategy reflects how aspnetcore configures its own `JsonSerializerOptions`

SteveSandersonMS: Awesome - sounds great. Thanks!

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: Correct.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -107,8 +98,8 @@ private sealed class ReflectionAIFunction : AIFunction
     {
         private readonly MethodInfo _method;
         private readonly object? _target;
-        private readonly Func<IReadOnlyDictionary<string, object?>, AIFunctionContext?, object?>[] _parameterMarshalers;
-        private readonly Func<object?, ValueTask<object?>> _returnMarshaler;
+        private readonly Func<IReadOnlyDictionary<string, object?>, AIFunctionContext?, object?>[] _parameterMarshallers;
+        private readonly Func<object?, ValueTask<object?>> _returnMarshaller;
```
stephentoub, line 102: We can change if you want, but single `l` is the US spelling.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryCreateOptions.cs
```diff
@@ -19,10 +18,8 @@ public sealed class AIFunctionFactoryCreateOptions
     /// <summary>
     /// Initializes a new instance of the <see cref="AIFunctionFactoryCreateOptions"/> class with default serializer options.
     /// </summary>
-    [RequiresUnreferencedCode(AIFunctionFactory.UsesReflectionJsonSerializerMessage)]
-    [RequiresDynamicCode(AIFunctionFactory.UsesReflectionJsonSerializerMessage)]
     public AIFunctionFactoryCreateOptions()
-        : this(JsonSerializerOptions.Default)
+        : this(JsonDefaults.Options)
```
stephentoub, line 22: Nice

src/Libraries/Microsoft.Extensions.AI/JsonDefaults.cs
```diff
@@ -30,9 +31,10 @@ private static JsonSerializerOptions CreateDefaultOptions()
             // Keep in sync with the JsonSourceGenerationOptions on JsonContext below.
             var options = new JsonSerializerOptions(JsonSerializerDefaults.Web)
             {
-                WriteIndented = true,
-                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
                 TypeInfoResolver = new DefaultJsonTypeInfoResolver(),
+                Converters = { new JsonStringEnumConverter() },
```
stephentoub, line 35: Are there any negative implications to this?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
stephentoub: I'd be ok if you wanted to consolidate them here. I'm more concerned about batching up breaking binary changes to the object model, as those are the things we'd expect nuget packages to be impacted by. AIFunctionFactory is less relevant there.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: Oh ok, I only changed it because my spell checker was complaining.

src/Libraries/Microsoft.Extensions.AI/JsonDefaults.cs
eiriktsarpalis: I can't think of any reason why one might want to serialize enums as integers.

PR #5508, opened 2024-10-10
Use the logging generator in LoggingChatClient / LoggingEmbeddingGenerator

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5508)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/LoggingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/LoggingChatClient.cs
@@ -10,13 +10,10 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Shared.Diagnostics;
 
-#pragma warning disable EA0000 // Use source generated logging methods for improved performance
-#pragma warning disable CA2254 // Template should be a static expression
-
 namespace Microsoft.Extensions.AI;
 
 /// <summary>A delegating chat client that logs chat operations to an <see cref="ILogger"/>.</summary>
-public class LoggingChatClient : DelegatingChatClient
+public partial class LoggingChatClient : DelegatingChatClient
 {
     /// <summary>An <see cref="ILogger"/> instance used for all logging.</summary>
     private readonly ILogger _logger;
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/LoggingChatClient.cs
```diff
@@ -103,52 +130,63 @@ public override async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteSt
...
-            }
-        }
-    }
+    private string AsJson<T>(T value) => JsonSerializer.Serialize(value, _jsonSerializerOptions.GetTypeInfo(typeof(T)));
+
+    [LoggerMessage(LogLevel.Debug, "{MethodName} invoked.")]
+    private partial void LogInvoked(string methodName);
+
+    [LoggerMessage(LogLevel.Trace, "{MethodName} invoked: {ChatMessages}. Options: {ChatOptions}. Metadata: {ChatClientMetadata}.")]
+    private partial void LogInvokedSensitive(string methodName, string chatMessages, string chatOptions, string chatClientMetadata);
```
gfoidl, line 173: As it's already checked if the `LogLevel` is enabled, there's no need for the SGen to generate that check too. So set
```suggestion
    [LoggerMessage(LogLevel.Trace, "{MethodName} invoked: {ChatMessages}. Options: {ChatOptions}. Metadata: {ChatClientMetadata}.", SkipEnabledCheck = true)]
    private partial void LogInvokedSensitive(string methodName, string chatMessages, string chatOptions, string chatClientMetadata);
```
in order to avoid the redundant check be emitted.

Same on other places.

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/LoggingChatClient.cs
stephentoub: I didn't do so because these are only Debug/Trace events, and in the case where such events are enabled, the extra interface call will be dwarfed by all the JSON serialization and other overheads. And while I could just enable it anyway, not all of the call sites are currently guarded, which makes it complicated for someone maintaining the code to know which call sites need guarding and which don't. Which then leads to consistently putting SkipEnabledCheck on all of the log methods and consistently guarding all call sites, but that in turn makes the code more expensive. Seemed best to just keep it simple and pay for the extra enabled check when there's already a ton of overhead / when debugging.

PR #5513, opened 2024-10-11
Expose an `AIJsonUtilities` class in M.E.AI and lower M.E.AI.Abstractions to STJv8

### LINKED ISSUE #5482: Cache JSON schema generated in CompleteAsync<T>
Currently each call to `CompleteAsync<T>` generates a schema:
https://github.com/dotnet/extensions/blob/a7e4f3b402620610eaa02b32978efe229bff32e6/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs#L137

This should be cached, e.g. via a ConditionalWeakTable associated with the corresponding JsonSerializer.

### LINKED ISSUE #5496: Expose `FunctionCallHelpers` as public API.
> To be clear, this should be reusing the same caching infrastructure used by the existing `InferSchema` helpers.
> 
> Only tangentially-related, you thought we should consider exposing some of that publicly, yes? If so, could you open an issue to track? Thanks. 

 _Originally posted by @stephentoub in [#5482](https://github.com/dotnet/extensions/issues/5482#issuecomment-2403350491)_

eiriktsarpalis:
Renames `FunctionCallHelpers` to `AIJsonUtilities` and exposes the class as public.

Fix #5496. Fix #5482.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5513)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
```diff
@@ -14,40 +15,68 @@
...
+    /// </returns>
+    public static string SanitizeMemberName(string memberName) =>
+        InvalidNameCharsRegex().Replace(memberName, "_");
 
     /// <summary>Parses a JSON object into a dictionary of objects encoded as <see cref="JsonElement"/>.</summary>
     /// <param name="json">A JSON object containing the parameters.</param>
     /// <param name="parsingException">If the parsing fails, the resulting exception.</param>
     /// <returns>The parsed dictionary of objects encoded as <see cref="JsonElement"/>.</returns>
-    public static Dictionary<string, object?>? ParseFunctionCallArguments(string json, out Exception? parsingException)
+    public static Dictionary<string, object?>? ParseFunctionCallArguments([StringSyntax("json")] string json, out Exception? parsingException)
```
stephentoub, line 72:
```suggestion
    public static Dictionary<string, object?>? ParseFunctionCallArguments([StringSyntax(StringSyntaxAttribute.Json)] string json, out Exception? parsingException)
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
```
eiriktsarpalis, line 9: This class shouldn't really be in the "Primitives" assembly, but there's really no other place it could be kept given the current project structure.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
```diff
@@ -80,10 +109,10 @@ internal static partial class FunctionCallHelpers
     /// <param name="parameters">The dictionary of parameters.</param>
     /// <param name="options">A <see cref="JsonSerializerOptions"/> governing serialization.</param>
     /// <returns>A JSON encoding of the parameters.</returns>
-    public static string FormatFunctionParametersAsJson(IDictionary<string, object?>? parameters, JsonSerializerOptions? options = null)
+    public static string FormatFunctionParametersAsJsonString(IDictionary<string, object?>? parameters, JsonSerializerOptions? options = null)
```
stephentoub, line 112: Does this need to be public? It's really just a call to JsonSerializer.Serialize.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
```diff
@@ -97,7 +126,7 @@ public static string FormatFunctionParametersAsJson(IDictionary<string, object?>
     public static JsonElement FormatFunctionParametersAsJsonElement(IDictionary<string, object?>? parameters, JsonSerializerOptions? options = null)
```
stephentoub, line 126: Ditto

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
```diff
@@ -108,10 +137,10 @@ public static JsonElement FormatFunctionParametersAsJsonElement(IDictionary<stri
     /// <param name="result">The result value to be serialized.</param>
     /// <param name="options">A <see cref="JsonSerializerOptions"/> governing serialization.</param>
     /// <returns>A JSON encoding of the parameter.</returns>
-    public static string FormatFunctionResultAsJson(object? result, JsonSerializerOptions? options = null)
+    public static string FormatFunctionResultAsJsonString(object? result, JsonSerializerOptions? options = null)
```
stephentoub, line 140: Same for these

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
```diff
@@ -136,12 +165,19 @@ public static JsonElement FormatFunctionResultAsJsonElement(object? result, Json
...
+    /// <param name="includeTypeInEnumSchemas">Whether to include the type keyword in schemas for enum types.</param>
+    /// <param name="disallowAdditionalProperties">Whether to emit an additionalProperties keyword set to false.</param>
     /// <returns>A JSON schema document encoded as a <see cref="JsonElement"/>.</returns>
     public static JsonElement InferParameterJsonSchema(
         AIFunctionParameterMetadata parameterMetadata,
         AIFunctionMetadata functionMetadata,
-        JsonSerializerOptions? options)
+        JsonSerializerOptions? options,
+        bool includeTypeInEnumSchemas = false,
+        bool disallowAdditionalProperties = false)
```
stephentoub, line 176: What's the liklihood we'll want to expose more configuration here? Wondering if this should instead be an options bag, or a flags enum.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
eiriktsarpalis: They could be removed. I suppose their benefit is that they give you ready access to a JsonSerializerContext containing `IDictionary<string, object?>` so that one doesn't need to be added in the service.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
eiriktsarpalis: It's something I've been thinking about. It's likely we'll need more of these in the future as we encounter more vendors imposing their own restrictions on accepted schemas.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
eiriktsarpalis: I should add that these particular helpers uphold the invariant required by `AIFunctionFactory`: that object types must be deserialized as `JsonElement`. This isn't always a given and we expose configuration flags in JSO that change this. In my view this is further evidence that the current shape of AIFunction/FunctionXContent is a leaky abstraction -- a non-polymorphic approach eliminates the need for such helpers altogether.

stephentoub: You were going to prototype the alternative you had in mind? It's hard to reason about possible deficiencies here without that.

eiriktsarpalis: I can try to revive my old prototype assuming there's enough buy-in from the rest of the team.

eiriktsarpalis: Removed the formatter helpers in https://github.com/dotnet/extensions/pull/5513/commits/514f846e359ca7abc283320e70c6ea6b17571537

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
eiriktsarpalis: Moved to a separate `JsonSchemaInferenceOptions` class.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallUtilities.cs
stephentoub: > I can try to revive my old prototype assuming there's enough buy-in from the rest of the team.

If it fixes more issues than it creates, we should certainly consider it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -14,40 +15,68 @@
...
+
+    /// <summary>
+    /// Removes characters from a .NET member name that shouldn't be used in an AI function name.
+    /// </summary>
+    /// <param name="memberName">The .NET member name that should be sanitized.</param>
+    /// <returns>
+    /// Replaces non-alphanumeric characters in the identifier with the underscore character.
+    /// Primarily intended to remove characters produced by compiler-generated method name mangling.
+    /// </returns>
+    public static string SanitizeMemberName(string memberName) =>
```
stephentoub, line 65: Nit: now that this is public, we should add a Throw.IfNull(memberName) in there.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -74,179 +103,141 @@ internal static partial class FunctionCallHelpers
...
     /// <param name="functionMetadata">The containing function metadata.</param>
-    /// <param name="options">The global <see cref="JsonSerializerOptions"/> governing serialization.</param>
+    /// <param name="serializerOptions">The global <see cref="JsonSerializerOptions"/> governing serialization.</param>
+    /// <param name="inferenceOptions">The options controlling schema inference.</param>
     /// <returns>A JSON schema document encoded as a <see cref="JsonElement"/>.</returns>
     public static JsonElement InferParameterJsonSchema(
         AIFunctionParameterMetadata parameterMetadata,
         AIFunctionMetadata functionMetadata,
-        JsonSerializerOptions? options)
+        JsonSerializerOptions? serializerOptions,
```
stephentoub, line 117: We don't want serializerOptions to be optional?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -74,179 +103,141 @@ internal static partial class FunctionCallHelpers
...
     /// <returns>A JSON schema document encoded as a <see cref="JsonElement"/>.</returns>
     public static JsonElement InferParameterJsonSchema(
         Type? type,
-        string name,
-        string? description,
-        bool hasDefaultValue,
-        object? defaultValue,
-        JsonSerializerOptions options)
+        string parameterName,
+        JsonSerializerOptions serializerOptions,
```
stephentoub, line 162: Here as well, I'd have expected serializerOptions to be optional and to be at the end of the parameter list. It feels strange to me coming between parameterName and other things about the parameter, like description and default value.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -74,179 +103,141 @@ internal static partial class FunctionCallHelpers
...
+            inferenceOptions.IncludeSchemaKeyword,
+            inferenceOptions.DisallowAdditionalProperties,
+            inferenceOptions.IncludeTypeInEnumSchemas);
+
+        return GetJsonSchemaCached(serializerOptions, key);
+    }
 
+    private static JsonElement GetJsonSchemaCached(JsonSerializerOptions options, FunctionParameterKey key)
+    {
         options.MakeReadOnly();
```
stephentoub, line 219: Do we need to document anything about using a JSO with these inference methods will make it read only? Or is that implied?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -269,13 +266,18 @@ private static JsonElement GetJsonSchemaCore(JsonSerializerOptions options, Func
...
                     : null;
 
-                schemaObj["default"] = defaultValueNode;
+                (schemaObj ??= [])["default"] = defaultValueNode;
             }
 
-            return JsonSerializer.SerializeToElement(schemaObj, FunctionCallHelperContext.Default.JsonNode);
+            return schemaObj is null
+                ? _trueJsonSchema
+                : JsonSerializer.SerializeToElement(schemaObj, FunctionCallUtilityContext.Default.JsonNode);
```
stephentoub, line 274: It's ok for this to use our built-in converter rather than using the supplied JSO?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -306,19 +312,34 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext ctx, JsonNode schema)
...
-                    : $"#/{PropertiesPropertyName}/{key.ParameterName}/{refUri.AsMemory("#/".Length)}";
-
-                refObj[RefPropertyName] = (JsonNode)refUri;
+                // The resulting schema might be a $ref using a pointer to a different location in the document.
+                // As JSON pointer doesn't support relative paths, parameter schemas need to fix up such paths
+                // to accommodate the fact that they're being nested inside of a higher-level schema.
+                if (key.ParameterName is not null && objSchema.TryGetPropertyValue(RefPropertyName, out JsonNode? paramName))
+                {
+                    // Fix up any $ref URIs to match the path from the root document.
+                    string refUri = paramName!.GetValue<string>();
```
stephentoub, line 323: TryGetPropertyValue doesn't guarantee a non-null node on success?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -327,14 +348,30 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext ctx, JsonNode schema)
...
+                    JsonNode descriptionNode = (JsonNode)key.Description!;
+                    int index = obj.IndexOf(DescriptionPropertyName);
+                    if (index < 0)
+                    {
+                        obj.Insert(0, DescriptionPropertyName, (JsonNode)key.Description!);
+                    }
+                    else
+                    {
+                        obj[index] = (JsonNode)key.Description!;
+                    }
```
stephentoub, line 361: There's no notion of just doing `obj[DescriptionPropertyName] = (JsonNode)key.Description!` and having it either update an existing property in place or add one?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonSchemaInferenceOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+/// </summary>
+public sealed class JsonSchemaInferenceOptions
+{
+    /// <summary>
+    /// Gets the default options instance.
+    /// </summary>
+    public static JsonSchemaInferenceOptions Default { get; } = new JsonSchemaInferenceOptions();
+
+    /// <summary>
+    /// Gets a value indicating whether to include the type keywork in inferred schemas for .NET enums.
```
stephentoub, line 17:
```suggestion
    /// Gets a value indicating whether to include the type keyword in inferred schemas for .NET enums.
```
Also, should `type` be `$type`?

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -356,20 +356,22 @@ private IEnumerable<OllamaChatRequestMessage> ToOllamaChatRequestMessages(ChatMe
...
                             Name = fcc.Name,
-                            Arguments = FunctionCallHelpers.FormatFunctionParametersAsJsonElement(fcc.Arguments, ToolCallJsonSerializerOptions),
+                            Arguments = JsonSerializer.SerializeToElement(fcc.Arguments, serializerOptions.GetTypeInfo(typeof(IDictionary<string, object?>))),
                         }, JsonContext.Default.OllamaFunctionCallContent)
                     };
                     break;
 
                 case FunctionResultContent frc:
-                    JsonElement jsonResult = FunctionCallHelpers.FormatFunctionResultAsJsonElement(frc.Result, ToolCallJsonSerializerOptions);
+                    JsonSerializerOptions serializerOptions1 = ToolCallJsonSerializerOptions ?? JsonContext.Default.Options;
```
stephentoub, line 373: You could surround the case bodies where variables are being defined with braces, and then you wouldn't need to use names like serializerOptions1.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonSchemaInferenceOptions.cs
eiriktsarpalis: It's just `type` in JSON schema world.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -120,26 +124,12 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
+            serializerOptions,
+            inferenceOptions: _inferenceOptions);
 
-                return node;
-            },
-        });
-        schemaNode.Insert(0, "$schema", "https://json-schema.org/draft/2020-12/schema");
-        schemaNode.Add("additionalProperties", false);
-        var schema = JsonSerializer.Serialize(schemaNode, JsonDefaults.Options.GetTypeInfo(typeof(JsonNode)));
+        var schema = JsonSerializer.Serialize(schemaNode, JsonDefaults.Options.GetTypeInfo(typeof(JsonElement)));
```
stephentoub, line 132: Similar question here... this doesn't need to use the supplied JSO?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -14,40 +15,68 @@
...
+    /// </returns>
+    public static string SanitizeMemberName(string memberName) =>
+        InvalidNameCharsRegex().Replace(memberName, "_");
 
     /// <summary>Parses a JSON object into a dictionary of objects encoded as <see cref="JsonElement"/>.</summary>
     /// <param name="json">A JSON object containing the parameters.</param>
     /// <param name="parsingException">If the parsing fails, the resulting exception.</param>
     /// <returns>The parsed dictionary of objects encoded as <see cref="JsonElement"/>.</returns>
-    public static Dictionary<string, object?>? ParseFunctionCallArguments(string json, out Exception? parsingException)
+    public static Dictionary<string, object?>? ParseFunctionCallArguments([StringSyntax(StringSyntaxAttribute.Json)] string json, out Exception? parsingException)
```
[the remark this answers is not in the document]
eiriktsarpalis: This is just following our usual practice of returning the most specific type. Any reason why we would want to abstract it away?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: This is checked by the regex method (added testing for that). I guessing you want this for cleaner stacktraces?

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
eiriktsarpalis: Can do. I figured I could get away with this to minimize the diff.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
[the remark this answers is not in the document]
stephentoub: > This is just following our usual practice of returning the most specific type. Any reason why we would want to abstract it away?

For future flexibility should we want to return a different kind of dictionary.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
stephentoub: > This is checked by the regex method (added testing for that). 

But that exception includes the wrong argument name.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: We can't make the JSO optional in this case since there's no `AIFunctionParameterMetadata` to tap into. We could of course use the same technique as `JsonDefaults` does, but that's defined in a separate assembly and it didn't make a lot of sense to try and duplicate in primitives as well. This overload is only intended for use by `AIFunctionFactory` or similar components but not chat service implementations the use the first overload.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: I would say that's generally speaking implied when trying to use a JSO for anything meaningful.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: It's ok, it's just there to convert between the two DOM types. Using the supplied JSO would introduce the implicit dependency for a `JsonNode` contract.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: It doesn't because JsonNode uses `null` to represent JSON nulls.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: It's the latter. I wanted to add a comment about this but then forgot.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
eiriktsarpalis: Good question. In this case the schema is being rendered as a string so one could argue that indentation concerns are significant. We could, although again this would be introducing an implicit dependency for `JsonNode` on the JSO which I would rather not be enforcing on the caller.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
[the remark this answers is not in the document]
eiriktsarpalis: Isn't the usual counterargument that we introduce an implicit dependency on the current implementation if it's public since users can do `result is Dictionary<string, object?>`?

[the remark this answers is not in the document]
stephentoub: We say folks can't rely on concrete type beyond what's in the signature, and we have changed it. We do not consider it to be a breaking change.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: Done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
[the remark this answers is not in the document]
eiriktsarpalis: I don't see a strong reason why we'd want to move away from `Dictionary<string, object?>`. This is a deserialization routine that assigns ownership of the dictionary to the caller, so unless the signature of `FunctionCallContent` changes dramatically I don't think this would need to change either.

One approach I briefly experimented with was changing the return type to be `FunctionalCallContent` and have the method take additional functionName and callId parameters. While this did clean the API signature a bit, it made setting additional properties a bit more clunky at the call sites. It would serve to eliminate the concern you're raising here though. Thoughts?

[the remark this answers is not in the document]
stephentoub: I think I like that better. Makes me wonder whether that helper should be on FunctionCallContent instead, but if we're trying to keep the JSON-ness separate, this makes sense.

eiriktsarpalis:
I believe I've addressed all feedback, this should be ready to merge.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -0,0 +1,448 @@
...
+    bool DisallowAdditionalProperties,
+    bool IncludeTypeInEnumSchemas);
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Provides a collection of static utility methods for marshalling JSON data in function calling.</summary>
+public static partial class JsonFunctionCallUtilities
+{
+    /// <summary>The uri used when populating the $schema keyword in inferred schemas.</summary>
+#pragma warning disable S1075 // URIs should not be hardcoded
```
stephentoub, line 41: I'd just put this at the top with the rest of the supressions.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
stephentoub: > The reason why I left it as required but nullable is that the AIFunctionMetadata might not have a JSO specified, in which case this is almost guaranteed to fail.

Why would it be guaranteed to fail? Couldn't we fall back to whatever default we otherwise use in other situations?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
stephentoub: > but that's defined in a separate assembly and it didn't make a lot of sense to try and duplicate in primitives as well.

Why not?

stephentoub: Should we expose from JsonFunctionCallUtilities the default JSO we use?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    private static JsonElement GetJsonSchemaCore(JsonSerializerOptions options, FunctionParameterKey key)
+    {
+        _ = Throw.IfNull(options);
+        options.MakeReadOnly();
+
+        if (options.ReferenceHandler == ReferenceHandler.Preserve)
+        {
+            throw new NotSupportedException("Schema generation not supported with ReferenceHandler.Preserve enabled.");
```
stephentoub, line 266: Can you add a comment about why this is the case?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -0,0 +1,448 @@
...
+    /// <summary>Infers a JSON schema for the specified type.</summary>
+    /// <param name="type">The type for which to generate the schema.</param>
+    /// <param name="serializerOptions">The options used to extract the schema from the specified type.</param>
+    /// <param name="description">The description of the parameter.</param>
+    /// <param name="hasDefaultValue">Whether the parameter is optional.</param>
+    /// <param name="defaultValue">The default value of the optional parameter, if applicable.</param>
+    /// <param name="inferenceOptions">The options controlling schema inference.</param>
+    /// <returns>A <see cref="JsonElement"/> representing the schema.</returns>
+    public static JsonElement InferJsonSchema(
+        Type? type,
```
stephentoub, line 216: Why do we allow a null Type?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: You added this :-) This is replicating a check in JsonSchemaExporter/JsonSchemaMapper so in many ways it's redundant. I'm interpreting this as an attempt to throw the same error regardless of whether the AI metadata is defining a type, but if not we can just remote it.

stephentoub: I didn't "add" it, I just copied it from https://github.com/eiriktsarpalis/stj-schema-mapper/blob/3cc062d1c4711fc83626fb59a5cf2d96cb76bc7c/src/JsonSchemaMapper/JsonSchemaMapper.cs#L410.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: `AIFunctionMetadata` allows it so this needs to allow it as well.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: Gotcha. We can remove it given that the check is made by `JsonSchemaExporter` as well.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -40,7 +39,7 @@ public static AIFunction Create(Delegate method, AIFunctionFactoryCreateOptions?
     /// <param name="method">The method to be represented via the created <see cref="AIFunction"/>.</param>
     /// <param name="name">The name to use for the <see cref="AIFunction"/>.</param>
     /// <param name="description">The description to use for the <see cref="AIFunction"/>.</param>
-    /// <param name="serializerOptions">The <see cref="JsonSerializerOptions"/> used to marshal function parameters.</param>
+    /// <param name="serializerOptions">The <see cref="JsonSerializerOptions"/> used to marshal function parameters and return value.</param>
```
stephentoub, line 42: Nit:
```suggestion
    /// <param name="serializerOptions">The <see cref="JsonSerializerOptions"/> used to marshal function parameters and any return value.</param>
```
or
```suggestion
    /// <param name="serializerOptions">The <see cref="JsonSerializerOptions"/> used to marshal any function parameters and return value.</param>
```
As written, though, it sounds strange to my ears without an article, e.g. "used to marshal return value".

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -19,6 +16,13 @@ namespace Microsoft.Extensions.AI;
 /// </summary>
 public static class ChatClientStructuredOutputExtensions
 {
+    private static readonly JsonSchemaInferenceOptions _inferenceOptions = new()
+    {
+        IncludeSchemaKeyword = true,
+        DisallowAdditionalProperties = true,
+        IncludeTypeInEnumSchemas = true
```
stephentoub, line 23: These values differ from the JsonSchemaInferenceOptions defaults. Do we have the right defaults? Or is this just a special case?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: > I would expect the handling of a null Type to live in the function that accepts an AIFunctionMetadata.

`AIFunctionMetadata` itself can contain schemas, so any function factory building the metadata needs to use an overload doesn't require it as input.

stephentoub: Both public APIs can use the same internal code. It'd just be the public one accepting a Type that would throw for a null Type before delegating the rest of its processing to the internal worker.

What's the scenario where we would encourage someone to call this public method with a null Type?

stephentoub: > AIFunctionMetadata itself can contain schemas, so any function factory building the metadata needs to use an overload doesn't require it as input.

What's the scenario where someone would be building up an AIFunctionMetadata with a null Type? I'm genuinely asking, I'm missing it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -0,0 +1,442 @@
...
+        {
+            // If the resolved options matches that of the function metadata,
+            // we can just return the precomputed JSON schema value.
+            return schema;
+        }
+
+        if (serializerOptions is null)
+        {
+            return _trueJsonSchema;
+        }
```
stephentoub, line 157: This seems like an odd policy. I don't provide a JSO so no inference takes place?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
eiriktsarpalis: Good question. It isn't clear to me if we need to adopt some of these defaults in AIFunctionFactory. The current divergences reflect the status quo.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: Refactored so that it resolves to the source generated JSO.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
```diff
@@ -0,0 +1,448 @@
...
+    /// <param name="type">The type for which to generate the schema.</param>
+    /// <param name="serializerOptions">The options used to extract the schema from the specified type.</param>
+    /// <param name="description">The description of the parameter.</param>
+    /// <param name="hasDefaultValue">Whether the parameter is optional.</param>
+    /// <param name="defaultValue">The default value of the optional parameter, if applicable.</param>
+    /// <param name="inferenceOptions">The options controlling schema inference.</param>
+    /// <returns>A <see cref="JsonElement"/> representing the schema.</returns>
+    public static JsonElement InferJsonSchema(
+        Type? type,
+        JsonSerializerOptions serializerOptions,
```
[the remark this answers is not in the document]
eiriktsarpalis: Changed all overloads to optional and having it default to `JsonDefatuls.Options`

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// An options class for configuring the behavior of <see cref="AIJsonUtilities"/> JSON schema creation functionality.
+/// </summary>
+public sealed class AIJsonSchemaCreateOptions
```
eiriktsarpalis, line 9: Should we try to hide these classes behind a nested namespace? `Microsoft.Extensions.AI.Utilities`?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: Done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/JsonFunctionCallUtilities.cs
eiriktsarpalis: Any nonstandard parameter type should guarantee that this fails.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallContent.cs
```diff
@@ -56,6 +56,46 @@ public FunctionCallContent(string callId, string name, IDictionary<string, objec
...
+    /// Creates a new instance of <see cref="FunctionCallContent"/> parsing arguments using a specified encoding and parser.
+    /// </summary>
+    /// <typeparam name="TEncoding">The encoding format from which to parse function call arguments.</typeparam>
+    /// <param name="encodedArguments">The input arguments encoded in <typeparamref name="TEncoding"/>.</param>
+    /// <param name="callId">The function call ID.</param>
+    /// <param name="name">The function name.</param>
+    /// <param name="argumentParser">The parsing implementation converting the encoding to a dictionary of arguments.</param>
+    /// <param name="exceptionFilter">Filters potential parsing exceptions that should be caught and included in the result.</param>
+    /// <returns>A new instance of <see cref="FunctionCallContent"/> containing the parse result.</returns>
+    public static FunctionCallContent CreateFromParsedArguments<TEncoding>(
```
eiriktsarpalis, line 69: This change removes all "helper" code from the Abstractions repo with the exception of this method required by client impls. I used a callback approach to invert dependencies if that's acceptable.

stephentoub: Seems ok, though at this point I'm not sure how much benefit this method really provides.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -16,12 +16,15 @@
...
 #pragma warning disable S1135 // Track uses of "TODO" tags
 #pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable SA1204 // Static elements should appear before instance elements
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>An <see cref="IChatClient"/> for an Azure AI Inference <see cref="ChatCompletionsClient"/>.</summary>
 public sealed partial class AzureAIInferenceChatClient : IChatClient
 {
+    private static readonly JsonElement _defaultParameterSchema = JsonDocument.Parse("{}").RootElement;
```
stephentoub, line 26: Dispose of the JsonDocument? Or use `JsonSerializer.Deserialize<JsonElement>`?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -428,9 +427,10 @@ private IEnumerable<ChatRequestMessage> ToAzureAIInferenceChatMessages(IEnumerab
                         string? result = resultContent.Result as string;
                         if (result is null && resultContent.Result is not null)
                         {
+                            JsonSerializerOptions options = ToolCallJsonSerializerOptions ?? JsonContext.Default.Options;
```
stephentoub, line 430: Could be a separate change if we decide to do it, but should ToolCallJsonSerializerOptions be made non-nullable like it is on the middleware clients, just defaulting it to the default options?

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -15,12 +15,15 @@
...
 
 #pragma warning disable EA0011 // Consider removing unnecessary conditional access operator (?)
+#pragma warning disable SA1204 // Static elements should appear before instance elements
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>An <see cref="IChatClient"/> for Ollama.</summary>
 public sealed class OllamaChatClient : IChatClient
 {
+    private static readonly JsonElement _defaultParameterSchema = JsonDocument.Parse("{}").RootElement;
```
stephentoub, line 25: Same question/comment

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -0,0 +1,348 @@
...
+
+    /// <summary>
+    /// Determines a JSON schema for the provided parameter metadata.
+    /// </summary>
+    /// <param name="parameterMetadata">The parameter metadata from which to infer the schema.</param>
+    /// <param name="functionMetadata">The containing function metadata.</param>
+    /// <param name="serializerOptions">The options used to extract the schema from the specified type.</param>
+    /// <param name="inferenceOptions">The options controlling schema inference.</param>
+    /// <returns>A JSON schema document encoded as a <see cref="JsonElement"/>.</returns>
+    public static JsonElement ResolveParameterSchema(
```
stephentoub, line 58: What's the rationale for ResolveParameterSchema vs ResolveParameterJsonSchema?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.csproj
```diff
@@ -25,8 +25,8 @@
     <InjectStringSyntaxAttributeOnLegacy>true</InjectStringSyntaxAttributeOnLegacy>
   </PropertyGroup>
 
-  <ItemGroup>
-    <PackageReference Include="System.Text.Json" />
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+    <PackageReference Include="System.Text.Json" VersionOverride="$(SystemTextJson8Version)" />
```
stephentoub, line 29: So this is basically saying:
- if targeting net462 or netstandard2.0, add a reference to the STJ 8.0.x package
- if targeting net8.0 or net9.0, don't add a package reference, in which case it'll just pick up the STJ from the corresponding platform

Yes? Makes sense.

eng/Versions.props
```diff
@@ -62,6 +62,7 @@
     <SystemSecurityCryptographyXmlVersion>9.0.0-rc.2.24473.5</SystemSecurityCryptographyXmlVersion>
     <SystemTextEncodingsWebVersion>9.0.0-rc.2.24473.5</SystemTextEncodingsWebVersion>
     <SystemTextJsonVersion>9.0.0-rc.2.24473.5</SystemTextJsonVersion>
+    <SystemTextJson8Version>8.0.5</SystemTextJson8Version>
```
stephentoub, line 65: If STJ revs, e.g. with a security patch, we'll need to remember to update this, or will it be automatically updated by darc? Do we need a corresponding entry in Version.Details.xml?

stephentoub (approved):
Thanks!

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallContent.cs
```diff
@@ -56,6 +56,46 @@ public FunctionCallContent(string callId, string name, IDictionary<string, objec
...
+        TEncoding encodedArguments,
+        string callId,
+        string name,
+        Func<TEncoding, IDictionary<string, object?>?> argumentParser,
+        Func<Exception, bool>? exceptionFilter = null)
```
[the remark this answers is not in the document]
eiriktsarpalis: For JSON stuff we only wrap `JsonException` and let `NotSupportedException` bubble up, but maybe it isn't necessary.

stephentoub:
Was this set to auto-merge and that's why it was merged without comments being addressed?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: It's possible, but it would require moving the defaults to `Abstractions` so that clients can reference it.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: Related to https://github.com/dotnet/runtime/issues/108930. Given that it's a singleton I figured never returning this buffer to the pool is probably acceptable.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
stephentoub: In general I think we should revisit using nested namespace for grouping. Everything is in a flat list right now.
cc: @SteveSandersonMS

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
stephentoub: I'm not understanding why this would require any movement. I'm simply suggesting changing:
```C#
public JsonSerializerOptions? ToolCallJsonSerializerOptions { get; set; }
```
to
```C#
private JsonSerializerOptions _toolCallOptions = JsonContext.Default.Options;
...
public JsonSerializerOptions? ToolCallJsonSerializerOptions
{
    get => _toolCallOptions;
    set => _toolCallOptions = Throw.IfNull(value);
}
```
and then here instead of:
```C#
JsonSerializerOptions options = ToolCallJsonSerializerOptions ?? JsonContext.Default.Options;
```
it'd just be:
```C#
JsonSerializerOptions options = ToolCallJsonSerializerOptions;
```

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: This is also a temporary measure depending on whether we decide to move the schema generator to `Abstractions` making it available to the client implemetation.

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: That's a mistake, I'll see if I can correct it in a follow-up.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.csproj
eiriktsarpalis: Yep!

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
stephentoub: Unless we're trying to avoid exposing this options instance?

eng/Versions.props
eiriktsarpalis: It's not clear to me that it would be updated automatically. Maybe @ViktorHofer knows.

eiriktsarpalis:
> Was this set to auto-merge and that's why it was merged without comments being addressed?

Exactly what happened.

eiriktsarpalis:
> We should either stop using auto-merge or I'll need to adapt and stop approving with comments.

Some repos don't trigger auto merges unless all review items have been resolved. Seems like the best of both worlds if we turned that on.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: I'm just pointing out that unless we expose a default options instance on `Abstractions` users will have to define their own contexts to fill in the default `_toolCallOptions`.

eng/Versions.props
ViktorHofer: Version.Details.xml is the input source for Maestro dependency flow which handles repo -> repo internal flow. If you want to get a version updated based on a new package being available on nuget.org, you would need to use Dependabot.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallContent.cs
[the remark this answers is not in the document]
eiriktsarpalis: We can't hardcode exception types given that it's a generic helper, so I'll just remove the filter.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
stephentoub: > I'm just pointing out that unless we expose a default options instance on Abstractions users will have to define their own contexts to fill in the default _toolCallOptions.

I don't understand why. The _current_ code doesn't require that, just substituting JsonContext.Default.Options when the user hasn't supplied their own. I'm simply suggesting moving around where we default back to that.

eiriktsarpalis: I see what you mean. It is appropriate to default to `JsonContext.Default.Options` in the particular locations where it is being done because they're serializing specific types (`JsonElement` for FCR and `IDictionary<string, object>` for FCC) which are known to be defined in the local `JsonContext`. I don't think it would be safe to assume the same thing in other locations where `ToolCallJsonSerializerOptions` is being used (e.g. when serializing `AdditionalProperties`), so if we wanted to default to something in that case we should be using the global options instead.

eng/Versions.props
RogerBarreto: Forgive my ignorance but if this extension doesn't have a dependency on .Net 9, why the package is a `9.x-preview` and not a `8.x-preview` (targeting NET 8 dependencies only) ?

stephentoub: Which package are you asking about?

Issue #5481, opened 2024-10-08
CachingChatClient is building up a string with concatenation


stephentoub:
https://github.com/dotnet/extensions/blob/a7e4f3b402620610eaa02b32978efe229bff32e6/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs#L95-L97

Every chat completion update that arrives in sequence ends up concatenating the new string onto the old. As we expect long sequences of short strings, this should be updated with something more scalable for string concatenation, like a StringBuilder.

PR #5514, opened 2024-10-11
Improve CachingChatClient's coalescing of streaming updates

stephentoub:
- Avoid O(N^2) memory allocation in the length of the received text
- Propagate additional metadata from coalesced nodes
- Propagate metadata on the coalesced TextContent, like ModelId
- Expose whether to coalesce as a setting on the client

Fixes https://github.com/dotnet/extensions/issues/5481
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5514)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
@@ -163,7 +163,6 @@ public abstract class CachingChatClient : DelegatingChatClient
                         TextContent nextContent = (TextContent)next.Contents[0];
                         _ = coalescedText.Append(nextContent.Text);
 
-                        coalesced.AdditionalProperties = MergeProperties(coalesced.AdditionalProperties, next.AdditionalProperties);
                         coalesced.AuthorName ??= next.AuthorName;
                         coalesced.CompletionId ??= next.CompletionId;
                         coalesced.CreatedAt ??= next.CreatedAt;
@@ -171,7 +170,6 @@ public abstract class CachingChatClient : DelegatingChatClient
                         coalesced.Role ??= next.Role;
 
                         coalescedContent.ModelId ??= nextContent.ModelId;
-                        coalescedContent.AdditionalProperties = MergeProperties(coalescedContent.AdditionalProperties, nextContent.AdditionalProperties);
                     }
 
                     // Complete the coalescing by patching the text of the coalesced node.
@@ -237,27 +235,4 @@ public abstract class CachingChatClient : DelegatingChatClient
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A <see cref="Task"/> representing the completion of the operation.</returns>
     protected abstract Task WriteCacheStreamingAsync(string key, IReadOnlyList<StreamingChatCompletionUpdate> value, CancellationToken cancellationToken);
-
-    /// <summary>Merges the properties from source into target.</summary>
-    /// <returns>
-    /// <paramref name="target"/> if <paramref name="target"/> is not <see langword="null"/> or if <paramref name="source"/> is <see langword="null"/> or empty.
-    /// Otherwise, a clone of <paramref name="source"/>.
-    /// </returns>
-    private static AdditionalPropertiesDictionary? MergeProperties(AdditionalPropertiesDictionary? target, AdditionalPropertiesDictionary? source)
-    {
-        if (source is { Count: > 0 })
-        {
-            if (target is null)
-            {
-                return source.Clone();
-            }
-
-            foreach (var entry in source)
-            {
-                target[entry.Key] = entry.Value;
-            }
-        }
-
-        return target;
-    }
 }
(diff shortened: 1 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
```diff
@@ -3,10 +3,13 @@
 
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
 
+#pragma warning disable S127 // "for" loop stop conditions should be invariant
```
joperezr, line 11: Just curious, why this wasn't needed before? Assume because before you were exclusively using foreach?

joperezr (commented):
Code changes LGTM but not familiar with the area so I'll defer to experts. FWIW and given this is a perf-oriented change, We do have benchmark projects added [here](https://github.com/dotnet/extensions/tree/main/bench) but they are only running manually for now. We could look into having these benchmarks run daily so we can track the impact of changes as time goes by.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
stephentoub: The PR changed the core loop for how the coalescing is done. It now has a structure like:
```C#
for (int start = 0; start < length; start++)
{
    int end;
    for (end = start + 1; end < length; end++)
    {
        ...
    }
    ...

    start = end;
}
```
and that analyzer doesn't like updating the iteration variable in that way.

joperezr: Ah, understood. I was mainly curious since if we don't think it's valuable we could probably consider turning it off globally, but I can see how some bugs could be introduced given your example above.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
```diff
@@ -50,58 +67,126 @@ public override async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteSt
...
+                    for (int i = startInclusive + 1; i < endExclusive; i++)
+                    {
+                        // Grab the next item.
+                        StreamingChatCompletionUpdate next = capturedItems[i];
+                        capturedItems[i] = null!;
+
+                        TextContent nextContent = (TextContent)next.Contents[0];
+                        _ = coalescedText.Append(nextContent.Text);
+
+                        coalesced.AdditionalProperties = MergeProperties(coalesced.AdditionalProperties, next.AdditionalProperties);
```
SteveSandersonMS, line 166: This seems potentially very expensive. Streaming output tends to be chunked on a very fine-grained level, usually individual words or maybe even tokens. Performing a full dictionary merge for every chunk could be a lot of work.

I know it's still linear in the response length, assuming the number of additional properties per chunk is unrelated to the number of chunks. But still potentially a lot.

Do you think there are definitely cases where a developer would expect to see a merged dictionary? And why would it be correct to have a "last chunk wins" rule for properties that vary on a per-chunk basis? Would be reasonable to say we just take metadata from the first chunk, and if a developer wants to preserve full granularity, they can disable the coalescing?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
```diff
@@ -152,4 +237,27 @@ public override async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteSt
...
+        if (source is { Count: > 0 })
+        {
+            if (target is null)
+            {
+                return source.Clone();
+            }
+
+            foreach (var entry in source)
+            {
+                target[entry.Key] = entry.Value;
```
SteveSandersonMS, line 257: I'm not sure we should do this, but if we do, shouldn't the precedence go the other way around (as in, we have a first-chunk wins rule rather than last-chunk wins)? That would be consistent with the logic for other non-metadata properties, e.g., `coalesced.CreatedAt ??= next.CreatedAt`.

SteveSandersonMS:
Everything else looks great!

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
stephentoub: > Do you think there are definitely cases where a developer would expect to see a merged dictionary?

There are cases where properties show up at some point after not having been on previous chunks. I'm not particularly concerned with cases where the same property shows up multiple times or with the last-write-wins policy, it was just easiest to implement, with the indexer (TryAdd would be first-wins but doesn't exist on netfx, and if we're concerned about cost, implementing it with TryGetValue and TryAdd makes it more exoensive).

I don't feel super strong about it, though. If you don't think it's worth it, I can remove the dictionary stuff.

PR #5518, opened 2024-10-14
Add embedding generator normalization

stephentoub:


src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/Embedding{T}.cs
```diff
@@ -19,4 +19,15 @@ public Embedding(ReadOnlyMemory<T> vector)
 
     /// <summary>Gets or sets the embedding vector this embedding represents.</summary>
     public ReadOnlyMemory<T> Vector { get; set; }
+
+    /// <summary>Gets or sets a value indicating whether the <see cref="Vector"/> has a unit norm.</summary>
```
eiriktsarpalis, line 23: Do we need to qualify that it's unital w.r.t. the Euclidean norm or is it standard fare in the context of embeddings?

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
```diff
@@ -0,0 +1,46 @@
...
+        IEnumerable<TInput> values, EmbeddingGenerationOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        var embeddings = await InnerGenerator.GenerateAsync(values, options, cancellationToken).ConfigureAwait(false);
+
+        foreach (var e in embeddings)
+        {
+            if (!e.Normalized)
+            {
+                var normalized = new TEmbeddingValue[e.Vector.Length];
+                TensorPrimitives.Divide(e.Vector.Span, TensorPrimitives.Norm(e.Vector.Span), normalized);
```
eiriktsarpalis, line 37: Should we check if the vector is zero here?

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <inheritdoc/>
+    public override async Task<GeneratedEmbeddings<Embedding<TEmbeddingValue>>> GenerateAsync(
+        IEnumerable<TInput> values, EmbeddingGenerationOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        var embeddings = await InnerGenerator.GenerateAsync(values, options, cancellationToken).ConfigureAwait(false);
+
+        foreach (var e in embeddings)
+        {
+            if (!e.Normalized)
```
eiriktsarpalis, line 34: Given that calculating the norm is a relatively speaking a cheap operation, couldn't we just decide to normalize on the basis of how close to 1 the value is?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/Embedding{T}.cs
eiriktsarpalis: I found a few resources online stating that in some cases it might be appropriate to use Manhattan or infinity, but I can't tell if they're niche or not.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
eiriktsarpalis: I mean the latter, i.e. `[0, 0, ... ,0]` or any other input that could result in a zero norm. I don't think this could ever be returned by a well behaved embedding generator but might make sense for the middleware to be resilient against it.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
eiriktsarpalis: (I assume that this middleware wouldn't be inserted unless the inner generator doesn't normalize embeddings or isn't reliable in how it normalizes embeddings and in either case this check seems redundant).

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#if NET // requires generic math interfaces
```
eiriktsarpalis, line 4: Should we consider authoring specialized implementations `double` and `float` in ns2.0?

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
eiriktsarpalis: Not necessarily, but this middleware could choose to ignore it if we keep it. Perhaps it makes more sense to advertise on the generator itself rather than individual vectors.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
stephentoub: > I mean the latter, i.e. `[0, 0, ..., 0]` or any other input that could result in a zero norm. I don't think this could ever be returned by a well behaved embedding generator but might make sense for the middleware to be resilient against it.

Define resilient? In such a case, this will end up returning a vector of NaNs, which seems like reasonable behavior. What would you recommend instead?

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
stephentoub: I think for now it's reasonable to say that this is only available on core.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
stephentoub: > Perhaps it makes more sense to advertise on the generator itself rather than individual vectors.

So remove Embedding<T>.Normalized and add EmbeddingGeneratorMetadata.CreatesNormalizedVectors, or something like that? It wouldn't make sense for some other kinds of embeddings, but in such cases it could just be false.

@SteveSandersonMS, opinions?

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
eiriktsarpalis: My expectation would be to forward the original vector as-is in that case.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
eiriktsarpalis: Yep. Either that or `IsNormalizingGenerator` sounds reasonable to me.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
eiriktsarpalis: In other words, I would roughly expect something like this:

```c#
T norm = TensorPrimitives.Norm(e.Vector.Span);
if (T.IsZero(norm) || T.Abs(norm - T.One) < T.CreateTruncating(SomeTolerance))
{
    continue;
}

// Existing normalization code as usual...
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/Embedding{T}.cs
SteveSandersonMS: The phrasing as given is likely *precise* enough without qualifying that we mean a Euclid norm.

However I think it's probably too terse and technical for developers without a mathematical background to understand intuitively. I have a math degree but even so would have to go searching to really definitely confirm that "has a unit norm" means "is normalized to magnitude 1 in the obvious way" as opposed to "comes with an extra thing called a 'unit norm', whatever that is".

Maybe:

```suggestion
    /// <summary>Gets or sets a value indicating whether the embedding claims to normalized. If true, this indicates that the magnitude or norm of <see cref="Vector"/> equals 1.</summary>
```

Getting the "claim" terminology in here helps a bit to make sense of why the property is settable, which otherwise seems really strange. I know this is also explained in more length in `<remarks>`.

It's also nice to avoid implying that, if the vaue is `false`, the vector is *not* normalized (since it might still be).

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
SteveSandersonMS: I thought the value of having this flag is so you can use it to make operations cheaper, e.g.:

```cs
var similarity = embedding1.Normalized && embedding2.Normalized
    ? TensorPrimitives.DotProduct(embedding1.Vector.Span, embedding2.Vector.Span)
    : TensorPrimitives.CosineSimilarity(embedding1.Vector.Span, embedding2.Vector.Span);
```

Having a flag on the embedding generator doesn't help with this unless the instance holds a reference to its original generator (which I don't think it does).

If it's not for this purpose, what purpose is it for?

The other main way to do something here would be to have a property like:

```
private float? _magnitude;

public float Magnitude {
    get {
        if (!magnitude.HasValue) {
            _magnitude = TensorPrimitives.WhateverComputesTheMagnitudeOrNormOrLength(Vector.Span);
        }

        return _magnitude.Value;
    }
    set => _magnitude = Value;
}
```

... so that people computing similarity can always do:

```cs
var similarity = TensorPrimitives.DotProduct(embedding1.Vector.Span, embedding2.Vector.Span)
    / (embedding1.Magnitude * embedding2.Magnitude);
```

... and the optimization takes effect if the generator chooses to assign `Magnitude = 1` because it knows the provider always does this.

stephentoub: I think Eirik's suggestion is that the consumer just knows what generator they're working with / assumes things are always normalized, and if they're not, this can be used to make them so. Eirik?

Personally, I like having the metadata on the embedding instance.

SteveSandersonMS: We haven't been super clear on what scenarios are improved by this change, which is making it quite hard to pick between different design possibilities. Perhaps we could frame it as two separate potential benefits:

 1. Guaranteeing that the vector is normalized
 2. Indicating whether it is normalized (either per-embedding or per-generator)

Somebody would want (1) if they have an embedding generator that actually produces non-normalized output but they want it to be normalized (e.g., so they can use dot product instead of cosine calculation elsewhere). This isn't a scenario we know exists because the generators we care about do produce normalized output. However, if someone is in this scenario, it's quite easy for them to add a `UseNormalization` step that normalizes unconditionally, or we can ship one in the library that we don't suggest you use with OpenAI/etc but it's there if you need it for some reason.

Somebody would want (2) if they want to optimize more aggressively, skipping the normalization step if it's prenormalized, or dynamically switching between dot product and cosine calculation for similarity on a per-embedding-pair basis.

 * For app developers, this is unlikely to be the case, since they likely know their vectors are prenormalized by the generator, and if not, they could put in an unconditional normalization step when they register the generator
 * For library authors, this is more plausible since they don't know or control which generator is used. However it's really niche to need this level of optimization instead of just computing the norm when they need it, or using `TensorPrimitives.CosineSimilarity` if it's a similarity calculation.

As for the risks of confusion/misuse:

 1. Shipping an unconditional `UseNormalization` is basically zero risk. People likely won't use it except in the rare case they need to. Even though we don't have a scenario where it's needed today, it will 100% produce the right output in all cases if used, so is not harmful.
 2. Keeping track of normalization status is a little more likely to get into confusion:
    * If it's at the embedding level, the flag could be wrong if someone misunderstands (e.g., they set `Normalized` to true because they *want* it to be normalized, not because they think it already is)
    * If it's at the generator level, the flag could be wrong because the truth might depend on which model/deployment is used, and the client library might not be able to tell from that string alone. Worst case is, e.g., OpenAI always produces normalized output today so it's hardcoded to `true` in all cases, but then one day they release a new embedding model that doesn't prenormalize.

Altogether I'm not hugely concerned what we do here, as it's all likely to be fine. If it were me choosing, I'd probably ship an unconditional `UseNormalization` and leave open the option to add further "is prenormalized" metadata in the future (which would then cause UseNormalization to no-op). But if you want to add the metadata now and/or track it on the embedding instances, it's all fine with me.

eiriktsarpalis: I don't necessarily mind having metadata on the embedding instance itself, the only concern is that the metadata could be lying. I tend to agree with @SteveSandersonMS's takeaway that we should just ship an unconditional normalizer.

> Having a flag on the embedding generator doesn't help with this unless the instance holds a reference to its original generator (which I don't think it does).

I think this suggests we might want to expose magnitude/normalization functionality on the embedding type itself:
```C#
public static class EmbeddingExtensions
{
    public static bool IsNormalized(this Embedding<T> embedding) where T : IRootFunctions<T> { }
    public static T Magnitude<T>(this Embedding<T> embedding) where T : IRootFunctions<T> {  }
    public static void Normalize<T>(this Embedding<T> embedding) where T : IRootFunctions<T> { }
}
```
A normalizing middleware can be trivially built on top of this and as an added bonus it avoids the ambiguities of exposing such APIs in non-floating point embeddings.

stephentoub: > I think this suggests we might want to expose magnitude/normalization functionality on the embedding type itself:

I guess I just don't see the benefit of this. It's a one liner with TensorPrimitives.

I'll just remove the metadata and make the normalization unconditional.

eiriktsarpalis: Perhaps there's benefit in defining the appropriate tolerances for declaring something as being normalized and handling zero vectors correctly.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
stephentoub: It'd be much more surprising to me if there were such opinions in the implementation.

stephentoub: It does tolerate it. It doesn't blow up, throw, fail fast. It produces NaN for garbage input.

eiriktsarpalis: Technically `[double.Epsilon, -double.Epsilon]` is a vector that defines a polarity but it isn't normalizable because its magnitude evaluates to zero. Overwriting it with NaN seems inappropriate because it loses information even if it is garbage input in the context of embeddings.

stephentoub: > Overwriting it with NaN seems inappropriate because it loses information

By definition normalization loses information and the developer is opting into normalization. It'd be surprising if the resulting vectors weren't normalized, so from my perspective allowing through a non-normalized vector would be the bug. Outputting NaNs conveys in the invalidity of the operation, while also ensuring that the likely subsequent operations are appropriately tainted. If someone ends up using CosineSimilarity, for example, with a [double.Epsilon, -double.Epsilon] vector and just does a "use whichever vector has the largest value" to find the best match, they'd end up incorrectly choosing this erroneous vector because it would result in infinity.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
stephentoub: > and handling zero vectors correctly

We obviously disagree here. I believe TP handles zero vectors correctly. And if someone wants tolerances, that's something I want them defining, not us.

eiriktsarpalis: > I believe TP handles zero vectors correctly. 

Does TP have a normalization routine?

stephentoub: That was a general statement. It has a Norm method and it has a Divide method.

src/Libraries/Microsoft.Extensions.AI/Embeddings/NormalizationEmbeddingGenerator.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <inheritdoc/>
+    public override async Task<GeneratedEmbeddings<Embedding<TEmbeddingValue>>> GenerateAsync(
+        IEnumerable<TInput> values, EmbeddingGenerationOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        var embeddings = await InnerGenerator.GenerateAsync(values, options, cancellationToken).ConfigureAwait(false);
+
+        foreach (var e in embeddings)
+        {
+            var normalized = new TEmbeddingValue[e.Vector.Length];
```
eiriktsarpalis, line 39: Why allocate a new buffer in cases where the embedding is already normalized?

stephentoub:
I'm just going to close this. We obviously have different beliefs on how this should work and when it should be used.

PR #5522, opened 2024-10-15
Add support for arrays, enums and primitive types

### LINKED ISSUE #5521: Arrays and primitive types are not supported in structured output (native or otherwise)
If you need an array (or list) of typed values from completion rather than a single object, things fail whether you're using native json schema support or not. These two tests fail, for example:

```
    [ConditionalFact]
    public virtual async Task CompleteAsync_StructuredOutputArray()
    {
        SkipIfNotEnabled();

        var response = await _chatClient.CompleteAsync<Person[]>("""
            Who are described in the following sentence?
            Jimbo Smith is a 35-year-old software developer from Cardiff, Wales.
            Josh Simpson is a 25-year-old software developer from Newport, Wales.
            """,
            useNativeJsonSchema: _chatClient.Metadata.ProviderName == "openai");

        Assert.Equal(2, response.Result.Length);
        Assert.Contains(response.Result, x => x.FullName == "Jimbo Smith");
        Assert.Contains(response.Result, x => x.FullName == "Josh Simpson");
    }

    [ConditionalFact]
    public virtual async Task CompleteAsync_StructuredOutputPrimitive()
    {
        SkipIfNotEnabled();

        var response = await _chatClient.CompleteAsync<int>("""
            As of today (october 2024), Jimbo Smith is a 35-year-old software developer from Cardiff, Wales.
            Which year was he born in?
            """,
            useNativeJsonSchema: _chatClient.Metadata.ProviderName == "openai");

        Assert.Equal(1989, response.Result);
    }

    [ConditionalFact]
    public virtual async Task CompleteAsync_StructuredOutputEnum()
    {
        SkipIfNotEnabled();

        var response = await _chatClient.CompleteAsync<Architecture>("""
            I'm using a Macbook Pro with an M2 chip. What architecture am I using?
            """,
            useNativeJsonSchema: _chatClient.Metadata.ProviderName == "openai");

        Assert.Equal(Architecture.Arm64, response.Result);
    }

    [ConditionalFact]
    public virtual async Task CompleteAsync_StructuredOutputString()
    {
        SkipIfNotEnabled();

        var response = await _chatClient.CompleteAsync<string>("""
            The software developer, Jimbo Smith, is a 35-year-old software developer from Cardiff, Wales.
            What's his full name?
            """,
            useNativeJsonSchema: _chatClient.Metadata.ProviderName == "openai");

        Assert.Equal("Jimbo Smith", response.Result);
    }

    [ConditionalFact]
    public virtual async Task CompleteAsync_StructuredOutputBool()
    {
        SkipIfNotEnabled();

        var response = await _chatClient.CompleteAsync<bool>("""
            The software developer, Jimbo Smith, is a 35-year-old software developer from Cardiff, Wales.
            Is he a medical doctor?
            """,
            useNativeJsonSchema: _chatClient.Metadata.ProviderName == "openai");

        Assert.False(response.Result);
    }
```

Except for the first test, the others don't even compile due to a constraint on the T being a `class`, but there is really no need for that to be the case and it seriously restricts the usefulness of the typed API, forcing the creation of intermediate types just to hold those values (i.e. a superfluous `Payload<T>`).

The native JSON schema in OpenAI does require the schema type to be an object (not an array or anything else). But this could be solved by having the API itself providing the wrapping object automatically, so that users fall more easily in the proverbial pit of success.

kzu:
Structured outputs in OpenAI require an `object` schema object. By detecting this situation and wrapping always in a `Payload<T>(T Data)` record, we significantly improve the developer experience by making the API transparent to that limitation (which might even be a temporary one?).

The approach works for both OpenAI as well as Azure Inference without native structured outputs. In order to signal a wrapped result to the `ChatCompletion<T>`, we use the `AdditionalProperties` dictionary with a non-standard `$wrapped` property which is a typical convention for JSON properties that are not intended for end-user consumption (like $schema).

NOTE: there will be significant conflicts with https://github.com/dotnet/extensions/pull/5513, so I'll adjust after that merge, if this is deemed a useful addition 🙏 

Fixes #5521
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5522)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -40,8 +40,7 @@ public static class ChatClientStructuredOutputExtensions
         IList<ChatMessage> chatMessages,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
+        CancellationToken cancellationToken = default) =>
         CompleteAsync<T>(chatClient, chatMessages, JsonDefaults.Options, options, useNativeJsonSchema, cancellationToken);
 
     /// <summary>Sends a user chat text message to the model, requesting a response matching the type <typeparamref name="T"/>.</summary>
@@ -61,8 +60,7 @@ public static class ChatClientStructuredOutputExtensions
         string chatMessage,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
+        CancellationToken cancellationToken = default) =>
         CompleteAsync<T>(chatClient, [new ChatMessage(ChatRole.User, chatMessage)], options, useNativeJsonSchema, cancellationToken);
 
     /// <summary>Sends a user chat text message to the model, requesting a response matching the type <typeparamref name="T"/>.</summary>
@@ -84,8 +82,7 @@ public static class ChatClientStructuredOutputExtensions
         JsonSerializerOptions serializerOptions,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
+        CancellationToken cancellationToken = default) =>
         CompleteAsync<T>(chatClient, [new ChatMessage(ChatRole.User, chatMessage)], serializerOptions, options, useNativeJsonSchema, cancellationToken);
 
     /// <summary>Sends chat messages to the model, requesting a response matching the type <typeparamref name="T"/>.</summary>
@@ -112,7 +109,6 @@ public static class ChatClientStructuredOutputExtensions
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
         CancellationToken cancellationToken = default)
-        where T : class
     {
         _ = Throw.IfNull(chatClient);
         _ = Throw.IfNull(chatMessages);
@@ -120,7 +116,7 @@ public static class ChatClientStructuredOutputExtensions
 
         serializerOptions.MakeReadOnly();
 
-        var schemaNode = (JsonObject)serializerOptions.GetJsonSchemaAsNode(typeof(T), new()
+        var exporterOptions = new JsonSchemaExporterOptions
         {
             TreatNullObliviousAsNonNullable = true,
             TransformSchemaNode = static (context, node) =>
@@ -136,7 +132,23 @@ public static class ChatClientStructuredOutputExtensions
 
                 return node;
             },
-        });
+        };
+
+        var schemaNode = (JsonObject)serializerOptions.GetJsonSchemaAsNode(typeof(T), exporterOptions);
+        var isObject = schemaNode.TryGetPropertyValue("type", out var schemaType) &&
+                schemaType?.GetValueKind() == JsonValueKind.String &&
+                schemaType.GetValue<string>() is { } type &&
+                type.Equals("object", System.StringComparison.Ordinal);
+
+        var wrapped = false;
+
+        // We wrap regardless of native structured output, since it also applies to Azure Inference
+        if (!isObject)
+        {
+            schemaNode = (JsonObject)serializerOptions.GetJsonSchemaAsNode(typeof(Payload<T>), exporterOptions);
+            wrapped = true;
+        }
+
         schemaNode.Insert(0, "$schema", "https://json-schema.org/draft/2020-12/schema");
         schemaNode.Add("additionalProperties", false);
         var schema = JsonSerializer.Serialize(schemaNode, JsonDefaults.Options.GetTypeInfo(typeof(JsonNode)));
@@ -176,6 +188,13 @@ public static class ChatClientStructuredOutputExtensions
         try
         {
             var result = await chatClient.CompleteAsync(chatMessages, options, cancellationToken).ConfigureAwait(false);
+            if (wrapped)
+            {
+                // We don't initialize the dictionary unless we need it, to avoid unnecessary allocations.
+                result.AdditionalProperties ??= [];
+                result.AdditionalProperties["$wrapped"] = true;
+            }
+
             return new ChatCompletion<T>(result, serializerOptions);
         }
         finally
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatCompletion{T}.cs
```diff
@@ -125,8 +125,24 @@ public bool TryGetResult([NotNullWhen(true)] out T? result)
...
 
+        T? deserialized = default;
+        var wrapped = AdditionalProperties?.TryGetValue("$wrapped", out var isWrappedValue) == true && isWrappedValue is bool isWrappedBool && isWrappedBool;
+
         // If there's an exception here, we want it to propagate, since the Result property is meant to throw directly
-        var deserialized = DeserializeFirstTopLevelObject(json!, (JsonTypeInfo<T>)_serializerOptions.GetTypeInfo(typeof(T)));
+
+        if (wrapped)
+        {
+            var result = DeserializeFirstTopLevelObject(json!, (JsonTypeInfo<Payload<T>>)_serializerOptions.GetTypeInfo(typeof(Payload<T>)));
```
eiriktsarpalis, line 135: This will invariably fail if the user is passing a source generated `_serializerOptions` which can't access `Payload<T>`. Assuming we decide to take this change, the result should just be parsed to a `JsonDocument` and then deserialize the nested "Data" property to the desired type.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -40,8 +40,7 @@ public static Task<ChatCompletion<T>> CompleteAsync<T>(
         IList<ChatMessage> chatMessages,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
```
eiriktsarpalis, line 44: Regardless of whether we take this change, I'd be in favor of removing this constraint since it isn't a good predictor of the JSON shape of the type. We can instead just fail at runtime depending on the value of the corresponding `JsonTypeInfo.Kind` property.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatCompletion{T}.cs
kzu: ah, yeah, absolutely!

kzu: Would this fail equally?

```
            schemaNode = (JsonObject)serializerOptions.GetJsonSchemaAsNode(typeof(Payload<T>), exporterOptions);
```

🤔

kzu:
Addressed feedback and improved code a bit.

SteveSandersonMS:
Thanks for your work on this, @kzu!

I'm not yet certain whether or not we'd add this feature, as per the comment at https://github.com/dotnet/extensions/issues/5521#issuecomment-2415955219. However, I do appreciate you've been able to do this without expanding the built-in prompt. And in many ways the end result is similar to what happens if the developer supplies a wrapper class manually.

One of the key questions that would impact whether we want to build this in is how the reliability would compare to if the developer provides their own wrapper class. A wrapper class has the advantage that its property is named semantically to reinforce what value is desired, whereas the auto-wrapper always calls the property `data` which is less clear.

It might be that it works fine, but like you say in #5521, we're looking for a "pit of success" here and so if the best results come from having a semantically-named property, we'd want to lead developers to do that even if it means a few more lines of code.

When it comes to reliability, I'm not too worried about OpenAI models (GPT 3.5T and later) since they seem to behave pretty solidly with almost any structured-output case, even the models that don't support native structured output. It's much more relevant to benchmark the reliability of small models, such as the 7B/8B parameter Lllama 2/3, Mistral, or phi3:medium. I found it pretty hard to get those to give valid structured output at all [1], since JSON schema is not a good way of describing the output format (examples are much better). Providing them with a JSON schema that's more abstract could add to the challenge.

If you have any quantification about the reliability of smaller models on tasks like "return a single enum value" or "return a single number" (which developers will often want to do) and how manual wrapping compares with autogenerating a wrapper, that could help with making a decision here. In the extreme, we might conclude that smaller models just aren't reliable for this in any case, which would lead to some other strategy around generating a JSON example instead of a JSON schema, and at that time loosen the rules about what value types you can request.

*[1] Sidenote: it turns out that changing the prompt augmentation to use `User` messages instead of `System` messages makes the Ollama-based models much more compliant about structured output, so we should do that. But it's still hard for them.*

kzu:
Fair enough. May I request then that instead of an opaque result ("couldn't convert to schema"), the API instead detects this situation and errors with a meaningful error, stating that a wrapper IS REQUIRED otherwise things won't work? It's a super easy "bug" to hit and the current behavior isn't great.

kzu:
On the semantic help you get when defining your own wrapper, I'd suggest you add this by default in the transform callback, since the description can add a lot of context for the model:

```
          TransformSchemaNode = (context, node) =>
          {
              var description = context.PropertyInfo?.AttributeProvider?.GetCustomAttributes(typeof(DescriptionAttribute), false)
                  .OfType<DescriptionAttribute>()
                  .FirstOrDefault()?.Description;

              if (description != null)
                  node["description"] = description;

              return node;
          },
```

Otherwise, you're just relying on the propery names alone.

SteveSandersonMS:
> Otherwise, you're just relying on the propery names alone. Should I report that as a separate issue?

That would be great. Thanks!

stephentoub:
@SteveSandersonMS, what would you like to do with this PR?

kzu:
> > Otherwise, you're just relying on the propery names alone. Should I report that as a separate issue?
> 
> That would be great. Thanks!

I see this is already being done after the recent merge: https://github.com/dotnet/extensions/blob/424e9748b0a4815fd291f8a7e6a4fbfb76418bbd/src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs#L241-L245

💯

SteveSandersonMS:
Closing so we can continue in https://github.com/dotnet/extensions/pull/5560

PR #5528, opened 2024-10-16
Add AdditionalPropertiesDictionary.TryGetValue<T>

stephentoub:
Fixes https://github.com/dotnet/extensions/issues/5526


eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalPropertiesDictionary.cs
```diff
@@ -103,6 +108,52 @@ public object? this[string key]
...
+    /// </param>
+    /// <returns>
+    /// <see langword="true"/> if a non-<see langword="null"/> value was found for <paramref name="key"/>
+    /// in the dictionary and converted to the requested type; otherwise, <see langword="false"/>.
+    /// </returns>
+    /// <remarks>
+    /// If a non-<see langword="null"/> is found for the key in the dictionary, but the value is not of the requested type but is
+    /// an <see cref="IConvertible"/> object, the method will attempt to convert the object to the requested type.
+    /// </remarks>
+    public bool TryGetValue<T>(string key, [NotNullWhen(true)] out T? value)
```
eiriktsarpalis, line 126: Should we explicitly ban nullable type parameters?
```suggestion
    public bool TryGetValue<T>(string key, [NotNullWhen(true)] out T? value) where T : notnull
```

stephentoub: What does that help? The parameter is already nullable in the case of reference types.

eiriktsarpalis: It explicitly prevents passing nullable type parameters even if it doesn't change the nullability of the out type parameter.

Issue #5523, opened 2024-10-16
Update UseOpenTelemetry for latest semantic convention


stephentoub:
The draft convention at https://opentelemetry.io/docs/specs/semconv/gen-ai/ was recently updated. The UseOpenTelemetry implementations need to be updated to match.

PR #5532, opened 2024-10-17
Update UseOpenTelemetry for latest genai spec updates

stephentoub:
- Events are now expected to be emitted as body fields, and the newly-recommended way to achieve that is via ILogger. So UseOpenTelemetry now takes an optional logger that it uses for emitting such data.
- I restructured the implementation to reduce duplication.
- Added logging of response format and seed.
- Added ChatOptions.TopK, as it's one of the parameters considered special by the spec.
- Updated the Azure.AI.Inference provider name to match the convention and what the library itself uses
- Updated the OpenAI client to use openai regardless of the kind of the actual client being used, per spec and recommendation

Closes https://github.com/dotnet/extensions/issues/5523
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5532)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
@@ -299,7 +299,7 @@ public sealed partial class AzureAIInferenceChatClient : IChatClient
             // These properties are strongly-typed on ChatOptions but not on ChatCompletionsOptions.
             if (options.TopK is int topK)
             {
-                result.AdditionalProperties["top_k"] = BinaryData.FromObjectAsJson(topK, ToolCallJsonSerializerOptions);
+                result.AdditionalProperties["top_k"] = BinaryData.FromObjectAsJson(topK, JsonContext.Default.Options);
             }
 
             if (options.AdditionalProperties is { } props)
@@ -505,5 +505,6 @@ public sealed partial class AzureAIInferenceChatClient : IChatClient
     [JsonSerializable(typeof(AzureAIChatToolJson))]
     [JsonSerializable(typeof(IDictionary<string, object?>))]
     [JsonSerializable(typeof(JsonElement))]
+    [JsonSerializable(typeof(int))]
     private sealed partial class JsonContext : JsonSerializerContext;
 }
--- a/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
@@ -48,7 +48,7 @@ public sealed partial class AzureAIInferenceChatClient : IChatClient
         var providerUrl = typeof(ChatCompletionsClient).GetField("_endpoint", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
             ?.GetValue(chatCompletionsClient) as Uri;
 
-        Metadata = new("AzureAIInference", providerUrl, modelId);
+        Metadata = new("az.ai.inference", providerUrl, modelId);
     }
 
     /// <summary>Gets or sets <see cref="JsonSerializerOptions"/> to use for any serialization activities related to tool call arguments and results.</summary>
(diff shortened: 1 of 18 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClient.cs
```diff
@@ -162,122 +182,97 @@ public override async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteSt
...
             string? modelId = options?.ModelId ?? _modelId;
 
             activity = _activitySource.StartActivity(
-                $"chat.completions {modelId}",
+                $"{OpenTelemetryConsts.GenAI.Chat} {modelId}",
                 ActivityKind.Client,
                 default(ActivityContext),
                 [
-                    new(OpenTelemetryConsts.GenAI.Operation.Name, "chat"),
+                    new(OpenTelemetryConsts.GenAI.Operation.Name, OpenTelemetryConsts.GenAI.Chat),
```
tarekgh, line 264: >                   new(OpenTelemetryConsts.GenAI.Operation.Name, OpenTelemetryConsts.GenAI.Chat), [](http://example.com/codeflow?start=1&length=97)

maybe you can delay adding the tags after checking if the activity is null? I am not sure if the allocation is sensitive here or not.

Also, we have TagList which can be used to avoid the array allocation?

tarekgh:
>         _meter = new(name);

I am wondering if OpenTelemetryChatClient will be created many times? I am asking because it creates ActivitySource, Logger, and Meter and dispose them when the object is not in use any more. If it is expected the component is used inside a container, would it be better to use IMeterFactory for instance like:

var meterFactory = services.GetRequiredService<IMeterFactory>();
var meter = meterFactory.Create("HatCo.Store");

---
Refers to: src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClient.cs:67 in 4594124. [](commit_id = 45941245feb238453d803e81c9b5f7425affe16e, deletion_comment = False)

tarekgh (approved):
Reviewed the usage of the telemetry stuff and left a couple of questions. LGTM!

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -297,13 +297,19 @@ private ChatCompletionsOptions ToAzureAIOptions(IList<ChatMessage> chatContents,
                 }
             }
 
+            // These properties are strongly-typed on ChatOptions but not on ChatCompletionsOptions.
+            if (options.TopK is int topK)
+            {
+                result.AdditionalProperties["top_k"] = BinaryData.FromObjectAsJson(topK, ToolCallJsonSerializerOptions);
```
eiriktsarpalis, line 303: Given that the type is known, couldn't you just use a source generated JSO?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClient.cs
```diff
@@ -117,35 +135,37 @@ public override async Task<ChatCompletion> CompleteAsync(IList<ChatMessage> chat
...
         catch (Exception ex)
         {
-            SetCompletionResponse(activity, requestModelId, null, ex, stopwatch);
+            TraceCompletion(activity, requestModelId, completion: null, ex, stopwatch);
             throw;
         }
 
-        var responseEnumerator = response.ConfigureAwait(false).GetAsyncEnumerator();
-        List<StreamingChatCompletionUpdate>? streamedContents = activity is not null ? [] : null;
+        var responseEnumerator = updates.ConfigureAwait(false).GetAsyncEnumerator();
```
eiriktsarpalis, line 166: I've never seen `ConfigureAwait` being applied to an IAE. Does the remove the need for doing `ConfigureAwait` on individual `MoveNextAsync()` calls or is it something else?

stephentoub:
>I am wondering if OpenTelemetryChatClient will be created many times? I am asking because it creates ActivitySource, Logger, and Meter and dispose them when the object is not in use any more. If it is expected the component is used inside a container, would it be better to use IMeterFactory for instance like:

It might be; it'll end up typically depending on how instances are registered in DI, whether as singleton or scoped.

It _could_ take an IMeterFactory, but that'd require adding a ctor parameter, and I don't see us doing that in many places. Is it recommended? It also wouldn't help with the ActivitySource.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClient.cs
stephentoub: Right, it effectively adds the ConfigureAwait onto each await performed implicitly as part of the foreach.

Issue #5529, opened 2024-10-16
Revisit AIContent.ModelId


stephentoub:
Why does AIContent have a ModelId property?

ChatCompletion has ModelId. StreamingChatCompletionUpdate doesn't.

Shouldn't ModelId be removed from AIContent and instead added to StreamingChatCompletionUpdate?

PR #5535, opened 2024-10-18
Remove AIContent.ModelId, add StreamingChatCompletionUpdate.ModelId

stephentoub:
Closes https://github.com/dotnet/extensions/issues/5529

This is a source and binary break. We should discuss when best to merge, and merge other similar breaks at the same time. We've had a few discussions about ctors to remove and things like that.
cc: @SteveSandersonMS, @eiriktsarpalis 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5535)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIContent.cs
@@ -32,11 +32,6 @@ public class AIContent
     [JsonIgnore]
     public object? RawRepresentation { get; set; }
 
-    /// <summary>
-    /// Gets or sets the model ID used to generate the content.
-    /// </summary>
-    public string? ModelId { get; set; }
-
     /// <summary>Gets or sets additional properties for the content.</summary>
     public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
 }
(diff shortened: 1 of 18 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIContent.cs
```diff
@@ -32,11 +32,6 @@ protected AIContent()
     [JsonIgnore]
     public object? RawRepresentation { get; set; }
 
-    /// <summary>
-    /// Gets or sets the model ID used to generate the content.
-    /// </summary>
-    public string? ModelId { get; set; }
-
     /// <summary>Gets or sets additional properties for the content.</summary>
     public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
```
eiriktsarpalis, line 36: Should we also consider scoping `AdditionalProperties` to the session rather than individual content values?

eiriktsarpalis (approved):

luisquintanilla:
@stephentoub I think I get this change for streaming scenarios. Not sure if I missed it but I don't see how the model ID is exposed for non-streaming scenarios.

luisquintanilla:
> @stephentoub I think I get this change for streaming scenarios. Not sure if I missed it but I don't see how the model ID is exposed for non-streaming scenarios.

Disregard 🙂 . I see it's in the client ctor.

stephentoub:
> I don't see how the model ID is exposed for non-streaming scenarios.

https://github.com/dotnet/extensions/blob/bdbda816432a90a3af550c9f8254ff58a0d6dc5f/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs#L62-L63

PR #5543, opened 2024-10-19
Make GenerateAsync extension just return the embedding

stephentoub:
@SteveSandersonMS, what we spoke about earlier today.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5543)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorExtensions.cs
```diff
@@ -18,7 +19,7 @@ public static class EmbeddingGeneratorExtensions
     /// <param name="options">The embedding generation options to configure the request.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The generated embedding for the specified <paramref name="value"/>.</returns>
-    public static Task<GeneratedEmbeddings<TEmbedding>> GenerateAsync<TValue, TEmbedding>(
+    public static async Task<TEmbedding> GenerateAsync<TValue, TEmbedding>(
```
eiriktsarpalis, line 22: Don't we have a rule about naming methods differently when the return type changes?

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorExtensions.cs
stephentoub: We try to avoid it. But unless we want to add Embeddings/Embedding to the interface/extension method names, respectively, I don't have a better option.

PR #5549, opened 2024-10-22
Fix a few issues in IChatClient implementations

stephentoub:
- Avoid null arg exception when constructing system message with null text
- Avoid empty exception when constructing user message with no parts
- Use all parts rather than just first text part for system message
- Handle assistant messages with both content and tools
- Avoid unnecessarily trying to weed out duplicate call ids
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5549)


src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -410,13 +409,16 @@ private sealed class AzureAIChatToolJson
...
         // Maps all of the M.E.AI types to the corresponding AzureAI types.
-        // Unrecognized content is ignored.
+        // Unrecognized or non-processable content is ignored.
 
         foreach (ChatMessage input in inputs)
         {
             if (input.Role == ChatRole.System)
             {
-                yield return new ChatRequestSystemMessage(input.Text);
+                if (input.Text is string text)
```
eiriktsarpalis, line 418: Do we want to skip the message altogether if the string is null?

stephentoub: What's the alternative?

stephentoub: >  In what cases could this happen?

new TextContent(null)

eiriktsarpalis: Not sure. Keep as-is? Replace with the empty string? It depends on what could be producing the null strings and whether skipping such content altogether is appropriate.

eiriktsarpalis: > > In what cases could this happen?
> 
> new TextContent(null)

As a user, if I was passing this intentionally I would have been surprised if it got intercepted by the client.

stephentoub: The azure client throws an exception if passed null. We could make it string.empty, but I'm not sure that's any better?

stephentoub: > Clearly it's either misuse or a mistake done by the user.

Why is that clear? A point of an abstraction like this is to enable generalized use of multiple implementations that may all differ in large or subtle ways. If I get an object model from one, I should be able to pass it to another.

Should an implementation throw if ChatOptions.Temperature is set and it doesn't have a notion of temp? Should an implementation throw if a content item has an additional property it's not familiar with? What about a content type it's not familiar with.

We've made a decision to be lenient in what's accepted in order to maximize the chances of an arbitrary consumer working with an arbitrary implementation, and I still believe that's the right decision.

If we want to prohibit null in TextContent, we found explore that, or normalize null to be empty string. There are other cases of this, though. For example, the OpenAI client blows up if you give it an empty list of content parts to an assistant message, whereas the azure inference client allows that. What did we do there. Explicitly throwing goes against our goals. Should we skip the content in that case? Manufacture an empty content? Right now it's skipping, but manufacturing empty wood be reasonable.

eiriktsarpalis: > We've made a decision to be lenient in what's accepted in order to maximize the chances of an arbitrary consumer working with an arbitrary implementation, and I still believe that's the right decision.

I suppose it's a necessary trade-off when trying to work with a shared abstraction. In that case I would probably err towards normalizing to the empty string, as in that case the structure of the chat log is preserved to a greater extent.

stephentoub: > I would probably err towards normalizing to the empty string

Done.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -569,13 +568,16 @@ private sealed class OpenAIChatToolJson
     private IEnumerable<OpenAI.Chat.ChatMessage> ToOpenAIChatMessages(IEnumerable<ChatMessage> inputs)
     {
         // Maps all of the M.E.AI types to the corresponding OpenAI types.
-        // Unrecognized content is ignored.
+        // Unrecognized or non-processable content is ignored.
 
         foreach (ChatMessage input in inputs)
         {
-            if (input.Role == ChatRole.System)
+            if (input.Role == ChatRole.System || input.Role == ChatRole.User)
```
eiriktsarpalis, line 575:
```suggestion
            if (input.Role is ChatRole.System or ChatRole.User)
```

stephentoub: That won't work, unfortunately... these aren't enums.

PR #5554, opened 2024-10-22
Lower M.E.AI.Ollama STJ dependency back to 8

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5554)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Json;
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -114,12 +115,14 @@ public async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteStreamingAs
...
             .ConfigureAwait(false);
 
-        await foreach (OllamaChatResponse? chunk in JsonSerializer.DeserializeAsyncEnumerable(
-            httpResponseStream,
-            JsonContext.Default.OllamaChatResponse,
-            topLevelValues: true,
-            cancellationToken).ConfigureAwait(false))
+        using var streamReader = new StreamReader(httpResponseStream);
+#if NET
+        while ((await streamReader.ReadLineAsync(cancellationToken).ConfigureAwait(false)) is { } line)
```
eiriktsarpalis, line 120: FWIW this slightly changes the behavior in that it requires newline separation between single-line objects, whereas the new DeserializeAsyncEnumerable method accepts arbitrary whitespace separation between objects that could themselves be indented. But I don't think there's much we can do to improve things without STJ support.

stephentoub: Yes, Ollama uses newlines.

PR #5557, opened 2024-10-22
Initial sketch of an SSE JSON serializer helper.

eiriktsarpalis:
Creating this as a draft to solicit some initial feedback on the design.

Fix https://github.com/dotnet/extensions/issues/5546.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5557)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -61,6 +61,7 @@
     <SystemSecurityCryptographyPkcsVersion>9.0.0-rc.2.24473.5</SystemSecurityCryptographyPkcsVersion>
     <SystemSecurityCryptographyXmlVersion>9.0.0-rc.2.24473.5</SystemSecurityCryptographyXmlVersion>
     <SystemTextEncodingsWebVersion>9.0.0-rc.2.24473.5</SystemTextEncodingsWebVersion>
+    <SystemNetServerSentEventsVersion>9.0.0-rc.2.24473.5</SystemNetServerSentEventsVersion>
     <SystemNumericsTensorsVersion>9.0.0-rc.2.24473.5</SystemNumericsTensorsVersion>
     <SystemTextJsonVersion>9.0.0-rc.2.24473.5</SystemTextJsonVersion>
     <!-- Dependencies from https://github.com/aspnet/AspNetCore -->
--- a/eng/packages/TestOnly.props
+++ b/eng/packages/TestOnly.props
@@ -17,6 +17,7 @@
     <PackageVersion Include="Polly.Testing" Version="8.4.2" />
     <PackageVersion Include="StrongNamer" Version="0.2.5" />
     <PackageVersion Include="System.Configuration.ConfigurationManager" Version="$(SystemConfigurationConfigurationManagerVersion)" />
+    <PackageVersion Include="System.Net.ServerSentEvents" Version="$(SystemNetServerSentEventsVersion)" />
     <PackageVersion Include="System.Numerics.Tensors" Version="$(SystemNumericsTensorsVersion)" />
     <PackageVersion Include="Verify.Xunit" Version="20.4.0" />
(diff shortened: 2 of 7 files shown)


src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Sse.cs
```diff
@@ -0,0 +1,105 @@
...
+#pragma warning disable SA1114 // Parameter list should follow declaration
+
+namespace Microsoft.Extensions.AI;
+
+public static partial class AIJsonUtilities
+{
+    private static readonly byte[] _sseEventFieldPrefix = "event: "u8.ToArray();
+    private static readonly byte[] _sseDataFieldPrefix = "data: "u8.ToArray();
+    private static readonly byte[] _sseIdFieldPrefix = "id: "u8.ToArray();
+    private static readonly byte[] _sseLineBreak = Encoding.UTF8.GetBytes(Environment.NewLine);
```
eiriktsarpalis, line 25: Using the OS-specific line break felt appropriate given that all variants are supported by the spec.

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Sse.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// <summary>
+    /// Serializes the specified server-sent events to the provided stream as JSON data.
+    /// </summary>
+    /// <typeparam name="T">Specifies the type of data payload in the event.</typeparam>
+    /// <param name="stream">The UTF-8 stream to write the server-sent events to.</param>
+    /// <param name="sseEvents">The events to serialize to the stream.</param>
+    /// <param name="options">The options configuring serialization.</param>
+    /// <param name="cancellationToken">The token taht can be used to cancel the write operation.</param>
+    /// <returns>A task representing the asynchronous write operation.</returns>
+    public static async ValueTask SerializeAsSseAsync<T>(
```
stephentoub, line 36: If we want this as a public API, I'd rather it be in System.Net.ServerSentEvents.  For now it can be a non-public implementation detail from anything that needs to use it. I don't think we should be exposing this publicly from M.E.AI.

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Sse.cs
stephentoub: Most implementations I've seen always use '\n' regardless of OS.

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Sse.cs
eiriktsarpalis: Should I file a proposal in runtime following this shape?

stephentoub: > Do you have any particular use case in mind? Like a specific streamer for IAsyncEnumerable<StreamingChatCompletionUpdate> or something else?

The use case I've been talking about all along, being able to write out the M.E.AI object model as an OpenAI-compatible response, both non-streaming and streaming varieties (this case being the latter).

> Should I file a proposal in runtime following this shape?

Sure. But in runtime ideally I'd want it to be something ASP.NET would rely on, so it'd be good to understand what its needs would be.

eiriktsarpalis:
Superseded by https://github.com/dotnet/runtime/pull/109832. Will follow up with a PR that adds the dotnet/runtime implementation as a polyfill.

PR #5560, opened 2024-10-23
Structured output improvements (continuation of PR 5522)

SteveSandersonMS:
Continuation of #5522. I had to start a new PR because #5522 has complicated merge conflicts and after fixing them, I can't rebase and push because it's on a branch. However this is still primarily the work of @kzu (thanks!).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5560)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -212,6 +212,7 @@ public static class ChatClientStructuredOutputExtensions
     {
         return element.ValueKind switch
         {
+            JsonValueKind.Null => null,
             JsonValueKind.Array => JsonArray.Create(element),
             JsonValueKind.Object => JsonObject.Create(element),
             _ => JsonValue.Create(element)
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -124,6 +124,7 @@ public static class ChatClientStructuredOutputExtensions
             type: typeof(T),
             serializerOptions: serializerOptions,
             inferenceOptions: _inferenceOptions);
+
         var isObject = schemaNode.TryGetPropertyValue("type", out var schemaType) &&
                 schemaType?.GetValueKind() == JsonValueKind.String &&
                 schemaType.GetValue<string>() is { } type &&
@@ -134,7 +135,11 @@ public static class ChatClientStructuredOutputExtensions
         // We wrap regardless of native structured output, since it also applies to Azure Inference
         if (!isObject)
         {
-            schemaNode = (JsonObject)serializerOptions.GetJsonSchemaAsNode(typeof(Payload<T>), exporterOptions);
+            schemaNode = new JsonObject
+            {
+                { "type", "object" },
+                { "properties", new JsonObject { { "data", schemaNode } } },
+            };
             wrapped = true;
         }
 
@@ -194,6 +199,4 @@ public static class ChatClientStructuredOutputExtensions
             }
         }
     }
-
-    private sealed record Payload<TValue>(TValue Data);
 }
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -168,7 +168,7 @@ public static class ChatClientStructuredOutputExtensions
 
             // When not using native structured output, augment the chat messages with a schema prompt
 #pragma warning disable SA1118 // Parameter should not span multiple lines
-            promptAugmentation = new ChatMessage(ChatRole.System, $$"""
+            promptAugmentation = new ChatMessage(ChatRole.User, $$"""
                 Respond with a JSON value conforming to the following schema:
                 ```
                 {{schema}}
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -122,20 +122,20 @@ public static class ChatClientStructuredOutputExtensions
 
         serializerOptions.MakeReadOnly();
 
-        var schemaNode = AIJsonUtilities.CreateJsonSchema(
+        var schemaElement = AIJsonUtilities.CreateJsonSchema(
             type: typeof(T),
             serializerOptions: serializerOptions,
             inferenceOptions: _inferenceOptions);
 
         var isWrappedInObject = false;
 
-        // We wrap regardless of native structured output, since it also applies to Azure Inference
-        if (!SchemaRepresentsObject(schemaNode))
+        if (!SchemaRepresentsObject(schemaElement))
         {
+            // We wrap regardless of native structured output, since it also applies to Azure Inference
             var schemaAsJsonObject = (JsonObject)JsonSerializer.Deserialize(
-                schemaNode,
+                schemaElement,
                 AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonObject)))!;
-            schemaNode = JsonSerializer.SerializeToElement(new JsonObject
+            schemaElement = JsonSerializer.SerializeToElement(new JsonObject
             {
                 { "$schema", "https://json-schema.org/draft/2020-12/schema" },
                 { "type", "object" },
@@ -145,7 +145,7 @@ public static class ChatClientStructuredOutputExtensions
             isWrappedInObject = true;
         }
 
-        var schema = JsonSerializer.Serialize(schemaNode, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
+        var schema = JsonSerializer.Serialize(schemaElement, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
 
         ChatMessage? promptAugmentation = null;
         options = (options ?? new()).Clone();
@@ -193,9 +193,9 @@ public static class ChatClientStructuredOutputExtensions
         }
     }
 
-    private static bool SchemaRepresentsObject(JsonElement schemaNode)
+    private static bool SchemaRepresentsObject(JsonElement schemaElement)
     {
-        foreach (var property in schemaNode.EnumerateObject())
+        foreach (var property in schemaElement.EnumerateObject())
         {
             if (property.NameEquals("type"u8))
             {
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
@@ -44,8 +44,7 @@ public static class ChatClientStructuredOutputExtensions
         IList<ChatMessage> chatMessages,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
+        CancellationToken cancellationToken = default) =>
         CompleteAsync<T>(chatClient, chatMessages, AIJsonUtilities.DefaultOptions, options, useNativeJsonSchema, cancellationToken);
 
     /// <summary>Sends a user chat text message to the model, requesting a response matching the type <typeparamref name="T"/>.</summary>
@@ -65,8 +64,7 @@ public static class ChatClientStructuredOutputExtensions
         string chatMessage,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
+        CancellationToken cancellationToken = default) =>
         CompleteAsync<T>(chatClient, [new ChatMessage(ChatRole.User, chatMessage)], options, useNativeJsonSchema, cancellationToken);
 
     /// <summary>Sends a user chat text message to the model, requesting a response matching the type <typeparamref name="T"/>.</summary>
@@ -88,8 +86,7 @@ public static class ChatClientStructuredOutputExtensions
         JsonSerializerOptions serializerOptions,
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
-        CancellationToken cancellationToken = default)
-        where T : class =>
+        CancellationToken cancellationToken = default) =>
         CompleteAsync<T>(chatClient, [new ChatMessage(ChatRole.User, chatMessage)], serializerOptions, options, useNativeJsonSchema, cancellationToken);
 
     /// <summary>Sends chat messages to the model, requesting a response matching the type <typeparamref name="T"/>.</summary>
@@ -116,7 +113,6 @@ public static class ChatClientStructuredOutputExtensions
         ChatOptions? options = null,
         bool? useNativeJsonSchema = null,
         CancellationToken cancellationToken = default)
-        where T : class
     {
         _ = Throw.IfNull(chatClient);
         _ = Throw.IfNull(chatMessages);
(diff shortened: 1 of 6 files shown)


kzu:
So the decision was to move forward with this? Yay! 🚀

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -116,20 +115,37 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
             type: typeof(T),
             serializerOptions: serializerOptions,
             inferenceOptions: _inferenceOptions);
 
-        var schema = JsonSerializer.Serialize(schemaNode, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
+        var isWrappedInObject = false;
+
+        if (!SchemaRepresentsObject(schemaElement))
+        {
+            // We wrap regardless of native structured output, since it also applies to Azure Inference
```
stephentoub, line 134: I don't understand this comment. Why is a specific service relevant here?

SteveSandersonMS: The comment comes from @kzu's code.

I understand it to mean that among the real services we care about, which includes Azure Inference as an example, it's common for native structured output only to allow "object"-type JSON schemas. So to make this safe to use with non-object-type `T` values, we use the schema wrapping technique even if you're using native structured output.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -116,20 +115,37 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
+            var schemaAsJsonObject = (JsonObject)JsonSerializer.Deserialize(
+                schemaElement,
+                AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonObject)))!;
+            schemaElement = JsonSerializer.SerializeToElement(new JsonObject
+            {
+                { "$schema", "https://json-schema.org/draft/2020-12/schema" },
+                { "type", "object" },
+                { "properties", new JsonObject { { "data", schemaAsJsonObject } } },
```
eiriktsarpalis, line 142: You can avoid creating an intermediate JsonObject representation for the schema:
```suggestion
                { "properties", new JsonObject { { "data", schemaElement } } },
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -116,20 +115,37 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
-        var schema = JsonSerializer.Serialize(schemaNode, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
+        var isWrappedInObject = false;
+
+        if (!SchemaRepresentsObject(schemaElement))
+        {
+            // We wrap regardless of native structured output, since it also applies to Azure Inference
+            var schemaAsJsonObject = (JsonObject)JsonSerializer.Deserialize(
+                schemaElement,
+                AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonObject)))!;
+            schemaElement = JsonSerializer.SerializeToElement(new JsonObject
```
eiriktsarpalis, line 138: Given that the final result of this method is the `schema` string, couldn't we just directly serialize the wrapping `JsonObject` to that?

eiriktsarpalis: Effectively
```csharp
string schema;
if (SchemaRepresentsObject(schemaElement))
{
    schema = JsonSerializer.Serialize(schemaElement, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
}
else
{
    JsonObject wrappedSchema = ...;
    schema = JsonSerializer.Serialize(wrappedSchema, ...);
}
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -176,4 +192,18 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
             }
         }
     }
+
+    private static bool SchemaRepresentsObject(JsonElement schemaElement)
+    {
+        foreach (var property in schemaElement.EnumerateObject())
```
eiriktsarpalis, line 198: This will throw if `schemaElement.ValueKind` is not `Object`.
```suggestion
        if (schemaElement.ValueKind is not JsonValueKind.Object)
        {
             return false;
        }

        foreach (var property in schemaElement.EnumerateObject())
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -176,4 +192,18 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
+        {
+            if (property.NameEquals("type"u8))
```
eiriktsarpalis, line 200: Do we want to support keywords containing arrays? `["object"]` is equivalent but it perhaps we can tolerate `["object","null"]`?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: We can't do that exactly as it would not build:

```
Argument 2: cannot convert from 'System.Text.Json.JsonElement' to 'System.Text.Json.Nodes.JsonNode?'
```

I'll check your other suggestions in case that leads to a different solution.

eiriktsarpalis: Pretty sure we had an implicit conversion for that, but seems we don't. `JsonValue.Create(schemaElement)` should work.

SteveSandersonMS: Update: as you've pointed out, we only need to produce a `string`, so we can just do this using string interpolation instead of building a `JsonElement`.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: Updated

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
eiriktsarpalis: Shouldn't we still be honoring indentation configuration as present on the run-time JsonSerializerOptions?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -116,20 +114,37 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
+            // For non-object-representing schemas, we wrap them in an object schema, because all
+            // the real LLM providers today require an object schema as the root. This is currently
+            // true even for providers that support native structured output.
+            isWrappedInObject = true;
+            schema = $$"""
+                {
+                    "$schema": "https://json-schema.org/draft/2020-12/schema",
+                    "type": "object",
+                    "properties": {
+                        "data": {{schema}}
```
eiriktsarpalis, line 142: `schema.ToString()` doesn't product valid JSON, unfortunately.

eiriktsarpalis: There's a `JsonElement.GetRawText()` method but that could be using a radically different indentation settings compared to the envelope. I would probably just wrap in a `JsonObject` and serialize using indentation settings present in the JSO.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: > Pretty sure we had an implicit conversion for that, but seems we don't. JsonValue.Create(schemaElement) should work.

Given this code:

```cs
schema = JsonSerializer.Serialize(new JsonObject
{
    { "$schema", "https://json-schema.org/draft/2020-12/schema" },
    { "type", "object" },
    { "properties", new JsonObject { { "data", JsonValue.Create(schemaElement) } } },
    { "additionalProperties", false },
}, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
```

... it fails at runtime with:

```
    System.InvalidOperationException : The element cannot be an object or array.
```

Did I interpret your suggestion correctly or is there something else you mean?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: Do you have a sense of when this situation would occur?

AFAIK even if we add a wrapper layer unnecessarily things will typically still work. So there's an argument that we should add the wrapper unless we can be confident it's not required. So we could say we don't have to detect all the cases where it's not required, and only deal with ones that will be particularly common.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: You mean going back to the original version that deserializes `schemaElement` into a `JsonObject` using `JsonSerializer.Deserialize`?

eiriktsarpalis: Yes, rather than deserializing you can wrap the element with a `JsonValue.Create` call.

SteveSandersonMS: Can you be really specific? It sounds like you mean:

```cs
JsonValue.Create(schemaElement)
```

... where `schemaElement` is the value returned by `AIJsonUtilities.CreateJsonSchema`. But this fails at runtime (`ThrowInvalidOperationException_NodeElementCannotBeObjectOrArray`). Do you mean something different?

SteveSandersonMS: Do you mean `JsonObject.Create(schemaElement)`?

eiriktsarpalis: Ugh. It's a known issue it seems: https://github.com/dotnet/runtime/issues/70427#issuecomment-1150960366 but there is a relatively simple workaround:
```c#
static JsonNode? FromJsonElement(JsonElement element)
{
    return element.ValueKind switch
    {
        JsonValueKind.Array => JsonArray.Create(element),
        JsonValueKind.Object => JsonObject.Create(element),
        _ => JsonValue.Create(element)
    };
}
```

SteveSandersonMS: Thanks. Done like this.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: Updated as per other conversation

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -176,4 +190,31 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
+        }
+
+        return false;
+    }
+
+    private static JsonNode? JsonElementToJsonNode(JsonElement element)
+    {
+        return element.ValueKind switch
+        {
+            JsonValueKind.Array => JsonArray.Create(element),
```
eiriktsarpalis, line 215: My initial workaround was missing handling for null. Not applicable to schemas, but adding for completeness in case it gets copied over to other contexts.
```suggestion
            JsonValueKind.Null => null,
            JsonValueKind.Array => JsonArray.Create(element),
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -116,20 +115,35 @@ public static async Task<ChatCompletion<T>> CompleteAsync<T>(
...
 
-        var schema = JsonSerializer.Serialize(schemaNode, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(JsonElement)));
+        var isWrappedInObject = false;
+        if (!SchemaRepresentsObject(schemaElement))
+        {
+            // For non-object-representing schemas, we wrap them in an object schema, because all
+            // the real LLM providers today require an object schema as the root. This is currently
+            // true even for providers that support native structured output.
+            isWrappedInObject = true;
+            schemaElement = JsonSerializer.SerializeToElement(new JsonObject
```
eiriktsarpalis, line 137: Per https://github.com/dotnet/extensions/pull/5560/files#r1814592713 I think you could still avoid creating an intermediate `JsonElement` in this branch given that the final `schema` is just a string.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
eiriktsarpalis: Given that we currently control schema generation, it won't ever occur for root-level schemas. Just posting the hypothetical question for completeness.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: Done

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: OK, then I think it's fine already. There are many possible theoretical inputs we're not optimized for. Being optimized for the ones we think will actually happen should suffice :)

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: Marking as resolved so I can merge. @stephentoub if there are any concerns, let me know and I'll go back to it.

stephentoub: Thanks. My concern is it looks like this is somehow an accomodation _specifically_ for an Azure service. It sounds like it's more of a case where this is a necessary accomodation for a whole class of services, in which case I'd rather it either not call out one specifically, or make it clear this is just an example of one.

SteveSandersonMS: Makes sense. I did update the comment along those lines: https://github.com/dotnet/extensions/pull/5560/files#diff-e069bbb2b8d33d886c9d0545784be18532ba6d51716563a084499a921a0659bcR140

Issue #5567, opened 2024-10-24
OpenAIChatClient streaming tool calls may be broken


SteveSandersonMS:
### Description

As of today, some app code that was working before is now failing with:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'bytes')
2024-10-24T15:00:53          at System.ArgumentNullException.Throw(String paramName)
2024-10-24T15:00:53          at System.Text.StringBuilder.Append(Object value)
2024-10-24T15:00:53          at Microsoft.Extensions.AI.OpenAIChatClient.CompleteStreamingAsync(IList`1 chatMessages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
2024-10-24T15:00:53          at Microsoft.Extensions.AI.OpenAIChatClient.CompleteStreamingAsync(IList`1 chatMessages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
2024-10-24T15:00:53          at Microsoft.Extensions.AI.OpenAIChatClient.CompleteStreamingAsync(IList`1 chatMessages, ChatOptions options, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
```

this happens if and only if I try to do function calling with `CompleteStreamingAsync`. It seems that the `FunctionArgumentsUpdate` may now be a `BinaryData` instance that internally holds an empty `_data` value. This causes the exception.

![Image](https://github.com/user-attachments/assets/e8630a4f-cf8c-46bd-866f-942cfd396b74)

Perhaps the underlying OpenAI/AzureOpenAI service has started returning data in a different form.

Or perhaps I've caused this in my own code by some hackery. But I've got a talk to do in a few hours so right now I'm just keeping track of this and putting in this workaround into `CompleteStreamingAsync`:

```cs
if (toolCallUpdate.FunctionArgumentsUpdate is not null
    /* workaround */ && toolCallUpdate.FunctionArgumentsUpdate.ToArray().Length > 0)
{
    _ = (existing.Arguments ??= new()).Append(toolCallUpdate.FunctionArgumentsUpdate);
}
```

### Reproduction Steps

This is in eShopSupport.

If this doesn't repro with other streaming function calling, I'll come back to work out more repro details.

### Expected behavior

No exception

### Actual behavior

Exception

stephentoub:
[...]

PR #5569, opened 2024-10-24
Work around fixed bug in System.Memory.Data

stephentoub:
BinaryData had a bug in its ToString that would throw an exception if _bytes was empty. That was fixed several years ago, but Azure SDK libraries are still referencing older versions of System.Memory.Data that don't have the fix. Thus if nothing else pulls in a newer System.Memory.Data, this can fail.

Closes https://github.com/dotnet/extensions/issues/5567
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5569)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -264,9 +264,10 @@ public sealed partial class OpenAIChatClient : IChatClient
 
                     existing.CallId ??= toolCallUpdate.ToolCallId;
                     existing.Name ??= toolCallUpdate.FunctionName;
-                    if (toolCallUpdate.FunctionArgumentsUpdate is not null)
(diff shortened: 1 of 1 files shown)


eiriktsarpalis (approved):

SteveSandersonMS (approved):

eerhardt:
> Thus if nothing else pulls in a newer System.Memory.Data, this can fail.

Would another option to fix this be to reference the `8.0.1` version of `System.Memory.Data` from `Microsoft.Extensions.AI.OpenAI`?

stephentoub:
> > Thus if nothing else pulls in a newer System.Memory.Data, this can fail.
> 
> Would another option to fix this be to reference the `8.0.1` version of `System.Memory.Data` from `Microsoft.Extensions.AI.OpenAI`?

Yes. I think we want the changes in this PR, anyway, but https://github.com/dotnet/extensions/pull/5576 adds the explicit dependency.

PR #5574, opened 2024-10-25
Add logging sampling

evgenyfedorov2:
1. Partially resolves https://github.com/dotnet/extensions/issues/5123 proposal
2. Fully resolves https://github.com/dotnet/extensions/issues/5955 proposal

, this PR is focused on the logging sampling only. The buffering part will be in a separate PR.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5574)


evgenyfedorov2:
@noahfalk @geeknoid

geeknoid:
In my doc, I had separated the idea of matching from what you do once you match. In this PR, these two things have been coupled as captured in the SamplingParameters struct.

The reason I think we need to separate "matching" from "actions upon matching" is that we have multiple actions possible. Once a record matches, we want to globally filter it, globally buffer it, filter it at the request level, and buffer it at the request level.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
```diff
@@ -0,0 +1,90 @@
...
+[Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
+public readonly struct SamplingParameters : IEquatable<SamplingParameters>
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SamplingParameters"/> struct.
+    /// </summary>
+    /// <param name="logLevel"><see cref="Microsoft.Extensions.Logging.LogLevel"/> of the log record.</param>
+    /// <param name="category">Category of the log record.</param>
+    /// <param name="eventId"><see cref="Microsoft.Extensions.Logging.EventId"/> of the log record.</param>
+    public SamplingParameters(LogLevel? logLevel, string? category, EventId? eventId)
```
samsp-msft, line 22: Should this also include an Activity/TraceId associated with the active request for the log message? In many cases the actual Id does not matter so much as whether the log message is being delivered in the context of a request or not.
If some kind of head sampling is being performed, then using the traceId so that you can sample at the request level rather than the log level may be important.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -266,6 +265,12 @@ private void ModernPath(LogLevel logLevel, EventId eventId, LoggerMessageState m
             ref readonly MessageLogger loggerInfo = ref loggers[i];
             if (loggerInfo.IsNotFilteredOut(logLevel))
             {
+                if (!config.Sampler.ShouldSample(new SamplingParameters(logLevel, loggerInfo.Category!, eventId)))
```
noahfalk, line 268: I assume we only want to sample a given record once and then reuse that result for all loggers. It would be quite surprising if a sampling policy of 'Random.Shared.GetNext() > 0.5' meant the message is randomly sent to some providers but not others.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
noahfalk: > if we have it already, lets pass it in

I don't think the code will already have it unforetunately. There is the opt-in feature to include TraceIds as part of a logging scope (https://source.dot.net/#Microsoft.Extensions.Logging/LoggerFactoryScopeProvider.cs,35), but that is an API the LoggerProvider calls back to after the LoggerFactory already invoked ILogger.Log() on the provider. I don't see a good way that to share the reference that wouldn't wind up being more expensive than doing two independent queries. AsyncLocal lookups certainly cost more than a field lookup, but thankfully not that much more, probably 5-10ns.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
```diff
@@ -43,6 +46,7 @@ public ExtendedLoggerFactory(
 #pragma warning restore S107 // Methods should not have too many parameters
     {
         _scopeProvider = scopeProvider;
+        _sampler = sampler ?? new AlwaysOnSampler();
```
noahfalk, line 49: Performance-wise its probably a little bit faster to execute `_sampler == null ? true : _sampler.ShouldSample()` instead of invoking `_sampler.ShouldSample()` when no sampler was provided. You can do a little microbenchmark to confirm.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    /// <summary>
+    /// Adds Trace-based logging sampling to the logging infrastructure. Sampling decisions
+    /// for logs match exactly the sampling decisions for the underlying <see cref="System.Diagnostics.Activity"/>.
+    /// </summary>
+    /// <param name="builder">The dependency injection container to add logging to.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    /// <remarks>Please configure Tracing Sampling separately as part of OpenTelemetry .NET.</remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
+    public static ILoggingBuilder AddTraceBasedSampling(this ILoggingBuilder builder)
```
noahfalk, line 28:
```suggestion
    public static ILoggingBuilder AddTraceBasedSampler(this ILoggingBuilder builder)
```
To keep the naming consistent with the other APIs

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/TraceBasedSampler.cs
```diff
@@ -0,0 +1,12 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Diagnostics.Sampling;
+internal sealed class TraceBasedSampler : LoggerSampler
+{
+    public override bool ShouldSample(SamplingParameters _) =>
+        Activity.Current?.Recorded ?? false;
```
noahfalk, line 11: We might want no Activity to be true, or maybe to be configurable via the API. This is a spot where some experimental feedback feels very useful.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        LogLevel = logLevel;
+        Category = category;
+        EventId = eventId;
+    }
+
+    /// <summary>
+    /// Gets the log category.
+    /// </summary>
+    public string? Category { get; }
```
noahfalk, line 32: Do these properties need to be nullable? I'm not sure when we wouldn't have the values.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSampler.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    /// <param name="probability">The desired probability of sampling. This must be between 0.0 and 1.0.
+    /// Higher the value, higher is the probability of a given log record to be sampled in.
+    /// </param>
+    /// <param name="logLevel">Apply sampling to the provided log level or below.</param>
+    /// <param name="category">Log category to apply sampling to.</param>
+    /// <param name="eventId">Log event ID to apply sampling to.</param>
+    public RatioBasedSampler(double probability, LogLevel? logLevel, string? category, EventId? eventId)
+    {
+        _sampleRate = (int)(probability * Hundred);
```
noahfalk, line 35: We may as well multiply by int.MaxValue to get better precision. For example if someone asks for 0.001 probability we should give them 0.1% of the events instead of rounding it to zero.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSampler.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private bool IsApplicable(SamplingParameters parameters)
+    {
+        if (_parameters.LogLevel is not null && parameters.LogLevel > _parameters.LogLevel)
+        {
+            return false;
+        }
+
+        if (_parameters.Category is not null && parameters.Category != _parameters.Category)
```
noahfalk, line 61: Other places that use category as a filter are treating it as a prefix, not an exact match. We probably want to do the same (if we have category at all).

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// <param name="level">The log level (and below) to apply the sampler to.</param>
+    /// <param name="category">The log category to apply the sampler to.</param>
+    /// <param name="eventId">The event ID to apply the sampler to.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
+    public static ILoggingBuilder AddRatioBasedSampler(
+        this ILoggingBuilder builder,
+        double probability,
+        LogLevel? level = null,
+        string? category = null,
```
noahfalk, line 51: I think we may want to make level be the only optional parameter and get rid of category/eventId. It feels very odd to say the sampler should only filter one category or one eventId given that users are limited to one sampler in total. It feels simpler to have them write their own sampler as soon as they need any category or eventId logic in their sampling. Alternately we could implement a more complicated RatioBasedSampler that allows specifying a list of different categories but I'd keep it simple until we've seen a clear need.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.Extensions.Logging;
+
+/// <summary>
+/// Contains the parameters helping make sampling decisions for logs.
+/// </summary>
+[Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
+public readonly struct SamplingParameters : IEquatable<SamplingParameters>
```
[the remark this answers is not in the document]
evgenyfedorov2: It was the result of fixing the warnings:

```
CA1815: Override equals and operator equals on value types
CA1066: Implement IEquatable when overriding Equals
```

noahfalk:
@tarekgh - not sure if you have seen this yet?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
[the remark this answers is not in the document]
noahfalk: Ah. I'm guessing it would be rare for someone to compare these but I suppose if some dev does decide to make a Dictionary<SamplingParameter,bool> we may as well help it run faster.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
evgenyfedorov2: But this is a `for` loop through different loggers which can (and usually do) have different categories, and since we pass the `Category` to the samples via `SamplingParameters`, there might be a custom sampler which makes sampling decisions based on the `Category` value.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
evgenyfedorov2: Removed `?`

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/TraceBasedSampler.cs
evgenyfedorov2: Updated

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSampler.cs
evgenyfedorov2: Updated

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSampler.cs
evgenyfedorov2: Removed the category/eventId as per your suggestion in another comment, closing this as irrelevant

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
evgenyfedorov2: Updated

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
evgenyfedorov2: I will keep this thread open and update later

tarekgh: agree with @noah. Also it will be a way to check if the logger is created with sampler or not too.

tarekgh: seeing AlwaysOnSampler is internal, this make my previous comment is not accurate.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/TraceBasedSampler.cs
tarekgh: nit: this can be written like 

`Activity.Current?.Recorded is true`. will invoke `Current` once.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SamplingParameters"/> struct.
+    /// </summary>
+    /// <param name="logLevel"><see cref="Microsoft.Extensions.Logging.LogLevel"/> of the log record.</param>
+    /// <param name="category">Category of the log record.</param>
+    /// <param name="eventId"><see cref="Microsoft.Extensions.Logging.EventId"/> of the log record.</param>
+    public SamplingParameters(LogLevel logLevel, string category, EventId eventId)
+    {
+        LogLevel = logLevel;
+        Category = category;
```
tarekgh, line 25: >Category = category; [](http://example.com/codeflow?start=8&length=20)

what happen if someone forced `null` value? should we intentionally not allow that here?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
tarekgh, line 14: >struct [](http://example.com/codeflow?start=16&length=6)

are we expecting this will be created a lot to make it struct?

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
tarekgh: Is it can be handled by creating a new sampler object that can sample on Activity info? I guess in some cases, users may create aggregated sampler encapsulating more than one sample inside. 

Also, should we expose samplers like `TraceBasedSampler` so users can manually create it and wrap inside other custom samplers?

tarekgh:
Reviewed, in general, looks good. I added a minor question comments.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
noahfalk: > But this is a for loop through different loggers which can (and usually do) have different categories

Thats not what I'd expect? Unless the R9 implementation is using the same named types to mean something different than they do in the M.E.L implementation, I'd expect you have one ExtendedLogger instance per category. The MessageLoggers array being looped over represents the Logger created by ILoggerProvider.CreateLogger() for that one category.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
noahfalk: > Is it can be handled by creating a new sampler object that can sample on Activity info?

Yeah, anyone could write a sampler that follows a similar approach to the TraceBasedSampler if they want to. They could also use the API that takes a delegate:
```
logging.AddSampler( p => Activity.Current?.Recorded );
```

> Also, should we expose samplers like TraceBasedSampler so users can manually create it and wrap inside other custom samplers?

My preference would be not to add more API surface given that developers could trivially reproduce the logic with 1 line of code if that's the behavior they want. Its not a big deal to me either way though.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
noahfalk: Are you proposing adding Throw.IfNull(category) check? I assume at the moment if you passed null then its possible you get a NullReferenceException inside the call to ShouldSample() depending on its implementation.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/TraceBasedSampler.cs
evgenyfedorov2: Taking into account the @noahfalk comment that we should sample in if there is no Activity, I think this makes sense
`Activity.Current?.Recorded ?? true`

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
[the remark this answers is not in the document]
evgenyfedorov2: [Here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Trace/Sampler/SamplingParameters.cs) it is called `SamplingParameters`, that's why I have decided to re-use this name. Options kind of names are usually used to represent config with the IOptions<> pattern, so might not be the best choice here. What do you think?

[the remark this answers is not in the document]
tarekgh: `SamplingParameters` is ok. I was only trying to get attention if we thought about it. Let us stick with that name if no-one else has any concern about it. By the way, I tried to look at OTEL specs just in case they suggest something but couldn't find any info there.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
evgenyfedorov2: Updated

evgenyfedorov2:
> In my doc, I had separated the idea of matching from what you do once you match. In this PR, these two things have been coupled as captured in the SamplingParameters struct.
> 
> The reason I think we need to separate "matching" from "actions upon matching" is that we have multiple actions possible. Once a record matches, we want to globally filter it, globally buffer it, filter it at the request level, and buffer it at the request level.

Discussed offline. Added configuration support allowing for specifying matching conditions per action. The action is only one for now - Ratio based sampler itself.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /// <summary>
+    /// Configures <see cref="RatioBasedSamplerOptions" /> from an instance of <see cref="IConfiguration" />.
+    /// </summary>
+    /// <param name="builder">The <see cref="ILoggingBuilder"/> to use.</param>
+    /// <param name="configuration">The <see cref="IConfiguration" /> to add.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
+    public static ILoggingBuilder AddRatioBasedSamplerConfiguration(this ILoggingBuilder builder, IConfiguration configuration)
```
noahfalk, line 44: Does this need to be public? I'm not sure what scenario would use it.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// <summary>
+    /// Adds Ratio-based sampler to the logging infrastructure. Matched logs will be sampled
+    /// according to the provided probability./>.
+    /// Higher the probability value, higher is the probability of a given log record to be sampled in.
+    /// </summary>
+    /// <param name="builder">The dependency injection container to add logging to.</param>
+    /// <param name="configuration">The <see cref="IConfiguration" /> to add.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="builder"/> is <see langword="null"/>.</exception>
+    public static ILoggingBuilder AddRatioBasedSampler(this ILoggingBuilder builder, IConfiguration configuration)
```
noahfalk, line 62: I'm thinking we may want to rename all the places that say 'RatioBasedSampler' to 'ProbabilitySampler'. What do other folks think?

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSampler.cs
```diff
@@ -0,0 +1,60 @@
...
+        return _randomInstance.Value!.Next(int.MaxValue) < int.MaxValue * probability;
+#endif
+    }
+
+    private bool TryApply(SamplingParameters parameters, out double probability)
+    {
+        probability = 0.0;
+
+        // TO DO: check if we can optimize this. It is a hot path and
+        // we should be able to minimize number of rule selections on every log record.
```
noahfalk, line 50: I agree we'd want to optimize it but it doesn't impact the design review much. I think its good you left it as a TODO 👍.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
evgenyfedorov2: Yes, my thinking was to have this method to be able to call it from HostBuilder, similar to `AddConfiguration()` for logging [here](https://github.com/dotnet/runtime/blob/36bcc2c96045c6793dcfe3151c51a0597537917a/src/libraries/Microsoft.Extensions.Hosting/src/HostBuilder.cs#L281)

noahfalk: Due to layering I don't expect HostBuilder will add a dependency to this assembly necessary to make the call. I'd suggest leaving it out for now.

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/ILoggerSamplerFilterRule.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// <summary>
+    /// Gets the maximum <see cref="LogLevel"/> of messages where this rule applies to.
+    /// </summary>
+    public int? EventId { get; }
+
+    /// <summary>
+    /// Gets the filter delegate that would be applied to messages that passed the <see cref="LogLevel"/>.
+    /// </summary>
+    public Func<string?, LogLevel?, int?, bool>? Filter { get; }
```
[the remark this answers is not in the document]
evgenyfedorov2: Removed

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -39,6 +38,12 @@ public ExtendedLogger(ExtendedLoggerFactory factory, LoggerInformation[] loggers
 
     public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
     {
+        if (MessageLoggers.Length == 0 || !_factory.Config.Sampler.ShouldSample(new SamplingParameters(logLevel, MessageLoggers[0].Category, eventId)))
```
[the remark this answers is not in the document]
evgenyfedorov2: Done

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSamplerFilterRule.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <inheritdoc/>
+    public int? EventId { get; set; }
+
+    /// <summary>
+    /// Gets or sets the probability for sampling in if this rule applies.
+    /// </summary>
+    public double Probability { get; set; }
+
+    /// <inheritdoc/>
+    public Func<string?, LogLevel?, int?, bool>? Filter { get; }
```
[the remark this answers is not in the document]
evgenyfedorov2: Removed

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
evgenyfedorov2: Made it internal

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplerRuleSelector.cs
```diff
@@ -0,0 +1,105 @@
...
+            where T : class, ILoggerSamplerFilterRule
+        {
+            bestRule = null;
+
+            // Filter rule selection:
+            // 1. Select rules with longest matching categories
+            // 2. If there nothing matched by category take all rules without category
+            // 3. If there is only one rule use it's level and filter
+            // 4. If there are multiple rules use last
+            // 5. If there are no applicable rules use global minimal level
```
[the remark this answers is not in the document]
evgenyfedorov2: I have preliminarily updated the logic, but not covered it by tests, so this is work in progress. Will complete it later, presumably after API review. Keeping this conversation open until then.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLoggerFactory.cs
evgenyfedorov2: Updated

src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Sampling/SamplingParameters.cs
evgenyfedorov2: Added a null check

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplingLoggerBuilderExtensions.cs
evgenyfedorov2: Renamed

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/SamplerRuleSelector.cs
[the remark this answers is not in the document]
evgenyfedorov2: Added tests

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSamplerOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+/// The options for ratio based sampling.
+/// </summary>
+[Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
+public class RatioBasedSamplerOptions
+{
+    /// <summary>
+    /// Gets or sets the collection of <see cref="RatioBasedSamplerFilterRule"/> used for filtering log messages.
+    /// </summary>
+#pragma warning disable CA1002 // Do not expose generic lists - List is necessary to be able to call .AddRange()
+#pragma warning disable CA2227 // Collection properties should be read only - setter is necessary for options pattern
```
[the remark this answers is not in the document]
evgenyfedorov2: API approved as is

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        // Skip rules with inapplicable event id
+        if (rule.EventId != null && rule.EventId != eventId)
+        {
+            return false;
+        }
+
+        // Skip rules with inapplicable category
+        string? categoryName = rule.CategoryName;
+        if (categoryName != null)
```
RussKie, line 62:
```suggestion
        if (categoryName is not null)
```

Applies universally

evgenyfedorov2: Done

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/RatioBasedSampler.cs
evgenyfedorov2: Added rule caching which first tries to check if there is a rule in the cache, and if there is not it it selects the best rule and adds it to the cache. The cache is invalidated on every config update. Resolving the comment, feel free to re-open if you don't agree with this.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
```diff
@@ -261,16 +261,33 @@ private void ModernPath(LogLevel logLevel, EventId eventId, LoggerMessageState m
...
         for (int i = 0; i < loggers.Length; i++)
         {
             ref readonly MessageLogger loggerInfo = ref loggers[i];
             if (loggerInfo.IsNotFilteredOut(logLevel))
             {
+                if (samplingDecision is null && config.Sampler is not null)
+                {
+                    var logEntry = new LogEntry<ModernTagJoiner>(logLevel, loggerInfo.Category, eventId, joiner, exception, static (s, e) =>
+                    {
+                        Func<LoggerMessageState, Exception?, string>? fmt = s.Formatter!;
```
tarekgh, line 274: >s.Formatter [](http://example.com/codeflow?start=76&length=11)

Is it possible at any case this can be null? if not, `Func<LoggerMessageState, Exception?, string> fmt` should be not nullable.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
```diff
@@ -0,0 +1,135 @@
...
+    where T : class, ILogSamplingFilterRule
+{
+    private readonly ConcurrentDictionary<(string, LogLevel, EventId), T?> _ruleCache = new();
+
+    public void InvalidateCache()
+    {
+        _ruleCache.Clear();
+    }
+
+    public void Select(IList<T> rules, string category, LogLevel logLevel, EventId eventId, out T? bestRule)
```
tarekgh, line 26: >public void Select(IList<T> rules, string category, LogLevel logLevel, EventId eventId, out T? bestRule) [](http://example.com/codeflow?start=4&length=104)

nit: why this doesn't just return the best rule instead having it out parameter?

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        // Skip rules with inapplicable category
+        string? categoryName = rule.CategoryName;
+        if (categoryName is not null)
+        {
+            const char WildcardChar = '*';
+
+            int wildcardIndex = categoryName.IndexOf(WildcardChar);
+            if (wildcardIndex != -1 &&
+                categoryName.IndexOf(WildcardChar, wildcardIndex + 1) != -1)
```
tarekgh, line 72: >!= -1 [](http://example.com/codeflow?start=70&length=5)

ditto

tarekgh (approved):
LGTM

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
evgenyfedorov2: oops, 90% of this code was borrowed as is from [LoggerRuleSelector.cs](https://source.dot.net/#Microsoft.Extensions.Logging/LoggerRuleSelector.cs,ca77ca515f6ede80). Thanks for catching this, I will fix.

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        // Skip rules with inapplicable category
+        string? categoryName = rule.CategoryName;
+        if (categoryName is not null)
+        {
+            const char WildcardChar = '*';
+
+            int wildcardIndex = categoryName.IndexOf(WildcardChar);
+            if (wildcardIndex != -1 &&
```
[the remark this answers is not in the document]
evgenyfedorov2: Done

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
evgenyfedorov2: Done

src/Libraries/Microsoft.Extensions.Telemetry/Sampling/LogSamplingRuleSelector.cs
evgenyfedorov2: Done

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs
evgenyfedorov2: Removed `?`

dariusclay (approved):

geeknoid (approved):

geeknoid:
Yay!

noahfalk:
Congrats! :D

PR #5577, opened 2024-10-26
Add changelogs for M.E.AI projects

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5577)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
@@ -1,0 +1,19 @@
+# Release History
+
+## 9.0.0-preview.9.24525.1
+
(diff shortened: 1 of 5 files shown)


eiriktsarpalis (approved):
Thanks. Should we be appending entries on a per-change basis or would this get compiled from the git history every time we push a release?

stephentoub:
> Thanks. Should we be appending entries on a per-change basis or would this get compiled from the git history every time we push a release?

I'm fine with either. I did this one by taking 10 minutes and just going through the commits. The repo also has an auto-generated changelog for https://github.com/dotnet/extensions/releases, but that's for the whole repo rather than per library and only for stable releases rather than previews. I spoke with Jose about possibly trying out various alternatives, but in the meantime, we can keep going with these AI-specific, manually-updated ones.

PR #5582, opened 2024-10-28
Lower `AIJsonUtilities` to STJv8 and move to Abstractions library.

eiriktsarpalis:
This PR effectively moves most of https://github.com/eiriktsarpalis/stj-schema-mapper (including its test suites) and makes it available as a polyfill to .NET 9's `JsonSchemaExporter` in dotnet/extensions. This makes it possible to lower the dependency of AIJsonUtilities to STJ version 8 and move it to the Abstractions library.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5582)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -176,6 +180,11 @@ private static JsonElement GetJsonSchemaCached(JsonSerializerOptions options, Fu
 #endif
     }
 
+#if !NET9_0_OR_GREATER
```
eiriktsarpalis, line 183: NB I made the decision to suppress trimmer warnings for this API to minimize friction when consuming the library in older TFMs.

src/Shared/Shared.csproj
```diff
@@ -33,6 +33,10 @@
     <PackageReference Include="System.Memory" Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
```
eiriktsarpalis, line 36: We also run the schema tests against .NET 9 to ensure the polyfill remains compatible with the in-box `JsonSchemaExporter`.

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
```diff
@@ -0,0 +1,1128 @@
...
+            return CompleteSchema(ref state, new JsonSchema { Ref = existingJsonPointer });
+        }
+
+        JsonSchema schema;
+        JsonConverter effectiveConverter = customConverter ?? typeInfo.Converter;
+        JsonNumberHandling effectiveNumberHandling = customNumberHandling ?? typeInfo.NumberHandling ?? typeInfo.Options.NumberHandling;
+
+        if (!IsBuiltInConverter(effectiveConverter))
+        {
+            // Return a `true` schema for types with user-defined converters.
```
stephentoub, line 126: Can you help me better understand this part? Adding a custom converter to something means that it no longer participates in schema generation?

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
```diff
@@ -0,0 +1,1128 @@
...
+    {
+        Debug.Assert(schemaType is JsonSchemaType.Integer or JsonSchemaType.Number, "schema type must be number or integer");
+        Debug.Assert(!isIeeeFloatingPoint || schemaType is JsonSchemaType.Number, "If specifying IEEE the schema type must be number");
+
+        string? pattern = null;
+
+        if ((numberHandling & (JsonNumberHandling.AllowReadingFromString | JsonNumberHandling.WriteAsString)) != 0)
+        {
+            pattern = schemaType is JsonSchemaType.Integer
+                ? @"^-?(?:0|[1-9]\d*)$"
```
stephentoub, line 640: These are patterns that show up only in the schema, right? We don't actually execute this in .NET code? (If we did, we could improve / specialize them a bit.)

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
```diff
@@ -0,0 +1,1128 @@
...
+        {
+            return null;
+        }
+
+        if (propertyInfo.AttributeProvider is { } provider)
+        {
+            return provider;
+        }
+
+        _memberNameProperty ??= typeof(JsonPropertyInfo).GetProperty("MemberName", BindingFlags.Instance | BindingFlags.NonPublic)!;
```
stephentoub, line 691: If we haven't already, we should add a test to runtime that does the same operation in the same way and will fail if we ever change it.

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
eiriktsarpalis: A custom converter means that we are not able to reason about what schema the type may support, so we use the schema allowing any JSON value.

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
eiriktsarpalis: No execution. This is expected to be executed by a validation component like the one present in JsonSchema.NET.

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
eiriktsarpalis: We should, we originally capped the lifetime of this component to November but I guess these can't ever be changed until net8.0 falls out of support.

src/Shared/JsonSchemaExporter/JsonSchemaExporter.JsonSchema.cs
```diff
@@ -0,0 +1,545 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
stephentoub, line 1: Could you add a README.md to this folder that highlights it's purely a polyfill for .NET 9? And/or make it more clear in the folder name?

stephentoub (approved):

stephentoub (approved):

PR #5585, opened 2024-10-30
Fix exception when generating boolean schemas

eiriktsarpalis:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5585)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
@@ -138,8 +138,6 @@ public static partial class AIJsonUtilities
         JsonSerializerOptions? serializerOptions = null,
         AIJsonSchemaCreateOptions? inferenceOptions = null)
     {
-        _ = Throw.IfNull(serializerOptions);
-
         serializerOptions ??= DefaultOptions;
         inferenceOptions ??= AIJsonSchemaCreateOptions.Default;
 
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -278,24 +276,24 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext ctx, JsonNode schema)
...
-
-            if (ctx.Path.IsEmpty)
-            {
-                // We are at the root-level schema node, update/append parameter-specific metadata
 
                 // Some consumers of the JSON schema, including Ollama as of v0.3.13, don't understand
                 // schemas with "type": [...], and only understand "type" being a single value.
                 // STJ represents .NET integer types as ["string", "integer"], which will then lead to an error.
-                if (TypeIsArrayContainingInteger(schema))
+                if (TypeIsArrayContainingInteger(objSchema))
```
eiriktsarpalis, line 283: @SteveSandersonMS I think we discussed this in the past, but I'm somewhat confused by this rule. If Ollama doesn't support arrays in type keywords, shouldn't we always be removing them? 

Note that `"integer"` doesn't necessarily get bundled with `"string"`. For example, the schema for `int?` would have a type keyword of `["integer", "null"]` which this transformation would render as `"integer"`.

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -138,8 +138,6 @@ public static JsonElement CreateJsonSchema(
         JsonSerializerOptions? serializerOptions = null,
         AIJsonSchemaCreateOptions? inferenceOptions = null)
     {
-        _ = Throw.IfNull(serializerOptions);
```
stephentoub, line 141: I'm surprised this slipped through. Are we missing tests for other parameters as well?

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -278,24 +276,24 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext ctx, JsonNode schema)
...
                 // Some consumers of the JSON schema, including Ollama as of v0.3.13, don't understand
                 // schemas with "type": [...], and only understand "type" being a single value.
                 // STJ represents .NET integer types as ["string", "integer"], which will then lead to an error.
-                if (TypeIsArrayContainingInteger(schema))
+                if (TypeIsArrayContainingInteger(objSchema))
                 {
                     // We don't want to emit any array for "type". In this case we know it contains "integer"
                     // so reduce the type to that alone, assuming it's the most specific type.
-                    // This makes schemas for Int32 (etc) work with Ollama
+                    // This makes schemas for Int32 (etc) work with Ollama.
```
stephentoub, line 287: Is this specific to Ollama? It feels wrong to me changing how we create schemas to cater to one specific (buggy?) implementation. Or is this a more general and reasonable restriction?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: Agreed. This was added as ad-hoc fix to unblock Ollama applications specifically until they fix it on their side (an issue has been filed for this).

src/Libraries/Microsoft.Extensions.AI/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: We were testing null being passed to non-null parameters but not the opposite. I scanned APIs for other occurrences but couldn't find any.

Issue #5581, opened 2024-10-28
Add strongly-typed `long? Seed` property to ChatOptions


stephentoub:
This seems to be ubiquitous enough that it warrants being strongly-typed rather than in AdditionalProperties.

PR #5587, opened 2024-10-31
Add ChatOptions.Seed

stephentoub:
Closes https://github.com/dotnet/extensions/issues/5581

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
@@ -27,6 +27,9 @@ public class ChatOptions
     /// <summary>Gets or sets the presence penalty for generating chat responses.</summary>
     public float? PresencePenalty { get; set; }
 
+    /// <summary>Gets or sets a seed value used by a service to control the reproducability of results.</summary>
+    public long? Seed { get; set; }
+
     /// <summary>
     /// Gets or sets the response format for the chat request.
     /// </summary>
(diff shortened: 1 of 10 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -27,6 +27,9 @@ public class ChatOptions
     /// <summary>Gets or sets the presence penalty for generating chat responses.</summary>
     public float? PresencePenalty { get; set; }
 
+    /// <summary>Gets or sets a seed value used by a service to control the reproducability of results.</summary>
+    public long? Seed { get; set; }
```
eiriktsarpalis, line 31: Unrelated to this change per se, but does specifying a seed override other configuration impacting outputs such as `Temperature`

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
stephentoub: Seed typically doesn't override other settings, it just impacts how the prng is seeded when a prng is used.

Temperature typically impacts how probabilities of next tokens are computed. A model will generate a raw score for each possible next token, and then it'll use a function like softmax to convert those scores into probabilities. Temperature impacts the distribution created by softmax, where a lower value makes the distribution sharper, causing more likely words to have a disproprionately higher probability.

PR #5596, opened 2024-11-01
Add logging/activities to FunctionInvokingChatClient

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5596)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -8,6 +8,8 @@ using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
@@ -34,8 +36,14 @@ namespace Microsoft.Extensions.AI;
 /// invocation requests to that same function.
 /// </para>
 /// </remarks>
-public class FunctionInvokingChatClient : DelegatingChatClient
+public partial class FunctionInvokingChatClient : DelegatingChatClient
 {
+    /// <summary><see cref="ActivitySource"/> used for tracking function invocations.</summary>
+    private static readonly ActivitySource _activitySource = new(typeof(FunctionInvokingChatClient).FullName!);
+
+    /// <summary>The logger to use for logging information about function invocation.</summary>
+    private readonly ILogger _logger;
+
     /// <summary>Maximum number of roundtrips allowed to the inner client.</summary>
     private int? _maximumIterationsPerRequest;
 
@@ -43,9 +51,11 @@ public class FunctionInvokingChatClient : DelegatingChatClient
     /// Initializes a new instance of the <see cref="FunctionInvokingChatClient"/> class.
     /// </summary>
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
-    public FunctionInvokingChatClient(IChatClient innerClient)
+    /// <param name="logger">An <see cref="ILogger"/> to use for logging information about function invocation.</param>
+    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null)
         : base(innerClient)
     {
+        _logger = logger ?? NullLogger.Instance;
     }
 
     /// <summary>
@@ -167,6 +177,15 @@ public class FunctionInvokingChatClient : DelegatingChatClient
             _maximumIterationsPerRequest = value;
         }
     }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether <see cref="Activity"/>s should be used to
+    /// provide telemetry for function invocation.
+    /// </summary>
+    /// <remarks>
+    /// The default value is <see langword="true"/>.
+    /// </remarks>
+    public bool EnableTelemetry { get; set; } = true;
 
     /// <inheritdoc/>
     public override async Task<ChatCompletion> CompleteAsync(IList<ChatMessage> chatMessages, ChatOptions? options = null, CancellationToken cancellationToken = default)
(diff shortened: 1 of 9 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -562,13 +572,93 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
-        return context.Function.InvokeAsync(context.CallContent.Arguments, cancellationToken);
+        using Activity? activity = _activitySource.StartActivity(context.Function.Metadata.Name);
+
+        long startingTimestamp = 0;
+        if (_logger.IsEnabled(LogLevel.Debug))
+        {
+            startingTimestamp = Stopwatch.GetTimestamp();
+            if (_logger.IsEnabled(LogLevel.Trace))
+            {
+                LogInvokingSensitive(context.Function.Metadata.Name, LoggingHelpers.AsJson(context.CallContent.Arguments));
```
eiriktsarpalis, line 587: Shouldn't this be using the JSO from the corresponding AIFunction? This would invariably fail for nontrivial types in Native AOT.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -34,18 +36,26 @@ namespace Microsoft.Extensions.AI;
 /// invocation requests to that same function.
 /// </para>
 /// </remarks>
-public class FunctionInvokingChatClient : DelegatingChatClient
+public partial class FunctionInvokingChatClient : DelegatingChatClient
 {
+    /// <summary><see cref="ActivitySource"/> used for tracking function invocations.</summary>
+    private static readonly ActivitySource _activitySource = new(typeof(FunctionInvokingChatClient).FullName!);
+
+    /// <summary>The logger to use for logging information about function invocation.</summary>
```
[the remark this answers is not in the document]
stephentoub: What would you do instead? This is the component that's invoking the functions and so it's the only one that has the data about it.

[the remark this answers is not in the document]
eiriktsarpalis: I can't think of an alternative. Just wondering if we might want to follow a similar approach in other implementations for consistency.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
eiriktsarpalis: Alternatively, if serialization is sourced from `ToolCallJsonSerializationOptions` of the underlying chat service then perhaps there needs to be a way for this layer to see it.

stephentoub: > Shouldn't this be using the JSO from the corresponding AIFunction? This would invariably fail for nontrivial types in Native AOT.

Yup, fixed.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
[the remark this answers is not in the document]
stephentoub: @SteveSandersonMS, opinions?

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI/LoggingHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+    public static string AsJson<T>(T value, JsonSerializerOptions? options = null)
+    {
+        if (options?.TryGetTypeInfo(typeof(T), out var typeInfo) is true ||
+            AIJsonUtilities.DefaultOptions.TryGetTypeInfo(typeof(T), out typeInfo))
+        {
+            return JsonSerializer.Serialize(value, typeInfo);
+        }
+
+        // Unable to get a type info for the value, so return an empty JSON object.
+        // We do not want lack of type info to disrupt application behavior with exceptions.
```
eiriktsarpalis, line 21: Should we also eat any exceptions that may be thrown by the `JsonSerializer` call above?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClientBuilderExtensions.cs
```diff
@@ -16,15 +17,19 @@ public static class FunctionInvokingChatClientBuilderExtensions
     /// </summary>
     /// <remarks>This works by adding an instance of <see cref="FunctionInvokingChatClient"/> with default options.</remarks>
     /// <param name="builder">The <see cref="ChatClientBuilder"/> being used to build the chat pipeline.</param>
+    /// <param name="loggerFactory">An optional <see cref="ILoggerFactory"/> to use to create a logger for logging function invocations.</param>
     /// <param name="configure">An optional callback that can be used to configure the <see cref="FunctionInvokingChatClient"/> instance.</param>
     /// <returns>The supplied <paramref name="builder"/>.</returns>
-    public static ChatClientBuilder UseFunctionInvocation(this ChatClientBuilder builder, Action<FunctionInvokingChatClient>? configure = null)
+    public static ChatClientBuilder UseFunctionInvocation(
+        this ChatClientBuilder builder,
+        ILoggerFactory? loggerFactory = null,
```
SteveSandersonMS, line 25: Making the developer pass in an `ILoggerFactory` manually is a little unexpected. I would have thought we'd pull it from DI, given we already have the `services` below. Having to pass it in manually will be pretty cumbersome as you'll need to call `GetRequiredService` or similar with this call.

Could we use `services.GetService()` here to get `ILoggerFactory` as an optional dependency, and use it if available? It's pretty common throughout ASP.NET Core to log stuff liberally - we don't need the app dev's explicit buy-in to do that. As long as it's logged at a low level (e.g., Debug or Trace) then it doesn't bother anyone who doesn't need to see the logs.

stephentoub: I intended to pull it from DI if this was null, but obviously neglected to. I'll fix that.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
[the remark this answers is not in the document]
SteveSandersonMS: Well technically we could add some other `UseFunctionCallLogging` middleware that people would independently put after `UseFunctionInvocation`. That would be able to log the function call/result contents, though it still wouldn't be able to have an activity around the actual function invocation itself, or to log exceptions from the function invocation. Also it wouldn't be obvious that people could/should put `UseFunctionCallLogging` in their pipeline.

So for logging at least, I feel pretty comfortable that the approach here is good. The consumer can pass in an `ILoggerFactory` if and only if they want it to be used.

For telemetry, though, I'm not as settled. As it stands it doesn't look like the consumer gets any way to control if or how this happens. It's a shame there's no `ITelemetryReceiver` or similar concept we could be passing in to make it controllable. I suppose we could pass in an `ActivitySource?` and use its presence to control this (and by default, `UseFunctionInvocation` would pass one in).

Maybe it would be sufficient just to have a `public bool EnableTelemetry { get; set; }` on `FunctionInvokingChatClient`, which either defaults to `true` or have `UseFunctionCallLogging` set it to true by default.

It's not that I have strong scenarios for needing *not* to log telemetry. Just perhaps in theory this may be used in scenarios where the app is definitely not collecting telemetry and all the work we're doing to format messages for it is pointless, or perhaps the app does collect unrelated telemetry but is not interested in function calling telemetry for some reason.

src/Libraries/Microsoft.Extensions.AI/LoggingHelpers.cs
stephentoub: In what situations might it throw? The docs call out:
```
[ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-8.0)
jsonTypeInfo is null.

[InvalidCastException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidcastexception?view=net-8.0)
value does not match the type of jsonTypeInfo.
```
neither of which should apply here. Are there other situations / exceptions? I'll change it to eat everything, that just always makes me slightly queasy.

eiriktsarpalis: There are many reasons that it can throw relating to the shape of the underlying object graph. For example, there could be a nested `object` value that has an unsupported runtime type, or failing logic in a custom converter.

Issue #5588, opened 2024-10-31
Create new Process object only once in Resource Monitoring Windows


evgenyfedorov2:
[GetCurrentProcessMemoryUsage](https://github.com/dotnet/extensions/blob/e18a05582e64603a0f2e2871bcf555e476570102/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Interop/ProcessInfo.cs#L42 ) method creates a new Process object on every method call:

```
public ulong GetCurrentProcessMemoryUsage()
{
    using Process process = Process.GetCurrentProcess();
    return (ulong)process.WorkingSet64;
}
```

Instead, it could have created it once and then reusing it for the application lifetime. This should improve performance.

haipz:
This might hit performance issue:
![Image](https://github.com/user-attachments/assets/a87ae347-fac2-4e9b-91d9-8c89b82bc629)

makazeu:
Not sure why GetCurrentProcess is really cpu-heavy

evgenyfedorov2:
Fixed in #5597

PR #5597, opened 2024-11-02
Cache current process object to avoid performance hit

haipz:
`GetCurrentProcessMemoryUsage` will be called by `src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs` from `GetSnapshot` method, which will be triggered every 1 second by default. And also this will be called by `_ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ProcessMemoryUtilization, observeValue: () => MemoryPercentage(() => _processInfo.GetCurrentProcessMemoryUsage()));` with default interval: 5 seconds.

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5597)
 
 Related issue: https://github.com/dotnet/extensions/issues/5588
 
 Refer to another similar fix in Otel: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2286

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Interop/ProcessInfo.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Interop/ProcessInfo.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 
@@ -41,7 +42,6 @@ internal sealed class ProcessInfo : IProcessInfo
 
     public ulong GetCurrentProcessMemoryUsage()
     {
-        using Process process = Process.GetCurrentProcess();
-        return (ulong)process.WorkingSet64;
+        return (ulong)Environment.WorkingSet;
     }
 }
--- a/test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Windows/ProcessInfoTests.cs
+++ b/test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Windows/ProcessInfoTests.cs
@@ -12,7 +12,6 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Windows.Test;
 /// </summary>
 /// <remarks>These tests are added for coverage reasons, but the code doesn't have
 /// the necessary environment predictability to really test it.</remarks>
-[OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX, SkipReason = "Windows specific.")]
 public sealed class ProcessInfoTests
 {
     [ConditionalFact]
--- a/test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Windows/ProcessInfoTests.cs
+++ b/test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Windows/ProcessInfoTests.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.Diagnostics.ResourceMonitoring.Windows.Interop;
+using Microsoft.TestUtilities;
+using Xunit;
+
+namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Windows.Test;
+
+/// <summary>
+/// Process Info Interop Tests.
+/// </summary>
+/// <remarks>These tests are added for coverage reasons, but the code doesn't have
+/// the necessary environment predictability to really test it.</remarks>
+[OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX, SkipReason = "Windows specific.")]
+public sealed class ProcessInfoTests
+{
+    [ConditionalFact]
+    public void GetCurrentProcessMemoryUsage()
+    {
+        var workingSet64 = new ProcessInfo().GetCurrentProcessMemoryUsage();
+        Assert.True(workingSet64 > 0);
+    }
+}
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsSnapshotProvider.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsSnapshotProvider.cs
@@ -109,8 +109,7 @@ internal sealed class WindowsSnapshotProvider : ISnapshotProvider
 
     internal static long GetMemoryUsageInBytes()
     {
-        using var process = Process.GetCurrentProcess();
-        return process.WorkingSet64;
+        return Environment.WorkingSet;
     }
 
     internal static ulong GetTotalMemoryInBytes()


evgenyfedorov2:
Did you test it and it worked as expected?

haipz:
> Did you test it and it worked as expected?

Process.WorkingSet64 won't change unless call Process.Refresh() and it will retrieve the process info while accessing WorkingSet64 again, need to add _currentProcess.Refresh() before return WorkingSet64.

danespinosa:
> > Did you test it and it worked as expected?
> 
> The value of Process.WorkingSet64 will not update unless Process.Refresh() is called. This method retrieves the latest process information, so it's necessary to call _currentProcess.Refresh() before accessing WorkingSet64.

We need to be wise about the refresh, refresh literally disposes all handles to the process and the first call to the Process object it gets the handle again and is able to get the new values, at that point getting a new process or refreshing the process is the same cost, so from that perspective we can't do much as far as I understand, as long as the extension respects the sampling rate, I think the best would be to have better defaults that are not so aggressive? 

Refresh source code: https://source.dot.net/#System.Diagnostics.Process/System/Diagnostics/Process.cs,173ad2a094ae7507,references

evgenyfedorov2:
> > > Did you test it and it worked as expected?
> > 
> > 
> > The value of Process.WorkingSet64 will not update unless Process.Refresh() is called. This method retrieves the latest process information, so it's necessary to call _currentProcess.Refresh() before accessing WorkingSet64.
> 
> We need to be wise about the refresh, refresh literally disposes all handles to the process and the first call to the Process object it gets the handle again and is able to get the new values, at that point getting a new process or refreshing the process is the same cost, so from that perspective we can't do much as far as I understand, as long as the extension respects the sampling rate, I think the best would be to have better defaults that are not so aggressive?
> 
> Refresh source code: https://source.dot.net/#System.Diagnostics.Process/System/Diagnostics/Process.cs,173ad2a094ae7507,references

That's why I am asking if it works without the Refresh() or not :) If it does, probably we can get away with it. If it does not (e.g. WorkingSet value is not getting updated), we can't do anything but use `new Process`/`Refresh()`

evgenyfedorov2:
@haipz have you had a chance to test if it works as proposed?

danespinosa:
What about using Environment.WorkingSet would seem like the best bet, but we would have to make sure it maps properly and the perf is better

danespinosa:
> What about using [Environment.WorkingSet ](https://learn.microsoft.com/en-us/dotnet/api/system.environment.workingset?view=net-8.0)would seem like the best bet, but we would have to make sure it maps properly and the perf is better
> 
> the working set does match/map (with a low discrepancy) We would still need to measure perf @haipz image

I did a quick BenchMark.Net and if we were to replace p.WorkingSet64 with Environment.WorkingSet we would see a big win.

image

danespinosa:
My last benchmark had a minor issue where I was getting the process and refreshing at the same time but the results still show a big improvement when using Environment.WorkingSet. Also @evgenyfedorov2  @haipz @Yun-Ting has done a Refresh() vs GetProcess benchmark here, there's no gains from using one or the other https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/717
image

haipz:
> My last benchmark had a minor issue where I was getting the process and refreshing at the same time but the results still show a big improvement when using Environment.WorkingSet. Also @evgenyfedorov2 @haipz @Yun-Ting has done a Refresh() vs GetProcess benchmark here, there's no gains from using one or the other [open-telemetry/opentelemetry-dotnet-contrib#717](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/717) image

@danespinosa  Awesome work. `Environment.WorkingSet` should be the right way which supports `Windows`, `Linux`, `SunOS`, `OSX`, `FreeBSD`, etc. On windows, it [GetCurrentProcess](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.GetCurrentProcess.cs#L12) then [GetProcessMemoryInfo](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.GetProcessMemoryInfo.cs#L28) from `Libraries.Kernel32`.

@evgenyfedorov2 What do you think?

evgenyfedorov2:
> > My last benchmark had a minor issue where I was getting the process and refreshing at the same time but the results still show a big improvement when using Environment.WorkingSet. Also @evgenyfedorov2 @haipz @Yun-Ting has done a Refresh() vs GetProcess benchmark here, there's no gains from using one or the other [open-telemetry/opentelemetry-dotnet-contrib#717](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/717) image
> 
> @danespinosa Awesome work. `Environment.WorkingSet` should be the right way which supports `Windows`, `Linux`, `SunOS`, `OSX`, `FreeBSD`, etc. On windows, it [GetCurrentProcess](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.GetCurrentProcess.cs#L12) then [GetProcessMemoryInfo](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.GetProcessMemoryInfo.cs#L28) from `Libraries.Kernel32`.
> 
> @evgenyfedorov2 What do you think?

Thanks, sounds good to me!

haipz:

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Windows/ProcessInfoTests.cs
```diff
@@ -0,0 +1,24 @@
...
+using Xunit;
+
+namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Windows.Test;
+
+/// <summary>
+/// Process Info Interop Tests.
+/// </summary>
+/// <remarks>These tests are added for coverage reasons, but the code doesn't have
+/// the necessary environment predictability to really test it.</remarks>
+[OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX, SkipReason = "Windows specific.")]
```
evgenyfedorov2, line 15:
```suggestion
[OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX, SkipReason = "Test for Windows specific code.")]
```

evgenyfedorov2: Otherwise, Environment.WorkingSet seems to be cross-platform

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Interop/ProcessInfo.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
```
evgenyfedorov2, line 1: Please update WindowsSnapshotProvider class with the same change as well

haipz: @evgenyfedorov2 @danespinosa WindowsSnapshotProvider is creating snapshot for (`PrivilegedProcessorTime`, `UserProcessorTime`) which requires `Process`:
```
    public Snapshot GetSnapshot()
    {
        using var process = Process.GetCurrentProcess();

        return new Snapshot(
            totalTimeSinceStart: TimeSpan.FromTicks(_timeProvider.GetUtcNow().Ticks),
            kernelTimeSinceStart: process.PrivilegedProcessorTime,
            userTimeSinceStart: process.UserProcessorTime,
            memoryUsageInBytes: (ulong)process.WorkingSet64);
    }
```
The replacement is `Environment.CpuUsage.PrivilegedTime` and `Environment.CpuUsage.UserTime` from [Environment.CpuUsage](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L373), `Process` will leverage this as well: [](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs#L235).
image
However, these are only been supported in .NET 9:
image
image

evgenyfedorov2: Thanks, that would be too much for this PR. Please create an issue for it instead.

But actually what I was asking about is this https://github.com/dotnet/extensions/blob/3f1f59cd52c58ab7e5666ba4c8ea6176f606f3d7/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsSnapshotProvider.cs#L110 - can be replaced with Environment.WorkingSet, right?

haipz: Sure, updated.

evgenyfedorov2: Thank you!

evgenyfedorov2 (approved):

Issue #5017, opened 2024-03-11
Support catch-all parameters in HTTP logging


xakep139:
Currently HTTP logging component doesn't support catch-all parameters, e.g.:

```cs
public class MyController : ControllerBase
{
    [HttpGet("/catch-all/{*url}")]
    public ActionResult GetData([PrivateData] string url)
    {
        // ...
    }
}
```

PR #5604, opened 2024-11-06
HttpRouteParser: Handle catch-all parameters

### LINKED ISSUE #5466: HttpRouteParser incorrectly handles catch-all routes
### Description

The `HttpRouteParser.TryExtractParameters` method in the `Microsoft.Extensions.Telemetry` library does not handle route parameters that are marked as catch-all. This can lead to either truncated export of the route parameter in structured logging, or misleading redacted data in the output.

### Reproduction Steps

1. Create a route with a catch-all parameter, for example: {*catchall}
2. Use the HttpRouteParser to parse this route.
3. Observe the parsing result.

### Expected behavior

The `HttpRouteParser` should correctly handle and parse catch-all route parameters.

### Actual behavior

The `HttpRouteParser` fails to correctly parse routes with catch-all parameters, leading to incorrect logs.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

When redaction is configured, redacted values may be unintuitive because only the first segment of the catch-all route is used. 

For example, the following produces the same results since only the segment `/1` is taken into account.

redaction: HmacRedactor

route template: `path/with/{*catchAll}`

requests:
- `path/with/1/2/3`
- `path/with/1/2`
- `path/with/1`
- `path/with/1/some/data`

dariusclay:
Closes #5466 #5017 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5604)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteParser.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteParser.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using Microsoft.Extensions.Compliance.Classification;
 using Microsoft.Extensions.Compliance.Redaction;
 using Microsoft.Extensions.Http.Diagnostics;
+using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.Http.Diagnostics;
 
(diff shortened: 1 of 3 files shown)


evgenyfedorov2 (approved):

src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteParser.cs
```diff
@@ -187,13 +187,42 @@ private static Segment GetParameterSegment(string httpRoute, ref int pos)
...
 
+            // The segment has '*' catch all parameter.
+            // When we meet the character it indicates param start position needs to be adjusted, so that we capture 'param' instead of '*param'
+            // *param can only appear after opening curly brace and position needs to be adjusted only once
+            else if (!catchAllParamFound && ch == '*' && pos > 0 && httpRoute[pos - 1] == '{')
+            {
+                paramNameStart++;
+
+                // Catch all parameters can start with one or two '*' characters.
+                if (httpRoute[paramNameStart] == '*')
```
eduherminio, line 198: Do we need to validate `paramNameStart` here against `httpRoute` length?
I'm thinking of ill-formed strings, not sure if there's any logic upstream that prevents that from ever happening.

dariusclay: In general, malformed routes are not handled properly in this parser, causing IndexOutOfRange exceptions. It's a bigger issue than just his part. I would consider doing this in a separate PR because of the refactoring required.

dariusclay: For incoming requests, the route template is statically checked by analyzers and issue diagnostic warnings when it's not correct. However, for outgoing requests using downstream dependency metadata route info, there are no such analyzers.

mariamgerges (approved):

PR #5605, opened 2024-11-06
Add ToChatCompletion{Async} methods for combining StreamingChatCompletionUpdates

stephentoub:
`WithMessageAddedAsync` enables writing code like:
```C#
List<ChatMessage> messages = ...;
await foreach (var update in client.CompleteStreamingAsync(messages).WithMessageAddedAsync(messages))
{
    Console.Write(update);
}
```
and upon completion of the loop, `messages` will contain a `ChatMessage` merged from all of the updates.

`ToChatCompletion` is what's used to achieve that, but is also exposed on its own so that code which has an enumerable of `StreamingChatCompletionUpdate`s can merge them into a `ChatCompletion`. This enables a non-streaming implementation to easily be authored in terms of a streaming one.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5605)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
@@ -1,0 +1,212 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+#if NET
+using System.Runtime.InteropServices;
+#endif
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+#pragma warning disable S109 // Magic numbers should not be used
+#pragma warning disable S127 // "for" loop stop conditions should be invariant
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Provides extension methods for working with <see cref="StreamingChatCompletionUpdate"/> instances.
+/// </summary>
+public static class StreamingChatCompletionUpdateExtensions
+{
+    /// <summary>Combines <see cref="StreamingChatCompletionUpdate"/> instances into a single <see cref="ChatCompletion"/>.</summary>
+    /// <param name="updates">The updates to be combined.</param>
+    /// <param name="coalesceContent">
+    /// <see langword="true"/> to attempt to coalesce contiguous <see cref="AIContent"/> items, where applicable,
+    /// into a single <see cref="AIContent"/>, in order to reduce the number of individual content items that are included in
+    /// the manufactured <see cref="ChatMessage"/> instances. When <see langword="false"/>, the original content items are used.
+    /// The default is <see langword="true"/>.
+    /// </param>
+    /// <returns>The combined <see cref="ChatCompletion"/>.</returns>
+    public static ChatCompletion ToChatCompletion(
+        this IEnumerable<StreamingChatCompletionUpdate> updates, bool coalesceContent = true)
+    {
+        _ = Throw.IfNull(updates);
+
+        ChatCompletion completion = new([]);
+        Dictionary<int, ChatMessage> messages = [];
+
+        foreach (var update in updates)
+        {
+            ProcessUpdate(update, messages, completion);
+        }
+
+        AddMessagesToCompletion(messages, completion, coalesceContent);
+
+        return completion;
+    }
+
+    /// <summary>Combines <see cref="StreamingChatCompletionUpdate"/> instances into a single <see cref="ChatCompletion"/>.</summary>
+    /// <param name="updates">The updates to be combined.</param>
+    /// <param name="coalesceContent">
+    /// <see langword="true"/> to attempt to coalesce contiguous <see cref="AIContent"/> items, where applicable,
+    /// into a single <see cref="AIContent"/>, in order to reduce the number of individual content items that are included in
+    /// the manufactured <see cref="ChatMessage"/> instances. When <see langword="false"/>, the original content items are used.
+    /// The default is <see langword="true"/>.
+    /// </param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The combined <see cref="ChatCompletion"/>.</returns>
+    public static Task<ChatCompletion> ToChatCompletionAsync(
+        this IAsyncEnumerable<StreamingChatCompletionUpdate> updates, bool coalesceContent = true, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(updates);
+
+        return ToChatCompletionAsync(updates, coalesceContent, cancellationToken);
+
+        static async Task<ChatCompletion> ToChatCompletionAsync(
+            IAsyncEnumerable<StreamingChatCompletionUpdate> updates, bool coalesceContent, CancellationToken cancellationToken)
+        {
+            ChatCompletion completion = new([]);
+            Dictionary<int, ChatMessage> messages = [];
+
+            await foreach (var update in updates.WithCancellation(cancellationToken).ConfigureAwait(false))
+            {
+                ProcessUpdate(update, messages, completion);
+            }
+
+            AddMessagesToCompletion(messages, completion, coalesceContent);
+
+            return completion;
+        }
+    }
+
+    /// <summary>Processes the <see cref="StreamingChatCompletionUpdate"/>, incorporating its contents into <paramref name="messages"/> and <paramref name="completion"/>.</summary>
+    /// <param name="update">The update to process.</param>
+    /// <param name="messages">The dictionary mapping <see cref="StreamingChatCompletionUpdate.ChoiceIndex"/> to the <see cref="ChatMessage"/> being built for that choice.</param>
+    /// <param name="completion">The <see cref="ChatCompletion"/> object whose properties should be updated based on <paramref name="update"/>.</param>
+    private static void ProcessUpdate(StreamingChatCompletionUpdate update, Dictionary<int, ChatMessage> messages, ChatCompletion completion)
+    {
+        completion.CompletionId ??= update.CompletionId;
+        completion.CreatedAt ??= update.CreatedAt;
+        completion.FinishReason ??= update.FinishReason;
+        completion.ModelId ??= update.ModelId;
+
+#if NET
+        ChatMessage message = CollectionsMarshal.GetValueRefOrAddDefault(messages, update.ChoiceIndex, out _) ??=
+            new(default, new List<AIContent>());
+#else
+        if (!messages.TryGetValue(update.ChoiceIndex, out ChatMessage? message))
+        {
+            messages[update.ChoiceIndex] = message = new(default, new List<AIContent>());
+        }
+#endif
+
+        ((List<AIContent>)message.Contents).AddRange(update.Contents);
+
+        message.AuthorName ??= update.AuthorName;
+        if (update.Role is ChatRole role && message.Role == default)
+        {
+            message.Role = role;
+        }
+
+        if (update.AdditionalProperties is not null)
+        {
+            if (message.AdditionalProperties is null)
+            {
+                message.AdditionalProperties = new(update.AdditionalProperties);
+            }
+            else
+            {
+                foreach (var entry in update.AdditionalProperties)
+                {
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
```diff
@@ -0,0 +1,266 @@
...
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Provides extension methods for working with <see cref="StreamingChatCompletionUpdate"/> instances.
+/// </summary>
+public static class StreamingChatCompletionUpdateExtensions
+{
+    /// <summary>
+    /// Augments an <see cref="IAsyncEnumerable{StreamingChatCompletionUpdate}"/> produced by
```
eiriktsarpalis, line 22: Nit: it doesn't need to have been produced by a chat completion right?
```suggestion
    /// Augments an <see cref="IAsyncEnumerable{StreamingChatCompletionUpdate}"/> typically produced by
```

eiriktsarpalis:
> `WithMessageAddedAsync` enables writing code like:

I hadn't realized that `CompleteStreamingAsync` doesn't update the chat history after the IAE has been enumerated. I'm guessing that this is influenced by the deferred execution of the resultant IAE, but it does seem possible we could go the other way and require coalescence to be a part of `CompleteStreamingAsync` for consistency with `CompleteAsync`.

stephentoub:
> > `WithMessageAddedAsync` enables writing code like:
> 
> I hadn't realized that `CompleteStreamingAsync` doesn't update the chat history after the IAE has been enumerated. I'm guessing that this is influenced by the deferred execution of the resultant IAE, but it does seem possible we could go the other way and require coalescence to be a part of `CompleteStreamingAsync` for consistency with `CompleteAsync`.

Neither of them update the chat history with the resulting message.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
```diff
@@ -0,0 +1,266 @@
...
+            finally
+            {
+                if (coalesceContent)
+                {
+                    CoalesceTextContent(updatesList);
+                }
+
+                if (ToChatCompletion(updatesList) is { Choices.Count: > 0 } completion)
+                {
+                    chatMessages.Add(completion.Message);
```
eiriktsarpalis, line 84: Should we be updating chat history on cancellation as well?

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/StreamingChatCompletionUpdateExtensionsTests.cs
```diff
@@ -0,0 +1,131 @@
...
+            new() { ChoiceIndex = 7, Text = "Shouldn't be included" },
+        ];
+
+        List<ChatMessage> messages =
+        [
+            new(ChatRole.System, "You are an AI assistant."),
+            new(ChatRole.User, "Hey!"),
+        ];
+        int i = 0;
+        await foreach (var update in Yield(updates).WithMessageAddedAsync(messages, coalesceText))
```
eiriktsarpalis, line 45: Do you have any tests exercising partial enumeration/cancellation?

SteveSandersonMS:
The naming of `WithMessageAddedAsync` feels odd. Phrasing it in the past tense sounds as if the message has already been added in some way, but it adds the message in the future after the streaming completes.

Do any of the following feel right to you?

```cs
await foreach (var update in client.CompleteStreamingAsync(messages).AppendMessageTo(messages))
await foreach (var update in client.CompleteStreamingAsync(messages).AppendResultTo(messages))
await foreach (var update in client.CompleteStreamingAsync(messages).WithAppendTo(messages))
await foreach (var update in client.CompleteStreamingAsync(messages).ThenAppendTo(messages))
```

I mildly lean towards `AppendResultTo` but don't feel strongly.

eiriktsarpalis:
I tend to associate the `Append` prefix with methods that return `void`; `With` more accurately reflects we're decorating the underlying IAE with behaviour; `Then` is typically used when appending callbacks in the style of `Task.ContinueWith` which isn't precisely what this method is doing.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
SteveSandersonMS: Great point. This likely needs to be controllable by a flag.

 * If the app presents interrupted messages as if they cease to exist (i.e., the disappear from the UI), then you wouldn't want to have the message be added if cancelled
 * In the more common case where interrupted messages are left in the UI, you would want the message to be added. Otherwise the AI would behave as if it never answered your previous question even when you can see a partial answer in the UI.

If cancellation is due to user actions, such as submitting a further message or clicking "Stop responding" or similar, it's more likely you do want the partial message to be appended. So it would make sense to default to that behavior.

I'm not sure if there are other common cases for cancellation that aren't triggered by the end user. Well, there are cases like "the application or current window has shut down" but in those cases it's not harmful to append given that we're likely abandoning the message thread anyway.

SteveSandersonMS, line 84: Appending at this point will work fine in interaction-blocking UIs like console apps because the user likely can't send any more messages until the prior streaming process is completed. But in UIs that allow continuous interaction, where a user can post more messages while a chatbot is replying to an older one (and if this doesn't auto-cancel earlier responses), then we'd end up with out-of-order messages.

If we want to make that case work more naturally, we'd need to do something like inserting a placeholder at the beginning of the process, and then populating it with the message at the end.

We could argue it's not really a valid scenario because we expect that, if you post a further message to the list, you'd also cancel any incomplete response that was still coming. But once we lock to a particular behavior it will be hard to change it in the future, so it may be safer to do the placeholder now.

stephentoub:
The hard part here is the ToChatCompletion. With that, someone can do this with:
```C#
List<StreamingChatCompletionUpdate> updates = [];
await foreach (var update in client.CompleteStreamingAsync(messages))
{
    Console.Write(update);
    updates.Add(update);
}
messages.Add(updates.ToChatCompletion().Message);
```
Given that all of the raised concerns are about WithMessageAddedAsync, I'll just remove that for now.

stephentoub:
This is now just:
```
public static ChatCompletion ToChatCompletion(
        this IEnumerable<StreamingChatCompletionUpdate> updates, bool coalesceContent = true)
```

stephentoub:
(though now that I write that out, I'm wondering if I should add one for IAsyncEnumerable as well)

stephentoub:
Refactored to have ToChatCompletion{Async} and to have them in the Abstractions library. Having them there means a leaf client can choose to use e.g. ToChatCompletionAsync to implement its CompleteAsync method around its CompleteStreamingAsync, if desired.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+    /// <param name="updates">The updates to be combined.</param>
+    /// <param name="coalesceContent">
+    /// <see langword="true"/> to attempt to coalesce contiguous <see cref="AIContent"/> items, where applicable,
+    /// into a single <see cref="AIContent"/>, in order to reduce the number of individual content items that are included in
+    /// the manufactured <see cref="ChatMessage"/> instances. When <see langword="false"/>, the original content items are used.
+    /// The default is <see langword="true"/>.
+    /// </param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The combined <see cref="ChatCompletion"/>.</returns>
+    public static Task<ChatCompletion> ToChatCompletionAsync(
```
eiriktsarpalis, line 60: This approach seems cleaner, but at the same time doesn't it negate the purpose of streaming completions? You wouldn't be able to read anything until the enumeration has finished. A mechanism for injecting side-effects as results arrive (e.g. by way of a `Select` or `Tee` method) becomes almost mandatory.

eiriktsarpalis: But since we don't have a readily available Linq for IAE, perhaps it makes sense to bake functionality into this method, by way of an optional `Action<StreamingChatCompletionUpdate>? onCompletionUpdate` callback parameter.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+    /// <param name="updates">The updates to be combined.</param>
+    /// <param name="coalesceContent">
+    /// <see langword="true"/> to attempt to coalesce contiguous <see cref="AIContent"/> items, where applicable,
+    /// into a single <see cref="AIContent"/>, in order to reduce the number of individual content items that are included in
+    /// the manufactured <see cref="ChatMessage"/> instances. When <see langword="false"/>, the original content items are used.
+    /// The default is <see langword="true"/>.
+    /// </param>
+    /// <returns>The combined <see cref="ChatCompletion"/>.</returns>
+    public static ChatCompletion ToChatCompletion(
+        this IEnumerable<StreamingChatCompletionUpdate> updates, bool coalesceContent = true)
```
eiriktsarpalis, line 33: I'm surprised to see a blocking enumerable variant being exposed. I'm guessing it's here for completeness when needing to map buffered results?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
stephentoub: The enumerable one isn't "blocking". The intent is it works on an already created collection, eg
```C#
List<StreamingChatCompletionUpdate> updates = [];
await foreach (var update in client.CompleteStreamingAsync(messages))
{
    Console.Write(update);
    updates.Add(update);
}
messages.Add(updates.ToChatCompletion().Message);
```

And the async equivalent is primarily so that a client can implement its non-streaming implementation in terms of its streaming, eg
```C#
public Task<ChatCompletion> CompleteAsync(
        IList<ChatMessage> chatMessages, ChatOptions? options = null, CancellationToken cancellationToken = default) =>
    CompleteStreamingAsync(chatMessages, options, cancellationToken).ToChatCompletionAsync();
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
stephentoub: See above

stephentoub:
> As `Microsoft.Extensions.AI.Abstractions` is accumulating more and more helper methods, I'm growing somewhat uncomfortable with the "Abstractions" suffix in this project. Would it make sense to rename this to be just `Microsoft.Extensions.AI` and then have the existing `Microsoft.Extensions.AI` called something along the lines of `Microsoft.Extensions.AI.Middlewares`?
> 
> cc @SteveSandersonMS

I think it's fine. These helpers are to aid in the development of implementations of the abstractions.

eiriktsarpalis (approved):

SteveSandersonMS (approved):
Looks good.

In a previous iteration of this PR, I think you were able to remove the equivalent logic from `CachingChatClient` and use your new helper instead. But in the current version of the PR that's no longer the case. Is it no longer possible to avoid having two versions of this logic?

stephentoub:
> Is it no longer possible to avoid having two versions of this logic?

I put these into m.e.ai.abstractions, and the caching stuff is up in m.e.ai. We could reconsolidate if we wanted to expose the coalescing publicly from abstractions.

stephentoub:
> > Is it no longer possible to avoid having two versions of this logic?
> 
> I put these into m.e.ai.abstractions, and the caching stuff is up in m.e.ai. We could reconsolidate if we wanted to expose the coalescing publicly from abstractions.

Went another way: https://github.com/dotnet/extensions/pull/5616

PR #5606, opened 2024-11-06
Rework UseChatOptions as ConfigureOptions

stephentoub:
- Renames UseChatOptions / UseEmbeddingGenerationOptions to be ConfigureOptions
- Changes the delegate to just be `Action<XxOptions>` rather than `Func<XxOptions?, XxOptions?>`
- The delegate is now always invoked with a clone of the original options (or a new instance if none was provided) so that the callback can simply mutate / configure the options instance.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5606)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
@@ -216,24 +216,12 @@ Console.WriteLine((await client.CompleteAsync("What is AI?")).Message);
 
 Every call to `CompleteAsync` or `CompleteStreamingAsync` may optionally supply a `ChatOptions` instance containing additional parameters for the operation. The most common parameters that are common amongst AI models and services show up as strongly-typed properties on the type, such as `ChatOptions.Temperature`. Other parameters may be supplied by name in a weakly-typed manner via the `ChatOptions.AdditionalProperties` dictionary.
 
-Options may also be baked into an `IChatClient` via the `UseChatOptions` extension method on `ChatClientBuilder`. This delegating client wraps another client and invokes the supplied delegate to populate a `ChatOptions` instance for every call. For example, to ensure that the `ChatOptions.ModelId` property defaults to a particular model name, code like the following may be used:
+Options may also be baked into an `IChatClient` via the `ConfigureOptions` extension method on `ChatClientBuilder`. This delegating client wraps another client and invokes the supplied delegate to populate a `ChatOptions` instance for every call. For example, to ensure that the `ChatOptions.ModelId` property defaults to a particular model name, code like the following may be used:
 ```csharp
 using Microsoft.Extensions.AI;
 
 IChatClient client = new ChatClientBuilder()
-    .UseChatOptions(options =>
-    {
-        // If no model was specified...
-        if (options?.ModelId is null)
-        {
-            // Clone the options to avoid mutating the original instance, or create a new
-            // one if no options were provided. Then set the desired model name.
-            options = options?.Clone() ?? new();
-            options.ModelId = "phi3";
-        }
-
-        return options;
-    })
+    .ConfigureOptions(options => options.ModelId ??= "phi3")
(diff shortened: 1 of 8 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
```diff
@@ -212,6 +212,34 @@ IChatClient client = new ChatClientBuilder()
...
+```csharp
+using Microsoft.Extensions.AI;
+
+IChatClient client = new ChatClientBuilder()
+    .UseChatOptions(options =>
+    {
+        // If no model was specified...
+        if (options?.ModelId is null)
+        {
+            // Clone the options to avoid mutating the original instance, or create a new
```
eiriktsarpalis, line 229: Unrelated to the docs per se, but this seems like a pitfall users can easily fall into. Would we consider creating a defensive copy before calling the delegate to avoid checks like this on the user's side?

stephentoub: I've pushed back on it previously, but as the feedback has now been provided by multiple folks independently, I'll change it.

stephentoub:
@glennc, FYI

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ConfigureOptionsChatClientBuilderExtensions.cs
```diff
@@ -12,41 +12,25 @@ namespace Microsoft.Extensions.AI;
...
-    /// </para>
+    /// This can be used to set default options. The <paramref name="configure"/> delegate is passed either a new instance of
+    /// <see cref="ChatOptions"/> if the caller didn't supply a <see cref="ChatOptions"/> instance, or a clone (via <see cref="ChatOptions.Clone"/>
+    /// of the caller-supplied instance if one was supplied.
     /// </remarks>
-    public static ChatClientBuilder UseChatOptions(
-        this ChatClientBuilder builder, Func<ChatOptions?, ChatOptions?> configureOptions)
+    /// <returns>The <paramref name="builder"/>.</returns>
+    public static ChatClientBuilder ConfigureOptions(
+        this ChatClientBuilder builder, Action<ChatOptions> configure)
```
eiriktsarpalis, line 29: Replacing Func with Action is an improvement and makes the API more consistent with aspnet conventions. The only downside is that the configuration layer must populate the provided value, and are not given the option to return an options instance of their own if that's necessary.

eiriktsarpalis (approved):
Thanks!

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ConfigureOptionsChatClientBuilderExtensions.cs
stephentoub: Yup.

I debated that, and we could still have it be a `Func<ChatOptions, ChatOptions>`. It just means for what is probably the most common case you're forced to always have `return options;` at the end even though you don't need it.

We could change it, though. @glennc, you'd previously provided feedback in this area... opinions? @SteveSandersonMS?

stephentoub: (We could also support both, if we really wanted to, now or later adding a Func overload.)

SteveSandersonMS: I definitely like the simplicity of the `Action` here - `.ConfigureOptions(options => options.ModelId ??= "phi3")` looks great.

If we need to add a `Func` overload later that would be fine too but right now I'm not clear on what scenarios there would be for completely replacing the incoming `chatOptions` with a different set of values, erasing any other options that were set at the call site.

SteveSandersonMS (approved):

PR #5616, opened 2024-11-11
Use ToChatCompletion / ToStreamingChatCompletionUpdates in CachingChatClient

stephentoub:
Adds a ToStreamingChatCompletionUpdates method that's the counterpart to the recently added ToChatCompletion.

Then uses both from CachingChatClient instead of its now bespoke coalescing implementation. When coalescing is enabled (the default), CachingChatClient caches everything as a ChatCompletion, rather than distinguishing streaming and non-streaming.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5616)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs
```diff
@@ -87,4 +87,53 @@ public ChatMessage Message
...
+            };
+
+            if (Usage is { } usage)
+            {
+                extra.Contents.Add(new UsageContent(usage));
+            }
+        }
+
+        int choicesCount = Choices.Count;
+        var updates = new StreamingChatCompletionUpdate[choicesCount + Convert.ToInt32(extra is not null)];
```
eiriktsarpalis, line 110: Why not just?
```suggestion
        var updates = new StreamingChatCompletionUpdate[choicesCount + (extra is null ? 0 : 1)];
```

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs
stephentoub: That's exactly what ToInt32 is

PR #5617, opened 2024-11-11
Tweak ChatMessage/StreamingChatCompletionUpdate.ToString

stephentoub:
Include all text rather than just the first text content.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5617)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
 using System.Text.Json.Serialization;
 using Microsoft.Shared.Diagnostics;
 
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs
```diff
@@ -86,5 +86,5 @@ public ChatMessage Message
 
     /// <inheritdoc />
     public override string ToString() =>
-        Choices is { Count: > 0 } choices ? string.Join(Environment.NewLine, choices) : string.Empty;
+        string.Join(Environment.NewLine, Choices);
```
eiriktsarpalis, line 89: Assuming individual choices contain multi-line content, wouldn't this make it difficult to distinguish between them?

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatCompletion.cs
stephentoub: Yeah. I could add a prefix when there are multiple, e.g.
```
Choice 0: ...
Choice 1: ...
```
or something like that.

eiriktsarpalis: I like that better. Perhaps also consider indenting individual choices if multi-line content?

PR #5619, opened 2024-11-11
Expose options for making schema generation conformant with the subset accepted by OpenAI.

### LINKED ISSUE #5612: [API Proposal]: Microsoft.Extensions.AI  - Use structured output with function calling
### Background and motivation

Allow developers to use structured output with function calling : https://platform.openai.com/docs/guides/function-calling#function-calling-with-structured-outputs

Open AI says:

"By default, when you use function calling, the API will offer best-effort matching for your parameters, which means that occasionally the model may miss parameters or get their types wrong when using complicated schemas.

[Structured Outputs](https://platform.openai.com/docs/guides/structured-outputs) is a feature that ensures model outputs for function calls will exactly match your supplied schema.

Structured Outputs for function calling can be enabled with a single parameter, just by supplying strict: true"

### API Proposal

```csharp
public static AIFunction Create(
      Delegate method,
bool isStrict,
      string? name = null,
      string? description = null,
      JsonSerializerOptions? serializerOptions = null)
    {
      ...
    }

```

### API Usage

AIFunctionFactory.Create(Tools.TimeTools.ConvertDateTime,strict:true), //set strict to true

### Alternative Designs

_No response_

### Risks

_No response_

eiriktsarpalis:
Makes the following changes:

* Exposes additional configuration settings in the built-in schema generator making it conformant with OpenAI constraints.
* Adds a `SchemaCreateOptions` property to the `AIFunctionFactoryCreateOptions` class for controlling schema generation in `AIFunctionFactory`.
* Defaults `AIFunctionFactory` to using configuration that is conformant with OpenAI restrictions.

Fix #5612.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5619)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -27,4 +27,25 @@ public sealed class AIJsonSchemaCreateOptions
     /// Gets a value indicating whether to include the $schema keyword in inferred schemas.
     /// </summary>
     public bool IncludeSchemaKeyword { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating whether to mark all properties as required in the schema.
+    /// </summary>
+    public bool RequireAllProperties { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating whether to filter keywords that are disallowed by certain AI vendors.
+    /// </summary>
+    /// <remarks>
+    /// Filters a number of non-essential schema keywords that are not yet supported by some AI vendors.
+    /// These include:
+    /// <list type="bullet">
+    /// <item>The "minLength", "maxLength", "pattern", and "format" keywords.</item>
+    /// <item>The "minimum", "maximum", and "multipleOf" keywords.</item>
+    /// <item>The "patternProperties", "unevaluatedProperties", "propertyNames", "minProperties", and "maxProperties" keywords.</item>
+    /// <item>The "unevaluatedItems", "contains", "minContains", "maxContains", "minItems", "maxItems", and "uniqueItems" keywords.</item>
+    /// </list>
+    /// See also https://platform.openai.com/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported.
+    /// </remarks>
+    public bool FilterDisallowedKeywords { get; init; }
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -16,12 +16,12 @@ public sealed class AIJsonSchemaCreateOptions
     /// <summary>
     /// Gets a value indicating whether to include the type keyword in inferred schemas for .NET enums.
     /// </summary>
-    public bool IncludeTypeInEnumSchemas { get; init; }
+    public bool IncludeTypeInEnumSchemas { get; init; } = true;
 
     /// <summary>
     /// Gets a value indicating whether to generate schemas with the additionalProperties set to false for .NET objects.
     /// </summary>
-    public bool DisallowAdditionalProperties { get; init; }
+    public bool DisallowAdditionalProperties { get; init; } = true;
 
     /// <summary>
     /// Gets a value indicating whether to include the $schema keyword in inferred schemas.
@@ -31,7 +31,7 @@ public sealed class AIJsonSchemaCreateOptions
     /// <summary>
     /// Gets a value indicating whether to mark all properties as required in the schema.
     /// </summary>
-    public bool RequireAllProperties { get; init; }
+    public bool RequireAllProperties { get; init; } = true;
 
     /// <summary>
     /// Gets a value indicating whether to filter keywords that are disallowed by certain AI vendors.
@@ -47,5 +47,5 @@ public sealed class AIJsonSchemaCreateOptions
     /// </list>
     /// See also https://platform.openai.com/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported.
     /// </remarks>
-    public bool FilterDisallowedKeywords { get; init; }
+    public bool FilterDisallowedKeywords { get; init; } = true;
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 #if !NET9_0_OR_GREATER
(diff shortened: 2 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -27,4 +27,25 @@ public sealed class AIJsonSchemaCreateOptions
...
+    /// These include:
+    /// <list type="bullet">
+    /// <item>The "minLength", "maxLength", "pattern", and "format" keywords.</item>
+    /// <item>The "minimum", "maximum", and "multipleOf" keywords.</item>
+    /// <item>The "patternProperties", "unevaluatedProperties", "propertyNames", "minProperties", and "maxProperties" keywords.</item>
+    /// <item>The "unevaluatedItems", "contains", "minContains", "maxContains", "minItems", "maxItems", and "uniqueItems" keywords.</item>
+    /// </list>
+    /// See also https://platform.openai.com/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported.
+    /// </remarks>
+    public bool FilterDisallowedKeywords { get; init; }
```
stephentoub, line 50: Should this instead be an enumerable / list of keywords to filter? It feels a bit skewed that we're hardcoding which keywords are disallowed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -275,11 +287,35 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext ctx, JsonNode schema)
...
                     objSchema.Add(AdditionalPropertiesPropertyName, (JsonNode)false);
                 }
 
+                // Mark all properties as required
+                if (key.RequireAllProperties &&
+                    objSchema.TryGetPropertyValue(PropertiesPropertyName, out JsonNode? properties) &&
+                    properties is JsonObject propertiesObj)
+                {
+                    _ = objSchema.TryGetPropertyValue(RequiredPropertyName, out JsonNode? required);
+                    if (required is not JsonArray { } requiredArray || requiredArray.Count != propertiesObj.Count)
```
stephentoub, line 303: Is it possible they could have the same length but not actually represent all the props, e.g. if the same prop were included twice?

test/Libraries/Microsoft.Extensions.AI.Tests/Functions/AIFunctionFactoryTest.cs
```diff
@@ -182,4 +182,17 @@ public void AIFunctionFactoryCreateOptions_ValuesPropagateToAIFunction()
...
+    {
+        var options = new AIFunctionFactoryCreateOptions();
+        var schemaOptions = options.SchemaCreateOptions;
+
+        Assert.NotNull(schemaOptions);
+        Assert.True(schemaOptions.IncludeTypeInEnumSchemas);
+        Assert.True(schemaOptions.FilterDisallowedKeywords);
+        Assert.True(schemaOptions.RequireAllProperties);
+        Assert.True(schemaOptions.DisallowAdditionalProperties);
+    }
```
stephentoub, line 197: Could we add an integration test that exercises this?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
eiriktsarpalis: I tried that initially, but putting any sort of collection in the options messes up with our schema caching strategy which keys on the configuration settings. I could try to do that, but it would require building a custom list with sequence equality semantics.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: Only if there is a bug in the underlying `JsonSchemaExporter`.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryCreateOptions.cs
```diff
@@ -16,6 +16,13 @@ namespace Microsoft.Extensions.AI;
 public sealed class AIFunctionFactoryCreateOptions
 {
     private JsonSerializerOptions _options = AIJsonUtilities.DefaultOptions;
+    private AIJsonSchemaCreateOptions _schemaCreateOptions = new()
+    {
+        IncludeTypeInEnumSchemas = true,
```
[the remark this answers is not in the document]
eiriktsarpalis: @stephentoub thoughts on this point?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
stephentoub: I see. I'm concerned that this will quickly get stale. Compat-wise, this will effectively always need to represent the same set of keywords, even as constraints evolve.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryCreateOptions.cs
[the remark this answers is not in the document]
stephentoub: It'd be nice to have the same defaults at all levels.

test/Libraries/Microsoft.Extensions.AI.Tests/Functions/AIFunctionFactoryTest.cs
stephentoub: Ok

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryCreateOptions.cs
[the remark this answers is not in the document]
eiriktsarpalis: Done.

eerhardt:
FYI - this introduces new trim/AOT warnings. I'm enabling the CI check in #5622.

```
    D:\git\extensions\src\Libraries\Microsoft.Extensions.AI.Abstractions\Utilities\AIJsonUtilities.Schema.cs(305): Trim analysis error IL2026: Microsoft.Extensions.AI.AIJsonUtilities.<>c__DisplayClass15_0.<GetJsonSchemaCore>g__TransformSchemaNode|0(JsonSchemaExporterContext,JsonNode): Using member 'System.Text.Json.Nodes.JsonArray.Add<String>(String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Creating JsonValue instances with non-primitive types is not compatible with trimming. It can result in non-primitive types being serialized, which may have their members trimmed. [D:\git\extensions\test\Libraries\Microsoft.Extensions.AotCompatibility.TestApp\Microsoft.Extensions.AotCompatibility.TestApp.csproj]
    D:\git\extensions\src\Libraries\Microsoft.Extensions.AI.Abstractions\Utilities\AIJsonUtilities.Schema.cs(305): AOT analysis error IL3050: Microsoft.Extensions.AI.AIJsonUtilities.<>c__DisplayClass15_0.<GetJsonSchemaCore>g__TransformSchemaNode|0(JsonSchemaExporterContext,JsonNode): Using member 'System.Text.Json.Nodes.JsonArray.Add<String>(String)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. Creating JsonValue instances with non-primitive types requires generating code at runtime. [D:\git\extensions\test\Libraries\Microsoft.Extensions.AotCompatibility.TestApp\Microsoft.Extensions.AotCompatibility.TestApp.csproj]
```

PR #5621, opened 2024-11-11
Move extension methods in same assembly to be instance methods

stephentoub:
Each of the middleware components for IChatClient / IEmbeddingGenerator currently has an additional class, each of which contains a single extension method for the corresponding builder. But they type being extended is in the same assembly; these may as well just be instance methods and avoid all the extra types (implementations in external assemblies would still use extensions, of course). Bonus, the code gets a bit cleaner as we don't have the extra arguments, arg validation, generic parameters, etc.


eiriktsarpalis (approved):
LGTM, although I often find that it's useful for a method to show up in Intellisense as an extension. It provides signal that it's a true extension as opposed to being an integral part of the type's API or something that accesses internal data.

SteveSandersonMS:
This isn't personally how I'd do it but won't object.

My preference is to use extension method to show a clean boundary around what's innate to the interface and what are the external additions to it. This is partly to make clear to app/library authors that these kinds of extensions are meant to be added from outside and that anyone can do it - it's not us using our special privilege to build these kinds of middleware (use of `internal` APIs notwithstanding). I guess that's similar to @eiriktsarpalis's point above.

PR #5627, opened 2024-11-12
Fix linker warning.

### LINKED ISSUE #5626: Microsoft.Extensions.AI.Abstractions is raising trim/AOT warnings
After https://github.com/dotnet/extensions/pull/5619 was merged, new warnings are coming from `Microsoft.Extensions.AI.Abstractions` when trimming/AOT'ing:

```
    D:\git\extensions\src\Libraries\Microsoft.Extensions.AI.Abstractions\Utilities\AIJsonUtilities.Schema.cs(305): Trim analysis error IL2026: Microsoft.Extensions.AI.AIJsonUtilities.<>c__DisplayClass15_0.<GetJsonSchemaCore>g__TransformSchemaNode|0(JsonSchemaExporterContext,JsonNode): Using member 'System.Text.Json.Nodes.JsonArray.Add<String>(String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Creating JsonValue instances with non-primitive types is not compatible with trimming. It can result in non-primitive types being serialized, which may have their members trimmed. [D:\git\extensions\test\Libraries\Microsoft.Extensions.AotCompatibility.TestApp\Microsoft.Extensions.AotCompatibility.TestApp.csproj]
    D:\git\extensions\src\Libraries\Microsoft.Extensions.AI.Abstractions\Utilities\AIJsonUtilities.Schema.cs(305): AOT analysis error IL3050: Microsoft.Extensions.AI.AIJsonUtilities.<>c__DisplayClass15_0.<GetJsonSchemaCore>g__TransformSchemaNode|0(JsonSchemaExporterContext,JsonNode): Using member 'System.Text.Json.Nodes.JsonArray.Add<String>(String)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. Creating JsonValue instances with non-primitive types requires generating code at runtime. [D:\git\extensions\test\Libraries\Microsoft.Extensions.AotCompatibility.TestApp\Microsoft.Extensions.AotCompatibility.TestApp.csproj]
```

cc @eiriktsarpalis

eiriktsarpalis:
Fix #5626.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
@@ -302,7 +302,7 @@ public static partial class AIJsonUtilities
                     _ = objSchema.TryGetPropertyValue(RequiredPropertyName, out JsonNode? required);
                     if (required is not JsonArray { } requiredArray || requiredArray.Count != propertiesObj.Count)
                     {
-                        requiredArray = [.. propertiesObj.Select(prop => prop.Key)];
+                        requiredArray = [.. propertiesObj.Select(prop => (JsonNode)prop.Key)];
                         objSchema[RequiredPropertyName] = requiredArray;
                     }
                 }


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -302,7 +302,7 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext ctx, JsonNode schema)
                     _ = objSchema.TryGetPropertyValue(RequiredPropertyName, out JsonNode? required);
                     if (required is not JsonArray { } requiredArray || requiredArray.Count != propertiesObj.Count)
                     {
-                        requiredArray = [.. propertiesObj.Select(prop => prop.Key)];
+                        requiredArray = [.. propertiesObj.Select(prop => (JsonNode)prop.Key)];
```
eiriktsarpalis, line 305: Casting to `JsonNode` results in the non-generic `JsonArray.Add` method being used which is linker-safe.

stephentoub: Why is the non-generic safe but the generic isn't? i.e. what is the explicit cast to JsonNode doing that's safe and that can't be done in the generic Add?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eerhardt: Can we add a quick comment explaining why the cast is necessary?

eiriktsarpalis: > Why is the non-generic safe but the generic isn't? i.e. what is the explicit cast to JsonNode doing that's safe and that can't be done in the generic Add?

The generic `Add` method encapsulates arbitrary .NET values and uses reflection to lazily serialize them to JSON. The cast uses the explicit conversions from known primitives to `JsonNode` which doesn't suffer from the same problem.

> Can we add a quick comment explaining why the cast is necessary?

If you search for casts to `JsonNode` in this source file, you'll see that this pattern is pervasive both here and generally any AOT compatible library that works with `JsonNode`. It sucks, but that's what got shipped before trimmer annotations became a design concern.

Issue #4377, opened 2023-09-07
Add truncation notice in logs


sebastienros:
When request/response bodies are truncated in logs (Http client logging and Http middleware logging) it should be visible that the field doesn't represent the whole body. The person trying to debug the code might otherwise believe the body not incomplete and assume this is an error.

dariusclay:
Closing as stale due to work in #5628

PR #5628, opened 2024-11-12
Allow logging of body without modifying the actual response

### LINKED ISSUE #5511: AddExtendedHttpClientLogging is truncating body response when it mustn't do it
### Description

I'm querying an API that returns a big json (around 250KB) and the application which makes this request have configured response body logging using the `AddExtendedHttpClientLogging` method from the Microsoft.Extensions.Http.Diagnostics 8.10.0 nuget library. For some reason, the body of the response is truncated when reading it from the HttpResponseMessage so I am not able to process the response. If i disable body logging the response is not truncated at all.

I tried a very small json response like around 1KB and the response is not truncated at all. However, I tried a json around 64KB and it is truncated. 

I tried increasing the body limit from the default value to 1.5MB when calling the method `AddExtendedHttpClientLogging`, which it shouldnt be necessary at all to be honest, and the 64KB test is no longer truncated so it is something related to this configuration. As far as I understand the library can truncate the emitted log but it shouldnt being truncating the final response.

### Reproduction Steps

[ConsoleApp2.zip](https://github.com/user-attachments/files/17343138/ConsoleApp2.zip)

### Expected behavior

Final response that you read from a HttpResponseMessage shouldnt be truncated in any scenario. Logs emitted by the library can be truncated depending on what you configured.

### Actual behavior

Final response that you read from a HttpResponseMessage is being truncated depending on what you configured in the library.

### Regression?

Cant say it.

### Known Workarounds

Havent found any, just stop using the library and do your own logging class.

### Configuration

I attached a console app in this issue but is also happening in an asp.net application.

Partial result of `dotnet --info`
.NET SDK:
 Version:           8.0.403
 Commit:            c64aa40a71
 Workload version:  8.0.400-manifests.e99c892e
 MSBuild version:   17.11.9+a69bbaaf5

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26100
 OS Platform: Windows
 RID:         win-x64

### Other information

_No response_

dariusclay:
Closes #5511 
Makes #4377 obsolete
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5628)


src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -54,90 +51,186 @@ public ValueTask<string> ReadAsync(HttpResponseMessage response, CancellationTok
...
 
-        var bufferWriter = _bufferWriterPool.Get();
-        var memory = bufferWriter.GetMemory(readSizeLimit).Slice(0, readSizeLimit);
-#if !NETCOREAPP3_1_OR_GREATER
-        byte[] buffer = memory.ToArray();
-#endif
-        try
+        var pipe = new Pipe();
+
+        var bufferedString = await BufferStreamAndWriteToPipeAsync(streamToReadFrom, pipe.Writer, readSizeLimit, cancellationToken).ConfigureAwait(false);
```
RussKie, line 95: [nit] No var here please, the type is not obvious

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -54,90 +51,186 @@ public ValueTask<string> ReadAsync(HttpResponseMessage response, CancellationTok
...
-#if NETCOREAPP3_1_OR_GREATER
-            var charsWritten = await streamToReadFrom.ReadAsync(memory, cancellationToken).ConfigureAwait(false);
-            bufferWriter.Advance(charsWritten);
-            return Encoding.UTF8.GetString(memory.Slice(0, charsWritten).Span);
+            await WriteStreamToPipeAsync(streamToReadFrom, pipe.Writer, cancellationToken).ConfigureAwait(false);
+        }, CancellationToken.None).ConfigureAwait(false);
+
+        // use the pipe reader as stream for the new content
+        var newContent = new StreamContent(pipe.Reader.AsStream());
+        foreach (var header in response.Content.Headers)
```
RussKie, line 116: Ditto

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -54,90 +51,186 @@ public ValueTask<string> ReadAsync(HttpResponseMessage response, CancellationTok
...
+
+        response.Content = newContent;
+
+        return bufferedString;
+    }
+
+#if NET6_0_OR_GREATER
+    private static async Task<string> BufferStreamAndWriteToPipeAsync(Stream stream, PipeWriter writer, int bufferSize, CancellationToken cancellationToken)
+    {
+        var memory = writer.GetMemory(bufferSize)[..bufferSize];
```
RussKie, line 129: Ditto

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -54,90 +51,186 @@ public ValueTask<string> ReadAsync(HttpResponseMessage response, CancellationTok
...
+        var sb = new StringBuilder();
 
-                await freshStream.WriteAsync(buffer, 0, writtenCount + bufferWriter.WrittenCount, cancellationToken).ConfigureAwait(false);
-#endif
-                freshStream.Seek(0, SeekOrigin.Begin);
+        int bytesRead = 0;
+
+        while (bytesRead < bufferSize)
+        {
+            var chunkSize = Math.Min(ChunkSize, bufferSize - bytesRead);
```
RussKie, line 189: Ditto

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -54,90 +51,186 @@ public ValueTask<string> ReadAsync(HttpResponseMessage response, CancellationTok
...
+            streamToReadFrom.Seek(0, SeekOrigin.Begin);
+
+            await pipe.Reader.CompleteAsync().ConfigureAwait(false);
+
+            return bufferedString;
+        }
+
+        // if stream is not seekable we need to write the rest of the stream to the pipe
+        // and create a new response content with the pipe reader as stream
+        _ = Task.Run(async () =>
```
iliar-turdushev, line 109: Why do we need to start a task here?

iliar-turdushev: We are calling `ConfigureAwait(false)` but don't await the task. Is that correct?

dariusclay: This approach leverages `Pipe` for efficient memory management and non-blocking operations. By consuming the original stream in a background task and writing it to the Pipe, we can quickly return control to the caller. The response content is replaced with a new `StreamContent` supplied by `PipeReader.AsStream()`, allowing the application to handle large responses asynchronously and efficiently. This method improves performance and scalability by utilizing Pipe's buffer management and concurrent read/write capabilities.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
dariusclay: Done, thanks

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
iliar-turdushev: Got it, then we don't need `ConfigurateAwait(false)` there since we don't await the task, right? Or am I missing something?

dariusclay: It was removed

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -54,90 +53,186 @@ public ValueTask<string> ReadAsync(HttpResponseMessage response, CancellationTok
...
         }
-        finally
+
+        // if stream is not seekable we need to write the rest of the stream to the pipe
+        // and create a new response content with the pipe reader as stream
+        _ = Task.Run(async () =>
         {
-            if (streamToReadFrom.CanSeek)
+            await WriteStreamToPipeAsync(streamToReadFrom, pipe.Writer, cancellationToken).ConfigureAwait(false);
+        }, CancellationToken.None);
```
eduherminio, line 114: Is this preferred over reusing `cancellationToken`?

dariusclay: I originally had it passed both to the inner task to write the stream to pipe, and the background task. My thoughts here is the inner task gets cancelled the thread will terminate anyway. I may be wrong about this. What do we prefer here?

eduherminio: I'm not sure, we're using it in [`TcpEndpointProbesService`](https://github.com/dotnet/extensions/blob/09094aebc27b03eec7656bed9c128c5cc523df39/src/Libraries/Microsoft.Extensions.Diagnostics.Probes/TcpEndpointProbesService.cs#L49) like this, but not in [FunctionInvokingChatClient ](https://github.com/dotnet/extensions/blob/09094aebc27b03eec7656bed9c128c5cc523df39/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L441).

After doing some reading, If believe it doesn't really matter in our scenario, given we are using it for the internal call and we don't have [children tasks inside of this `Task.Run()`](https://learn.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-cancel-a-task-and-its-children#example), and we're relying on the status of the cancelled task (`cancelled` vs `faulted`).

There seem to be opinions both for and [against](https://blog.stephencleary.com/2022/02/cancellation-1-overview.html) using it, so your call.

eduherminio (approved):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
dariusclay: > The CancellationToken passed to Task.Run just cancels the scheduling of the delegate to the thread pool; once that delegate starts running (which happens pretty much immediately), that cancellation token is ignored.

To keep semantics, I'm leaving it as-is. That way we are clearly showing that the cancellation should happen for the delegate itself. If we pass it to `Task.Run` it would only prevent the scheduling of the task if it were already cancelled.

dariusclay:
@RussKie @iliar-turdushev do any of you have other comments or concerns about this change?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpResponseBodyReader.cs
```diff
@@ -22,15 +22,14 @@ internal sealed class HttpResponseBodyReader
     /// </summary>
     internal readonly TimeSpan ResponseReadTimeout;
 
-    private static readonly ObjectPool<BufferWriter<byte>> _bufferWriterPool = BufferWriterPool.SharedBufferWriterPool;
+    private const int ChunkSize = 8 * 1024;
```
RussKie, line 25: [nit] May worth adding a comment why 8KB size was chosen?

dariusclay: Done

darraghjones:
how can we find out what version of the library will this fix be in?

RussKie:
> how can we find out what version of the library will this fix be in?

9.1, which is expected to be shipped in Jan 25.

PR #5630, opened 2024-11-12
Replace STJ boilerplate in the leaf clients with AIJsonUtilities calls.

eiriktsarpalis:
With `AIJsonUtilities` being moved to M.E.AI.Abstractions many of these helpers aren't necessary. Also makes all `ToolCallJsonSerializerOptions` non-nullable.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5630)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
```diff
@@ -16,6 +17,19 @@ public static partial class AIJsonUtilities
     /// <summary>Gets the <see cref="JsonSerializerOptions"/> singleton used as the default in JSON serialization operations.</summary>
     public static JsonSerializerOptions DefaultOptions { get; } = CreateDefaultOptions();
 
+    /// <summary>
+    /// Gets a strongly typed <see cref="JsonTypeInfo{T}"/> instance for the specified <typeparamref name="T"/>.
+    /// </summary>
+    /// <typeparam name="T">The type for which to resolve the JSON contract.</typeparam>
+    /// <param name="options">The options from which to resolve the JSON metadata.</param>
+    /// <returns>A <see cref="JsonTypeInfo{T}"/> instance for the specified type.</returns>
+    public static JsonTypeInfo<T> GetTypeInfo<T>(this JsonSerializerOptions options)
```
eiriktsarpalis, line 26: On the fence on whether this should be an extension method for JSO. It's a fairly useful helper when building AOT compatible libraries.

stephentoub: I don't think M.E.AI should be in the business of supplying non-MEAI-related extension methods onto non-MEAI types.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -51,7 +51,7 @@ public AzureAIInferenceChatClient(ChatCompletionsClient chatCompletionsClient, s
     }
 
     /// <summary>Gets or sets <see cref="JsonSerializerOptions"/> to use for any serialization activities related to tool call arguments and results.</summary>
-    public JsonSerializerOptions? ToolCallJsonSerializerOptions { get; set; }
+    public JsonSerializerOptions ToolCallJsonSerializerOptions { get; set; } = AIJsonUtilities.DefaultOptions;
```
stephentoub, line 54: If we're going to depend on this being non-null, we should validate in the setter that the provided value is non-null.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -317,7 +317,7 @@ private ChatCompletionsOptions ToAzureAIOptions(IList<ChatMessage> chatContents,
                         default:
                             if (prop.Value is not null)
                             {
-                                byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, JsonContext.GetTypeInfo(prop.Value.GetType(), ToolCallJsonSerializerOptions));
+                                byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, ToolCallJsonSerializerOptions.GetTypeInfo(typeof(object)));
```
stephentoub, line 320: Why is this using typeof(object) rather than the actual type of the value?

stephentoub: Also, the JsonContext.GetTypeInfo method was first doing the lookup on the supplied JSO and then falling back to doing it on the default JSO. This is only doing it on the supplied one.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -449,7 +449,7 @@ private IEnumerable<ChatRequestMessage> ToAzureAIInferenceChatMessages(IEnumerab
                              callRequest.CallId,
                              new FunctionCall(
                                  callRequest.Name,
-                                 JsonSerializer.Serialize(callRequest.Arguments, JsonContext.GetTypeInfo(typeof(IDictionary<string, object>), ToolCallJsonSerializerOptions)))));
+                                 JsonSerializer.Serialize(callRequest.Arguments, ToolCallJsonSerializerOptions.GetTypeInfo(typeof(IDictionary<string, object>))))));
```
stephentoub, line 452: Similarly here. My intent was that if the supplied JSO didn't contain the relevant type, it'd fall back to using the built-in one. Why don't we want that?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceEmbeddingGenerator.cs
```diff
@@ -173,7 +173,7 @@ private EmbeddingsOptions ToAzureAIOptions(IEnumerable<string> inputs, Embedding
             {
                 if (prop.Value is not null)
                 {
-                    byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, JsonContext.GetTypeInfo(prop.Value.GetType(), null));
+                    byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(object)));
```
stephentoub, line 176: I'm not clear on why it's ok to change these from value.GetType() to typeof(object)

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
eiriktsarpalis: Does keeping it as a static seem fair? Or should we just remove it/replicate it in the leaf clients?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: > Why is this using typeof(object) rather than the actual type of the value?

It's equivalent. Serializing something as object has the serializer use the contract for the runtime type.

> Also, the JsonContext.GetTypeInfo method was first doing the lookup on the supplied JSO and then falling back to doing it on the default JSO. This is only doing it on the supplied one.

I removed that intentionally. Even if `JsonContext` does provide a `JsonTypeInfo<object>`, because of how `object` serialization works it will then immediately try to look up the runtime type from `JsonContext` only. For any type graph that isn't closed (i.e. contains nodes with `object`) we should default to consulting `ToolCallJsonSerializerOptions` so that users can specify serializers for any runtime types that may appear in the objects.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: See https://github.com/dotnet/extensions/pull/5630#discussion_r1839703731. We should use `ToolCallJsonSerializerOptions` for any type containing `object`.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: Done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
eiriktsarpalis: Removed.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -490,5 +498,6 @@ private static List<ChatMessageContentItem> GetContentParts(IList<AIContent> con
 
     private static FunctionCallContent ParseCallContentFromJsonString(string json, string callId, string name) =>
         FunctionCallContent.CreateFromParsedArguments(json, callId, name,
-            argumentParser: static json => JsonSerializer.Deserialize(json, JsonContext.Default.IDictionaryStringObject)!);
+            argumentParser: static json => JsonSerializer.Deserialize(json,
+                (JsonTypeInfo<IDictionary<string, object>>)AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(IDictionary<string, object>)))!);
```
stephentoub, line 502: We don't want to use ToolCallJsonSerializerOptions here as well?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -668,60 +674,19 @@ private static List<ChatMessageContentPart> GetContentParts(IList<AIContent> con
 
     private static FunctionCallContent ParseCallContentFromJsonString(string json, string callId, string name) =>
         FunctionCallContent.CreateFromParsedArguments(json, callId, name,
-            argumentParser: static json => JsonSerializer.Deserialize(json, JsonContext.Default.IDictionaryStringObject)!);
+            argumentParser: static json => JsonSerializer.Deserialize(json,
+                (JsonTypeInfo<IDictionary<string, object>>)AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(IDictionary<string, object>)))!);
```
stephentoub, line 678: Same question

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: We could, but the callback delegate would no longer be static. Deserializing `object` is much more straightforward compared to serializing; it just uses `JsonElement` as the runtime type and so the default works for all intents and purposes.

PR #5634, opened 2024-11-13
Add API allowing to disable retries for a given list of HTTP methods

### LINKED ISSUE #5248: Fix default HTTP resilience not to retry non-idempotent requests (POST, PUT, etc.)
## Original issue description added by @SteveSandersonMS 

Currently if a backend service takes a while to handle a request (e.g., > 5 sec), the default HTTP resilience policy will retry.

This is fine for idempotent requests (e.g., GET), but dangerous/harmful for other request types (POST, PUT). While building eShopSupport I often encountered cases where it would create duplicate messages/tickets just because either:

[1] I was debugging, and hence causing things to be slow
[2] Or, some lower-level service was still starting up and loading an AI model or similar, causing it to be temporarily delayed

It's fine for developers to configure a retry policy for POST/PUT/etc themselves, but I don't think the default should be to retry those and was very surprised when finding that it is.

---

### Background and motivation

[`Microsoft.Extensions.Http.Resilience`](https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience) package provides a delegating handler called `StandardResilienceHandler` that applies various resilience strategies (retry, timeout, rate limiting) to HttpClient. For example, it automatically makes retries of an HTTP request if it fails with a transient error. You can register `StandardResilienceHandler` for HttpClient as following:

```csharp
services.AddHttpClient("test").AddStandardResilienceHandler();
```

or you can use an overload that accepts options configuring resilience strategies:

```csharp
services.AddHttpClient("test").AddStandardResilienceHandler(options =>
{
    // You can configure resilience options here.
});
```

Each resilience strategy of `StandardResilienceHandler` has a predicate deciding whether to apply the strategy or not based on the outcome of an HTTP request, i.e. whether the outcome is a transient error or not. `Retry` strategy by default is configured to make retries for all HTTP methods, i.e. it will make retries for both idempotent and non-idempotent HTTP methods. That brings the issue described above at the beginning of this comment. While the current default behavior might be undesired for some services, we don't want to change it because:

* In general, we cannot make assumptions about idempotency of APIs based on HTTP methods being used to access those APIs.
* Such a change will be breaking and would lead to unpleasant results for existing services that already use `StandardResilienceHandler`.

As a compromise solution to the issue we would like to introduce a new API allowing to easily disable the default behavior and don't do retries for non-idempotent HTTP methods.

### API Proposal

```csharp
namespace Microsoft.Extensions.Http.Resilience;

/// <summary>
/// Extensions for HttpRetryStrategyOptionsExtensions.
/// </summary>
public static class HttpRetryStrategyOptionsExtensions
{
    /// <summary>
    /// Disables retries for POST, PATCH, PUT, DELETE, and CONNECT.
    /// </summary>
    /// <param name="option">Retry strategy options.</param>
    public static void DisableForUnsafeHttpMethods(this HttpRetryStrategyOptions option);

    /// <summary>
    /// Disables retries for the given list of HTTP methods.
    /// </summary>
    /// <param name="options">Retry strategy options.</param>
    /// <param name="methods">List of HTTP methods.</param>
    public static void DisableFor(this HttpRetryStrategyOptions options, params HttpMethod[] methods);
}
```

According to [RFC safe HTTP methods](https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1) are those whose semantics is read-only. POST, PATCH, PUT, DELETE, and CONNECT are unsafe. While for PUT and DELETE doing retries might be safe, we propose an API method `DisableForUnsafeHttpMethods` disabling retries for all unsafe HTTP methods, i.e. HTTP methods that could be harmful for the server. Additionally, to have some flexibility we want to introduce a method `DisableFor` allowing users to specify which HTTP methods their service considers as non-retriable.

### API Usage

`DisableForUnsafeHttpMethods`:

```csharp
services.AddHttpClient("test").AddStandardResilienceHandler(options =>
{
    options.Retry.DisableForUnsafeHttpMethods();
});
```

`DisableFor`:

```csharp
services.AddHttpClient("test").AddStandardResilienceHandler(options =>
{
    options.Retry.DisableFor(HttpMethod.Post, HttpMethod.Patch);
});
```

### Alternative Designs

Extend existing `AddStandardResilienceHandler` methods with an argument accepting a list of HTTP methods for which users don't want to do retries.

API:

```csharp
namespace Microsoft.Extensions.DependencyInjection;

public static partial class ResilienceHttpClientBuilderExtensions
{
    // Registers StandardResilienceHanlder in the given IHttpClientBuilder and configures its retry strategy to not do retries for the given list of HTTP methods.
    public static IHttpStandardResiliencePipelineBuilder AddStandardResilienceHandler(this IHttpClientBuilder builder, params HttpMethod[] nonIdempotentHttpMethods);

    // Like the first method + accepts a delegate to configure resilience options.
    public static IHttpStandardResiliencePipelineBuilder AddStandardResilienceHandler(this IHttpClientBuilder builder, Action<HttpStandardResilienceOptions> configure, params HttpMethod[] nonIdempotentHttpMethods);

    // Like the first method + accepts a configuration section to configure resilience options.
    public static IHttpStandardResiliencePipelineBuilder AddStandardResilienceHandler(this IHttpClientBuilder builder, IConfigurationSection section, params HttpMethod[] nonIdempotentHttpMethods);
}
```

Usage:

```csharp
services.AddHttpClient("test").AddStandardResilienceHandler(HttpMethod.Post, HttpMethod.Patch);
```

The downside of this approach is that:

- We have to duplicate each overload of `AddStandardResilienceHandler` method with an argument accepting a list of non-idempotent HTTP methods.
- A list of non-idempotent HTTP methods affects only `Retry` strategy, but we pass it to a method registering a resilience handler also combining other resilience strategies, e.g. timeout, rate limiting, circuit breaker.

### Risks

Not aware of any risks.

iliar-turdushev:
The PR introduces an API that allows to disable automatic retries for

1. a given list of HTTP methods;
2. a list of unsafe/non-readonly HTTP methods: POST, PATCH, PUT, DELETE, and CONNECT.

Here is the approved API https://github.com/dotnet/extensions/issues/5248#issuecomment-2457909322.

The documentation will be added in a separate PR https://github.com/dotnet/docs/pull/43630.

Closes #5248.


src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+#if !NET8_0_OR_GREATER
+    private static readonly HttpMethod _connect = new("CONNECT");
+    private static readonly HttpMethod _patch = new("PATCH");
+#endif
+
+    /// <summary>
+    /// Disables retry attempts for POST, PATCH, PUT, DELETE, and CONNECT HTTP methods.
+    /// </summary>
+    /// <param name="options">The retry strategy options.</param>
+    public static void DisableForUnsafeHttpMethods(this HttpRetryStrategyOptions options)
```
iliar-turdushev, line 29: Alternatively, we can return `HttpRetryStrategyOptions` to make this and other extension-methods configuring `HttpRetryStrategyOptions` "chainable". Honestly, at the moment I don't see a necessity for that, but probably in future we'll want to introduce more API and make them fluent.

I would personally leave the API as-is returning `void`. What do you think?

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        var shouldHandle = options.ShouldHandle;
+
+        options.ShouldHandle = async args =>
+        {
+            var result = await shouldHandle(args).ConfigureAwait(args.Context.ContinueOnCapturedContext);
+
+            if (result &&
+                args.Outcome.Result is HttpResponseMessage response &&
+                response.RequestMessage is HttpRequestMessage request)
```
iliar-turdushev, line 58: We need this condition to satisfy the compiler because `RequestMessage` is nullable `HttpRequestMesssage?`. I don't think in reality it could be `null`. Alternatively, we can throw an exception here, but I think it is better not to fail here, but rather let the original `ShouldHandle` predicate to take a decision whether to do retries or not.

iliar-turdushev, line 58: We need this condition to satisfy the compiler because `RequestMessage` is nullable `HttpRequestMesssage?`. I don't think in reality it could be `null`. Alternatively, we can throw an exception here, but I think it is better not to fail here, but rather let the original `ShouldHandle` predicate to take a decision whether to do retries or not.

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
RussKie: This is more a question for the API Review Board. Here are the general rules: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md.

I _think_ that changing the return type from `void` to `T` should be an acceptable change in the future, but I may be wrong... @terrajobst @bartonjs what's your take on this?

bartonjs: > I think that changing the return type from void to T should be an acceptable change in the future, but I may be wrong...

Methods in .NET are named for their full signature, including the declared return type.  So any callers from before you change it will get a MissingMethodException if they run against a library after you change it.

If you want to change it between previews that will annoy your users, but you can do it (with a breaking change notice).  Once you abandon the preview label it is what it is forever (you'd have to leave this one as-is and make some other overload that happens to return the data, but clearly can't have the same calling signature).

RussKie: Thank you, @bartonjs for clarifications.

bartonjs: FWIW, in that breaking change guidelines doc, it says it's OK to change the return _value_, but that it's not OK to change the return _type_.

> Returning a value of a more derived type for a property, field, return or out value
>
> **Note, again, that the static type cannot change.** e.g. it is OK to return a string instance where an object was returned previously, but it is not OK to change the return type from object to string.

<sub>Emphasis mine</sub>

iliar-turdushev:
@martintmk @SteveSandersonMS Please, review this PR and the PR with documentation https://github.com/dotnet/docs/pull/43630. Thank you.

dariusclay (approved):

mobratil (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Disables retry attempts for the given list of HTTP methods.
+    /// </summary>
+    /// <param name="options">The retry strategy options.</param>
+    /// <param name="methods">The list of HTTP methods.</param>
+    public static void DisableFor(this HttpRetryStrategyOptions options, params HttpMethod[] methods)
+    {
+        _ = Throw.IfNull(options);
+        _ = Throw.IfNullOrEmpty(methods);
+
+        var shouldHandle = options.ShouldHandle;
```
martintmk, line 50: nit: To prevent runtime errors, this should be validated to be not null. 

Albeit we have nullable annotations here, one can set `ShouldHandle` to `null` before calling `DisableFor`

martintmk (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
iliar-turdushev: Good point, I'll add a check ensuring `options.ShouldHandle` is not null.

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
iliar-turdushev: Let's keep the methods void. I searched dotnet/extensions and dotnet/runtime repositories, and didn't find similar public Extensions API for Options class, so there is no pattern we can follow. Usually, users directly configure Options objects by setting their properties, and API we are introducing is kind of helper-methods that allow users to easily set up existing property of the RetryOptions object. I'm not sure there will be a need to make such API "fluent" in future, but if there is we can introduce new API. For now, let's keep the API simpler, i.e. return `void`.

SteveSandersonMS (approved):
Looks great!

src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptionsExtensions.cs
iliar-turdushev: If nobody has objections, I'll proceed with current API.

PR #5632, opened 2024-11-12
Merge branch release/9.0 into main

joperezr:
This should be merged and not squashed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5632)

PR #5640, opened 2024-11-14
Generalise how package version isn't stabilized for dev/preview stages

RussKie:
Follow up on #5632
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5640)

### DIFF
--- a/eng/MSBuild/ProjectStaging.targets
+++ b/eng/MSBuild/ProjectStaging.targets
@@ -4,6 +4,10 @@
     <!-- Preview packages: do not use stable branding and do not warn about lack of [Experimental] -->
     <SuppressFinalPackageVersion Condition="'$(Stage)' == 'preview'">true</SuppressFinalPackageVersion>
     <NoWarn Condition="'$(Stage)' == 'preview'">$(NoWarn);LA0003</NoWarn>
+
+    <!-- Amend the description based on stage -->
+    <Description Condition="'$(Stage)' == 'dev'">Experimental package. $(Description)</Description>
+    <Description Condition="'$(Stage)' == 'obsolete'">Obsolete Package. $(Description)</Description>
   </PropertyGroup>
 
   <!-- Produce errors if we don't have all the right properties defined -->
@@ -22,10 +26,11 @@
     <Error Condition="'$(MinMutationScore)' != 'n/a' AND ('$(MinMutationScore)' &lt; 50)" Text="MinMutationScore property must be >= 50 for normal stage." />
   </Target>
 
-  <!-- Amend the description based on stage -->
-  <PropertyGroup>
-    <Description Condition="'$(Stage)' == 'dev'">Experimental package. $(Description)</Description>
-    <Description Condition="'$(Stage)' == 'obsolete'">Obsolete Package. $(Description)</Description>
-  </PropertyGroup>
+  <Target Name="_ValidateVersion" AfterTargets="GenerateNuspec" Condition="'$(Stage)' == 'dev' or '$(Stage)' == 'preview'">
+    <PropertyGroup>
+      <_ExpectedVersionSuffix>$(_PreReleaseLabel)$(_BuildNumberLabels)</_ExpectedVersionSuffix>
+    </PropertyGroup>
 
+    <Error Condition=" '$(VersionSuffix)' != '$(_ExpectedVersionSuffix)' " Text="Unexpected package version suffix. Expected suffix for '$(Stage)' stage: '$(_ExpectedVersionSuffix)', received: '$(VersionSuffix)'." />
+  </Target>
 </Project>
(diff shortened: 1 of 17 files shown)


eng/MSBuild/ProjectStaging.targets
```diff
@@ -28,4 +28,10 @@
     <Description Condition="'$(Stage)' == 'obsolete'">Obsolete Package. $(Description)</Description>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <SuppressFinalPackageVersion>false</SuppressFinalPackageVersion>
```
joperezr, line 32: NIT: Not super important, but in arcade the default is actually to have unset, rather than set to false. Doesn't matter unless there are checks that do `SuppressFinalPackageVersion` != ``
```suggestion
    <SuppressFinalPackageVersion />
```

joperezr (commented):
Thank you so much for the help with this @RussKie! this is great. Given these properties could be impacted by race condition, can we also add a target that runs after GenerateNuspec that validates that version contains a prerelease label?

src/Generators/Microsoft.Gen.MetricsReports/Directory.Build.props
```diff
@@ -4,7 +4,6 @@
   the project itself. -->
   <PropertyGroup>
     <Stage>dev</Stage>
-    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
iliar-turdushev, line 7: Do we also need to remove `SuppressFinalPackageVersion` here?
 https://github.com/dotnet/extensions/blob/7ebb34d4b5a0d1e18c4323cbb23a7e1b6c4622bc/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj#L28

RussKie: Yes, looks like I missed that. Thanks.
I found it another day performing the validations locally, but I haven't had a chance to update the PR as I'm very busy with other stuff.

RussKie:
> build.cmd -ci -pack /p:OfficialBuildId=20241212.01

This command doesn't work (don't ask me how I know).
```
 D:\Development\dotnet-extensions\.dotnet\sdk\9.0.100-rtm.24479.2\NuGet.RestoreEx.targets(19,5): error :
      '9.1.0-preview.1.24612.01' is not a valid version string. (Parameter 'value')
```

The following do:
```
.\build.cmd -ci -restore
.\build.cmd -ci -build /p:OfficialBuildId=20241120.4
.\build.cmd -ci -pack /p:Restore=false /p:Build=false /p:OfficialBuildId=20241120.4
```

I added a pack-time check to ensure the version suffix is correct:
image

eng/Versions.props
```diff
@@ -8,13 +8,21 @@
...
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <!--
+      Makes it such that the package version won't be stabilized even when the rest of the repo is going stable.
+      https://github.com/dotnet/arcade/blob/main/Documentation/CorePackages/Versioning.md#package-version
+     -->
+    <SuppressFinalPackageVersion />
+    <SuppressFinalPackageVersion Condition="'$(Stage)' == 'dev' or '$(Stage)' == 'preview'">true</SuppressFinalPackageVersion>
+  </PropertyGroup>
```
RussKie, line 24: Version suffix calculations are performed in Version.BeforeCommonTargets.targets, which too late for the functionality placed in eng/MSBuild/ProjectStaging.targets
Arguably, we could add eng/MSBuild/ProjectStaging.props, but it feels more logical to have versioning-related configurations in one place (which is here).
image

joperezr: Will Stage be defined this early though?

joperezr: Oh, I see it will because you are creating a Directory.Build.props everywhere to make sure you control the order of imports. While that would work, I'd argue that we are probably defeating the point of the benefit of the change. The initial motivation was to simplify things, and reducing 2 lines in projects (setting the stage, and setting suppressFinalVersion) to a single one, but this is instead now making us having to add new files per project, and it also means that now you need to know to look into this file to see if a package is stable or not and also to know what is the stage.

Thoughts?

joperezr: Probably an even more meta question (sorry for having this discussion in your PR) is: do we even really need package stages? What benefit are we getting out of them? if it is literally just that the description has a few other words as a prefix, then perhaps we should just have stable and unstable packages, and adjust the descriptions of packages accordingly. That way, we can go back to having a single property in the package that marks it's "stableness"

RussKie: > Probably an even more meta question (sorry for having this discussion in your PR) is: do we even really need package stages? What benefit are we getting out of them? if it is literally just that the description has a few other words as a prefix, then perhaps we should just have stable and unstable packages, and adjust the descriptions of packages accordingly. That way, we can go back to having a single property in the package that marks it's "stableness"

The stages drive the test coverage requirements
https://github.com/dotnet/extensions/blob/c08e5ac737d0830ff83c1c5ae8b084e6fce2b538/eng/MSBuild/ProjectStaging.targets#L19-L23

joperezr: Right but that could easily just be controlled without stages, right? You can basically just change that logic to read whether a package is stable or not via supressfinalpackageversion

RussKie:
Superseded by https://github.com/dotnet/extensions/pull/5864

PR #5642, opened 2024-11-14
Change ChatClientBuilder to register singletons and support lambda-less chaining

### LINKED ISSUE #5572: Consider alternate chaining pattern for DI/builder
(Written by @SteveSandersonMS; I'm copying this into the repo from an offline location...)

---

## Now

Currently the API looks like:

```cs
services.AddChatClient(pipeline => pipeline
    .UseDistributedCache()
    .UseFunctionCalling()
    .Use(new SomeConcreteClient()); // Or pass a factory (Func<IServiceProvider, IChatClient>)
```

... and the proposed Aspire+MEAI integration looks like:

```cs
services.AddOpenAIChatClient("service-name", pipeline => pipeline
    .UseDistributedCache()
    .UseFunctionCalling());
```

## Possibility

However, we could change `AddChatClient` so that it actually returns the `ChatClientBuilder` instead of the `IServiceCollection`. Then it would look like:

```cs
services.AddChatClient(new SomeConcreteClient()) // Or pass a factory (Func<IServiceProvider, IChatClient>)
    .UseDistributedCache()
    .UseFunctionCalling();
```

... and the proposed Aspire+MEAI integration would become:

```cs
services.AddOpenAIChatClient("service-name")
    .UseDistributedCache()
    .UseFunctionCalling();
```

In my opinion this looks a lot better, as it removes the lambda, nested brackets, and the meaningless-looking `pipeline => pipeline` that hangs at the end of the first line.

This would be a legitimate pattern for DI with plenty of prior art of `AddX` calls that return a builder you can chain further setup steps on.

## Challenges

* We'd have to change `ChatClientBuilder` so that, instead of accepting `IServiceProvider` as a constructor arg and exposing it as a property, it would have to work without access to an `IServiceProvider` until the DI service is resolved.
  * Mostly this is fine because the `Use(Func<ChatClientBuilder, ChatClientBuilder>)` method already has an overload that takes `Func<IServiceProvider, ChatClientBuilder, ChatClientBuilder>` and so that would become the only overload
  * The one thing that would have to change is its `Use(IChatClient)` that terminates the pipeline. This would have to change to `Use(Func<IServiceProvider, IChatClient>)`. And we should probably rename it to `UseInnerClient` or back to `Build` otherwise it gets really hard to differentiate.
    * Alternatively we remove that method entirely and change the constructor to have two overloads, one accepting `IChatClient` and the other accepting `Func<IServiceProvider, IChatClient>`. Then the usage pattern is that you pass the inner client (or factory for one) to the constructor, and any `Use` calls are assumed to come before that in the pipeline.
* It becomes harder for us to detect whether you've configured a pipeline or not. Today we can check if the `pipeline` func is null, and if so, have a default pipeline.
  * We could do something slightly trickier where `ChatClientBuilder` keeps track of whether any `Use` calls have been made, and then at service resolution time, we apply a default pipeline if not
  * Or we move away from the concept of a default pipeline, which is dubious anyway because it's weird that *adding* a pipeline phase erases the default pipeline you had before

Altogether I think the improved aesthetics of doing this without `pipeline => pipeline` make it worthwhile to pursue.

SteveSandersonMS:
Fixes #5572 as well as changing the default DI lifetime from scoped to singleton.

This is also needed for the API design we want in https://github.com/dotnet/aspire/pull/6225

### Usage with DI (common)

Before:

```cs
serviceCollection.AddChatClient(pipeline => pipeline
    .UseDistributedCache()
    .UseFunctionCalling()
    .Use(innerClientOrFactory));
```

After:

```cs
serviceCollection.AddChatClient(innerClientOrFactory)
    .UseDistributedCache()
    .UseFunctionCalling();
```

### Usage without DI (uncommon)

Before:

```cs
var client = new ChatClientBuilder(serviceProvider)
    .UseDistributedCache()
    .UseFunctionCalling()
    .Use(innerClientOrFactory);
```

After:

```cs
var client = new ChatClientBuilder(innerClientOrFactory)
    .UseDistributedCache()
    .UseFunctionCalling()
    .Build(serviceProvider);    // Or just .Build() to use an empty service provider for scenarios where you're really not using DI at all
```

### Summary

In the non-DI case it doesn't make much difference. It's a small improvement though, since it avoids the `Use` method having totally different semantics based on which oveload is called.

But in the DI case which is more common, it's simpler and more approachable as there's no need for the lambda.

It also has value for the Aspire helpers as it make it possible to do [this API](https://github.com/dotnet/aspire/pull/6225#issuecomment-2470704222), which we can't otherwise achieve since you can't specify the inner client up front and can only do so once the rest of the pipeline is built.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5642)


SteveSandersonMS:
Note that I'll do the equivalent thing for `EmbeddingGeneratorBuilder` before this is merged.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilderServiceCollectionExtensions.cs
```diff
@@ -11,37 +11,71 @@ namespace Microsoft.Extensions.DependencyInjection;
...
+    /// <param name="serviceCollection">The <see cref="IServiceCollection"/> to which the client should be added.</param>
     /// <param name="serviceKey">The key with which to associate the client.</param>
-    /// <param name="clientFactory">The factory to use to construct the <see cref="IChatClient"/> instance.</param>
-    /// <returns>The <paramref name="services"/> collection.</returns>
+    /// <param name="innerClientFactory">A callback that produces the inner <see cref="IChatClient"/> that represents the underlying backend.</param>
+    /// <returns>A <see cref="ChatClientBuilder"/> that can be used to build a pipeline around the inner client.</returns>
     /// <remarks>The client is registered as a scoped service.</remarks>
-    public static IServiceCollection AddKeyedChatClient(
-        this IServiceCollection services,
+    public static ChatClientBuilder AddKeyedChatClient(
```
stephentoub, line 68: Did you explicitly choose not to have the generic version as well that you added for `AddChatClient<T>`?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
```diff
@@ -10,32 +10,43 @@ namespace Microsoft.Extensions.AI;
...
+    private Func<IServiceProvider, IChatClient> _innerClientFactory;
+
     /// <summary>The registered client factory instances.</summary>
     private List<Func<IServiceProvider, IChatClient, IChatClient>>? _clientFactories;
 
     /// <summary>Initializes a new instance of the <see cref="ChatClientBuilder"/> class.</summary>
-    /// <param name="services">The service provider to use for dependency injection.</param>
-    public ChatClientBuilder(IServiceProvider? services = null)
+    /// <param name="innerClient">The inner <see cref="IChatClient"/> that represents the underlying backend.</param>
+    public ChatClientBuilder(IChatClient innerClient)
```
stephentoub, line 20: Can you update the README.md's as well?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilderServiceCollectionExtensions.cs
```diff
@@ -11,37 +11,71 @@ namespace Microsoft.Extensions.DependencyInjection;
...
+
+    /// <summary>Adds a chat client to the <see cref="IServiceCollection"/>.</summary>
+    /// <typeparam name="T">The type of the inner <see cref="IChatClient"/> that represents the underlying backend. This will be resolved from the service provider.</typeparam>
+    /// <param name="serviceCollection">The <see cref="IServiceCollection"/> to which the client should be added.</param>
+    /// <returns>A <see cref="ChatClientBuilder"/> that can be used to build a pipeline around the inner client.</returns>
+    /// <remarks>The client is registered as a singleton service.</remarks>
+    public static ChatClientBuilder AddChatClient<T>(
+        this IServiceCollection serviceCollection)
+        where T : IChatClient
+        => AddChatClient(serviceCollection, services => services.GetRequiredService<T>());
```
stephentoub, line 31: What would happen if someone literally called `AddChatClient<IChatClient>`?

SteveSandersonMS: That's equivalent to doing something like:

```cs
serviceCollection.AddSingleton(services => services.GetRequiredService<IMyInterface>());
```

i.e., registering some `T` service as a factory that itself tries to resolve a `T`. In practice this behaves as an infinite loop, both in the `AddChatClient<IChatClient>` case and in the line of code I posted in this comment. That's not great but it's the same as the more general behavior in the case of recursive registrations.

There is another design aspect that's a bit ambiguous here which is that if you do:

```cs
serviceCollection.AddChatClient<SomeChatClient>();
```

... or similar, then this does *not* register `SomeChatClient` in DI, but rather only registers `IChatClient`. That makes sense in the context of this scenario alone, but feels inconsistent with other DI `Add...<T>` methods such as `AddSingleton<T>`, since they actually do register the `T` service in DI.

We could fix this by calling `TryAddSingleton<T>()` inside the `AddChatClient<T>` method. However it's not clear on how often that would be useful in practice since in most cases you may need to call some other method to register the concrete `T` service anyway (e.g., because it needs some other params).

Given these two pitfalls I'm not 100% sold on adding the generic-typed overloads here. Do you have an opinion about it? I'm probably leaning towards not having the generic-typed overloads, since then we're free to add them in the future if we get enough evidence of the need.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilderServiceCollectionExtensions.cs
SteveSandersonMS: At the time yes it was intentional, but I'm glad you challenged this because my reasoning didn't actually hold up on closer inspection. I've now added this, but am [leaning towards removing them both](https://github.com/dotnet/extensions/pull/5642#pullrequestreview-2434998332).

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
SteveSandersonMS: Done

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilderServiceCollectionExtensions.cs
SteveSandersonMS: Update: to keep this moving forwards, I've removed them.

stephentoub: Sounds good

eiriktsarpalis (approved):
Thanks, the new approach looks much more intuitive.

PR #5647, opened 2024-11-14
EmbeddingGeneratorBuilder API updates

SteveSandersonMS:
Brings it into line with what we just did in #5642 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5647)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
@@ -432,10 +432,11 @@ var tracerProvider = OpenTelemetry.Sdk.CreateTracerProviderBuilder()
 
 // Explore changing the order of the intermediate "Use" calls to see that impact
 // that has on what gets cached, traced, etc.
-IEmbeddingGenerator<string, Embedding<float>> generator = new EmbeddingGeneratorBuilder<string, Embedding<float>>()
+var generator = new EmbeddingGeneratorBuilder<string, Embedding<float>>(
+        new SampleEmbeddingGenerator(new Uri("http://coolsite.ai"), "my-custom-model"))
     .UseDistributedCache(new MemoryDistributedCache(Options.Create(new MemoryDistributedCacheOptions())))
     .UseOpenTelemetry(sourceName)
-    .Use(new SampleEmbeddingGenerator(new Uri("http://coolsite.ai"), "my-custom-model"));
+    .Build();
 
 var embeddings = await generator.GenerateAsync(
 [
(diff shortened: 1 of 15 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
```diff
@@ -432,10 +432,11 @@ var tracerProvider = OpenTelemetry.Sdk.CreateTracerProviderBuilder()
 
 // Explore changing the order of the intermediate "Use" calls to see that impact
 // that has on what gets cached, traced, etc.
-IEmbeddingGenerator<string, Embedding<float>> generator = new EmbeddingGeneratorBuilder<string, Embedding<float>>()
+var generator = new EmbeddingGeneratorBuilder<string, Embedding<float>>(
+        new SampleEmbeddingGenerator(new Uri("http://coolsite.ai"), "my-custom-model"))
```
stephentoub, line 436: To consider separately, I wonder if we should have an extension on `IEmbeddingGenerator<...>` that creates a corresponding builder. Then in this situation it could infer the generic arguments, e.g.
```C#
var generator = new SampleEmbeddingGenerator(new Uri("http://coolsite.ai"), "my-custom-model"))
    .WithBuilder()
    .UseDistributedCache(...)
    .UseOpenTelemetry(sourceName)
    .Build();
```

Alternatively, maybe for these non-DI cases we don't even need a builder. We could expose "With" variants of the "Use" methods, just simplifying the act of wrapping one of the middleware clients around one of the leaf clients, e.g.
```C#
var generator = new SampleEmbeddingGenerator(new Uri("http://coolsite.ai"), "my-custom-model"))
    .WithDistributedCache(...)
    .WithOpenTelemetry(...);
```
which would be equivalent to:
```C#
var generator =
    new OpenTelemetryEmbeddingGenerator<..>(
        new DistributedCachingEmbeddingGenerator<...>(
            new SampleEmbeddingGenerator(...),
            new MemoryDistributedCache(...),
        sourceName));
```

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
SteveSandersonMS: I like the idea of doing something like this.

The trickiest bit to me seems that the `.WithSomething` approach would register things in the opposite order than the normal pipeline builder.

The `WithBuilder` approach looks a more solid option. It's more or less equivalent to the lambda-based option we previously speculated about doing, e.g.:

```cs
var generator = new SampleEmbeddingGenerator(new Uri("http://coolsite.ai"), "my-custom-model")).Wrap(builder => builder
    .UseDistributedCache(...)
    .UseOpenTelemetry(...));
```

Personally I'm equally happy with your `.WithBuilder()....Build()` chaining style as with the lambda.

SteveSandersonMS: I have to mark this conversation as resolved in order to merge. Hope that's OK!

PR #5650, opened 2024-11-16
Add anonymous chat clients / embedding generators

stephentoub:
Provides an `AnonymousChatClient`/`EmbeddingGenerator` and corresponding `UseAnonymous` methods, that enable adding simple middleware functionality without needing to derive a custom type from `DelegatingChatHandler`. For example, a client that wants to add a message before delegating and remove it after can just do:
```C#
.UseAnonymous(async (args, next) =>
{
    ChatMessage m = new(ChatRole.User, "Always respond in Italian");
    args.ChatMessages.Add(m);
    try { await next(args.ChatMessages, args.Options, args.CancellationToken); }
    finally { args.ChatMessages.Remove(m); }
})
```
With that overload, the same delegate is used for both non-streaming and streaming, with the batch or streaming results forwarded to the consumer, such that this delegate doesn't get a chance to mutate or augment the results. If such mutation is desired, another overload exists that allows for the non-streaming and streaming to be provided using their own dedicated delegates. If just one of the delegates is provided, than the other mode is implemented in terms of it.
```C#
.UseAnonymous(async args =>
{
    ChatMessage m = new(ChatRole.User, "Always respond in Italian");
    args.ChatMessages.Add(m);
    try
    {
        ChatCompletion completion = await args.InnerClient.CompleteAsync(args.ChatMessages, args.Options, args.CancellationToken);
        foreach (ChatMessage choice in completion.Choices)
        {
            choice.Text += "Caio!";
        }
        return completion;
    }
    finally { args.ChatMessages.Remove(m); }
})
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5650)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -63,6 +63,7 @@
     <SystemTextEncodingsWebVersion>9.0.0</SystemTextEncodingsWebVersion>
     <SystemNumericsTensorsVersion>9.0.0</SystemNumericsTensorsVersion>
     <SystemTextJsonVersion>9.0.0</SystemTextJsonVersion>
+    <SystemThreadingChannelsVersion>9.0.0</SystemThreadingChannelsVersion>
     <!-- Dependencies from https://github.com/aspnet/AspNetCore -->
     <MicrosoftAspNetCoreAppRefVersion>9.0.1</MicrosoftAspNetCoreAppRefVersion>
     <MicrosoftAspNetCoreAppRuntimewinx64Version>9.0.1</MicrosoftAspNetCoreAppRuntimewinx64Version>
@@ -112,6 +113,7 @@
     <SystemTextEncodingsWebLTSVersion>8.0.0</SystemTextEncodingsWebLTSVersion>
     <SystemNumericsTensorsLTSVersion>8.0.0</SystemNumericsTensorsLTSVersion>
     <SystemTextJsonLTSVersion>8.0.5</SystemTextJsonLTSVersion>
+    <SystemThreadingChannelsLTSVersion>8.0.0</SystemThreadingChannelsLTSVersion>
     <!-- Dependencies from https://github.com/aspnet/AspNetCore -->
     <MicrosoftAspNetCoreAppRefLTSVersion>8.0.11</MicrosoftAspNetCoreAppRefLTSVersion>
     <MicrosoftAspNetCoreAppRuntimewinx64LTSVersion>8.0.11</MicrosoftAspNetCoreAppRuntimewinx64LTSVersion>
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousDelegatingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousDelegatingChatClient.cs
@@ -1,0 +1,213 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+#pragma warning disable VSTHRD003 // Avoid awaiting foreign Tasks
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>A delegating chat client that wraps an inner client with implementations provided by delegates.</summary>
+public sealed class AnonymousDelegatingChatClient : DelegatingChatClient
+{
+    /// <summary>The delegate to use as the implementation of <see cref="CompleteAsync"/>.</summary>
+    private readonly Func<IList<ChatMessage>, ChatOptions?, IChatClient, CancellationToken, Task<ChatCompletion>>? _completeFunc;
+
+    /// <summary>The delegate to use as the implementation of <see cref="CompleteStreamingAsync"/>.</summary>
+    /// <remarks>
+    /// When non-<see langword="null"/>, this delegate is used as the implementation of <see cref="CompleteStreamingAsync"/> and
+    /// will be invoked with the same arguments as the method itself, along with a reference to the inner client.
+    /// When <see langword="null"/>, <see cref="CompleteStreamingAsync"/> will delegate directly to the inner client.
+    /// </remarks>
+    private readonly Func<IList<ChatMessage>, ChatOptions?, IChatClient, CancellationToken, IAsyncEnumerable<StreamingChatCompletionUpdate>>? _completeStreamingFunc;
+
+    /// <summary>The delegate to use as the implementation of both <see cref="CompleteAsync"/> and <see cref="CompleteStreamingAsync"/>.</summary>
+    private readonly CompleteSharedFunc? _sharedFunc;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AnonymousDelegatingChatClient"/> class.
+    /// </summary>
+    /// <param name="innerClient">The inner client.</param>
+    /// <param name="sharedFunc">
+    /// A delegate that provides the implementation for both <see cref="CompleteAsync"/> and <see cref="CompleteStreamingAsync"/>.
+    /// In addition to the arguments for the operation, it's provided with a delegate to the inner client that should be
+    /// used to perform the operation on the inner client. It will handle both the non-streaming and streaming cases.
+    /// </param>
+    /// <remarks>
+    /// This overload may be used when the anonymous implementation needs to provide pre- and/or post-processing, but doesn't
+    /// need to interact with the results of the operation, which will come from the inner client.
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="innerClient"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="sharedFunc"/> is <see langword="null"/>.</exception>
+    public AnonymousDelegatingChatClient(IChatClient innerClient, CompleteSharedFunc sharedFunc)
+        : base(innerClient)
+    {
+        _ = Throw.IfNull(sharedFunc);
+
+        _sharedFunc = sharedFunc;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AnonymousDelegatingChatClient"/> class.
+    /// </summary>
+    /// <param name="innerClient">The inner client.</param>
+    /// <param name="completeFunc">
+    /// A delegate that provides the implementation for <see cref="CompleteAsync"/>. When <see langword="null"/>,
+    /// <paramref name="completeStreamingFunc"/> must be non-null, and the implementation of <see cref="CompleteAsync"/>
+    /// will use <paramref name="completeStreamingFunc"/> for the implementation.
+    /// </param>
+    /// <param name="completeStreamingFunc">
+    /// A delegate that provides the implementation for <see cref="CompleteStreamingAsync"/>. When <see langword="null"/>,
+    /// <paramref name="completeFunc"/> must be non-null, and the implementation of <see cref="CompleteStreamingAsync"/>
+    /// will use <paramref name="completeFunc"/> for the implementation.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="innerClient"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException">Both <paramref name="completeFunc"/> and <paramref name="completeStreamingFunc"/> are <see langword="null"/>.</exception>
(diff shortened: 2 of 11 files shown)


SteveSandersonMS:
Thanks for moving this forwards! I'll review the code on Monday but in terms of naming, how would you feel about something like `UsePipelineStep` or `UseStep`?

I wasn't able to guess what "anonymous chat clients" meant when I first just saw the PR title alone.

stephentoub:
> how would you feel about something like UsePipelineStep or UseStep?

Those aren't more clear to me, as there's nothing in the names of the other UseXx methods that says they're "steps".

How about just Use?

stephentoub:
Or UseDelegate or UseCustom?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// </remarks>
+    private readonly Func<Arguments, IAsyncEnumerable<StreamingChatCompletionUpdate>>? _completeStreamingFunc;
+
+    /// <summary>The delegate to use as the implementation of both <see cref="CompleteAsync"/> and <see cref="CompleteStreamingAsync"/>.</summary>
+    private readonly Func<Arguments, Func<IList<ChatMessage>, ChatOptions?, CancellationToken, Task>, Task>? _sharedFunc;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AnonymousChatClient"/> class.
+    /// </summary>
+    /// <param name="innerClient">The inner generator.</param>
```
eiriktsarpalis, line 41: Presumably this was copied over by the equivalent embedding generator?
```suggestion
    /// <param name="innerClient">The inner client.</param>
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
```diff
@@ -0,0 +1,242 @@
...
+        Func<Arguments, IAsyncEnumerable<StreamingChatCompletionUpdate>>? completeStreamingFunc)
+    {
+        if (completeFunc is null && completeStreamingFunc is null)
+        {
+            Throw.ArgumentNullException(nameof(completeFunc), $"At least one of the {nameof(completeFunc)} or {nameof(completeStreamingFunc)} delegates must be non-null.");
+        }
+    }
+
+    /// <summary>Represents the arguments passed to the delegates used by <see cref="AnonymousChatClient"/>.</summary>
+    public sealed class Arguments
```
eiriktsarpalis, line 210: I'm guessing you chose this approach over a custom delegate to allow future extensibility/minimize lambda parameters?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
```diff
@@ -0,0 +1,242 @@
...
+        public ChatOptions? Options { get; }
+
+        /// <summary>Gets the <see cref="CancellationToken"/> to monitor for cancellation requests.</summary>
+        public CancellationToken CancellationToken { get; }
+
+        /// <summary>Gets the inner client to which to delegate.</summary>
+        public IChatClient InnerClient { get; }
+
+        /// <summary>Gets any services associated with the chat client.</summary>
+        public IServiceProvider? Services { get; }
```
eiriktsarpalis, line 240: I'm guessing by the presence of this parameter that the inner service provider can differ from the outer one?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// </param>
+    /// <param name="completeStreamingFunc">
+    /// A delegate that provides the implementation for <see cref="CompleteStreamingAsync"/>. When <see langword="null"/>,
+    /// <paramref name="completeFunc"/> must be non-null, and the implementation of <see cref="CompleteStreamingAsync"/>
+    /// will use <paramref name="completeFunc"/> for the implementation.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="innerClient"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException">Both <paramref name="completeFunc"/> and <paramref name="completeStreamingFunc"/> are <see langword="null"/>.</exception>
+    public AnonymousChatClient(
+        IChatClient innerClient,
```
eiriktsarpalis, line 85: Unless I'm misunderstanding something, the only reason we require an inner client is to drive the shared func, which isn't a mode being exercised at all by the particular constructor.

In general, I think requiring a inner client negates the purpose of exposing an anonymous interface implementation.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
stephentoub: That's part of it. But also because otherwise there's a large number of parameters that must be specified in the lambda, and this consolidates them all to one. We could re-expand it if that's desirable.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
stephentoub: I'm not sure what you mean. The services here are passed in from the builder.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
```diff
@@ -0,0 +1,242 @@
...
+    /// <remarks>
+    /// This overload may be used when the anonymous implementation needs to provide pre- and/or post-processing, but doesn't
+    /// need to interact with the results of the operation, which will come from the inner client.
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="innerClient"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="sharedFunc"/> is <see langword="null"/>.</exception>
+    public AnonymousChatClient(
+        IChatClient innerClient,
+        IServiceProvider? services,
+        Func<Arguments, Func<IList<ChatMessage>, ChatOptions?, CancellationToken, Task>, Task> sharedFunc)
```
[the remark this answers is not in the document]
stephentoub: Because then you need to provide the same implementation twice, and deal with not being able to supply an async iterator as a lambda.

SteveSandersonMS:
> How about just Use?

Use sounds great.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
eiriktsarpalis: Thanks, your example makes it clearer to me what this type represents. Effectively it's modelling an aspnetcore-style middleware? I interpreted the term "anonymous client" as a helper that just that creates thin anonymous objects using delegates.

> No, it's necessary to be passed to the base because it's used by the other delegating members, i.e.. Metadata, GetService, and Dispose.

I can see how it accelerates the definition of those ancillary methods, but not needing to specify an underlying client feels like goodness too. Some of these ancillary members could be specified using additional parameters.

My overall feeling is that the two constructors are different enough that they should be factored into separate types with different names.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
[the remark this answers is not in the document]
eiriktsarpalis: Could this be converter to a bespoke delegate type? Explicitly labelling the nested func with `next` would help a lot in conveying what it does.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
```diff
@@ -76,4 +78,62 @@ public ChatClientBuilder Use(Func<IChatClient, IServiceProvider, IChatClient> cl
...
+    /// provided with a delegate to the inner client that should be used to perform the operation on the inner client.
+    /// It will handle both the non-streaming and streaming cases.
+    /// </param>
+    /// <returns>The updated <see cref="ChatClientBuilder"/> instance.</returns>
+    /// <remarks>
+    /// This overload may be used when the anonymous implementation needs to provide pre- and/or post-processing, but doesn't
+    /// need to interact with the results of the operation, which will come from the inner client.
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="sharedFunc"/> is <see langword="null"/>.</exception>
+    public ChatClientBuilder Use(AnonymousDelegatingChatClient.CompleteSharedFunc sharedFunc)
```
SteveSandersonMS, line 98: Sidenote: while this will be fine to use when copy-pasting an example, it will be challenging to use just through intellisense alone. AFAIK VS doesn't provide any prompt to indicate what set of args a lambda should take in order to conform to a delegate definition. Obviously there's nothing you can do about that here and it's fine for this PR.

But I wonder if we should consider asking for a C#/Intellisense improvement whereby, when the cursor is in a place where you're being asked for a delegate, it should include some sort of "generate lambda" option in the list of available completions which would inject code like `(T0 arg0, T1 arg1, T2 arg2) => { }`.

I remember in the past having this issue with Blazor's grid data provider callback. The delegate type is just too complicated to remember so I can never work out how to actually write the code for a grid data provider and have to go looking for an example to copy-paste.

SteveSandersonMS (approved):
The "shared" delegate type is definitely an interesting approach. If the audience gets on well with it, it's a very nice way to simplify things. We'll have to see how well this lands in practice or whether folks don't get why that version can't see mutate the results.

In any case this was clearly a tricky thing to design well so it's great to see it done.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
eiriktsarpalis: You can also access the services on the `InnerClient` property.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousDelegatingChatClient.cs
```diff
@@ -0,0 +1,213 @@
...
+#pragma warning restore CA2016
+            {
+                Exception? error = null;
+                try
+                {
+                    await _sharedFunc(chatMessages, options, async (chatMessages, options, cancellationToken) =>
+                    {
+                        await foreach (var update in InnerClient.CompleteStreamingAsync(chatMessages, options, cancellationToken).ConfigureAwait(false))
+                        {
+                            await updates.Writer.WriteAsync(update, cancellationToken).ConfigureAwait(false);
```
eiriktsarpalis, line 143: That's pretty clever. Do you think this unifying pattern can be extended beyond delegating wrappers? IOW could the behaviour of chat client be fully defined in terms of a variation of this higher-order function, and we expose `CompleteAsync` and `CompleteAsync` as extension methods?

I would guess not since the nested callback inherently relies on an inner client with discrete streaming and non-streaming functionalities.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousChatClient.cs
stephentoub: Are you referring to GetService? That's not the same. It does not provide access to IServiceProvider.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousDelegatingChatClient.cs
stephentoub: Yeah, I don't see a good way to do that. It is possible to define non streaming in terms of streaming, though, albeit with some minimal loss of fidelity.

eng/Versions.props
```diff
@@ -63,6 +63,7 @@
     <SystemTextEncodingsWebVersion>9.0.0</SystemTextEncodingsWebVersion>
     <SystemNumericsTensorsVersion>9.0.0</SystemNumericsTensorsVersion>
     <SystemTextJsonVersion>9.0.0</SystemTextJsonVersion>
+    <SystemThreadingChannelsVersion>9.0.0</SystemThreadingChannelsVersion>
```
RussKie, line 66: @stephentoub, I just stumbled on this change. This won't work as you expect. For these packages to be updated via the automated dependency flow there must be an associated entry in Version.Details.xml.
The maestro bot is querying Version.Details.xml to determine what dependencies must be updated.

PR #5651, opened 2024-11-16
Remove duplicate GetCacheKey methods

stephentoub:
Consolidate to only the `ReadOnlySpan<object>`-based method.

@shyamnamboodiripad, does this address your concerns (https://github.com/dotnet/extensions/pull/5641#discussion_r1844931054)?
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5651)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
@@ -92,7 +92,10 @@ public class DistributedCachingChatClient : CachingChatClient
         await _storage.SetAsync(key, newJson, cancellationToken).ConfigureAwait(false);
     }
 
-    /// <inheritdoc />
+    /// <summary>Computes a cache key for the specified values.</summary>
+    /// <param name="values">The values to inform the key.</param>
+    /// <returns>The computed key.</returns>
+    /// <remarks>The <paramref name="values"/> are serialized to JSON using <see cref="JsonSerializerOptions"/> in order to compute the key.</remarks>
     protected override string GetCacheKey(params ReadOnlySpan<object?> values)
     {
         _jsonSerializerOptions.MakeReadOnly();
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
@@ -20,12 +20,6 @@ namespace Microsoft.Extensions.AI;
 /// </remarks>
 public class DistributedCachingChatClient : CachingChatClient
 {
-    /// <summary>A boxed <see langword="true"/> value.</summary>
-    private static readonly object _boxedTrue = true;
-
-    /// <summary>A boxed <see langword="false"/> value.</summary>
-    private static readonly object _boxedFalse = false;
-
     /// <summary>The <see cref="IDistributedCache"/> instance that will be used as the backing store for the cache.</summary>
     private readonly IDistributedCache _storage;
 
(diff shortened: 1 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -99,14 +93,7 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
...
-
-    /// <summary>Gets a cache key based on the supplied values.</summary>
-    /// <param name="values">The values to inform the key.</param>
-    /// <returns>The computed key.</returns>
-    /// <remarks>This provides the default implementation for <see cref="GetCacheKey(bool, IList{ChatMessage}, ChatOptions?)"/>.</remarks>
-    protected string GetCacheKey(ReadOnlySpan<object?> values)
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values)
     {
         _jsonSerializerOptions.MakeReadOnly();
         return CachingHelpers.GetCacheKey(values, _jsonSerializerOptions);
```
shyamnamboodiripad, line 99: Should `CachingHelpers.GetCacheKey` be made public so that the 'default' hash algorithm is always available for any derived class that does not want to use an overridden implementation present in its immediate base class?

stephentoub: Example?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -99,14 +93,7 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
...
     /// <inheritdoc />
-    protected override string GetCacheKey(bool streaming, IList<ChatMessage> chatMessages, ChatOptions? options) =>
-        GetCacheKey([streaming ? _boxedTrue : _boxedFalse, chatMessages, options]);
-
-    /// <summary>Gets a cache key based on the supplied values.</summary>
-    /// <param name="values">The values to inform the key.</param>
-    /// <returns>The computed key.</returns>
-    /// <remarks>This provides the default implementation for <see cref="GetCacheKey(bool, IList{ChatMessage}, ChatOptions?)"/>.</remarks>
-    protected string GetCacheKey(ReadOnlySpan<object?> values)
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values)
```
shyamnamboodiripad, line 96: The default implementation in `DistributedCachingChatClient` relies on the supplied objects being json serializable. Should we clarify this via comments somehow?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
shyamnamboodiripad: Consider the following case.

`BaseCachingChatClient` inherits `DistributedChatClinet` and overrides `GetCacheKey` to use a completely custom hashing algorithm.

`DerivedCachingChatClient` inherits `BaseCachingChatClient`, but it does not like the cache keys it gets back from the custom hashing algorithm implemented in `BaseCachingChatClient`, and wants to use the 'default' hashing algorithm used in `DistributedCachingChatClient` instead (or any other off the shelf hashing algorithm for that matter so long as it does not need to implement, test and secure its own new hashing algorithm from scratch)...

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
stephentoub: C inherits B inherits A. If C doesn't use B's override, it'll be missing out on any additional state B felt it necessary to include in the hash key. That's a bug, since now C might have two identical keys that B considers to be distinct so as to not share a cache entry.

PR #5654, opened 2024-11-16
Tweak CachingHelpers.GetCacheKey to clean up better on failure

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5654)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/CachingHelpers.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/CachingHelpers.cs
@@ -36,19 +36,41 @@ internal static class CachingHelpers
         // invalidating any existing cache entries that may exist in whatever IDistributedCache was in use.
 
 #if NET
-        IncrementalHashStream? stream = IncrementalHashStream.ThreadStaticInstance ?? new();
-        IncrementalHashStream.ThreadStaticInstance = null;
-
-        foreach (object? value in values)
+        IncrementalHashStream? stream = IncrementalHashStream.ThreadStaticInstance;
+        if (stream is not null)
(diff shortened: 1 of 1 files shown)


SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI/CachingHelpers.cs
```diff
@@ -36,19 +36,41 @@ public static string GetCacheKey(ReadOnlySpan<object?> values, JsonSerializerOpt
...
-        IncrementalHashStream? stream = IncrementalHashStream.ThreadStaticInstance ?? new();
-        IncrementalHashStream.ThreadStaticInstance = null;
+        IncrementalHashStream? stream = IncrementalHashStream.ThreadStaticInstance;
+        if (stream is not null)
+        {
+            // We need to ensure that the value in ThreadStaticInstance is always ready to use.
+            // If we start using an instance, write to it, and then fail, we will have left it
+            // in an inconsistent state. So, when renting it, we null it out, and we only put
+            // it back upon successful completion after resetting it.
+            IncrementalHashStream.ThreadStaticInstance = null;
```
eiriktsarpalis, line 46: Why not just dispose it in a try/finally? Is it expected that this could be called recursively?

stephentoub: It shouldn't be disposed if it's going to be cached. So it's disposed in a catch rather than a finally.

PR #5666, opened 2024-11-18
Add OpenAIRealtimeExtensions with ToConversationFunctionTool

SteveSandersonMS:
Realised this will involve upgrading the `OpenAI` library dependency from 2.0.0 to 2.1.0-beta.2.

I think the main things affected will be the `Aspire.OpenAI` and `Aspire.Azure.AI.OpenAI` packages, since they will then get a dependency on OpenAI 2.1.0-beta.2. Most likely that won't be a problem since they aren't going to be shipping with stable version numbers anyway (since they depend on M.E.AI) but @eerhardt / @sebastienros if you anticipate this being an issue, please let us know!

### Testing

It's difficult to test this but I've done what I think is reasonable at present.

 * For unit testing, we're constrainted by the fact that some of the realtime API types can't be instantiated or subclassed. So we can't actually create (for example) `ConversationItemStreamingFinishedUpdate` to use in a test.
 * For integration testing, we're constrained by rate limits. It's only allowed to create one conversation session every 6 seconds. So for now there's only one integration test, and it combines a whole bunch of different behaviors to show the end-to-end is working.

If we go further with realtime API support in the future we'll need to find ways to test more, but hopefully this is sufficient coverage for now.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5666)


src/Libraries/Microsoft.Extensions.AI.OpenAI/JsonContext.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Source-generated JSON type information.</summary>
+[JsonSourceGenerationOptions(JsonSerializerDefaults.Web,
+    UseStringEnumConverter = true,
+    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+    WriteIndented = true)]
+[JsonSerializable(typeof(OpenAIChatClient.OpenAIChatToolJson))]
+[JsonSerializable(typeof(OpenAIRealtimeExtensions.ConversationFunctionToolParametersSchema))]
+internal sealed partial class OpenAIJsonContext : JsonSerializerContext;
```
stephentoub, line 16: Nit: rename the file to match?

eng/packages/General.props
```diff
@@ -11,7 +11,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisVersion)" />
     <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="OpenAI" Version="2.0.0" />
+    <PackageVersion Include="OpenAI" Version="2.1.0-beta.2" />
```
eerhardt, line 14: Do we have an idea on when `2.1.0` will ship stable? Do we anticipate it will be before MEAI ships stable?

SteveSandersonMS: I have no inside information about when OpenAI client 2.1.0 ships, but would guess it would be a *long* time before MEAI becomes stable.

In the very unlikely event that this is not the case, we'd be forced to move the realtime parts out into a separate preview package in order for MEAI to ship as stable. But I really doubt that will happen.

src/Libraries/Microsoft.Extensions.AI.OpenAI/JsonContext.cs
SteveSandersonMS: Done. PR is ready to go now.

eng/packages/General.props
eerhardt: Sounds good to me. Thanks for the info.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <returns>A <see cref="ConversationFunctionTool"/> that can be used with <see cref="RealtimeConversationClient"/>.</returns>
+    public static ConversationFunctionTool ToConversationFunctionTool(this AIFunction aiFunction)
+    {
+        _ = Throw.IfNull(aiFunction);
+
+        var parametersSchema = new ConversationFunctionToolParametersSchema
+        {
+            Type = "object",
+            Properties = aiFunction.Metadata.Parameters
+                .Where(p => p.Schema is JsonElement)
```
eiriktsarpalis, line 34: I don't think we want to filter out parameters that don't have a schema since that would result in an incorrect function arity being reported. I would recommend defaulting to the `true` or `{}` schema instead, similar to how the main leaf client is doing it.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// it can be used with <see cref="RealtimeConversationClient"/>.
+    /// </summary>
+    /// <returns>A <see cref="ConversationFunctionTool"/> that can be used with <see cref="RealtimeConversationClient"/>.</returns>
+    public static ConversationFunctionTool ToConversationFunctionTool(this AIFunction aiFunction)
+    {
+        _ = Throw.IfNull(aiFunction);
+
+        var parametersSchema = new ConversationFunctionToolParametersSchema
+        {
+            Type = "object",
```
eiriktsarpalis, line 32: Doesn't OpenAI require the type "function" for its top-level schema?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <param name="update">The <see cref="ConversationUpdate"/> being processed.</param>
+    /// <param name="tools">The available tools.</param>
+    /// <param name="detailedErrors">An optional flag specifying whether to disclose detailed exception information to the model. The default value is <see langword="false"/>.</param>
+    /// <param name="jsonSerializerOptions">An optional <see cref="JsonSerializerOptions"/> that controls JSON handling.</param>
+    /// <param name="cancellationToken">An optional <see cref="CancellationToken"/>.</param>
+    /// <returns>A <see cref="Task"/> that represents the completion of processing, including invoking any asynchronous tools.</returns>
+    public static async Task HandleToolCallsAsync(
+        this RealtimeConversationSession session,
+        ConversationUpdate update,
+        IReadOnlyList<AIFunction> tools,
```
eiriktsarpalis, line 69: Should we perhaps wrap the function list in a dedicated option type, in case more configuration is going to be necessary in the future?

SteveSandersonMS: It's an interesting idea, but I'm trying to minimize the invention of new APIs/abstractions here since we don't yet know if/when we'd really create abstractions around realtime, and if we did, they would end up in a different package entirely.

We're definitely not trying to create some final APIs for realtime here - anything we deliver at this point is going to have breaking changes later when OpenAI's realtime APIs have further breaking changes. So I'd rather keep it simple and minimal and be ready to adapt as needed in the future.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
SteveSandersonMS: This isn't the top-level schema; it's the list of parameters. It's equivalent to this sample code: https://github.com/Azure-Samples/aoai-realtime-audio-sdk/blob/main/dotnet/samples/console-from-file/Program.cs#L230-L249

So, `object` is apparently the right value here and does work.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
SteveSandersonMS: Good point. Done.

stephentoub (approved):

PR #5670, opened 2024-11-19
Removes Experimental attribute from ResilienceHandler class

### LINKED ISSUE #5669: [Http.Resilience] Make experimental class ResilienceHandler stable
`ResilienceHandler` is currently experimental. The API looks solid and can be made stable. We need to remove `Experimental` attribute.

Originally the issue was discussed there https://github.com/dotnet/extensions/issues/4893#issuecomment-2482058978.

iliar-turdushev:
The PR removes `Experimental` attribute from ResilienceHandler class.

Fixes #5669
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5670)


scripts/MakeApiBaselines.ps1
```diff
@@ -16,7 +16,7 @@ Write-Output "Installing required toolset"
 InitializeDotNetCli -install $true | Out-Null
 
 $Project = $PSScriptRoot + "/../eng/Tools/ApiChief/ApiChief.csproj"
-$Command = $PSScriptRoot + "/../artifacts/bin/ApiChief/Debug/net8.0/ApiChief.dll"
+$Command = $PSScriptRoot + "/../artifacts/bin/ApiChief/Debug/net9.0/ApiChief.dll"
```
iliar-turdushev, line 19: We now build the `ApiChief` tool for `net9.0` that is why I changed this.

amadeuszl (approved):

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Name": "Microsoft.Extensions.Http.Resilience, Version=8.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
+  "Name": "Microsoft.Extensions.Http.Resilience, Version=9.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
```
iliar-turdushev, line 2: The API json file hasn't been updated since dotnet/extensions 8.0.0 therefore the changes to this file also include APIs that were changed/updated earlier.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
```diff
@@ -25,6 +25,12 @@
     <MinMutationScore>100</MinMutationScore>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Disable "Published symbols cannot be deleted to maintain compatibility" warning
+         because we have different APIs for different TFMs. -->
+    <NoWarn Condition="'$(TargetFramework)' == 'net462'">$(NoWarn);LA0006</NoWarn>
+  </PropertyGroup>
```
iliar-turdushev, line 32: We have to suppress `LA0006 - Published symbols cannot be deleted to maintain compatibility` warning. The reason is the following. Together with making the `ResilienceHandler` class stable, we also made stable protected members `SendAsync` and `Send`. These members are inherited from `DelegatingHandler` and `ResilienceHandler` overrides them. The `Send` method is not available on `net462` and this causes the build pipeline to fail with the `LA0006` warning. To make the build pipeline "green" we have to suppress the `LA0006` warning.

RussKie:
Approving on the assumption the @dotnet/dotnet-extensions-resilience team has done the due diligence for the API review.

iliar-turdushev:
Tagging @RussKie @joperezr as infra-area owners. Please, review the PR. Thank you.

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
joperezr: This is suppressing all of the warnings on public symbols though, right? Is there not a way to suppress that single change?

joperezr: Also, I'm not very familiar to which analyzer is throwing those errors, but FWIW we do run package validation when we pack projects, which would catch compatibility issues in packages. In this case, net462 and net6.0+ aren't even supposed to be compatible with each other so perhaps the analyzer tooling being used here is not correct.

joperezr (commented):
Left a couple of minor comments, but looks good otherwise

src/Libraries/Microsoft.Extensions.Http.Resilience/Microsoft.Extensions.Http.Resilience.csproj
iliar-turdushev: @joperezr 

> This is suppressing all of the warnings on public symbols though, right? Is there not a way to suppress that single change?

Sometime ago I tried to suppress the warning in a source file for a particular change, but it didn't work. Here is another fresher PR https://github.com/dotnet/extensions/pull/5690 where I tried it again and it didn't work. Unfortunately, we have to suppress the warning for the whole project.

> Also, I'm not very familiar to which analyzer is throwing those errors, but FWIW we do run package validation when we pack projects, which would catch compatibility issues in packages. In this case, net462 and net6.0+ aren't even supposed to be compatible with each other so perhaps the analyzer tooling being used here is not correct.

Yes, I agree. I don't understand what exactly the analyzer is catching. Adding a non-abstract method that is available in net6+ and is not on net462 is alright in regards of compatibility, i.e. upgrading from net462 to net6+ will not break. Right?

iliar-turdushev: @joperezr I'm going to resolve this discussion and merge the PR. Do you know who we can follow up with on this analyzer warning? We need to figure out whether the warning the analyzer triggers is incorrect.

RussKie: @iliar-turdushev please open a new issue and let's continue the discussion there.

iliar-turdushev: @RussKie Here is the issue https://github.com/dotnet/extensions/issues/5721.

Issue #5668, opened 2024-11-19
FunctionCallingChatClient reports incorrect UsageData


lucasmeijer:
the FunctionInvokingChatClient.CompleteAsync()

has a loop that will call its inner chatclient multiple times to service any FunctionInvocationsRequest.

It forgets to sum up all the UsageData information from those additional requests, and include those in the result,
leading to under-reporting of used tokens.

SideNote: congrats on this library.  It looks super solid, and as I'm contemplating migrating away from my own solution to this
one, so far this arguably minor issue is the only one I've ran into so far.  great job.

stephentoub:
> SideNote: congrats on this library. It looks super solid, and as I'm contemplating migrating away from my own solution to this
one, so far this arguably minor issue is the only one I've ran into so far. great job.

Thanks. Please do :)

> It forgets to sum up all the UsageData information from those additional requests, and include those in the result,
leading to under-reporting of used tokens.

This should only be a problem for non-streaming, right? With streaming, every iteration yields all of the results from the inner client, including the usage data. For non-streaming, the usage data is stored on the ChatCompletion object, and the intermediate instances are discarded, so that usage data is also discarded.

I see two options for handling this:
1. Make streaming consistent with non-streaming, by removing UsageContent from the stream initially, and then at the end of everything producing a single UsageContent that sums everything up. And then for both sum things up to ensure that the final one produced is the total.
2. Make non-streaming consistent with streaming, ensuring that the intermediate chat messages added to history include UsageContent manufactured from the ChatCompletion.

My preference is (2), in part because I don't think we can accurately do (1). We could sum up the counts we know about, but various services output additional count information (e.g. reasoning tokens, cached tokens, etc.), which are stored in AdditionalProperties, and we can't reason about those.

@SteveSandersonMS?

stephentoub:
e.g. https://github.com/dotnet/extensions/compare/main...stephentoub:extensions:fixfuncinvokeresponsecontent

stephentoub:
Actually, this is going to need a bit more thought...

stephentoub:
I opened https://github.com/dotnet/extensions/pull/5676 to address the usage issue called out here. It does so by ensure the usage details end up in a usage content in the history.

I opened https://github.com/dotnet/extensions/issues/5675 to track the other issues I called out, which I'll address separately.

SteveSandersonMS:
[...]

PR #5676, opened 2024-11-20
Ensure non-streaming usage data from function calling is in history

stephentoub:
It's already yielded during streaming, but it's not being surfaced for non-streaming. Do so by manufacturing a new UsageContent for the UsageDetails and adding that to the response message that's added to the history.

Fixes https://github.com/dotnet/extensions/issues/5668
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5676)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
@@ -163,13 +163,25 @@ public abstract class ChatClientIntegrationTests : IDisposable
 
         int secretNumber = 42;
 
-        var response = await chatClient.CompleteAsync("What is the current secret number?", new()
+        List<ChatMessage> messages =
+        [
+            new(ChatRole.User, "What is the current secret number?")
+        ];
+
+        var response = await chatClient.CompleteAsync(messages, new()
         {
             Tools = [AIFunctionFactory.Create(() => secretNumber, "GetSecretNumber")]
         });
 
         Assert.Single(response.Choices);
         Assert.Contains(secretNumber.ToString(), response.Message.Text);
+
+        if (response.Usage is { } finalUsage)
+        {
+            UsageContent? intermediate = messages.SelectMany(m => m.Contents).OfType<UsageContent>().FirstOrDefault();
+            Assert.NotNull(intermediate);
+            Assert.True(finalUsage.TotalTokenCount > intermediate.Details.TotalTokenCount);
+        }
     }
 
     [ConditionalFact]
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -252,6 +252,13 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
                     }
                 }
 
+                // If the original chat completion included usage data,
+                // add that into the message so it's available in the history.
+                if (KeepFunctionCallingMessages && response.Usage is { } usage)
+                {
+                    response.Message.Contents = [.. response.Message.Contents, new UsageContent(usage)];
+                }
+
                 // Add the responses from the function calls into the history.
                 var modeAndMessages = await ProcessFunctionCallsAsync(chatMessages, options, functionCallContents, iteration, cancellationToken).ConfigureAwait(false);
                 if (modeAndMessages.MessagesAdded is not null)


test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
```diff
@@ -163,13 +163,25 @@ public virtual async Task FunctionInvocation_AutomaticallyInvokeFunction_Paramet
...
         });
 
         Assert.Single(response.Choices);
         Assert.Contains(secretNumber.ToString(), response.Message.Text);
+
+        if (response.Usage is { } finalUsage)
+        {
+            UsageContent? intermediate = messages.SelectMany(m => m.Contents).OfType<UsageContent>().FirstOrDefault();
+            Assert.NotNull(intermediate);
+            Assert.True(finalUsage.TotalTokenCount > intermediate.Details.TotalTokenCount);
```
SteveSandersonMS, line 183: I must be missing something - can you clarify why `finalUsage.TotalTokenCount > intermediate.Details.TotalTokenCount`?

I would have thought the final message token count was independent of the token count on preceding messages.

SteveSandersonMS: Oh is the point that `TotalTokenCount` includes input tokens as well as output tokens, and the input for the final message includes the output from preceding messages?

stephentoub: > Oh is the point that `TotalTokenCount` includes input tokens as well as output tokens, and the input for the final message includes the output from preceding messages?

Yup. So the first request sends A tokens and the response has B tokens, then the next request include A and B tokens (since they're in the history) plus an additional C tokens, and has a response with D tokens. As long as C and D aren't 0, the final total should be greater than the intermediate total.

I think... right?

SteveSandersonMS:
While this is a valid solution, I suspect it's not what #5668 had in mind. It sounds like they were not expecting to have to set `KeepFunctionCallingMessages = true` or to have to sum over a set of auto-added messages (somehow working out which ones to sum over), and just expected the response object to contain the sum over usage details across all sub-requests.

I'm fairly sure you've done it this way because there could be usage information that we can't sum automatically, i.e., the `UsageDetails.AdditionalProperties` entries.

Even with this we're still discarding other intermediate state such as the `AdditionalProperties` and `RawRepresentation` on the original `ChatCompletion` objects, but then `FunctionInvokingChatClient` is meant as a convenient but lossy process that primarily just gives you the final post-function-call results and doesn't try to preserve and expose all the intermediate states. People who must have all the intermediate states verbatim would need not to use `FunctionInvokingChatClient`.

Given the lossiness of `FunctionInvokingChatClient` I'd also be open to the alternative of auto-summing known properties on `UsageDetails` and ignoring its `AdditionalProperties`. I know the casualty there would be gpt4-o1 reasoning tokens - at least in the short term, since we could potentially add that as a first-class thing to `UsageDetails` in the long term.

Overall I'm fine with whatever you prefer here.

SteveSandersonMS (approved):

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
SteveSandersonMS: Sounds correct!

stephentoub:
> It sounds like they were not expecting to have to set KeepFunctionCallingMessages = true

That's the default, fwiw

> I'm fairly sure you've done it this way because there could be usage information that we can't sum automatically, i.e., the UsageDetails.AdditionalProperties entries.

Yes

>  I know the casualty there would be gpt4-o1 reasoning tokens

And cached tokens. And predicted tokens. All of those have emerged in just the last few months. Presumably there will be more in the future.

> I'd also be open to the alternative of auto-summing known properties on UsageDetails and ignoring its AdditionalProperties

That makes me nervous. The mitigating factor is at least all the ones we're aware of are included in the totals. But they also all have different pricing structures.

I liked this PR's approach because it logically matches the streaming case, but I understand the hesitancy and the rationale that the resulting ChatCompletion should represent the totality of the operation. I'm just concerned about the summing due to the financial aspect of tokens.

Another approach would be to include all of the additional properties in the final one, and assume that anything that's a property with an int value is summable. Thats not necessarily a valid assumption, but we could do it anyway. Or find another way to include everything, like putting in lists instead of indovidual values, or having a slot in additional properties thats a list of all the individual usagr details.

SteveSandersonMS:
> That's the default, fwiw

Thanks for the reminder. I usually expect bool flags to default to false (in ASP.NET Core that's pretty much a rule, even if it makes naming harder) but it does make sense that we have this behavior as default.

That adds extra weight to the validity of this design.

> Another approach would be to include all of the additional properties in the final one, and assume that anything that's a property with an int value is summable. Thats not necessarily a valid assumption, but we could do it anyway. Or find another way to include everything, like putting in lists instead of indovidual values, or having a slot in additional properties thats a list of all the individual usagr details.

Summing all ints seems very sketchy. Tracking all the UsageDetails in a list in AdditionalProperties would technically preserve the information but it's likely harder to use that information as a consumer because (1) it's not accessible in a strongly-typed way and (2) even if you find the list, it's hard to match up entries with their original sources.

I think the approach you've taken here is good, at least as good as we can do. It retains precision at the cost of a bit of ease-of-use. As long as it's relatively uncommon for people to need to track token usage, there's not much drawback to what this PR does. If it does turn into a common requirement we could look at adding some helper method that does the counting given an `IEnumerable<ChatMessage>` or something like that (or encourage `IChatClient` implementations to provide one, since they know about their own custom usage details extension data).

lucasmeijer:
> I liked this PR's approach because it logically matches the streaming case, but I understand the hesitancy and the rationale that the resulting ChatCompletion should represent the totality of the operation. I'm just concerned about the summing due to the financial aspect of tokens.

Original issue submitter here:  as a user of this library, the reason I use it is to abstract away the small and big differences between the different LLM providers.   Being able to determine cost, or at least relevant token counts seems like a good idea from a user perspective, implementation challenges aside.

I have some sympathy for the hesitancy due to financial impacts of tokens, but right now you're just punting this problem to the user, who is probably more likely to get the logic wrong.

Sidenote: when working with raw LLM's over http, especially in streaming mode, I've found semantic differences between providers.  IIRC anthropic's usage data looked like I was supposed to only take the last one, and they did the summing on their end,  where IIRC openai's ones looked like the responsibility for the summing is on my end.

Tricky situation.  Ideally there was a nice abstraction library that would shield me from this problem :-)

SteveSandersonMS:
> How so? This makes all the data available, and in a standard way.

My guess is that @lucasmeijer means there's some nontrivial, possibly error-prone step of converting that raw data into the desired interpretation, which means interpreting and aggregating vendor-specific data in a useful way that ends up exactly matching real-world costs. @lucasmeijer would prefer this library to take on the burden of handling that so that individual apps don't have to. If we don't, we're "punting this problem to the user".

I can totally sympathise with that. However I think it would have to be some kind of additional abstraction within this library rather than being something we just do automatically. This library can't just innately know how to aggregate cost data for all LLM backends; we'd need some abstraction that LLM backends can optionally implement in order to do that aggregation based on their vendor-specific data. This is what I meant above with "encourage IChatClient implementations to provide [counting]".

stephentoub:
> I can totally sympathise with that. However I think it would have to be some kind of additional abstraction within this library rather than being something we just do automatically.

Alternatively,  automatically sum input/output/total counts and
- Explicitly drop UsageDetails.AdditionalProperties on those sum instances, or
- Declare by design that numeric properties in UsageDetails.AdditionalProperties are considered summable, or
- Add a separate AdditionalCounts dictionary that's summable, or
- Put a collection of the original UsageDetails into AdditionalProperties, or
- ...

SteveSandersonMS:
Of those, the `AdditionalCounts` concept seems most interesting as it avoids the pitfalls of the others (either discarding info because we don't understand it, making assumptions we can't be sure of, or failing to do the aggregation so it remains the app's problem).

I like the idea of `UsageDetails` containing an `AdditionalCounts` dictionary that we declare must only contain summable values. That seems a lot more useful than `AdditionalProperties`, to such an extent I wonder if `UsageDetails` should even have `AdditionalProperties` anyway (anything you can put there could also be stored at the `ChatCompletion` level).

stephentoub:
> I like the idea of UsageDetails containing an AdditionalCounts dictionary that we declare must only contain summable values. That seems a lot more useful than AdditionalProperties, to such an extent I wonder if UsageDetails should even have AdditionalProperties anyway (anything you can put there could also be stored at the ChatCompletion level).

@SteveSandersonMS, want to submit a pr?

SteveSandersonMS:
> want to submit a pr?

https://github.com/dotnet/extensions/pull/5707

PR #5677, opened 2024-11-20
Expose a schema transformer on AIJsonSchemaCreateOptions.

eiriktsarpalis:
With https://github.com/eiriktsarpalis/stj-schema-mapper being migrated to M.E.AI and archived, `AIJsonUtilities` has become the de facto standard API for exporting JSON schemas that's available in both .NET 8 and .NET 9. Partner teams that have been relying on `stj-schema-mapper` source copies will soon need to start using `AIJsonUtilities` instead. 

In order to make this functionality more useful from a customizability perspective, this PR exposes the schema transformer functionality on `AIJsonSchemaCreateOptions`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5677)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateContext.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateContext.cs
@@ -1,0 +1,101 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json.Schema;
+using System.Text.Json.Serialization.Metadata;
+
+#pragma warning disable CA1815 // Override equals and operator equals on value types
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Defines the context in which a JSON schema within a type graph is being generated.
+/// </summary>
+public readonly struct AIJsonSchemaCreateContext
+{
+    private readonly JsonSchemaExporterContext _exporterContext;
+
+    internal AIJsonSchemaCreateContext(JsonSchemaExporterContext exporterContext)
+    {
+        _exporterContext = exporterContext;
+    }
+
+    /// <summary>
+    /// Gets the path to the schema document currently being generated.
+    /// </summary>
+    public ReadOnlySpan<string> Path => _exporterContext.Path;
+
+    /// <summary>
+    /// Gets the <see cref="JsonTypeInfo"/> for the type being processed.
+    /// </summary>
+    public JsonTypeInfo TypeInfo => _exporterContext.TypeInfo;
+
+    /// <summary>
+    /// Gets the type info for the polymorphic base type if generated as a derived type.
+    /// </summary>
+    public JsonTypeInfo? BaseTypeInfo => _exporterContext.BaseTypeInfo;
+
+    /// <summary>
+    /// Gets the <see cref="JsonPropertyInfo"/> if the schema is being generated for a property.
+    /// </summary>
+    public JsonPropertyInfo? PropertyInfo => _exporterContext.PropertyInfo;
+
+    /// <summary>
+    /// Gets the declaring type of the property or parameter being processed.
+    /// </summary>
+    public Type? DeclaringType =>
+#if NET9_0_OR_GREATER
+        _exporterContext.PropertyInfo?.DeclaringType;
+#else
+        _exporterContext.DeclaringType;
+#endif
+
+    /// <summary>
+    /// Gets the <see cref="ICustomAttributeProvider"/> corresponding to the property or field being processed.
+    /// </summary>
+    public ICustomAttributeProvider? PropertyAttributeProvider =>
+#if NET9_0_OR_GREATER
+        _exporterContext.PropertyInfo?.AttributeProvider;
+#else
+        _exporterContext.PropertyAttributeProvider;
+#endif
+
+    /// <summary>
+    /// Gets the <see cref="System.Reflection.ICustomAttributeProvider"/> of the
+    /// constructor parameter associated with the accompanying <see cref="PropertyInfo"/>.
+    /// </summary>
+    public ICustomAttributeProvider? ParameterAttributeProvider =>
+#if NET9_0_OR_GREATER
+        _exporterContext.PropertyInfo?.AssociatedParameter?.AttributeProvider;
+#else
+        _exporterContext.ParameterInfo;
+#endif
+
+    /// <summary>
+    /// Retrieves a custom attribute of a specified type that is applied to the specified schema node context.
+    /// </summary>
+    /// <typeparam name="TAttribute">The type of attribute to search for.</typeparam>
+    /// <param name="inherit">If <see langword="true"/>, specifies to also search the ancestors of the context members for custom attributes.</param>
+    /// <returns>The first occurrence of <typeparamref name="TAttribute"/> if found, or <see langword="null"/> otherwise.</returns>
+    /// <remarks>
+    /// This helper method resolves attributes from context locations in the following order:
+    /// <list type="number">
+    /// <item>Attributes specified on the property of the context, if specified.</item>
+    /// <item>Attributes specified on the constructor parameter of the context, if specified.</item>
+    /// <item>Attributes specified on the type of the context.</item>
+    /// </list>
+    /// </remarks>
+    public TAttribute? GetCustomAttribute<TAttribute>(bool inherit = false)
+        where TAttribute : Attribute
+    {
+        return GetCustomAttr(PropertyAttributeProvider) ??
+            GetCustomAttr(ParameterAttributeProvider) ??
+            GetCustomAttr(TypeInfo.Type);
+
+        TAttribute? GetCustomAttr(ICustomAttributeProvider? provider) =>
+            (TAttribute?)provider?.GetCustomAttributes(typeof(TAttribute), inherit).FirstOrDefault();
+    }
+}
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateContext.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateContext.cs
@@ -14,6 +14,10 @@ namespace Microsoft.Extensions.AI;
 /// <summary>
 /// Defines the context in which a JSON schema within a type graph is being generated.
 /// </summary>
+/// <remarks>
+/// This struct is being passed to the user-provided <see cref="AIJsonSchemaCreateOptions.TransformSchemaNode"/> 
+/// callback by the <see cref="AIJsonUtilities.CreateJsonSchema"/> method and cannot be instantiated directly.
+/// </remarks>
 public readonly struct AIJsonSchemaCreateContext
 {
     private readonly JsonSchemaExporterContext _exporterContext;
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -40,20 +40,4 @@ public sealed class AIJsonSchemaCreateOptions
     /// Gets a value indicating whether to mark all properties as required in the schema.
     /// </summary>
     public bool RequireAllProperties { get; init; } = true;
-
-    /// <summary>
-    /// Gets a value indicating whether to filter keywords that are disallowed by certain AI vendors.
-    /// </summary>
-    /// <remarks>
-    /// Filters a number of non-essential schema keywords that are not yet supported by some AI vendors.
-    /// These include:
-    /// <list type="bullet">
-    /// <item>The "minLength", "maxLength", "pattern", and "format" keywords.</item>
-    /// <item>The "minimum", "maximum", and "multipleOf" keywords.</item>
-    /// <item>The "patternProperties", "unevaluatedProperties", "propertyNames", "minProperties", and "maxProperties" keywords.</item>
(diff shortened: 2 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -443,30 +430,46 @@ private static int IndexOf(this JsonObject jsonObject, string key)
...
+        public string? ParameterName { get; }
+        public string? Description { get; }
+        public bool HasDefaultValue { get; }
+        public object? DefaultValue { get; }
+        public bool IncludeSchemaKeyword { get; }
+        public bool DisallowAdditionalProperties { get; }
+        public bool IncludeTypeInEnumSchemas { get; }
+        public bool RequireAllProperties { get; }
+        public bool FilterDisallowedKeywords { get; }
+        public Func<AIJsonSchemaCreateContext, JsonNode, JsonNode>? TransformSchemaNode { get; }
```
eiriktsarpalis, line 473: Note that this is the first time a type with reference equality is being introduced to the components of the key. While there is precedent in us doing this elsewhere (notably in the `JsonSerializerOptions` cache in STJ), an application using ephemeral transformer delegates would quickly saturate the cache.

I briefly considered nesting another ConditionalWeakTable that is keyed on the delegate, but I suspect this is overkill.

eiriktsarpalis: A viable alternative would involve adding the JSO as a property of `AIJsonSchemaCreateOptions` and then key the existing CWT on that. It is a breaking change though.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -13,6 +16,11 @@ public sealed class AIJsonSchemaCreateOptions
     /// </summary>
     public static AIJsonSchemaCreateOptions Default { get; } = new AIJsonSchemaCreateOptions();
 
+    /// <summary>
+    /// Gets a callback that is invoked for every schema that is generated within the type graph.
```
stephentoub, line 20: I'm losing track. How much of the functionality this is providing is about creating something that works without STJ 9.0, and how much of this is about functionality that's relevant outside of AI scenarios? I'm concerned we're using M.E.AI.Abstractions as a way to ship something that's not specific to AI and that will become instant legacy when the equivalent functionality is added to STJ, or that we're adding functionality to M.E.AI.Abstractions that's already legacy when targeting STJ 9.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateContext.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Text.Json.Schema;
+using System.Text.Json.Serialization.Metadata;
+
+#pragma warning disable CA1815 // Override equals and operator equals on value types
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Defines the context in which a JSON schema within a type graph is being generated.
+/// </summary>
```
stephentoub, line 16: How is someone supposed to create one of these? There's no public ctor, and it looks like using `default(AIJsonSchemaCreateContext)` will blow up if any of its members are used. Presumably then this is only supposed to be created by this component, in order to be passed into the delegate? Consider mentioning that here.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
eiriktsarpalis: > How much of the functionality this is providing is about creating something that works without STJ 9.0, and how much of this is about functionality that's relevant outside of AI scenarios?

It's really all about the former. Granted, we do provide flags that make AI specific tweaks like requiring all properties etc. but these are relatively straightforward to recreate and not strictly speaking necessary.

> I'm concerned we're using M.E.AI.Abstractions as a way to ship something that's not specific to AI

Haven't we done so already? If the concern is about fit we could perhaps entertain the possibility of shipping it as a separate package (say M.E.JsonSchema) that polyfills the implementation for older versions of STJ. This would still show up as a dependency for M.E.AI.Abstractions though.

stephentoub: > Haven't we done so already?

As implementation detail. This appears to be adding public surface area with little other purpose than replicating what's in STJ 9.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: One possible option is inserting the JSO as a property on the `AIJsonCreateSchemaOptions` and keying the CWT on instances of the latter. This would allow caching of schemas for standalone types but disable caching of AIFunction parameters since they provide their own JSO -- this is probably fine though since schemas are created and stored when the AIFunction is constructed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
eiriktsarpalis: > > Haven't we done so already?
> 
> As implementation detail.

We already expose public functions on `AIJsonUtilities` for exporting schemas.

> This appears to be adding public surface area with little other purpose than replicating what's in STJ 9.

Sure, but I think there is a good reason for exposing transformer delegates specifically. It unblocks users facing restrictions from particular AI vendors and removes the onus from us to update M.E.AI.Abstractions so that these restrictions are supported. In other words, I want [this kind of code currently present in SK](https://github.com/microsoft/semantic-kernel/blob/f5facce9ec0dc86e97a13dad0ac4c2476f218bb7/dotnet/src/Connectors/Connectors.OpenAI/Core/OpenAIJsonSchemaTransformer.cs#L31-L49) to still be possible.

stephentoub: > We already expose public functions on AIJsonUtilities for exporting schemas.

Yes, and we did even when this was only ever using STJ 9, providing helpers specific to the scenarios needed by AI consumers.

> It unblocks users facing restrictions from particular AI vendors and removes the onus from us to update M.E.AI.Abstractions so that these restrictions are supported

So then it's not just exposing STJ functionality?

eiriktsarpalis: > > We already expose public functions on AIJsonUtilities for exporting schemas.
> 
> Yes, and we did even when this was only ever using STJ 9, providing helpers specific to the scenarios needed by AI consumers.

I'm probably not understanding what your feedback is. I thought the issue was with exposing schema APIs in M.E.AI.

> So then it's not just exposing STJ functionality?

This PR is forwarding STJ functionality (and the corresponding polyfill functionality) to the M.E.AI schema API.

stephentoub: > I'm probably not understanding what your feedback is.

I'll rephrase.

If M.E.AI.Abstractions required STJ 9 and had zero need for polyfills, would we still be adding this?

And if yes, how much of this is something we'll likely see added to STJ vNext making this functionality obsolete?

eiriktsarpalis: > If M.E.AI.Abstractions required STJ 9 and had zero need for polyfills, would we still be adding this?

I the _this_ refers to the transformer API specifically, I think probably not. We would just direct users to use the equivalent thing in the STJ layer. If it refers to the presence of `AIJsonSchemaUtilities.CreateSchema` in general, then yes, exposing a component that produces schemas compatible with common AI vendors would still be necessary.

> And if yes, how much of this is something we'll likely see added to STJ vNext making this functionality obsolete?

I find it very unlikely we'd ever backport any of the M.E.AI specific knobs to STJ vNext. FWIW this PR is doing the opposite, surfacing general-purpose functionality already present in STJ and the polyfill in the MEAI layer.

stephentoub: > I the "this" refers to the transformer API specifically

Right, this == this PR.

> I think probably not. We would just direct users to use the equivalent thing in the STJ layer.

Then I'm confused why we're adding it.

We agree we'd still have AIJsonSchemaUtilities.CreateSchema even if we required STJ 9. And we're saying we wouldn't be doing this PR if we required STJ 9. So doesn't that mean there's a path where if someone needs this extra transformer logic we say "use STJ 9"?

I'm obviously missing something.

I'm fine with M.E.AI.Abstractions containing a polyfill for STJ 9 functionality to enable the AI-related surface area we want to have regardless of STJ version and that's still desirable when using STJ 9. I start to be concerned when we add M.E.AI.Abstractions surface area specifically to avoid folks needing to reference STJ 9 and that wouldn't be necessary if they're using STJ 9.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: I ended up disabling caching for inputs that specify a transformer -- https://github.com/dotnet/extensions/pull/5677/commits/bffe54d0aa9eaf96fc0f316d18f7411dd0e5c413

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
stephentoub: I spoke offline with Eirik.

Effectively:
- Even in a world where everyone has STJ 9, we still want AIJsonUtilities.CreateXx because it represents hundreds of lines of utility code a developer needs to write on top of the exporter in STJ.
- We don't anticipate that code moving back into STJ, as much of it is relevant to the AI scenario.
- There's a cliff today in what you can do with AIJsonUtilities, where once you hit it, you can fall back to using STJ 9 directly, but then you need to recreate those hundreds of lines of code.
- By providing this extra transformers knob, we effectively remove that cliff.

Given that, I'm good with adding this.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -155,27 +146,23 @@ public static JsonElement CreateJsonSchema(
...
 
-    private static JsonElement GetJsonSchemaCached(JsonSerializerOptions options, FunctionParameterKey key)
+    private static JsonElement GetJsonSchemaCached(JsonSerializerOptions options, SchemaGenerationKey key)
     {
         options.MakeReadOnly();
-        ConcurrentDictionary<FunctionParameterKey, JsonElement> cache = _schemaCaches.GetOrCreateValue(options);
+        ConcurrentDictionary<SchemaGenerationKey, JsonElement> cache = _schemaCaches.GetOrCreateValue(options);
 
-        if (cache.Count >= CacheSoftLimit)
+        if (cache.Count >= CacheSoftLimit || key.TransformSchemaNode is not null)
```
stephentoub, line 165: cache.Count takes all the locks in the CD and is not cheap. It'd be a bit better to do the TransformSchemaNode check first.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
stephentoub: > I ended up disabling caching for inputs that specify a transformer -- [bffe54d](https://github.com/dotnet/extensions/commit/bffe54d0aa9eaf96fc0f316d18f7411dd0e5c413)

Why do we still need it in the key then?

stephentoub: I see, it looks like it's out of convenience so you can construct the key in one place, look at this property in another, and use a record struct to simplify adding the equality implementation, which requires having this property?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: Yep. Technically I could have extracted it to a separate parameter but this would require more refactoring.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: True, although that would only be optimizing for the case where a transformer is specified, which does not cache and is therefore slow by definition.

stephentoub: Yes, but in the common case, both of these conditions are always checked. It can only help to swap the order.

eiriktsarpalis, line 165:
```suggestion
        if (key.TransformSchemaNode is not null || cache.Count >= CacheSoftLimit)
```

PR #5697, opened 2024-11-26
Add OpenAI serialization helper methods.

eiriktsarpalis:
Very early draft implementation of helper methods for serializing M.E.AI.Abstractions using the OpenAI wire format. Contains three separate commits that should be reviewed separately:

1. a0a66969a328a0bef3444eada5bd8d405c9f5a9d adds a System.Net.ServerSentEvents polyfill.
2. 339d3ad69cae225ec8f42609fbbb56096a552fd8 refactors the chat client so that model mapping is pulled into a standalone class.
3. ac2a89f2616d8789d954b42622c2c545c089d658 provides the initial skeleton implementation of the serialization methods.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5697)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,134 @@
...
+public static class OpenAISerializationHelpers
+{
+    /// <summary>
+    /// Deserializes a stream using the OpenAI wire format into a pair of <see cref="ChatMessage"/> and <see cref="ChatOptions"/> values.
+    /// </summary>
+    /// <param name="stream">The stream containing a message using the OpenAI wire format.</param>
+    /// <param name="options">The <see cref="JsonSerializerOptions"/> governing deserialization of function call content.</param>
+    /// <param name="cancellationToken">A token used to cancel the operation.</param>
+    /// <returns>The deserialized list of chat messages and chat options.</returns>
+    public static async Task<(IList<ChatMessage> Messages, ChatOptions? Options)> DeserializeFromOpenAIAsync(
```
stephentoub, line 33: Is this about deserializing a request, deserializing a non-streaming response, or deserializing a streaming response?

eiriktsarpalis: It's for deserializing requests. API names are largely placeholders, will refine.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,134 @@
...
+        Stream stream,
+        JsonSerializerOptions? options = null,
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(stream);
+        _ = Throw.IfNull(chatCompletion);
+        options ??= AIJsonUtilities.DefaultOptions;
+
+        OpenAI.Chat.ChatCompletion openAiChatCompletion = OpenAIModelMappers.ToOpenAIChatCompletion(chatCompletion);
+        BinaryData binaryData = JsonModelHelpers.Serialize(openAiChatCompletion);
```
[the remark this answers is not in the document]
eiriktsarpalis: The OpenAI models use the [`IJsonModel<T>` pattern](https://learn.microsoft.com/en-us/dotnet/api/system.clientmodel.primitives.ijsonmodel-1.write?view=azure-dotnet) which doesn't support async streaming serialization.

src/Shared/Shared.csproj
```diff
@@ -42,4 +42,8 @@
     <InternalsVisibleToTest Include="$(AssemblyName).Tests" />
     <InternalsVisibleToDynamicProxyGenAssembly2 Include="*" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="ServerSentEvents\" />
+  </ItemGroup>
```
SteveSandersonMS, line 48: Is this redundant now the files exist? It looks like it would be.

eiriktsarpalis: Not sure how this slipped by. Removed.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIJsonContext.cs
```diff
@@ -11,6 +12,7 @@ namespace Microsoft.Extensions.AI;
     UseStringEnumConverter = true,
     DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
     WriteIndented = true)]
-[JsonSerializable(typeof(OpenAIChatClient.OpenAIChatToolJson))]
 [JsonSerializable(typeof(OpenAIRealtimeExtensions.ConversationFunctionToolParametersSchema))]
+[JsonSerializable(typeof(OpenAIModelMappers.OpenAIChatToolJson))]
+[JsonSerializable(typeof(IDictionary<string, object?>))]
```
SteveSandersonMS, line 17: Were we just missing this by accident before? If so it's a shame our tests can't catch it. Or maybe this type was included transitively and you're adding it here to be more explicit.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
```diff
@@ -6,13 +6,97 @@
...
+                        {
+                            // If the content can't be deserialized, leave it as a string.
+                        }
+#pragma warning restore CA1031 // Do not catch general exception types
+                    }
+
+                    yield return new ChatMessage
+                    {
+                        Role = ChatRole.Tool,
+                        Contents = new AIContent[] { new FunctionResultContent(toolMessage.ToolCallId, name: string.Empty, result) },
```
SteveSandersonMS, line 77: Could we look up the name by matching the `ToolCallId` in any preceding `FunctionCallContent`?

I'm not sure what anyone will do with this name, just suggesting this to more closely align with the behavior we have in the opposite direction.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIJsonContext.cs
eiriktsarpalis: It's a new addition -- rather than speculatively using `ToolCallJsonSerializerOptions.GetTypeInfo(typeof(IDictionary<string, object>))` I'm hardcoding a source generated contract for the dictionary. This is because `object` always gets deserialized as `JsonElement` under the hood so it doesn't require user customization.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
```diff
@@ -6,13 +6,97 @@
...
+
+                case ToolChatMessage toolMessage:
+                    string textContent = string.Join(string.Empty, toolMessage.Content.Where(part => part.Kind is ChatMessageContentPartKind.Text).Select(part => part.Text));
+                    object? result = textContent;
+                    if (!string.IsNullOrEmpty(textContent))
+                    {
+#pragma warning disable CA1031 // Do not catch general exception types
+                        try
+                        {
+                            result = JsonSerializer.Deserialize(textContent, options.GetTypeInfo(typeof(object)));
```
SteveSandersonMS: I'm not certain what benefit we provide by doing this deserialization to an unknown type. In practice it will always come out as a `JsonElement` or something, right? If so that means anyone processing the `FunctionResultContent` will still have to deserialize to a particular type to do anything with it, so it makes me wonder why not just leave it as a string unconditionally.

Perhaps the benefit of deserializing to `JsonElement` is that anyone who re-serializes the `FunctionResultContent` will get a correctly round-tripped object instead of unintentionally converting the incoming object into a nested JSON string. Is that the reasoning?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
```diff
@@ -6,13 +6,97 @@
...
+                    yield return new ChatMessage
+                    {
+                        Role = ChatRole.User,
+                        AuthorName = userMessage.ParticipantName,
+                        Contents = FromOpenAIChatContent(userMessage.Content),
+                    };
+                    break;
+
+                case ToolChatMessage toolMessage:
+                    string textContent = string.Join(string.Empty, toolMessage.Content.Where(part => part.Kind is ChatMessageContentPartKind.Text).Select(part => part.Text));
```
SteveSandersonMS, line 58: Why would we join up all the content parts into a single string, instead of emitting them each as separate AIContent items?

I'm not 100% certain if OpenAI's schema is documented clearly enough to say whether they would ever split a JSON value across multiple content items in the same message, or whether they may embed multiple independent function result content items in the same message. But if they ever do have multiple independent function results in the same `tool` message, it would be incorrect to concatenate their texts together.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -0,0 +1,616 @@
...
+                    OpenAI.Chat.ChatResponseFormat.CreateJsonObjectFormat();
+            }
+        }
+
+        return result;
+    }
+
+    private static AITool FromOpenAIChatTool(ChatTool chatTool)
+    {
+        Dictionary<string, object?> additionalProperties = new();
```
stephentoub, line 346: Should we use AdditionalPropertiesDictionary here? It's not required, but it would provide consistency, including using OrdinalIgnoreCase.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -0,0 +1,616 @@
...
+            {
+                Description = "Return parameter",
+                Schema = _defaultParameterSchema,
+            }
+        };
+
+        return new MetadataOnlyAIFunction(metadata);
+    }
+
+    private sealed class MetadataOnlyAIFunction(AIFunctionMetadata metadata) : AIFunction
```
stephentoub, line 378: Separate from this PR, I wonder if we should expose an `AIFunction.Create` that just takes the metadata and a delegate for the InvokeCoreAsync implementation. Pro, it'd avoid needing to define a class like this for anyone who needs it. Con, it might be confusing with AIFunctionFactory.Create.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
eiriktsarpalis: This is just performing the inverse operation of the existing `ToOpenAIChatMessages`. I think `JsonElement` is appropriate because it preserves the invariant of the `FunctionResultContent`s returned by `AIFunctionFactory` and the invoking middleware

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -0,0 +1,616 @@
...
+    [return: NotNullIfNotNull("role")]
+    private static ChatMessageRole? ToOpenAIChatRole(ChatRole? role) =>
+        role switch
+        {
+            null => null,
+            _ when role == ChatRole.System => ChatMessageRole.System,
+            _ when role == ChatRole.User => ChatMessageRole.User,
+            _ when role == ChatRole.Assistant => ChatMessageRole.Assistant,
+            _ when role == ChatRole.Tool => ChatMessageRole.Tool,
+            _ => ChatMessageRole.System,
```
stephentoub, line 519: System seems like the wrong default here; it's effectively promoting something unknown to the most sensitive category. It should probably be User or Assistant.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -0,0 +1,616 @@
...
+        };
```
stephentoub, line 520: Also, the use of switch here with when blocks seems like unnecessary ceremony. It could instead be:
```suggestion
        role is null ? null :
        role == ChatRole.System ? ChatMessageRole.System :
        role == ChatRole.User ? ChatMessageRole.User :
        role == ChatRole.Assistant ? ChatMessageRole.Assistant :
        role == ChatRole.Tool ? ChatMessageRole.Tool :
        ChatMessageRole.System;
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -0,0 +1,616 @@
...
+
+    /// <summary>Converts an Extensions finish reason to an OpenAI finish reason.</summary>
+    private static OpenAI.Chat.ChatFinishReason ToOpenAIFinishReason(ChatFinishReason? finishReason) =>
+        finishReason switch
+        {
+            _ when finishReason == ChatFinishReason.Length => OpenAI.Chat.ChatFinishReason.Length,
+            _ when finishReason == ChatFinishReason.ContentFilter => OpenAI.Chat.ChatFinishReason.ContentFilter,
+            _ when finishReason == ChatFinishReason.ToolCalls => OpenAI.Chat.ChatFinishReason.ToolCalls,
+            _ or null => OpenAI.Chat.ChatFinishReason.Stop,
+        };
```
stephentoub, line 580: Similarly here, this seems like an unnecessarily verbose way to express this.
```suggestion
        finishReason == ChatFinishReason.Length ? OpenAI.Chat.ChatFinishReason.Length :
        finishReason == ChatFinishReason.ContentFilter ? OpenAI.Chat.ChatFinishReason.ContentFilter :
        finishReason == ChatFinishReason.ToolCalls => OpenAI.Chat.ChatFinishReason.ToolCalls :
        OpenAI.Chat.ChatFinishReason.Stop;
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
eiriktsarpalis: Sounds reasonable, although I suspect it would require a separate method group.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
eiriktsarpalis: That was my original intent, but the analyzers in extensions find nested ternary operators to be disagreeable.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -0,0 +1,616 @@
...
+        if (chatCompletion.Usage is UsageDetails usageDetails)
+        {
+            chatTokenUsage = ToOpenAIUsage(usageDetails);
+        }
+
+        return OpenAIChatModelFactory.ChatCompletion(
+            id: chatCompletion.CompletionId,
+            model: chatCompletion.ModelId,
+            createdAt: chatCompletion.CreatedAt ?? default,
+            role: ToOpenAIChatRole(chatCompletion.Message.Role).Value,
```
SteveSandersonMS, line 59: Seems like `ToOpenAIChatRole(chatCompletion.Message.Role)` could be null

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
stephentoub: I continue to question why FunctionResultContent needs the name at all.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
stephentoub: Just suppress it at the top of the file. And subsequently I think we should disable some of these analyzers for the whole repo. @joperezr was amenable for any of these that are overreaching.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
eiriktsarpalis: It's annotated `[return: NotNullIfNotNull("role")]` which is why the compiler isn't complaining about this.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
stephentoub: > unless the OpenAI object model has a way to represent unrecognized role strings.

It doesn't at the moment, but it's likely to get one.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
```diff
@@ -0,0 +1,251 @@
...
+                            var callContent = ParseCallContentFromBinaryData(toolCall.FunctionArguments, toolCall.Id, toolCall.FunctionName);
+                            callContent.RawRepresentation = toolCall;
+
+                            message.Contents.Add(callContent);
+                        }
+                    }
+
+                    if (assistantMessage.Refusal is not null)
+                    {
+                        message.AdditionalProperties ??= new();
```
stephentoub, line 103: Nit: `new()` => `[]`?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
```diff
@@ -0,0 +1,251 @@
...
+        if (parts.Count == 0)
+        {
+            parts.Add(ChatMessageContentPart.CreateTextPart(string.Empty));
+        }
+
+        return parts;
+    }
+
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+    private static readonly Func<ChatCompletionOptions, IList<OpenAI.Chat.ChatMessage>> _getMessagesAccessor =
```
stephentoub, line 238: Can we ensure there's an open issue in the OpenAI repo about exposing the ability to access this information?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
eiriktsarpalis: `FunctionResultContent` doesn't currently model splitting results into multiple parts though. What do you suggest as an alternative? Picking the first part only? Emitting separate results contents for each part? The latter doesn't feel right to me.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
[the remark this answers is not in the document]
stephentoub: @annelo-msft, how are System.ClientModel types intended to be serialized asynchronously?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+        _ = Throw.IfNull(stream);
+        _ = Throw.IfNull(chatCompletion);
+        options ??= AIJsonUtilities.DefaultOptions;
+
+        OpenAI.Chat.ChatCompletion openAiChatCompletion = OpenAIModelMappers.ToOpenAIChatCompletion(chatCompletion, options);
+        BinaryData binaryData = JsonModelHelpers.Serialize(openAiChatCompletion);
+#if NET
+        await stream.WriteAsync(binaryData.ToMemory(), cancellationToken).ConfigureAwait(false);
+#else
+        await stream.WriteAsync(binaryData.ToArray(), 0, binaryData.Length, cancellationToken).ConfigureAwait(false);
```
stephentoub, line 61: This could first try to extract the array from the `ReadOnlyMemory<T>` and only fall back to ToArray when it can't. 99.9% of the time it should be able to.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
stephentoub: Do we think we need to revisit FunctionResultContent?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <param name="stream">The stream containing a message using the OpenAI wire format.</param>
+    /// <param name="cancellationToken">A token used to cancel the operation.</param>
+    /// <returns>The deserialized list of chat messages and chat options.</returns>
+    public static async Task<OpenAIChatCompletionRequest> DeserializeChatCompletionRequestAsync(
+        Stream stream, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(stream);
+
+        BinaryData binaryData = await BinaryData.FromStreamAsync(stream, cancellationToken).ConfigureAwait(false);
+        ChatCompletionOptions openAiChatOptions = JsonModelHelpers.Deserialize<ChatCompletionOptions>(binaryData);
```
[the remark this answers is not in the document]
eiriktsarpalis: Not really. The `IJsonModel<T>` abstraction used by the Azure models inherently does not support streaming. I experimented with using a pooled `JsonDocument` instead of `BinaryData` but with currently available APIs this forces copying the data to two separate buffers. `BinaryData` seemed simple enough.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// <summary>
+    /// Serializes a Microsoft.Extensions.AI streaming completion using the OpenAI wire format.
+    /// </summary>
+    /// <param name="stream">The stream to write the value.</param>
+    /// <param name="streamingChatCompletionUpdates">The streaming chat completions to serialize.</param>
+    /// <param name="options">The <see cref="JsonSerializerOptions"/> governing function call content serialization.</param>
+    /// <param name="cancellationToken">A token used to cancel the serialization operation.</param>
+    /// <returns>A task tracking the serialization operation.</returns>
+    public static Task SerializeAsSseEventsAsync(
```
stephentoub, line 73: SerializeStreamingAsync?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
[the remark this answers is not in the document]
stephentoub: > The IJsonModel<T> abstraction used by the Azure models inherently does not support streaming

Write accepts a Utf8JsonWriter which can be constructed around a Stream. I assume then you mean it doesn't support _asynchronous_ streaming.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        static async IAsyncEnumerable<SseItem<BinaryData>> ToSseEventsAsync(IAsyncEnumerable<OpenAI.Chat.StreamingChatCompletionUpdate> updates)
+        {
+            await foreach (var update in updates.ConfigureAwait(false))
+            {
+                BinaryData binaryData = JsonModelHelpers.Serialize(update);
+                yield return new(binaryData);
+            }
+
+            yield return new(new BinaryData("[DONE]"u8.ToArray()));
```
stephentoub, line 94: Can we cache this BinaryData as a singleton?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
eiriktsarpalis: I think specifying the format on the method name is important, given that all other methods in the class use pure JSON.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,100 @@
...
+        _ = Throw.IfNull(stream);
+        _ = Throw.IfNull(streamingChatCompletionUpdates);
+        options ??= AIJsonUtilities.DefaultOptions;
+
+        var mappedUpdates = OpenAIModelMappers.ToOpenAIStreamingChatCompletionAsync(streamingChatCompletionUpdates, options, cancellationToken);
+        return SseFormatter.WriteAsync(ToSseEventsAsync(mappedUpdates), stream, FormatAsSseEvent, cancellationToken);
+
+        static async IAsyncEnumerable<SseItem<BinaryData>> ToSseEventsAsync(IAsyncEnumerable<OpenAI.Chat.StreamingChatCompletionUpdate> updates)
+        {
+            await foreach (var update in updates.ConfigureAwait(false))
```
stephentoub, line 88: Should this be factoring in the cancellation token?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
stephentoub: The wire format details are irrelevant to the consumer of these APIs, though. They want the response written out as OpenAI would, for non-streaming or streaming. That streaming uses SSE is an implementation detail.

src/Shared/ServerSentEvents/README.md
```diff
@@ -0,0 +1,11 @@
+# System.Net.ServerSentEvents
+
+Polyfill for the System.Net.ServerSentEvents library, including the `SseFormatter` component available in .NET 10.
```
stephentoub, line 3: Do we need `#if NET_10_OR_GREATER` in the relevant files then?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
eiriktsarpalis: I'm not sure that it's necessary. It's being passed both in the enumerator that it wraps and the formatter that consumes this IAE.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
SteveSandersonMS: OK I think I see now, and agree with you.

Previously I thought that `ToolChatMessage` might contain multiple `Content` entries that may each represent separate function results.

However on closer inspection I see that the CallId itself only exists at the `ToolChatMessage` level, and not within any of the `Content` entries. As such it's impossible for the content entries to be separate function results, since they would have no way to indicate which call they are for.

So I agree that all the `Content` entries within a given `ToolChatMessage` must refer to the same function call result. AFAIK we likely could just return the first or even throw if there's more than one. But what you're doing with concatenation seems safe, even if it never actually has to deal with more than one content entry in practice.

src/Shared/ServerSentEvents/README.md
eiriktsarpalis: Given that it's not bundled in the shared framework, it seems orthogonal? If and when we add a dependency to the 10.0.0 version of the package, it would be added for all targets and the polyfill can be removed.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
eiriktsarpalis: Turns out this wasn't necessary since the Sse polyfill includes a ns2.0 implementation for the ROM overload.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
eiriktsarpalis: Done.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
SteveSandersonMS: Nitpicky, but is it worth making a static singleton out of `new BinaryData("[DONE]"u8.ToArray())` for reuse across calls?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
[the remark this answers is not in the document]
eiriktsarpalis: But even we were considering serialization, I don't think using Utf8JsonWriter over streams is something we could consider. IIRC aspnetcore streams will throw an exception if you try to invoke any of the sync APIs.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
eiriktsarpalis: Dupe of https://github.com/dotnet/extensions/pull/5697#discussion_r1878252663. I've pushed an update that lifts it to a static.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatMessage.cs
eiriktsarpalis: It became necessary at some point in one of the intermediate refactorings of `AIFunctionFactory` where parameter marshalling was the responsibility of the leaf clients. Given that we've pivoted back to encapsulating this inside the function itself, it should be possible to mark it optional again.

That being said, trying to recover the function name from the chat history seems reasonable to me.

SteveSandersonMS (approved):
I know there are still some open points (e.g., about async serialization). Besides those this looks good to me.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAISerializationHelpers.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+using OpenAI.Chat;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Defines a set of helpers used to serialize Microsoft.Extensions.AI content using the OpenAI wire format.
+/// </summary>
+public static class OpenAISerializationHelpers
```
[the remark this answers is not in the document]
stephentoub: Seems reasonable. I don't have anything significantly better.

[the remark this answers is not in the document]
SteveSandersonMS: Actually I should have mentioned before I marked 'approve' -

I do wonder whether this functionality should all really be in the same package as the client library. Presumably 99.9% of people will be using MEAI.OpenAI to act as a client, so exposing all the opposite-direction APIs for those people may be confusing, or even limiting for us in the future if we want to reference server-specific helpers but can't do so because we don't want extra package dependencies.

Is there any major payoff from this all being in a single MEAI.OpenAI package? I guess it makes it easier to use `internal` helpers but we could IVT for that if needed.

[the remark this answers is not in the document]
eiriktsarpalis: It's something we could consider assuming the server-side components do accrue additional dependencies (for example when we replace the polyfill with the S.N.SSE package after next November). I think it's a bit premature to do so at this stage -- I don't believe the presence of a helper class is enough to put people off about the purpose of the library.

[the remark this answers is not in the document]
stephentoub: Thinking forward to when the abstraction implementations are in the OpenAI library itself, what would we want the distribution vehicle of these additional helpers to be?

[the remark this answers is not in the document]
SteveSandersonMS: I'm raising it because it would be a breaking change to lift these pieces into another package, so if we think it's likely we want to depend on the SSE package within the forseeable future, we could save ourselves the future trouble of dealing with a breaking change.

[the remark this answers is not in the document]
eiriktsarpalis: > Thinking forward to when the abstraction implementations are in the OpenAI library itself, what would we want the distribution vehicle of these additional helpers to be?

My expectation is that an official OpenAI implementation would bundle both pieces of functionality in the same package. I don't believe the SSE dependency is going to be a huge differentiator -- they already bundle some sort of SSE implementation for the parsing side of things.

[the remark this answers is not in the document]
SteveSandersonMS: It sounds like you're saying yes, there will be a breaking change in the future, but it will be to deprecate MEAI.OpenAI entirely and migrate it into the OpenAI client package. Is that right? If so then I agree we don't really benefit from splitting MEAI.OpenAI at this stage.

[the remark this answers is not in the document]
eiriktsarpalis: I think that's our ideal endgame. Keeping the mappers in sync and up to date with newer versions of the OpenAI library isn't sustainable longer term.

PR #5707, opened 2024-11-29
Support aggregating usage data, and do so in FunctionInvokingChatClient

SteveSandersonMS:
This generally works out fine, though I wonder if I just went too far with `AdditionalUsageValues`. It could just be an `IDictionary<string, object?>` or an `AdditionalPropertiesDictionary`, with the convention that we will sum supported numeric values and discard unknown value types. That would have eliminated most of the code from this PR.

However:

 * This wouldn't stop people from storing non-numeric values `AdditionalUsageValues`, which would then get lost. Or we'd need to implement `IDictionary<...>` and block the writes based on `value.GetType()`.
 * Every read or write would be boxing. Middleware might do that many times during the course of a single `CompleteAsync`.

So I implemented a custom collection that can only hold designated numeric types does so without boxing. It is quite laborious, though not super complicated. This in turn requires custom JSON serialization since we can't expect people providing a custom JSO to attach any nonobvious or private types to their source generator.

I discounted use of `INumber<T>` or `IAdditionOperators<T>` since they aren't available on netstandard2.0. Limiting to a fixed set of types (int/long/float/double/decimal) is highly likely to suffice.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalUsageValues.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalUsageValues.cs
@@ -6,6 +6,8 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Runtime.InteropServices;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
@@ -14,6 +16,7 @@ namespace Microsoft.Extensions.AI;
 /// A key-value store for usage values. Values may be of the following types:
 /// <see cref="int"/>, <see cref="long"/>, <see cref="float"/>, <see cref="double"/>, <see cref="decimal"/>.
 /// </summary>
+[JsonConverter(typeof(AdditionalUsageValuesJsonConverter))]
 public class AdditionalUsageValues : IEnumerable<KeyValuePair<string, object>>
 {
     private readonly Dictionary<string, Entry> _dictionary = new();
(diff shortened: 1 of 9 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/UsageDetails.cs
```diff
@@ -19,8 +21,23 @@ public class UsageDetails
     /// <summary>Gets or sets the total number of tokens used to produce the response.</summary>
     public int? TotalTokenCount { get; set; }
 
-    /// <summary>Gets or sets additional properties for the usage details.</summary>
-    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
```
SteveSandersonMS, line 23: We don't strictly have to remove `AdditionalProperties`, but I think it's a lot clearer if we do.

I'm not sure why someone would need to store usage data on `AdditionalProperties` now they have `AdditionalValues` (or could use `ChatCompletion.AdditionalProperties` if it's not usage data), and if they did so we couldn't aggregate it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalUsageValues.cs
```diff
@@ -0,0 +1,349 @@
...
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// A key-value store for usage values. Values may be of the following types:
+/// <see cref="int"/>, <see cref="long"/>, <see cref="float"/>, <see cref="double"/>, <see cref="decimal"/>.
+/// </summary>
+[JsonConverter(typeof(AdditionalUsageValuesJsonConverter))]
+public class AdditionalUsageValues : IEnumerable<KeyValuePair<string, object>>
+{
```
stephentoub, line 21: Are there other places we think we'll use this collection? I'm wondering if we'd be better off just using an `IDictionary<string, long>` for an `AdditionalCounts` property on UsageDetails, rather than trying to make it applicable to lots of different TValue types. Bonus, that would end up being cheaper for enumeration, since it wouldn't involve boxing.

SteveSandersonMS: I'm not aware of other places this would be used, and yes I'd be happy to eliminate all this code.

The reason I didn't just go for `IDictionary<string, long>` is thinking there may be other summable value types people would want - floats and decimals in particular, e.g., to represent things like CPU/GPU time consumed and costs. Perhaps that's over-generalized and we can rely on people modelling things as whole-numbered `long` values anyway (e.g., GPU-microseconds and something like thousandths-of-cents or whatever smallest currency unit is needed).

Do you have a view on what's needed?

SteveSandersonMS: Including `TimeSpan` on the discriminated union here would be fine, however it's not a huge benefit vs storing a nanosecond count as a `long` (especially since Ollama itself is documented to produce nanosecond counts). From my point of view this doesn't strongly push us one way or the other.

However I do agree we should move the timing info onto the `UsageDetails` so it can be aggregated. I'll update my PR to do that, once we've decided whether we're going with this one or #5709.

SteveSandersonMS:
Closing in favour of #5709.

@stephentoub and I discussed, and we agreed that staying with the simpler option of only supporting `long` "additional count" data is preferable as it covers the known scenarios and avoids introducing a nontrivial extra bit of machinery (the custom discriminated-union cache, which could end up seeming quite retro in the long term if C# gets native DUs).

In the future, if we have to add an `AdditionalValues` property to `UsageDetails` supporting other types than `long` we could still do so, even if it partially repeats the functionality in `AdditionalCounts`.

PR #5709, opened 2024-12-02
Usage aggregation via Dictionary<string, long>

SteveSandersonMS:
This is a much simpler version of https://github.com/dotnet/extensions/pull/5707 in case we prefer this approach. It limits additional usage data to be of type `long`, with the payoff that we require much less implementation and new public API surface.

If we think this will suffice, it's certainly advantageous vs #5707. Arguably we could proceed with this as-is, and then in the future if further "additional usage" value types are needed, we'd still have the option to add some other property called `AdditionalValues` (or similar) to `UsageDetails` which accepts a wider range of types.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5709)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
@@ -100,7 +100,6 @@ public sealed class OllamaChatClient : IChatClient
             CompletionId = response.CreatedAt,
             ModelId = response.Model ?? options?.ModelId ?? Metadata.ModelId,
             CreatedAt = DateTimeOffset.TryParse(response.CreatedAt, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset createdAt) ? createdAt : null,
-            AdditionalProperties = ParseOllamaChatResponseProps(response),
             FinishReason = ToFinishReason(response),
             Usage = ParseOllamaChatResponseUsage(response),
         };
@@ -153,7 +152,6 @@ public sealed class OllamaChatClient : IChatClient
             {
                 Role = chunk.Message?.Role is not null ? new ChatRole(chunk.Message.Role) : null,
                 CreatedAt = DateTimeOffset.TryParse(chunk.CreatedAt, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset createdAt) ? createdAt : null,
-                AdditionalProperties = ParseOllamaChatResponseProps(chunk),
                 FinishReason = ToFinishReason(chunk),
                 ModelId = modelId,
             };
@@ -193,29 +191,24 @@ public sealed class OllamaChatClient : IChatClient
 
     private static UsageDetails? ParseOllamaChatResponseUsage(OllamaChatResponse response)
     {
-        if (response.PromptEvalCount is not null || response.EvalCount is not null)
+        AdditionalPropertiesDictionary<long>? additionalCounts = null;
+        OllamaUtilities.TransferNanosecondsTime(response, static r => r.LoadDuration, "load_duration", ref additionalCounts);
+        OllamaUtilities.TransferNanosecondsTime(response, static r => r.TotalDuration, "total_duration", ref additionalCounts);
+        OllamaUtilities.TransferNanosecondsTime(response, static r => r.PromptEvalDuration, "prompt_eval_duration", ref additionalCounts);
+        OllamaUtilities.TransferNanosecondsTime(response, static r => r.EvalDuration, "eval_duration", ref additionalCounts);
+
+        if (additionalCounts is not null || response.PromptEvalCount is not null || response.EvalCount is not null)
         {
             return new()
             {
                 InputTokenCount = response.PromptEvalCount,
                 OutputTokenCount = response.EvalCount,
                 TotalTokenCount = response.PromptEvalCount.GetValueOrDefault() + response.EvalCount.GetValueOrDefault(),
+                AdditionalCounts = additionalCounts,
             };
         }
 
         return null;
-    }
-
-    private static AdditionalPropertiesDictionary? ParseOllamaChatResponseProps(OllamaChatResponse response)
-    {
-        AdditionalPropertiesDictionary? metadata = null;
-
-        OllamaUtilities.TransferNanosecondsTime(response, static r => r.LoadDuration, "load_duration", ref metadata);
-        OllamaUtilities.TransferNanosecondsTime(response, static r => r.TotalDuration, "total_duration", ref metadata);
-        OllamaUtilities.TransferNanosecondsTime(response, static r => r.PromptEvalDuration, "prompt_eval_duration", ref metadata);
-        OllamaUtilities.TransferNanosecondsTime(response, static r => r.EvalDuration, "eval_duration", ref metadata);
-
-        return metadata;
     }
 
     private static ChatFinishReason? ToFinishReason(OllamaChatResponse response) =>
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaEmbeddingGenerator.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaEmbeddingGenerator.cs
@@ -126,17 +126,18 @@ public sealed class OllamaEmbeddingGenerator : IEmbeddingGenerator<string, Embed
         }
 
         // Convert response into result objects.
-        AdditionalPropertiesDictionary? responseProps = null;
-        OllamaUtilities.TransferNanosecondsTime(response, r => r.TotalDuration, "total_duration", ref responseProps);
-        OllamaUtilities.TransferNanosecondsTime(response, r => r.LoadDuration, "load_duration", ref responseProps);
+        AdditionalPropertiesDictionary<long>? additionalCounts = null;
+        OllamaUtilities.TransferNanosecondsTime(response, r => r.TotalDuration, "total_duration", ref additionalCounts);
+        OllamaUtilities.TransferNanosecondsTime(response, r => r.LoadDuration, "load_duration", ref additionalCounts);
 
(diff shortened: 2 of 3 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/UsageDetails.cs
```diff
@@ -19,8 +22,25 @@ public class UsageDetails
     /// <summary>Gets or sets the total number of tokens used to produce the response.</summary>
     public int? TotalTokenCount { get; set; }
 
-    /// <summary>Gets or sets additional properties for the usage details.</summary>
-    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+    /// <summary>Gets additional usage counts.</summary>
+    [JsonObjectCreationHandling(JsonObjectCreationHandling.Populate)]
+    public IDictionary<string, long> AdditionalCounts { get; } = new Dictionary<string, long>(StringComparer.OrdinalIgnoreCase);
```
stephentoub, line 27: This differs from how we've done AdditionalProperties everywhere else. Why have this be get-only with populate rather than being a nullable property with a setter?

SteveSandersonMS: > Is it because this is using the interface and so this is the only way to guarantee what comparer is used

Yes that's exactly it. Or in fact you could generalize the concern to say that allowing settable `IDictionary`-type properties anywhere is dubious if these are intended as exchange types (or we could say "collaboration types") that multiple unrelated pieces of software are meant to use as a shared workspace. Whoever writes the dictionary first gets to pick its semantics, which might have arbitrary behaviors unlike a standard `Dictionary<T, U>`.

Swapping this to use `AdditionalPropertiesDictionary` (which becomes generic) makes sense. I'll see if it works out.

SteveSandersonMS: This is done. It works out OK.

The only drawback is we need to either:

 * Explicitly surface `AdditionalPropertiesDictionary<object?>` in a bunch of places on the public API, which makes it look and feel more complicated
 * Or, also derive a subclass `AdditionalPropertiesDictionary` with TValue=object, which is totally fine but then the base type has to be unsealed

In practice I don't know of a problem from having the generic base class be unsealed. It doesn't even have any `virtual` members. So I think keeping the public API surface looking clean in all other places outweighs this, and hence I have added a nongeneric subclass too.

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalPropertiesDictionary{TValue}.cs
```diff
@@ -0,0 +1,249 @@
...
+    public ICollection<string> Keys => _dictionary.Keys;
+
+    /// <inheritdoc />
+    public ICollection<TValue> Values => _dictionary.Values;
+
+    /// <inheritdoc />
+    public int Count => _dictionary.Count;
+
+    /// <inheritdoc />
+    public bool IsReadOnly => false;
```
stephentoub, line 78: Dogs you purposefully change this to be public rather than explicitly implementing the interface?

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalPropertiesDictionary{TValue}.cs
```diff
@@ -0,0 +1,249 @@
...
+            }
+        }
+
+        // Unable to find the value or convert it to the requested type.
+        value = default;
+        return false;
+    }
+
+    /// <inheritdoc />
+    public bool TryGetValue(string key, [NotNullWhen(true)] out TValue value) => _dictionary.TryGetValue(key, out value!);
```
stephentoub, line 193: This attribute is no longer correct. It should be MaybeNullWhen(false).

src/Libraries/Microsoft.Extensions.AI.Abstractions/UsageDetails.cs
```diff
@@ -19,8 +21,34 @@ public class UsageDetails
     /// <summary>Gets or sets the total number of tokens used to produce the response.</summary>
     public int? TotalTokenCount { get; set; }
 
-    /// <summary>Gets or sets additional properties for the usage details.</summary>
-    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+    /// <summary>Gets or sets a dictionary of additional usage counts.</summary>
```
stephentoub, line 24: Maybe add a remark about how all values in here are considered to be summable?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -370,6 +344,50 @@ private sealed class FunctionCallInfo
...
+            }
+
+            destination.AdditionalCounts.Add(
+                $"{nameof(ChatTokenUsage.OutputTokenDetails)}.{nameof(ChatOutputTokenUsageDetails.ReasoningTokenCount)}",
+                outputDetails.ReasoningTokenCount);
+        }
+
+        return destination;
+    }
+
```
stephentoub, line 390: Were you also going to update the ollama impl to move the timings to usage details?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalPropertiesDictionary.cs
```diff
@@ -1,248 +1,39 @@
...
-[DebuggerDisplay("Count = {Count}")]
-public sealed class AdditionalPropertiesDictionary : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
+public sealed class AdditionalPropertiesDictionary : AdditionalPropertiesDictionary<object?>
 {
-    /// <summary>The underlying dictionary.</summary>
-    private readonly Dictionary<string, object?> _dictionary;
-
     /// <summary>Initializes a new instance of the <see cref="AdditionalPropertiesDictionary"/> class.</summary>
     public AdditionalPropertiesDictionary()
+        : base()
```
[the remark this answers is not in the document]
SteveSandersonMS: No, you're correct. I was just editing the code mechanically. Removed this.

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalPropertiesDictionary{TValue}.cs
SteveSandersonMS: Just changing the attribute to `[MaybeNullWhen(false)]` results in build errors, e.g.:

```
Nullability of reference types in type of parameter 'value' of 'bool AdditionalPropertiesDictionary<TValue>.TryGetValue(string key, out TValue value)' doesn't match implicitly implemented member 'bool IDictionary<string, TValue>.TryGetValue(string key, out TValue value)' (possibly because of nullability attributes).
```

I'm not super familiar with this but it looks like it's an API difference across .NET versions. Does the following look right to you?

```cs
#if NET
    /// <inheritdoc />
    public bool TryGetValue(string key, [MaybeNullWhen(false)] out TValue value) => _dictionary.TryGetValue(key, out value);
#else
    /// <inheritdoc />
    public bool TryGetValue(string key, [NotNullWhen(true)] out TValue value) => _dictionary.TryGetValue(key, out value!);
#endif
```

stephentoub: > Does the following look right to you?

Unfortunately, no. It means downlevel you could have TryGrtValue return true,  have the value be null (because null was put into the dictionary as a valid object? value), but then not get an nrt warning trying to deference the value.

I'm on my phone right now. For now,  can you add an explicit impl of the interface method,  with whatever suppressions you need to make it compile, separate from the public method annotated correctly? The annotations on the explicitly impl won't matter,  and we can sort it out to try to remove those extra impls later.

src/Libraries/Microsoft.Extensions.AI.Abstractions/AdditionalPropertiesDictionary{TValue}.cs
SteveSandersonMS: Yes, after unsealing, it was to deal with:

```
error CA1033: Make 'AdditionalPropertiesDictionary' sealed (a breaking change if this class has previously shipped), implement the method non-explicitly, or implement a new method that exposes the functionality of 'System.Collections.Generic.ICollection<System.Collections.Generic.KeyValuePair<System.String,TValue>>.get_IsReadOnly' and is visible to derived classes (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1033)
```

Same applies to `Contains` and `CopyTo`.

Do you feel we should solve this in another way?

SteveSandersonMS: > For example I could just suppress S4039 and CA1033 in this file.

I've done that for now. Given that the three members concerned (`IsReadOnly`, `Contains(keyvaluepair)` and `CopyTo(collection of keyvaluepair)`) are all extremely unlikely to be useful on this type, making them hard to access is unlikely to be a problem.

stephentoub: I'd just suppress the diagnostic here.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
SteveSandersonMS: Note that I did *not* add a `_ns` suffix to the keys, based on our separate discussion.

Issue #5583, opened 2024-10-29
[Microsoft.Extensions.AI] Add EndUserId as tag to span/traces/metrics


kzu:
It's typically quite useful to be able to associate genai telemetry with a particular user, for multiple reasons (i.e. track token consumption per user, rate limit per user, etc.). OpenAI and Claude support such an identifier specifically, but I couldn't find a similar mechanism for Gemini or Cohere.

While this is easy to do add this tag for the activity/span itself (via a delegating chat client that adds that to the Activity.Current), it's impossible to do the same for the token consumption and operation duration metrics.

It would be ideal if arbitrary (serializable/primitive?) properties in the `ChatOptions.AdditionalProperties` were automatically added to both traces and metrics automatically by the `OpenTelemetryChatClient`, or alternatively, at least just the `EndUserId`. It could also be an explicit setting for the client, to be configured just like `EnableSensitiveData`, such as `EnableAdditionalProperties` or `EnableAdditionalTags`).

----
Workaround: how to add EndUserId to traces (not metrics though)

```csharp
    class UserIdChatClient(IChatClient client) : DelegatingChatClient(client)
    {
        public override Task<Microsoft.Extensions.AI.ChatCompletion> CompleteAsync(IList<Microsoft.Extensions.AI.ChatMessage> chatMessages, ChatOptions? options = null, CancellationToken cancellationToken = default)
        {
            if (Activity.Current is { } activity && options?.AdditionalProperties?.TryGetValue("EndUserId", out var endUserId) == true)
                activity.SetTag("user.id", endUserId);

            return base.CompleteAsync(chatMessages, options, cancellationToken);
        }

        public override IAsyncEnumerable<Microsoft.Extensions.AI.StreamingChatCompletionUpdate> CompleteStreamingAsync(IList<Microsoft.Extensions.AI.ChatMessage> chatMessages, ChatOptions? options = null, CancellationToken cancellationToken = default)
        {
            if (Activity.Current is { } activity && options?.AdditionalProperties?.TryGetValue("EndUserId", out var endUserId) == true)
                activity.SetTag("user.id", endUserId);

            return base.CompleteStreamingAsync(chatMessages, options, cancellationToken);
        }
    }
```

And ensure it's added right after the open telemetry client:

```csharp
services.AddChatClient(builder => builder
        .UseOpenTelemetry()
        .Use(client => new UserIdChatClient(client))
        .Use(new OpenAIClient(...));
```

stephentoub:
[...]

PR #5712, opened 2024-12-02
Update otel chat client / embedding generator for 1.29 draft of the spec

stephentoub:
Also address feedback to include additional properties as tags.
Closes https://github.com/dotnet/extensions/issues/5583
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5712)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
@@ -136,17 +136,33 @@ public static class StreamingChatCompletionUpdateExtensions
     {
         if (messages.Count <= 1)
         {
+            // Add the single message if there is one.
             foreach (var entry in messages)
             {
                 AddMessage(completion, coalesceContent, entry);
             }
+
+            // In the vast majority case where there's only one choice, promote any additional properties
+            // from the single message to the chat completion, making them more discoverable and more similar
+            // to how they're typically surfaced from non-streaming services.
+            if (completion.Choices.Count == 1 &&
+                completion.Choices[0].AdditionalProperties is { } messageProps)
+            {
+                completion.Choices[0].AdditionalProperties = null;
+                completion.AdditionalProperties = messageProps;
+            }
         }
         else
         {
+            // Add all of the messages, sorted by choice index.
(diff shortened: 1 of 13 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
```diff
@@ -136,17 +136,33 @@ private static void AddMessagesToCompletion(Dictionary<int, ChatMessage> message
...
+
+            // In the vast majority case where there's only one choice, promote any additional properties
+            // from the single message to the chat completion, making them more discoverable and more similar
+            // to how they're typically surfaced from non-streaming services.
+            if (completion.Choices.Count == 1 &&
+                completion.Choices[0].AdditionalProperties is { } messageProps)
+            {
+                completion.Choices[0].AdditionalProperties = null;
+                completion.AdditionalProperties = messageProps;
+            }
```
SteveSandersonMS, line 153: If the result sometimes has one choice and sometimes has multiple choices, it may become difficult to know where to find a particular "AdditionalProperties" value. Someone who needs to access it programmatically would have to replicate this logic (i.e., if there's one choice look in one place, otherwise look in another place).

Maybe that's acceptable or can never happen for some reason.

Is there a high payoff from moving the `AdditionalProperties` over?

stephentoub: > Is there a high payoff from moving the AdditionalProperties over?

In the majority case having the same behavior with regards to additional properties between a ChatCompletion from CompleteAsync and one from ToChatCompletion (from a list of streaming updates). Otherwise, with the former top-level properties show up on the ChatCompletion but with the latter they show up on the ChatMessage.

stephentoub: > If the result sometimes has one choice and sometimes has multiple choices

That also seems very unlikely for a given call site. It's possible someone could variable set n, but that would be an explicit action that would already impact things based on one vs multiple choices, like whether automatic function invocation works.

SteveSandersonMS (approved):

Issue #5708, opened 2024-11-30
[Microsoft.Extensions.AI] The Cohere Command R+ always returns "Bad Request"


yamanaiyuki:
### Description

I am creating a program using the AzureAIInferenceChatClient, and there was an issue.
The Mistral-large-2407 is working normally, but the Cohere-command-r-plus-08-2024 throws an exception.
I prepared a short sample for verification.
According to the sample, when CompleteAsync() is executed, "400 (Bad Request)" is returned.

Sample Code:

````csharp
using Azure;
using Azure.AI.Inference;
using Microsoft.Extensions.AI;

string? AZURE_COHERE_ENDPOINT = Environment.GetEnvironmentVariable("AZURE_COHERE_ENDPOINT");
string? AZURE_COHERE_API_KEY = Environment.GetEnvironmentVariable("AZURE_COHERE_API_KEY");
string? modelId = Environment.GetEnvironmentVariable("AZURE_COHERE_MODELID");

ArgumentNullException.ThrowIfNull(AZURE_COHERE_ENDPOINT);
ArgumentNullException.ThrowIfNull(AZURE_COHERE_API_KEY);
ArgumentNullException.ThrowIfNull(modelId);

Uri endpoint = new(AZURE_COHERE_ENDPOINT);
AzureKeyCredential credential = new(AZURE_COHERE_API_KEY);

ChatCompletionsClient chatCompletionsClient = new(endpoint, credential);

Response<ModelInfo> modelInfo = chatCompletionsClient.GetModelInfo();
Console.WriteLine($"Model name: {modelInfo.Value.ModelName}");
Console.WriteLine($"Model type: {modelInfo.Value.ModelType}");
Console.WriteLine($"Model provider name: {modelInfo.Value.ModelProviderName}");

try
{
    AzureAIInferenceChatClient client = new(chatCompletionsClient, modelId);

    ChatCompletion result = await client.CompleteAsync("What is AI?");
    Console.WriteLine(result);
}
catch (Exception ex)
{
    Console.WriteLine(ex);
}
````

Message printed to the console:

````
Model name: Cohere Command R+
Model type: chat-completion
Model provider name: Cohere
Azure.RequestFailedException: {"message":"invalid type: parameter messages.content is of type array but should be of type string"}
Status: 400 (Bad Request)
ErrorCode: Bad Request

Content:
[...]

PR #5714, opened 2024-12-02
Fix handling of text-only user messages in AzureAIInferenceChatClient

stephentoub:
`ChatRequestUserMessage` has three constructors:
```csharp
public ChatRequestUserMessage(string content)
public ChatRequestUserMessage(IEnumerable<ChatMessageContentItem> content)
public ChatRequestUserMessage(params ChatMessageContentItem[] content)
```
but even though all of the parameters are named `content` and represent the message's content, they behave differently. The first assigns the string content to the instance's `Content` property and leaves its `MultimodalContentItems` property null, and the others leave `Content` null and set `MultimodalContentItems` to the property. For models that don't support multi-modal, using the latter two constructors breaks, even when the content is a single text item.

I think this should be improved in Azure.AI.Inference, but regardless, this fixes the ToAzureAIInferenceChatMessages helper to special-case text-only inputs and use the first `string`-based constructor rather than always using the `IEnumerable<ChatMessageContentItem>`-based one.

Fixes https://github.com/dotnet/extensions/issues/5708
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5714)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
@@ -405,7 +407,7 @@ public sealed class AzureAIInferenceChatClient : IChatClient
     }
 
     /// <summary>Converts an Extensions chat message enumerable to an AzureAI chat message enumerable.</summary>
-    private IEnumerable<ChatRequestMessage> ToAzureAIInferenceChatMessages(IEnumerable<ChatMessage> inputs)
+    private IEnumerable<ChatRequestMessage> ToAzureAIInferenceChatMessages(IList<ChatMessage> inputs)
     {
         // Maps all of the M.E.AI types to the corresponding AzureAI types.
         // Unrecognized or non-processable content is ignored.
@@ -441,7 +443,9 @@ public sealed class AzureAIInferenceChatClient : IChatClient
             }
             else if (input.Role == ChatRole.User)
             {
-                yield return new ChatRequestUserMessage(GetContentParts(input.Contents));
+                yield return input.Contents.All(c => c is TextContent) ?
+                    new ChatRequestUserMessage(string.Concat(input.Contents)) :
+                    new ChatRequestUserMessage(GetContentParts(input.Contents));
             }
             else if (input.Role == ChatRole.Assistant)
             {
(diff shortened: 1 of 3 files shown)


stephentoub:
cc: @trangevi

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -441,7 +443,9 @@ private IEnumerable<ChatRequestMessage> ToAzureAIInferenceChatMessages(IEnumerab
             }
             else if (input.Role == ChatRole.User)
             {
-                yield return new ChatRequestUserMessage(GetContentParts(input.Contents));
+                yield return input.Contents.All(c => c is TextContent) ?
+                    new ChatRequestUserMessage(string.Concat(input.Contents)) :
```
SteveSandersonMS, line 447: I know it's a super-niche case, but should we treat separate `Contents` entries as if they have some kind of separator between them, for example a linebreak?

Concatenating with no separator at all could theoretically result in different meanings than were intended.

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
stephentoub: Do we have evidence to suggest that's desirable? With streaming, for example, each TextContent that's produced contains its own separators where necessary, e.g. the streamed contents for "hello world" will often be an update for "hello " and an update for "world", which we then represent as a TextContent for "hello " and a TextContent for "world". I wouldn't want to then artificially insert additional whitespace.

SteveSandersonMS: Another option is we could just pass through the bug/limitation from AzureAIInferenceChatClient without changing any semantics:

 * If the supplied value is exactly one TextContent, we could use the `string` overload that works universally
 * Otherwise, we use the `IEnumerable<ChatMessageContentItem>` overload that may or may not work depending on the backend model. Later this might get fixed in `AzureAIInferenceChatClient` or in the service itself to work universally as long as all the content is text, but that happens independently from this library.

I know in practice it likely isn't harmful to polyfill over the bug/limitation (and if it becomes fixed in the underlying client/service in the future, we could remove our polyfill without thinking of that as a breaking change), so I don't feel strongly about this.

stephentoub: Ok. Until we get any feedback to the contrary, I'm going to leave it as-is.

PR #5718, opened 2024-12-03
Fix streaming function calling

SteveSandersonMS:
There was part of the `FunctionInvokingChatClient` implementation that I couldn't understand, and after investigation, I think it's a bug.

Our tests don't cover any cases where streaming returns multiple updates that contain function calls. I *think* that's a legitimate scenario, but I'm not sure that it can be modelled by the `InvokeAndAssertStreamingAsync` helper we use in those tests, because it treats each "plan" entry as the end of a response.

I've added a test showing what I think is this legitimate response format with multiple updates that have function calls. Without the fix in `FunctionInvokingChatClient`, it only removes one of the two FCCs from the result (and so the other one is both added to history *and* returned, which the method's XML docs say should not happen). The fix is completely trivial.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5718)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -325,7 +325,7 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
                     // If there were any, remove them from the update. We do this before yielding the update so
                     // that we're not modifying an instance already provided back to the caller.
                     int addedFccs = functionCallContents.Count - preFccCount;
-                    if (addedFccs > preFccCount)
+                    if (addedFccs > 0)
                     {
                         update.Contents = addedFccs == update.Contents.Count ?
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -325,7 +325,7 @@ public override async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteSt
                     // If there were any, remove them from the update. We do this before yielding the update so
                     // that we're not modifying an instance already provided back to the caller.
                     int addedFccs = functionCallContents.Count - preFccCount;
-                    if (addedFccs > preFccCount)
+                    if (addedFccs > 0)
```
stephentoub, line 328: Yeah, this is just a bug. If memory serves, I originally had `if (functionCallContents.Count > preFccCount)`, then realized I needed the difference, separated it out into a local, but missed fixing up the rhs.

PR #5730, opened 2024-12-09
Ollama support for streaming function calling and native structured output

### LINKED ISSUE #5726: [Microsoft.Extensions.AI] - Add new Structured output support for JSON in Ollama
Would love to see support for the [new structured outputs](https://ollama.com/blog/structured-outputs) that ollama supports in a new `format` option to describe the json schema object we want to enforce.

```
curl -X POST http://localhost:11434/api/chat -H "Content-Type: application/json" -d '{
  "model": "llama3.1",
  "messages": [{"role": "user", "content": "Tell me about Canada."}],
  "stream": false,
  "format": {
    "type": "object",
    "properties": {
      "name": {
        "type": "string"
      },
      "capital": {
        "type": "string"
      },
      "languages": {
        "type": "array",
        "items": {
          "type": "string"
        }
      }
    },
    "required": [
      "name",
      "capital", 
      "languages"
    ]
  }
}'
```

As far as the API for this, I'd think we could just add a new `Object` option to `ResponseFormat`, but not sure exactly what i'd expect for the actual schema definition offhand?

```
new ChatOptions<MyExpectedShape>()
{
    ResponseFormat = ChatResponseFormat.Object,
    Schema = typeof(MyExpectedShape)
}

// ...

new ChatOptions()
{
    ResponseFormat = ChatResponseFormat.Object,
    Schema = typeof(MyExpectedShape),
    SchemaSerializerOptions = new JsonSerializerOptions
    {
        PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
        DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull
    }
}
```

regardless, this would greatly enhance the capabilities for json handling and would be a great value add!

SteveSandersonMS:
Fixes #5726 by adding support for native structured output added in [Ollama 0.5.0](https://github.com/ollama/ollama/releases/tag/v0.5.0).

Also adds support for streaming function calling added in [Ollama 0.4.6](https://github.com/ollama/ollama/releases/tag/v0.4.6).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5730)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Extensions.AI;
 public sealed class OllamaChatClient : IChatClient
 {
     private static readonly JsonElement _defaultParameterSchema = JsonDocument.Parse("{}").RootElement;
-    private static readonly JsonElement _defaultJsonSchema = JsonDocument.Parse("\"json\"").RootElement;
+    private static readonly JsonElement _schemalessJsonResponseFormatValue = JsonDocument.Parse("\"json\"").RootElement;
 
     /// <summary>The api/chat endpoint URI.</summary>
     private readonly Uri _apiChatEndpoint;
@@ -263,7 +263,7 @@ public sealed class OllamaChatClient : IChatClient
     {
         if (format is ChatResponseFormatJson jsonFormat)
         {
-            return jsonFormat.Schema ?? _defaultJsonSchema;
+            return jsonFormat.Schema ?? _schemalessJsonResponseFormatValue;
         }
         else
         {
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
@@ -263,14 +263,7 @@ public sealed class OllamaChatClient : IChatClient
     {
         if (format is ChatResponseFormatJson jsonFormat)
         {
-            if (!string.IsNullOrEmpty(jsonFormat.Schema))
-            {
-                return JsonDocument.Parse(jsonFormat.Schema!).RootElement;
-            }
-            else
-            {
-                return _defaultJsonSchema;
-            }
+            return jsonFormat.Schema ?? _defaultJsonSchema;
         }
         else
         {
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.AI;
 public sealed class OllamaChatClient : IChatClient
 {
     private static readonly JsonElement _defaultParameterSchema = JsonDocument.Parse("{}").RootElement;
+    private static readonly JsonElement _defaultJsonSchema = JsonDocument.Parse("\"json\"").RootElement;
 
     /// <summary>The api/chat endpoint URI.</summary>
     private readonly Uri _apiChatEndpoint;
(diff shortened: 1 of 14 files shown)


src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -247,11 +253,36 @@ private static ChatMessage FromOllamaMessage(OllamaChatResponseMessage message)
         return new ChatMessage(new(message.Role), contents);
     }
 
+    private static FunctionCallContent ToFunctionCallContent(OllamaFunctionToolCall function)
+    {
+        var id = Guid.NewGuid().ToString().Substring(0, 8);
```
stephentoub, line 258: Nit: on core this could be:
```C#
var id = RandomNumberGenerator.GetHexString(8); // with `, lowercase = true` if we care about that
```

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -247,11 +253,36 @@ private static ChatMessage FromOllamaMessage(OllamaChatResponseMessage message)
...
+        return new FunctionCallContent(id, function.Name, function.Arguments);
+    }
+
+    private static JsonElement? ToOllamaChatResponseFormat(ChatResponseFormat? format)
+    {
+        if (format is ChatResponseFormatJson jsonFormat)
+        {
+            if (!string.IsNullOrEmpty(jsonFormat.Schema))
+            {
+                return JsonDocument.Parse(jsonFormat.Schema!).RootElement;
```
SteveSandersonMS, line 268: One option is to change `ChatResponseFormatJson.Schema` to be a `JsonElement?` instead of a `string?` so that we don't have to parse it here.

I'm not massively worried about the perf of parsing this, since it's only once-per-call and only in the Ollama native structured output case, this is more a question of what is the right abstraction. There isn't any question that the response format is JSON here (we already know it's a `ChatResponseFormatJson`); the only question is whether it's fair to force JSON schemas to be represented as a `JsonElement`. My sense is that yes, `JsonElement` is at least as reasonable a representation as `string` and is already what we do for the almost-equivalent function calling parameter schema.

@eiriktsarpalis / @stephentoub Do you have an opinion?

stephentoub, line 268: This is going to leak an ArrayPool array. Can you use `JsonSerializer.Deserialize<JsonElement>` instead?

SteveSandersonMS: Thanks for the tip, but please see my other PR comment before we decide on this.

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.AI;
 public sealed class OllamaChatClient : IChatClient
 {
     private static readonly JsonElement _defaultParameterSchema = JsonDocument.Parse("{}").RootElement;
+    private static readonly JsonElement _defaultJsonSchema = JsonDocument.Parse("\"json\"").RootElement;
```
eiriktsarpalis, line 26: What does this string represent? Technically, it's not a valid schema document.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
eiriktsarpalis: Is there anything preventing us from using a full Guid for the identifier?

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
eiriktsarpalis: One benefit of using `JsonElement` over `string` is that it works better when you're serializing it as a nested property. A string value will output the schema as a single escaped JSON string whereas `JsonElement` serializes as a nested value. Ultimately though, it depends on what pre-existing wire format you need to cater to.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
SteveSandersonMS: It's the default value for the `format` property, i.e., it causes us to produce `format: "json"`. I'll rename this to clarify.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
SteveSandersonMS: > JsonElement serializes as a nested value

In the Ollama case, we do want to serialize it as a nested value, which is what prompted this change.

In the OpenAI case, the API requires us to pass it as a serialized UTF8 value (in a `BinaryData`) so this change makes that slightly less convenient, but not enough to be problematic.

Anyway, this is now done.

SteveSandersonMS: Resolving as no longer applicable.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
SteveSandersonMS: > Is there anything preventing us from using a full Guid for the identifier?

We're copying the convention used in OpenAI's function calling convention here, since Ollama doesn't have a clear convention of its own.

My intuition (which I haven't actually confirmed) is that it's the LLM that's matching the function call results to entries in the set of prior function calls, and hence shorter values are more likely to work with it being innately fuzzy. We benefit from the ID being unique, but disbenefit from it being long. 8 chars is the tradeoff that OpenAI chose and we're reusing that decision here. Again, I haven't actually tested this intuition in any systematic way, but doing something we know works is reasonably safe.

SteveSandersonMS: To be honest I'm not even sure that Ollama uses the call IDs at all, but it's not harmful to make them available like this.

eiriktsarpalis (approved):

PR #5748, opened 2024-12-17

mgravell:

Outstanding:

- additional logging
  - add log when L2 data is rejected for (reason)
  - add metric of the number of tag expirations being tracked in L1
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5748)


sebastienros:
I am not sure `GetTimestamp()` should be used to store the moment, I tried this code on Linux and Windows and got different results:

```c#
Console.WriteLine(TimeProvider.System.GetTimestamp());
Console.WriteLine(TimeProvider.System.GetUtcNow());
Console.WriteLine(System.Runtime.InteropServices.RuntimeInformation.OSDescription);
```

```
8226721287470
12/20/2024 5:27:32 PM +00:00
Microsoft Windows 10.0.17763
```

```
5640002254299340
12/20/2024 17:27:24 +00:00
Ubuntu 24.04.1 LTS
```

mgravell:
There is TimestampFrequency, but I will change to use GetUtcNow() instead.

sebastienros:
Use ticks (but same normalization issue) or a sortable numeric format (yyyymmddhhmmssxxxx), or use an EPOCH so 0 starts at 1/1/2025 ;) then numbers are much smaller.

mgravell:
I considered that - if you think it won't be contentious, sure I can use an atypical epoch. At stackoverflow, we used "stack seconds" in a few key places :) however, the fixed size is highly desirable, and I'd rather not get into 32-bit pain. Honestly, I think it is better to just pay for 64 and not worry about it.

jodydonetti:
FWIW `DateTimeOffset.Ticks`/`UtcTicks` seems to be consistent cross platform.

PR #5749, opened 2024-12-18
Update to public versions

RussKie:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5749)

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,13 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
-    <!--  Begin: Package sources from dotnet-aspnetcore -->
-    <add key="darc-pub-dotnet-aspnetcore-401ae7c" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspnetcore-401ae7cb/nuget/v3/index.json" />
-    <!--  End: Package sources from dotnet-aspnetcore -->
-    <!--End: Package sources managed by Dependency Flow automation. Do not edit the sources above.-->
-    <!-- Arcade -->
-    <add key="darc-pub-dotnet-aspnetcore-be19faf1" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspnetcore-be19faf1/nuget/v3/index.json" />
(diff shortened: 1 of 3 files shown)


NuGet.config
```diff
@@ -32,34 +22,14 @@
...
-      <package pattern="*" />
-    </packageSource>
-    <packageSource key="dotnet10-transport">
-      <package pattern="*" />
-    </packageSource>
     <packageSource key="richnav">
       <package pattern="*" />
     </packageSource>
   </packageSourceMapping>
-  <disabledPackageSources>
```
joperezr, line 57: I think it should be better to keep the section, and just leave it empty as it is. That way, when we flow internal dependencies in preparation for our next release, maestro can still add sources bellow in the section that they should go.

eng/Version.Details.xml
```diff
@@ -148,39 +148,39 @@
       <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-runtime</Uri>
       <Sha>9d5a6a9aa463d6d10b0b0ba6d5982cc82f363dc3</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.AspNetCore.App.Ref" Version="9.0.1">
+    <Dependency Name="Microsoft.AspNetCore.App.Ref" Version="9.0.0">
```
joperezr, line 151: Out of curiosity, how did you change these versions? If you did it through darc, I would have expected the Sha's to change.

RussKie: I change those manually. Darc pulls the non-public versions (i.e., 9.0.1)

NuGet.config
RussKie: Maestro can add this section automatically, I believe

NuGet.config
```diff
@@ -32,34 +22,14 @@
...
-    <packageSource key="dotnet10">
-      <package pattern="*" />
-    </packageSource>
-    <packageSource key="dotnet10-transport">
-      <package pattern="*" />
-    </packageSource>
     <packageSource key="richnav">
       <package pattern="*" />
     </packageSource>
   </packageSourceMapping>
```
RussKie, line 34:
```suggestion
  </packageSourceMapping>
  <disabledPackageSources />
```

RussKie:
Tested in https://github.com/dotnet/extensions-samples/pull/617

PR #5774, opened 2025-01-06
Add Obsolete attribute on IResourceMonitor and its friends

### LINKED ISSUE #5773: Add Obsolete attribute to public API which use IResourceMonitor
Just add the Obsolete attribute without making any behavior changes

evgenyfedorov2:
Resolves #5773 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5774)


amadeuszl:
There's `SourceIpAddresses` in `ResourceMonitoringOptions.Windows` that I believe should be Obsolete as well

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
amadeuszl, line 1: Should `AddResourceMonitoring()` overload also be marked as obsolete?

amadeuszl:
Could we specify in readme (unless there's some guide somewhere else) what's obsolete and what's new recommended way of the library usage? It may be quite important as at least half of the codebase of the library will be obsolete at this point.

evgenyfedorov2:
> There's `SourceIpAddresses` in `ResourceMonitoringOptions.Windows` that I believe should be Obsolete as well

I did not make it obsolete intentionally because it is used in `WindowsTcpStateInfo`, which is not related to `IResourceMonitor`

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
evgenyfedorov2: But we still need some entry point for the whole ResourceMonitoring library, I think that will be the one.

evgenyfedorov2:
> Could we specify in readme (unless there's some guide somewhere else) what's obsolete and what's new recommended way of the library usage? It may be quite important as at least half of the codebase of the library will be obsolete at this point.

We already provide quick Diagnostics IDs [here](https://github.com/dotnet/extensions/blob/422bd6d3087d6a54df20e910d2ba72f67dcb1d8e/src/Shared/DiagnosticIds/DiagnosticIds.cs#L126). Conceptual documentation of Resource Monitoring is [here](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/diagnostic-resource-monitoring) and of course it will need to be updated too, but it is a different repo, so created an issue [there](https://github.com/dotnet/docs/issues/44176). Is this what you are suggesting or is it something different?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
amadeuszl: If I understand correctly it will be implemented within #5779. It that case it makes sense to leave this API open, so we can just change internal implementation without disrupting end user. 

What's the estimation of work for implementing 5779? Just want to avoid situation when in the next release, user have all APIs marked as obsolete without having an alternative.

evgenyfedorov2: Users start using this library by calling `.AddResourceMonitoring()` method. Then, they either inject `IResourceMonitor` interface and call `.GetUtilization()` on it, or simply use any metric exporter to export metrics to a back-end. So, we are now deprecating the `IResourceMonitor` part, but the metrics will stay as is. We don't need to implement anything new.

Also, this class [src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Network/WindowsNetworkMetrics.cs ](https://github.com/dotnet/extensions/blob/422bd6d3087d6a54df20e910d2ba72f67dcb1d8e/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Network/WindowsNetworkMetrics.cs) has just metrics without `.GetSnapshot()` implementation, and this is what we will be aiming for for `WindowsSnapshotProvider`/`WindowsContainerSnapshotProvider` and `LinuxUtilizationProvider` classes as well, they will eventually have only metrics without any `.GetSnapshot()` kind of stuff.

amadeuszl: All clear now

amadeuszl (approved):
Looks good, no more questions

PR #5781, opened 2025-01-08

mgravell:
(note: this is a re-do of #5748 which went "a bit wrong" when rebasing)

Outstanding:

- [ ] additional logging
  - [ ] add log when L2 data is rejected for (reason)
  - [ ] add metric of the number of tag expirations being tracked in L1
- [x] resolve netfx test failure (actually: niche timing problem in tag invalidation)
- [ ] address option member inheritance (or not - per @jodydonetti comments here)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5781)

### DIFF
--- a/eng/MSBuild/LegacySupport.props
+++ b/eng/MSBuild/LegacySupport.props
@@ -15,7 +15,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)\..\..\src\LegacySupport\CallerAttributes\*.cs" LinkBase="LegacySupport\CallerAttributes" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(InjectSkipLocalsInitAttributeOnLegacy)' == 'true' AND ('$(TargetFramework)' == 'net462' or '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'netcoreapp3.1')">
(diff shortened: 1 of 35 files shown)


mgravell:
inviting @sebastienros and @jodydonetti from #5748

jodydonetti:
Hi Marc, not strictly related to tagging, but I'm looking at some code change in the PR and I spotted this xml comment regarding `HybridCacheEntryOptions`:

> If options are specified at the individual call level, the non-null values are merged
> (with the per-call options being used in preference to the global options). If no value is
> specified for a given option (globally or per-call), the implementation can choose a reasonable default.

When we touched on this subject some time ago I remember we agreed on the fact that inheritance (from per-call up to the `DefaultEntryOptions`) would have to be done on an all-or-nothing approach, and not on a per-single-option approach.
What I mean is that "if `null` is passed as an `HybridCacheEntryOptions` param" than it would fall back to the `DefaultEntryOptions`, but single options inside that object (like `Expiration`, `LocalCacheExpiration`) should not have inheritance of values.

All of this was because of the impossibility to express "undefined" or, to better say it, the impossibility to differentiate between `null` meaning "I'm saying `null` because I want a local fallback" (for example from `Expiration` to `LocalCacheExpiration`) and `null` meaning "I'm saying `null` because I want a global fallback" (for example to the same option in the `DefaultEntryOptions`").

Can you clarify what is the current rationale here?

I'm asking for 2 reasons: first I'm interested in general, and second because as you know I'm creating a [3rd party implementation](https://github.com/ZiggyCreatures/FusionCache/issues/350), and I'd like te external behaviour to be the same.

Thanks!


ps: will post more about the tagging implementation as soon as I'll finish reading the massive code changes.

mgravell:
@jodydonetti  thanks; added checklist item

Issue #5694, opened 2024-11-25
FakeTimeProvider does not advance when GetTimestamp is called


ItsVeryWindy:
### Description

This is how I expect it to work, not sure if it's a bug, intentional or an oversight.

```csharp
var advanceAmount = TimeSpan.FromSeconds(5);

var provider = new FakeTimeProvider
{
    AutoAdvanceAmount = TimeSpan.FromSeconds(5)
};

/// .. non test code

var startingTimestamp = provider.GetTimestamp();

/// ... things happening in between

var elapsedTime = provider.GetElapsedTime(startingTimestamp);

/// ... test code

// expected to be true but is 0
Assert.That(elapsedTime, Is.EqualTo(advanceAmount))

```

### Reproduction Steps

As above ^^

### Expected behavior

I would expect that when checking the elapsed time that it should match what is specified in AutoAdvanceAmount.

### Actual behavior

The elapsed time is zero.

### Regression?

No idea.

### Known Workarounds

Currently I can work around the issue by advancing the timer myself in between the calls to GetTimestamp and GetElapsedTime.

### Configuration

.NET 8

### Other information

https://github.com/dotnet/extensions/blob/cfed375f3161f2e553e946b4f968b818e8e858f1/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs#L165

This shows that it just returns the number of ticks from now.

Maybe it needs to do something similar to https://github.com/dotnet/extensions/blob/cfed375f3161f2e553e946b4f968b818e8e858f1/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs#L83

amadeuszl:
Thank you for raising this question. I did some checks and current implementation works as expected. There's one important reason why `GetTimestamp()` does not AutoAdvance. `GetElapsedTime()` runs `GetTimestamp()` as well in the implementation in `TimeProvider`, that would cause unintended move in time when just checking the time with `GetElapsedTime()`. Here's implementation for [GetElapsedTime()](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/TimeProvider.cs#L125).

ItsVeryWindy:
Personally I'd consider that an acceptable trade off, I'm not even sure I would call it a trade off...

If I did this, would I normally expect the same value twice?

```csharp
var elapsedTime = provider.GetElapsedTime(startingTimestamp);
[...]

PR #5783, opened 2025-01-09
Respect AutoAdvance for GetTimestamp in FakeTimeProvider

amadeuszl:
Solves #5694
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5783)


src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
```diff
@@ -175,7 +175,17 @@ public override long GetTimestamp()
...
+        DateTimeOffset result;
+
+        lock (Waiters)
+        {
+            result = _now;
+            _now += _autoAdvanceAmount;
+        }
+
+        WakeWaiters();
+        return result.Ticks;
```
evgenyfedorov2, line 188: Can something like this work:

```suggestion

        return GetUtcNow().Ticks;
```

amadeuszl: We can, changed

evgenyfedorov2 (approved):

PR #5785, opened 2025-01-09

mgravell:
(note: this is a re-do of #5748 / #5781 which went "a bit wrong" when rebasing)

Outstanding:

- [x] additional logging
  - [x] add log when L2 data is rejected for (reason)
  - [x] add metric of the number of tag expirations being tracked in L1
- [x] resolve netfx test failure (actually: niche timing problem in tag invalidation)
- [x] address option member inheritance (or not - per @jodydonetti comments here)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5785)


jodydonetti:
> [!NOTE]  
> Reposting my comment here from [there](https://github.com/dotnet/extensions/pull/5781) for ease of discussion.

Hi Marc, not strictly related to tagging, but I'm looking at some code change in the PR and I spotted this xml comment regarding `HybridCacheEntryOptions`:

> If options are specified at the individual call level, the non-null values are merged
> (with the per-call options being used in preference to the global options). If no value is
> specified for a given option (globally or per-call), the implementation can choose a reasonable default.

When we touched on this subject some time ago I remember we agreed on the fact that inheritance (from per-call up to the `DefaultEntryOptions`) would have to be done on an all-or-nothing approach, and not on a per-single-option approach.
What I mean is that "if `null` is passed as an `HybridCacheEntryOptions` param" than it would fall back to the `DefaultEntryOptions`, but single options inside that object (like `Expiration`, `LocalCacheExpiration`) should not have inheritance of values.

All of this was because of the impossibility to express "undefined" or, to better say it, the impossibility to differentiate between `null` meaning "I'm saying `null` because I want a local fallback" (for example from `LocalCacheExpiration` to `Expiration`) and `null` meaning "I'm saying `null` because I want a global fallback" (for example to the same option in the `DefaultEntryOptions`").
A practical example of a problem this may create: in the `DefaultEntryOptions` I set `Expiration` to `5min` and `LocalCacheExpiration` to `10s` to refresh L1 more frequently. Then in a certain call I pass a specific `HybridCacheEntryOptions` instance where I set `Expiration` to `1h` and leave `LocalCacheExpiration` to `null`, thinking it will be the same as `Expiration`, when in reality it will become `10s`. Because of this, there will be for me no way to keep the 2 options in-sync at every call, unless I keep specifying the same value for both `Expiration` and `LocalCacheExpiration` at every call.

Can you clarify what is the current rationale here?

I'm asking for 2 reasons: first I'm interested in general, and second because as you know I'm creating a [3rd party implementation](https://github.com/ZiggyCreatures/FusionCache/issues/350), and I'd like te observable external behaviour to be the same for both implementations.

Thanks!

ps: if you like I can open a different issue for this, just let me know.
pps: will post more about the tagging implementation as soon as I'll finish reading the massive code changes.

jodydonetti:
An extra point to think about, again related the entry options inheritance thing, is how [this method](https://github.com/dotnet/runtime/blob/289aa17cac46fc4c2fb3e5889030f5379ca8e92a/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/Hybrid/HybridCacheEntryOptions.cs#L35) (marked internal, and maybe it will not be used in the end) would be affected.

BrennanConroy:
> to allow persistence of invalidation between sessions when a L2 (backend cache) is in play, we additionally store the timeout data via additional simple values in the cache

Shouldn't we be concerned about machine clocks not being synchronized? i.e. Machine 1 says it's 10:05:03 and adds a timeout for 10:10:03, but Machine 2 says it's 10:12:05 and when it reads the cache item it will determine it to be invalid.

jodydonetti:
> Shouldn't we be concerned about machine clocks not being synchronized? i.e. Machine 1 says it's 10:05:03 and adds a timeout for 10:10:03, but Machine 2 says it's 10:12:05 and when it reads the cache item it will determine it to be invalid.

Eh, welcome to the world of clock shifting/drifting/skewing & friends: it's not something that's solvable via software, at least not at this level (to the best of my knowledge).

Just to make an example: typically Microsoft/Meta/Google/AWS have their own atomic clocks to avoid this problem (as much as possible), dedicated to specific services (like Google Spanner).
Some reads:
- https://en.wikipedia.org/wiki/Clock_synchronization
- https://en.wikipedia.org/wiki/Clock_drift
- https://www.theverge.com/2012/11/26/3692392/google-spanner-atomic-clocks-GPS
- https://engineering.fb.com/2021/08/11/open-source/time-appliance/
- https://cloud.google.com/spanner/docs/true-time-external-consistency
- https://www.bigdatawire.com/2024/12/03/aws-takes-on-google-spanner-with-atomic-clock-powered-distributed-dbs/
- https://people.mpi-inf.mpg.de/~clenzen/pubs/LLSW10clock.pdf

mgravell:
Clock skew: ultimately as stated, not easily solvable. But in the context
of cache servers: we don't need super complexity or accuracy: IMO if any
layer is unhappy re expiration, that's good enough to consider it invalid.
As a side-effect, it also makes the logic time-testable.

jodydonetti:
Hi Marc, I'm looking at [this](https://github.com/dotnet/extensions/commit/f78e41d1607c0dc76094f4709cd1e535b9a29ca4#diff-c9af9be99941920f5835bb198a67dd56204dc4165fd49f7ea8a96ddedcbcdd84) and wow, those are a lot of tests 💪

Just to doublecheck so we are on the same page: the idea is still [this](https://github.com/dotnet/extensions/pull/5785#issuecomment-2580928409)?

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Extensions.Caching.Hybrid.Internal;
+
+/// <summary>
+/// Represents zero (null), one (string) or more (string[]) tags, avoiding the additional array overhead when necessary.
+/// </summary>
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1066:Implement IEquatable when overriding Object.Equals", Justification = "Equals throws by intent")]
+internal readonly struct TagSet
```
sebastienros, line 16: Since there is a nice ToString implementation

```suggestion
[System.Diagnostics.DebuggerDisplay("{ToString()}")]
internal readonly struct TagSet
```

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
```diff
@@ -0,0 +1,235 @@
...
+        {
+            switch (collection.Count)
+            {
+                case 0:
+                    return Empty;
+                case 1 when collection is IList<string> list:
+                    return new TagSet(list[0]);
+                case 1:
+                    // avoid the GetEnumerator() alloc
+                    var arr = ArrayPool<string>.Shared.Rent(1);
```
sebastienros, line 109: For my own knowledge, your opinion on this vs. a ThreadStatic string[1]? (it's even possible that ArrayPool does that internally)

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
```diff
@@ -0,0 +1,235 @@
...
+        string tag when index == 0 => tag,
+        string[] tags => tags[index],
+        _ => throw new IndexOutOfRangeException(nameof(index)),
+    };
+
+    public void CopyTo(Span<string> target)
+    {
+        switch (_tagOrTags)
+        {
+            case string tag:
```
sebastienros, line 82: no `null` case? (TagSet.Empty)

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
```diff
@@ -0,0 +1,235 @@
...
+        CopyTo(arr);
+        return arr;
+    }
+
+    internal const string WildcardTag = "*";
+
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("StyleCop.CSharp.ReadabilityRules", "SA1122:Use string.Empty for empty strings", Justification = "Not needed")]
+    internal bool TryFind(ReadOnlySpan<char> span, [NotNullWhen(true)] out string? tag)
+    {
+        switch (_tagOrTags)
```
sebastienros, line 176: No `null` or `default` case?

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
mgravell: there would be nothing to do - it would be `case null: break;`

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
mgravell: Given that there's no chance of re-entrancy, that would be a very valid implementation, and probably a lot less mess than arraypool - which IIRC is already going to be doing some thread-bucketing; I'll go ahead and switch over to that - see a5b816eed83d505af43e89fc3393c4ca0a54d109

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/TagSet.cs
mgravell: not needed - it would be `tag = null; return false;` which is what already happens *after* the `switch` (sharing the "nope" logic from all ways of saying "nope")

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.cs
```diff
@@ -20,8 +21,11 @@ namespace Microsoft.Extensions.Caching.Hybrid.Internal;
 /// <summary>
 /// The inbuilt implementation of <see cref="HybridCache"/>, as registered via <see cref="HybridCacheServiceExtensions.AddHybridCache(IServiceCollection)"/>.
 /// </summary>
+[SkipLocalsInit]
```
DeagleGross, line 24: why `SkipLocalsInit` exists on specific types only? Not on module?

DeagleGross (approved):
_Approved_; but **would be happy if someone else takes a look** (PR is long living). Below asked a couple of specific code questions; overall purpose of PR is kind of understandable, but I dont have any background with HybridCache

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.cs
mgravell: I'd be fine with either, but I don't really *need* it globally; least surprises, etc

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
```diff
@@ -65,35 +71,114 @@ static async Task<BufferChunk> AwaitedLegacyAsync(Task<byte[]?> pending, Default
...
+    internal ValueTask InvalidateL2TagAsync(string tag, long timestamp, CancellationToken token)
+    {
+        if (!HasBackendCache)
+        {
+            return default; // no L2
+        }
+
+        byte[] oversized = ArrayPool<byte>.Shared.Rent(sizeof(long));
+        BinaryPrimitives.WriteInt64LittleEndian(oversized, timestamp);
+        var pending = SetDirectL2Async(TagKeyPrefix + tag, new BufferChunk(oversized, 0, sizeof(long), false), _tagInvalidationEntryOptions, token);
```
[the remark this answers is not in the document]
mgravell: yes, this is to avoid the state machine overhead in the happy path; that *isn't terribad*, but: it is sometimes desirable to hand-crank it

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
```diff
@@ -128,6 +213,20 @@ internal void SetL1<T>(string key, CacheItem<T> value, HybridCacheEntryOptions?
...
+    private async ValueTask WritePayloadAsync(string key, CacheItem cacheItem, BufferChunk payload, HybridCacheEntryOptions? options, CancellationToken token)
+    {
+        // bundle a serialized payload inside the wrapper used at the DC layer
+        var maxLength = HybridCachePayload.GetMaxBytes(key, cacheItem.Tags, payload.Length);
+        var oversized = ArrayPool<byte>.Shared.Rent(maxLength);
+
+        var length = HybridCachePayload.Write(oversized, key, cacheItem.CreationTimestamp, GetL2AbsoluteExpirationRelativeToNow(options),
+            HybridCachePayload.PayloadFlags.None, cacheItem.Tags, payload.AsSequence());
+
+        await SetDirectL2Async(key, new(oversized, 0, length, true), GetL2DistributedCacheOptions(options), token).ConfigureAwait(false);
```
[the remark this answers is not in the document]
mgravell: for environments that have a sync-context, which usually *means* lower FXs; however, as library code: we don't know about the sync-context state, so: best to be explicit

I *believe* the plan to allow a global setting for this is still being discussed; I will check whether that finally shipped!

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/BufferChunk.cs
```diff
@@ -58,7 +62,7 @@ public byte[] ToArray()
         }
 
         var copy = new byte[length];
-        Buffer.BlockCopy(Array!, 0, copy, 0, length);
+        Buffer.BlockCopy(OversizedArray!, Offset, copy, 0, length);
```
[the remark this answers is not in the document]
mgravell: nice; I will integrate that next cycle

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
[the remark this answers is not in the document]
mgravell: if this were "aspnetcore code, only targeting netcurrent", then yes: we could probably get away without specifying this; however, HC a: can be used in non-aspnetcore contexts (WinForms, etc), and b: supports netstandard, netfx, etc

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
```diff
@@ -128,6 +213,20 @@ internal void SetL1<T>(string key, CacheItem<T> value, HybridCacheEntryOptions?
         }
     }
 
+    private async ValueTask WritePayloadAsync(string key, CacheItem cacheItem, BufferChunk payload, HybridCacheEntryOptions? options, CancellationToken token)
+    {
+        // bundle a serialized payload inside the wrapper used at the DC layer
+        var maxLength = HybridCachePayload.GetMaxBytes(key, cacheItem.Tags, payload.Length);
```
RussKie, line 219: We follow runtime's style - i.e., no `var` unless the type is obvious.
https://github.com/dotnet/extensions/blob/main/CONTRIBUTING.md#dos-and-donts

Applies universally.

RussKie (approved):

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultHybridCache.L2.cs
mgravell: @RussKie I will do a `var` sweep in a follow up PR - should be a far less daunting one!

RussKie: Sure thing. Thank you

jodydonetti:
Congrats for the big milestone 🎉

mgravell:
@jodydonetti  *if* I made the "snap" last night, this *should* go out ~11th as a final month of `[Experimental]`, then it'll be GA the following month; I have some docs improvements from internal review, which apply to both HC and FC (things like guidance on appropriate key construction), but: eesh, that was indeed a slog.

jodydonetti:
> @jodydonetti _if_ I made the "snap" last night, this _should_ go out ~11th as a final month of `[Experimental]`,

every time I see an `[Experimental]` usage I get a dumb smile thinking about the dense fella who [proposed](https://github.com/dotnet/runtime/issues/31542) it 🤣

> then it'll be GA the following month; I have some docs improvements from internal review,

Eh, enjoy every second of stress test in the public preview, it's really worth it. I got random minor edge case stuff came up with FC (and something is still coming up) so that's really precious, although frustrating sometimes 🥲

> which apply to both HC and FC (things like guidance on appropriate key construction)

Eagerly awaiting this part to see all coming together on both HC/FC, can't wait.

> but: eesh, that was indeed a slog.

Amazing job 💪

PR #5786, opened 2025-01-09
Add `FunctionInvokingChatClient.CurrentContext`

### LINKED ISSUE #5705: There doesn't seem to be a way to set the context.Terminate in a AIFunction call
From what I see in the source the context is created internally and by default it will continue the calls:

- https://github.com/dotnet/extensions/blob/d8e2cd29165360b2eeb30f6305ccdf09d3509511/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L529

Perhaps it would be better if the return type of the `AIFunction` is `void` to set the `FunctionInvocationContext.Terminate` prop to `true` by default. Or somehow allow users to set it manually when needed.

MackinnonBuck:
Allows AI functions to access the current `FunctionInvokingChatClient.FunctionInvocationContext`.

This can be used to access the chat history or terminate the function invocation loop.

Fixes #5705
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5786)


test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/FunctionInvokingChatClientTests.cs
```diff
@@ -546,6 +546,61 @@ public async Task SupportsConsecutiveStreamingUpdatesWithFunctionCalls()
...
 
+    [Fact]
+    public async Task CanTerminateInvocationLoopFromFunction()
+    {
+        var options = new ChatOptions
+        {
+            Tools =
+            [
+                AIFunctionFactory.Create((int i) =>
+                {
```
stephentoub, line 557: Could we also assert that other state on CurrentContext is as we expect it?

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/FunctionInvokingChatClientTests.cs
```diff
@@ -546,6 +546,61 @@ public async Task SupportsConsecutiveStreamingUpdatesWithFunctionCalls()
         Assert.Equal("OK bye", singleUpdateContent.Text);
     }
 
+    [Fact]
```
stephentoub, line 549: It'd be good to have a few more tests validating:
- It's null when we expect it to be null.
- It does indeed propagate across async hops.

stephentoub (approved):
Thanks!

stephentoub (approved):

PR #5788, opened 2025-01-10
Fix schema generation for floating point types

### LINKED ISSUE #5787: Structured output schema generation is too loose for decimal/double/float
If you have a type like:

```cs
public record Product(int Name, decimal Price);
```

... and make a call like:

```cs
var response = await chatClient.CompleteAsync<Product>(
    "Extract info about the following: we sell eggs for fifty dollars");
if (response.TryGetResult(out var product))
{
    Console.WriteLine($"{product.Name} costs {product.Price:c}");
}
```

... you would expect it to work quite reliably. But it does not - often the result JSON can't be parsed.

### The cause

On debugging, I found this is because the JSON schema we generate contains this:

```json
"price": {
    "type": ["string", "number" ]
}
```

The LLM sees the invitation to supply a `string`, and often returns JSON like:

```json
{ "Name": "Eggs", "Price": "$50" }
```

or sometimes

```json
{ "Name": "eggs", "Price": "fifty dollars" }
```

... neither of which can be JSON-parsed as a `Product`.

### The solution

By default, for `decimal`/`double`/`float`, the schema should not invite the LLM to respond with a `string`, because suggesting it can return an arbitrary string will allow it to put in literally anything. It has no reason even to favour values that are numeric-like.

@eiriktsarpalis Should we be passing some other option to the JSON schema generator to advise it not to use `string` for numeric types?

eiriktsarpalis:
Fix #5787.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5788)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
@@ -234,7 +234,7 @@ public static class AIJsonUtilitiesTests
             """).RootElement;
 
         JsonSerializerOptions options = new(JsonSerializerOptions.Default) { NumberHandling = JsonNumberHandling.AllowReadingFromString };
-        AIFunction func = AIFunctionFactory.Create((int a, int? b, long c, short d) => { }, serializerOptions: options);
+        AIFunction func = AIFunctionFactory.Create((int a, int? b, long c, short d, float e, double f, decimal g) => { }, serializerOptions: options);
 
         AIFunctionMetadata metadata = func.Metadata;
         foreach (var param in metadata.Parameters)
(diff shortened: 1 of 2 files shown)


test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
```diff
@@ -234,7 +234,7 @@ public static void CreateParameterJsonSchema_TreatsIntegralTypesAsInteger_EvenWi
             """).RootElement;
 
         JsonSerializerOptions options = new(JsonSerializerOptions.Default) { NumberHandling = JsonNumberHandling.AllowReadingFromString };
-        AIFunction func = AIFunctionFactory.Create((int a, int? b, long c, short d) => { }, serializerOptions: options);
+        AIFunction func = AIFunctionFactory.Create((int a, int? b, long c, short d, float e, double f, decimal g) => { }, serializerOptions: options);
```
stephentoub, line 237: What will happen if a type like Half or BigInteger is used?

eiriktsarpalis: `Half` works the same as `float` and `double` for the targets that support it. `BigInteger` currently has no built-in support in STJ, its schema is an object containing all its public properties.

SteveSandersonMS (approved):
Nice, thanks!

PR #5795, opened 2025-01-14
Ensure Ollama streaming updates specify a CompletionId.

eiriktsarpalis:
Discovered while working on https://github.com/eiriktsarpalis/blackbeard-extension-cs/pull/1. Non-streaming completions use the `CreatedAt` timestamp as the completion id, however this is being omitted for the case of streaming completions.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5795)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
@@ -142,6 +142,7 @@ public sealed class OllamaChatClient : IChatClient
 
             StreamingChatCompletionUpdate update = new()
             {
+                CompletionId = chunk.CreatedAt,
                 Role = chunk.Message?.Role is not null ? new ChatRole(chunk.Message.Role) : null,
                 CreatedAt = DateTimeOffset.TryParse(chunk.CreatedAt, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset createdAt) ? createdAt : null,
                 FinishReason = ToFinishReason(chunk),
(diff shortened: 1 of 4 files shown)


src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -142,6 +142,7 @@ public async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteStreamingAs
 
             StreamingChatCompletionUpdate update = new()
             {
+                CompletionId = chunk.CreatedAt,
```
stephentoub, line 145: We have other components that create updates. Do we want to manufacture id's for all those?

And if an OpenAI compatible client requires id's, shouldn't the serialization helpers add id's when they don't exist?

eiriktsarpalis: Done. I'm keeping the existing change as well since like I said it's the only place where we don't populate the field.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -584,6 +585,8 @@ private static FunctionCallContent ParseCallContentFromBinaryData(BinaryData ut8
     private static T? GetValueOrDefault<T>(this AdditionalPropertiesDictionary? dict, string key) =>
         dict?.TryGetValue(key, out T? value) is true ? value : default;
 
+    private static string CreateCompletionId() => $"chatcmpl-{Guid.NewGuid().ToString("N", CultureInfo.InvariantCulture)}";
```
stephentoub, line 588: The culture isn't necessary here, fwiw. Guid.ToString/TryFormat completely ignores the format provider.

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
eiriktsarpalis: An analyzer complained. Seemed preferable to suppressing yet another rule.

PR #5798, opened 2025-01-16
Make HealthChecks.ResourceUtilization use observable instruments

### LINKED ISSUE #5779: Make HealthChecks.ResourceUtilization use observable instruments
Make HealthChecks.ResourceUtilization use observable instruments, e.g. metrics from ResourceMonitoring instead of IResourceMonitor.GetUtilization obsolete API

evgenyfedorov2:
Fixes #5779 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5798)


src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
```diff
@@ -91,4 +135,43 @@ public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, Canc
...
+        }
+    }
+
+    private void OnMeasurementRecorded(
+        Instrument instrument, double measurement,
+        ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        switch (instrument.Name)
+        {
+            case "process.cpu.utilization":
```
amadeuszl, line 167: It doesn't have to be done withing this PR, but we should bring meter names to Shared project so we reduce chance of messing up the names across different projects. WDYT?

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
```diff
@@ -13,21 +15,53 @@ namespace Microsoft.Extensions.Diagnostics.HealthChecks;
 /// <summary>
 /// Represents a health check for in-container resources <see cref="IHealthCheck"/>.
 /// </summary>
-internal sealed class ResourceUtilizationHealthCheck : IHealthCheck
+internal sealed class ResourceUtilizationHealthCheck : IHealthCheck, IDisposable
```
amadeuszl, line 18: Could we have different separate implementation of `IHealthCheck` with `IMeterListener`? With `UseObservableResourceMonitoringInstruments` deciding which to use? Is would make it easier to deprecate old path and also gives more flexibility if we would like to extend any of these implementations.

The only common part is the description string builder, which can be stored wherever.

evgenyfedorov2: Ok, looks like we can't do exactly that, because AddCheck<T>() needs to know type T in advance, while we will know it only later when `ServiceProvider` is built where we would check the resulting value of the `UseObservableResourceMonitoringInstruments `flag and register the right type accordingly.
Nevertheless, I split it into partial classes, I think this is the best we can do.

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
evgenyfedorov2: Created the issue https://github.com/dotnet/extensions/issues/5813

src/Libraries/Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization/ResourceUtilizationHealthCheck.cs
amadeuszl: Looks good

amadeuszl (approved):

PR #5799, opened 2025-01-16
Make a number of improvements to the OpenAI serialization helpers.

eiriktsarpalis:
1. Makes a number of fixes in the OpenAI mapping logic.
2. Adds a custom converter for the `OpenAIChatCompletionRequest` envelope type that uses OpenAI wire format. This should make it possible to bind to that type directly in the context of MVC apps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5799)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatCompletionRequest.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatCompletionRequest.cs
@@ -1,13 +1,21 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.Generic;
+using System.ComponentModel;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using OpenAI.Chat;
+
+#pragma warning disable CA1034 // Nested types should not be visible
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>
 /// Represents an OpenAI chat completion request deserialized as Microsoft.Extension.AI models.
 /// </summary>
+[JsonConverter(typeof(Converter))]
 public sealed class OpenAIChatCompletionRequest
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatCompletionRequest.cs
```diff
@@ -29,4 +37,33 @@ public sealed class OpenAIChatCompletionRequest
     /// Gets the model id requested by the chat completion.
     /// </summary>
     public string? ModelId { get; init; }
+
+    /// <summary>
+    /// Converts an OpenAIChatCompletionRequest object to and from JSON.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class Converter : JsonConverter<OpenAIChatCompletionRequest>
```
eiriktsarpalis, line 45: Needs to be public so that it can be picked up by source generators.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatCompletionRequest.cs
```diff
@@ -29,4 +37,33 @@ public sealed class OpenAIChatCompletionRequest
...
+        }
+
+        /// <summary>
+        /// Writes the specified value as JSON.
+        /// </summary>
+        /// <param name="writer">The writer.</param>
+        /// <param name="value">The value to write.</param>
+        /// <param name="options">The serializer options.</param>
+        public override void Write(Utf8JsonWriter writer, OpenAIChatCompletionRequest value, JsonSerializerOptions options) =>
+            throw new NotSupportedException("Request body serialization is not supported.");
```
eiriktsarpalis, line 67: Populating the corresponding OpenAI DTO requires access to internal APIs. Rather than using reflection I chose to make this an unsupported operation for now.

stephentoub: What APIs?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -148,7 +149,12 @@ public static ChatOptions FromOpenAIOptions(OpenAI.Chat.ChatCompletionOptions? o
 
         if (options is not null)
         {
-            result.ModelId = _getModelIdAccessor.Invoke(options, null)?.ToString();
+            result.ModelId = _getModelIdAccessor.Invoke(options, null)?.ToString() switch
+            {
+                null or "" => null,
```
stephentoub, line 154: Is this sufficient, or is there any situation where there might be non-empty whitespace which should also be null?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
eiriktsarpalis: I thought about it, but in absence of evidence that it occurs I figured I should avoid overcompensating.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatCompletionRequest.cs
eiriktsarpalis: The OpenAI wire format is being produced first by mapping MEAI types to the OpenAI library DTOs and then serializing the latter. However, many of the properties of the OpenAI DTOs are currently marked internal, so we can't actually populate it without using reflection.

stephentoub: Previously I'd suggested just going directly to writing out JSON rather than going through mapping to OpenAI's types.

PR #5801, opened 2025-01-17
API that allows to remove all resilience handlers from the HTTP client

### LINKED ISSUE #5695: [API Proposal, Microsoft.Extensions.Http.Resilience]: API for removing and replacing the standard resilience/hedging handlers
## Background and motivation

.NET provides the [`ConfigureHttpClientDefaults`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientfactoryservicecollectionextensions.configurehttpclientdefaults?view=net-8.0) method that allows you to configure the default behavior of the `HttpClient`. With the help of this method you can register the [`StandardResiliencePipeline`](https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience#resilience) as part of the default configuration of the `HttpClient`:

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());
```

This is convenient, since you don't need to configure the resilience pipeline for each `HttpClient`. But with the API we currently provide users face a few issues/challenges when they need to change the default configuration. Below are a couple of known issues/challanges.

### Issue 1: Removing the existing default resilience pipeline and adding a custom one

For example, a user registers the `StandardResilienceHandler` as default configuration, but at the same time he/she wants to use the `StandardHedgingHandler` for a particular named `HttpClient`:

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to remove the StandardResilienceHandler and add instead the StandardHedgingHandler.
services.AddHttpClient("custom")./*Remove StandardResilienceHandler*/.AddStandardHedgingHandler();
```

Currently, there is no API that allows users to remove the `StandardResilienceHandler` and register a new one. Now users can work that around as described here https://github.com/dotnet/extensions/issues/4814#issuecomment-2374345866.

### Issue 2: Override configuration of the default resilience pipeline for a particular `HttpClient`

A user registers the `StandardResilienceHandler` as default configuration, and wants to provide custom configuration for a particular `HttpClient`:

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to customize settings of the StandardResilienceHandler.
services.AddHttpClient("custom")./*Configure default StandardResilienceHandler*/;
```

Currently, there is no API to override the default configuration, see https://github.com/dotnet/extensions/issues/4814.

### Issue 3: Replace the default resilience pipeline with a custom one

In oppose to the "Issue 1" a user wants to replace the default resilience pipeline with a custom one. He/she wants to replace it, because it might be important to keep the order where the default resilience pipeline was registered, for example, it could be important for metering purposes.

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to replace the StandardResilienceHandler with a custom one.
services.AddHttpClient("custom")./*We want to replace default StandardResilienceHandler, i.e. the new pipeline will be in the same order as the default one*/;
```

Supporting such scenario could be useful to fix issues like the following https://github.com/dotnet/extensions/issues/5021. Because the default resilience pipeline and its state are shared by all `HttpClient` instances it could lead to issues, like the one mentioned in the previous sentence. And having an ability to replace the default pipeline with a custom one for a particular `HttpClient` could be a solution to that issue.

#### Remove all resilience handlers:

```diff
namespace Microsoft.Extensions.DependencyInjection;

public static class ResilienceHttpClientBuilderExtensions
{
+   /// <summary>
+   /// Removes all resilience handlers registered earlier.
+   /// </summary>
+   /// <param name="builder">The HTTP client builder.</param>
+   /// <returns>The value of <paramref name="builder"/>.</returns>
+   public static IHttpClientBuilder RemoveAllResilienceHandlers(this IHttpClientBuilder builder);
}
```

#### Add or replace the `StandardResilienceHandler`:

```diff
namespace Microsoft.Extensions.DependencyInjection;

public static class ResilienceHttpClientBuilderExtensions
{
+    /// <summary>
+    /// If there is no standard Resilience or Hedging handlers registered,
+    /// then the method adds the standard Resilience handler to the HttpClient request pipeline.
+    /// Otherwise, the existing standard Resilience or Hedging handler is replaced with the standard Resilience handler.
+    /// </summary>
+    /// <param name="builder">The HTTP client builder.</param>
+    /// <returns>A <see cref="IHttpStandardResiliencePipelineBuilder"/> instance that can be used to configure the standard resilience handler behavior.</returns>
+    public static IHttpStandardResiliencePipelineBuilder AddOrReplaceStandardResilienceHandler(
+        this IHttpClientBuilder builder);

+    /// <summary>
+    /// If there is no standard Resilience or Hedging handlers registered,
+    /// then the method adds the standard Resilience handler to the HttpClient request pipeline.
+    /// Otherwise, the existing standard Resilience or Hedging handler is replaced with the standard Resilience handler.
+    /// </summary>
+    /// <param name="builder">The HTTP client builder.</param>
+    /// <param name="section">The section that the options will bind against.</param>
+    /// <returns>A <see cref="IHttpStandardResiliencePipelineBuilder"/> instance that can be used to configure the standard resilience handler behavior.</returns>
+    public static IHttpStandardResiliencePipelineBuilder AddOrReplaceStandardResilienceHandler(
+        this IHttpClientBuilder builder,
+        IConfigurationSection section);

+    /// <summary>
+    /// If there is no standard Resilience or Hedging handlers registered,
+    /// then the method adds the standard Resilience handler to the HttpClient request pipeline.
+    /// Otherwise, the existing standard Resilience or Hedging handler is replaced with the standard Resilience handler.
+    /// </summary>
+    /// <param name="builder">The HTTP client builder.</param>
+    /// <param name="configure">The callback that configures the options.</param>
+    /// <returns>A <see cref="IHttpStandardResiliencePipelineBuilder"/> instance that can be used to configure the standard resilience handler behavior.</returns>
+    public static IHttpStandardResiliencePipelineBuilder AddOrReplaceStandardResilienceHandler(
+        this IHttpClientBuilder builder,
+        Action<HttpStandardResilienceOptions> configure);
}
```

#### Add or replace the `StandardHedgingHandler`:

```diff
namespace Microsoft.Extensions.DependencyInjection;

public static partial class ResilienceHttpClientBuilderExtensions
{
+    /// <summary>
+    /// If there is no standard Resilience or Hedging handlers registered,
+    /// then the method adds the standard Hedging handler to the HttpClient request pipeline.
+    /// Otherwise, the existing standard Resilience or Hedging handler is replaced with the standard Hedging handler.
+    /// </summary>
+    /// <param name="builder">The HTTP client builder.</param>
+    /// <returns>A <see cref="IStandardHedgingHandlerBuilder"/> instance that can be used to configure the standard hedging behavior.</returns>
+    public static IStandardHedgingHandlerBuilder AddOrReplaceStandardHedgingHandler(
+        this IHttpClientBuilder builder);

+    /// <summary>
+    /// If there is no standard Resilience or Hedging handlers registered,
+    /// then the method adds the standard Hedging handler to the HttpClient request pipeline.
+    /// Otherwise, the existing standard Resilience or Hedging handler is replaced with the standard Hedging handler.
+    /// </summary>
+    /// <param name="builder">The HTTP client builder.</param>
+    /// <param name="configure">Configures the routing strategy associated with this handler.</param>
+    /// <returns>A <see cref="IStandardHedgingHandlerBuilder"/> instance that can be used to configure the standard hedging behavior.</returns>
+    public static IStandardHedgingHandlerBuilder AddOrReplaceStandardHedgingHandler(
+        this IHttpClientBuilder builder,
+        Action<IRoutingStrategyBuilder> configure);
}
```

## API Usage

With the help of the proposed API the issues described in the beginning could be fixed as following.

### Issue 1: Removing the existing default resilience pipeline and adding a custom one

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to remove the StandardResilienceHandler and add instead the StandardHedgingHandler.
services.AddHttpClient("custom")
    .RemoveAllResilienceHandlers()
    .AddStandardHedgingHandler();
```

### Issue 2: Override configuration of the default resilience pipeline for a particular `HttpClient`

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to customize settings of the StandardResilienceHandler.
services.AddHttpClient("custom")
    .AddOrReplaceStandardResilienceHandler(options =>
    {
        // Configure options here.
    });
```

### Issue 3: Replace the default resilience pipeline with a custom one

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to replace the StandardResilienceHandler with a custom one. There are 2 cases:

// Case 1: we replace the original StandardResilienceHandler with a new StandardResilienceHandler with custom configuration options.
// This case is similar to Issue 2.
services.AddHttpClient("custom")
    .AddOrReplaceStandardResilienceHandler(options =>
    {
        // Configure options here.
    });

// Case 2: we replace the original StandardResilienceHandler with the StandardHedgingHandler.
services.AddHttpClient("custom").AddOrReplaceStandardHedgingHandler();
```

## Alternative Designs

The API design above considers HttpClient's delegating handlers as a list (the order of the items is important), therefore it introduces the `AddOrReplace` methods that implicitly "say" that the standard resilience handler (which is a delegating handler) will be replaced if any. Otherwise, it will be added to the end of the list.

**_The alternative design suggests not to introduce `AddOrReplace` methods, but instead to implement `AddOrReplace` semantics in the existing `Add` methods. The justification behind that is: having more than one standard resilience handlers in the HttpClient is INCORRECT._**

Since it is not correct to have more than one standard resilience handlers in the HttpClient, we consider reasonable to implement `AddOrReplace` semantics in the existing `Add` methods:

* if there is already a standard resilience handler in the HttpClient, then we replace it and reconfigure
* otherwise, we add a new handler to the HttpClient

With such an approach we'll need to introduce the only method `RemoveAllResilienceHandlers` mentioned above, and then the issues described above could be fixed as following:

### Issue 1: Removing the existing default resilience pipeline and adding a custom one

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to remove the StandardResilienceHandler and add instead the StandardHedgingHandler.
services.AddHttpClient("custom")
    .RemoveAllResilienceHandlers()
    .AddStandardHedgingHandler();
```

### Issue 2: Override configuration of the default resilience pipeline for a particular `HttpClient`

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to customize settings of the StandardResilienceHandler.
services.AddHttpClient("custom")
    .AddStandardResilienceHandler(options =>
    {
        // Configure options here.
    });
```

### Issue 3: Replace the default resilience pipeline with a custom one

```csharp
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to replace the StandardResilienceHandler with a custom one. There are 2 cases:

// Case 1: we replace the original StandardResilienceHandler with a new StandardResilienceHandler with custom configuration options.
// This case is similar to Issue 2.
services.AddHttpClient("custom")
    .AddStandardResilienceHandler(options =>
    {
        // Configure options here.
    });

// Case 2: we replace the original StandardResilienceHandler with the StandardHedgingHandler.
services.AddHttpClient("custom").AddStandardHedgingHandler();
```

**Risks.** It could be confusing for customers that `Add` methods has `AddOrReplace` semantics. We'll need to properly document that.

## Risks

There are a few risks:

#### 1. Names of the methods `AddOrReplace...` are not perfect

Names `AddOrReplaceStandardResilienceHandler` and `AddOrReplaceStandardHedgingHandler` are not perfect. The suffixes `ReplaceStandardResilienceHandler` and `ReplaceStandardHedgingHandler` might bring confusion that the methods replace only the corresponding standard resilience handler type, i.e. either the Resilience handler or the Hedging handler. In fact each method replaces any of the standard handlers. So the names of the methods don't fully describe what the methods do. The name that will fully describe the methods would be `AddStandardResilienceHandler_Or_ReplaceAnyStandardHandlerWithStandardResilienceHandler`, but it is too long, therefore we have to find some balance.

#### 2. We'll need to introduce a breaking change to the behavior of the existing methods registering standard Resilience and Hedging handlers

In order to introduce `AddOrReplace` methods we'll need to ensure that HttpClient request pipeline has only one Resilience or Hedging handler. For that we'll need to change the behavior of the existing methods registering Resilience and Hedging handlers in HttpClient request pipeline. In particular, the new behavior will throw an exception when a user attempts to register one of those handlers twice. Currently, we don't throw an exception but register the handler twice which is also not correct. Therefore this risk could be not so high/important.

#### 3. Additional effort to maintaining new `AddOrReplace` methods

1. With the introduction of new `AddOrReplace` methods we'll have to maintain them, meaning that adding some capability to existing `AddStandardResilienceHandler` methods would result in mirroring it to the `AddOrReplace` methods.

2. Currently we have 2 standard handlers: Resilience and Hedging. Probably, introducing a new standard handler would lead to creating similar `AddOrReplace` methods to have feature parity.

rainsxng:
**Background and motivation**

.NET provides the [ConfigureHttpClientDefaults](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientfactoryservicecollectionextensions.configurehttpclientdefaults?view=net-8.0) method that allows you to configure the default behavior of the HttpClient. With the help of this method you can register the [StandardResiliencePipeline](https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience#resilience) as part of the default configuration of the HttpClient:

`services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());`

This is convenient, since you don't need to configure the resilience pipeline for each HttpClient. But with the API we currently provide users face a few issues/challenges when they need to change the default configuration. This change focuses on the following issue:

Removing the existing default resilience pipeline and adding a custom one

For example, a user registers the `StandardResilienceHandler` as default configuration, but at the same time he/she wants to use the `StandardHedgingHandler` for a particular named `HttpClient`:

```
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());

// For a named HttpClient "custom" we want to remove the StandardResilienceHandler and add instead the StandardHedgingHandler.
services.AddHttpClient("custom")./*Remove StandardResilienceHandler*/.AddStandardHedgingHandler();
```

**API description**

This API provides functionality to remove all assigned handlers from the HttpClient.

**API Usage**

```
services.ConfigureHttpClientDefaults(builder => builder.AddStandardResilienceHandler());
// For a named HttpClient "custom" we want to remove the StandardResilienceHandler and add instead the StandardHedgingHandler.
services.AddHttpClient("custom")
    .RemoveAllResilienceHandlers()
    .AddStandardHedgingHandler();
```

This PR partially resolves #5695 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5801)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Net.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Diagnostics.ExceptionSummarization;
@@ -90,7 +91,10 @@ public static partial class ResilienceHttpClientBuilderExtensions
         {
             for (int i = handlers.Count - 1; i >= 0; i--)
             {
-                handlers.RemoveAt(i);
+                if (handlers[i] is ResilienceHandler)
+                {
+                    handlers.RemoveAt(i);
+                }
             }
         });
         return builder;
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using System.Net.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Diagnostics.ExceptionSummarization;
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
@@ -2,11 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Net.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Diagnostics.ExceptionSummarization;
 using Microsoft.Extensions.Http.Resilience;
 using Microsoft.Extensions.Http.Resilience.Internal;
+using Microsoft.Shared.DiagnosticIds;
 using Microsoft.Shared.Diagnostics;
 using Polly;
 using Polly.Registry;
@@ -80,6 +82,7 @@ public static partial class ResilienceHttpClientBuilderExtensions
     /// </summary>
     /// <param name="builder">The builder instance.</param>
     /// <returns>The value of <paramref name="builder"/>.</returns>
+    [Experimental(diagnosticId: DiagnosticIds.Experiments.Resilience, UrlFormat = DiagnosticIds.UrlFormat)]
     public static IHttpClientBuilder RemoveAllResilienceHandlers(this IHttpClientBuilder builder)
     {
         _ = Throw.IfNull(builder);
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
@@ -75,6 +75,24 @@ public static partial class ResilienceHttpClientBuilderExtensions
         return pipelineBuilder;
     }
 
+    /// <summary>
+    /// Removes all resilience handlers registered earlier.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    public static IHttpClientBuilder RemoveAllResilienceHandlers(this IHttpClientBuilder builder)
+    {
+        _ = Throw.IfNull(builder);
+        _ = builder.ConfigureAdditionalHttpMessageHandlers(static (handlers, _) =>
+        {
+            for (int i = handlers.Count - 1; i >=0; i--)
+            {
+                handlers.RemoveAt(i);
+            }
+        });
+        return builder;
+    }
+
     private static Func<HttpRequestMessage, ResiliencePipeline<HttpResponseMessage>> CreatePipelineSelector(IServiceProvider serviceProvider, string pipelineName)
     {
         var resilienceProvider = serviceProvider.GetRequiredService<ResiliencePipelineProvider<HttpKey>>();
--- a/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
@@ -88,7 +88,7 @@ public static partial class ResilienceHttpClientBuilderExtensions
         _ = Throw.IfNull(builder);
         _ = builder.ConfigureAdditionalHttpMessageHandlers(static (handlers, _) =>
         {
-            for (int i = handlers.Count - 1; i >=0; i--)
+            for (int i = handlers.Count - 1; i >= 0; i--)
             {
                 handlers.RemoveAt(i);
             }
--- a/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
+++ b/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
@@ -286,6 +286,46 @@ public sealed partial class HttpClientBuilderExtensionsTests
         Assert.NotNull(factory.CreateClient("my-client"));
     }
 
+    [Fact]
+    public void RemoveAllResilienceHandlers_ArgumentValidation()
+    {
+        var services = new ServiceCollection();
+        IHttpClientBuilder? builder = null;
+        Assert.Throws<ArgumentNullException>(() => builder!.AddResilienceHandler("pipeline-name", _ => { }));
+        Assert.Throws<ArgumentNullException>(() => builder!.AddResilienceHandler("pipeline-name", (_, _) => { }));
+    }
+
+    [Fact]
+    public void RemoveAllResilienceHandlers_EnsureServicesRemoved()
+    {
+        var services = new ServiceCollection().AddLogging().AddMetrics().AddRedaction();
+
+        IHttpClientBuilder? customBuilder = services.AddHttpClient("custom");
+
+        customBuilder.AddResilienceHandler("pipeline-name", _ => {});
+        customBuilder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Single(handlers);
+        });
+        customBuilder.RemoveAllResilienceHandlers();
+
+        customBuilder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Empty(handlers);
+        });
+
+        customBuilder.RemoveAllResilienceHandlers().AddStandardResilienceHandler();
+
+        customBuilder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Single(handlers);
+        });
+
+        using ServiceProvider serviceProvider = services.BuildServiceProvider();
+        services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
+    }
+
+
     private void ConfigureBuilder(ResiliencePipelineBuilder<HttpResponseMessage> builder) => builder.AddTimeout(TimeSpan.FromSeconds(1));
 
     private class TestMetricsEnricher : MeteringEnricher
--- a/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
+++ b/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
@@ -336,7 +336,6 @@ public sealed partial class HttpClientBuilderExtensionsTests
         services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
     }
 
-
     private void ConfigureBuilder(ResiliencePipelineBuilder<HttpResponseMessage> builder) => builder.AddTimeout(TimeSpan.FromSeconds(1));
 
     private class TestMetricsEnricher : MeteringEnricher
--- a/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
+++ b/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
@@ -330,6 +331,28 @@ public sealed partial class HttpClientBuilderExtensionsTests
         builder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
         {
             Assert.Single(handlers);
+        });
+
+        using ServiceProvider serviceProvider = services.BuildServiceProvider();
+        services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
+    }
+
+    [Fact]
+    public void RemoveAllResilienceHandlers_EnsureOnlyResilienceHandlersRemoved()
+    {
+        var services = new ServiceCollection();
+
+        IHttpClientBuilder? builder = services.AddHttpClient("custom");
+
+        builder.AddHttpMessageHandler(() => new TestHandlerStub(HttpStatusCode.OK));
+        builder.AddStandardResilienceHandler();
+
+        builder.RemoveAllResilienceHandlers();
+
+        builder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Single(handlers);
+            Assert.Equal("TestHandlerStub", handlers.First().GetType().Name);
         });
 
         using ServiceProvider serviceProvider = services.BuildServiceProvider();
--- a/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
+++ b/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
@@ -293,7 +293,6 @@ public sealed partial class HttpClientBuilderExtensionsTests
         var services = new ServiceCollection();
         IHttpClientBuilder? builder = null;
         Assert.Throws<ArgumentNullException>(() => builder!.RemoveAllResilienceHandlers());
-        Assert.Throws<ArgumentNullException>(() => builder!.RemoveAllResilienceHandlers());
     }
 
     [Fact]
@@ -356,7 +355,7 @@ public sealed partial class HttpClientBuilderExtensionsTests
         });
 
         using ServiceProvider serviceProvider = services.BuildServiceProvider();
-        services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
+        serviceProvider.GetRequiredService<IHttpClientFactory>().CreateClient("custom");
     }
 
     private void ConfigureBuilder(ResiliencePipelineBuilder<HttpResponseMessage> builder) => builder.AddTimeout(TimeSpan.FromSeconds(1));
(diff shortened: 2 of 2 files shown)


src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
```diff
@@ -75,6 +77,25 @@ public static IHttpResiliencePipelineBuilder AddResilienceHandler(
...
+    /// <returns>The value of <paramref name="builder"/>.</returns>
+    [Experimental(diagnosticId: DiagnosticIds.Experiments.Resilience, UrlFormat = DiagnosticIds.UrlFormat)]
+    public static IHttpClientBuilder RemoveAllResilienceHandlers(this IHttpClientBuilder builder)
+    {
+        _ = Throw.IfNull(builder);
+        _ = builder.ConfigureAdditionalHttpMessageHandlers(static (handlers, _) =>
+        {
+            for (int i = handlers.Count - 1; i >= 0; i--)
+            {
+                handlers.RemoveAt(i);
```
iliar-turdushev, line 93: This will remove all http message handler types, but we need to remove only `ResilienceHandler`.

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
```diff
@@ -75,6 +78,28 @@ public static IHttpResiliencePipelineBuilder AddResilienceHandler(
...
+        _ = Throw.IfNull(builder);
+        _ = builder.ConfigureAdditionalHttpMessageHandlers(static (handlers, _) =>
+        {
+            for (int i = handlers.Count - 1; i >= 0; i--)
+            {
+                if (handlers[i] is ResilienceHandler)
+                {
+                    handlers.RemoveAt(i);
+                }
+            }
```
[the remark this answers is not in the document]
rainsxng: Handlers doesn't have the `RemoveAll` method. Instead, it has the `Remove` method that deletes the first occurrence of the specific object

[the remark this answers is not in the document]
evgenyfedorov2: oh, it is `IList`, okay.

They cannot decide for two years https://github.com/dotnet/runtime/issues/76375

src/Libraries/Microsoft.Extensions.Http.Resilience/Resilience/ResilienceHttpClientBuilderExtensions.Resilience.cs
rainsxng: Updated

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
```diff
@@ -286,6 +286,56 @@ public void AddResilienceHandler_AuthorityByCustomSelector_NotValidated()
         Assert.NotNull(factory.CreateClient("my-client"));
     }
 
+    [Fact]
```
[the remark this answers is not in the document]
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
```diff
@@ -286,6 +287,78 @@ public void AddResilienceHandler_AuthorityByCustomSelector_NotValidated()
...
+    public void RemoveAllResilienceHandlers_ArgumentValidation()
+    {
+        var services = new ServiceCollection();
+        IHttpClientBuilder? builder = null;
+        Assert.Throws<ArgumentNullException>(() => builder!.AddResilienceHandler("pipeline-name", _ => { }));
```
iliar-turdushev, line 295: I believe we should invoke `RemoveAllResilienceHandlers` here instead of `AddResilienceHanlder`.

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
```diff
@@ -286,6 +287,78 @@ public void AddResilienceHandler_AuthorityByCustomSelector_NotValidated()
...
+
+        builder.RemoveAllResilienceHandlers();
+
+        builder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Empty(handlers);
+        });
+
+        using ServiceProvider serviceProvider = services.BuildServiceProvider();
+        services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
```
iliar-turdushev, line 321:
```suggestion
        serviceProvider.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
```

We already have the `serviceProvider` so we can use it.

iliar-turdushev: Please, fix similar issue in the tests below too.

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
```diff
@@ -286,6 +287,78 @@ public void AddResilienceHandler_AuthorityByCustomSelector_NotValidated()
...
+
+        builder.AddHttpMessageHandler(() => new TestHandlerStub(HttpStatusCode.OK));
+        builder.AddStandardResilienceHandler();
+
+        builder.RemoveAllResilienceHandlers();
+
+        builder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Single(handlers);
+            Assert.Equal("TestHandlerStub", handlers.First().GetType().Name);
```
iliar-turdushev, line 355: Can we compare types here instead of type names? That would be more reliable in case somebody decides to rename the type.

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
```diff
@@ -286,6 +287,78 @@ public void AddResilienceHandler_AuthorityByCustomSelector_NotValidated()
         Assert.NotNull(factory.CreateClient("my-client"));
     }
 
+    [Fact]
+    public void RemoveAllResilienceHandlers_ArgumentValidation()
+    {
+        var services = new ServiceCollection();
+        IHttpClientBuilder? builder = null;
+        Assert.Throws<ArgumentNullException>(() => builder!.RemoveAllResilienceHandlers());
+        Assert.Throws<ArgumentNullException>(() => builder!.RemoveAllResilienceHandlers());
```
iliar-turdushev, line 296: This line is the copy of the above one. We can remove it.
```suggestion
```

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
```diff
@@ -286,6 +287,78 @@ public void AddResilienceHandler_AuthorityByCustomSelector_NotValidated()
...
+        builder.RemoveAllResilienceHandlers();
+
+        builder.ConfigureAdditionalHttpMessageHandlers((handlers, _) =>
+        {
+            Assert.Single(handlers);
+            Assert.Equal(typeof(TestHandlerStub), handlers.First().GetType());
+        });
+
+        using ServiceProvider serviceProvider = services.BuildServiceProvider();
+        services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
```
iliar-turdushev, line 359:
```suggestion
        serviceProvider().GetRequiredService<IHttpClientFactory>().CreateClient("custom");
```

iliar-turdushev (approved):
@rainsxng The PR looks good. Thank you for the contribution.

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Resilience/HttpClientBuilderExtensionsTests.Resilience.cs
rainsxng: Done

PR #5806, opened 2025-01-20
Add container.cpu.time metric

### LINKED ISSUE #5800: Add new metric to Resource Monitoring

evgenyfedorov2:
Fixed #5800


evgenyfedorov2:
@andrey-noskov FYI

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
```diff
@@ -205,6 +208,15 @@ private double MemoryPercentage(Func<ulong> getMemoryUsage)
         }
     }
 
+    private IEnumerable<Measurement<double>> GetCpuTime()
+    {
+        using var jobHandle = _createJobHandleObject();
+        var basicAccountingInfo = jobHandle.GetBasicAccountingInfo();
+
+        yield return new(basicAccountingInfo.TotalUserTime / TicksPerSecoundDouble, [new KeyValuePair<string, object?>("cpu.mode", "user")]);
```
amadeuszl, line 216: What's the difference between UserTime and KernelTime from perspective of container running in Cosmic?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -65,6 +67,7 @@ public LinuxUtilizationProvider(IOptions<ResourceMonitoringOptions> options, ILi
         var meter = meterFactory.Create(ResourceUtilizationInstruments.MeterName);
 #pragma warning restore CA2000 // Dispose objects before losing scope
 
+        _ = meter.CreateObservableCounter(name: ResourceUtilizationInstruments.ContainerCpuTime, observeValues: GetCpuTime, unit: "s", description: "CPU time used by the container.");
```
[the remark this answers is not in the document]
evgenyfedorov2: The counter is by [convention](https://github.com/open-telemetry/semantic-conventions/blob/c828bebb75977b2139b53c2a0aad89b9bff767e3/docs/system/container-metrics.md#metric-containercputime).
Also for reference, .NET runtime uses counter - [doc](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-runtime#metric-dotnetprocesscputime) and [source code](https://github.com/dotnet/runtime/blob/a2bb26d4b72a8fd19c0582040aa9d3efce23de9f/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/RuntimeMetrics.cs#L151)

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
evgenyfedorov2: Not sure if I got the question right, but I don't think it is different from what you have without containers/k8s. See also https://unix.stackexchange.com/questions/87625/what-is-difference-between-user-space-and-kernel-space

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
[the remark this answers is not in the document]
amadeuszl: Makes sense

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
amadeuszl: For Linux (cgroupv2) I see that we're going to report (correct me if I'm wrong):
`cpu.mode=user` is `/sys/fs/cgroup/cpu.stat` so CPU time from container related cgroup
`cpu.mode=system` is `/proc/stat`) so host CPU time, which is whole machine, which in fact can host hundreds of containers.

For Windows based on [JOBOBJECT](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-jobobject_basic_accounting_information):
`cpu.mode=user` - TotalUserTime - `The total amount of user-mode execution time for all active processes associated with the job`, what's the job in this case? And what's user-mode?
`cpu.mode=system` - TotalKernelTime - `The total amount of kernel-mode execution time for all active processes associated with the job`, what's the job in this case? And what's kernel-mode?
Basically, I'm trying to double check what those values are representing, so tags make sense, and that they represent same thing in both OSes.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
```diff
@@ -15,6 +16,7 @@ namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring.Windows;
 internal sealed class WindowsContainerSnapshotProvider : ISnapshotProvider
 {
     private const double Hundred = 100.0d;
+    private const double TicksPerSecoundDouble = TimeSpan.TicksPerSecond;
```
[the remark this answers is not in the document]
evgenyfedorov2: Done

RussKie (approved):

mScott224 (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
evgenyfedorov2: `what's the job in this case? And what's kernel-mode?` - on a very high level, job object is similar to Linux cgroups in terms of managing resources for containers. This is also what K8s uses to manage resources when running containers on Windows.

`Basically, I'm trying to double check what those values are representing, so tags make sense, and that they represent same thing in both OSes.`
We approach it this way: User mode is our app process, and Kernel (System) mode is all other processes. This makes most sense in a container environment where you typically have your one application process running in user mode, all other processes are system because they are run by OS.

Anyway, those values are not something new, we already use them for calculating other existing CPU metrics in Resource Monitoring.

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/AcceptanceTest.cs
```diff
@@ -382,25 +385,25 @@ public Task ResourceUtilizationTracker_And_Metrics_Report_Same_Values_With_Cgrou
...
         var memoryLimitFromGauge = 0.0d;
-        using var e = new ManualResetEventSlim();
 
         object? meterScope = null;
         listener.InstrumentPublished = (Instrument instrument, MeterListener meterListener)
             => OnInstrumentPublished(instrument, meterListener, meterScope);
-        listener.SetMeasurementEventCallback<double>((m, f, _, _)
-            => OnMeasurementReceived(m, f, ref cpuFromGauge, ref cpuLimitFromGauge, ref cpuRequestFromGauge, ref memoryFromGauge, ref memoryLimitFromGauge));
+        listener.SetMeasurementEventCallback<double>((m, f, tags, _)
+            => OnMeasurementReceived(m, f, tags, ref cpuUserTime, ref cpuKernelTime, ref cpuFromGauge, ref cpuLimitFromGauge, ref cpuRequestFromGauge, ref memoryFromGauge, ref memoryLimitFromGauge));
```
amadeuszl: For readability let's break line per parameter here too

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/AcceptanceTest.cs
```diff
@@ -457,10 +453,12 @@ private static void OnInstrumentPublished(Instrument instrument, MeterListener m
 #pragma warning restore S1067 // Expressions should not be too complex
     }
 
+#pragma warning disable S107 // Methods should not have too many parameters
     private static void OnMeasurementReceived(
-        Instrument instrument, double value,
-        ref double cpuFromGauge, ref double cpuLimitFromGauge, ref double cpuRequestFromGauge,
-        ref double memoryFromGauge, ref double memoryLimitFromGauge)
+        Instrument instrument, double value, ReadOnlySpan<KeyValuePair<string, object?>> tags,
```
amadeuszl: Let's break line on each parameter for readability

evgenyfedorov2: Done

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/AcceptanceTest.cs
evgenyfedorov2: Done

amadeuszl (approved):

PR #5809, opened 2025-01-21
Enable overriding FunctionInvokingChatClient function discovery

stephentoub:
Allows a derived instance to more easily pick which `AIFunction` to use based on the call content being provided.

Also augment failure message for a function not found with the name of the function requested, to help mitigate hallucination.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5809)


SteveSandersonMS (approved):
This seems totally fine though I'm not sure what the use case is. A subclass would rarely be able to select anything other than the function that matches by name, since the supplied args won't match any other function in general.

I'd be interested if you could clarify the use case, and it would be good to have a test case showing this works. But no objection to the new API.

SteveSandersonMS:
I haven't experienced the need for it and am not certain in what cases a developer would realistically be able to fix any incorrect function names they received. If you're confident there are use cases, it seems reasonably safe to allow it, but if you're not then I'd be fine with leaving this until a need becomes clearer.

Issue #5719, opened 2024-12-03
Revisit DataContent and friends


stephentoub:
A few issues to sort out:
 
- **Derived types.** We currently have DataContent (base type), ImageContent (derives from DataContent), and AudioContent (derives from DataContent). It's not clear what significant value ImageContent / AudioContent add: they're just constructors and enable applying more strong-typing based on media type. But presumably chat client implementations should also handle cases where DataContent is used rather than the derived type (we're not currently), and once you're handling both, what's the benefit of having the derived types.

- **Videos**. We currently lack a VideoContent. If we're going to have ImageContent and AudioContent, we should also have VideoContent.

- **Non-in-memory data**. We currently don't have any Stream-based support; everything needs to be loaded into memory. We should look at either adding a StreamingDataContent or augmenting DataContent with support for Stream. As these can be stored in chat message lists, we'll need to work through that. Should it require any Stream to be seekable? What should happen with JSON serialization employed as part of logging / caching / etc.?

- **Realtime**. Do these content types work fine for representing partial data as might be sent or received as part of a streaming request / response?

cc: @SteveSandersonMS, @RogerBarreto, @eiriktsarpalis

SteveSandersonMS:
> It's not clear what significant value ImageContent / AudioContent add ...  once you're handling both, what's the benefit of having the derived types

I get that the benefit of the subclasses is nonobvious, hence I was originally proposing to remove them too. However, the benefit of these subclasses isn't for `IChatClient` implementors - it's for app developers.

For example, if you want your app to support displaying image responses, you can check if a content item `is ImageContent` and if so, use its binary data in an `<img>` tag (if it's a web app) or equivalents for native apps. If there was no `ImageContent` and instead all image/audio/video data was represented as `DataContent`, then the app would need to do some `MediaType` parsing and detect a set of known MIME types, which is a whole layer of inconvenience that app devs likely don't want to deal with nor think they should have to. Note that it's not sufficient just to check `MediaType.StartsWith("image/")` etc., given the existence of media types like `multipart/voice-message` or the `application/*` ones.

I'd be fine with eliminating `ImageContent`/`AudioContent` if we also add some way other than MIME type to identify "this is an image" vs audio etc. For example we could have a mandatory `ContentType` enum with values (image, audio, binary). That's almost equivalent to using subclasses as a discriminator.

> Videos. We currently lack a VideoContent. If we're going to have ImageContent and AudioContent, we should also have VideoContent.

Sounds good

[...]

PR #5814, opened 2025-01-24
Remove `ImageContent` and `AudioContent`

MackinnonBuck:
Removes the `ImageContent` and `AudioContent` types.

The recommendation will be that chat client implementors and app developers use `DataContent.MediaType` to distinguish between the various content types.

Fixes https://github.com/dotnet/extensions/issues/5719
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5814)


src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -484,12 +484,16 @@ private static List<ChatMessageContentItem> GetContentParts(IList<AIContent> con
                     parts.Add(new ChatMessageTextContentItem(textContent.Text));
                     break;
 
-                case ImageContent imageContent when imageContent.Data is { IsEmpty: false } data:
-                    parts.Add(new ChatMessageImageContentItem(BinaryData.FromBytes(data), imageContent.MediaType));
-                    break;
+                case DataContent dataContent when dataContent.HasImageMediaType():
+                    if (dataContent.Data is { IsEmpty: false } data)
```
stephentoub, line 488: This is fine, but fwiw I find the IsEmpty: false double-negative a little confusing. This could also be:
```C#
if (dataContent.ContainsData)
{
    ... dataContent.Data ...
}
```

src/Shared/AIExtensions/DataContentMediaTypeExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+#endif
+internal static class DataContentMediaTypeExtensions
+{
+    /// <summary>
+    /// Determines whether the <see cref="DataContent"/> represents an image.
+    /// </summary>
+    /// <param name="content">The <see cref="DataContent"/>.</param>
+    /// <returns><see langword="true"/> if the content represents an image; otherwise, <see langword="false"/>.</returns>
+    public static bool HasImageMediaType(this DataContent content)
+        => content.MediaType is { } mediaType && mediaType.StartsWith("image/", StringComparison.OrdinalIgnoreCase);
```
stephentoub, line 22:
```suggestion
        => content.MediaType?.StartsWith("image/", StringComparison.OrdinalIgnoreCase) is true;
```
?

If this is something we expect every consumer to need to do, should we bake it in with a helper instance method on DataContent, e.g.
```C#
public bool HasMediaTypePrefix(string prefix)
```
? ("Prefix" isn't ideal naming, but HasMediaTypeType doesn't sound right, either :))

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -375,7 +375,7 @@ private IEnumerable<OllamaChatRequestMessage> ToOllamaChatRequestMessages(ChatMe
         OllamaChatRequestMessage? currentTextMessage = null;
         foreach (var item in content.Contents)
         {
-            if (currentTextMessage is not null && item is not ImageContent)
+            if (currentTextMessage is not null && item is not DataContent)
```
stephentoub, line 378: What ends up happening to the logic if it's a DataContent but HasImageMediaType below is false? Might we end up coalescing incorrectly / yielding items out of order?

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/StreamingChatCompletionUpdateTests.cs
```diff
@@ -91,8 +91,8 @@ public void Text_GetSet_UsesFirstTextContent()
             Role = ChatRole.User,
             Contents =
             [
-                new AudioContent("http://localhost/audio"),
-                new ImageContent("http://localhost/image"),
+                new DataContent("http://localhost/audio", mediaType: "audio/mpeg"),
+                new DataContent("http://localhost/image", mediaType: "image/png"),
```
stephentoub, line 95: Did these tests actually require providing a media type now when wasn't provided before?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: I usually go for `is not []` but this requires polyfilling `System.Index` in netstandard2.0

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
MackinnonBuck: Hm, I don't think we'd get any out-of-order messages, because the switch cases that directly `yield return` a result still can't get reached if `currentTextMessage` is non-`null`.

However, there is a subtle change in coalescing behavior here. Consider the following message content:
1. Text ("Hello, world!")
2. Audio ("hello.wav")
3. Image ("hello.png")

Then, prior to this change, the return value would be:
```
[
  OllamaChatRequestMessage { Content = "Hello, world!" },
  OllamaChatRequestMessage { Images = ["hello.png"] }
]
```

But now, it would be:
```
[
  OllamaChatRequestMessage { Content = "Hello, world!", Images = ["hello.png"] }
]
```

I don't know if I would necessarily say the new behavior is incorrect. But it might be better deduplicate the logic a bit and not have two checks that share the responsibility of determining when content can be added to an existing message. This would allow us to preserve the existing behavior.

src/Shared/AIExtensions/DataContentMediaTypeExtensions.cs
MackinnonBuck: Yeah, adding a new instance member sounds fine to me 🙂

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/StreamingChatCompletionUpdateTests.cs
MackinnonBuck: No, I was just did this to replicate how we'd expect image and audio content to be created in the real world (it's more important now that `mediaType` is specified, since it's the main way to determine the content's type).

Happy to revert this change if it just adds confusion.

Also - should we consider removing the default value for the `mediaType` constructor parameter? This makes it so that you have to deliberately specify a `null` media type if you want one. It might reduce the chance that someone would leave it out by accident.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
MackinnonBuck: I've just updated the logic to retain the original behavior.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/StreamingChatCompletionUpdateTests.cs
MackinnonBuck: Update: Removed `mediaType` where isn't not needed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -105,6 +105,14 @@ public DataContent(ReadOnlyMemory<byte> data, string? mediaType = null)
         _data = data;
     }
 
+    /// <summary>
+    /// Determines whether the <see cref="MediaType"/> has the specified prefix.
+    /// </summary>
+    /// <param name="prefix">The media type prefix.</param>
+    /// <returns><see langword="true"/> if the <see cref="MediaType"/> has the specified prefix, otherwise <see langword="false"/>.</returns>
+    public bool HasMediaTypePrefix(string prefix)
```
stephentoub, line 113: Reading this again, simply "MediaTypeStartsWith" instead of "HasMediaTypePrefix" might be a better name.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -484,12 +484,16 @@ private static List<ChatMessageContentItem> GetContentParts(IList<AIContent> con
                     parts.Add(new ChatMessageTextContentItem(textContent.Text));
                     break;
 
-                case ImageContent imageContent when imageContent.Data is { IsEmpty: false } data:
-                    parts.Add(new ChatMessageImageContentItem(BinaryData.FromBytes(data), imageContent.MediaType));
-                    break;
+                case DataContent dataContent when dataContent.HasMediaTypePrefix("image/"):
+                    if (dataContent.ContainsData)
+                    {
+                        parts.Add(new ChatMessageImageContentItem(BinaryData.FromBytes(dataContent.Data!.Value), dataContent.MediaType));
```
stephentoub, line 490: I'm surprised the `!` is needed here on `Data!`. The MemberNotNullWhen on Data is supposed to address this:
https://github.com/dotnet/extensions/blob/6e20a080cb52b7a842610714118b35fb0baaf3dc/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs#L183-L185
Are you actually getting a warning here without the exclamation? If yes, might be worth following up with Roslyn folks to understand why.

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
MackinnonBuck: Yeah, I was getting a warning. Looks like the issue was that the `[MemberNotNullWhen]` attribute should have been on the `ContainsData` property, not the `Data` property.

After moving the attribute, the warning goes away (and a few other existing locations were made simpler too).

PR #5823, opened 2025-01-28
Inheritdoc fixes

gewarren:
- [learn.microsoft.com doesn't understand inheritdoc from something that itself has inheritdoc](https://buildapi.docs.microsoft.com/Output/PullRequest/500c749d-de78-3d4a-00e4-269d599bf9ec/202501050810051900-10808/BuildReport?accessString=8150976b02bdfaba2a8d4273cf5fc53d484938991b2766fda7369ec3e0c63520), so added cref to type that provides the docs
- FakeHostOptions had a bug where the inheritdoc cref pointed to the type instead of the member, so [docs were incorrect](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.testing.fakehostoptions?view=net-9.0-pp#properties):

  image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5823)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
@@ -44,7 +44,7 @@ public abstract class CachingChatClient : DelegatingChatClient
     /// </remarks>
     public bool CoalesceStreamingUpdates { get; set; } = true;
 
-    /// <inheritdoc />
+    /// <inheritdoc cref="IChatClient"/>
     public override async Task<ChatCompletion> CompleteAsync(IList<ChatMessage> chatMessages, ChatOptions? options = null, CancellationToken cancellationToken = default)
     {
         _ = Throw.IfNull(chatMessages);
@@ -63,7 +63,7 @@ public abstract class CachingChatClient : DelegatingChatClient
         return result;
     }
 
-    /// <inheritdoc />
+    /// <inheritdoc cref="IChatClient"/>
     public override async IAsyncEnumerable<StreamingChatCompletionUpdate> CompleteStreamingAsync(
         IList<ChatMessage> chatMessages, ChatOptions? options = null, [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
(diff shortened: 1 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
```diff
@@ -44,7 +44,7 @@ protected CachingChatClient(IChatClient innerClient)
     /// </remarks>
     public bool CoalesceStreamingUpdates { get; set; } = true;
 
-    /// <inheritdoc />
+    /// <inheritdoc cref="IChatClient.CompleteAsync(IList{ChatMessage}, ChatOptions?, CancellationToken)"/>
```
RussKie, line 47: @gewarren what's the reason for needing the explicit cref?

RussKie (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
gewarren: @RussKie [learn.microsoft.com doesn't understand inheritdoc from something that itself has inheritdoc](https://buildapi.docs.microsoft.com/Output/PullRequest/500c749d-de78-3d4a-00e4-269d599bf9ec/202501050810051900-10808/BuildReport?accessString=8150976b02bdfaba2a8d4273cf5fc53d484938991b2766fda7369ec3e0c63520). If you don't like this change, I can try to get the Learn platform team to fix this instead.

gewarren: At least I assume it's not working on Learn because the parent type also has `inheritdoc` on this member. I'm not 100% sure it's the problem, but it's an educated guess.

stephentoub: It'd be great if the platform could fix it. Otherwise given there's compiler / VS support, we'll likely continue to have a long tail of such issues reoccurring (in this and other repos).

gewarren: https://dev.azure.com/ceapex/Engineering/_workitems/edit/1026690

PR #5826, opened 2025-01-29
Move AIFunction parameter schematization from parameter level to function level.

### LINKED ISSUE #5655: Microsoft.Extensions.AI.AIFunctionMetadata should allow to be initialized with a json schema string
### Description

I built a C# code generator to implement the boilerplate for Tools/Agents/Functions.
I already generate at compile time the json schema string complying to the OpenAI specification for any agent.

The AIFunctionMetadata should provide a way to be initialized with the json schema directly.
Currently I can only specify (but there is no example, therefore I don't know how) the schema in `AIFunctionReturnParameterMetadata` and `AIFunctionParameterMetadata`.

Anyway, the problem with `AIFunctionParameterMetadata` is that the json section specifying whether the  parameter is required or not, belongs to the parent key (`properties`). This means that even if I can initialize this class with each parameter schema, I cannot specify the `required` field.

To get rid of all the problems and completely avoid the serialization at runtime, the root class AIFunctionMetadata should allow to take the entire schema. This also provides better performance.

### Reproduction Steps

```csharp

    private AIFunctionMetadata CreateMetadata(
        string functionName,
        string functionDescription,
        string returnSchema,
        IEnumerable<(string name, string schema)> parametersSchema)
    {
        List<AIFunctionParameterMetadata> parameters = new();
        foreach (var par in parametersSchema)
        {
            parameters.Add(new AIFunctionParameterMetadata(par.name)
            {
                Schema = par.schema,
            });
        }

        // how can I specify the 'required' field for each parameter?

        var returnParameter = new AIFunctionReturnParameterMetadata()
        {
            Schema = returnSchema,
        };

        var meta = new AIFunctionMetadata(functionName)
        {
            Description = functionDescription,
            ReturnParameter = returnParameter,
            Parameters = parameters,
        };

        return meta;
    }

```

### Expected behavior

I expect:
- to be able to initialize the `AIFunctionMetadata ` with the entire schema, as for the OpenAI specification.
- to be able to avoid any serialization or reflection in order to obtain better performance

### Actual behavior

See example above: the ability to specify the schema per-parameter is not sufficient and does not provide any benefit

### Regression?

No

### Known Workarounds

Use reflection and serialize at runtime

### Configuration

.NET 9.0.100

### Other information

_No response_

eiriktsarpalis:
Fix #5655.


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -100,6 +101,35 @@ public AIFunctionReturnParameterMetadata ReturnParameter
...
+    /// <summary>Gets a JSON Schema describing the function and its input parameters.</summary>
+    /// <remarks>
+    /// <para>
+    /// When specified, declares a self-contained JSON schema document that describes the function and its input parameters.
+    /// A simple example of a JSON schema for a function that adds two numbers together is shown below:
+    /// </para>
+    /// <code>
+    /// {
+    ///   "title" : "addNumbers",
+    ///   "description": "A simple function that adds two numbers together.",
```
stephentoub, line 113: Why include title and description? Do any of the services we're aware of expect these? We're carrying that information separately. If this schema is sent to OpenAI or another service that directly supports this format, seems like worst case the schema would be rejected because of the extra properties, and best case they wouldn't be rejected but folks will end up paying for the additional tokens used for this unnecessary data?

eiriktsarpalis: For the leaf client implementations I'm actually mapping this out to the existing tool model types so that any redundant keywords get filtered out. I'm mostly doing it out of an abundance of caution, I haven't tested what happens if you let arbitrary schema documents pass through to the request. Looking at the OpenAI/Azure AI clients specifically, given that the schema type is `BinaryData` I would assume that any schema document would be acceptable provided it met OpenAI's definition of JSON schema.

eiriktsarpalis: > Why include title and description? Do any of the services we're aware of expect these? 

Even though I'm not doing it in this PR yet, the idea is that you could extract this information from the schema documentation as opposed to using the corresponding metadata object. But even if we don't end up going there, adding the title and description represents the most metadata-rich JSON schema document we can extract from a function.

stephentoub: Ok. We should keep the filtering the leaf clients, if for no other reason than cost / token count.

eiriktsarpalis: I'm fine doing it like this for now, although in certain ways it contradicts the spirit of #5655 i.e. let users specify their schemas for a given function.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -25,100 +25,90 @@ namespace Microsoft.Extensions.AI;
...
+    private const string PropertiesPropertyName = "properties";
+    private const string RequiredPropertyName = "required";
+    private const string AdditionalPropertiesPropertyName = "additionalProperties";
+    private const string DefaultPropertyName = "default";
+    private const string RefPropertyName = "$ref";
+
     /// <summary>The uri used when populating the $schema keyword in inferred schemas.</summary>
     private const string SchemaKeywordUri = "https://json-schema.org/draft/2020-12/schema";
 
-    /// <summary>Soft limit for how many items should be stored in the dictionaries in <see cref="_schemaCaches"/>.</summary>
```
eiriktsarpalis, line 31: Note that this change removes the JSON schema caching since there's no longer a need for leaf clients to infer the schema on their side -- AIFunctions should always supply a valid schema now.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -100,6 +104,43 @@ public AIFunctionReturnParameterMetadata ReturnParameter
...
+    ///   "type": "object",
+    ///   "properties": {
+    ///     "a" : { "type": "number" },
+    ///     "b" : { "type": "number", "default": 1 }
+    ///   }, 
+    ///   "required" : ["a"]
```
stephentoub, line 122: It's a bit inconsistent now that we have the return schema contained separately from the function schema. Is there any standard way with schemas for functions to represent a return value, such that we could augment the function schema with it rather than having it in the separate return instance? I realize none of the AI vendors care about it today.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -376,33 +373,19 @@ private ChatCompletionsOptions ToAzureAIOptions(IList<ChatMessage> chatContents,
     /// <summary>Converts an Extensions function to an AzureAI chat tool.</summary>
     private static ChatCompletionsToolDefinition ToAzureAIChatTool(AIFunction aiFunction)
     {
-        BinaryData resultParameters = AzureAIChatToolJson.ZeroFunctionParametersSchema;
-
-        var parameters = aiFunction.Metadata.Parameters;
-        if (parameters is { Count: > 0 })
+        BinaryData functionParameters = AzureAIChatToolJson.ZeroFunctionParametersSchema;
+        if (aiFunction.Metadata.Schema is { } schema)
```
stephentoub, line 377: It's no longer nullable, right? Why is this check needed?

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -466,20 +465,21 @@ private IEnumerable<OllamaChatRequestMessage> ToOllamaChatRequestMessages(ChatMe
         }
     }
 
-    private static OllamaTool ToOllamaTool(AIFunction function) => new()
+    private static OllamaTool ToOllamaTool(AIFunction function)
     {
-        Type = "function",
-        Function = new OllamaFunctionTool
+        OllamaFunctionToolParameters toolParameters = function.Metadata.Schema is { } schema
```
stephentoub, line 470: Same question... isn't the else dead code?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -398,28 +398,16 @@ private static ChatTool ToOpenAIChatTool(AIFunction aiFunction)
             strictObj is bool strictValue ?
             strictValue : null;
 
-        BinaryData resultParameters = OpenAIChatToolJson.ZeroFunctionParametersSchema;
-
-        var parameters = aiFunction.Metadata.Parameters;
-        if (parameters is { Count: > 0 })
+        BinaryData functionParameters = OpenAIChatToolJson.ZeroFunctionParametersSchema;
+        if (aiFunction.Metadata.Schema is { } schema)
```
stephentoub, line 402: Is this still the right check? Will ZeroFunctionParametersSchema ever be used now?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -29,22 +28,18 @@ public static ConversationFunctionTool ToConversationFunctionTool(this AIFunctio
     {
         _ = Throw.IfNull(aiFunction);
 
-        var parametersSchema = new ConversationFunctionToolParametersSchema
+        BinaryData functionParameters = OpenAIChatToolJson.ZeroFunctionParametersSchema;
+        if (aiFunction.Metadata.Schema is { } schema)
```
stephentoub, line 32: Ditto

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -10,6 +10,7 @@
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
 using OpenAI.RealtimeConversation;
+using static Microsoft.Extensions.AI.OpenAIModelMappers;
```
stephentoub, line 13: I'd prefer if we avoided `using static` like this. I have a hard time reading the code to know where things come from, and end up scouring around looking for private methods.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: I did a last minute change to non-nullable but forgot to update the leaf clients since it didn't warn. Removed the dead code.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
eiriktsarpalis: Not that I could find. Unless the AI vendors legislate a schema keyword for the return type into existence, this would need to be kept as a separate property for now. It don't believe it's going to be necessary though -- I think we might as well remove it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -17,10 +17,13 @@ namespace Microsoft.Extensions.AI;
 public sealed class AIFunctionMetadata
 {
     /// <summary>The name of the function.</summary>
-    private string _name = string.Empty;
+    private readonly string _name = string.Empty;
 
     /// <summary>The description of the function.</summary>
-    private string _description = string.Empty;
+    private readonly string _description = string.Empty;
```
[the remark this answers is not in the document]
eiriktsarpalis: Done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -100,6 +104,43 @@ public AIFunctionReturnParameterMetadata ReturnParameter
...
+    /// <para>
+    /// The metadata present in the schema document plays an important role in guiding AI function invocation.
+    /// Functions should incorporate as much detail as possible. The arity of the "properties" keyword should
+    /// also match the length of the <see cref="Parameters"/> list.
+    /// </para>
+    /// <para>
+    /// When no schema is specified, consuming chat clients should assume the "{}" or "true" schema, indicating that any JSON input is admissible.
+    /// </para>
+    /// </remarks>
+    public JsonElement Schema
```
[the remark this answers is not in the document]
stephentoub: How would SK layer on top if it lacked that information?

Is there not still value / scenarios in having the .NET type information for parameters exposed, so that another consumer of the AIFunction can use that to understand what .NET types it's actually expecting? Seems like this is inching back towards having the arguments dictionary being strongly typed as JsonElement.

And as we spoke about offline, we've seen folks take the return schema information and add it into function call results to help the LLM better understand what the return information represents. Unless we incorporate the return schema into the function schema, we'll be cutting off that possibility, no?

I'm all for simplifying, but I worry this would be cutting off useful information.

@SteveSandersonMS?

stephentoub (approved):

PR #5839, opened 2025-02-04
Small fixes for chat template

MackinnonBuck:
Fixes some issues that unblock local development:
1. Add a `launchSettings.json` file. Without this, the environment defaults to `Production` and this causes the page to fail to load when running with `dotnet run`.
2. Fix the `TailwindBuild` target to always run after the `BuildNpmDependencies` target. This just required changing the `AfterTargets` to a `DependsOnTargets`.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5839)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Properties/launchSettings.json
```diff
@@ -0,0 +1,23 @@
...
+      "applicationUrl": "http://localhost:5500",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    },
+    "https": {
+      "commandName": "Project",
+      "dotnetRunMessages": true,
+      "launchBrowser": true,
+      "applicationUrl": "https://localhost:5501;http://localhost:5500",
```
SteveSandersonMS, line 17: This looks good but do we need to generate these port numbers dynamically like the ASP.NET Core templates do?

I think the code needed is something like this: https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/.template.config/template.json#L58-L101

MackinnonBuck: Yeah, good idea. Just added this.

SteveSandersonMS:
Excellent - thanks!

I don't know how much difference it makes but the ASP.NET Core templates also do something to suppress these `kestrelHttp(s)Port` parameters from appearing in the CLI: https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/.template.config/dotnetcli.host.json#L11-L16

If you think it's reasonable to copy that pattern here that's great, but like I say I'm not certain what impact it has.

SteveSandersonMS (approved):

MackinnonBuck:
> If you think it's reasonable to copy that pattern here that's great, but like I say I'm not certain what impact it has.

Yep, seems reasonable to me 🙂

PR #5848, opened 2025-02-05
Allow custom filtering logic for FakeLogger #5615

Demo30:


Demo30:

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
```diff
@@ -34,6 +35,17 @@ public class FakeLogCollectorOptions
...
+    /// <summary>
+    /// Gets or sets custom filters used to filter out records to be collected.
+    /// </summary>
+    /// <value>Default is an empty array.</value>
+    /// <remarks>
+    /// Defaults to an empty array, which doesn't filter any records.
+    /// If not empty, only records for which the filter function returns <see langword="true" /> will be collected by the fake logger.
+    /// </remarks>
+    [Experimental(DiagnosticIds.Experiments.Telemetry)]
+    public IEnumerable<Func<FakeLogRecord, bool>> CustomFilters { get; set; } = Array.Empty<Func<FakeLogRecord, bool>>();
```
RussKie, line 47: This setter will allow setting the property to null. Is this desirable?

test/Libraries/Microsoft.Extensions.Diagnostics.Testing.Tests/Logging/FakeLoggerTests.cs
```diff
@@ -283,4 +281,59 @@ public void Scopes()
...
+    {
+        const string NotIgnoredMessage1 = "Not ignored message 1";
+        const string NotIgnoredMessage2 = "Not ignored message 2";
+        const string IgnoredMessage = "Ignored message";
+
+        // Given
+        var options = new FakeLogCollectorOptions
+        {
+            CustomFilters = [
+                r => !r.Message.Equals(IgnoredMessage, StringComparison.Ordinal),
```
RussKie, line 296: If you're using Ordinal comparison, then more idiomatic to write as follows:
```suggestion
                r => r.Message != IgnoredMessage,
```

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollector.cs
```diff
@@ -129,6 +129,15 @@ internal void AddRecord(FakeLogRecord record)
...
         }
 
+        foreach (var condition in _options.CustomFilters)
+        {
+            if (!condition(record))
+            {
+                // record was filtered out by a custom filter
+                return;
+            }
+        }
```
RussKie, line 139: Would this work?
```suggestion
        if (_options.CustomFilters.Any(condition => !condition(record))
        {
            // record was filtered out by a custom filter
            return;
        }
```

RussKie: This looks sus - if any record hasn't been filtered (i.e., `!filter(record)`) then exit because "record was filtered out by a custom filter". Don't you think?

RussKie:
@Demo30 when you're get closer to completion, please provide more details in the body of the commit message as well as in the [original message](https://github.com/dotnet/extensions/pull/5848#issue-2832836221) explaining the nature of the changes. This is to allow future maintainers to look at the git commit and understand why the change was done instead of needing to locate the original CR.

Also, in the future please use [GitHub keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in the body of the git commit and not in the subject.

Thank you!

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
RussKie: > Since Nullable is enabled, attempt to setting it to null issues a warning. 

"Nullable" only provides annotations which only work at compile-time. That is, those _may_ result in warnings, however, they don't block the user code from assigning null (which results in NRE at runtime and may crash the app).

> How about defining it as nullable, defaulting to null and having a null check at the place of usage?

It's definitely an option to expect a developer to implement null-checks, but since the API is annotated the good developer won't do that (because they trust us) and a mediocre developer won't do that (because they are mediocre). Either way, the likelihood of a user receiving an NRE is pretty high because **our** code isn't living up to the annotation and not handling nulls.

Null handling either implies throwing if null is assigned or setting (or just returning) an empty array (like we're doing for the init state). Which implementation better depends on usecases you're trying to address.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollector.cs
RussKie: > RE: linq vs not-linq

It's may worth checking perf switching to linq. If this is a hot path, then non-linq version may behave better. ¯\\_(ツ)\_/¯

> the reversed semantic might work better

To me the current implementation looks confusing to me (which means it'll likely confuse others)...
I sense the predicate's logic is inverse - i.e., "_is the record caught by the filter => false_", whereas _perhaps_ it should be "_is the record handled/filtered by the filter => true_". This will remove the inversion in the iterator, i.e., `if (_options.CustomFilters.Any(filter)) return true`.

WDYT?

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: Yes, that's want I meant, let's define the contract as nullable with null as default and handle the null gracefully **on our side**. This way the user will never reach NRE.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollector.cs
Piedone: My idea under the issue was that only those records will be collected that are allowed by the filter, i.e. those that the filter returns `true` for (and if not filter is specified, then all records are allowed). This is following the logic of the other filtering here above: e.g., if `FilteredLevels` contains items, then only those records that have the level contained there will be allowed.

So, what's currently in the PR is IMO correct.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
```diff
@@ -34,6 +35,17 @@ public class FakeLogCollectorOptions
     /// </remarks>
     public ISet<LogLevel> FilteredLevels { get; set; } = new HashSet<LogLevel>();
 
+    /// <summary>
+    /// Gets or sets custom filters used to filter out records to be collected.
```
RussKie, line 39:
```suggestion
    /// Gets or sets filters which control what records would be collected by the fake logger.
```

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
```diff
@@ -34,6 +35,17 @@ public class FakeLogCollectorOptions
...
+    /// </summary>
+    /// <value>Default is an empty array.</value>
+    /// <remarks>
+    /// Defaults to an empty array, which doesn't filter any records.
+    /// If not empty, only records for which the filter function returns <see langword="true" /> will be collected by the fake logger.
```
RussKie, line 44: I think this is the point of the disconnect for me: "_only records for which the filter function returns <see langword="true" /> will be collected by the fake logger._". 
To me if a filter function returns true, it means the record was filtered out.

I recommend inverting the meaning here, which will invert the logic in the collector implementation too.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
RussKie, line 47:
```suggestion
    /// <summary>
    /// Gets or sets filters which control what records would be collected by the fake logger.
    /// </summary>
    /// <remarks>
    ///  The default value is an empty array, which means all records would be collected.
    ///  To filter a record out, the filter function must return <see langword="true" />.
    /// </remarks>
    [Experimental(DiagnosticIds.Experiments.Telemetry)]
    public IEnumerable<Func<FakeLogRecord, bool>> CustomFilters { get; set; } = Array.Empty<Func<FakeLogRecord, bool>>();
```

```suggestion
```

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollector.cs
Demo30: The inverse was proposed in this comment: https://github.com/dotnet/extensions/pull/5848#discussion_r1950240397

I think both versions are possible, but I suggested going with the original proposal as that seems to be more widely used (e.g.: `.Where` in LINQ, `.filter` in javascript, `stream.filter` in Java) and also consistent with the behavior of the other filters in the same options type as mentioned by @Piedone.

Piedone: I guess indeed. My concern is really only about keeping it consistent with the other filter options here, what the current implementation is.

evgenyfedorov2:
@Demo30 what are the next steps in this PR?

Demo30:
> @Demo30 what are the next steps in this PR?

unfortunately same as 6 months ago, waiting to get scheduled for API review. At the moment, I'm closer to the top, but may change as it has already in the past. Any suggestion on how to move forward more effectively?
image

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: this alternative design was mentioned during [api review](https://github.com/dotnet/extensions/issues/5615#issuecomment-3074621717), but the proposal passed with current semantics

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
```diff
@@ -34,6 +35,17 @@ public class FakeLogCollectorOptions
...
 
+    /// <summary>
+    /// Gets or sets custom filter for which records are collected.
+    /// </summary>
+    /// <value>The default is <see langword="null" />.</value>
+    /// <remarks>
+    /// Defaults to <see langword="null" /> which doesn't apply any additional filter to the records.
+    /// If not empty, only records for which the filter function returns <see langword="true" /> will be collected by the fake logger.
+    /// </remarks>
+    [Experimental(DiagnosticIds.Experiments.Telemetry)]
```
Demo30, line 46: Check experimental tag, during [api review](https://github.com/dotnet/extensions/issues/5615#issuecomment-3074621717) it was mentioned that the tag should not be necessary

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: The proposal passed [api review](https://github.com/dotnet/extensions/issues/5615#issuecomment-3074621717) with nullable type defaulting to null. I believe we can close this thread

test/Libraries/Microsoft.Extensions.Diagnostics.Testing.Tests/Logging/FakeLoggerTests.cs
Demo30: changed according to suggestion in tests, closing this thread

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: The wording was changed to `/// If not empty, only records for which the filter function returns <see langword="true" /> will be collected by the fake logger.` to match existing `FilteredLevels` and `FilteredCategories` descriptions.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollector.cs
Demo30: This thread has become obfuscated by different topics being discussed. The potential discussion regarding semantics of the filter will be followed in the other thread that also comments on it: https://github.com/dotnet/extensions/pull/5848#discussion_r1950240397 Closing this one

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: Closing this as it is not apparent what is being discussed here. The reversed semantics of custom filter is included here which is being discussed in another thread: https://github.com/dotnet/extensions/pull/5848#discussion_r1950240397

test/Libraries/Microsoft.Extensions.Diagnostics.Testing.Tests/Logging/FakeLoggerTests.cs
```diff
@@ -6,8 +6,6 @@
 using System.Globalization;
 using System.Linq;
 using System.Threading;
```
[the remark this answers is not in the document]
Demo30: no true, does not cause compilation error due to matching parent nested namespace

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
```diff
@@ -3,7 +3,8 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.Extensions.Logging;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.DiagnosticIds;
```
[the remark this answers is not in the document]
Demo30: no true, does not cause compilation error due to matching parent nested namespace

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: the automated analyzers logic requires it though; (different arbitrary instance where experimental attribute was added in similar case: https://github.com/dotnet/extensions/pull/5798/files#:~:text=%5BExperimental(diagnosticId%3A%20DiagnosticIds.Experiments.HealthChecks%2C%20UrlFormat%20%3D%20DiagnosticIds.UrlFormat)%5D) image

Demo30: ➕ it was discussed whether "telemetry" is the right category

dariusclay:
API proposal shows approved, is this ready to go?

dariusclay (approved):
Overall, looks good. I'm on the edge between the naming of "CustomFilter" vs "FilteredRecords" though

Demo30:
> API proposal shows approved, is this ready to go?

from the API Review point of view yes, the api has been discussed on review and approved. We have some comments open here and after resolving those I think we're good to merge

Demo30:
> Overall, looks good. I'm on the edge between the naming of "CustomFilter" vs "FilteredRecords" though

I was thinking simplifying it to just `Filter`. I don't like `FilteredRecords` very much even though it seems to follow the other two filters, but in those cases: filtered levels is a set of levels, filtered categories is a list of categories -> for "filtered records" i'd expect a set of records, but this is a predicate so I personally would find that naming confusing because i'd expect to get/set some records here

iliar-turdushev (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: discussed with @dariusclay, the experimental attribute belongs here (under the Telemetry category) for now as this is newly introduced symbol to the class and will be removed later when confirmed by the usage. Closing this comment

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: I believe we are in agreement in this thread and it can be closed. Keeping this open for next few days to give room for additional comments @RussKie. Otherwise the PR is green, approved on API review and approved by the maintainers so we'll continue by merging it.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: Closing this thread as it also ties to this discussion here: https://github.com/dotnet/extensions/pull/5848#discussion_r1950240397 if by any chance we decided to change the semantics of the predicate, we would also update the summary.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
Demo30: We are not in agreement in this thread, however the current design has passed the API review in this form and has been approved by other maintainers. The argument for this design is that is follows common conventions and also the other two existing filter properties.

Keeping this open for next few days to give room for additional comments @RussKie. Otherwise the PR is green, approved on API review and approved by the maintainers so we'll continue by merging it.

Demo30: Closing this thread. We are proceeding with the merge.

src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Logging/FakeLogCollectorOptions.cs
RussKie: The comment was about the wording in the xmldocs

PR #5849, opened 2025-02-06
Use UseZeroToOneRangeForMetrics in Health Checks

evgenyfedorov2:
Use `UseZeroToOneRangeForMetrics` property in `Microsoft.Extensions.Diagnostics.HealthChecks.ResourceUtilization `package to determine the cpu and metric values range [0;1] vs. [0;100] instead of relying purely on operating system which became inaccurate after the bug fix https://github.com/dotnet/extensions/issues/5472.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5849)


amadeuszl (approved):
All good

PR #5852, opened 2025-02-06
Add AsChatClient for OpenAI's AssistantClient

stephentoub:
- Adds ChatOptions/ChatCompletion/StreamingChatCompletionUpdate.ChatThreadId.
- Adds an AsChatClient extension method for creating an IChatClient from an AssistantClient.
- Updates FunctionInvokingChatClient to have KeepFunctionCallingMessages
  be false by default, and to handle ChatThreadId.
- Fixes handling of ChatCompletion.Usage in ToChatCompletion{Async}.


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -71,6 +74,7 @@ public virtual ChatOptions Clone()
     {
         ChatOptions options = new()
         {
+            ConversationId = ConversationId,
```
stephentoub, line 77: Open to other naming suggestions

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+                {
+                    var assistant = await _assistantClient.CreateAssistantAsync(model, cancellationToken: cancellationToken).ConfigureAwait(false);
+                    assistantId = assistant.Value.Id;
+                }
+                else
+                {
+                    Throw.ArgumentException(nameof(options), "No assistant was available, and no model was supplied for which to create an assistant.");
+                }
+            }
+        }
```
stephentoub, line 112: This behavior may not be intuitive. Not sure what would be better.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        {
+            var creationOptions = new ThreadCreationOptions();
+            foreach (var message in runOptions.AdditionalMessages)
+            {
+                creationOptions.InitialMessages.Add(message);
+            }
+
+            runOptions.AdditionalMessages.Clear();
+
+            updates = _assistantClient.CreateThreadAndRunStreamingAsync(assistantId, creationOptions, runOptions, cancellationToken: cancellationToken);
```
stephentoub, line 146: Similarly here. It's creating a new thread and storing the ID back into ChatOptions. Not sure what would be better.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        if (instructions is not null)
+        {
+            runOptions.AdditionalInstructions = instructions.ToString();
+        }
+
+        // Now that we've gathered everything we need from the chat messages, clear it out. We don't want
+        // the same messages resubmitted in the next request.
+        chatMessages.Clear();
```
stephentoub, line 375: This might also be unexpected, but not sure how else to mature the model work well.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
SteveSandersonMS: Looks like OpenAI's term is "thread ID", but perhaps you didn't use that because of the clash with .NET's unrelated notion of thread IDs

From a quick look, I don't see any equivalent in Gemini, Mistral, or Anthropic. Context.ai does have [its own equivalent and calls it a "thread"](https://docs.context.ai/product-analytics/api-ingestion-methods/thread-conversation) (but not this is an eval too, not an LLM itself).

Is there any prior art for the term "conversation ID"? If not and we wanted to avoid making up novel terminology, we could go with either `ThreadId` or `ChatThreadId` to avoid ambiguity with `System.Threading`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionCallContent.cs
```diff
@@ -45,6 +45,13 @@ public FunctionCallContent(string callId, string name, IDictionary<string, objec
     /// </summary>
     public IDictionary<string, object?>? Arguments { get; set; }
 
+    /// <summary>Gets or sets an additional context object specific to the provider that created the content.</summary>
+    /// <remarks>
+    /// If not <see langword="null"/>, the value of this property should be stored into any <see cref="FunctionResultContent"/>
+    /// created to represent the result of this function call.
+    /// </remarks>
+    public object? Context { get; set; }
```
SteveSandersonMS, line 53: This is a bit weird because you might ask why not define `CallId` as type `object`, and then the provider can bundle any amount of ID/state in there that it likes, and the FRC creator would not need to write any extra code since they already know they have to provide the same ID on the result.

But then the call ID would typically get lost when round-tripping through serialization, and perhaps that's a problem we need to avoid, and is solved by having `CallId` be `string`. So my remaining question is: is it OK for `Context` to be lost when round-tripping through serialization? And if so, how can we communicate that to someone who's dealing with chat history serialization?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+            serviceKey is not null ? null :
+            serviceType == typeof(AssistantClient) ? _assistantClient :
+            serviceType.IsInstanceOfType(this) ? this :
+            null;
+    }
+
+    /// <inheritdoc />
+    public Task<ChatCompletion> CompleteAsync(
+        IList<ChatMessage> chatMessages, ChatOptions? options = null, CancellationToken cancellationToken = default) =>
+        CompleteStreamingAsync(chatMessages, options, cancellationToken).ToChatCompletionAsync(coalesceContent: true, cancellationToken);
```
SteveSandersonMS, line 81: Presumably we could change this to use the nonstreaming APIs in the future if we wanted, and wouldn't consider that a breaking change.

Just asking because historically there have been behavioral/capability differences between streaming and nonstreaming, leading to practical reasons why the nonstreaming calls need to use the nonstreaming APIs.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        IList<ChatMessage> chatMessages, ChatOptions? options = null, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        // Extract necessary state from chatMessages and options.
+        (string? assistantId, string? threadId) = GetIds(options);
+        (RunCreationOptions runOptions, List<FunctionResultContent>? toolResults) = CreateRunOptions(chatMessages, options);
+
+        // Ensure we have an assistant, finding one if none was specified, and creating one if none was available.
+        if (assistantId is null)
+        {
+            await foreach (var assistant in _assistantClient.GetAssistantsAsync(cancellationToken: cancellationToken).ConfigureAwait(false))
```
SteveSandersonMS, line 94: This doesn't feel right - it's just picking an arbitrary assistant. But if I understand correctly, assistants have existing configuration (e.g., "instructions" (system prompt), tools, vector store, files). So in general a developer wouldn't do well to pick one arbitrarily, since in any practical case they need to have one correctly configured for their scenario.

I think we should make `assistantId` be a required parameter.

SteveSandersonMS: The alternative would be to drop lines 94-98 so that, if `assistantId` is not specified, we always create a new one. But I don't think that's necessarily what people want either as then the OpenAI account will end up with a vast number of randomly named assistants without a good way for the developer to clean them up later.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
SteveSandersonMS: For now I'd be happy to bypass the problem and make `assistantId` be required, so that developers understand they have to think about how the assistant is defined and configured (e.g., giving it instructions, tools, etc.).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
SteveSandersonMS: It is a bit quirky. Maybe I've forgotten but is this the first case where we'd mutate `ChatOptions` automatically? It might not compose well with middleware since middleware may use `options.Clone()`.

The alternative that follows our existing patterns would be to put the ID onto the resulting `ChatCompletion`/`StreamingChatCompletionUpdate`. Developers would then assign this to their `ChatOptions` manually, which is the equivalent to appending the resulting message to a `List<ChatMessage>`.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        if (threadId is null)
+        {
+            if (toolResults is not null)
+            {
+                Throw.ArgumentException(nameof(chatMessages), "No thread ID was provided, but chat messages includes tool results.");
+            }
+            else if (options is null)
+            {
+                Throw.ArgumentException(nameof(options), "No thread ID was provided, and no options was supplied to store a new thread's ID into.");
+            }
```
SteveSandersonMS, line 124: Perhaps we could avoid this restriction as per https://github.com/dotnet/extensions/pull/5852/files#r1946310023

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -9,6 +9,9 @@ namespace Microsoft.Extensions.AI;
 /// <summary>Represents the options for a chat request.</summary>
 public class ChatOptions
 {
+    /// <summary>Gets or sets an optional identifier used to associate a request with an existing conversation.</summary>
+    public string? ConversationId { get; set; }
```
eiriktsarpalis, line 13: Per our recent offline conversation, this feels to me like information that could be encapsulated by a particular `IChatClient` implementation. I know that our approach so far has been to offload all context minus connection strings to the message log and options, but it makes me wonder if the new assistants API is signal that we might want to consider encapsulating information pertaining to a specific session/thread/conversation.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+    /// <param name="threadId">The ID of the thread to use.</param>
+    public OpenAIAssistantClient(AssistantClient assistantClient, string? assistantId, string? threadId)
+    {
+        _assistantClient = Throw.IfNull(assistantClient);
+
+        _assistantId = assistantId;
+        _threadId = threadId;
+
+        // https://github.com/openai/openai-dotnet/issues/215
+        // The endpoint isn't currently exposed, so use reflection to get at them, temporarily. Once packages
```
eiriktsarpalis, line 54: 😭

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+    void IDisposable.Dispose()
+    {
+        // Nothing to dispose. Implementation required for the IChatClient interface.
+    }
+
+    /// <summary>Gets the thread ID and assistant ID to target.</summary>
+    private (string? AssistantId, string? ThreadId) GetIds(ChatOptions? options)
+    {
+        // Get the assistant ID to use.
+        string? assistantId = options?.AdditionalProperties?.TryGetValue(AssistantIdKey, out string? s) is true ? s : _assistantId;
```
eiriktsarpalis, line 218: I would personally prefer it if there was only one way of specifying these id's, ideally from the constructor. The `ChatOptions` could have been deserialized from the wire, so in a hypothetical agent implementation these could be overridden by callers.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
stephentoub: That's what I had initially, I can go back to it.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
stephentoub: > The alternative that follows our existing patterns would be to put the ID onto the resulting ChatCompletion/StreamingChatCompletionUpdate

That doesn't work well with middleware like function invocation unless we somehow standardize that pattern.

As with assistant ID, I could also revert here back to what I had initially, which was requiring a thread I'd to be specified.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,379 @@
...
+    private readonly string? _assistantId;
+
+    /// <summary>The thread ID to use if none is supplied in <see cref="ChatOptions.ConversationId"/>.</summary>
+    private readonly string? _threadId;
+
+    /// <summary>Initializes a new instance of the <see cref="OpenAIAssistantClient"/> class for the specified <see cref="AssistantClient"/>.</summary>
+    /// <param name="assistantClient">The underlying client.</param>
+    /// <param name="assistantId">The ID of the assistant to use.</param>
+    /// <param name="threadId">The ID of the thread to use.</param>
+    public OpenAIAssistantClient(AssistantClient assistantClient, string? assistantId, string? threadId)
```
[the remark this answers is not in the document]
stephentoub: > Presuambly our argument for why we want to add this feature to MEAI.Abstractions right now is that we want to be sure we can extend to support this

Yes

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
stephentoub: We already allow things like model and dimensions to be specified in multiple places, with one baked into the instance and options used for ovverides or if one wasn't specified to construction. And in practice, there's nothing we can do to avoid that... anyone can define an IChatClient with any defaults built into it for things otherwise specified by ChatOptions.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -9,6 +9,9 @@ namespace Microsoft.Extensions.AI;
 /// <summary>Represents the options for a chat request.</summary>
 public class ChatOptions
 {
+    /// <summary>Gets or sets an optional identifier used to associate a request with an existing chat thread.</summary>
+    public string? ChatThreadId { get; set; }
```
stephentoub, line 13: I renamed this per the previous feedback. I also added the same named property to ChatCompletion and StreamingChatCompletionUpdate. Now when we create a thread behind the scenes, we have a way to propagate that back to the caller. When non-null, the caller can then put that thread ID into ChatOptions.ChatThreadId rather than maintaining the history manually, instead relying on Assistants doing it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/StreamingChatCompletionUpdateExtensions.cs
```diff
@@ -178,20 +193,6 @@ static void AddMessage(ChatCompletion completion, bool coalesceContent, KeyValue
...
-                {
-                    if (content is UsageContent c)
-                    {
-                        completion.Usage = c.Details;
-                        entry.Value.Contents = entry.Value.Contents.ToList();
-                        _ = entry.Value.Contents.Remove(c);
-                        break;
-                    }
-                }
-            }
```
stephentoub, line 194: This is only a semi-related bug fix. Now that we're tracking all token counts, we should be doing so across all messages when creating a ChatCompletion and not just for a single usage from a single choice.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,350 @@
...
+    /// <param name="assistantClient">The underlying client.</param>
+    /// <param name="assistantId">The ID of the assistant to use.</param>
+    /// <param name="threadId">
+    /// The ID of the thread to use. If not supplied here, it should be supplied per request in <see cref="ChatOptions.ChatThreadId"/>.
+    /// If none is supplied, a new thread will be created for a request.
+    /// </param>
+    public OpenAIAssistantClient(AssistantClient assistantClient, string assistantId, string? threadId)
+    {
+        _assistantClient = Throw.IfNull(assistantClient);
+        _assistantId = Throw.IfNull(assistantId);
```
stephentoub, line 49: I made the assistant ID required, per the previous feedback.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,350 @@
...
+                    chatMessage.Role == ChatRole.Assistant ? MessageRole.Assistant : MessageRole.User,
+                    messageContents));
+            }
+        }
+
+        if (instructions is not null)
+        {
+            runOptions.AdditionalInstructions = instructions.ToString();
+        }
+
```
stephentoub, line 309: I no longer clear the user's history; the need for that is gone with the new semantics around {Streaming}ChatCompletion{Update}.ChatThreadId.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,350 @@
...
+                            OutputTokenCount = usage.OutputTokenCount,
+                            TotalTokenCount = usage.TotalTokenCount,
+                        }));
+                    }
+
+                    if (ru is RequiredActionUpdate rau && rau.ToolCallId is string toolCallId && rau.FunctionName is string functionName)
+                    {
+                        ruUpdate.Contents.Add(
+                            new FunctionCallContent(
+                                JsonSerializer.Serialize(new[] { ru.Value.Id, toolCallId }, OpenAIJsonContext.Default.StringArray!),
```
stephentoub, line 160: I got rid of that additional Context property and am following the suggested approach of putting the run ID into the tool call ID.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -155,16 +155,22 @@ public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = nul
...
     /// as to whether function calling messages are kept during an in-flight request.
     /// </para>
+    /// <para>
+    /// If the underlying <see cref="IChatClient"/> responds with <see cref="ChatCompletion.ChatThreadId"/>
+    /// set to a non-<see langword="null"/> value, this property may be ignored and behaves as if it is
+    /// <see langword="false"/>, with any such intermediate messages not stored in the messages list.
+    /// </para>
     /// </remarks>
-    public bool KeepFunctionCallingMessages { get; set; } = true;
+    public bool KeepFunctionCallingMessages { get; set; }
```
stephentoub, line 173: I decided to swap the default for KeepFunctionCallingMessages. Based on all usage we've seen thus far, it feels like the common case is really that these intermediate messages are an implementation detail that someone doesn't care about, but can opt-in to still getting if they do. The default of false feels better as well in this ChatThreadId / Assistants context, where those intermediate messages are even more of an implementation detail and aren't propagated back as part of the original list being mutated.

SteveSandersonMS: Even if developers don't realise they want to retain FCCs/FRCs, dropping them does make a behavioral difference - it means the LLM will become more prone to (a) repeating calls and (b) hallucinations. When the LLM sees a chat history like:

 * [user] What's the weather in A
 * [tool] `{ description: "sunny" }`
 * [assistant] It's sunny in A
 * [user] What about in B

... it's likely to repeat the pattern of doing a tool call. But if the tool call is absent from the history, it's more likely to repeat the pattern of just stating some weather without justification, since the chat history suggests the LLM should just state an example of weather without doing a tool call.

Secondly if you're using the FCCs/FRCs to trigger a UI display of what tool calls have taken place, for example as in the RAG/chat template we're currently finalizing, dropping that info makes it way harder because you need some extra sidechannel to send back the info, and then you get into race condition issues where it's tough to ensure you're showing the call at the right position in the chat in the UI.

I know it's tricky that in the assistants case you have the second problem regardless, but I think in the traditional chat case in a large percentage of cases (likely the substantial majority) people would be wise to enable retention, so it feels like having it as a default would still be beneficial.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -0,0 +1,350 @@
...
+        {
+            Throw.ArgumentException(nameof(chatMessages), "No thread ID was provided, but chat messages includes tool results.");
+        }
+
+        // Get the updates to process from the assistant. If we have any tool results, this means submitting those and ignoring
+        // our runOptions. Otherwise, create a run, and a thread if we don't have one.
+        IAsyncEnumerable<StreamingUpdate> updates;
+        if (GetRunId(toolResults, out List<ToolOutput>? toolOutputs) is string existingRunId)
+        {
+            updates = _assistantClient.SubmitToolOutputsToRunStreamingAsync(threadId, existingRunId, toolOutputs, cancellationToken);
```
SteveSandersonMS, line 94: Should we also throw if, at this point, we see there are any message contents *other* than FRCs?

Am I right to think that if there are, that additional content would just get lost at the point? (Not 100% certain if I'm visualising the flow correctly)

SteveSandersonMS (approved):
Looks good!

* Personally I'd be more inclined to leave `KeepFunctionCallingMessages` on by default as per https://github.com/dotnet/extensions/pull/5852/files?w=1#r1948821474 but I know this is a subjective call
* Should there be some additional tests covering the `FunctionInvokingChatClient` behaviors around its detection/handling of `ChatThreadId`?
* Technically also we have a load of integration tests for `OpenAIChatClient` that don't have an equivalent in the new `OpenAIAssistantClient`. However I know it's a huge pain to create them so will leave it with you to decide if there's enough value in doing so.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: > it means the LLM will become more prone to (a) repeating calls and (b) hallucinations.

Ok, that's a compelling reason to keep the default as true. I'll change it back.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
stephentoub: > Am I right to think that if there are, that additional content would just get lost at this point?

Yes

> Should we also throw if, at this point, we see there are any message contents other than FRCs?

I'm a little hesitant to do that. In general we've been operating under a scheme where unrecognized content / options are just ignored rather than producing an error.

stephentoub:
> Should there be some additional tests covering the FunctionInvokingChatClient behaviors around its detection/handling of ChatThreadId?

Yes, added. Thanks.

PR #5853, opened 2025-02-07
Automate chat template JS dependency updates

MackinnonBuck:
# Automate chat template JS dependency updates

Following are the main changes in this PR:
1. Add a `package.json` to specify project template dependencies and how they should be copied into template content
    * Note that this is only a development-time dependency and doesn't ship with the templates. That is, we're not adding back a dependency on NPM for the shipping templates.
2. Add a `copyDependency.mjs` script that carries out the actual dependency copying.
    * It also updates the package version in the README file
3. Reduce the number of assets required for PDF viewing (more info below).

## PDF viewer changes

The `pdfjs` dependency was trickier than the others, because its NPM package ([`pdfjs-dist`](https://www.npmjs.com/package/pdfjs-dist)) does _not_ include the built-in PDF viewer that was being utilized by the chat template. There are numerous issues in the pdf.js GitHub repo asking for the viewer to be included, but the maintainers have [stated](https://github.com/mozilla/pdf.js/wiki/Setup-pdf.js-in-a-website#official-releases) that they don't want people to use the PDF viewer "as-is" in actual code:

> The viewer is built on the display layer and is the UI for the PDF viewer in Firefox and the other browser extensions within the project. It can be a good starting point for building your own viewer. ***However, we do ask if you plan to embed the viewer in your own site, that it not just be an unmodified version. Please re-skin it or build upon it.***

Given this statement and the fact that it was intentionally made difficult to use the built-in PDF viewer, it's probably best to find an alternative.

Thankfully, it's pretty straightforward to recreate a super minimal PDF viewer using pdf.js directly. I've added two assets to the `wwwroot` folder in the template that define a custom viewer:
* `pdf_viewer.html`
* `pdf_viewer.js`

These assets won't get updated automatically, but they probably won't need to be anyway unless there's a significant breaking change in pdf.js.

It's pretty minimal, but I don't think we need anything more advanced for a template:
pdf-viewer

One advantage to this is that we can significantly reduce the number of assets required for PDF viewing (all the images except for the loading icon can be removed).

In the future, we could even replace `pdf_viewer.html` with a Blazor component that displays the citation in a modal dialog. But I'm hoping we would address that in a follow-up.

## Future work

This PR does not add a CI pipeline to update template dependencies. It's still "manual" in that you have to clone the repo and run these two commands from the `src/ProjectTemplates` folder:
```sh
npm install
npm run copy-dependencies
```

However, this is how it works in the ASP.NET Core repo as well, and I'd imagine it saves about 90% of the effort required to update dependencies.

This PR also doesn't do anything to automate .NET dependency updates. If we think this would be helpful, I'll address it in a follow-up PR to keep this one focused.

Fixes https://github.com/dotnet/ai-private-planning/issues/276
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5853)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/wwwroot/pdf_viewer.html
```diff
@@ -0,0 +1,34 @@
...
+
+  <script src="./pdfjs-dist/dist/build/pdf.mjs" type="module"></script>
+  <script src="./pdfjs-dist/dist/web/pdf_viewer.mjs" type="module"></script>
+</head>
+<body tabindex="1">
+  <div id="viewerContainer">
+    <div id="viewer" class="pdfViewer"></div>
+  </div>
+
+  <script src="pdf_viewer.mjs" type="module"></script>
```
SteveSandersonMS, line 32: There are two JS files both called `pdf_viewer.mjs`? (This one and the one on L25)

Assuming these aren't duplicates, would it work to give them different names? It could get confusing, e.g., when working on this or if people are trying to upgrade or report issues otherwise.

src/ProjectTemplates/package.json
```diff
@@ -0,0 +1,33 @@
...
+      "chat": "./Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/wwwroot"
+    }
+  },
+  "scripts": {
+    "copy-dependency": "node copyDependency.mjs",
+    "copy-dependencies": "npm run copy-dompurify && npm run copy-marked && npm run copy-tailwindcss && npm run copy-pdfjs",
+    "copy-dompurify": "npm run copy-dependency -- --name dompurify --dest-root chat --src-root dist --files purify.es.mjs;purify.es.mjs.map",
+    "copy-marked": "npm run copy-dependency -- --name marked --dest-root chat --src-root lib --files marked.esm.js;marked.esm.js.map",
+    "copy-tailwindcss": "npm run copy-dependency -- --name tailwindcss --dest-root chat --files preflight.css",
+    "copy-pdfjs": "npm run copy-dependency -- --name pdfjs-dist --dest-root chat --files build/pdf.mjs;build/pdf.mjs.map;build/pdf.worker.mjs;build/pdf.worker.mjs.map;web/pdf_viewer.css;web/pdf_viewer.mjs;web/pdf_viewer.mjs.map;web/images/loading-icon.gif"
```
SteveSandersonMS, line 17: How would you feel about omitting all the `.map` files? I don't think that we need to do anything for debugging these files since they are effectively vendor binaries. Would be nice just to minimize the number of files people have to make sense of.

Vaguely related, what about putting these vendor-type dependencies into `wwwroot/lib/*` instead of `wwwroot/*`? Just to make it easier to distinguish them from the developer's own folder structure of custom wwwroot content.

SteveSandersonMS:
> Thankfully, it's pretty straightforward to recreate a super minimal PDF viewer using pdf.js directly. I've added two assets to the wwwroot folder in the template that define a custom viewer:

This is wonderful. I love it. Much better than having that massive dump of files we had before. Brilliant.

> This PR also doesn't do anything to automate .NET dependency updates. If we think this would be helpful, I'll address it in a follow-up PR to keep this one focused.

I definitely think that would that would be helpful, even more so than the JS files since the .NET deps are going to rev continually. Hopefully we could use an equivalent technique to what we do in dotnet/aspnetcore so we automatically use whatever version this host repo itself depends on, so the automatic dependency flow keeps things up-to-date for us: https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/GenerateContent.targets

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/wwwroot/pdf_viewer.html
MackinnonBuck: Good catch, I'll change their names.

src/ProjectTemplates/package.json
MackinnonBuck: > How would you feel about omitting all the .map files?

I was thinking about doing this as well. If we do that, we could probably also opt to use minified JS when available.

> Vaguely related, what about putting these vendor-type dependencies into wwwroot/lib/* instead of wwwroot/*?

Agreed. I'll change that too.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/wwwroot/pdf_viewer/viewer.mjs
```diff
@@ -0,0 +1,62 @@
...
+// Create the PDF viewer.
+const pdfViewer = new PDFViewer({
+  container,
+  eventBus,
+  linkService: pdfLinkService,
+  findController: pdfFindController,
+});
+pdfLinkService.setViewer(pdfViewer);
+
+// Allow navigation to a citaiton from the URL hash.
```
SteveSandersonMS, line 36:
```suggestion
// Allow navigation to a citation from the URL hash.
```

PR #5855, opened 2025-02-07
AI Templates - Fixes for Ollama and OpenAI scenarios

jmatthiesen:
Fixes several issues for the different combinations of environment choices:
* Fixed Azure AI Search integration by adding a check in the data ingestor to see if there are documents to delete before trying to delete them.
* Adjusted the main chat prompt so that citations work with Ollama/Llama3.2 after recent changes - verified it works for all other supported AI model providers
* Fixed using statements when pairing OpenAI with Azure AI Search (was missing a `using Azure` statement)
* Removed the Attach Media button after a recent decision to not include that for now.
* Fixed the conditionals in Chat.razor to include the proper code for Ollama.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5855)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Pages/Chat/Chat.razor
```diff
@@ -52,7 +56,7 @@
         chatSuggestions?.Clear();
         await chatInput!.FocusAsync();
 
-#if (IsOllama)
+@*#if (IsOllama)
```
jmatthiesen, line 59: This looks wrong, but it is the conditional syntax used by the templating engine for .razor files.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Services/Ingestion/DataIngestor.cs
```diff
@@ -40,8 +40,11 @@ public async Task IngestDataAsync(IIngestionSource source)
         {
             logger.LogInformation("Processing {file}", modifiedDoc.Id);
 
-            await vectorCollection.DeleteBatchAsync(modifiedDoc.Records.Select(r => r.Id));
-
+            if (modifiedDoc.Records.Count > 0)
```
jmatthiesen, line 43: Without this, Azure AI Search integration breaks.

SteveSandersonMS: Great, thanks.

The previous version of this check seemed to have the braces in a different place and was suppressing the `CreateRecordsForDocumentAsync` call as well, but maybe something weird happened in the Git merge.

In any case this version of the check looks totally safe!

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Pages/Chat/Chat.razor
```diff
@@ -23,9 +23,13 @@
...
-        Use the search tool to find relevant information. When you do this, end your
-        reply with citations in the special format, always formatted as XML:
-        <citation filename='string' page_number='number'>exact quote here</citation>.
+         Use the search tool to find relevant information. When you do this, end your
+        reply with citations in the special XML format:
+
+        <citation filename='string' page_number='number'>exact quote here</citation>
+
+         Always include the citation in your response if there are results.
+
```
SteveSandersonMS, line 32: Nbd but something quirky about the indentation:

```suggestion
        Use the search tool to find relevant information. When you do this, end your
        reply with citations in the special XML format:

        <citation filename='string' page_number='number'>exact quote here</citation>

        Always include the citation in your response if there are results.
```

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Pages/Chat/Chat.razor
```diff
@@ -23,9 +23,13 @@
...
 
-        Use the search tool to find relevant information. When you do this, end your
-        reply with citations in the special format, always formatted as XML:
-        <citation filename='string' page_number='number'>exact quote here</citation>.
+         Use the search tool to find relevant information. When you do this, end your
+        reply with citations in the special XML format:
+
+        <citation filename='string' page_number='number'>exact quote here</citation>
+
+         Always include the citation in your response if there are results.
```
SteveSandersonMS, line 31: Nbd but something quirky about the indentation. Not sure if it's a tabs/spaces thing:

```suggestion
        Use the search tool to find relevant information. When you do this, end your
        reply with citations in the special XML format:

        <citation filename='string' page_number='number'>exact quote here</citation>

        Always include the citation in your response if there are results.
```

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Pages/Chat/Chat.razor
SteveSandersonMS: Actually the syntax here isn't quite right because it will break the application if you run it directly from sources here. Note that in VS it's displayed like this:

image

That is, **both** branches of the if/else are greyed out, because as far as Razor syntax is concerned, this is all just one big comment block. So if you run the application like this, neither branch will execute and the app won't work. This will block us from being able to work on the application code directly.

I think what we need is the following syntax instead:

image

Now Razor itself sees the Ollama part of the code as being a comment, and the rest as being real code. So the app will still work if we run it directly from sources.

I *think* at least that the `dotnet new` templating system will also understand this as an if/else and emit the desired code in both cases.

jmatthiesen: Excellent - thanks for the details and verification! I hadn't thought about running these from source, but only ever installing and using the template. Nice improvement!

1
MackinnonBuck, line 1: Looks like this file might have been added by mistake?

jmatthiesen: Oops! :)

SteveSandersonMS (approved):

PR #5860, opened 2025-02-07
Address M.E.AI API feedback

stephentoub:
- Remove DataContent.ContainsData. While ContainsData could be used to avoid lazily-instantiating a `ReadOnlyMemory<byte>` from a data URI, in all cases examined ContainsData was being used to guard accessing Data, in which case Data.HasValue is sufficient.
- Make ToolMode optional. We'd previously said that an implementation could use null to imply None if it wanted, but with this being optional we'd want null to be the same as auto, so I added in an explicit null options. That matches as well with what various other client libs do.
- Remove IChatClient/IEmbeddingGenerator.Metadata. GetService can be used instead, reducing what an implementer must implement and what's exposed to a consumer.
- Add Complete{Streaming}Async for a single message. We have such accelerators in other places but not here, and it provides a natural grow-up from string to ChatMessage to `IList<ChatMessage>`.
- Change UsageDetails.XxTokenCount properties from int? to long?. The AdditionalCounts is already based on longs, and in theory the token counts could exceed int, especially in a situation where UsageData is being used to sum many other call data.
- Rename GenerateEmbeddingVectorAsync's TEmbedding to TEmbeddingElement. Everywhere else TEmbedding is used where it represents an Embedding, but here it represents a numerical element in an embedding vector.
- Remove setters on FunctionCall/ResultContent for required ctor parameters. Having those setters goes against .NET design guidelines.
- Remove FunctionResultContent.Name. It's unnecessary and is actually causing us to do more work in places.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5860)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -42,7 +42,25 @@ public static Task<ChatCompletion> CompleteAsync(
...
+
+    /// <summary>Sends a chat message to the model and returns the response messages.</summary>
+    /// <param name="client">The chat client.</param>
+    /// <param name="chatMessage">The chat message to send.</param>
+    /// <param name="options">The chat options to configure the request.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The response messages generated by the client.</returns>
+    public static Task<ChatCompletion> CompleteAsync(
+        this IChatClient client,
+        ChatMessage chatMessage,
```
eiriktsarpalis, line 56: This new set of extensions can only be used in cases where the chat history consists of only one message. How common is such a scenario? Could it lead callers into incorrectly assuming that they're _appending_ a chat message to a log that is being encapsulated by the client?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -61,16 +64,14 @@ public JsonSerializerOptions ToolCallJsonSerializerOptions
     }
 
     /// <inheritdoc />
-    public ChatClientMetadata Metadata { get; }
```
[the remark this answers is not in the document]
stephentoub: I'd rather we expose it as an extension property once those are possible.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
stephentoub: > This new set of extensions can only be used in cases where the chat history consists of only that one message. How common is such a scenario?

It's pretty common, for at least:
- Getting started scenarios, where you want to send a single input
- Non-chat scenarios, where you're sending input to have the LLM give you back a response as part of the logic of you're program.
- IChatClient implementations where the state is stored server-side, ala assistants

It's no different from the existing string-based overloads that are helpful in similar circumstances but that are limited to TextContent; these overloads allow you to customize beyond that.

PR #5864, opened 2025-02-10
Generalise how package version isn't stabilized for dev/preview stages

RussKie:
Follow up on #5632
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5864)

PR #5873, opened 2025-02-11
Add MEAI.Evaluation libraries

shyamnamboodiripad:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5873)


shyamnamboodiripad:
FYI @stephentoub @luisquintanilla @RussKie @joperezr @peterwald

eng/Versions.props
```diff
@@ -78,6 +79,7 @@
   <PropertyGroup Label="Manual">
     <!-- Packages from dotnet/runtime -->
     <MicrosoftBclAsyncInterfacesLTSVersion>8.0.0</MicrosoftBclAsyncInterfacesLTSVersion>
+    <MicrosoftBclMemoryLTSVersion>9.0.1</MicrosoftBclMemoryLTSVersion>
```
RussKie, line 82: 🤔 I think this should be 8.0.0, otherwise there's a risk of pulling non-LTS dependencies.

scripts/ConfigureEvaluationTests.ps1
```diff
@@ -0,0 +1,45 @@
+param (
```
RussKie, line 1: It'd be great to have explanations for this file and the input args similar to other ps1s in this folder.

Also, if this file is M.E.AI.Eval test specific, then would it make more sense to locate it in the project's folder?

test/Libraries/Microsoft.Extensions.AI.Evaluation.Tests/Microsoft.Extensions.AI.Evaluation.Tests.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>Microsoft.Extensions.AI.Evaluation.Tests</RootNamespace>
```
RussKie, line 4: [nit] If the file name is the same as the project name, then this is redundant.

scripts/ConfigureEvaluationTests.ps1
shyamnamboodiripad: FYI @peterwald - perhaps we should move this down to the test folder per @RussKie's recommendation?

shyamnamboodiripad: Actually, since this is used by a couple of test projects, we will need to split out two copies if we want to move it down... Keeping it in the same location (and adding explanations) may work better from that perspective.

eng/Versions.props
shyamnamboodiripad: Ah good point - I will switch it to 8.*

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <see cref="ScenarioRun"/> via
+    /// <see cref="ReportingConfiguration.CreateScenarioRunAsync(string, string, IEnumerable{string}?, CancellationToken)"/>.
+    /// </summary>
+    public const string DefaultIterationName = "1";
+
+    /// <summary>
+    /// A <see cref="TimeSpan"/> that specifies the default amount of time that cached AI responses should survive
+    /// in the <see cref="IResponseCacheProvider"/>'s cache before they are considered expired and evicted.
+    /// </summary>
+    public static readonly TimeSpan DefaultTimeToLiveForCacheEntries = TimeSpan.FromDays(14);
```
stephentoub, line 32: This should be a property rather than a field.

Do the consts above need to be consts, in order to be used in only places consts are valid, or could they also be made properties?

Properties give you flexibility to change them in the future, both their values and the implementation. Consts don't provide any hooks for code to run and get baked into the call sites.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Formats/Html/HtmlReportWriter.cs
```diff
@@ -0,0 +1,94 @@
...
+        IEnumerable<ScenarioRunResult> scenarioRunResults,
+        CancellationToken cancellationToken = default)
+    {
+        var dataset =
+            new Dataset(
+                scenarioRunResults.ToList(),
+                createdAt: DateTime.UtcNow,
+                generatorVersion: Constants.Version);
+
+        using var stream = new FileStream(reportFilePath, FileMode.Create, FileAccess.Write, FileShare.None);
```
stephentoub, line 37: All writing is being done async; ideally this would also be opened with FileOptions.Asynchronous.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
```diff
@@ -0,0 +1,240 @@
...
+        {
+            IChatClient originalChatClient = chatConfiguration.ChatClient;
+
+            IEnumerable<string> cachingKeys = CachingKeys;
+
+            cachingKeys = [scenarioName, iterationName, .. cachingKeys];
+            if (additionalCachingKeys is not null)
+            {
+                cachingKeys = [.. cachingKeys, .. additionalCachingKeys];
+            }
```
stephentoub, line 183:
```suggestion
            cachingKeys = additionalCachingKeys is not null ?
                [scenarioName, iterationName, .. cachingKeys, .. additionalCachingKeys] :
                [scenarioName, iterationName, .. cachingKeys];
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
```diff
@@ -0,0 +1,240 @@
...
+            ExecutionName,
+            Evaluators,
+            ResultStore,
+            chatConfiguration,
+            EvaluationMetricInterpreter);
+    }
+
+    private static IEnumerable<string> GetCachingKeysForChatClient(IChatClient chatClient)
+    {
+        var metadata = chatClient.GetService<ChatClientMetadata>();
```
stephentoub, line 221: Nit:
```suggestion
        var metadata = chatClient.GetService<ChatClientMetadata>();

```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunExtensions.cs
```diff
@@ -0,0 +1,134 @@
...
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Evaluation.Reporting;
+
+/// <summary>
+/// Extension methods for <see cref="ScenarioRun"/>.
+/// </summary>
+public static class ScenarioRunExtensions
+{
+    private static IEnumerable<ChatMessage> EmptyMessages { get; } = [];
```
stephentoub, line 16: Do you need this, or can you just use `[]` wherever you need it?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Storage/DiskBasedResultStore.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        return default;
+    }
+
+    /// <inheritdoc/>
+    public async IAsyncEnumerable<string> GetLatestExecutionNamesAsync(
+        int? count = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        await Task.CompletedTask.ConfigureAwait(false);
```
stephentoub, line 184: Better to just suppress the warning rather than add unnecessary awaits

src/Libraries/Microsoft.Extensions.AI.Evaluation/CompositeEvaluator.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        IAsyncEnumerable<EvaluationResult> resultsStream =
+            EvaluateAndStreamResultsAsync(
+                messages,
+                modelResponse,
+                chatConfiguration,
+                additionalContext,
+                cancellationToken);
+
+        await foreach (EvaluationResult result in resultsStream.ConfigureAwait(false))
```
stephentoub, line 120:
```suggestion
        await foreach (EvaluationResult result in resultsStream.WithCancellation(cancellationToken).ConfigureAwait(false))
```
?

src/Libraries/Microsoft.Extensions.AI.Evaluation/CompositeEvaluator.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            cancellationToken.ThrowIfCancellationRequested();
```
stephentoub, line 122: I haven't been commenting on all of these ThrowIfCancellationRequested, but it's very unusual to see them used so heavily. Consider searching and deleting any that aren't truly necessary.

src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluationResult.cs
```diff
@@ -0,0 +1,141 @@
...
+        var metricsDictionary = new Dictionary<string, EvaluationMetric>();
+
+        foreach (EvaluationMetric metric in metrics)
+        {
+            if (metricsDictionary.ContainsKey(metric.Name))
+            {
+                Throw.ArgumentException(nameof(metrics), $"Cannot add multiple metrics with name '{metric.Name}'.");
+            }
+
+            metricsDictionary[metric.Name] = metric;
```
stephentoub, line 64: This can be TryAdd on .NET Core. Not sure it's worth ifdef'ing, though.

src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluatorExtensions.cs
```diff
@@ -0,0 +1,196 @@
...
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Evaluation;
+
+/// <summary>
+/// Extension methods for <see cref="IEvaluator"/>.
+/// </summary>
+public static class EvaluatorExtensions
+{
+    private static IEnumerable<ChatMessage> EmptyMessages { get; } = [];
```
stephentoub, line 16: Same question as earlier

src/Libraries/Microsoft.Extensions.AI.Evaluation/Utilities/TaskExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            foreach (Task<T> task in concurrentTasks)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                yield return await task.ConfigureAwait(false);
+            }
+        }
+        else
+        {
```
stephentoub, line 57: This is Task.WhenEach in .NET 9.

stephentoub:
I rebased this on main after merging API changes, and added a commit here that adapts to those changes (just a few tweaks needed).

eng/Versions.props
shyamnamboodiripad: @RussKie looks like `Microsoft.Bcl.Memory` is new. I only see 9.* versions here https://www.nuget.org/packages/Microsoft.Bcl.Memory#versions-body-tab. Any recommendations on how to fix this up?

scripts/ConfigureEvaluationTests.ps1
shyamnamboodiripad: I think we can leave it in this folder for now - we can figure out alternate location in the future if needed. (We will likely be making more changes to this and other scripts as part of enabling the above tests to run in CI.)

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
```diff
@@ -0,0 +1,28 @@
...
+        string storageRootPath = storageRootDir.FullName;
+        logger.LogInformation("Storage root path: {storageRootPath}", storageRootPath);
+        logger.LogInformation("Deleting expired cache entries...");
+
+        var cacheProvider = new DiskBasedResponseCacheProvider(storageRootPath);
+
+        await logger.ExecuteWithCatchAsync(
+            () => cacheProvider.DeleteExpiredCacheEntriesAsync(cancellationToken)).ConfigureAwait(false);
+
+        return 0;
```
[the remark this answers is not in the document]
shyamnamboodiripad: The `System.CommandLine` APIs should turn the returned `int` into the process exit code for the tool execution on the command line. Nothing in our tooling uses the process exit codes at the moment - however the user is free to use them in their scripts if needed.

As you point out we never explicitly return anything other than `0` for our existing 'commands'. (If an exception escapes this function, then `System.CommandLine` should return a non-zero code.) However, there will likely be cases / new commands in the future where we will want to explicitly return a non-zero exit code. (For example, one potential future command we have discussed is one where a user could specify some kind of failure criteria for the overall evaluation and running the tool with the specified criteria could return a non-zero exit code when the criteria is not met...)

I am not sure this is worth changing - but if you feel strongly, I can change the `Task<int>` to `Task` and based on a quick look `System.CommandLine` should preserve the same behavior for the existing commands...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Extensions.AI.Evaluation.Console.Commands;
+
+internal sealed class CleanCacheCommand(ILogger logger)
+{
+    internal async Task<int> InvokeAsync(DirectoryInfo storageRootDir, CancellationToken cancellationToken = default)
+    {
+        string storageRootPath = storageRootDir.FullName;
+        logger.LogInformation("Storage root path: {storageRootPath}", storageRootPath);
+        logger.LogInformation("Deleting expired cache entries...");
```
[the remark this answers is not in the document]
shyamnamboodiripad: I am not sure we need to. Currently, the above messages are meant to always be displayed on the command line for the dotnet tool, and the output is too small to worry about verbosity.

We could revisit and allow user more control over verbosity (or toggling logging on / off) in the future if the tool's output ends up becoming more verbose over time...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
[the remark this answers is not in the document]
stephentoub: It's fine. Thanks.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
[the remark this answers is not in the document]
stephentoub: Ok

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+        int lastN,
+        CancellationToken cancellationToken = default)
+    {
+        string storageRootPath = storageRootDir.FullName;
+        logger.LogInformation("Storage root path: {storageRootPath}", storageRootPath);
+
+        var resultStore = new DiskBasedResultStore(storageRootPath);
+
+        await logger.ExecuteWithCatchAsync(
+            async ValueTask () =>
```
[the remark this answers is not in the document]
shyamnamboodiripad: Yes, it does - we added overloads for both `Task` and `ValueTask` returning `Func`s.

We could potentially remove one or the other overload since the task allocations should not matter in the context of this command line tool. But not sure if it's worth changing for just the above reason.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Utilities/LoggerExtensions.cs
```diff
@@ -0,0 +1,274 @@
...
+        }
+        catch (Exception ex) when (!ex.IsCancellation() && logger.LogException(ex) && swallowUnhandledExceptions)
+        {
+            // Do nothing. The exception is logged in the when clause above.
+        }
+    }
+
+    internal static async Task ExecuteWithCatchAsync<TArgument>(
+        this ILogger logger,
+        Func<TArgument, Task> action,
```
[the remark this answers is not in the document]
shyamnamboodiripad: The helpers were intended to support both, but you are correct that all current usages just use the `ValueTask` overloads.

We could delete the Task overloads - it would just make it harder to call for `Func<Task>`-based signatures if and when we need to call those in the future... Can certainly change this if you feel strongly :)

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Utilities/LoggerExtensions.cs
```diff
@@ -0,0 +1,274 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Evaluation.Console.Utilities;
+
+internal static class LoggerExtensions
+{
+    internal static bool LogException(this ILogger logger, Exception exception)
+    {
+        logger.LogError("{errorMessage}", exception.ToString());
```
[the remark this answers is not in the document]
shyamnamboodiripad: Thanks - I missed that for some reason. However, looks like it requires passing both the exception and a message (and it was a bit unclear whether I can just pass `null` for the message if I just want to log the `exception.ToString()` value)...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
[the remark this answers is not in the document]
stephentoub: The problem with having both is that it is ambiguous, and you're forced to disambiguate by specifying the return type as you did. In my opinion, it's simpler and cleaner to just have the one.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Utilities/LoggerExtensions.cs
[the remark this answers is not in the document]
shyamnamboodiripad: Fixed - copilot tells me that I can pass `null` for the message :)

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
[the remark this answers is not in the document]
shyamnamboodiripad: (Resolving since there is another comment tracking the same issue in another file)

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+                    await foreach (string executionName in
+                        resultStore.GetLatestExecutionNamesAsync(lastN, cancellationToken).ConfigureAwait(false))
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+
+                        _ = toPreserve.Add(executionName);
+                    }
+
+                    await foreach (string executionName in
+                        resultStore.GetLatestExecutionNamesAsync(
```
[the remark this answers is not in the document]
shyamnamboodiripad: Ah in this case we want to preserve the `lastN` latest executions.

The default value of the `lastN` command line arg is `0` - but if the user specifies a value, we first collect the names of the latest `lastN` executions that are to be preserved first, and then call `GetLatestExecutionNamesAsync` (again - this time without the `lastN` argument) to fetch all executions. We then remove all executions that are not included in `toPreserve` set...

eng/Versions.props
RussKie: In this case I don't think we can use it in the LTS releases. You may need to copy the required functionality and secure it with `#if !NET9_0_OR_GREATER` pragmas.
There are examples under /src/Shared. If the code is needed by several projects, then it should go under Shared, otherwise it'd live in the project.

scripts/ConfigureEvaluationTests.ps1
RussKie: Sounds good

> It does indeed apply across several M.E.AI.Eval projects, so if we move it down, we'd need to find a good intermediate folder under Tests or something.

...or it could be linked ;)
E.g., `<Compile Include="$(MSBuildThisFileDirectory)\..\..\<path>\...ps1" LinkBase="scripts" />`

eng/Versions.props
shyamnamboodiripad: On closer look looks like we are no longer using `Microsoft.Bcl.Memory`. I've gone ahead and removed it.

src/Libraries/Microsoft.Extensions.AI.Evaluation/CompositeEvaluator.cs
shyamnamboodiripad: Thanks for catching this - I've audited and removed all redundant calls to `ThrowIfCancellartionRequested`.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Program.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Evaluation.Console;
+
+internal sealed class Program
+{
+    private const string Name = "Microsoft.Extensions.AI.Evaluation.Console";
+    private const string Banner = $"{Name} [{Constants.Version}]";
+
+#pragma warning disable EA0014 // Async methods should support cancellation.
```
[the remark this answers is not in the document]
shyamnamboodiripad: We have taken a similar approach in most cases (especially when there are multiple instances of a given warning in a file). In a few cases like above, the warnings seemed useful in general even though it does not apply for the specific instance, so I decided to restore it.

If we end up disabling some of these warnings globally, it would be great to go back and clean up the corresponding suppressions from the code / project files...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/RelevanceTruthAndCompletenessEvaluator.Prompts.cs
```diff
@@ -0,0 +1,224 @@
...
+
+namespace Microsoft.Extensions.AI.Evaluation.Quality;
+
+public partial class RelevanceTruthAndCompletenessEvaluator
+{
+    private static class Prompts
+    {
+        internal static string BuildEvaluationPrompt(string userQuery, string modelResponse, string history)
+        {
+#pragma warning disable S103 // Lines should not be too long
```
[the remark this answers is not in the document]
shyamnamboodiripad: Agreed - although I would prefer to fix up long lines that can otherwise be wrapped. We only had a handful of instances of this warning, and they were mostly restricted to files that contained long string literals (for prompts like above, or for error messages). So, I decided to just suppress the specific instances for now.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/RelevanceTruthAndCompletenessEvaluatorOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+/// <param name="includeReasoning">
+/// If <paramref name="includeReasoning"/> is set to <see langword="true"/>, this instructs the
+/// <see cref="RelevanceTruthAndCompletenessEvaluator"/> to include <see cref="EvaluationDiagnostic"/>s (with
+/// <see cref="EvaluationDiagnostic.Severity"/> set to <see cref="EvaluationDiagnosticSeverity.Informational"/>) as
+/// part of the returned <see cref="NumericMetric"/>s for 'Relevance' 'Truth' and 'Completeness' that explain the
+/// reasoning behind the corresponding scores. By default, <paramref name="includeReasoning"/> is set to
+/// <see langword="false"/>.
+/// </param>
+public class RelevanceTruthAndCompletenessEvaluatorOptions(bool includeReasoning = false)
```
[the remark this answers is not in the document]
shyamnamboodiripad: Good catch. Yes, this one should have been sealed. I've pushed a commit to fix this.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Utilities/JsonOutputFixer.cs
```diff
@@ -0,0 +1,78 @@
...
+#else
+        ReadOnlySpan<char> trimmed = json.ToCharArray();
+#endif
+
+        // Trim whitespace and markdown characters from beginning and end.
+        trimmed = trimmed.Trim().Trim(['`']);
+
+        // Trim 'json' marker from markdown if it exists.
+        int markerLength = "json".Length;
+        if (trimmed.Length > markerLength && trimmed[0..markerLength].SequenceEqual(['j', 's', 'o', 'n']))
```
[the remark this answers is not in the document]
shyamnamboodiripad: Fixed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Storage/AzureStorageResponseCache.cs
```diff
@@ -0,0 +1,332 @@
...
+                }
+            }
+        }
+    }
+
+    private (string entryFilePath, string contentsFilePath) GetPaths(string key)
+    {
+        string keyPath = $"{_iterationPath}/{key}";
+        string entryFilePath = $"{keyPath}/{EntryFileName}";
+        string contentsFilePath = $"{keyPath}/{ContentsFileName}";
```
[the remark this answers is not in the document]
shyamnamboodiripad: Thanks! Fixed.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
shyamnamboodiripad: Thanks for catching that - fixed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunExtensions.cs
shyamnamboodiripad: I was not sure whether `[]` would result in a new instance being allocated each time... But this may be a case of premature optimization even if this does allocate... I've gone ahead and replaced all usages of `EmptyMessages` with `[]`.

src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluatorExtensions.cs
shyamnamboodiripad: I've gone ahead and replaced all usages of `EmptyMessages` with `[]`.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunExtensions.cs
stephentoub: `[]` target-typed to `IEnumerable<T>` expands by the compiler to `Array.Empty<T>()`.

src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluationResult.cs
shyamnamboodiripad: Fixed

src/Libraries/Microsoft.Extensions.AI.Evaluation/Utilities/TaskExtensions.cs
shyamnamboodiripad: Thanks - I've `#if NET9`-ed that in

src/Libraries/Microsoft.Extensions.AI.Evaluation/CompositeEvaluator.cs
shyamnamboodiripad: Ah is that required in even if the `IAsyncEnumerable` is being produced by a function with the same `cancellationToken` being passed via `[EnumeratorCancellation]`?

stephentoub: No, effectively the compiler-generated code will check if the tokens are the same, and only if they're not it'll produce a new one that combines both.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Microsoft.Extensions.AI.Evaluation.Quality.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <Description>A library containing a set of evaluators for evaluating the quality (coherence, relevance, truth, completeness, groundedness, fluency, equivalence etc.) of responses received from an LLM.</Description>
+    <TargetFrameworks>$(TargetFrameworks);netstandard2.0</TargetFrameworks>
+    <RootNamespace>Microsoft.Extensions.AI.Evaluation.Quality</RootNamespace>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Workstream>AIEval</Workstream>
+    <Stage>preview</Stage>
+    <ForceLatestDotnetVersions>true</ForceLatestDotnetVersions>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
RussKie, line 13: This is no longer needed after #5864.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Microsoft.Extensions.AI.Evaluation.Reporting.Azure.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <RootNamespace>Microsoft.Extensions.AI.Evaluation.Reporting</RootNamespace>
+    <!-- EA0002: Use System.TimeProvider to make the code easier to test. -->
+    <NoWarn>$(NoWarn);EA0002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Workstream>AIEval</Workstream>
+    <Stage>preview</Stage>
+    <ForceLatestDotnetVersions>true</ForceLatestDotnetVersions>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
RussKie, line 15: This is no longer needed after #5864.

```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Microsoft.Extensions.AI.Evaluation.Reporting.csproj
```diff
@@ -0,0 +1,77 @@
...
+    <RootNamespace>Microsoft.Extensions.AI.Evaluation.Reporting</RootNamespace>
+    <!-- EA0002: Use System.TimeProvider to make the code easier to test. -->
+    <NoWarn>$(NoWarn);EA0002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Workstream>AIEval</Workstream>
+    <Stage>preview</Stage>
+    <ForceLatestDotnetVersions>true</ForceLatestDotnetVersions>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
RussKie, line 15: This is no longer needed after #5864.

```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Evaluation/Microsoft.Extensions.AI.Evaluation.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <Description>A library containing core abstractions for evaluating responses received from an LLM.</Description>
+    <TargetFrameworks>$(TargetFrameworks);netstandard2.0</TargetFrameworks>
+    <RootNamespace>Microsoft.Extensions.AI.Evaluation</RootNamespace>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Workstream>AIEval</Workstream>
+    <Stage>preview</Stage>
+    <ForceLatestDotnetVersions>true</ForceLatestDotnetVersions>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
RussKie, line 13: This is no longer needed after #5864.

```suggestion
```

shyamnamboodiripad: Removed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Microsoft.Extensions.AI.Evaluation.Reporting.csproj
shyamnamboodiripad: Removed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Microsoft.Extensions.AI.Evaluation.Reporting.Azure.csproj
shyamnamboodiripad: Removed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Microsoft.Extensions.AI.Evaluation.Quality.csproj
shyamnamboodiripad: Removed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Storage/AzureStorageResponseCache.CacheEntry.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public static CacheEntry Read(
+            DataLakeFileClient fileClient,
+            CancellationToken cancellationToken = default)
+        {
+            Response<DataLakeFileReadResult> content = fileClient.ReadContent(cancellationToken);
+
+            CacheEntry cacheEntry =
+                JsonSerializer.Deserialize(
+                    content.Value.Content.ToStream(),
```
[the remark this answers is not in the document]
shyamnamboodiripad: The static type of `content.Value.Content` is `System.BinaryData`. Also tagging @peterwald since he has more context around the DataLake APIs used in this file.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Storage/AzureStorageResponseCache.CacheEntry.cs
```diff
@@ -0,0 +1,101 @@
...
+        public void Write(
+            DataLakeFileClient fileClient,
+            CancellationToken cancellationToken = default)
+        {
+            MemoryStream stream = new();
+
+            JsonSerializer.Serialize(stream, this, AzureStorageSerializerContext.Default.CacheEntry);
+
+            _ = stream.Seek(0, SeekOrigin.Begin);
+            _ = fileClient.Upload(stream, overwrite: true, cancellationToken);
```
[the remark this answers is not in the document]
shyamnamboodiripad: Tagging @peterwald who should have more context around the DataLake APIs - but I believe we had to use the above approach since no such API was available.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Storage/DiskBasedResultStore.cs
shyamnamboodiripad: Fixed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
shyamnamboodiripad: I've converted the `static readonly` field above to a property.

The `const`s above need to be preserved since they are used exclusively as default values of method parameters.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/SingleNumericMetricEvaluator.cs
```diff
@@ -0,0 +1,88 @@
...
+namespace Microsoft.Extensions.AI.Evaluation.Quality;
+
+/// <summary>
+/// An <see langword="abstract"/> base class that can be used to implement an AI-based <see cref="IEvaluator"/> that
+/// produces an <see cref="EvaluationResult"/> containing a single <see cref="NumericMetric"/>.
+/// </summary>
+public abstract class SingleNumericMetricEvaluator : ChatConversationEvaluator
+{
+    /// <inheritdoc/>
+    public sealed override IReadOnlyCollection<string> EvaluationMetricNames => [MetricName];
```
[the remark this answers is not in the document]
shyamnamboodiripad: We chose Collection over List because this could be a HashSet (which wouldn't support indexing). IEnumerable was considered, however, we have generally avoided its use in properties to avoid questions related to deferred execution...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/ChatConversationEvaluator.cs
```diff
@@ -0,0 +1,397 @@
...
+                    tokenCounter.CountTokens("] ") +
+                    tokenCounter.CountTokens(content) +
+                    tokenCounter.CountTokens("\n")
+                : tokenCounter.CountTokens("[") +
+                    tokenCounter.CountTokens(author!) +
+                    tokenCounter.CountTokens(" (") +
+                    tokenCounter.CountTokens(role) +
+                    tokenCounter.CountTokens(")] ") +
+                    tokenCounter.CountTokens(content) +
+                    tokenCounter.CountTokens("\n");
```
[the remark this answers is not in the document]
shyamnamboodiripad: Yes, we did consider this. However, we also wanted to avoid allocating (potentially large) strings in cases where we happen to be short on tokens... (Perhaps we also need better token counting APIs that can help for this?)

The token counting being performed above is not perfect, and in extreme cases, it could end up rejecting some messages that might otherwise fit. But it should be conservative, and should generally always reject messages that can't fit...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/SingleNumericMetricEvaluator.cs
```diff
@@ -0,0 +1,88 @@
...
+        $"""
+        You are an AI assistant. You will be given the definition of an evaluation metric for assessing the quality of
+        a response in a question-answering task. Your job is to compute an accurate evaluation score for the provided
+        evaluation metric based on the provided scoring guidance.
+
+        This evaluation score should always be an integer between 1 and 5. So your response should be 1 or 2 or 3 or 4
+        or 5.
+
+        Your response should be a single character containing only the evaluation score. Do not include any other text
+        in your response besides the evaluation score.
```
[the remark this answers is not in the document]
shyamnamboodiripad: We haven't tried that yet. I remember that this had also come up during our API review earlier. It would definitely be worth trying - although I am not sure whether or not that would make these evaluation prompts less portable between different models?

I have added a comment to revisit this for now - I'll convert this to an issue later once we start moving our backlog here.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Utilities/JsonOutputFixer.cs
```diff
@@ -0,0 +1,78 @@
...
+    internal static ReadOnlySpan<char> TrimMarkdownDelimiters(string json)
+    {
+#if NET
+        ReadOnlySpan<char> trimmed = json;
+#else
+        ReadOnlySpan<char> trimmed = json.ToCharArray();
+#endif
+
+        // Trim whitespace and markdown characters from beginning and end.
+        trimmed = trimmed.Trim().Trim(['`']);
```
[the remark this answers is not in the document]
shyamnamboodiripad: I believe the intent was to do just enough to trim out the markdown prefix and suffix characters and handle some very specific problems we had run into with some of the GPT-3.5 responses a while back - @peterwald may remember more. This should work for the intended purpose.

That said, I am not sure we need this code anymore as newer models like GPR-4o may no longer exhibit the same problematic behaviors. (Unless perhaps we want to support running the same evaluation prompts against smaller / local models....)

We also set `ChatResponse.ResponseFormat` to `ChatResponseFormat.Json` for this evaluator now - although we could probably go one step further and include a JSOJN schema to structure the output more effectively as you proposed in another comment.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Utilities/JsonOutputFixer.cs
```diff
@@ -0,0 +1,78 @@
...
+            };
+
+        var messages = new List<ChatMessage>
+        {
+            new ChatMessage(ChatRole.System, SystemPrompt),
+            new ChatMessage(ChatRole.User, fixPrompt)
+        };
+
+        ChatCompletion result =
+            await chatConfig.ChatClient.CompleteAsync(
```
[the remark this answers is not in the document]
shyamnamboodiripad: I remember that we had touched on this during the API review, but we haven't explored that yet.

I have included a comment in the code for now - will convert this to an issue once we move our backlog to the current repo.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/FluencyEvaluator.cs
```diff
@@ -0,0 +1,92 @@
...
+            stars: 4
+
+            question: Can you describe your morning routine?
+            answer: Every morning, I wake up at 6 am, drink a glass of water, and do some light stretching. After that,
+            I take a shower and get dressed for work. Then, I have a healthy breakfast, usually consisting of oatmeal
+            and fruits, before leaving the house around 7:30 am.
+            stars: 5
+
+            question: {{renderedUserRequest}}
+            answer: {{renderedModelResponse}}
```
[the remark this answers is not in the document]
shyamnamboodiripad: The above prompts were mostly ported from equivalent python evaluators (from promptflow / azure python sdk). We do have a backlog item to refresh the prompts to reflect changes that have since happened in the original source.

That said, we don't have safeguards for the cases you describe above at the moment. I am guessing using the structured output / supplying JSON schema for the output would be one way to mitigate this, wouldn't it?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Utilities/ExceptionUtilities.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    internal static bool IsCancellation(this Exception exception)
+    {
+        if (exception is OperationCanceledException)
+        {
+            return true;
+        }
+        else if (exception is AggregateException aggregateException)
+        {
+            return aggregateException.InnerExceptions.All(IsCancellation);
```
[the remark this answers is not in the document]
shyamnamboodiripad: Yes, this should probably be updated to avoid recursion. I'll look into this tomorrow.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Storage/AzureStorageResponseCache.CacheEntry.cs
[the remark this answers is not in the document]
stephentoub: This would be better then as `content.Value.Content.ToMemory().Span`

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/RelevanceTruthAndCompletenessEvaluator.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// <see cref="RelevanceTruthAndCompletenessEvaluator"/> for 'Completeness'.
+    /// </summary>
+    public static string CompletenessMetricName => "Completeness";
+
+    /// <inheritdoc/>
+    public override IReadOnlyCollection<string> EvaluationMetricNames { get; } =
+        [RelevanceMetricName, TruthMetricName, CompletenessMetricName];
+
+    /// <inheritdoc/>
+    protected override ChatOptions? ChatOptions { get; } =
```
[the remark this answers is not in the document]
stephentoub: My concern wasn't setting the property, but changing values on the returned ChatOptions, e.g. if one test was able to do:
```C#
RelevanceTruthAndCompletenessEvaluator.Instance.Options.Temperature = 2f;
```
and then other tests that happened to run later will horked.

If every test creates their own instance, no worries.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/FluencyEvaluator.cs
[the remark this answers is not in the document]
stephentoub: I don't think structured output mitigates it. If the renderedUserRequest includes complicated output that contains e.g. "answer: blah blah blah", that will end up corrupting what the model sees in the few shot examples.

It may not be a real concern for these scenarios, but we should think through such possibilities, whether they be malicious or accidental.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/RelevanceTruthAndCompletenessEvaluator.cs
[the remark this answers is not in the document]
peterwald: This is a good point. The user can create evaluators for each scenario, or reuse them. We expect that they can be reused, and we do reuse them in our tests and examples. If someone mutated the Options on this evaluator, the modified options would be used by subsequent tests.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Storage/AzureStorageResponseCache.CacheEntry.cs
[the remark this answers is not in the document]
peterwald: Fixed

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/Storage/AzureStorageResponseCache.CacheEntry.cs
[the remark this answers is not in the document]
peterwald: AFAIK, there are only two ways to Upload to a DataLake file. We can provide a Stream or a path to a file on disk. So, no, I don't know of a better way to do this.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Utilities/JsonOutputFixer.cs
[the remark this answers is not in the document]
shyamnamboodiripad: I have created #5888 to track the structured output exploration

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/RelevanceTruthAndCompletenessEvaluator.cs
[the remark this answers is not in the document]
peterwald: Good point. I missed that.

peterwald (approved):
There are items we need to follow up on, but nothing that should block at this point in my opinion.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Utilities/ExceptionUtilities.cs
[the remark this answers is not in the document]
shyamnamboodiripad: Fixed.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
```diff
@@ -0,0 +1,66 @@
...
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Extensions.AI.Evaluation\Microsoft.Extensions.AI.Evaluation.csproj" />
+    <ProjectReference Include="..\Microsoft.Extensions.AI.Evaluation.Reporting\CSharp\Microsoft.Extensions.AI.Evaluation.Reporting.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleToTest Include="Microsoft.Extensions.AI.Evaluation.Console.Tests" />
```
RussKie, line 34: [nitpick] You could consider this instead:
```suggestion
    <InternalsVisibleToTest Include="$(AssemblyName).Tests" />
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Microsoft.Extensions.AI.Evaluation.Reporting.csproj
```diff
@@ -0,0 +1,91 @@
...
+
+  <ItemGroup>
+    <InternalsVisibleToTest Include="Microsoft.Extensions.AI.Evaluation.Integration.Tests" />
+    <InternalsVisibleToTest Include="Microsoft.Extensions.AI.Evaluation.Reporting.Tests" />
+  </ItemGroup>
+
+  <!-- Build the report code before this, so it can be embedded.
+  Setting the BeforeTargets=DispatchToInnerBuilds allows it to build only once before
+  the multiple parallel builds are invoked for each target framework. -->
+  <Target Name="BuildTypescriptBundle"
```
RussKie, line 39: Consider moving targets out of proj files into Directory.Build.targets. This would reduce clutter in proj files and could help navigating the code.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Microsoft.Extensions.AI.Evaluation.Reporting.csproj
```diff
@@ -0,0 +1,91 @@
...
+    </ItemGroup>
+  </Target>
+
+  <!-- Generate a version file that can be accessed from the script that builds the Azure DevOps Extension pacakge. -->
+  <Target Name="StampVSIXPackageVersion" BeforeTargets="CoreCompile">
+    <PropertyGroup>
+      <VSIXPackageVersionFile>..\TypeScript\azure-devops-report\VSIXPackageVersion.json</VSIXPackageVersionFile>
+      <VSIXVersion>$(VersionPrefix).42424242</VSIXVersion>
+      <VSIXVersion Condition=" '$(VersionSuffixDateStamp)'!='' ">$(VersionPrefix).$(VersionSuffixDateStamp)$(VersionSuffixBuildOfTheDayPadded)</VSIXVersion>
+      <Lines>{"PackageVersion":"$(VSIXVersion)"}</Lines>
```
RussKie, line 83: Some of these properties clash with the Arcade: https://github.com/dotnet/arcade/blob/caa0982cc4f4a9c744f33336d899e6fdf310640d/src/Microsoft.DotNet.Arcade.Sdk/tools/VisualStudio.targets#L4-L11

If you can't use the Arcade, consider renaming your properties. A common practice is to prefix properties with the underscore to indicate the "private" nature of the property.

eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -45,6 +51,11 @@ steps:
             $(_OfficialBuildIdArgs)
     displayName: Build
 
+  - ${{ if eq(parameters.isWindows, 'true') }}:
+    - pwsh: |
+          $(Build.SourcesDirectory)/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/azure-devops-report/build.ps1 -OutputPath $(Build.ArtifactStagingDirectory)\VSIX
```
RussKie, line 56: We don't use `$(Build.ArtifactStagingDirectory)` as we use the Arcade.

We have few help variables defined:
https://github.com/dotnet/extensions/blob/c086106865b9808eba1acd78db1f91680952e065/azure-pipelines-public.yml#L48-L54

RussKie (approved):

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
shyamnamboodiripad: Ah this is not always applicable since we have 6 product projects that are tested by 3 test projects. I'd prefer to leave this for now until we start having more 1-1 mappings.

eng/pipelines/templates/BuildAndTest.yml
shyamnamboodiripad: Tagging @peterwald. I synced up with @russkie offline - we can handle this in the follow up PRs you will be creating for setting up for the VSIX build.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Microsoft.Extensions.AI.Evaluation.Reporting.csproj
shyamnamboodiripad: Tagging @peterwald. Resolving for now but we should handle this in the follow up change you are planning for the VSIX build.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Microsoft.Extensions.AI.Evaluation.Reporting.csproj
shyamnamboodiripad: Good call. Tagging @peterwald - potentially something we can standardize in your follow up PR.

PR #5886, opened 2025-02-12
Replace `AIFunctionParameterMetadata` with `MethodInfo`

eiriktsarpalis:
Removes both `AIFunctionParameterMetadata` and `AIFunctionReturnParameterMetadata`, replacing both with an optional `MethodInfo? UnderlyingMethod` property.

Me and @stephentoub discussed possibly removing `AIFunctionMetadata` altogether and flattening all its remaining properties on `AIFunction` itself. This PR stops short of that so we can gather feedback on the basis of this initial removal.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5886)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -48,15 +39,13 @@ public AIFunctionMetadata(string name)
     /// <summary>Initializes a new instance of the <see cref="AIFunctionMetadata"/> class as a copy of another <see cref="AIFunctionMetadata"/>.</summary>
     /// <exception cref="ArgumentNullException">The <paramref name="metadata"/> was null.</exception>
     /// <remarks>
-    /// This creates a shallow clone of <paramref name="metadata"/>. The new instance's <see cref="Parameters"/> and
-    /// <see cref="ReturnParameter"/> properties will return the same objects as in the original instance.
+    /// This creates a shallow clone of <paramref name="metadata"/>.
     /// </remarks>
     public AIFunctionMetadata(AIFunctionMetadata metadata)
```
stephentoub, line 44: I thought we were getting rid of this class entirely...?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -76,33 +65,15 @@ public string Description
...
-
-    /// <summary>Gets parameter metadata for the return parameter.</summary>
-    /// <remarks>If the function has no return parameter, the value is a default instance of an <see cref="AIFunctionReturnParameterMetadata"/>.</remarks>
-    public AIFunctionReturnParameterMetadata ReturnParameter
-    {
-        get => _returnParameter;
-        init => _returnParameter = Throw.IfNull(value);
-    }
+    /// <summary>
+    /// Gets the underlying .NET method that the function could be wrapping.
```
stephentoub, line 69:
```suggestion
    /// Gets a <see cref="MethodInfo"/> for the underlying .NET method this <see cref="AIFunction"/> represents.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -76,33 +65,15 @@ public string Description
...
+    /// </summary>
+    /// <remarks>
+    /// Used to provide additional metadata on the function and its signature.
+    /// Setting this property is optional and should have no impact on function invocation or its JSON schema,
+    /// which is how <see cref="IChatClient"/> implementations interface with AI functions primarily.
```
stephentoub, line 74:
```suggestion
    /// This property provides additional metadata on the function and its signature.
    /// Setting this property is optional and should have no impact on function invocation or its JSON schema,
    /// which is how <see cref="IChatClient"/> implementations interface with AI functions primarily.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -46,39 +46,47 @@ public static partial class AIJsonUtilities
...
     /// <returns>A JSON schema document encoded as a <see cref="JsonElement"/>.</returns>
     public static JsonElement CreateFunctionJsonSchema(
+        MethodBase methodInfo,
         string? title = null,
         string? description = null,
-        IReadOnlyList<AIFunctionParameterMetadata>? parameters = null,
         JsonSerializerOptions? serializerOptions = null,
         AIJsonSchemaCreateOptions? inferenceOptions = null)
     {
+        _ = Throw.IfNull(methodInfo);
```
stephentoub, line 64: This method could previously be invoked with no arguments. Now it fails if you don't provide a methodInfo. That's by design?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -417,46 +405,36 @@ static bool IsAsyncMethod(MethodInfo method)
...
-                IsRequired = !parameter.IsOptional,
-                ParameterType = parameter.ParameterType,
-            };
         }
 
         /// <summary>
         /// Gets a delegate for handling the result value of a method, converting it into the <see cref="Task{FunctionResult}"/> to return from the invocation.
         /// </summary>
-        private static Type GetReturnMarshaller(MethodInfo method, out Func<object?, ValueTask<object?>> marshaller)
+        private static Func<object?, ValueTask<object?>> GetReturnMarshaller(MethodInfo method, out Type returnType)
```
stephentoub, line 413: Curious what motivated swapping the return and out?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -417,46 +405,36 @@ static bool IsAsyncMethod(MethodInfo method)
...
         {
             // Handle each known return type for the method
-            Type returnType = method.ReturnType;
+            returnType = method.ReturnType;
```
stephentoub, line 416: Can this be moved to the end just before the final return?

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/PromptBasedFunctionCallingChatClient.cs
```diff
@@ -181,14 +183,14 @@ answer the user's question without repeating the same tool call.
     {
         Name = tool.Metadata.Name,
         Description = tool.Metadata.Description,
-        Arguments = tool.Metadata.Parameters.ToDictionary(
-            p => p.Name,
+        Arguments = tool.Metadata.UnderlyingMethod!.GetParameters().ToDictionary(
```
stephentoub, line 186: What guarantees UnderlyingMethod is non-null here? If this is processing arbitrary AIFunctions coming in from Tools, couldn't this be null? Same question about the Name below.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: The name of the method suggests that the return type should be the marshaller. Results in cleaner code in the method body I think.

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/PromptBasedFunctionCallingChatClient.cs
eiriktsarpalis: It's test code hitting AIFunctions that we define. I could coalesce the null case if you want.

stephentoub: It's not really a test. It's a sample for writing a custom IChatClient middleware component.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: I've changed the signature of the method altogether. It didn't accept `MethodInfo` at all before this change.

stephentoub: I know. And before this change, nothing was required, you could call the method with no arguments and still get back something. That's no longer the case. Was that not a scenario?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: It's being used in a bunch of locations inside the method body. Obviously I could just query `method.ReturnType` directly but that just translates to more virtual calls.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: The only scenario for this particular method was to act as an ad hoc helper for AIFunctionFactory. It used to be the case that it accepted a number of optional parameters that would populate relevant schema keywords. I experimented with keeping the same signature as before using an optional list of `ParameterInfo` but it still felt a bit ad hoc to me.

Having it take a `MethodInfo` makes it more clear that it's about generating schemas for .NET functions. It nicely complements the other public method in this class that takes a `Type` parameter.

eiriktsarpalis:
@SteveSandersonMS what do you think? Is this a step in the right direction?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
```diff
@@ -76,33 +65,15 @@ public string Description
...
-    }
+    /// <summary>
+    /// Gets a <see cref="MethodInfo"/> for the underlying .NET method this <see cref="AIFunction"/> represents.
+    /// </summary>
+    /// <remarks>
+    /// This property provides additional metadata on the function and its signature.
+    /// Setting this property is optional and should have no impact on function invocation or its JSON schema,
+    /// which is how <see cref="IChatClient"/> implementations interface with AI functions primarily.
+    /// </remarks>
+    public MethodInfo? UnderlyingMethod { get; init; }
```
SteveSandersonMS, line 76: Without yet fully understanding the design intent, I'm not clear on this from the XML docs alone. Why would someone set an `UnderlyingMethod` if it "has no impact on function invocation or its JSON schema"? Or as a consumer, when should I expect this to be available or not available?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -46,39 +46,47 @@ public static partial class AIJsonUtilities
...
     /// </summary>
+    /// <param name="methodInfo">The method from which to extract schema information.</param>
     /// <param name="title">The title keyword used by the method schema.</param>
     /// <param name="description">The description keyword used by the method schema.</param>
-    /// <param name="parameters">The AI function parameter metadata.</param>
     /// <param name="serializerOptions">The options used to extract the schema from the specified type.</param>
     /// <param name="inferenceOptions">The options controlling schema inference.</param>
     /// <returns>A JSON schema document encoded as a <see cref="JsonElement"/>.</returns>
     public static JsonElement CreateFunctionJsonSchema(
+        MethodBase methodInfo,
```
SteveSandersonMS, line 58: I'd recommend naming this simply `method`, because `methodInfo` sounds like it should be an instance of `MethodInfo`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -46,39 +46,47 @@ public static partial class AIJsonUtilities
     private static readonly string[] _schemaKeywordsDisallowedByAIVendors = ["minLength", "maxLength", "pattern", "format"];
 
     /// <summary>
-    /// Determines a JSON schema for the provided AI function parameter metadata.
+    /// Determines a JSON schema for the provided method signature.
```
SteveSandersonMS, line 49:
```suggestion
    /// Determines a JSON schema for the provided method.
```

SteveSandersonMS:
@eiriktsarpalis @stephentoub Can you summarize the design intent?

I've read through the implementation and simplifying things certainly looks good and beneficial. But it's not obvious initially why we can get away with just removing all the metadata, unless it was just not serving a purpose before.

My guess is that you've identified that it was an unnecessary extra level of indirection, e.g.:

 * Before:
   * App provides `AIFunctionMetadata`, or a `MethodInfo` which we convert to `AIFunctionMetadata`
   * IChatClient implementation receives `AIFunctionMetadata` and converts it to its own wire format, usually JSON schema
* After:
  * App provides JSON schema, or a `MethodInfo` which we convert to JSON schema
  * IChatClient uses the JSON schema or, in the unusual case where it needs a different format, needs to parse the JSON schema and convert this it its own other wire format

At least, this lines up with my memory of the initial design, though I know that may have already changed. I know there was initially a lot of debate about to what extent we're willing to make a strong assumption that JSON schema is the one true format for function descriptions.

> Me and @stephentoub discussed possibly removing AIFunctionMetadata altogether and flattening all its remaining properties on AIFunction itself.

It's not clear to me which properties would remain. In this PR you've removed our abstract representation for parameter metadata, but left the abstract representation for function-level metadata (name, description).

Is the intent that `AIFunction` would gain `Name` and `Description` properties? If so, can you clarify why we would need those if we don't need metadata about parameters? In the OpenAI function metadata format at least, the schema innately contains `name` and `description` I'm guessing we could eliminate `Name` and `Description` as separate properties.

eiriktsarpalis:
> Can you summarize the design intent?

One use case for `AIFunction` that has come up both in my own Copilot extensions work and in feedback received from partner teams is being able to deserialize a "tools" payload into a symbolic `AIFunction` instance (aka an instance that declares a signature but does not encapsulate behavior). Instances like that are meant to be forwarded to other chat services as opposed to being executed by a local function invoking middleware.

The presence of `AIFunctionMetadata` in this case confuses implementers, compelling them to populate them with unnecessary information. The specifc feedback we got from the VS Copilot team was that they didn't know what to use in the `AIFunctionParameterMetadata.Type` property; even though it's optional they didn't understand the ramifications of not using something there.

Given that `AIFunctionMetadata` is really a proxy for `ParameterInfo`, we initially debated using an `IReadOnlyList<ParameterInfo>` but ultimately ended up on `MethodInfo?`. It more clearly communicates that this particular instance is backed by a .NET method, provides richer metadata, and it substantially simplifies function implementations. Because `MethodInfo` can be inherited, it doesn't prevent users from defining synthetic metadata if they so choose.

> I know there was initially a lot of debate about to what extent we're willing to make a strong assumption that JSON schema is the one true format for function descriptions.

I think this has always been the assumption, what we've been changing recently (particularly with https://github.com/dotnet/extensions/pull/5826) is how explicit this is being made in the API contract. We originally exposed an `object? Schema` property scoped to `AIFunctionParameterMetadata` that we were documenting as being a JSON schema. Every component in our infrastructure and leaf client implementations expected this to be `JsonElement`. That PR changed the schema to explicitly be a `JsonElement` scoped to the function level as opposed to the parameter level. It came following [customer feedback](https://github.com/dotnet/extensions/issues/5655) that it wasn't possible to simply specify a self-contained JSON schema, and users could only work with schema fragments scoped to individual parameters.

Assuming a new schematization format does come around in the future, the `MethodInfo` should contain all metadata necessary to reconstitute that on demand. If a new format proves ubiquitous enough, we can add it as a separate (strongly typed) schema property.

> Is the intent that AIFunction would gain Name and Description properties? If so, can you clarify why we would need those if we don't need metadata about parameters? In the OpenAI function metadata format at least, the schema innately contains name and description, so I'm guessing we could eliminate Name and Description as separate properties.

I've kept those since they are required values in [envelopes that contain the function schema](https://github.com/dotnet/extensions/blob/89b0d7e84d8c716b1635e44d8a5d9f1e41a2bde2/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaFunctionTool.cs#L8-L9). The "description" keyword is optional in JSON schema and "name" does not exist per se -- the canonical keyword in this case is "title".

@stephentoub made the point that he would like to see `Name` (and possible `Description`?) lifted to to the `AITool` base class since he needs to introduce a new derived tool type that needs this information as well.

SteveSandersonMS:
> One use case for AIFunction that has come up both in my own Copilot extensions work and in feedback received from partner teams is being able to deserialize a "tools" payload into a symbolic AIFunction instance (aka an instance that declares a signature but does not encapsulate behavior). Instances like that are meant to be forwarded to other chat services as opposed to being executed by a local function invoking middleware.

I guess I'm less clear now on how to understand the situation. Could you give an outline of what's happening in this situation? Who's providing data to who else and what are they going to do with it?

My guess is that, if you're on the *server* end of an `IChatClient` interaction (e.g., implementing an OpenAI-style REST API) then you'd receive incoming `/chat` calls that contain tool descriptors. I would have expected that you need all the metadata, including the parameter types, in order to be able to generate valid calls to those tools.

You mention that the instances *are meant to be forwarded to other chat services* but surely somewhere at the end of this line, the final recipient needs the full metadata (including parameter types) in order to be able to generate valid calls to those tools.

In what situation do you need a name and description but no definition of the parameter schema?

> I've kept those since they are required values in [envelopes that contain the function schema](https://github.com/dotnet/extensions/blob/89b0d7e84d8c716b1635e44d8a5d9f1e41a2bde2/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaFunctionTool.cs#L8-L9). The "description" keyword is optional in JSON schema and "name" does not exist per se -- the canonical keyword in this case is "title".

Not sure how to interpret the example given (`OllamaFunctionTool`). That class requires not only `Name` and `Description`, but also `Parameters`. So how does it follow that we can remove `Parameters` but not `Name` and `Description`?

> The "description" keyword is optional in JSON schema and "name" does not exist per se -- the canonical keyword in this case is "title".

This sounds like JSON schema has an official spec for how functions would be represented. Is that the case? I'm [guessing not](https://stackoverflow.com/a/69478880).

So if this is an informally-defined way to represent functions, basically just taking the de-facto standard from OpenAI, what difference does it make what is canonical for JSON schema?

Apologies if my lack of context here is slowing things down. Please don't interpret any of it as blocking this change, which almost certainly is a good thing since it's removing complexity.

eiriktsarpalis:
> You mention that the instances _are meant to be forwarded to other chat services_ but surely somewhere at the end of this line, the final recipient needs the full metadata (including parameter types) in order to be able to generate valid calls to those tools.

Why would it need the .NET types to do so? The final recipient needn't be a .NET process; the contract is to generate argument data based on a schema that has been specified on the wire. Only an AIFunctoin that needs to invoke an underlying .NET method requires type information so that it can marshal and unmarshal data.

> Not sure how to interpret the example given (`OllamaFunctionTool`). That class requires not only `Name` and `Description`, but also `Parameters`. So how does it follow that we can remove `Parameters` but not `Name` and `Description`?

The "parameters" property used in Ollama (and OpenAI as well) is a bit of a misnomer since it actually contains a complete JSON schema and not just parameter keywords. Here's what the corresponding model in our code looks like:

https://github.com/dotnet/extensions/blob/89b0d7e84d8c716b1635e44d8a5d9f1e41a2bde2/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaFunctionToolParameters.cs#L9-L14

In other words, "parameters" should be thought of as the JSON schema for parameterizing the function, but the actual schemas for individual inputs are nested inside a "properties" keyword -- in accordance with JSON schema specification.

> In what situation do you need a name and description but no definition of the parameter schema?

I haven't tested, but my expection is that it would always be mandatory. Since my previous change the `AIFunctionMetadata.Schema` is non-nullable and always returns a valid JSON schema for functions.

> This sounds like JSON schema has an official spec for how functions would be represented. Is that the case? I'm [guessing not](https://stackoverflow.com/a/69478880).

Correct. De facto, schemas for functions model a pseudo-object that lists all function parameters as its properties. There are some deviations here, for example OpenAI has coined the "function" type although in my experiements it seems happy with `"type" : "object"` declarations as well.

> So if this is an informally-defined way to represent functions, basically just taking the de-facto standard from OpenAI, what difference does it make what is canonical for JSON schema?

`AIFunctionFactory` will always populate the function schema with a valid and self-contained schema document per the latest spec. But we do apply transformations to it so that it does conform with OpenAI (and presumably other major vendor) restrictions.

SteveSandersonMS:
Thanks @eiriktsarpalis and @stephentoub. Lots of it is clearer now. In particular the benefit of retaining name/description properties to simplify telemetry/debugging etc makes a lot of sense.

I'm understanding now that `AIFunctionMetadata` was itself already duplicative, in the sense that the JSON schema already carried all the information. Having the same info in a different form didn't really achieve anything, now we've settled on taking JSON schema as the standard.

So yes this all does seem like a good direction. Thanks for explaining the details!

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionMetadata.cs
eiriktsarpalis: Does this look better?
```
    /// <summary>
    /// Gets the underlying <see cref="MethodInfo"/> that this <see cref="AIFunction"/> might be wrapping.
    /// </summary>
    /// <remarks>
    /// Provides additional metadata on the function and its signature. Implementations not wrapping .NET methods may return <see langword="null"/>.
    /// </remarks>
    public virtual MethodInfo? UnderlyingMethod { get; init; }
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: Changed, thanks.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -1,20 +1,95 @@
...
+    private readonly string _description = string.Empty;
+
+    /// <summary>The JSON schema describing the function and its input parameters.</summary>
+    private readonly JsonElement _jsonSchema = AIJsonUtilities.DefaultJsonSchema;
+
+    /// <summary>Optional additional properties in addition to the named properties already available on this class.</summary>
+    private readonly IReadOnlyDictionary<string, object?> _additionalProperties = EmptyReadOnlyDictionary<string, object?>.Instance;
+
+    /// <summary>Gets the name of the function.</summary>
+    public abstract string Name { get; }
```
eiriktsarpalis, line 31: @stephentoub I've now flattened the properties inside `AIFunction`. Using a protected constructor created akwardness in a number of implementations so I opted for making `Name` abstract instead. I've also marked all other properties easier for easier extension.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -1,20 +1,95 @@
...
+    /// }
+    /// </code>
+    /// <para>
+    /// The metadata present in the schema document plays an important role in guiding AI function invocation.
+    /// </para>
+    /// <para>
+    /// When no schema is specified, consuming chat clients should assume the "{}" or "true" schema, indicating that any JSON input is admissible.
+    /// </para>
+    /// </remarks>
+    public virtual JsonElement JsonSchema
```
eiriktsarpalis, line 74: Making this property virtual means that implementers can bypass schema validation. I'm mostly ok with this since our leaf clients would catch this and it's intended to provide guardrails for simpler implementations.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -1,20 +1,95 @@
...
+
+    /// <summary>Gets the name of the function.</summary>
+    public abstract string Name { get; }
+
+    /// <summary>Gets a description of the function, suitable for use in describing the purpose to a model.</summary>
+    [AllowNull]
+    public virtual string Description
+    {
+        get => _description;
+        init => _description = value ?? string.Empty;
```
stephentoub, line 38: Should this be abstract like Name instead of an init?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -1,20 +1,95 @@
...
+    }
+
+    /// <summary>
+    /// Gets the underlying <see cref="MethodInfo"/> that this <see cref="AIFunction"/> might be wrapping.
+    /// </summary>
+    /// <remarks>
+    /// Provides additional metadata on the function and its signature. Implementations not wrapping .NET methods may return <see langword="null"/>.
+    /// </remarks>
+    public virtual MethodInfo? UnderlyingMethod { get; init; }
```
stephentoub, line 47: Same here. I guess it's a question for everything here... how are we deciding some things are init vs abstract?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
eiriktsarpalis: It would be one more thing that users would need to implement. We may decide that descriptions are an essential feature of AI functions (since it's the primary means with which we convey hints to the LLM about the function's behavior) and therefore require implementers be deliberate about what they set here.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
eiriktsarpalis: I reserved abstract for the properties that are required. Everything else I left as virtual to make overriding more convenient but they don't need to be.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
stephentoub: If that's the concern, it could just be virtual, with the base returning string.Empty. I'm not sure what init buys us here. It could also be problematic... if someone implements a public derived type, they lack any ability to prevent Description from directly being set and validating it beyond the (lack of) validation provided here.

eiriktsarpalis: Should I let you determine the appropriate signature as you factor in the changes from #5898?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
stephentoub: Can we also remove the "init"?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
stephentoub: Ok

eiriktsarpalis: > I'm not sure what init buys us here.

It lets constructors of derived types assign values to properties without needing to implement them. If there's no init we might as well leave it abstract.

eiriktsarpalis: Perhaps `protected init` is warranted?

stephentoub: > It lets constructors of derived types assign values to properties without needing to implement them.

That's the difference between:
```C#
Description = value;
```
and
```C#
Description = value;
...
public override string Description { get; }
```
which does not seem onerous.

And having `protected init` forces the base to have storage for the field, even if the derived type doesn't actually use it.

eiriktsarpalis: Removed.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryCreateOptions.cs
```diff
@@ -35,10 +35,10 @@ public JsonSerializerOptions SerializerOptions
     /// <summary>
     /// Gets or sets the <see cref="AIJsonSchemaCreateOptions"/> governing the generation of JSON schemas for the function.
     /// </summary>
-    public AIJsonSchemaCreateOptions SchemaCreateOptions
+    public AIJsonSchemaCreateOptions JsonSchemaCreateOptions
```
stephentoub, line 38: Would it make sense at this point for JsonSchemaCreateOptions and SerializerOptions to be nullable, and then just substitute the default values if they're null when actually constructing the AIFunction? It's not clear to me at this point why that's how Name/Description/AdditionalProperties work but not how these work.

eiriktsarpalis: We could do that. Making JSO non-nullable specifically is a holdover from when the factory APIs not accepting an explicit JSO were being marked RUC/RDC.

PR #5898, opened 2025-02-13
Add marker AITool for enabling code interpreters

stephentoub:
Various AI services (Gemini, Bedrock, OpenAI, etc.) now support server-side code interpreting, where the model can generate code it can then execute in a sandbox. For good reason, they all require opting in. We can model that well as an AITool-derived type.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5898)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
@@ -1,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a tool that can be specified to an AI service to enable it to execute code it generates.</summary>
+/// <remarks>
+/// This tool does not itself implement code interpration. It is a marker that can be used to inform a service
+/// that the service is allowed to execute its generated code if the service is capable of doing so.
+/// </remarks>
+public class CodeInterpreterTool : AITool
+{
+    /// <summary>Initializes a new instance of the <see cref="CodeInterpreterTool"/> class.</summary>
+    public CodeInterpreterTool()
+    {
+    }
+
+    /// <summary>Gets or sets additional properties that impact the behavior of the tool.</summary>
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+}
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
@@ -1,6 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
+using Microsoft.Shared.Collections;
+
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a tool that can be specified to an AI service to enable it to execute code it generates.</summary>
@@ -11,10 +14,11 @@ namespace Microsoft.Extensions.AI;
 public class CodeInterpreterTool : AITool
 {
     /// <summary>Initializes a new instance of the <see cref="CodeInterpreterTool"/> class.</summary>
-    public CodeInterpreterTool()
+    public CodeInterpreterTool(IReadOnlyDictionary<string, object?>? additionalProperties = null)
     {
+        AdditionalProperties = additionalProperties ?? EmptyReadOnlyDictionary<string, object?>.Instance;
     }
 
-    /// <summary>Gets or sets additional properties that impact the behavior of the tool.</summary>
-    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+    /// <inheritdoc/>
+    public override IReadOnlyDictionary<string, object?> AdditionalProperties { get; }
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
@@ -1,9 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Collections.Generic;
-using Microsoft.Shared.Collections;
-
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a tool that can be specified to an AI service to enable it to execute code it generates.</summary>
@@ -14,11 +11,7 @@ namespace Microsoft.Extensions.AI;
 public class CodeInterpreterTool : AITool
 {
     /// <summary>Initializes a new instance of the <see cref="CodeInterpreterTool"/> class.</summary>
-    public CodeInterpreterTool(IReadOnlyDictionary<string, object?>? additionalProperties = null)
+    public CodeInterpreterTool()
     {
-        AdditionalProperties = additionalProperties ?? EmptyReadOnlyDictionary<string, object?>.Instance;
     }
-
-    /// <inheritdoc/>
-    public override IReadOnlyDictionary<string, object?> AdditionalProperties { get; }
 }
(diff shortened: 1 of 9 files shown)


src/Analyzers/Microsoft.Analyzers.Local/ApiLifecycle/Utils.cs
```diff
@@ -90,6 +90,6 @@ private static void GetBaseTypesImpl(List<string> results, string baseTypesStrin
             }
         }
 
-        results.Add(baseTypesString.Substring(start, baseTypesString.Length - start).Trim());
+        results.Add(baseTypesString.Substring(start).Trim());
```
stephentoub, line 93: This and a couple of the other line changes in the PR are unrelated and are just addressing some easy analyzer suggestions in the solution.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdateExtensions.cs
```diff
@@ -89,10 +89,11 @@ static async Task<ChatResponse> ToChatResponseAsync(
     /// <param name="response">The <see cref="ChatResponse"/> object whose properties should be updated based on <paramref name="update"/>.</param>
     private static void ProcessUpdate(ChatResponseUpdate update, Dictionary<int, ChatMessage> messages, ChatResponse response)
     {
-        response.ResponseId ??= update.ResponseId;
+        response.ChatThreadId ??= update.ChatThreadId;
```
stephentoub, line 92: The lack of this was an existing bug from my having added this property.

src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
```diff
@@ -0,0 +1,20 @@
...
+public class CodeInterpreterTool : AITool
+{
+    /// <summary>Initializes a new instance of the <see cref="CodeInterpreterTool"/> class.</summary>
+    public CodeInterpreterTool()
+    {
+    }
+
+    /// <summary>Gets or sets additional properties that impact the behavior of the tool.</summary>
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+}
```
SteveSandersonMS, line 20: How confident are you that this will be useful end-to-end across a wide range of backends?

I'm asking because there could be a lot of standard and semi-standard tools built into different AI services and it might not be obvious which of them to immortalize in the framework. For example, "file search" is built into OpenAI Assistants too - would we add that here?

In the case of code interpreter, baking this in implies it would work across multiple backends without additional parameterization. For example it wouldn't be valid for some provder to add a required "language" parameter or any other required params/config.

stephentoub: > How confident are you that this will be useful end-to-end across a wide range of backends?

Is the question whether this should be an OpenAICodeInterpreterTool in an OpenAI-specific library vs this which I've created as a general thing in abstractions?

OpenAI Assistants support it: https://platform.openai.com/docs/assistants/tools/code-interpreter#enabling-code-interpreter
Gemini supports it: https://ai.google.dev/gemini-api/docs/code-execution?lang=python#get-started
AWS Bedrock supports it (at least for its agents right now... search this for "actionGroupName='code-interpreter'"): https://aws-samples.github.io/amazon-bedrock-samples/agents-and-function-calling/bedrock-agents/features-examples/10-create-agent-with-code-interpreter/10-create-agent-with-code-interpreter/

It appears to be becoming a staple.

You think we should hold off?

> For example, "file search" is built into OpenAI Assistants too - would we add that here?

I initially had CodeInterpreterTool as a singleton, but changed it to be instantiable with AdditionalProperties exactly because of scenarios like this, where you might want to give the code interpreter tool additional service-specific context. In theory OpenAI supports that, but the OpenAI client nuget package doesn't appear to.

SteveSandersonMS: Personally I'm totally fine with us adding things like this, but since we don't have a precedent for something exactly like this (unless you count function calling), wanted to discuss the general premise and clarify what sort of precedent scope we aim it to be.

Given this makes sense for three backends now, and how general it feels as a concept, I'm happy enough. We might even feel the bar for inclusion should be fairly low, since what cost does it have? If a given backend doesn't support something, there'll be an error, but that's no worse than having no way to express the developer intent.

stephentoub: > If a given backend doesn't support something, there'll be an error

I think it'd be more of a nop. Backends that don't support function calling simply don't invoke functions, even if they're provided them, as is the case for services that nominally support function calling but simply don't end up invoking tools. If a backend doesn't support the notion of code interpretation, than it can just ignore this, just as if it simply chose not to use it.

eiriktsarpalis: I notice that `AdditionalProperties` is mutable here whereas in `AIFunction` it is not. Should we consolidate the too?

src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
```diff
@@ -0,0 +1,20 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a tool that can be specified to an AI service to enable it to execute code it generates.</summary>
+/// <remarks>
+/// This tool does not itself implement code interpration. It is a marker that can be used to inform a service
+/// that the service is allowed to execute its generated code if the service is capable of doing so.
+/// </remarks>
+public class CodeInterpreterTool : AITool
```
eiriktsarpalis, line 11: Do we want to support polymorphic serialization for `AITool` like we do for `AIContent`? I don't believe we have a use case for persisting these objects directly today, but maybe one will come up. 

For `AIFunction` specifically we can't safely roundtrip behavior but it should be possible to roundtrip the schema.

stephentoub:
@SteveSandersonMS, @eiriktsarpalis, please take a look at the changes to AIFunction / AITool. Thanks.

SteveSandersonMS (approved):
Looks good

src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>Initializes a new instance of the <see cref="CodeInterpreterTool"/> class.</summary>
+    public CodeInterpreterTool(IReadOnlyDictionary<string, object?>? additionalProperties = null)
```
eiriktsarpalis, line 17: Why not just make it an init property? It feel like this type could admit more properties in the future.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
```diff
@@ -212,19 +212,25 @@ private static (RunCreationOptions RunOptions, List<FunctionResultContent>? Tool
...
+
+                            var functionParameters = BinaryData.FromBytes(
+                                JsonSerializer.SerializeToUtf8Bytes(
+                                    JsonSerializer.Deserialize(aiFunction.JsonSchema, OpenAIJsonContext.Default.OpenAIChatToolJson)!,
+                                    OpenAIJsonContext.Default.OpenAIChatToolJson));
+
+                            runOptions.ToolsOverride.Add(ToolDefinition.CreateFunction(aiFunction.Name, aiFunction.Description, functionParameters, strict));
+                            break;
+
+                        case CodeInterpreterTool:
```
eiriktsarpalis, line 231: Shouldn't we handle this case in the inverse mapper as well?

stephentoub: Good point. Probably.

stephentoub: Turns out the OpenAI ChatToolKind only currently enables Function.

src/Libraries/Microsoft.Extensions.AI.Abstractions/CodeInterpreterTool.cs
stephentoub: It's worth thinking about, but we don't really have a good way to roundtrip the majority case we care about, as you say.

PR #5905, opened 2025-02-14
Move AIFunctionFactory to abstractions project.

eiriktsarpalis:
I tend to regard `AIFunctionFactory` to be more foundational infrastructure and less an extension that is typical of components found in the `Microsoft.Extensions.AI` project. This PR proposes we move it to the `Abstractions` so that:

1. We remove a hard dependency on STJ 9 for function factories.
2. Leaf client implementations can rely on that infrastructure to create canonical AIFunction instances of their own.

In additional to moving over the relevant source files, this PR is also renaming `AIFunctionFactoryCreateOptions` to `AIFunctionFactoryOptions` since the "Create" verb seemed redundant.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5905)


stephentoub:
> Leaf client implementations can rely on that infrastructure to create canonical AIFunction instances of their own.

I've not seen any need for that. Have you?

> We remove a hard dependency on STJ 9 for function factories.

I've also not seen that as a problem for apps, which are the general consumer of this.

> This PR proposes we move it to the Abstractions so that:

My main concerns are:
1. These aren't abstractions nor are they needed by the 90% case of implementors. They're typically an app level concern, and apps reference M.E.AI. If you need AI Function Factory, there's a super high chance you also need FunctionInvokingChatClient.
2. This is just one creation method. What happens if we decide to add an OpenAI import method, or for model context protocol, or other such imports. Those would naturally belong here as well, but also not in abstractions.
3. We don't currently imbue the generated functions with metrics or other forms of telemetry, but we might want to (SK does). This would either prevent that or force more dependencies into abstractions.

Doesn't seem like the risk is worth the reward. And nothing prevents us from type forwarding them down later if we decide then we actually want this.

PR #5906, opened 2025-02-14
Rename AIFunctionFactoryOptions and make all properties nullable.

eiriktsarpalis:
Salvages the API changes from #5905
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5906)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryOptions.cs
@@ -6,40 +6,34 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Reflection;
 using System.Text.Json;
-using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>
 /// Represents options that can be provided when creating an <see cref="AIFunction"/> from a method.
 /// </summary>
-public sealed class AIFunctionFactoryCreateOptions
+public sealed class AIFunctionFactoryOptions
 {
-    private JsonSerializerOptions _options = AIJsonUtilities.DefaultOptions;
-    private AIJsonSchemaCreateOptions _jsonSchemaCreateOptions = AIJsonSchemaCreateOptions.Default;
-
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryOptions.cs
```diff
@@ -6,40 +6,34 @@
...
 using System.Text.Json;
-using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>
 /// Represents options that can be provided when creating an <see cref="AIFunction"/> from a method.
 /// </summary>
-public sealed class AIFunctionFactoryCreateOptions
+public sealed class AIFunctionFactoryOptions
```
stephentoub, line 15: "Create" was there in case we added other non-Create methods to AIFunctionFactory in the future. But if we think any factory APIs we add would want to support these same options, removing "Create" seems ok. The one member I'm a little hesitant about is JsonSchemaCreateOptions, which is very tied to the notion of creating a schema from a .NET method; what would we do if we wanted to import something that already had a schema? I guess we'd just say it'd be ignored?

eiriktsarpalis: > he one member I'm a little hesitant about is JsonSchemaCreateOptions, which is very tied to the notion of creating a schema from a .NET method

Isn't deriving an AIFunction from a .NET method the whole point of AIFunctionFactory? Do we expect that it might be used to create functions from other things? Today this use case is covered by extending the base class directly, which I think is appropriate.

stephentoub: > Isn't deriving an AIFunction from a .NET method the whole point of AIFunctionFactory?

It's the point of the current Create overloads we expose, not necessarily of the whole class. SK, for example, has its KernelFunctionFactory counterpart, which has methods on it like CreateFromMethod (our Create) but also CreateFromPrompt.

stephentoub (approved):

PR #5908, opened 2025-02-15
Reinstate caching in schema generation

eiriktsarpalis:
I spent a bit of time investigating the performance of `AIFunctionFactory.Create` where I discovered that schema generation is the most expensive factor by far is schema derivation. Even though the schema generation routines used to employ a caching mechanism in the past, this got removed during the recent refactorings since it no longer is necessary for leaf clients to resolve schemas on the fly.

This PR reinstates a different version of the caching mechanism for the purpose of speeding up `AIFunctionFactory.Create` when applied multiple times on the same delegate.

## Benchmarks

```C#
[MemoryDiagnoser]
public class Benchmark
{
    private static readonly Func<string, int, MyPoco?, MyPoco> _delegate = MyFunc;

    [Benchmark]
    public JsonElement CreateFunctionSchema() => AIJsonUtilities.CreateFunctionJsonSchema(_delegate.Method);

    [Benchmark]
    public JsonElement CreateFunctionSchema_CustomDescription() => AIJsonUtilities.CreateFunctionJsonSchema(_delegate.Method, description: "custom description");

    [Benchmark]
    public JsonElement CreateFunctionSchema_NewOptions() => AIJsonUtilities.CreateFunctionJsonSchema(_delegate.Method, serializerOptions: new(JsonSerializerOptions.Default));

    [Benchmark]
    public AIFunction CreateAIFunction() => AIFunctionFactory.Create(_delegate);

    [Benchmark]
    public AIFunction CreateAIFunction_CustomDescription() => AIFunctionFactory.Create(_delegate, description: "custom description");

    [Benchmark]
    public AIFunction CreateAIFunction_NewOptions() => AIFunctionFactory.Create(_delegate, serializerOptions: new(JsonSerializerOptions.Default));

    [Description("My awesome function")]
    public static MyPoco MyFunc([Description("Parameter 1")]string param1, int param2 = 42, MyPoco? other = null) => throw new NotImplementedException();

    public record MyPoco(int X, int Y);
}
```

### Main

| Method                                 | Mean     | Error     | StdDev    | Gen0   | Allocated |
|--------------------------------------- |---------:|----------:|----------:|-------:|----------:|
| CreateFunctionSchema                   | 6.345 us | 0.0848 us | 0.0708 us | 2.4109 |   9.95 KB |
| CreateFunctionSchema_CustomDescription | 6.041 us | 0.0357 us | 0.0298 us | 2.3499 |    9.7 KB |
| CreateFunctionSchema_NewOptions        | 6.438 us | 0.0260 us | 0.0230 us | 2.1973 |    9.1 KB |
| CreateAIFunction                       | 6.706 us | 0.0344 us | 0.0322 us | 2.5940 |  10.65 KB |
| CreateAIFunction_CustomDescription     | 6.298 us | 0.0393 us | 0.0368 us | 2.5330 |  10.45 KB |
| CreateAIFunction_CustomOptions         | 6.683 us | 0.0300 us | 0.0281 us | 2.3804 |   9.85 KB |

### PR

| Method                                 | Mean        | Error     | StdDev    | Gen0   | Allocated |
|--------------------------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateFunctionSchema                   | 6,679.98 ns | 35.900 ns | 31.824 ns | 2.4109 |   10193 B |
| CreateFunctionSchema_CustomDescription | 6,126.56 ns | 43.308 ns | 38.391 ns | 2.3499 |    9929 B |
| CreateFunctionSchema_NewOptions        | 6,494.00 ns | 59.036 ns | 52.334 ns | 2.1973 |    9318 B |
| CreateAIFunction                       |    20.20 ns |  0.097 ns |  0.081 ns |      - |         - |
| CreateAIFunction_CustomDescription     |    28.19 ns |  0.182 ns |  0.171 ns | 0.0134 |      56 B |
| CreateAIFunction_NewOptions            | 7,309.19 ns | 72.739 ns | 64.481 ns | 2.5635 |   10854 B |

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5908)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -4,12 +4,14 @@
 using System;
 using System.Text.Json.Nodes;
 
+#pragma warning disable S1067 // Expressions should not be too complex
+
 namespace Microsoft.Extensions.AI;
 
 /// <summary>
 /// Provides options for configuring the behavior of <see cref="AIJsonUtilities"/> JSON schema creation functionality.
 /// </summary>
-public sealed class AIJsonSchemaCreateOptions
+public sealed class AIJsonSchemaCreateOptions : IEquatable<AIJsonSchemaCreateOptions>
 {
     /// <summary>
     /// Gets the default options instance.
@@ -40,4 +42,21 @@ public sealed class AIJsonSchemaCreateOptions
     /// Gets a value indicating whether to mark all properties as required in the schema.
     /// </summary>
     public bool RequireAllProperties { get; init; } = true;
+
+    /// <inheritdoc/>
+    public bool Equals(AIJsonSchemaCreateOptions? other)
+    {
+        return other is not null &&
+            TransformSchemaNode == other.TransformSchemaNode &&
+            IncludeTypeInEnumSchemas == other.IncludeTypeInEnumSchemas &&
+            DisallowAdditionalProperties == other.DisallowAdditionalProperties &&
+            IncludeSchemaKeyword == other.IncludeSchemaKeyword &&
+            RequireAllProperties == other.RequireAllProperties;
+    }
+
+    /// <inheritdoc />
+    public override bool Equals(object? obj) => obj is AIJsonSchemaCreateOptions other && Equals(other);
+
+    /// <inheritdoc />
+    public override int GetHashCode() => (TransformSchemaNode, IncludeTypeInEnumSchemas, DisallowAdditionalProperties, IncludeSchemaKeyword, RequireAllProperties).GetHashCode();
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Collections.Concurrent;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -20,7 +19,6 @@ using Microsoft.Shared.Diagnostics;
 #pragma warning disable S1075 // URIs should not be hardcoded
 #pragma warning disable SA1118 // Parameter should not span multiple lines
 #pragma warning disable S109 // Magic numbers should not be used
-#pragma warning disable S1067 // Expressions should not be too complex
 
 namespace Microsoft.Extensions.AI;
 
@@ -42,12 +40,6 @@ public static partial class AIJsonUtilities
 
     /// <summary>The uri used when populating the $schema keyword in inferred schemas.</summary>
     private const string SchemaKeywordUri = "https://json-schema.org/draft/2020-12/schema";
-
-    /// <summary>The maximum number of schema entries to cache per JsonSerializerOptions instance.</summary>
-    private const int InnerCacheSoftLimit = 512;
-
-    /// <summary>A global cache for generated schemas, weakly keyed on JsonSerializerOptions instances.</summary>
-    private static readonly ConditionalWeakTable<JsonSerializerOptions, ConcurrentDictionary<JsonSchemaCacheKey, JsonElement>> _schemaCache = new();
 
     // List of keywords used by JsonSchemaExporter but explicitly disallowed by some AI vendors.
     // cf. https://platform.openai.com/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported
@@ -73,64 +65,58 @@ public static partial class AIJsonUtilities
         serializerOptions ??= DefaultOptions;
         inferenceOptions ??= AIJsonSchemaCreateOptions.Default;
         title ??= method.Name;
+        description ??= method.GetCustomAttribute<DescriptionAttribute>()?.Description;
 
-        JsonSchemaCacheKey cacheKey = new(member: method, title, description, hasDefaultValue: false, defaultValue: null, inferenceOptions);
-        return GetOrAddSchema(serializerOptions, cacheKey, CreateSchema);
-
-        static JsonElement CreateSchema(JsonSchemaCacheKey key, JsonSerializerOptions serializerOptions)
+        JsonObject parameterSchemas = new();
+        JsonArray? requiredProperties = null;
+        foreach (ParameterInfo parameter in method.GetParameters())
         {
-            JsonObject parameterSchemas = new();
-            JsonArray? requiredProperties = null;
-            foreach (ParameterInfo parameter in ((MethodBase)key.Member!).GetParameters())
+            if (string.IsNullOrWhiteSpace(parameter.Name))
             {
-                if (string.IsNullOrWhiteSpace(parameter.Name))
-                {
-                    Throw.ArgumentException(nameof(parameter), "Parameter is missing a name.");
-                }
-
-                JsonNode parameterSchema = CreateJsonSchemaCore(
-                    type: parameter.ParameterType,
-                    parameterName: parameter.Name,
-                    description: parameter.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description,
-                    hasDefaultValue: parameter.HasDefaultValue,
-                    defaultValue: parameter.HasDefaultValue ? parameter.DefaultValue : null,
-                    serializerOptions,
-                    key.Options);
-
-                parameterSchemas.Add(parameter.Name, parameterSchema);
-                if (!parameter.IsOptional)
-                {
-                    (requiredProperties ??= []).Add((JsonNode)parameter.Name);
-                }
+                Throw.ArgumentException(nameof(parameter), "Parameter is missing a name.");
             }
 
-            JsonObject schema = new();
-            if (key.Options.IncludeSchemaKeyword)
+            JsonNode parameterSchema = CreateJsonSchemaCore(
+                type: parameter.ParameterType,
+                parameterName: parameter.Name,
+                description: parameter.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description,
+                hasDefaultValue: parameter.HasDefaultValue,
+                defaultValue: parameter.HasDefaultValue ? parameter.DefaultValue : null,
+                serializerOptions,
+                inferenceOptions);
+
+            parameterSchemas.Add(parameter.Name, parameterSchema);
+            if (!parameter.IsOptional)
             {
-                schema[SchemaPropertyName] = SchemaKeywordUri;
+                (requiredProperties ??= []).Add((JsonNode)parameter.Name);
             }
-
-            if (!string.IsNullOrWhiteSpace(key.Title))
-            {
-                schema[TitlePropertyName] = key.Title;
-            }
-
-            string? description = key.Description ?? key.Member.GetCustomAttribute<DescriptionAttribute>()?.Description;
-            if (!string.IsNullOrWhiteSpace(description))
-            {
-                schema[DescriptionPropertyName] = description;
-            }
-
-            schema[TypePropertyName] = "object"; // Method schemas always hardcode the type as "object".
-            schema[PropertiesPropertyName] = parameterSchemas;
-
-            if (requiredProperties is not null)
-            {
-                schema[RequiredPropertyName] = requiredProperties;
-            }
-
-            return JsonSerializer.SerializeToElement(schema, JsonContext.Default.JsonNode);
         }
+
+        JsonObject schema = new();
+        if (inferenceOptions.IncludeSchemaKeyword)
+        {
+            schema[SchemaPropertyName] = SchemaKeywordUri;
+        }
+
+        if (!string.IsNullOrWhiteSpace(title))
+        {
+            schema[TitlePropertyName] = title;
+        }
+
+        if (!string.IsNullOrWhiteSpace(description))
+        {
+            schema[DescriptionPropertyName] = description;
+        }
+
+        schema[TypePropertyName] = "object"; // Method schemas always hardcode the type as "object".
+        schema[PropertiesPropertyName] = parameterSchemas;
+
+        if (requiredProperties is not null)
+        {
+            schema[RequiredPropertyName] = requiredProperties;
+        }
+
+        return JsonSerializer.SerializeToElement(schema, JsonContext.Default.JsonNode);
     }
 
     /// <summary>Creates a JSON schema for the specified type.</summary>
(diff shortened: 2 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -411,6 +422,68 @@ private static int IndexOf(this JsonObject jsonObject, string key)
...
+
+        if (innerCache.Count >= InnerCacheSoftLimit)
+        {
+            return schemaFactory(cacheKey, serializerOptions);
+        }
+
+#if NET
+        return innerCache.GetOrAdd(cacheKey, schemaFactory, serializerOptions);
+#else
+        return innerCache.GetOrAdd(cacheKey, cacheKey => schemaFactory(cacheKey, serializerOptions));
```
stephentoub, line 442: This is going to add allocation on every call, even when the TryGetValue earlier has a cache hit.

We've also already done a pre-emptive TryGetValue, so the only way the key would be here is if a concurrent call added one between the TryGetValue above and here. And since we never remove, you can restructure the method to instead be something like this, without specialization for core/netfx:
```C#
    private static JsonElement GetOrAddSchema(JsonSerializerOptions serializerOptions, JsonSchemaCacheKey cacheKey, Func<JsonSchemaCacheKey, JsonSerializerOptions, JsonElement> schemaFactory)
    {
        ConcurrentDictionary<JsonSchemaCacheKey, JsonElement> innerCache = _schemaCache.GetOrCreateValue(serializerOptions);
        if (!innerCache.TryGetValue(cacheKey, out JsonElement schema))
        {
            schema = schemaFactory(cacheKey, serializerOptions);
            if (innerCache.Count < InnerCacheSoftLimit)
            {
                innerCache.TryAdd(cacheKey, schema);
            }
        }

        return schema;
}
```

stephentoub:
> Benchmark

What do the numbers before/after look like if you first add 512 methods to the cache and then benchmark calling Create with a different one? I'm wondering what the overhead of accessing the ConditionalWeakTable, accessing the ConcurrentDictionary, running the GetHashCode, etc. looks like when it doesn't pay dividends. Presumably the common case will be that across the whole process, typically the default JsonSerializerOptions will be used, so the 512 limit will end up applying to all use of AIFunctionFactory across the process.

eiriktsarpalis:
> We expect that to be something folks do, recreating an AIFunction for the same delegate over and over?

When working on the demo I felt enticed to incorporate the AIFunction declaration directly on the method like so:

```C#
chatOptions.Tools = [AIFunctionFactory.Create(GetNearestPorts)];
```

Which got me thinking that users would likely fall into the same performance trap.

> I'm wondering what the overhead of accessing the ConditionalWeakTable, accessing the ConcurrentDictionary, running the GetHashCode, etc. looks like when it doesn't pay dividends.

That scenario should be represented by the benchmarks creating a new JSO every time. As you can see, it does introduce performance regression compared to what we have in main today.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: > even when the TryGetValue earlier has a cache hit.

Can you clarify? I don't see how the wrapping delegate would be allocated if the first `TryGetValue` has found a value.

stephentoub: `cacheKey => schemaFactory(cacheKey, serializerOptions)` is closing over serializerOptions. That display class will be allocated at the beginning of the method.

stephentoub: e.g. [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0ATEBqAPgAQCYBGAWACh8BmAAiJoGEaBvCm9u2/FGgWQAoAlgDsALjQBuAQwA2AVxhoawCBBk1IwjINGCIwgJRsOrch3M1BAMxr9N23fqNmLr/AHYA3Mdc+LA/gMaAF4APjpiAE5+aXkYAwNvF3YAXz8/AAcoQWlRGAikOh4BfGJCGikwR0MWNPIUoA)

stephentoub:
> When working on the demo I felt enticed to incorporate the AIFunction declaration directly on the method like so:

Should we cache the AI Functions instead then?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: I must say I'm surprised that codegen is the way it is. Why is the closure type allocated at the start of the method when the delegate itself is allocated inside the branch?

eiriktsarpalis:
> > When working on the demo I felt enticed to incorporate the AIFunction declaration directly on the method like so:
> 
> Should we cache the AI Functions instead then?

I tried that and have a working prototype locally. What I did discover is that schema generation accounts for ~95% of the cost (as can be seen from the benchmark results) and caching on the `AIFunctionMetadata` level is trickier since there are more variables at play. It seemed like an obvious win to do it on the schema generation APIs instead.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
stephentoub: https://github.com/dotnet/roslyn/issues/20777

stephentoub:
Those numbers are very suspect to me, e.g.
```html
CreateAIFunction | 56.31 ns | 0.404 ns | 0.378 ns | 0.0172 | 72 B
```
That's suggesting with this change AIFunctionFactory.Create(delegate) only allocates 72 bytes and only took 56 nanoseconds? Without caching the AIFunction? It's doing a non-trivial amount of reflection. The AIFunction object alone must be close to 72 bytes, plus the delegates that get created for marshaling.

So I pulled down your change, and instead I see numbers like this:

| Method | Mean     | Error     | StdDev    | Allocated |
|------- |---------:|----------:|----------:|----------:|
| CreateAIFunction | 1.214 us | 0.0194 us | 0.0182 us |     816 B |

eiriktsarpalis:
Here's updated numbers from a new run:

| Method                                 | Mean        | Error      | StdDev     | Gen0   | Gen1   | Allocated |
|--------------------------------------- |------------:|-----------:|-----------:|-------:|-------:|----------:|
| CreateFunctionSchema                   |    26.01 ns |   0.093 ns |   0.087 ns |      - |      - |         - |
| CreateFunctionSchema_CustomDescription |    38.41 ns |   0.172 ns |   0.161 ns |      - |      - |         - |
| CreateFunctionSchema_NewOptions        | 7,505.62 ns | 139.908 ns | 130.870 ns | 2.3804 |      - |   10167 B |
| CreateAIFunction                       |   612.79 ns |   5.975 ns |   5.589 ns | 0.1945 |      - |     816 B |
| CreateAIFunction_CustomDescription     |   224.36 ns |   1.081 ns |   0.903 ns | 0.1452 |      - |     608 B |
| CreateAIFunction_NewOptions            | 7,657.49 ns |  61.634 ns |  54.637 ns | 2.5635 | 0.0610 |   10779 B |

And here's the same benchmarks with additional caching implemented on the AIFunction layer:

| Method                                 | Mean        | Error      | StdDev     | Gen0   | Allocated |
|--------------------------------------- |------------:|-----------:|-----------:|-------:|----------:|
| CreateFunctionSchema                   |    27.15 ns |   0.085 ns |   0.079 ns |      - |         - |
| CreateFunctionSchema_CustomDescription |    32.35 ns |   0.215 ns |   0.201 ns |      - |         - |
| CreateFunctionSchema_NewOptions        | 7,648.95 ns | 140.465 ns | 196.912 ns | 2.4109 |   10167 B |
| CreateAIFunction                       |    60.11 ns |   0.307 ns |   0.287 ns | 0.0172 |      72 B |
| CreateAIFunction_CustomDescription     |    66.00 ns |   0.256 ns |   0.227 ns | 0.0305 |     128 B |
| CreateAIFunction_NewOptions            | 8,604.40 ns |  60.222 ns |  53.385 ns | 2.7466 |   11692 B |

So the numbers I quoted originally were most likely recorded with those changes accidentally unstashed. I've updated the OP to reflect the correct numbers.

eiriktsarpalis:
@stephentoub as discussed, I pushed updates using caching on the AIFunctionFactory layer only. Here are the updated benchmark numbers:

| Method                                 | Mean        | Error     | StdDev    | Gen0   | Allocated |
|--------------------------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateFunctionSchema                   | 6,679.98 ns | 35.900 ns | 31.824 ns | 2.4109 |   10193 B |
| CreateFunctionSchema_CustomDescription | 6,126.56 ns | 43.308 ns | 38.391 ns | 2.3499 |    9929 B |
| CreateFunctionSchema_NewOptions        | 6,494.00 ns | 59.036 ns | 52.334 ns | 2.1973 |    9318 B |
| CreateAIFunction                       |    20.20 ns |  0.097 ns |  0.081 ns |      - |         - |
| CreateAIFunction_CustomDescription     |    28.19 ns |  0.182 ns |  0.171 ns | 0.0134 |      56 B |
| CreateAIFunction_NewOptions            | 7,309.19 ns | 72.739 ns | 64.481 ns | 2.5635 |   10854 B |

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.Utilities.cs
```diff
@@ -30,4 +31,25 @@ internal static string SanitizeMemberName(string memberName)
...
+        {
+            return method.Invoke(target, BindingFlags.Default, binder: null, arguments, culture: null);
+        }
+        catch (TargetInvocationException e) when (e.InnerException is not null)
+        {
+            // If we're targeting .NET Framework, such that BindingFlags.DoNotWrapExceptions
+            // is ignored, the original exception will be wrapped in a TargetInvocationException.
+            // Unwrap it and throw that original exception, maintaining its stack information.
+            System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(e.InnerException).Throw();
+            return null;
```
stephentoub, line 51: Nit:
```suggestion
            throw;
```

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -432,42 +450,38 @@ static bool IsAsyncMethod(MethodInfo method)
...
-                // Unwrap it and throw that original exception, maintaining its stack information.
-                System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(e.InnerException).Throw();
-                return null;
+                if (returnTypeInfo.Kind is JsonTypeInfoKind.None)
+                {
+                    // Special-case trivial contracts to avoid the more expensive general-purpose serialization path.
+                    return JsonSerializer.SerializeToElement(result, returnTypeInfo);
+                }
+
+                // Serialize asynchronously to support potential IAsyncEnumerable responses.
```
stephentoub, line 476: Should we add to STJ a SerializeToElementAsync?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -432,42 +450,38 @@ static bool IsAsyncMethod(MethodInfo method)
...
+                using MemoryStream stream = new();
```
stephentoub, line 477: Should we be looking at avoiding the underlying buffer allocations via using a stream that uses pooled buffers?

stephentoub (approved):
Nice, thanks.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -40,4 +42,21 @@ public sealed class AIJsonSchemaCreateOptions
...
+            DisallowAdditionalProperties == other.DisallowAdditionalProperties &&
+            IncludeSchemaKeyword == other.IncludeSchemaKeyword &&
+            RequireAllProperties == other.RequireAllProperties;
+    }
+
+    /// <inheritdoc />
+    public override bool Equals(object? obj) => obj is AIJsonSchemaCreateOptions other && Equals(other);
+
+    /// <inheritdoc />
+    public override int GetHashCode() => (TransformSchemaNode, IncludeTypeInEnumSchemas, DisallowAdditionalProperties, IncludeSchemaKeyword, RequireAllProperties).GetHashCode();
```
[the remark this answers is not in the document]
eiriktsarpalis: We can, just not on netstandard. I felt this is acceptable over duplicating the implementation.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.Utilities.cs
eiriktsarpalis: Isn't that redundant given that the previous line always throws?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -278,30 +210,111 @@ static bool IsAsyncMethod(MethodInfo method)
                 args[i] = paramMarshallers[i](argDict, context);
             }
 
-            object? result = await _returnMarshaller(ReflectionInvoke(_method, _target, args)).ConfigureAwait(false);
+            return await FunctionDescriptor.ReturnParameterMarshaller(ReflectionInvoke(FunctionDescriptor.Method, Target, args), cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
eiriktsarpalis: It can't because it needs to also handle potentially asynchronous serialization of the final result (e.g. when it's IAsyncEnumerable)

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.Utilities.cs
stephentoub: No more redundant than `return null;`, which will also never be reached. But using `throw` makes it easy to see at a glance that you never exit this catch block successfully.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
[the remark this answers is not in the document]
stephentoub: I'm not referring to ReturnParameterMarshaller, I'm referring to InvokeCoreAsync. It has a single await at the very end. Can it just return the result of invoking ReturnParameterMarshaller rather than awaiting the result?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: It's possible. Currently the only async methods we expose target `Stream` although we could conceivably expose variants for all other encoding mediums.

PR #5926, opened 2025-02-18
Make sure npm build tasks happen reliably and only once.

peterwald:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5926)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
```diff
@@ -1,32 +1,42 @@
...
 
   <PropertyGroup>
-    <_TypescriptRootPath>$(MSBuildThisFileDirectory)/TypeScript</_TypescriptRootPath>
-    <_NpmInstallStampFile>$(_TypescriptRootPath)/node_modules/.installed</_NpmInstallStampFile>
+    <_TypescriptRootPath>$(MSBuildThisFileDirectory)TypeScript</_TypescriptRootPath>
+    <_NpmInstallStampFile>$(_TypescriptRootPath)\node_modules\.installed</_NpmInstallStampFile>
   </PropertyGroup>
 
   <Target Name="NpmInstall"
-          Condition="'$(EnableEvaluationReportBuild)'=='true'"
```
shyamnamboodiripad, line 10: Was removing the condition intentional? Won't this fail on machines where node is missing?

shyamnamboodiripad: Ah looks like `BuildTypeScriptBundle` is still conditioned and looks like this target can only be called via that one... So this is probably fine.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
```diff
@@ -1,32 +1,42 @@
...
   </Target>
 
+  <!-- Make sure the typescript build is up to date before the the multi-targetted builds. -->
+  <Target Name="_BuildTypeScriptBundleBeforeOuterBuild"
+    DependsOnTargets="BuildTypescriptBundle"
+    BeforeTargets="DispatchToInnerBuilds"
+  />
+
+  <!-- Delegate inner builds back to the outer build in case it tries to finish before the outer build -->
+  <Target Name="_BuildTypeScriptBundleBeforeInnerBuild"
```
shyamnamboodiripad, line 33: Any reason why some of the target names begin with `_` while others don't? Should all of them begin with `_` (including `GenerateVersionConstant` below)? If yes, please also fix this in the `Directory.Build.targets` file in the `MEAI.Extensions.Console` project.

shyamnamboodiripad (approved):
👍🏾 modulo a couple of questions in the comments

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
```diff
@@ -1,32 +1,42 @@
 <Project>
-  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)..\'))" />
```
shyamnamboodiripad, line 2: Looks like same issue also exists in the below `Directory.Build.props` which was copied from this one. Could you please fix that one too?

https://github.com/shyamnamboodiripad/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Directory.Build.targets#L2

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
```diff
@@ -1,32 +1,42 @@
...
-          Inputs="$(_TypescriptRootPath)/**/*.*"
-          Outputs="$(_TypescriptRootPath)/html-report/dist/index.html">
+          Inputs="$(_TypescriptRootPath)\**\*.*"
+          Outputs="$(_TypescriptRootPath)\html-report\dist\index.html">
     <Message Text="Build the Typescript bundle" Importance="high" />
     <Exec Command="node build-if-out-of-date.js" WorkingDirectory="$(_TypescriptRootPath)" />
   </Target>
 
+  <!-- Make sure the typescript build is up to date before the the multi-targetted builds. -->
+  <Target Name="_BuildTypeScriptBundleBeforeOuterBuild"
```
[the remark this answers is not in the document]
RussKie: If all the targets and property definitions are conditioned on `'$(EnableEvaluationReportBuild)'=='true'` then perhaps it's worth considering renaming this file to something like `EvaluationReportBuild.targets` and conditionally importing it from the csproj. This way you'd be able to simplify implementations in this file.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
peterwald: Yes it was intentional since it was not necessary.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
peterwald: Yes this is intentional. The `_` prefix is a way of indicating that these targets are "private". I made these targets start with `_` since they are essentially infrastructure targets and not useful on their own. This same pattern is commonly used through the .NET SDK.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/Directory.Build.targets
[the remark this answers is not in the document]
peterwald: > Should we move the `Condition="'$(EnableEvaluationReportBuild)'=='true'"` from the above target to `_BuildTypeScriptBundleBeforeOuterBuild` and `_BuildTypeScriptBundleBeforeInnerBuild` below since they are the main entry points? It is a bit confusing that a target in the middle of the chain is the only one that short-circuits. Short-circuiting the entry points could potentially make some future binlog investigations less confusing.

Maybe I'm in the minority, but personally I find it less confusing to have the condition on the `BuildTypeScriptBundle` task since that is the main one that is doing all the work and it is the one people are going to look to first to see what is going on. The others are only there as plumbing to tie the target into the rest of the build. I don't feel super strongly about it either way, but that is my rationale.

RussKie (approved):

test/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Tests/EmbeddingTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.Extensions.AI.Evaluation.Reporting.Tests;
+
+public class EmbeddingTests
+{
+    [ConditionalFact]
+    public void CIBuildsMustIncludeEmbeddedHTML()
+    {
+        // TF_BUILD should be set in our CI pipeline
+        if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TF_BUILD")))
```
shyamnamboodiripad, line 17: Should we say `if (!Settings.Current.Configured)` instead? I'd assume that we also want the test to fail locally if this happens to be missing?

peterwald: I thought about this. I couldn't think of a way to have it work locally and also stay out of the way when people don't have EnableEvaluationReportBuild enabled.

shyamnamboodiripad: Could we perhaps also require them to pass when `Settings.Configured` is true?

peterwald: I'm happy to iterate on this, but let's do that in another PR

test/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Tests/EmbeddingTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            throw new SkipTestException("Skipping test because it is not running in CI");
+        }
+
+        Assert.NotEmpty(HtmlReportWriter.HtmlTemplateBefore);
+        Assert.NotEmpty(HtmlReportWriter.HtmlTemplateAfter);
```
shyamnamboodiripad, line 23: minor: It may be better to actually test this in a more black-box way by cracking open the assembly and reading the resource string... (Just in case we change how the static fields work in the future but forget to update / remove this test - for example if we switch to using a default / hardcoded template for some reason...)

I don't feel too strongly about this - but if you do end up changing this now, please flip the fields back to private.

peterwald: I'm going to leave it for now, but I think we can continue to iterate.  One other thought I had was to add a command line option that would check the embedded data when running `aieval version`  for example. Then we could run the console command from the pipeline to see if it succeeds. The only problem with this is that console packages multiple TFMs and I don't know how to invoke each TFM separately without installing multiple SDKs.

shyamnamboodiripad: One idea may be to add a test under `Console.Tests` to programmatically invoke the corresponding command in the `aieval` tool. This test would then run for all TFMs for which the tool is built.

Issue #5888, opened 2025-02-12
[AI Evaluation] Explore use of structured output for eval prompts


shyamnamboodiripad:
Evaluators such as `RelevanceTruthAndCompletenessEvaluators`, `FluencyEvaluators` currently include prompts that instruct the LLM to produce evaluation ratings in a certain fixed format (e.g. JSON, OR a single character output when the evaluation score to be returned is a single integer).

Loging this issue to explore use of structured output and `IChatClient.GetResponseAsync<T>` to reinforce the above prompts and improve their overall effectiveness.

shyamnamboodiripad:
FYI @peterwald @stephentoub

shyamnamboodiripad:
@stephentoub I tried this out for the two use cases we have under [Microsoft.Extensions.AI.Evaluation.Quality](https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality).

In the first case, we already prompt the LLM to return JSON (see [`RelevanceTruthAndCompletenessEvaluator.cs`](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/RelevanceTruthAndCompletenessEvaluator.Prompts.cs#L15C98-L15C121)). `GetResponseAsync<MyJsonType>` works great in this case!

In the second case, we prompt the LLM to return a single `int` value (see [`SingleNumericMetricEvaluator.cs`](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/SingleNumericMetricEvaluator.cs#L45-L50)). In this case, `GetResponseAsync<int>` does not work. Looks like the linked system prompt instructions conflict with the JSON schema included by `GetResponseAsync<T>`. We could change these prompts to require JSON (similar to `RelevanceTruthAndCompletenessEvaluator` above). However, the single `int` response would be more compact (fewer tokens) and potentially less error prone (although that may be model dependent).

❔Is there some way to support primitive types like `int` for `GetResponseAsync<T>` or do you think it is better to switch these prompts to require JSON?

[...]

PR #5945, opened 2025-02-19
Use structured output for RTC evaluator

shyamnamboodiripad:
Also change the base types so that derived types have more control over whether or not structured output should be used.

Also adds some tests to ensure that the RTC works well both when the returned JSON contains reasoning elements as well as if it does not.

Addresses #5888
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5945)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/SingleNumericMetricEvaluator.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/SingleNumericMetricEvaluator.cs
@@ -24,18 +24,6 @@ public abstract class SingleNumericMetricEvaluator : ChatConversationEvaluator
     protected abstract string MetricName { get; }
 
     /// <inheritdoc/>
-    protected override ChatOptions? ChatOptions { get; } =
-        new ChatOptions
-        {
-            MaxOutputTokens = 1,
-            Temperature = 0.0f,
-            TopP = 1.0f,
-            PresencePenalty = 0.0f,
-            FrequencyPenalty = 0.0f,
-            ResponseFormat = ChatResponseFormat.Text
-        };
-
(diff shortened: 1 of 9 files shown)


shyamnamboodiripad:
FYI @andhesky

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/SingleNumericMetricEvaluator.cs
```diff
@@ -49,8 +37,16 @@ Your response should be a single character containing only the evaluation score.
...
-    // TASK: Explore using structured output and providing a JSON schema to better enforce the LLM response format
-    // requirements above. Tracked by https://github.com/dotnet/extensions/issues/5888.
+    private readonly ChatOptions _chatOptions =
+        new ChatOptions
+        {
+            MaxOutputTokens = 1,
+            Temperature = 0.0f,
+            TopP = 1.0f,
+            PresencePenalty = 0.0f,
+            FrequencyPenalty = 0.0f,
```
stephentoub, line 47: Have you found that setting these is helpful? Are these not the default values used in most services?

shyamnamboodiripad: That's a good question. `MaxOutputTokens` is certainly useful in this context. The rest were copied over from the original python-based implementation. I am not sure what the defaults for these parameters are, and whether the defaults are the same across different models / providers. Most of our real-world usage / testing has been against Azure Open AI, and perhaps some of these are indeed redundant. However, given that the deployed evaluations have been working well, I'd rather not change it at this point unless there is a strong reason to...

PR #5949, opened 2025-02-20
Add `Microsoft.Extensions.AI.Templates` third party notices

MackinnonBuck:
Adds a `THIRD-PARTY-NOTICES.TXT` file to the `Microsoft.Extensions.AI.Templates` NuGet package.

In addition to the JS dependencies directly included in the template, I've added notices for 3rd party .NET packages used by the template, even though the template doesn't directly include content from those packages. We can remove those if we don't think it's necessary to include them.

Fixes https://github.com/dotnet/ai-private-planning/issues/259
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5949)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/Microsoft.Extensions.AI.Templates.csproj
```diff
@@ -24,6 +24,7 @@
 
   <ItemGroup>
     <Content Include="src\ChatWithCustomData\**\*" Exclude="**\bin\**;**\obj\**;**\node_modules\**;**\*.user;**\*.csproj.in;**\*.out.js;**\*.generated.css;**\package-lock.json" />
+    <None Include="THIRD-PARTY-NOTICES.TXT" Pack="true" PackagePath="." />
```
jmatthiesen, line 27: Do we typically include these with .NET templates? Looks like this will be included in the project once created, right?

MackinnonBuck: It will be included in the NuGet package but not included as content in the generated template (similar to the `README.md` in the same folder).

MackinnonBuck: The same thing is done in, e.g., [Microsoft.AspNetCore.SignalR.Client](https://nuget.info/packages/Microsoft.AspNetCore.SignalR.Client/9.0.2).

jmatthiesen: Ah, OK. Perfect

SteveSandersonMS (approved):
Looks great, thanks!

I don't know for sure whether we need to include copies of licenses for things we only consume via NuGet package references and don't redistribute ourselves. @jeffhandley do you know?

jeffhandley:
> I don't know for sure whether we need to include copies of licenses for things we only consume via NuGet package references and don't redistribute ourselves. @jeffhandley do you know?

I don't think so, since we're not distributing it. I'll double-check though.

richlander:
It depends on the vantage point. It's hard to tell since this has something to do with templates.

In the normal case, we don't add 3PN entries for components we don't ship.

For templates, it would be useful to the user to have a 3PN file that reflects their app. That would include the packages.

jeffhandley (approved):

PR #5956, opened 2025-02-22
Enable AIFunctionFactory to resolve parameters from an IServiceProvider

stephentoub:
In a discussion last week, @eiriktsarpalis suggested that sooner or later we'll want to support AIFunctionFactory being able to populate parameters from DI. The second commit here is how I think we could do that. (Ignore the first commit, it's just https://github.com/dotnet/extensions/pull/5954.) What do we think? Is this something we'd want to proceed with, or if not, are we comfortable with this as a future solution, in which case this demonstrates it's possible without breakage?

Effectively, it enables someone to write:
```C#
AIFunctionFactory.Create(([FromServiceProvider] WhateverType wt = null) =>
{
    ...
});
```
ala what you can do with MVC and SK and others. When constructing a FunctionInvokingChatClient, it'll hold onto whatever IServiceProvider was used to construct it, and then feed that into AIFunction.InvokeAsync via an AIFunctionArguments type that AIFunctionFactory type checks for.

cc: @eiriktsarpalis, @SteveSandersonMS, @halter73 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5956)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -76,6 +77,22 @@ public static JsonElement CreateFunctionJsonSchema(
...
 
+            if (parameter.ParameterType == typeof(CancellationToken))
+            {
+                // CancellationToken is a special case that, by convention, we don't want to include in the schema.
+                // Invocations of methods that include a CancellationToken argument should also special-case CancellationToken
+                // to pass along what relevant token into the method's invocation.
+                continue;
+            }
+
+            if (parameter.GetCustomAttribute<FromServiceProviderAttribute>(inherit: true) is not null)
```
eiriktsarpalis, line 88: This is introducing a concern that is specific to AIFunctionFactory infrastructure. I wonder if it's further evidence that the two components should be merged as one.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,84 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Shared.Collections;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Provides context for an in-flight function invocation.</summary>
+public sealed class FunctionInvocationContext
```
eiriktsarpalis, line 12: I take it this change is based on top of the other PR branch you have open atm?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -320,40 +318,43 @@ static bool IsAsyncMethod(MethodInfo method)
...
+                            ksp.GetKeyedService(parameterType, serviceKey) is object keyedService)
+                        {
+                            return keyedService;
+                        }
+                    }
+                    else if (services.GetService(parameterType) is object service)
+                    {
+                        return service;
+                    }
+                }
```
eiriktsarpalis, line 356: I'm not sure if I'm a fan of the marshaller falling through to deserialization if no service has been found for the attributed parameter. It's not part of the advertised schema so no LLM would be able to provide valid JSON for it (and that's assuming that the service is JSON-serializable). I get that the parameter _could_ still be passed in assuming the object is an instance of the parameter type, but I don't think any canonical use case for AIFunction would be able to do this.

For the sake of simplicity then, I would prefer it if these parameters became unambiguously a DI dependency (aka use `GetRequiredService` in a separate marshaller).

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -320,40 +318,43 @@ static bool IsAsyncMethod(MethodInfo method)
...
-            // Create a marshaller that simply looks up the parameter by name in the arguments dictionary.
-            return (arguments, _) =>
+            // Create a marshaller for the parameter. This produces a value for the parameter based on an ordered
+            // collection of rules.
+            return (arguments, cancellationToken) =>
             {
+                // If the parameter is [FromServiceProvider], try to satisfy it from the service provider
+                // provided via arguments.
+                if (fspAttr is not null &&
+                    (arguments as AIFunctionArguments)?.ServiceProvider is IServiceProvider services)
```
eiriktsarpalis, line 342: Couldn't we just have `InvokeAsync` accept an optional `IServiceProvider` parameter?

src/Libraries/Microsoft.Extensions.AI.Abstractions/FromServiceProviderAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Indicates that a parameter to an <see cref="AIFunction"/> should be sourced from an associated <see cref="IServiceProvider"/>.</summary>
+[AttributeUsage(AttributeTargets.Parameter)]
+public sealed class FromServiceProviderAttribute : Attribute
```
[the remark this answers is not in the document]
stephentoub: There's one in ASP.NET's MVC library. I'm not aware of one in M.E.DI.

I would like:
- To move the attr up to M.E.AI
- To move the schema support up to M.E.AI.
- Then somehow factory AIFF into the name.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvocationContext.cs
stephentoub: Yes, it's mentioned in the PR description.

> Ignore the first commit, it's just https://github.com/dotnet/extensions/pull/5954.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
stephentoub: This PR is showing how it can be added later, and changing InvokeAsync can't be done later. I also think including IServiceProvider in InvokeAsync is an arbitrary additional piece of information that doesn't belong; nothing about AIFunction is related to DI; that's a higher level concern.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
stephentoub: Note that this PR also highlights that AIFunctionFactory should not be moved down. It needs to reference IKeyedServiceProvider, which is in the DI assembly, and we won't be adding the DI assembly as a dependency of the abstractions lib.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -58,11 +58,13 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
     /// </summary>
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
     /// <param name="logger">An <see cref="ILogger"/> to use for logging information about function invocation.</param>
-    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null)
+    /// <param name="services">An optional <see cref="IServiceProvider"/> to use for resolving services required by the <see cref="AIFunction"/> instances being invoked.</param>
+    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null, IServiceProvider? services = null)
         : base(innerClient)
     {
-        _logger = logger ?? NullLogger.Instance;
+        _logger = logger ?? (ILogger?)services?.GetService(typeof(ILogger<FunctionInvokingChatClient>)) ?? NullLogger.Instance;
```
halter73, line 65:
```suggestion
        _logger = logger ?? NullLogger.Instance;
```

Nit: falling back to using `GetService` for the logger is weird especially when you can provide the logger seperately. What if someone wanted to prevent `FunctionInvokingChatClient` from logging but still offer the service provider via arguments to their functions? It feels like an orthogonal concern. People could obviously adjust log levels for the appropriate category in their configuration, but what if they only want to disable logging for certain `FunctionInvokingChatClient` instances?

Also, it's odd for something that can be registered/resolved as a service to take a plain `ILogger` parameter. I see that [FunctionInvokingChatClientBuilderExtensions makes the appropriate `CreateLogger`](https://github.com/dotnet/extensions/blob/576fa221f57f80423a4852e446cee2aa2e937aab/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClientBuilderExtensions.cs#L35) for you when wiring this up, but any service type that was registered directly with DI would need to take either an `ILogger<FunctionInvokingChatClient>` or `ILoggerFactory` for it to work.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FromServiceProviderAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>Initializes a new instance of the <see cref="FromServiceProviderAttribute"/> class.</summary>
+    /// <param name="serviceKey">Optional key to use when resolving the service.</param>
+    public FromServiceProviderAttribute(object? serviceKey = null)
+    {
+        ServiceKey = serviceKey;
+    }
+
+    /// <summary>Gets the key to use when resolving the service.</summary>
+    public object? ServiceKey { get; }
```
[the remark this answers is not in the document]
stephentoub: Is there a reason to have two different attributes rather than one that serves both purposes?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -340,6 +346,25 @@ static bool IsAsyncMethod(MethodInfo method)
...
+                {
+                    if (fspAttr.ServiceKey is object serviceKey)
+                    {
+                        if (services is IKeyedServiceProvider ksp &&
+                            ksp.GetKeyedService(parameterType, serviceKey) is object keyedService)
```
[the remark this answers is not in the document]
stephentoub: > Would we also want to infer whether or not the service is required based on whether or not it has a default value (i.e. = null)?

Eirik also suggested in another comment that he doesn't like the fallback behavior I have here. If we get rid of that, then we'd make the implementation throw (effectively GetRequiredService) if the parameter wasn't optional, and use the default value if it was.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FromServiceProviderAttribute.cs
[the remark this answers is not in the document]
halter73: Mostly because I think people would expect it to work. We support it in minimal APIs, MVC and SignalR for example. See https://source.dot.net/#Microsoft.Extensions.DependencyInjection.Abstractions/FromKeyedServicesAttribute.cs,662de84ebb722cbd,references for a bunch of places we reference it outside of DI internals.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
halter73: I don't think I've ever seen a method or constructor with both an `IServiceProvider` parameter and another parameter for a specific service that falls back to the service provider like this, so I wouldn't expect it just from the signature, but I can see it being convenient if you're manually newing up a `FunctionInvokingChatClient`. I'm fine with leaving the fallback there. You could always pass in `NullLogger.Instance` if you really don't like the fallback.

> We could change the latter to be ILoggerFactory if you think that's important.

On second thought. I think you can leave it as is given the design is never to register this type as a service directly. One upside of taking an `ILogger` over an `ILoggerFactory` is that people can pick their own category name if that care to.

tl;dr: leave everything as-is.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FromServiceProviderAttribute.cs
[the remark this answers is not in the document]
stephentoub: Just seems unnecessary to have two different attributes here. Why did we design it that way for asp.net? `[FromServices("key")]`  seems quite clear. Is it ambiguous somehow that it needs "keyed" in the name to differentiate it somehow?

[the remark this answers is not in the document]
stephentoub: Oh, wait, FromKeyedServices is in the DI library but FromServices is not?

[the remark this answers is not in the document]
stephentoub: (I'm also not sure how to reasonably support that attribute when the schema code in M.E.AI.Abstractions would need to know about it.)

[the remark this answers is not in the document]
stephentoub: I think we'd need to add a mechanism to CreateFunctionJsonSchema overload taking a MethodInfo that would be a list of the parameters to include or exclude. Then we could get rid of the base Skip attribute, and AIFunctionFactory could call CreateFunctionJsonSchema passing in the parameters it cares about. @eiriktsarpalis?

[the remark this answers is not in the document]
eiriktsarpalis: Sounds reasonable, although for caching purposes the schema options type is IEquatable so any filter/allow list mechanism should be made cache friendly. I _think_ exposing it as a simple predicate `Func<ParameterInfo, bool>` should suffice so long as we don't need to pass it any additional state.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -343,28 +342,40 @@ static bool IsAsyncMethod(MethodInfo method)
...
+            if (parameter.GetCustomAttribute<FromServiceProviderAttribute>(inherit: true) is FromServiceProviderAttribute fspAttr)
             {
-                // If the parameter is [FromServiceProvider], try to satisfy it from the service provider
-                // provided via arguments.
-                if (fspAttr is not null &&
-                    (arguments as AIFunctionArguments)?.ServiceProvider is IServiceProvider services)
+                return (arguments, _) =>
                 {
-                    if (fspAttr.ServiceKey is object serviceKey)
+                    if ((arguments as AIFunctionArguments)?.ServiceProvider is IServiceProvider services)
```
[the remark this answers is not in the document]
eiriktsarpalis: One possibility I considered is using a dependency inversion scheme where the `InvokeAsync` method accepts a `Func<Type, object>` either directly or through an "AIFunctionInvocationContext" type that contains it as a property.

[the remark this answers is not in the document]
stephentoub: > consider requiring that input dictionaries are always of this type regardless?

I'd rather not, both because it limits what input types are possible but also because it makes the IServiceProvider a prominent thing that's part of the base abstraction, when it's not really.

> Making it optional does seem like a pit of failure for library authors whose test coverage doesn't include DI scenaria. Should we consider requiring that input dictionaries are always of this type regardless?

I wouldn't expect this to be an issue for most library authors. What are these libraries doing? The only component that would need to instantiate this type to flow additional data through would be something that's invoking arbitrary AIFunction instances, ala FunctionInvokingChatClient.

I agree it's not ideal. I just don't have a better answer that keeps AIFunction.InvokeAsync simple and clean while still allowing for other ambient context to be passed through. If we want to change InvokeAsync now, before it's stable, options would include changing the type of the argument, adding an additional dictionary argument for ambient context in addition to named arguments, adding some kind of callback argument that the AIFunction could use to retrieve the value for an argument, e.g. `Func<string, object?>? argumentFactory` (where we establish a search order between the dictionary and the callback), etc. I'm not sure any of those is better, though I'm open to arguments (no pun intended).

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
BrennanConroy, line 65: Could you update this to
```suggestion
        _logger = logger ?? services?.GetService<ILogger<FunctionInvokingChatClient>>() ?? NullLogger.Instance;
```
(need to add DependencyInjection.Abstractions namespace to usings)

stephentoub: > Could you update this to

I can use the generic GetService, but I can't remove the cast; C# would complain about the types there with the ??

stephentoub:
@SteveSandersonMS, could you review this as well?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FromServicesAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+
+/// <summary>Indicates that a parameter to an <see cref="AIFunction"/> should be sourced from an associated <see cref="IServiceProvider"/>.</summary>
+[AttributeUsage(AttributeTargets.Parameter)]
+public sealed class FromServicesAttribute : Attribute
+{
+    /// <summary>Initializes a new instance of the <see cref="FromServicesAttribute"/> class.</summary>
+    public FromServicesAttribute()
+    {
+    }
+}
```
SteveSandersonMS, line 16: I guess it's a challenge with .NET in general, but we do already have [a type with this name](https://github.com/dotnet/aspnetcore/blob/1f82fa558dd7afb0ac94a43c7b80f40499bad98c/src/Mvc/Mvc.Core/src/FromServicesAttribute.cs#L26) and of course the same meaning. This could lead to some pretty awkward debugging for people who believe they are using `[FromServices]` correctly but it's not working (either breaking MVC code or breaking MEAI code).

Possible solutions:

 * Use a different name here
 * Try to push MVC's `FromServicesAttribute` up into `System.ComponentModel` or something
 * Use a different pattern here

If we wanted a different pattern here, we could embrace the whole "AIFunctionCallContext" parameter notion. It could deal with services and cancellation tokens in one, and would make it easy for us to add any other per-call context in the future without having to special-case any other parameter types. I know we considered that before, but maybe adding further per-call context starts to tip the balance another way.

What do you think? Is this enough of a problem to warrant a different approach?

SteveSandersonMS: I've now read the other comments on this PR and see there's already been talk of this. Looks like one blocker is that we would need AIFF etc to move from M.E.AI.Abstractions to M.E.AI. But is there a plan to do that regardless?

If we could avoid creating a duplicate-named attribute that certainly sounds preferable to me. I'm surprised this didn't seem to bother @halter73 so maybe I'm overblowing the issue. What do you think, @halter73?

stephentoub: > But is there a plan to do that regardless?

No, that would require M.E.AI.Abstractions depending on DI, which I want to avoid.

stephentoub: > name

I'd started with a different name and was urged by Stephen and others to use the same one. Happy to go back to a different one if folks can agree. 

> Try to push MVC's FromServicesAttribute up into System.ComponentModel or something

It'd be really weird to depend on something in the Microsoft.AspNetCore.Mvc name space, even if we could figure out how to move it into a different assembly on an appropriate time line.

Just to confirm, you're factoring into your concerns that it's in an MVC namespace? 

> we could embrace the whole "AIFunctionCallContext" parameter

Got a lot of push back on that during API review, especially around CT... we won't be putting the CT into such a thing even if we add it back, it's too hard to discover and doesn't match . LNET conventions. But I'm also not clear how that helps, unless you're suggesting it would just carry an IServiceProvider and the ap code would query directly? We could also do that just by special-casing IServiceProvider if we wanted to.

SteveSandersonMS: > Just to confirm, you're factoring into your concerns that it's in an MVC namespace?

Yes, I do know that. It means we're relying on the developer understanding that, in a typical ASP.NET Core application, there are two different `[FromServices]` attributes and that they have to get the right one in each case, with the compiler/IDE not helping identify the correct one. Picking the wrong one means it just won't work at runtime.

> But I'm also not clear how that helps, unless you're suggesting it would just carry an IServiceProvider and the ap code would query directly?

Yes, that's what I meant.

> We could also do that just by special-casing IServiceProvider if we wanted to.

That's a perfectly fine solution as long as we consider this a relatively niche case.

If we thought that, say, 50%+ of AIFunctions would need to obtain DI services directly (and couldn't acquire them from context, e.g., by being instance methods on some type that already has access to the desired services, which I think is also common), then it would make sense to try to make it more first-class.

One option is, for now, to support `IServiceProvider` parameters and thus bypass the other issues. It would still leave open the option to add other ways to resolve params-as-services in the future.

Altogether this isn't something that worries me deeply. If you and others are keen to have `[FromServices]` and don't think the duplicate type name will be an issue, that's OK.

stephentoub:
The purpose of this PR was to demonstrate that we can add such functionality without taking any breaking changes, and we can (there'd be a tiny possible behavioral breaking change, in that you could have a parameter attributed as [FromKeyedService] and this functionality would change how that's handled, but if you're doing that, it's not behaving today how you almost certainly expect it to). We can either expose it as is done in this PR, or @SteveSandersonMS also suggested it could be simplified to just special-casing IServiceProvider in AIFunctionFactory (similarly, there'd be a small behavioral change, in that we'd start preferring to pull that from DI and excluding it from the schema, but AI can't do anything reasonable to populate an IServiceProvider instance, anyway).

Given that, I'll close this for now until there's a demonstrated need.

Issue #5962, opened 2025-02-24
AddResourceMonitoring() do no-operation if not supported OS


juliankock:
### Description

I'm currently working on MacOS and it seems like https://source.dot.net/#Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs,7ba4439b710f2460,references throws an exception if the OS is not supported.

I would prefer it did a no-op. We go through all OS' and we need to do:
```
if (!OperatingSystem.IsMacOS())
{
    builder.Services.AddResourceMonitoring();
}
```
In the calling code.

### Reproduction Steps

Run `builder.Services.AddResourceMonitoring();`.

### Expected behavior

Just do no-op and return early if not supported OS. 

### Actual behavior

Throws exception

### Known Workarounds

Add if statement at consuming code

evgenyfedorov2:
Can you please explain your use case in more details, please?
Otherwise, the proposed change would be misleading for users, the library does not indeed support MacOs

evgenyfedorov2:
>  I do believe the docs state clearly that the metrics are only supported by _MacOS and Linux_, therefore consumers should expect it not to appear as metrics.

Did you mean _Windows and Linux_?

> However, running on a non-Windows or Linux host is quite common when developing .NET, so I wouldn't say it's exception worthy.

This exception is about being explicit and return early.

> Another approach would be to add to the docs that it throws a PlatformNotSupportedException in the XML comment, to clearly state that it is thrown.

Can we go this route instead, please?

evgenyfedorov2:
I sometimes develop on a MacOS as well and suffer from not being able to run unit tests for this library, even though most of them are mock-based, e.g. do not technically require a specific OS.

Having said that, if you have time, feel free to propose and, possibly, implement another change, which would include removal of the PlatformNotSupportException on MacOs and unit test refactoring (where necessary) so that as many tests as possible would run on MacOS. Otherwise, I will try do it myself sooner or later

martincostello:
#5963 fixes an exception being thrown immediately by `AddResourceMonitoring()`, but instead defers it to when the service starts, which breaks integration-style tests with `WebApplicationFactory<T>`.

```text
 [2026-08-26 17:04:37Z] fail: Microsoft.Extensions.Hosting.Internal.Host[11]
       Hosting failed to start
[...]

PR #5963, opened 2025-02-24
#5962 Change to early return of OS instead of throwing

juliankock:
This will make sure that the executed code only works in scenarios that are supported, which is Windows and Linux. This is particularly important for teams working on different operating systems, and running the code without containerization.

Addresses: #5962 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5963)


vidarasberg:
I am looking forward to this change as I too am on a mac and am experiencing the same issue.

Why is this only a draft? It looks good to me :)

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
```diff
@@ -63,13 +63,17 @@ private static IServiceCollection AddResourceMonitoringInternal(
         this IServiceCollection services,
```
evgenyfedorov2, line 63: This method now can be fully covered by unit tests, please remove the comment and `ExcludeFromCodeCoverage `attribute

evgenyfedorov2:
> @evgenyfedorov2 It passes tests now. Is there anything we would like more?

Yeah, please just add more tests, I left a comment.

marcopelegrini (approved):

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringServiceCollectionExtensions.cs
evgenyfedorov2: Added

evgenyfedorov2:
@juliankock please accept CLA and then we can merge

juliankock:
@dotnet-policy-service agree

PR #5966, opened 2025-02-25
Finalizing template name and sample data

jmatthiesen:
This update includes a final set of sample PDFs to use for the first Preview of the template:
* Added two sample PDFs from the eShopSupport project
* Added a note in each PDF to clarify that they were generated by AI for testing purposes
* Updated Chat.razor with an initial "welcome message" when running the app to help the user understand the scenario and think of some prompts.

image

This also includes a final name for this preview release that reflects the Chat use case.


SteveSandersonMS (approved):
Excellent, @jmatthiesen!

I really like how you made the landing page look before the chat starts. Much clearer what's going on now.

BTW I added a couple of commits because after merging the latest into this branch, PDF file paths had changed.

PR #5968, opened 2025-02-25
Chat template static files layout update

SteveSandersonMS:
Now that we're putting `Data` under `wwwroot`, it's a little more important that `wwwroot` itself is clean and minimal so that people aren't distracted by its contents when doing the very first thing: adding custom files to `Data`.

Updates in this PR:

 * Renamed `lib.js` to `app.js` to align with `app.css` (also, the old name `lib.js` made sense when it was the result of running a rollup build on libraries, but now it doesn't do that the name doesn't make sense anyway)
 * Moved `pdf_viewer` under `lib`
   * I know it's technically our code, but effectively it's indistinguishable from 'lib' content to the app developer - it's an artifact they won't see or edit in most cases
* Updated `App.razor` to use `@Assets[...]` consistently

I considered eliminating `app.js`/`lib.js` completely and inlining it into `ChatMessageList.razor.js`. We could do that and it would work fine. However it then turns that file into a miscellaneous bunch of stuff instead of having a single clear role, so I marginally prefer not doing that.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5968)


jmatthiesen (approved):
Good catch on the cleanup!

PR #5974, opened 2025-02-26
Addressing dogfooding feedback, adding GitHub Models support

jmatthiesen:
Updates to address a bug with Azure AI Search ingestion, also adding GitHub Models support that was temporarily disabled (was waiting for a fix in GH Models) .
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5974)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Pages/Chat/Chat.razor
```diff
@@ -9,7 +9,11 @@
 <ChatHeader OnNewChat="@ResetConversationAsync" />
 
 <ChatMessageList Messages="@messages" InProgressMessage="@currentResponseMessage">
-    <NoMessagesContent>Ask me anything about ChatWithCustomData.Web.</NoMessagesContent>
+    <NoMessagesContent>
+        <div>Ask me anything about the example PDFs.</div>
+        <ChatCitation File="Example - Emergency Survival Kit.pdf"/>
+        <ChatCitation File="Example - GPS Watch.pdf"/>
```
stephentoub, line 15: Shouldn't these be based on whatever files are actually in the folder? i.e. if the template consumer drops different PDFs, shouldn't this reflect that? Seems like this should not be hardcoded.

SteveSandersonMS: Not really - the intent here is just to give the app developer a sense of what to do when they first run it. They are not expected to leave this markup in the application once they start adding custom content. In general it wouldn't work to list all the source documents (there could be thousands, or you might not even be using a directory of PDFs and might instead reference arbitrary external content from your semantic search).

However I appreciate this will be confusing. We might need to make some change like:

```html
<div>To help you get started, we've added the following example documents. You can replace these with your own data and remove or replace this message.</div>
<ChatCitation File="Example - Emergency Survival Kit.pdf"/>
<ChatCitation File="Example - GPS Watch.pdf"/>
```

@stephentoub Do you think people would understand that or would it still be confusing?

stephentoub: > Do you think people would understand that or would it still be confusing?

That'd be better.

SteveSandersonMS (approved):

PR #5979, opened 2025-02-26
HybridCache JSON serialization improvements

### LINKED ISSUE #5716: HybridCache support for Tuples
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

I've been trying to implement a system based on eTag to return a 304 in case the existing resource has not changed in the Database.

I have a HybridCache instance working alongside this ETag in the API layer. When returning a tuple ( obj, string ) from the cache, they are returned as null, even if values are attributed in the service call.  Provided sample code:

```
var tags = new List<string> { "test" };
var entryOptions = new HybridCacheEntryOptions
{
    Expiration = TimeSpan.FromMinutes(1),
    LocalCacheExpiration = TimeSpan.FromMinutes(1)
};

var clientETag = httpContext.Request.Headers["If-None-Match"].FirstOrDefault();
var  (result, eTag) = await _cache.GetOrCreateAsync($"identifier", async get =>
{
    var (result, eTag) = await _companyService.GetPaginatedBillsAasync(companyId, filter, clientETag); // values retrieved from service 
    return  (result, eTag); // values are attributed.
}, entryOptions, tags); // values are still shown in breakpoints
// After reaching this point, result and eTag are now null.

if (result == null) // Here result is always null. 
{
    return Results.StatusCode(304); 
}

```

### Describe the solution you'd like

I expect the tuple to retain it's value after the Func has finished it's execution.

### Additional context

-

### LINKED ISSUE #5827: [API Proposal]: HybridCache - make it easier to use custom JsonSerializerOptions
### Background and motivation

The default HybridCache serializer uses `JsonSerializer`, but there is no way to configure the `JsonSerializerOptions`.

I think it is fairly common to have a custom `JsonSerializerOptions` which is helpful (or even necessary) for serializing cached types. Thus I think it would be helpful if it were easier to configure the `JsonSerializerOptions` used by the HybridCache.

As it stands now, you have to write your own serializer factory which is a lot of boilerplate.

### API Proposal

In general I see 2 main options:

1. The `DefaultJsonSerializerFactory` tries to obtain a `JsonSerializerOptions` from the DI container (no API change)
2. The `DefaultJsonSerializerFactory` is made public, taking a `JsonSerializerOptions` parameter, so we can use it with `.AddSerializerFactory(...)` instead of re-implementing it

### API Usage

For option 1, there's no change, you'd just need to add a singleton `JsonSerializerOptions` to DI

For option 2, it would look something like this:

```cs
var jsonSerializerOptions = BuildJsonSerializerOptions();
services.AddHybridCache(opts =>
{
    opts.DefaultEntryOptions = new HybridCacheEntryOptions { Expiration = TimeSpan.FromMinutes(5) };
}).AddSerializerFactory(new DefaultJsonSerializerFactory(jsonSerializerOptions));
```

### Alternative Designs

_No response_

### Risks

Option 1 could be considered a breaking change since the JSON serialization behaviour would change, however the hybrid cache is still experimental. Option 2 should not break anything.

mgravell:
- support value-tuples (requires field mode)
- support global custom JSON options via keyed DI

More context in the two linked issues below

fix #5716 - value-tuples aren't serializing correctly because they lack property accessors
fix #5827 - provide a basic mechanism to specify custom JSON options *without* requiring additional API surface

a better API for #5827 (i.e. an explicit API to register global or per-type JSON options) may be provided at a later date, but requires API review

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
@@ -11,24 +11,30 @@ namespace Microsoft.Extensions.Caching.Hybrid.Internal;
 
 internal sealed class DefaultJsonSerializerFactory : IHybridCacheSerializerFactory
 {
+    private readonly IServiceProvider _serviceProvider;
+
     internal static JsonSerializerOptions FieldEnabledJsonOptions { get; } = new() { IncludeFields = true };
 
     internal JsonSerializerOptions Options { get; }
 
-    public DefaultJsonSerializerFactory([FromKeyedServices(typeof(HybridCache))] JsonSerializerOptions? options = null)
+    public DefaultJsonSerializerFactory(IServiceProvider serviceProvider)
     {
+        // store the service provider and obtain the default JSON options, keyed by the **open** generic interface type
+        _serviceProvider = serviceProvider;
+
 #pragma warning disable IDE0079 // unnecessary suppression: TFM-dependent
 #pragma warning disable IL2026, IL3050 // AOT bits
-        Options = options ?? JsonSerializerOptions.Default;
+        Options = serviceProvider.GetKeyedService<JsonSerializerOptions>(typeof(IHybridCacheSerializer<>)) ?? JsonSerializerOptions.Default;
 #pragma warning restore IL2026, IL3050
 #pragma warning restore IDE0079
     }
 
     public bool TryCreateSerializer<T>([NotNullWhen(true)] out IHybridCacheSerializer<T>? serializer)
     {
-        // no restriction - accept any type
+        // no restriction - accept any type (i.e. always return true)
 
-        var options = Options;
+        // see if there is a per-type options registered (keyed by the **closed** generic type), otherwise use the default
+        JsonSerializerOptions options = _serviceProvider.GetKeyedService<JsonSerializerOptions>(typeof(IHybridCacheSerializer<T>)) ?? Options;
         if (IsValueTuple(typeof(T)) && !options.IncludeFields)
         {
             // value-tuples expose fields, not properties; special-case this as a common scenario
--- a/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
@@ -1,42 +1,73 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Buffers;
 using System.Diagnostics.CodeAnalysis;
 using System.Text.Json;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Extensions.Caching.Hybrid.Internal;
 
 internal sealed class DefaultJsonSerializerFactory : IHybridCacheSerializerFactory
 {
+    internal static JsonSerializerOptions FieldEnabledJsonOptions { get; } = new() { IncludeFields = true };
+
+    internal JsonSerializerOptions Options { get; }
+
+    public DefaultJsonSerializerFactory([FromKeyedServices(typeof(HybridCache))] JsonSerializerOptions? options = null)
+    {
+#pragma warning disable IDE0079 // unnecessary suppression: TFM-dependent
+#pragma warning disable IL2026, IL3050 // AOT bits
+        Options = options ?? JsonSerializerOptions.Default;
+#pragma warning restore IL2026, IL3050
+#pragma warning restore IDE0079
+    }
+
     public bool TryCreateSerializer<T>([NotNullWhen(true)] out IHybridCacheSerializer<T>? serializer)
     {
-        // no restriction
-        serializer = new DefaultJsonSerializer<T>();
+        // no restriction - accept any type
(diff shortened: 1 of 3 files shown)


mgravell:
The first edition of this PR only allows global registration, but it may often be desirable to allow per-type; I propose to update this to use the interfaces as the keys instead - the factory interface for global, and the typed interface for per-type, with inheritance.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -1,42 +1,73 @@
...
+#pragma warning restore IDE0079
+    }
+
     public bool TryCreateSerializer<T>([NotNullWhen(true)] out IHybridCacheSerializer<T>? serializer)
     {
-        // no restriction
-        serializer = new DefaultJsonSerializer<T>();
+        // no restriction - accept any type
+
+        var options = Options;
```
RussKie, line 31:
```suggestion
        JsonSerializerOptions options = Options;
```

RussKie (approved):

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
mgravell: resolved in https://github.com/dotnet/extensions/pull/5979/commits/b7dd36294b1f68a2e6a3f4e0cd26094cfa507d18, thanks

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -1,42 +1,79 @@
...
+    internal JsonSerializerOptions Options { get; }
+
+    public DefaultJsonSerializerFactory(IServiceProvider serviceProvider)
+    {
+        // store the service provider and obtain the default JSON options, keyed by the **open** generic interface type
+        _serviceProvider = serviceProvider;
+
+#pragma warning disable IDE0079 // unnecessary suppression: TFM-dependent
+#pragma warning disable IL2026, IL3050 // AOT bits
+        Options = serviceProvider.GetKeyedService<JsonSerializerOptions>(typeof(IHybridCacheSerializer<>)) ?? JsonSerializerOptions.Default;
```
DeagleGross, line 27: Will definitely work; I wonder if that is easy to discover, and maybe we can provide some definition like `HybridCacheOptions.JsonSerializerOptionsName` and then resolve via `serviceProvider.GetServiceByName(...)` in order to be more explicit.

_(more a question about why this decision is done than actual comment - the simpler the better I guess)_

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -1,42 +1,79 @@
...
+    }
+
     public bool TryCreateSerializer<T>([NotNullWhen(true)] out IHybridCacheSerializer<T>? serializer)
     {
-        // no restriction
-        serializer = new DefaultJsonSerializer<T>();
+        // no restriction - accept any type (i.e. always return true)
+
+        // see if there is a per-type options registered (keyed by the **closed** generic type), otherwise use the default
+        JsonSerializerOptions options = _serviceProvider.GetKeyedService<JsonSerializerOptions>(typeof(IHybridCacheSerializer<T>)) ?? Options;
```
DeagleGross, line 37: why is resolving every time required? Did not we do it in ctor already? Or the idea is that if it was not yet available in ctor, but when performing `TryCreateSerializer<T>` it will be?

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -1,42 +1,79 @@
...
+        internal JsonSerializerOptions Options { get; }
+
+        public DefaultJsonSerializer(JsonSerializerOptions options)
         {
-            var reader = new Utf8JsonReader(source);
+            Options = options;
+        }
+
 #pragma warning disable IDE0079 // unnecessary suppression: TFM-dependent
 #pragma warning disable IL2026, IL3050 // AOT bits
```
DeagleGross, line 61: question: AOT is not supported by design?

DeagleGross (approved):

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
mgravell: It isn't; this method is only called once per `T`, with the selected serializer then cached in `DefaultHybridCache` - but we don't know which various `T` we will need until we see them at runtime.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
mgravell: The problem then becomes the per-type aspect; that's why we're keying by the *open* generic type for the global option, and the *closed* generic type for the per-type option. I mentioned in the PR that we may add a discoverable API for this (that just pokes the same things), but that this needs API design/review.

PR #5988, opened 2025-02-27
Improve debugger display for ChatMessage

stephentoub:
Before:
before

After:
after

We can subsequently change exactly how it's implemented and what's displayed, maybe JSON serializing the AIContent instead of doing something custom for each type. But it's much easier to debug with this than with what was there before.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5988)


SteveSandersonMS (approved):
Very nice

PR #5994, opened 2025-02-28
AI Template Preview 1 - ID cleanup & link to survey

jmatthiesen:
A couple random items that came from discussions today, about final updates before we publish a preview 1 of the AI template.
* Changed the template ID to reflect this is a webChat template, and leave us open to future chat templates
* Added a survey prompt with an aka.ms link for a feedback survey (the link goes nowhere for now).

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5994)

### DIFF
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
@@ -2,7 +2,7 @@
   "$schema": "http://json.schemastore.org/template",
   "author": "Microsoft",
   "classifications": [ "Common", "AI", "Web" ],
-  "identity": "Microsoft.Extensions.AI.Templates.Chat.CSharp",
+  "identity": "Microsoft.Extensions.AI.Templates.WebChat.CSharp",
   "name": "AI Chat Web App",
   "description": "A project template for creating an AI chat application, which uses retrieval-augmented generation (RAG) to chat with your own data.",
   "shortName": "aichatweb",
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Layout/SurveyPrompt.razor
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Layout/SurveyPrompt.razor
@@ -1,13 +1,13 @@
 <div class="surveyContainer">
-    <span class="tool-icon" aria-hidden="true">
+    <div class="tool-icon" aria-hidden="true">
         <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="size-5">
             <path stroke-linecap="round" stroke-linejoin="round" d="M16.862 4.487l1.687-1.688a1.875 1.875 0 112.652 2.652L10.582 16.07a4.5 4.5 0 01-1.897 1.13L6 18l.8-2.685a4.5 4.5 0 011.13-1.897l8.932-8.931zm0 0L19.5 7.125" />
         </svg>
-    </span>
+    </div>
 
-    <span>
+    <div>
         How well is this template working for you? Please take our
         <a target="_blank" href="https://aka.ms/dotnet-chat-template-survey">brief survey</a>
         and tell us what you think.
-    </span>
+    </div>
 </div>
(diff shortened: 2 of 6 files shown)


jeffhandley (approved):

SteveSandersonMS (changes requested):
Would it be OK to make the survey link **much** more subtle? I get that you're making it look like the gold bar from VS, but that's very distracting and is just an eyesore until the developer removes it.

It would also be nice not to have this on the screen all the time.

Instead, what about adding it at the bottom of `NoMessagesContent` and simply having it be in subtle grey text, no yellow background? Then everyone who uses the template will still be certain to see it, but it doesn't interfere with really trying it out and won't interfere with our demo flows.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "http://json.schemastore.org/template",
   "author": "Microsoft",
   "classifications": [ "Common", "AI", "Web" ],
-  "identity": "Microsoft.Extensions.AI.Templates.Chat.CSharp",
+  "identity": "Microsoft.Extensions.AI.Templates.WebChat.CSharp",
```
stephentoub, line 5: Should "AI Chat Web App" be "AI Web Chat App"? I can see it going both ways.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Layout/SurveyPrompt.razor
```diff
@@ -0,0 +1,20 @@
...
+    <span>
+        <strong>@Title</strong>
+        Please take our
+        <a target="_blank" href="https://aka.ms/dotnet-chat-template-survey">brief survey</a>
+        and tell us what you think.
+    </span>
+</div>
+
+@code {
+    // Demonstrates how a parent component can supply parameters
```
stephentoub, line 17: Demonstrates to whom? Isn't this whole page temporary?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
RussKie: Is it "AI Chat" webapp or "AI web chat" app? As in, if there could be an "AI Chat" winapp then it should be the former. I wouldn't think an "AI win chat" app would be very technologically different from an "AI web chat" app....

jmatthiesen: I wanted to focus on scenario first ("AI Chat") in the name and then app type, e.g. ("Web App", or "Web API")

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Components/Layout/SurveyPrompt.razor
jmatthiesen: Hold over from the original Blazor template where I got this. I've removed it now.

jmatthiesen:
> Would it be OK to make the survey link **much** more subtle? I get that you're making it look like the gold bar from VS, but that's very distracting and is just an eyesore until the developer removes it.
> 
> It would also be nice not to have this on the screen all the time.
> 
> Instead, what about adding it at the bottom of `NoMessagesContent` and simply having it be in subtle grey text, no yellow background? Then everyone who uses the template will still be certain to see it, but it doesn't interfere with really trying it out and won't interfere with our demo flows.

Updated to be less "in your face". I didn't want to move it to NoMessagesContent because it would disappear as you start using the template. I'd like it to stick around so they see it after a few interactions, at least for now while we work to get feedback.

SteveSandersonMS:
> Updated to be less "in your face"

Thanks for the update!

For preview 1 I think it's OK to have the survey link on the screen all the time as you have, but for subsequent previews I hope we can avoid doing so as it's still quite invasive. It might not even put survey-fillers in a good frame of mind as they'd be filling it out through exhaustion rather than because they feel inclined to give feedback.

Likewise we'd benefit from reducing the total number of UI elements that appear by default, so perhaps for subsequent previews we could simplify it to:

image

This also avoids giving the impression that there should be a default link to every source document.

I'm also happy to do this right now but if you prefer to keep the link at the bottom for preview 1 I'm fine with that too.

SteveSandersonMS (approved):

SteveSandersonMS:
Actually the appearance with the current code can be weird at some screen widths, e.g.:

image

SteveSandersonMS:
@jmatthiesen I did a pass of testing this on small screens (mobile) and there were quite a few layout issues to fix. The result now is workable though it's still very busy-looking on a small screen.

If in a subsequent preview we remove both the survey prompt and the NoContent links to PDFs, it will make it look hugely nicer.

Issue #5909, opened 2025-02-15
[AI] FunctionCallContent is not included in StreamingChatCompletionUpdate when KeepFunctionCallingMessages enabled


Gigabyte0x1337:
### Description

When using `Microsoft.Extensions.AI` with function calling and streaming, the `FunctionCallContent` is not included, even with `KeepFunctionCallingMessages` enabled.

The reason I need this is to notify my front-end that a function call is busy. Of course, I could implement the whole thing myself, but what's the point of having it if it doesn't serve this purpose?

### Reproduction Steps

Minimal example code to repreduce this.

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
	  <OutputType>Exe</OutputType>
	  <TargetFramework>net9.0</TargetFramework>
	  <ImplicitUsings>enable</ImplicitUsings>
	  <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.AI" Version="9.1.0-preview.1.25064.3" />
    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.1.0-preview.1.25064.3" />
    <PackageReference Include="Microsoft.Extensions.AI.OpenAI" Version="9.1.0-preview.1.25064.3" />
    <PackageReference Include="Microsoft.Extensions.Options" Version="9.0.2" />
  </ItemGroup>

</Project>
```

```csharp
var chatClient = new ChatClientBuilder(new OpenAIChatClient(new OpenAIClient(openAiKey), "gpt-4o"))
    .UseFunctionInvocation(null, client =>
    {
        client.KeepFunctionCallingMessages = true;
        client.RetryOnError = true;
        client.ConcurrentInvocation = true;
    })
    .Build();


var messageStream = chatClient.CompleteStreamingAsync([
    new ChatMessage(ChatRole.System, [new TextContent("You are a helpful assistant")]),
    new ChatMessage(ChatRole.User,
    [
        new TextContent(
            "Hello, can you call the test function, try `test` as input. Follow the instruction solve the riddle! No need for confirmation do it yourself!")
    ])
], new ChatOptions()
{
    Tools = [AIFunctionFactory.Create((string? password) => password != "The Goat" ? 
        "Call me again with `The Goat` and see what happens next!" : 
        "Pass this along will you, 'The goat has been milked'!", "test_function", "This is function to test function calling.")]
});


[...]

Issue #5675, opened 2024-11-20
Streaming messages in response before function call request aren't being captured


stephentoub:
These should be sent back to the AI, but they're not. It's rare to see this with some models, like those from OpenAI, but it's more common with others.

stephentoub:
Fixed by Proposal for tweaks to IChatClient contract #5998

PR #5998, opened 2025-02-28
Proposal for tweaks to IChatClient contract

stephentoub:
- Reverts to an immutable model for chat history
- Updates ChatResponse to contain multiple response messages
- Removes Choices and ChoiceIndex

OLD:
-----
As we march towards marking Microsoft.Extensions.AI.Abstractions stable in the next few months, I've been re-evaluating a few fundamental aspects of the design. This is based on usage, feedback, the code developers writing to use the abstractions, as well as on some open issues that highlight some design limitations. While we're trying to lock things down and avoid breaking changes, this PR proposes some tweaks to hopefully address all of the above and set the abstractions up for even more success.

The core of the change is that, right now, it's the developer's responsibility to update the chat history with the response message. For example, a typical getting-started chat bot loop looks like:
```C#
List<ChatMessage> history = [];
while (true)
{
    Console.Write("Q: ");
    history.Add(new ChatMessage(ChatRole.User, Console.ReadLine());

    var response = await client.GetResponseAsync(history);
    Console.WriteLine(response);

    history.Add(response.Message);
}
```
The developer here has to transfer the response message back into the history list. (At least when dealing with a stateless service. When dealing with a stateful service, that's not necessary or desirable.) The same challenge exists for streaming, where it's even harder. We've provided the ToChatResponse helper to try to make this easier, but the developer still needs to track all updates in order to combine them back into the history:
```C#
List<ChatMessage> history = [];
List<ChatResponseUpdates> updates = [];
while (true)
{
    Console.Write("Q: ");
    history.Add(new ChatMessage(ChatRole.User, Console.ReadLine());

    updates.Clear();
    await foreach (var update in client.GetStreamingResponseAsync(history))
    {
        updates.Add(update);
        Console.Write(update);
    }
    Console.WriteLine();

    history.Add(updates.ToChatResponse());
}
```

We expect orders of magnitude more consumers of these abstractions than implementers of these abstractions, so it'd be nice if this extra effort could be pushed into the implementations. If it were the responsibility of the IChatClient to augment the history, the developer wouldn't need to care about it, e.g.
```C#
List<ChatMessage> history = [];
while (true)
{
    Console.Write("Q: ");
    history.Add(new ChatMessage(ChatRole.User, Console.ReadLine());

    Console.WriteLine(await client.GetResponseAsync(history));
}
```
and
```C#
List<ChatMessage> history = [];
while (true)
{
    Console.Write("Q: ");
    history.Add(new ChatMessage(ChatRole.User, Console.ReadLine());

    await foreach (var update in client.GetStreamingResponseAsync(history))
    {
        Console.Write(update);
    }
    Console.WriteLine();
}
```

The history is already declared to be mutable, but we're currently in a weird state where only _some_ of the messages are automatically added into the history. Consider FunctionInvokingChatClient. It adds to the history all responses and generated messages incurred as part of its back and forth with the underlying IChatClient, except for the last message, which it returns. But that's not the case for streaming, where it necessarily needs to stream back all response content (if it were to instead buffer the streaming content until it knew whether there were a function call, it would no longer be streaming).  However, in such a case, the streaming response might span multiple actual response messages, which means any resulting individual message that's created lacks fidelity with the original messages. In such a case as well, it means the final history likely includes out of order messages, because the FunctionInvokingChatClient is updating the history with the function calling content before the caller has the full streaming results and can update the history with it. And that then causes possible problems for subsequent use where a service gets a history containing out of order messages; some services.

Local function invocation is not the only case where there might be multiple response messages. Service-side use of tools like code interpretation and web search can also logically result in multiple response messages, as can agent-based scenarios where a single input message triggers a conversation to produce a single output.

To address all of this and more, I'm proposing the following:
- A IChatClient.Get{Streaming}ResponseAsync implementation has the responsibility to add any messages it generates into the chat history. For typical leaf clients, that means GetResponseAsync just adds to the history any message it's also going to return, and for GetStreamingResponseAsync, it maintains a `List<ChatResponseUpdate>` as I did in my earlier code snippet, and again adds to the history the result of converting those into a `ChatMessage` (via the `ToChatMessage` helper). For most middleware that's not producing its own messages, no changes are needed. For middleware like FunctionInvokingChatClient, it no longer needs to augment the history with the (streaming) response messages it gets back from its inner client, because it's that inner client's responsibility to do so; the only thing the FunctionInvokingChatClient needs to add is add the function result content to the history, which it's already doing.
- ChatResponse.Choices is removed. Of the many IChatClient implementations that now exist, literally zero support multiple choices in a response. The industry appears to have moved on from the concept, and it's not something we need to represent in the abstraction; having it just adds complication (and makes the rest of this proposal challenging, because the implementations would need to choose just one choice as the result). The main argument for services supporting choices over submitting the same request multiple times was as a cost-cutting measure for re-processing the same input multiple times, but that's largely been addressed via services supporting caching, making it all the more unlikely this will re-emerge as a need. But if it ever does, it can be added back in a niche manner, e.g. just as today the only way to ask a service to provide multiple responses is via AdditionalProperties, AdditionalProperties on the response could also be used to bring back alternatives, at least for the non-streaming case.
- Add the ability for ChatResponse to carry multiple messages, not just one. GetResponseAsync implementations then return all messages generated as part of the request rather than just the last. This is useful in cases where the implementation isn't maintaining a history at all (e.g. its stateful in the service), where the caller isn't maintaining a history (e.g. where they're just passing in a string prompt that results in a temporary history list being created), or where middleware mucks with the history in a way that makes it harder for the consumer to know what's new... but even without that, providing back all of the messages just eases the burden on the consumer writing code. ChatResponse.Message still behaves the same, returning the last message from the stored list of messages.
- Get rid of FunctionInvokingChatClient.KeepFunctionCallingMessages. It defaulted to true (another indication of this being the right direction), and it's now just how the whole system works, anyway.

As with other breaking changes we've made in preview, this will entail reaction in both implementations of the abstraction and consumption of it. On the consumption side, the largest impact to not reacting is that there might be some duplicate messages in the chat history, which is generally not problematic. But we'll still want to help consumers move to the new, simpler consumption patterns.

This also implicitly addresses https://github.com/dotnet/extensions/issues/5909 and https://github.com/dotnet/extensions/issues/5675.

cc: @SteveSandersonMS, @eiriktsarpalis, @luisquintanilla, @matthewbolanos, @RogerBarreto, @davidfowl, @DamianEdwards, @peterwald, @shyamnamboodiripad, @MackinnonBuck, @halter73 @colombod, @tghamm
 
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/5998)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -30,11 +30,6 @@ public static ChatCompletion ToOpenAIChatCompletion(ChatResponse response, JsonS
     {
         _ = Throw.IfNull(response);
 
-        if (response.Choices.Count > 1)
-        {
-            throw new NotSupportedException("Creating OpenAI ChatCompletion models with multiple choices is currently not supported.");
-        }
-
```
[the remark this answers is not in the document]
stephentoub: Thanks, I'll put this back. I was overzealous in my deletion of Choices and missed that this one is on the OpenAI object.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -139,40 +144,6 @@ public static FunctionInvocationContext? CurrentContext
...
-    /// Changing the value of this property while the client is in use might result in inconsistencies
-    /// as to whether function calling messages are kept during an in-flight request.
-    /// </para>
-    /// <para>
-    /// If the underlying <see cref="IChatClient"/> responds with <see cref="ChatResponse.ChatThreadId"/>
-    /// set to a non-<see langword="null"/> value, this property may be ignored and behave as if it is
-    /// <see langword="false"/>, with any such intermediate messages not stored in the messages list.
-    /// </para>
-    /// </remarks>
-    public bool KeepFunctionCallingContent { get; set; } = true;
```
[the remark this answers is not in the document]
stephentoub: > I noticed a scenario where function calling responses back to AI persisted in the chat history may cause the AI Model to assume further calls to the same function aren't needed as it uses the previous role=tool answer as a "Cached" response.

That sounds like it's just how the model reasoned, no? I don't want us chasing the current behavior / reasonining capabilities of today's models.  Can you elaborate on the case?

KeepFunctionCallingContent (which defaulted to true) was also just a nicety, so that a dev that didn't want that content could easily remove it themselves. If we _really_ want to add such functionality back, I'd prefer we expose it in a more cross-cutting way, e.g. `static void Remove(this IList<ChatMessage> message, Predicate<AIContent> filter)`, or some such thing.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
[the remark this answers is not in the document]
stephentoub: Wait, no, this is our ChatResponse here (github diff wasn't showing it to me).

What are you suggesting still be checked here?

src/Shared/Throw/Throw.cs
```diff
@@ -294,6 +294,24 @@ public static IEnumerable<T> IfNullOrEmpty<T>([NotNull] IEnumerable<T>? argument
...
+    /// <returns>The original value of <paramref name="argument"/>.</returns>
+    public static ICollection<T> IfReadOnly<T>(ICollection<T> argument, [CallerArgumentExpression(nameof(argument))] string paramName = "")
+    {
+        if (argument.IsReadOnly)
+        {
+            ArgumentException(paramName, "Collection is read-only");
+        }
+
+        return argument;
+    }
```
RussKie, line 313: Please add test coverage for this.

RussKie (approved):
Approving for the "fundamentals" side.

shyamnamboodiripad (approved):
The proposed changes sound reasonable. I don't see significant impact from the standpoint of MEAI.Evaluation, since the conversation history that is passed to the evaluation APIs (for evaluation) is only read (i.e., it is embedded as part of the prompts that perform evaluation) and is not mutable.

The evaluation itself is also performed using single shot prompts at the moment (at least for the built-in evaluators that ship as part of the MEAI.Evaluation.Quality assembly today). So, mutating the conversation history for evaluation also shouldn't have any significant impact - and in more advanced future cases where multi-turn conversation may be needed for evaluation, the proposed changes should prove just as beneficial as they would for the general case.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -3,58 +3,100 @@
...
+
+    /// <summary>Gets or sets the last chat response message.</summary>
+    /// <remarks>
+    /// When getting <see cref="Message"/>, if there are no messages, <see cref="Message"/> will add a new
+    /// empty message to the list and return that message; if there are messages, the last will be returned.
+    /// When setting <see cref="Message"/>, if there are messages, the last message will be replaced by the
+    /// newly set instance; if there are no messages, the newly set instance will be added to the list.
     /// </remarks>
     [JsonIgnore]
     public ChatMessage Message
```
shyamnamboodiripad, line 79: I find it a bit confusing that this returns / replaces the 'last' message. Should the property be renamed to clarify the semantics?

shyamnamboodiripad:
> The proposed changes sound reasonable. I don't see significant impact from the standpoint of MEAI.Evaluation, since the conversation history that is passed to the evaluation APIs (for evaluation) is only read (i.e., it is embedded as part of the prompts that perform evaluation) and is not mutable.

Actually, thinking a bit more, this may have an impact on [`IEvaluator.EvaluateAsync`](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation/IEvaluator.cs#L51) which has the following signature -
```
ValueTask<EvaluationResult> EvaluateAsync(
        IEnumerable<ChatMessage> messages,
        ChatMessage modelResponse,
        ChatConfiguration? chatConfiguration = null,
        IEnumerable<EvaluationContext>? additionalContext = null,
        CancellationToken cancellationToken = default);
```

The assumption so far was that this would be convenient for callers since they would normally have the 'history minus the response' (i.e., `messages`) and 'the response' (i.e., `modelResponse`) available separately. However, if the common case going forward will be one where the caller always has 'history including the response', then calling `IEvaluator.EvaluateAsync` above would become cumbersome.

Should we get rid of the `modelResponse` parameter and update the xml comments to clarify that the last message in `messages` will be treated as the `modelResponse` to be evaluated?

We also have a [couple of convenience extension methods](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluatorExtensions.cs) for cases where the caller wants to supply a question answer pair, or just the model response alone (without any conversation history. Those seem fine and should not need any changes.

peterwald (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMappers.StreamingChatCompletion.cs
```diff
@@ -199,7 +203,10 @@ public static async IAsyncEnumerable<ChatResponseUpdate> FromOpenAIStreamingChat
                 (responseUpdate.AdditionalProperties ??= [])[nameof(ChatCompletion.SystemFingerprint)] = fingerprint;
             }
 
+            responseUpdates.Add(responseUpdate);
             yield return responseUpdate;
         }
+
```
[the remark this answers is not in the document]
SteveSandersonMS: > The former would be my preference; having an incomplete message in the history doesn't seem desirable.

Arguably it is desirable because most likely you'd want the object model to reflect whatever you're showing in the UI. In the case where a response was truncated, would the UI be likely to remove the text that was already presented, or to retain it with some kind of (stopped) indicator at the end?

Then if the user poses a follow-up question, would you expect the LLM to understand it as being in the context of the previously-truncated response, or think of it as being independent of the previous response?

For example if the conversation goes:

 * User: Tell me about lemonade
 * Assistant: Do you mean the drink? I'm asking because it's also the name of [stopped]
 * User: Yes

... the continuation wouldn't make sense without the truncated assistant message.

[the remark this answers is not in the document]
SteveSandersonMS: I just checked and, if my approach is valid, it appears the OpenAI Assistants at least will retain partially-streamed responses in the conversation history even if cancelled part-way through.

To check this, I used AI Foundry and asked an assistant to generate and run a bunch of code samples, and while it was going through this slow process, I cancelled the response. Then I asked some specific questions about information present in its previous incomplete response and it was able to answer them correctly.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
SteveSandersonMS: Agreed this is a tricky spot in the API design. I was also going to ask about it.

I know that:

 * It's really helpful to be able to get `response.Message.Text` in basic scenarios
 * Our previous design also assumed there was one special primary message in the response

... and so I definitely see the merits of retaining that, and in a sense it might appear not to be a change. On the other hand:

 * Arguably this is a change of semantics. Previously, the inner client was *required* to produce exactly one primary output message (in the nonstreaming case). If for some reason it had multiple response messages, it would have no choice but to coalesce them in some provider-specific way into a single primary response message. But with the change here, the inner client is free not to coalesce multiple response messages. In turn this forces app code to allow for the possibility of multiple response messages, and failing that, it could have a bug where some of the response messages are just ignored. Worse still, the final message might not the a "primary" one in any sense; it might for example be some kind of metadata, and then app code won't work at all if it uses `.Message`.
 * The setter here is extremely odd. I'd recommend that either:
   * We throw if there's more than one message
   * Or, better still, don't have a setter

Beyond that, other possible alternatives to `Message`:

 * We could eliminate `Message` and have a property called `Text`, that coalesces the text across all the `TextContent` messages. That's at least more general. We'd have to introduce some arbitrary behavior like "inserts two linebreaks between any nonempty `Text` message value". Develops who want to see anything beyond text would then need to process the whole `Messages` collection, which is really what we want them to do.
 * Or we could just require everyone to look at the `Messages` collection in all cases. They can always use `LastOrDefault()` if they really know they only care about the last message.

SteveSandersonMS:
Thanks for pushing this forwards, @stephentoub!

I think this is a big improvement to the clarity of the model. It feels like this is explainable and neatly creates equivalences across streaming/nonstreaming and stateless/assistant modes.

The two smaller design points I'd like to resolve are:

 * [ChatResponse.Message](https://github.com/dotnet/extensions/pull/5998#discussion_r1977892712)
 * [Cancellation of streaming](https://github.com/dotnet/extensions/pull/5998#discussion_r1977829093)

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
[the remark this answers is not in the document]
RogerBarreto: @stephentoub Yes, I suggest it to still be checked and `NotSupported`.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMappers.StreamingChatCompletion.cs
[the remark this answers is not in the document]
RogerBarreto: @SteveSandersonMS, thanks for the added context, this behavior make sense, as a cancelled AI message might still be used in follow up iterations, sometimes the user just realized the way started the wrong path and want to correct it (referring to not follow the previous path).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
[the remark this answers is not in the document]
stephentoub: > @stephentoub Yes, I suggest it to still be checked and `NotSupported`.

_What_ should be checked? There's no Choices property anymore.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
stephentoub: @SteveSandersonMS, @RogerBarreto, what behavior would you expect if there are no updates, either because an exception occurred before yielding any or because it completed successfully yielding nothing? In both cases, would you expect an empty message in the history?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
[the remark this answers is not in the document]
RogerBarreto: No much to elaborate, you have a good point, the backend API should dictate or allow the setting of what is the AI Model behavior for caching or not previous tool responses not the abstraction/setting.

- https://github.com/microsoft/semantic-kernel/issues/10648

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
stephentoub: Our current ChatMessage.Text property picks the first TextContent it finds and gets / modifies that. Here picking the last message in the response list and targeting that doesn't seem any more or less confusing, and I wouldn't want ChatResponse.Text to concatenate everything while ChatMessage.Text doesn't.

@SteveSandersonMS, `ChatResponse<T>` currently relies on Message. What behavior would you expect it to employ?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMappers.StreamingChatCompletion.cs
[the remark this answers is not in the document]
SteveSandersonMS: It seems fine to not add a message to history if it would be empty. Unless you're anticipating this leads to problems?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
SteveSandersonMS: > Our current ChatMessage.Text property picks the first TextContent it finds and gets / modifies that. Here picking the last message in the response list and targeting that doesn't seem any more or less confusing, and I wouldn't want ChatResponse.Text to concatenate everything while ChatMessage.Text doesn't.

These rules are quite nonobvious. A person might think that if ChatResponse.Message gets the last candidate, then ChatMessage.Text should also get the last candidate rather than the first.

A more consistent pattern would be:

* ChatResponse.Message concatenates all the text values, and doesn't have a setter
* ChatMessage.Text concatenates all the text values, and doesn't have a setter

Is it a big drawback not to have setters? Generally we'd expect the values only to be set by inner clients or middleware, both of which are expected to understand the MEAI object model in more detail than application code does.

> @SteveSandersonMS, ChatResponse<T> currently relies on Message. What behavior would you expect it to employ?

It could throw if there's more than one message, or it could consume an all-concatenating `Text` property, since it explicitly tells the LLM to respond in the form of a JSON object, which doesn't permit other commentary alongside it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -3,58 +3,100 @@
...
     {
         get
         {
-            var choices = Choices;
-            if (choices.Count == 0)
+            if (Messages.Count == 0)
             {
-                throw new InvalidOperationException($"The {nameof(ChatResponse)} instance does not contain any {nameof(ChatMessage)} choices.");
+                Messages.Add(new ChatMessage(ChatRole.Assistant, []));
```
RogerBarreto, line 85: If there aren't messages why are we adding a non-existing message assuming it is a `Assistant` with no content. Maybe consider `Message` to be nullable and return null if `Count == 0`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
RogerBarreto: Thinking about renaming the `Message` as 
1. a nullable `LastMessage` or 
2. a method `bool TryGetLastMessage(out Message)` or
3. a method `Message? TryGetLastMessage()`

With the above you provide a clear picture what is Property/Method is about for most of the scenarios.

The setter here is indeed odd, given the reasons above as well I would rather not have a setter initially so this would also encourage users to rely on the proper `Messages` list for adding/removing messages.

RogerBarreto (approved):
Overall I appreciate the changes, removing Choices clears the design and simplify most of the scenarios, I still think would be interesting adding breaking glass examples how the 1% would handle multiple choices.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMappers.StreamingChatCompletion.cs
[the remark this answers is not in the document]
RogerBarreto: @stephentoub If nothing was yielded back I also assume there was "no message" to add.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
stephentoub: > These rules are quite nonobvious.

I don't disagree, but I believe this all stems from your refactoring here... do you remember why you did it this way?

> Is it a big drawback not to have setters?

For ChatResponse, probably not. For ChatMessage, possibly, when the caller is creating it for user messages? Presumably the majority case there would include passing it the constructor, and there's nothing the caller couldn't do themselves.

> If middleware wants to replace all the text on a response, it can do something like:

And also update the history, presumably? (That's separate from this Text conversation, though.)

> Thinking about renaming the Message as

I'd rather folks just use Messages.LastOrDefault() rather than add such methods.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMappers.StreamingChatCompletion.cs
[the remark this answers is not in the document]
stephentoub: > Unless you're anticipating this leads to problems?

I'm just trying to avoid inconsistencies. If GetResponseAsync fails, presumably we'd expect nothing would be added to history, though it's hard to guarantee that, especially with a middleware pipeline that could introduce failures after the inner client already successfully added something. In the case of exceptions, it's hard to guarantee any kind of consistency.

Certainly the easiest thing would be just not adding a message to the history if nothing was yielded.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
SteveSandersonMS: > I don't disagree, but I believe this all stems from your initial refactoring here... do you remember why you did it this way with ChatMessage.Text?

I think `ChatMessage.Text` derives from SK's `ChatMessageContent.Content`: https://github.com/microsoft/semantic-kernel/blob/04f5fbbc5a62483486c98dc50c3a3d2943b89fed/dotnet/src/SemanticKernel.Abstractions/Contents/ChatMessageContent.cs#L43 which has the equivalent getter logic.

luisquintanilla: Not to deviate too much from the conversation, but as we introduce other modalities, not sure `Text` to get the last message makes sense here. Keeping it more generic would be my preference whether that's using the built-in collection methods or adding methods to the abstractions to return the last message / history content.

SteveSandersonMS: > And also update the history, presumably? (That's separate from this Text conversation, though.)

Yes I guess so.

shyamnamboodiripad: > I'd rather folks just use Messages.LastOrDefault() rather than add such methods.
I tend to agree, unless we can find a good name that conveys the intended purpose of `Message` or what makes it special - `LastMessage` does not convey this suitably unfortunately.

Couple of other thoughts though (and some of this may be based on my incomplete understanding of the APIs / overall semantics :))

How would the consumer know that the last message is the important one that they should be looking at? Or should clients assume that there will always be just a single message within `Messages` (at least in the non-streaming case)? If more than one is found, what should the recommended client behavior be - throw an exception?

In cases where there are (supposed to be) multiple messages, what happens if there is a bug in the implementation of the `IChatClient` where `Messages` is not correctly updated to include the important final response? Is there anything we can do in the API to guard against such cases (e.g., detect and throw an exception as opposed to silently returning the wrong message)?

I am guessing there is no easy way to know whether such bugs exist today because there is no easy way to distinguish one kind of message from another. Just thinking aloud, but could we make the primary / last message special in some way (by using a special `Message.Kind` enum property OR a special metadata tag or annotation OR a special derived type perhaps)? That would make it easier to support a first-class property to fetch this special message (and throw an exception if it is not found). However, I am not sure how that compares to other similar AI APIs out there and whether it would be too surprising / heavy weight if we were to introduce a special concept around this just in MEAI.

stephentoub:
@SteveSandersonMS, @RogerBarreto, please take another look.

@shyamnamboodiripad, please confirm the eval changes are appropriate.

Thanks.

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -3,58 +3,90 @@
...
+                return string.Empty;
             }
 
-            return choices[0];
+            int count = messages.Count;
+            return count switch
+            {
+                0 => string.Empty,
+                1 => messages[0].Text,
+                _ => messages.SelectMany(m => m.Contents).ConcatText(),
```
SteveSandersonMS, line 88: Do you think there would be a case for having at least a linebreak between the text from successive messages?

It looks like, as-is, separate messages would be combined without any spacing at all, so a message sequence like:

 - I'll see if I can find information for you.
 - Sorry, there was no information.

... would appear as: `I'll see if I can find information for you.Sorry, there was no information.` with incorrectly missing spacing between the two sentences.

I know this pulls us into taking an opinion about text formatting which we might prefer to avoid, but without having *some* kind of spacing it may just seem buggy. A linebreak might be the most culture-agnostic way to separate two blocks of text.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -26,27 +26,35 @@ namespace Microsoft.Extensions.AI;
...
-    /// <param name="options">The chat options to configure the request.</param>
+    /// <param name="chatMessages">The list of chat messages to send and to be augmented with generated messages.</param>
+    /// <param name="options">The chat options with which to configure the request.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The response messages generated by the client.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="chatMessages"/> is <see langword="null"/>.</exception>
     /// <remarks>
-    /// The returned messages aren't added to <paramref name="chatMessages"/>. However, any intermediate messages generated implicitly
-    /// by the client, including any messages for roundtrips to the model as part of the implementation of this request, are included.
+    /// The response updates generated by <see cref="GetStreamingResponseAsync"/> are both stream from the method as well as automatically
```
SteveSandersonMS, line 53:
```suggestion
    /// The response updates generated by <see cref="GetStreamingResponseAsync"/> are streamed from the method as well as automatically
```

Or "returned" instead of "streamed" would do.

I recommend we avoid the word "both" in this context since in one interpretation it would suggest there are going to be exactly two response updates.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -26,27 +26,35 @@ namespace Microsoft.Extensions.AI;
...
+    /// added into <paramref name="chatMessages"/>. Any intermediate messages generated implicitly as part of the interaction are
+    /// also added to the chat history. For example, if as part of satisfying this request, the <see cref="GetStreamingResponseAsync"/> method
```
SteveSandersonMS, line 55: Not sure if I'm being too fussy but the mention of "intermediate messages" seems ambiguous. Are they also returned, or are they only added to chat history? It may be unclear what makes a message "intermediate" vs a regular message, and if they are all returned as well as added, we could just not differentiate them in this description.

src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
```diff
@@ -143,10 +62,7 @@ while (true)
     Console.Write("Q: ");
     history.Add(new(ChatRole.User, Console.ReadLine()));
 
-    ChatResponse response = await client.GetResponseAsync(history);
-
-    Console.WriteLine(response);
-    history.Add(response.Message);
+    Console.WriteLine(await client.GetResponseAsync(history));
```
SteveSandersonMS, line 65: Much nicer!

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
```diff
@@ -83,6 +87,11 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
                 {
                     yield return chunk;
                 }
+
+                if (chatResponse.ChatThreadId is null)
```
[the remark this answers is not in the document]
SteveSandersonMS: Sorry for resurrecting this, but I'm also unsure why we'd skip adding the messages in this case.

I would expect that `CachingChatClient` wants to produce the same outcomes in both the cache-hit and cache-miss cases. In the cache-miss case, whether or not the messages are added to history is determined by the inner client, and in general I think we do expect inner clients to append to history regardless of `ChatThreadId` right?

That said I'm not actually sure how caching makes sense in general for assistant-type requests. We have no way to know what content already is associated with `ChatThreadId`, so how can we match any cache entry? If it was some sort of "continuation ID" representing an immutable state then we could include that in the cache key, albeit for not much gain since it would be rare to fork an existing continuation. But for the assistant case, the chat thread is itself mutable and could have been written to by another client, so there's no way to produce a valid cache hit ever it seems.

@stephentoub What are your thoughts? Am I missing something major?

SteveSandersonMS (approved):
Great! This is a pretty big and important change - thanks for making it happen.

Two remaining points from me:

* [Spacing](https://github.com/dotnet/extensions/pull/5998/files#r1978932811) - happy to discuss, would be ideal to resolve within this PR
* [Caching](https://github.com/dotnet/extensions/pull/5998/files#r1979012801) - might be a separate matter for a different PR, but interested to know what you think

peterwald (approved):
LGTM with respect to M.E.AI.Eval

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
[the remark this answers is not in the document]
SteveSandersonMS: > We don't currently when there's a response I'd set, and this cache impl is just trying to match that behavior. If we want to change that and day the inner client should always add to the history, regardless of whether the state is being tracked in the service, then we'd not special case this in the cache layer either.

Thanks for clarifying. I suppose the rationale is that, in the case of server-side state, there's no benefit in updating the history since the developer should never re-supply it all in later calls anyway, is it? On the other hand there's also no drawback in doing so (for the same reason).

Are there any other reasons either way?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -3,58 +3,90 @@
...
+    /// <summary>Gets the text of the response.</summary>
     /// <remarks>
-    /// If there are multiple choices, this property returns the first choice.
-    /// If <see cref="Choices"/> is empty, this property will throw. Use <see cref="Choices"/> to access all choices directly.
+    /// This property concatenates the <see cref="ChatMessage.Text"/> of all <see cref="ChatMessage"/>
+    /// instances in <see cref="Messages"/>.
     /// </remarks>
     [JsonIgnore]
-    public ChatMessage Message
+    public string Text
```
RogerBarreto, line 73: Do the `caller` needs to know what is the Role of a response? 
Will it always be an `Assistant` for the last?

I wonder if we shouldn't just flatten all message properties, not just the `Text` in the `ChatResponse`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
RogerBarreto: @stephentoub This was marked as resolved, what's the rational to not returning null if there's no messages?

Wasn't clear why `null` is not an option where `throw` was acceptable before.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
SteveSandersonMS: > Do the caller needs to know what is the Role of a response?

It's rare since the response role is almost always "assistant". But if the developer really wants to do see the role, they will need to look at the `Messages` where they see everything with full fidelity. And we have no reason to guarantee that all sub-messages will have the same role.

The role of the `Text` property is just convenience for cases where app developers don't think or care about there being multiple sub-messages and just want the text. It keeps a lot of cases simple and streamlines all the samples, and in a large percentage of cases just skips to the same result the dev would get if they manually walked through all the sub-messages manually.

In general I don't think we need to hide the fact that there may be multiple sub-messages that could have different values for any of the metadata. `Text` is special-cased just because it's so overwhelmingly common in use, and because there's a sensible way to collate multiple values (i.e., concatenate them, which is nearly always equivalent to what the app dev would have to do manually otherwise).

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
stephentoub: > This was marked as resolved

Because this property no longer exists.

stephentoub:
One more peek @SteveSandersonMS before this merges?

shyamnamboodiripad:
@stephentoub Just wanted to make sure you saw [this comment](https://github.com/dotnet/extensions/pull/5998#issuecomment-2693310104)  that I had posted above. This could be something that we change after this PR is merged - however, if we are planning to include this PR in the snap for the next release of packages, I think this may be something that we should resolve before release.

cc @peterwald

stephentoub:
Sorry, @shyamnamboodiripad, I missed that.

> Should we get rid of the modelResponse parameter and update the xml comments to clarify that the last message in messages will be treated as the modelResponse to be evaluated?

Either that, or we could do what `ChatResponse<T>` does and use response.Text, which is the concatenation of all TextContent in messages in the response. But that probably doesn't make as much sense for arbitrary evaluators, which could reasonably want to evaluate content besides text (images, audio, etc.)

Regarding the last message, is that always the right answer for any evaluator? Would there not be some evaluator that would want to understand the full exchange, e.g. evaluate what functions were called on the way to producing a final answer?

Maybe it should just take the ChatResponse, assuming that's what's being evaluated, and its up to the evaluator what to evaluate?

peterwald:
> Either that, or we could do what `ChatResponse<T>` does and use response.Text, which is the concatenation of all TextContent in messages in the response. But that probably doesn't make as much sense for arbitrary evaluators, which could reasonably want to evaluate content besides text (images, audio, etc.)

We definitely want to support evaluators that can adapt to different content types.

> Regarding the last message, is that always the right answer for any evaluator? Would there not be some evaluator that would want to understand the full exchange, e.g. evaluate what functions were called on the way to producing a final answer?

Yes, in fact we already have customers using measures of function-calling as part of the evaluation. So, we do want allow the evaluator to understand the full exchange.

> Maybe it should just take the ChatResponse, assuming that's what's being evaluated, and its up to the evaluator what to evaluate?

`ChatResponse` does seem to be the most reasonable/flexible choice. It might make it slightly more difficult for users who just have a blob of text that they want to evaluate, but we already have helpers for this and could easily add more. i.e. `text.ToChatResponse()`.

cc @shyamnamboodiripad

shyamnamboodiripad:
>>Maybe it should just take the ChatResponse, assuming that's what's being evaluated, and its up to the evaluator what to evaluate?

>Yes, replacing the modelResponse parameter to be the ChatResponse sounds like a good call. If we do this for the interface method above, we may also want to do the same in the convenience overloads I linked above.

@stephentoub One other wrinkle is that `ChatResponse` is only available in the non-streaming case. But users can call `.ToChatResponse` on the streaming updates to get a `ChatResponse` with all the constituent messages.

I and Peter were just chatting about this, and we couldn't see an easy way to avoid evaluator-implementers having to separate out the history from the response messages after the history `IList` has already been update - they would have to look for all messages at the end that are not `Role.System` and `Role.Assistant` in the history and recognize that they represent the response.

We could offer a public helper to simplify this separation (i.e. something that removes any `ChatMessage` that also exists in the `ChatResponse` from the history that is used for evaluation - the model response portion for evaluation would be directly available in the `ChatResponse`). `IEvaluator.EvaluateAsync` would then look like below and we would also make corresponding updates to the convenience extension methods.

```
ValueTask<EvaluationResult> EvaluateAsync(
        IEnumerable<ChatMessage> messages,
        **ChatResponse** modelResponse,
        ChatConfiguration? chatConfiguration = null,
        IEnumerable<EvaluationContext>? additionalContext = null,
        CancellationToken cancellationToken = default);
```

Thoughts?

stephentoub:
> One more peek @SteveSandersonMS before this merges?

I'm second guessing myself. Now that we're returning all additional messages as part of ChatResponse, we no longer _have_ to mutate the history... we could say that there's no mutation of the history, and even change the type from IList to IReadOnlyList if we wanted.

Having the history be mutated to include all the history makes the typical chat loop a bit easier and more similar between stateless and stateful modes.

But avoiding mutating the history is less surprising, and makes it easier to reason about what was there before the call and what's new since. In cases where you're not needing the history updated, it's also cheaper, e.g. multiple components are having to buffer up all streaming updates in order to turn them into messages to be added into the history, and that would only need to be done in cases where the history actually needs updating.

shyamnamboodiripad:
> Now that we're returning all additional messages as part of ChatResponse, we no longer have to mutate the history... we could say that there's no mutation of the history, and even change the type from IList to IReadOnlyList if we wanted.

> Having the history be mutated to include all the history makes the typical chat loop a bit easier and more similar between stateless and stateful modes.

> But avoiding mutating the history is less surprising, and makes it easier to reason about what was there before the call and what's new since. 

This makes sense and also seems to match most other APIs for AI chat that I have come across. (For example, SemanticKernl with its dedicated `ChatHistory` and OpenAI's [`ChatClient`](https://github.com/openai/openai-dotnet/blob/main/src/Custom/Chat/ChatClient.cs#L102) both make it the caller's responsibility to mutate history). Not mutating history would keep the high-level semantics between these APIs similar and would lead to fewer surprises in scenarios where callers need to switch between libraries / use more than one library in their projects (for example, use the OpenAI API directly for one part of the implementation and then use some library based on MEAI (like MEAI.Evaluation) for another part).

> Part of my second guessing is the evaluations support, which really wants a clear divide between the request and the response. Having the history contain everything means a caller needs to copy the history before making the call, which is possible, but unexpected I think.

Based on what we discussed, I think we still want to make the following change to the evaluation API (i.e. change the type of the `modelResponse` parameter from `ChatMessage` to `ChatResponse`) so that implementers have access to all the messages in the response that may need to be evaluated. cc @peterwald 

```
ValueTask<EvaluationResult> EvaluateAsync(
        IEnumerable<ChatMessage> messages,
        **ChatResponse** modelResponse,
        ChatConfiguration? chatConfiguration = null,
        IEnumerable<EvaluationContext>? additionalContext = null,
        CancellationToken cancellationToken = default);
```

@stephentoub Were you still planning to include the current PR in the snap for this month's release?

SteveSandersonMS:
> I'm second guessing myself. Now that we're returning all additional messages as part of ChatResponse, we no longer have to mutate the history...

OK, interesting.

For me, the main way either approach generates extra clarity/simplicity is by dropping the distinction between "primary" outputs (that are not auto-appended) and other secondary outputs (that are). We only had that distinction in the past to ease the usage patterns.

I hadn't anticipated that the combination of putting multiple messages on `ChatResponse` and simplifying their aggregation through `Text` would give us almost equivalent ease of use to what we had before even without us having to mutate history. I think the usage would be close to equivalent to what we had before:

* If you are just building text-based interactions or using structured output, it's exactly the same or even microscopically easier (`response.Message.Text` becomes `response.Text`, and beyond that no change)
* If you are consuming non-text content (e.g., image or citation responses) it becomes very slightly harder: you now need to loop over `response.Messages.SelectMany(m => m.Contents)` instead of just `response.Message.Contents`. We could simplify that by having a `Contents` property on `ChatResponse` that returns `Messages.SelectMany(m => m.Contents)`.
* In either case, instead of `history.Add(response.Message)` you'd now need `history.AddRange(response.Messages)` which is less obvious. Again we could add a helper if we felt it beneficial, e.g., because `IList` doesn't have `AddRange`.

Altogether it feels safer for us not to mutate history by default, as it makes MEAI less responsible for any bad effects of doing that. It also reduces the chances of `IChatClient` implementations not being consistent with each other, since it's easier for them consistently not to do something than to consistently do something.

SteveSandersonMS:
Also, under the auto-mutate model, we did still have a slightly rough inconsistency between stateless and stateful chat clients, in that we were auto-updating history, but we were not auto-updating `ChatThreadId` which is the analogous action. It's a bit strange to do one but not the other.
 
Having a model where `GenerateAsync` typically has no side effects on the inputs in both stateless/stateful cases is clearer. I know we can't and don't want to actually block mutations altogother, but we'd be making it more of a niche thing where the reasoning about it has move to whoever's doing that mutation (the middleware), instead of being a complexity tax on all client and app authors.

stephentoub:
Long winding journey here, but we're now back to what I think is a sane place to be: the incoming messages list provided to Get{Streaming}ResponseAsync is considered immutable, and all messages generated by the operation are returned/streamed. @SteveSandersonMS, @RogerBarreto, look good?

@shyamnamboodiripad / @peterwald, please take a look at the changes I made to Eval and let me know if they look good or I'm out to lunch.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/ChatConversationEvaluator.cs
```diff
@@ -250,21 +250,26 @@ await PerformEvaluationAsync(
...
-    protected virtual ValueTask<string> RenderAsync(ChatMessage message, CancellationToken cancellationToken)
+    /// <remarks>
+    /// The default implementation considers only the last message of <paramref name="messages"/>.
+    /// </remarks>
+    protected virtual ValueTask<string> RenderAsync(IEnumerable<ChatMessage> messages, CancellationToken cancellationToken)
     {
-        _ = Throw.IfNull(message, nameof(message));
+        _ = Throw.IfNullOrEmpty(messages);
+
+        ChatMessage message = messages.Last();
```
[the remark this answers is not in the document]
shyamnamboodiripad: Or better yet, we can create two overloads (both `protected virtual`) - 

* The first one would take a `ChatMessage` and render it (same as before).
* The second one would take a `ChatResponse` and render `ChatResponse.Text` (which would not include any tool call text by default - derived types can override to include something for tool calls if needed)

Looking at the usage in derived types I think this should work well since we always either call it for a single existing message in the history or for the complete model response.

cc @peterwald

shyamnamboodiripad (commented):
Looks mostly good to me but it would be great to address the comment around `RenderAsync` (and perhaps split it into two overloads as I commented [here](https://github.com/dotnet/extensions/pull/5998#discussion_r1982196541)).

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/ChatConversationEvaluator.cs
```diff
@@ -250,21 +250,26 @@ await PerformEvaluationAsync(
...
 
     /// <summary>
-    /// Renders the supplied <paramref name="message"/> to a string that can be included as part of the evaluation
+    /// Renders the supplied <paramref name="messages"/> to a string that can be included as part of the evaluation
     /// prompt that this <see cref="IEvaluator"/> uses.
     /// </summary>
-    /// <param name="message">
-    /// A message that is part of the conversation history for the response being evaluated and that is to be rendered
+    /// <param name="messages">
+    /// Messages that are part of the conversation history for the response being evaluated and that is to be rendered
```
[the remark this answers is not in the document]
shyamnamboodiripad: This comment change can be rolled back if we split `RenderAsync` into two overloads as mentioned [here](https://github.com/dotnet/extensions/pull/5998#discussion_r1982196541).

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -16,40 +16,33 @@ namespace Microsoft.Extensions.AI;
 /// </para>
 /// <para>
 /// However, implementations of <see cref="IChatClient"/> might mutate the arguments supplied to <see cref="GetResponseAsync"/> and
```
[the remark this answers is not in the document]
stephentoub: They can still mutate the arguments, e.g. options. This part is about thread-safety.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/ChatConversationEvaluator.cs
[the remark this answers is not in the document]
stephentoub: It's not strange that the two overloads would render messages differently? The existing one does '[role] content', but the one taking multiple messages would just be 'content'? Would it work to have them compose, such that by default the one taking multiple messages would do something like 'string.Join(Environment.NewLine, messages.SelectMany(m => RenderAsync(m)))'? I don't know the intent or use cases for these methods, so you tell me what makes sense :)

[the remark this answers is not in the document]
shyamnamboodiripad: If you decide to ignore tool calls via early out in the other `RenderAsync` overload that takes a single `ChatMessage` as I proposed above, then we should also update `CanRender` with a similar early out for tool calls.

This adds some complication since we would need to return `true` for `CanRender` in the early out case (even though the message is rejected) because `false` is used to signal that the token budget was breached, and no further messages from the supplied history should be included in the rendered evaluation prompt...

However, adding this early out in `CanRender` (in addition to `RenderAsync(ChatMessage)` as proposed above) would also be more correct - since the derived evaluators that use this check today do not care about tool calls in the history by default.

shyamnamboodiripad (approved):
Latest iteration looks great - thanks! :shipit:

stephentoub:
@SteveSandersonMS, if you're good with this, we can merge it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -0,0 +1,304 @@
...
+/// Provides extension methods for working with <see cref="ChatResponse"/> and <see cref="ChatResponseUpdate"/> instances.
+/// </summary>
+public static class ChatResponseExtensions
+{
+    /// <summary>Adds all of the messages from <paramref name="response"/> into <paramref name="list"/>.</summary>
+    /// <param name="list">The destination list into which the messages should be added.</param>
+    /// <param name="response">The response containing the messages to add.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="list"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="response"/> is <see langword="null"/>.</exception>
+    public static void AddMessages(this IList<ChatMessage> list, ChatResponse response)
```
SteveSandersonMS, line 26: From a discoverability point of view, I suspect that:

```cs
response.AppendTo(history);
```

... would be more obvious than:

```cs
history.AddMessages(response);
```

... on the grounds that `IList<T>` wouldn't be expected to know anything about `ChatResponse`, whereas a `ChatResponse`  would be expected to know about an `IList<ChatMessage>`.

I know that for people copy-pasting samples it won't make a difference either way. But for people coming at it from first principles it may be clearer.

This isn't a big deal. Will leave up to you whether you swap it round or not.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -16,40 +16,33 @@ namespace Microsoft.Extensions.AI;
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The response messages generated by the client.</returns>
-    /// <remarks>
-    /// The returned messages aren't added to <paramref name="chatMessages"/>. However, any intermediate messages generated implicitly
-    /// by the client, including any messages for roundtrips to the model as part of the implementation of this request, are included.
-    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="messages"/> is <see langword="null"/>.</exception>
     Task<ChatResponse> GetResponseAsync(
-        IList<ChatMessage> chatMessages,
+        IEnumerable<ChatMessage> messages,
```
SteveSandersonMS, line 34: Curious about the rationale for `IEnumerable<ChatMessage>` over (say) `IReadOnlyList<ChatMessage>`. I would imagine it's common for middleware to want to access, say, the last or last N entries without fear that they are introducing some O(N) step by forcing an enumerable to walk its entire length.

Perhaps we have some reasoning like:

 * We don't expect the history to become really long, e.g., hundreds of entries, since that would entail bigger costs anyway
 * In most cases it would perform the same anyway because the history is usually stored in an `IList<ChatMessage>`, and `.Last`/`.TakeLast` presumably checks for that special case

Does that line up with your thinking?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
SteveSandersonMS: Actually having seen more of the usage samples below now, I'm really not bothered about this. It feels fine either way.

SteveSandersonMS (approved):
Excellent! The additional clarity with this is really nice.

I added a couple of small questions but neither are intended to block you from merging here.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
RogerBarreto: I'm curious to understand what's the biggest tradeoff of losing the indexing/count information from the messages in the benefit of an delayed enumeration approach.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
```diff
@@ -38,6 +38,29 @@ public class ChatResponseUpdate
...
+    /// <summary>Initializes a new instance of the <see cref="ChatResponseUpdate"/> class.</summary>
+    [JsonConstructor]
+    public ChatResponseUpdate()
+    {
+    }
+
+    /// <summary>Initializes a new instance of the <see cref="ChatResponseUpdate"/> class.</summary>
+    /// <param name="role">The role of the author of the update.</param>
+    /// <param name="contents">The text contents of the update.</param>
+    public ChatResponseUpdate(ChatRole? role, string? contents)
```
RogerBarreto, line 50: Since this reflects a single `TextContent` i would keep the singular name for the property.

```suggestion
    public ChatResponseUpdate(ChatRole? role, string? textContent)
```

```suggestion
    public ChatResponseUpdate(ChatRole? role, string? content)
```

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -39,9 +39,12 @@ public sealed class AzureAIInferenceChatClient : IChatClient
     /// <summary>Initializes a new instance of the <see cref="AzureAIInferenceChatClient"/> class for the specified <see cref="ChatCompletionsClient"/>.</summary>
     /// <param name="chatCompletionsClient">The underlying client.</param>
     /// <param name="modelId">The ID of the model to use. If null, it can be provided per request via <see cref="ChatOptions.ModelId"/>.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="chatCompletionsClient"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="modelId"/> is empty or composed entirely of whitespace.</exception>
```
RogerBarreto, line 43: nit

```suggestion
    /// <exception cref="ArgumentNullException"><paramref name="modelId"/> is <see langword="null"/>, empty, or composed entirely of whitespace.</exception>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
stephentoub: I went back and forth on this, but ultimately landed on IEnumerable, and I think it's the right choice.

The type here dictates the contract, which is purely that a read-only sequence of messages is supplied. And that's what IEnumerable exists for, as the universal representation of a sequence of elements, that both all collections implement and that can be used for lazily-produced sequences. It is the most flexible input, enabling the best experience for consumers, who can pass in effectively any collection of messages they have.

This does not limit what the implementations of the abstractions can do. They can always type-test to see if the incoming messages provides additional interface or concrete type implementations, e.g. querying for IList and using it for direct indexing if it's available. LINQ itself does this all over the place, e.g. Count() will check if a collection is ICollection and return its Count if it is.

IReadOnlyList is also a bit problematic because IList does not inherit it (today). That means if you have something typed as `IList<ChatMessage>`, you would not be able to call `Get{Streaming}ResponseAsync` with it if those methods were typed to accept `IReadOnlyList<ChatMessage>`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
stephentoub: I went back and forth on this one. But since you raise it, I can change it back.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
stephentoub: It doesn't throw if it's null.

src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
```diff
@@ -27,105 +27,24 @@ of the abstractions.
...
-
-#### Sample Implementation
+The `IChatClient` interface defines a client abstraction responsible for interacting with AI services that provide "chat" capabilities. It defines methods for sending and receiving messages comprised of multi-modal content (text, images, audio, etc.), with responses providing either a complete result or one streamed incrementally. Additionally, it allows for retrieving strongly-typed services that may be provided by the client or its underlying services.
 
 .NET libraries that provide clients for language models and services may provide an implementation of the `IChatClient` interface. Any consumers of the interface are then able to interoperate seamlessly with these models and services via the abstractions.
 
-Here is a sample implementation of an `IChatClient` to show the general structure.
```
gewarren, line 36: @stephentoub Did you intentionally remove the sample IChatClient implementation here? For Learn docs, we're trying to figure out if/where to show the sample implementation. We're thinking about adding an "advanced" or "library authors" section to the .NET AI docs and putting it there.

stephentoub: >Did you intentionally remove the sample IChatClient implementation here?

Yes, to avoid the very first thing the README showed being a fake, non-functional implementation.

> We're thinking about adding an "advanced" or "library authors" section to the .NET AI docs and putting it there.

Sounds good.

PR #6002, opened 2025-02-28
Expose default cache key use by DistributedCachingCacheClient

stephentoub:
We've had requests for this to be available separate from DCCC. This just moves the method from the internal CacheHelpers type to DistributedCachingChatClient; the implementation isn't changed at all, other than in arg validation.


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -95,10 +104,155 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
...
     {
-        _jsonSerializerOptions.MakeReadOnly();
-        return CachingHelpers.GetCacheKey(values, _jsonSerializerOptions);
+        if (serializerOptions is null)
+        {
+            serializerOptions = AIJsonUtilities.DefaultOptions;
+        }
+        else
+        {
+            serializerOptions.MakeReadOnly();
```
stephentoub, line 136: @eiriktsarpalis, can you remind me why we need to do this?

paulbatum:
Thanks for exposing this. We have implemented a deduping logger that uses this key generation logic to make it easier to read test output that involves multi-turn conversations.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -95,10 +104,155 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
...
+    /// <remarks>
+    /// <para>
+    /// The <paramref name="values"/> are serialized to JSON using <see cref="JsonSerializerOptions"/> in order to compute the key.
+    /// </para>
+    /// <para>
+    /// The generated cache key is not guaranteed to be stable across releases of the library.
+    /// </para>
+    /// </remarks>
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values) =>
+        GetDefaultCacheKey(values, _jsonSerializerOptions);
```
SteveSandersonMS, line 116: Would there be any value in making this virtual instance method `public` as well? That would provide a path for people to get the correct cache key even if the logic is overridden by a subclass.

I know it's not as convenient as the static method since it forces the developer to instantiate (or have access to an existing instance of) the DCCC. But at least it would be possible.

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
stephentoub: > Would there be any value in making this virtual instance method public as well? That would provide a path for people to get the correct cache key even if the logic is overridden by a subclass.

Do you have an example for where that would be useful? If you're looking to emulate the exact behavior of a derived type in order to get the same cache keys that instance is using, you not only need to have the algorithm, but you also need to know what inputs are being passed to the method, and that's an implementation detail.

I'm not against it, I'm just unclear what it unlocks.

SteveSandersonMS: I don't have an end-to-end scenario in mind, so I'm glad you brought this up.

My comment assumed that whatever end-to-end scenario applies for the static logic would apply to per-type logic too. But now I'm wondering if the use case is different.

Are you saying the use case for the static logic isn't to anticipate the cache keys that we'd be using, but rather as a more general-purpose "hash a load of MEAI objects for reasons unrelated to DCCC"? If that's the case then arguably the API shouldn't be on DCCC.

stephentoub: > Are you saying the use case for the static logic isn't to anticipate the cache keys that we'd be using, but rather as a more general-purpose "hash a load of MEAI objects for reasons unrelated to DCCC"?

Yes, albeit still related to caching, just not via DCCC. Exposing it here seemed like a reasonable middle ground between exposing such that it _could_ be used for other things even if not promoted for that. But we could also move it to AIJsonUtilities if we want it to be an obviously broader thing.

stephentoub: Meh, I'll just move it. It's just a hash for hashing json serialized anything.

stephentoub: Moved it to be AIJsonUtilities.HashDataToString. ("HashData" matches the name used on our crypto types, and I included the "ToString" suffix because a) the crypto HashData produces bytes, and b) we might want a byte-based version in the future as well.)

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.cs
```diff
@@ -53,6 +63,101 @@ public static void AddAIContentType(this JsonSerializerOptions options, Type con
...
+        {
+            stream.Dispose();
+            throw;
+        }
+
+        IncrementalHashStream.ThreadStaticInstance = stream;
+
+        return Convert.ToHexString(hashData);
+#else
+        MemoryStream stream = new();
```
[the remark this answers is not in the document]
stephentoub: MemoryStream's dispose is a nop. There is no memory leak. As soon as it's no longer being used, it's available for garbage collection.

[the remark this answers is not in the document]
eddynaka: Is there a way to know what kind of Streams also work that way? I thought all kinds of streams should be disposed if not used. Is there a doc that I can read more about this behavior?
Thanks!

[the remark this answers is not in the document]
stephentoub: Technically anything that's IDisposable should be disposed when you're done with it. I just happen to know that MemoryStream.Dispose is a nop; it'd be fine to add a Dispose call in this code, but not having it doesn't hurt anything. In general, though, not disposing of a stream is not a memory leak. The stream, like any IDisposable, may be referencing resources, and ideally you'd clean them up promptly, but if you don't, eventually resources will be cleaned up by finalization when they're no longer used.

PR #6025, opened 2025-03-04
Generate the vector store index name using project name. Fix symbol handling for CSS and dimensions.

### LINKED ISSUE #6013: [Template] CSS issue when project name includes a dash
### Description

I used the template to create a project called `aiwebchat-ollama` and the page does not render correctly. I found the CSS bundle was returning a 404.

The `App.Razor` file was set to render a stylesheet with the href of `@Assets["aichatweb_ollama.styles.css"]`, replacing the `-` from my project name with an `_`. This fails because the assembly name retains the `-` without getting replaced by an `_`.

According to the https://github.com/dotnet/templating/wiki/Naming-and-default-value-forms, we need to be more explicit with our default `sourceName` so that the content can correctly identify which replaced form to use in each occurrence. For this particular occurrence, we should use a string that coerces the replacement to use the `identity` form.

To fix this properly, we should find all occurrences of `"ChatWithCustomData.Web"` and replace them with strings that have unambiguous replacement formats, and update the `sourceName` in `template.json` to reflect the new name.

### Reproduction Steps

```sh
mkdir aichatweb-ollama
cd aichatweb-ollama
dotnet new aichatweb -A ollama

dotnet run
```

### Expected behavior

The page should render with correct CSS using `aichatweb-ollama.styles.css`

### Actual behavior

The `aichatweb_ollama.styles.css` stylesheet returns a 404.

### Regression?

_No response_

### Known Workarounds

Update `App.razor` to fix the name to have the `-`.

### Configuration

_No response_

### Other information

_No response_

### LINKED ISSUE #6021: [Template] Prefix the "data" index name with the project name
When using the Microsoft.Extensions.AI.Templates aiwebchat template multiple times, it's easy to hit the issue of having a collision for the "data" index in the Azure AI Search index. The README calls out that the index might need to be deleted to address this pitfall.

We discussed prefixing the "data" index name with the project name to reduce collisions. To do so, we will need to take the [Azure AI Search naming rules](https://learn.microsoft.com/en-us/rest/api/searchservice/naming-rules?utm_source=chatgpt.com) into account as part of the template symbol replacements.

- Length: 2 - 128
- Case: Lower
- Allowable Characters: Letters, numbers, dashes (-), underscores (_)
- First character must be a letter or number
- No consecutive dashes or underscores

When this is implemented, the README needs to be updated to reflect this, using the symbol replacement there as well. The project code would use the symbol in these two locations:

https://github.com/dotnet/extensions/blob/8d1e6cb3e6c9619ea276ac31578796f62d1a34ab/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Services/Ingestion/DataIngestor.cs#L22

https://github.com/dotnet/extensions/blob/8d1e6cb3e6c9619ea276ac31578796f62d1a34ab/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web/Services/SemanticSearch.cs#L13

jeffhandley:
Fixes #6013
Fixes #6021

**[[Template] CSS issue when project name includes a dash (#6013)](https://github.com/dotnet/extensions/issues/6013)**
The CSS bundle asset name is generated using the assembly name, and it needs to support dashes, spaces, and other characters disallowed in C# symbols. The default project name we're using doesn't force that behavior per [Naming and default value forms · dotnet/templating Wiki](https://github.com/dotnet/templating/wiki/Naming-and-default-value-forms) documentation.

Other projects templates (such as the Blazor templates) use a template project name ending in `-CSharp` which coerces the value to use the "identity" form. Rather than renaming the project, this fix applies a minimal touch but using a custom token in that location and introducing a new symbol in the template configuration that is forced to use "identity" form.

**[[Template] Prefix the "data" index name with the project name (#6021)](https://github.com/dotnet/extensions/issues/6021)**
To accomplish this and not introduce new bugs, the generated name needs to adhere to the [Azure AI Search Naming Rules](https://learn.microsoft.com/en-us/rest/api/searchservice/naming-rules). This is handled by introducing a symbol with a custom transform that chains together manipulation to ensure the final value conforms to those rules.

**Also fixes the `[VectorStoreRecordVector]` dimensions for Ollama**
There was an error in the template, referencing the wrong symbol for checking if using Ollama. This resulted in `1536` for the dimensions instead of the expected `384` for Ollama.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6025)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -52,6 +52,12 @@
       "type": "bind",
       "binding": "HostIdentifier"
     },
+    "assemblyName": {
+      "type": "derived",
+      "valueSource": "name",
+      "valueTransform": "identity", // Avoids transformation,
```
RussKie, line 58: [nit]
```suggestion
      "valueTransform": "identity", // Avoids transformation
```

RussKie (approved):

SteveSandersonMS (approved):
@jeffhandley Thanks!

I have left a couple of comments about how this may break us when running from sources, but given the goal to ship so soon, perhaps we shouldn't block ourselves on that. If you've already gone through the process of validating these changes with various project names and template flags, let's go ahead with what you have here right now.

> Other projects templates (such as the Blazor templates) use a template project name ending in -CSharp which coerces the value to use the "identity" form.

I'd love it if we could follow up pretty much immediately - and I don't mind doing it myself - with an update to use the `-CSharp` naming format so that it will work both from sources and as a template package. I mean once preview 1 has shipped, that is.

Does that sound OK?

jeffhandley:
Thanks, @SteveSandersonMS. I didn't realize we could run from source too; great catch. I'll fix that in a new commit.

I had also gone through the `-CSharp` approach and that worked too. I'll illustrate that and we can choose which approach we want for today.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
jeffhandley: Thanks. Changed the approach to not need this; so the block is deleted.

artl93 (approved):
# Test Report (initial run)



| Directory                                | New Project | Build  | Run         |
|------------------------------------------|------------|--------|------------|
| test_A_azureopenai_V_azureaisearch_U_true  | Success    | Success | Success    |
| test_A_azureopenai_V_azureaisearch_U_false | Success    | Success | Success    |
| test_A_azureopenai_V_local_U_true         | Success    | Success | Success    |
| test_A_azureopenai_V_local_U_false        | Success    | Success | Success    |
| test_A_githubmodels_V_azureaisearch_U_true | Success    | Success | JEFF tested |
| test_A_githubmodels_V_azureaisearch_U_false | Success    | Success | not tested |
| test_A_githubmodels_V_local_U_true        | Success    | Success | Success    |
| test_A_githubmodels_V_local_U_false       | Success    | Success | Success    |
| test_A_ollama_V_azureaisearch_U_true      | Success    | Success | Success - manual    |
| test_A_ollama_V_azureaisearch_U_false     | Success    | Success | not tested |
| test_A_ollama_V_local_U_true              | Success    | Success | Success    |
| test_A_ollama_V_local_U_false             | Success    | Success | Success    |
| test_A_openai_V_azureaisearch_U_true      | Success    | Success | Success - manual    |
| test_A_openai_V_azureaisearch_U_false     | Success    | Success | not tested |
| test_A_openai_V_local_U_true              | Success    | Success | Success - manual    |
| test_A_openai_V_local_U_false             | Success    | Success | not tested |
 

Mind you, "run" only validates ingestion - it does not validate that the end-to-end works in the browser.

RussKie (approved):

MackinnonBuck (approved):

PR #6041, opened 2025-03-05
Fix load order of ProjectStaging and restrict package version validation to only official builds

RussKie:
Reinstate #5864.


RussKie:
Tested in in the internal CI leg as a [normal](https://dev.azure.com/dnceng/internal/_build/results?buildId=2656260&view=results) and a [release](https://dev.azure.com/dnceng/internal/_build/results?buildId=2656259&view=results) branches.

RussKie:
Tested in https://dev.azure.com/dnceng/internal/_build/results?buildId=2660769 and compared artifacts to the latest release (9.3).

evgenyfedorov2 (approved):

eng/MSBuild/ProjectStaging.props
```diff
@@ -0,0 +1,26 @@
...
+    <Description Condition="'$(Stage)' == 'obsolete'">Obsolete Package. $(Description)</Description>
+
+    <!--
+        When DotNetFinalVersionKind is set to 'release', this branch will produce stable outputs for 'Shipping' packages
+
+        In 8.0, we shipped Microsoft.AspNetCore.Testing package as stable even when it was in the dev stage, so we
+        keep it as stable for compatiblity.
+      -->
+    <_IsStable Condition="('$(Stage)' != 'dev' and '$(Stage)' != 'preview') Or '$(MSBuildProjectName)' == 'Microsoft.AspNetCore.Testing'">true</_IsStable>
+    <DotNetFinalVersionKind Condition=" '$(DotNetFinalVersionKind)' == '' And '$(_IsStable)' == 'true' ">release</DotNetFinalVersionKind>
```
joperezr, line 15: We don't want this to be in main. In main we want to have this set to empty.

eng/MSBuild/ProjectStaging.props
```diff
@@ -0,0 +1,26 @@
...
+
+    <!-- Preview packages: do not use stable branding and do not warn about lack of [Experimental] -->
+    <NoWarn Condition="'$(Stage)' == 'dev' or '$(Stage)' == 'preview'">$(NoWarn);LA0003</NoWarn>
+    <!--
+        Makes it such that the package version won't be stabilized even when the rest of the repo is going stable.
+        https://github.com/dotnet/arcade/blob/main/Documentation/CorePackages/Versioning.md#package-version
+      -->
+    <SuppressFinalPackageVersion />
+    <SuppressFinalPackageVersion Condition=" '$(configureDotNetFinalVersionKind)' == 'true' And '$(_IsStable)' != 'true' ">true</SuppressFinalPackageVersion>
```
joperezr, line 24: Is `configureDotNetFinalVersionKind` right? Or did you mean `DotNetFinalVersionKind`?

joperezr (commented):
Other than the above comments, the rest of the changes look good to me. I'm assuming you've ran a stable build with this and validated that the right packages are still marked as prerelease?

eng/MSBuild/ProjectStaging.props
joperezr: Ok I do see you did mean the right property based on the pipeline changes, but can you help me understand what is this for? I'm not sure I'm following as we tipically wouldn't want to set this globally.

RussKie: Boo, I copied the wrong property name. I'll fix it, thank you.

RussKie:
> Other than the above comments, the rest of the changes look good to me. I'm assuming you've ran a stable build with this and validated that the right packages are still marked as prerelease?

I'm running tests now.

eng/MSBuild/ProjectStaging.props
RussKie: This property is only set whenever we're producing a signed build, which excludes the main and pull-requests. We don't want to stabilise for any other branch, right?

eng/MSBuild/ProjectStaging.props
```diff
@@ -0,0 +1,28 @@
...
+    <DotNetFinalVersionKind Condition=" '$(configureDotNetFinalVersionKind)' == 'true' And '$(DotNetFinalVersionKind)' == '' And '$(_IsStable)' == 'true' ">release</DotNetFinalVersionKind>
+
+    <!-- Preview packages: do not use stable branding and do not warn about lack of [Experimental] -->
+    <NoWarn Condition="'$(Stage)' == 'dev' or '$(Stage)' == 'preview'">$(NoWarn);LA0003</NoWarn>
+    <!--
+        Makes it such that the package version won't be stabilized even when the rest of the repo is going stable.
+        https://github.com/dotnet/arcade/blob/main/Documentation/CorePackages/Versioning.md#package-version
+      -->
+    <SuppressFinalPackageVersion />
+    <SuppressFinalPackageVersion Condition=" '$(_IsStable)' != 'true' ">true</SuppressFinalPackageVersion>
```
joperezr, line 26: You shouldn't need to set this. By default when DotNetFinalVersionKind is not release then packages will not be marked as stable, so no need to set this explicitly unless the rest of the repo is going stable and you want to opt-out in a package. I'm pretty sure you can remove both of these lines (25-26) and the behavior should be the same.

eng/MSBuild/ProjectStaging.props
```diff
@@ -0,0 +1,28 @@
+<Project>
+
+  <PropertyGroup>
+    <!-- Amend the description based on stage -->
+    <Description Condition="'$(Stage)' == 'dev'">Experimental package. $(Description)</Description>
+    <Description Condition="'$(Stage)' == 'obsolete'">Obsolete Package. $(Description)</Description>
+
+    <_IsStable Condition="('$(Stage)' != 'dev' and '$(Stage)' != 'preview') Or '$(MSBuildProjectName)' == 'Microsoft.AspNetCore.Testing'">true</_IsStable>
```
joperezr, line 8: This should be set to false for main.

eng/MSBuild/ProjectStaging.props
```diff
@@ -0,0 +1,28 @@
...
+
+    <!--
+        When DotNetFinalVersionKind is set to 'release' (only for the release branches),
+        the build will produce stable outputs for 'Shipping' packages.
+
+        In 8.0, we shipped Microsoft.AspNetCore.Testing package as stable even when it was in the dev stage, so we
+        keep it as stable for compatiblity.
+      -->
+    <DotNetFinalVersionKind Condition=" '$(configureDotNetFinalVersionKind)' == 'true' And '$(DotNetFinalVersionKind)' == '' And '$(_IsStable)' == 'true' ">release</DotNetFinalVersionKind>
```
RussKie, line 17:
```suggestion
    <DotNetFinalVersionKind Condition=" '$(ProduceStablePackages)' == 'true' And '$(DotNetFinalVersionKind)' == '' And '$(_IsStable)' == 'true' ">release</DotNetFinalVersionKind>
```

evgenyfedorov2 (approved):

eng/MSBuild/ProjectStaging.targets
```diff
@@ -22,10 +16,17 @@
...
-    <Description Condition="'$(Stage)' == 'dev'">Experimental package. $(Description)</Description>
-    <Description Condition="'$(Stage)' == 'obsolete'">Obsolete Package. $(Description)</Description>
-  </PropertyGroup>
+  <!--
+   Produce errors if we don't have all the right property values for non-production stages.
+
+   Note, that Arcade resets $(_PreReleaseLabel) to "ci" for non-official builds, which breaks the validation.
+   See Arcade SDK/Version.BeforeCommonTargets.targets for more details.
+    -->
+  <Target Name="_ValidateVersion" AfterTargets="GenerateNuspec" Condition=" '$(OfficialBuild)' == 'true' and '$(_IsStable)' != 'true' ">
```
joperezr, line 25: Where is `OfficialBuild` set? And what is the difference between that and `OfficialBuildId` which is the common one that gets set?

eng/Versions.props
```diff
@@ -9,9 +9,9 @@
     <ApiCompatBaselineVersion>9.3.0</ApiCompatBaselineVersion>
     <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
     <!--
-        When DotNetFinalVersionKind is set to 'release', this branch will produce stable outputs for 'Shipping' packages
+        When StabilizePackageVersion is set to 'true', this branch will produce stable outputs for 'Shipping' packages
     -->
-    <DotNetFinalVersionKind />
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
```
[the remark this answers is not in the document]
joperezr: Awesome, thanks!

src/Libraries/Microsoft.AspNetCore.Testing/Microsoft.AspNetCore.Testing.csproj
```diff
@@ -12,6 +12,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <Stage>dev</Stage>
```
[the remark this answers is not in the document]
joperezr: It has always been in the dev stage, we just pushed it as stable at the beginning by mistake, so we have continued to do so. This means that while the package is stable, all of it's APIs are still marked as experimental. If we are ready to stop doing that, then we can bump the stage in a subsequent PR.

src/Libraries/Microsoft.Extensions.Hosting.Testing/Microsoft.Extensions.Hosting.Testing.csproj
```diff
@@ -13,6 +13,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <Stage>dev</Stage>
```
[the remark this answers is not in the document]
joperezr: Same comment for all of these. This PR isn't changing the stage of any packages, and instead it is just changing where it is declared.

Directory.Build.props
```diff
@@ -105,6 +105,19 @@
   <!-- Enable the Evaluation report builds on all CI builds -->
   <PropertyGroup>
     <EnableEvaluationReportBuild Condition="'$(ContinuousIntegrationBuild)'=='true'">true</EnableEvaluationReportBuild>
+
+    <!--
+      ProjectStaging.props depends on the $(Stage) property, which is defined in the project file.
```
joperezr, line 110: Thanks for the detailed comment :smiley:!

azure-pipelines.yml
```diff
@@ -31,6 +31,7 @@ pr:
       - main
       - dev
       - release/*
+      - validation/*
```
joperezr, line 34: Do we really want this to run for PRs against validation too? Or simply with manually queued builds?

joperezr (approved):
Left a couple of super minor comments, but this looks good to go otherwise. Thanks a lot for the hard work on this and on getting it back and enabled.

azure-pipelines.yml
RussKie: Good spot, I put this into the wrong trigger group, moved to the branches group.

eng/MSBuild/ProjectStaging.targets
RussKie: It is being set by the Arcade SDK [here](https://github.com/dotnet/arcade/blob/1560b3406a8ebf9e267f7a1dddbf5238dd378081/src/Microsoft.DotNet.Arcade.Sdk/tools/DefaultVersions.props#L9-L12):
```xml
  <PropertyGroup>
    <OfficialBuild>false</OfficialBuild>
    <OfficialBuild Condition="'$(OfficialBuildId)' != ''">true</OfficialBuild>
  </PropertyGroup>
```

RussKie:
Ok, the build is green, merging.

MackinnonBuck (approved):

PR #6058, opened 2025-03-09
Address M.E.VectorData feedback for IEmbeddingGenerator

stephentoub:
Necessary, according to @roji:
- Adds a non-generic IEmbeddingGenerator, with GetService then moving down to the non-generic from the generic. This also simplifies the GetService/GetRequiredService extension methods, which no longer need to be generic on TInput/TEmbedding.

Optional (we should discuss which of these we actually want to take):
- Deletes GeneratedEmbeddings and makes GenerateAsync return an `IAsyncEnumerable<TEmbedding>` instead of `GeneratedEmbeddings<TEmbedding>`. The positive aspect of this is it allows for embeddings to be streamed as they're generated (though the main embedding generation services today don't stream), and also enables TEmbedding to be covariant (though it's not clear how much benefit that actually yields). The negative is we lose a place to hang metadata that's about the whole batch, in particular usage data. This moves that usage data to instead be on the Embedding, which is helpful for cases where such information is available per embedding, but isn't ideal for cases where it's about the whole batch. This PR deals with that by just putting such information onto one of the generated embeddings. We'll need to discuss the tradeoff. The alternative is we keep GeneratedEmbeddings but make it an IAsyncEnumerable instead of an IList. That gives a place to hang such batch information, but only if it's all available at the start of the streaming; however, it's also a bit harder to consume, as you need to await the initial call and then await foreach the embeddings.
- Adds a SupportedInputMediaTypes set to EmbeddingGeneratorMetadata, so that a generator can be explicit about what inputs it supports.
- Adds an optional MediaType parameter to TextContent. This would allow, for example, tagging code by language. 
- Makes the DataContent's ReadOnlyMemory constructor require a media type to be supplied. Ideally we'd also require this for non-data urls, but that's a bit trickier, as we can't distinguish between data and non-data urls in the type system. We could check for it dynamically. (This one wasn't actually mentioned by @roji.)

Not addressed:
- Removing IDisposable from IEmbeddingGenerator (and IChatClient), or adding IAsyncDisposable. Not having these implementations makes middleware awkward, as all middleware components need to expose the interface. It also makes it harder to know you might need to dispose the result of a method like AsEmbeddingGenerator. To my knowledge, we don't know of cases where async disposal is necessary.
- Add back a Metadata property. I still think `GetService<EmbeddingGeneratorMetadata>()` is sufficient, but we can discuss it further. We might want to consider an analyzer that would make diagnostic suggestions about supporting it in GetService and filling in as much data as possible.
- Splitting off a UrlContent from DataContent, to differentiate in a strongly typed way cases where data is included or not.

cc: @roji, @SteveSandersonMS 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6058)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
@@ -28,18 +28,18 @@ namespace Microsoft.Extensions.AI;
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 public class DataContent : AIContent
 {
-    // Design notes:
-    // - Ideally DataContent would be based in terms of Uri. However, Uri has a length limitation that makes it prohibitive
-    //   for the kinds of data URIs necessary to support here. As such, this type is based in strings.
+    // Design note:
+    // Ideally DataContent would be based in terms of Uri. However, Uri has a length limitation that makes it prohibitive
+    // for the kinds of data URIs necessary to support here. As such, this type is based in strings.
+
+    /// <summary>Parsed data URI information.</summary>
+    private readonly DataUriParser.DataUri? _dataUri;
 
     /// <summary>The string-based representation of the URI, including any data in the instance.</summary>
     private string? _uri;
 
     /// <summary>The data, lazily initialized if the data is provided in a data URI.</summary>
     private ReadOnlyMemory<byte>? _data;
-
-    /// <summary>Parsed data URI information.</summary>
-    private DataUriParser.DataUri? _dataUri;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="DataContent"/> class.
@@ -120,15 +120,17 @@ public class DataContent : AIContent
     }
 
     /// <summary>
-    /// Determines whether the <see cref="MediaType"/> has the specified prefix.
+    /// Determines whether the <see cref="MediaType"/>'s top-level type matches the specified <paramref name="topLevelType"/>.
     /// </summary>
-    /// <param name="prefix">The media type prefix.</param>
-    /// <returns><see langword="true"/> if the <see cref="MediaType"/> has the specified prefix; otherwise, <see langword="false"/>.</returns>
+    /// <param name="topLevelType">The type to compare against <see cref="MediaType"/>.</param>
+    /// <returns><see langword="true"/> if the type portion of <see cref="MediaType"/> matches the specified value; otherwise, false.</returns>
     /// <remarks>
-    /// This performs an ordinal case-insensitive comparison of the <see cref="MediaType"/> against the specified <paramref name="prefix"/>.
+    /// A media type is primarily composed of two parts, a "type" and a "subtype", separated by a slash ("/").
+    /// The type portion is also referred to as the "top-level type"; for example,
+    /// "image/png" has a top-level type of "image". <see cref="HasTopLevelMediaType"/> compares
+    /// the specified <paramref name="topLevelType"/> against the type portion of <see cref="MediaType"/>.
     /// </remarks>
-    public bool MediaTypeStartsWith(string prefix) =>
-        MediaType.StartsWith(prefix, StringComparison.OrdinalIgnoreCase) is true;
+    public bool HasTopLevelMediaType(string topLevelType) => DataUriParser.HasTopLevelMediaType(MediaType, topLevelType);
 
     /// <summary>Gets the data URI for this <see cref="DataContent"/>.</summary>
     /// <remarks>
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
@@ -8,17 +8,17 @@ using System.Text.Json.Serialization;
 using Microsoft.Shared.Diagnostics;
 
 #pragma warning disable S3996 // URI properties should not be strings
+#pragma warning disable CA1054 // URI-like parameters should not be strings
 #pragma warning disable CA1056 // URI-like properties should not be strings
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>
-/// Represents data content, such as an image or audio.
+/// Represents binary content with an associated media type (also known as MIME type).
 /// </summary>
 /// <remarks>
 /// <para>
-/// The represented content may either be the actual bytes stored in this instance, or it may
-/// be a URI that references the location of the content.
+/// The content represents in-memory data. For references to data at a remote URI, use <see cref="UriContent"/> instead.
 /// </para>
 /// <para>
 /// <see cref="Uri"/> always returns a valid URI string, even if the instance was constructed from
@@ -28,9 +28,9 @@ namespace Microsoft.Extensions.AI;
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 public class DataContent : AIContent
 {
-    // Design note:
-    // Ideally DataContent would be based in terms of Uri. However, Uri has a length limitation that makes it prohibitive
-    // for the kinds of data URIs necessary to support here. As such, this type is based in strings.
+    // Design notes:
+    // - Ideally DataContent would be based in terms of Uri. However, Uri has a length limitation that makes it prohibitive
+    //   for the kinds of data URIs necessary to support here. As such, this type is based in strings.
 
     /// <summary>The string-based representation of the URI, including any data in the instance.</summary>
     private string? _uri;
@@ -44,8 +44,15 @@ public class DataContent : AIContent
     /// <summary>
     /// Initializes a new instance of the <see cref="DataContent"/> class.
     /// </summary>
-    /// <param name="uri">The URI of the content. This can be a data URI.</param>
-    /// <param name="mediaType">The media type (also known as MIME type) represented by the content.</param>
+    /// <param name="uri">The data URI containing the content.</param>
+    /// <param name="mediaType">
+    /// The media type (also known as MIME type) represented by the content. If not provided,
+    /// it must be provided as part of the <paramref name="uri"/>.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="uri"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="uri"/> is not a data URI.</exception>
+    /// <exception cref="ArgumentException"><paramref name="uri"/> did not contain a media type and <paramref name="mediaType"/> was not supplied.</exception>
+    /// <exception cref="ArgumentException"><paramref name="mediaType"/> is an invalid media type.</exception>
     public DataContent(Uri uri, string? mediaType = null)
         : this(Throw.IfNull(uri).ToString(), mediaType)
     {
@@ -54,42 +61,48 @@ public class DataContent : AIContent
     /// <summary>
     /// Initializes a new instance of the <see cref="DataContent"/> class.
     /// </summary>
-    /// <param name="uri">The URI of the content. This can be a data URI.</param>
+    /// <param name="uri">The data URI containing the content.</param>
     /// <param name="mediaType">The media type (also known as MIME type) represented by the content.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="uri"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="uri"/> is not a data URI.</exception>
+    /// <exception cref="ArgumentException"><paramref name="uri"/> did not contain a media type and <paramref name="mediaType"/> was not supplied.</exception>
+    /// <exception cref="ArgumentException"><paramref name="mediaType"/> is an invalid media type.</exception>
     [JsonConstructor]
     public DataContent([StringSyntax(StringSyntaxAttribute.Uri)] string uri, string? mediaType = null)
     {
         _uri = Throw.IfNullOrWhitespace(uri);
 
-        ValidateMediaType(ref mediaType);
-        MediaType = mediaType;
-
-        if (uri.StartsWith(DataUriParser.Scheme, StringComparison.OrdinalIgnoreCase))
+        if (!uri.StartsWith(DataUriParser.Scheme, StringComparison.OrdinalIgnoreCase))
         {
-            _dataUri = DataUriParser.Parse(uri.AsMemory());
+            Throw.ArgumentException(nameof(uri), "The provided URI is not a data URI.");
+        }
 
-            // If the data URI contains a media type that's different from a non-null media type
-            // explicitly provided, prefer the one explicitly provided as an override.
-            if (MediaType is not null)
+        _dataUri = DataUriParser.Parse(uri.AsMemory());
+
+        if (mediaType is null)
+        {
+            mediaType = _dataUri.MediaType;
+            if (mediaType is null)
             {
-                if (MediaType != _dataUri.MediaType)
-                {
-                    // Extract the bytes from the data URI and null out the uri.
-                    // Then we'll lazily recreate it later if needed based on the updated media type.
-                    _data = _dataUri.ToByteArray();
-                    _dataUri = null;
-                    _uri = null;
-                }
-            }
-            else
-            {
-                MediaType = _dataUri.MediaType;
+                Throw.ArgumentNullException(nameof(mediaType), $"{nameof(uri)} did not contain a media type, and {nameof(mediaType)} was not provided.");
             }
         }
-        else if (!System.Uri.TryCreate(uri, UriKind.Absolute, out _))
+        else
         {
-            throw new UriFormatException("The URI is not well-formed.");
+            if (mediaType != _dataUri.MediaType)
+            {
+                // If the data URI contains a media type that's different from a non-null media type
+                // explicitly provided, prefer the one explicitly provided as an override.
+
+                // Extract the bytes from the data URI and null out the uri.
+                // Then we'll lazily recreate it later if needed based on the updated media type.
+                _data = _dataUri.ToByteArray();
+                _dataUri = null;
+                _uri = null;
+            }
         }
+
+        MediaType = DataUriParser.ThrowIfInvalidMediaType(mediaType);
     }
 
     /// <summary>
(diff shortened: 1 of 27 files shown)


stephentoub:
> Specifically for language

I was referring to programming languages, for eg code generated by the llm as part of a code interpreter tool, like text/x-python, text/x-csharp, or application/typescript.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/TextContent.cs
```diff
@@ -33,6 +37,27 @@ public string Text
         set => _text = value;
     }
 
+    /// <summary>Gets or sets the text media type (also known as MIME type) of the content.</summary>
+    [JsonPropertyOrder(1)]
+    public string? MediaType
```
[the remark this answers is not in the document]
stephentoub: I don't think it rises to that level of importance. For data content, you really need the mime type in any situation to know how to interpret the bytes. That's not the case for text content.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -97,9 +97,13 @@ public DataContent([StringSyntax(StringSyntaxAttribute.Uri)] string uri, string?
     /// </summary>
     /// <param name="data">The byte contents.</param>
     /// <param name="mediaType">The media type (also known as MIME type) represented by the content.</param>
-    public DataContent(ReadOnlyMemory<byte> data, string? mediaType = null)
+    /// <exception cref="ArgumentNullException"><paramref name="mediaType"/> is null.</exception>
+    /// <exception cref="ArgumentException"><paramref name="mediaType"/> is empty or composed entirely of whitespace.</exception>
+    public DataContent(ReadOnlyMemory<byte> data, string mediaType)
```
westey-m, line 102: Another common format we may want to support is base64 encoded strings.
I did some comparisons between the most popular image embedding services and the two most popular formats were data urls and base64 encoded strings.

See https://github.com/microsoft/semantic-kernel/issues/10492#issuecomment-2657456868

stephentoub: That would just be an optimization to avoid decoding the data, right? eg
https://github.com/dotnet/extensions/blob/736529fccdeba82484bcfef434fae9208b9219ad/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceEmbeddingGenerator.cs#L97
and it would need to be a situation where you weren't going to decode it, anyway. Does that show up?

stephentoub: Also, IEmbeddingGenerator isn't returning DataContent, it's returning Embedding.

stephentoub: (And a custom Embedding-derived type could handle that.)

westey-m: This is for data input to the `EmbeddingGenerator` for images.
I'm assuming we want to use `DataContent` as input in this case?

It would be an optimization to avoid having to either convert the base64 string to a data uri (not terribly hard granted) or a byte array before calling `IEmbeddingGenerator.GenerateEmbeddingAsync`.

stephentoub: Where is the base64 string input coming from?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/TextContent.cs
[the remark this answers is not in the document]
SteveSandersonMS: I might be wrong but my gut reaction here is that adding this muddies the concepts.

Before this change, app developers knew that `TextContent` is a shorthand for "human-readable text you can present in your UI". But after adding this, app devs can no longer assume that and must put in additional checks everywhere to reason about how to handle some text they have received.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorMetadata.cs
```diff
@@ -48,4 +68,43 @@ public EmbeddingGeneratorMetadata(string? providerName = null, Uri? providerUri
...
+
+    /// <summary>
+    /// Gets a read-only list of media types (also known as MIME types) the generator has declared to support as inputs.
+    /// </summary>
+    /// <remarks>
+    /// If <see langword="null"/>, the generator has not made any declarations about supported types.
+    /// If non-<see langword="null"/>, the generator supports only the listed types, and providing data
+    /// that is of another type may lead to exceptions or unexpected results.
+    /// </remarks>
+    public ISet<string>? SupportedInputMediaTypes { get; }
```
SteveSandersonMS, line 80: It's a bit odd to use media types to define the set of supported input types. Media types, as a concept, only apply to serialized data and a `TInput` isn't necessarily any kind of serialized data.

SteveSandersonMS: I guess more broadly I'm not clear on how this information would get used. In this PR, I don't see examples of it getting used. Could you give an example of how application code would interact with this?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
westey-m: That would depend on the source of the data that needs to be embedded.  Images could be extracted from an existing document, or be loaded from disk or be downloaded from a database or fetched from a rest service.

stephentoub:
@roji, @westey-m, @SteveSandersonMS, I've pushed a revision based on our discussion. This now only includes:
- Lowering GetService down to a non-generic IEmbeddingGenerator
- Splitting DataContent into DataContent and UriContent. The former requires bytes, the latter is about remote content.

@RogerBarreto, FYI

@shyamnamboodiripad, @peterwald, I don't think this should affect you much if at all, but FYI.

stephentoub:
cc: @luisquintanilla, since you were previously asking about DataContent representing both in-memory data and remote content

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -167,10 +175,9 @@ public string Uri
     /// If the media type was explicitly specified, this property returns that value.
     /// If the media type was not explicitly specified, but a data URI was supplied and that data URI contained a non-default
     /// media type, that media type is returned.
-    /// Otherwise, this property returns null.
     /// </remarks>
-    [JsonPropertyOrder(1)]
-    public string? MediaType { get; private set; }
+    [JsonIgnore]
```
SteveSandersonMS, line 179: Is this because the JSON representation is meant to be just the data URI which should include this information anyway?

stephentoub: Yes

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/UriContent.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// <exception cref="ArgumentNullException"><paramref name="mediaType"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="mediaType"/> is an invalid media type.</exception>
+    /// <exception cref="UriFormat"><paramref name="uri"/> is an invalid URL.</exception>
+    /// <remarks>
+    /// A media type must be specified, so that consumers know what to do with the content.
+    /// If an exact media type is not known, but the category (e.g. image) is known, a wildcard
+    /// may be used (e.g. "image/*").
+    /// </remarks>
+    public UriContent(string uri, string mediaType)
+        : this(new Uri(Throw.IfNull(uri)), mediaType)
```
SteveSandersonMS, line 40: Should we consider throwing if `uri` is a data URI, forcing people to use `DataContent` for those?

I'm guessing you didn't because a data URI *is* a URI and hence there's no strict reason why we should block it here.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/IEmbeddingGenerator.cs
```diff
@@ -2,42 +2,17 @@
...
-/// avoid using shared instances of these arguments for concurrent invocations or should otherwise ensure by construction that
-/// no <see cref="IEmbeddingGenerator{TInput, TEmbedding}"/> instances are used which might employ such mutation.
-/// </para>
+/// This base interface is used to allow for embedding generators to be stored in a non-generic manner.
+/// To use the generator to create embeddings, instances typed as this base interface first need to be
+/// cast to the generic interface <see cref="IEmbeddingGenerator{TInput, TEmbedding}"/>.
 /// </remarks>
-public interface IEmbeddingGenerator<in TInput, TEmbedding> : IDisposable
-    where TEmbedding : Embedding
+public interface IEmbeddingGenerator : IDisposable
```
SteveSandersonMS, line 14: Do you think `EmbeddingGeneratorBuilderServiceCollectionExtensions` should be updated to register the service as both `IEmbeddingGenerator<TInput, TEmbedding>` and `IEmbeddingGenerator`?

My guess is probably not because we don't really have use cases for people acquiring `IEmbeddingGenerator` from DI.

stephentoub: Hmm. I guess folks could enumerate all registered IEmbeddingGenerators and then use GetService to determine which they really want to use? There are other ways to achieve the end goals that would likely be in service of. So, I don't know. I could go either way. What would you like the answer to be?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -109,20 +123,14 @@ public DataContent(ReadOnlyMemory<byte> data, string? mediaType = null)
...
-        {
-            Throw.ArgumentException(nameof(mediaType), "Invalid media type.");
-        }
-    }
+    /// <returns><see langword="true"/> if the <see cref="MediaType"/> has the specified prefix; otherwise, <see langword="false"/>.</returns>
+    /// <remarks>
+    /// This performs an ordinal case-insensitive comparison of the <see cref="MediaType"/> against the specified <paramref name="prefix"/>.
+    /// </remarks>
+    public bool MediaTypeStartsWith(string prefix) =>
+        MediaType.StartsWith(prefix, StringComparison.OrdinalIgnoreCase) is true;
```
SteveSandersonMS, line 131: This is a bit dubious because `MediaType` might be something like `image/jpeg2000`, and someone might check for `MediaTypeStartsWith("image/jpeg")` and that would return `true` even though it's an incompatible binary format.

(That's actually a fake example because the media type for JPEG2000 is `image/jp2`, probably for this exact reason.)

I'm not 100% sure what the benefit of this API is, versus letting people call `MediaType.StartsWith` directly. If we are going to have a special API for it, that could be an opportunity to make it a bit smarter, e.g., by only matching when the prefix is followed by a word separator.

If we avoided having a special API then at least it would be completely clear that developers are responsible for handling the word boundary cases themselves.

Not a big deal either way, but curious if you've got specific thoughts around this.

SteveSandersonMS (approved):
Nice!

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/UriContent.cs
SteveSandersonMS: The only payoff to being more constrained is if we think `IChatClient` implementations are likely not to understand the case where `UriContent` is a data URI and then for something to go wrong. For example an `IChatClient` implementation might be forced by its own upstream APIs to distinguish between "remote URIs" and "data URIs" and might incorrectly use our `UriContent`/`DataContent` distinction to model that.

I think we don't really gain much by imposing an extra restriction here. If the `IChatClient` implementation does something wrong, that's their bug and they should fix it.

So overall no, I'd vote with keeping it as-is.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
stephentoub: I copied this from what we already have in DataContent.

The one in DataContent was added when MediaType might be null, so it was a bit more complicated, albeit not much:
```C#
=> _mediaType?.StartsWith(prefix, StringComparer.OrdinalIgnoreCase) is true;
```
Now that it can't be null, it's simplified slightly:
```C#
=> _mediaType.StartsWith(prefix, StringComparer.OrdinalIgnoreCase)
```
and it really is just delegation to StartsWith.

But it's also something that needs to be used on most consumption of the type, to determine how to route it, so we do want it to be as simple as possible.

If there's a better / more robust implementation, happy to have this be that. By separated by a word separator, you mean have it not only check the prefix but that what comes next is either not a regex word character or is end-of-string, i.e. as if it were doing a regex with `^{prefix}\b`?

stephentoub: Also if we want it to be more than delegating to StartsWith, maybe we want it to have a slightly different name, like MediaTypeHasPrefix, so that `MediaTypeStartsWith(...)` and `MediaType.StartsWith(...)` aren't confused.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/IEmbeddingGenerator.cs
westey-m: I'm for registering it. Even if we only have one `IEmbeddingGenerator` registered, it will be useful if we want to build a common/reusable decorator for `VectorStoreRecordCollection` that automatically generates embeddings as needed.  We won't know what type the input is until `Search` is called, so we cannot expect a specific typed IEmbeddingGenerator interface on the `EmbeddingVectorStoreRecordCollection` constructor.

```csharp
class EmbeddingVectorStoreRecordCollection<TKey, TRecord>(IEmbeddingGenerator embeddingGenerator)
{
    public async Task<Results> Search<TInputData>(TInputData data, VectorSearchOptions options)
    {
        // Take the type of TInput and check if the embedding generator supports this as input,
        // with the output type of the target vector (options.VectorProperty).
    }
}
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -28,9 +28,9 @@ namespace Microsoft.Extensions.AI;
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 public class DataContent : AIContent
 {
-    // Design note:
-    // Ideally DataContent would be based in terms of Uri. However, Uri has a length limitation that makes it prohibitive
-    // for the kinds of data URIs necessary to support here. As such, this type is based in strings.
+    // Design notes:
```
roji, line 31: Unrelated nit: `_dataUri` looks like it can be readonly

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/IEmbeddingGenerator.cs
roji: Yeah, as discussed in the meeting, there's the idea of AddQdrant() automatically pulling a DI-registered non-generic IEmbeddingGenerator to use as the default (@SteveSandersonMS this is the question you weren't convinced on, and I don't think we reached a clear decision). That would be the main reason to have the non-generic registration IMHO.

As I said, according to my mental model, the 95% case would be applications doing text->`ROM<float>`, and it's nice for those to be able to just register once and have it propagate - as an overridable default - everywhere, rather than having to configure different components separately with the same generator.

Re what happens with multiple generator registrations: I think that's going to be quite a rare case, and is also a general problem in DI that isn't limited to embedding generators (am not sure if the standard behavior is first- or last-one wins, but either one seems reasonable).

So to summarize, I think we should to this to save the typical basic user from having to go through more registration/setup ceremony, just because the more advanced scenarios with multiple generators would require a bit more special care. But I don't think this question is critical either.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
roji: Yeah, I had this thought too at some point.

The argument for this is e.g. people who get an image embedded inside JSON (where base64 is the standard binary encoding) and can forward it directly for embedding generation; that would help avoid decoding that to byte[], only to have that re-encoded to base64 for providers which require it.

But in most other cases where people get image data (disk, database, REST service...), I doubt the data would come encoded as base64... The downside of adding this API option is that providers which **don't** support base64 would have to implement decoding internally, or not support base64 and throw (limiting the usability of the abstraction).

I'm on the fence... Just not sure having base64 image data on-hand is that common...

roji (approved):
LGTM (see comments)!

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -137,8 +147,8 @@ public string Uri
             {
                 if (_dataUri is null)
                 {
-                    Debug.Assert(Data is not null, "Expected Data to be initialized.");
-                    _uri = string.Concat("data:", MediaType, ";base64,", Convert.ToBase64String(Data.GetValueOrDefault()
+                    Debug.Assert(_data is not null, "Expected _data to be initialized.");
+                    _uri = string.Concat("data:", MediaType, ";base64,", Convert.ToBase64String(_data.GetValueOrDefault()
```
RogerBarreto, line 151: Once the `_dataUri` is generated once from the `_data`, since it's a readonly whouldn't it be nice to cache it and vice-versa?

```suggestion
                    _uri = string.Concat("data:", MediaType, ";base64,", Convert.ToBase64String(_data.GetValueOrDefault()
```

stephentoub: Where would _dataUri end up being used again?

PR #6061, opened 2025-03-10
Fix evaluation report generation

shyamnamboodiripad:
The .NET side code for the `ScenarioRunResult` was recently changed (#5998) to include `ChatResponse` (which can contain multiple `ChatMessage`s) in place of a single `ChatMessage`. Unfortunately, we missed updating the TypeScript reporting code to account for this.

This change fixes the problem by updating the deserialization code in TypeScript to match what .NET code serializes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6061)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/EvalTypes.d.ts
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/EvalTypes.d.ts
@@ -13,9 +13,13 @@ type ScenarioRunResult = {
     executionName: string;
     creationTime?: string;
     messages: ChatMessage[];
-    modelResponse: ChatMessage;
+    modelResponse: ChatResponse;
     evaluationResult: EvaluationResult;
 };
+
+type ChatResponse = {
+    messages: ChatMessage[];
+}
 
 type ChatMessage = {
     authorName?: string;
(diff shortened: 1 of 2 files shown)


shyamnamboodiripad:
@peterwald I also noticed that the previously cached responses for the tests are no longer compatible after the changes in #5998. I didn't look too closely, but iirc the cached response was being read and returned without any exceptions - however, the tests (like `AdditionalContextIsNotPassed`) were throwing when they tried to get the response message to be evaluated by calling `response.Messages.Single()` - iirc, `response.Messages` was empty.

I am guessing this is because `ChatResponse` had a different shape previously (and only contained a single `ChatMessage`) and `CachingChatClient` ends up doing a best effort deserialization of this cached response which fails to deserialize the single existing `ChatMessage` into the `Messages` collection.

I didn't look too closely since the tests passed without any issues after I blew away and rebuilt the cache and I can't reproduce the problem again without syncing back to an older commit to rebuild the problematic cache state... 

But if you do run into it, it may be a good idea to confirm the root cause above :) The behavior is probably expected and ok for now since both MEAI and MEAI.Evaluation are in preview (and since we will not be making compat-breaking changes once everything is GA). But it may be interesting to understand why this happened and double-check that more innocuous (additive) changes to types like `ChatMesage` won't cause similar problems in the future.

peterwald:
On my local env, I added code that can detect the old version and display it properly. We should discuss whether that is the best way forward. We could consider adding a console feature to "upgrade" old data.

peterwald:
> On my local env, I added code that can detect the old version and display it properly. We should discuss whether that is the best way forward. 

On further investigation, this probably won't work. Any usage of the Console program to generate the report will run through the serializer and remove the old formatted data.

> We could consider adding a console feature to "upgrade" old data.

This could work. Not sure how manageable it would be though.

peterwald:
I did confirm that it doesn't crash with old reporting data however.

peterwald (approved):

Issue #6062, opened 2025-03-10
AI Extensions - strict schema support for OpenAI models


danielwinkler:
### Description

I noticed that the structured output of OpenAI chat completions did not match the schema reliably after switching from the native openai client to the AI extensions library.

One such example is: 

```json
{
  "summary": "....",
  "question": "....",
  "answer": "....",
  "title": "...."
}
``` 

sometimes gets returned as 

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "summary": "....",
    "question": "....",
    "answer": "....",
    "title": "...."
  },
  "required": [
    "summary",
    "question",
    "answer",
    "title"
  ],
  "additionalProperties": false
}
```

Investigating the library, a possible reason for this could be in the mapping to OpenAI options:

```csharp
            else if (options.ResponseFormat is ChatResponseFormatJson jsonFormat)
            {
                result.ResponseFormat = jsonFormat.Schema is { } jsonSchema ?
                    OpenAI.Chat.ChatResponseFormat.CreateJsonSchemaFormat(
                        jsonFormat.SchemaName ?? "json_schema",
                        BinaryData.FromBytes(
                            JsonSerializer.SerializeToUtf8Bytes(jsonSchema, OpenAIJsonContext.Default.JsonElement)),
                        jsonFormat.SchemaDescription) :
                    OpenAI.Chat.ChatResponseFormat.CreateJsonObjectFormat();
            }
```

namely, the `jsonSchemaIsStrict` parameter of `CreateJsonSchemaFormat` receives the default value `null`, see

```xml
<param name="jsonSchemaIsStrict">
    <para>
        Whether to enable strict schema adherence when generating the response. If set to <c>true</c>, the
        model will follow the exact schema defined in <paramref name="jsonSchema"/>.
    </para>
    <para>
        Only a subset of the JSON schema specification is supported when this is set to <c>true</c>. Learn more
        in the
        <see href="https://platform.openai.com/docs/guides/structured-outputs">structured outputs guide</see>.
    </para>
</param> 
```

```csharp
public static ChatResponseFormat CreateJsonSchemaFormat(string jsonSchemaFormatName, BinaryData jsonSchema, string jsonSchemaFormatDescription = null, bool? jsonSchemaIsStrict = null)
```

### Reproduction Steps

- Create a record for a schema
- `private record ExampleSchema(string summary, string question, string answer, string title)`
- Create a schema
[...]

PR #6064, opened 2025-03-10
[AI extensions] Add jsonSchemaIsStrict option to OpenAI options mapping

danielwinkler:
See https://github.com/dotnet/extensions/issues/6062

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
@@ -403,7 +403,8 @@ internal static partial class OpenAIModelMappers
                         jsonFormat.SchemaName ?? "json_schema",
                         BinaryData.FromBytes(
                             JsonSerializer.SerializeToUtf8Bytes(jsonSchema, OpenAIJsonContext.Default.JsonElement)),
-                        jsonFormat.SchemaDescription) :
+                        jsonFormat.SchemaDescription,
+                        jsonSchemaIsStrict: true) :
                     OpenAI.Chat.ChatResponseFormat.CreateJsonObjectFormat();
             }
         }
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantClient.cs
@@ -267,7 +267,8 @@ internal sealed class OpenAIAssistantClient : IChatClient
                     AssistantResponseFormat.CreateJsonSchemaFormat(
                         jsonFormat.SchemaName ?? "json_schema",
                         BinaryData.FromBytes(JsonSerializer.SerializeToUtf8Bytes(jsonSchema, OpenAIJsonContext.Default.JsonElement)),
-                        jsonFormat.SchemaDescription) :
+                        jsonFormat.SchemaDescription,
+                        jsonSchemaIsStrict: true) :
                     AssistantResponseFormat.JsonObject;
             }
         }


danielwinkler:
@dotnet-policy-service agree

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -403,7 +403,8 @@ public static ChatCompletionOptions ToOpenAIOptions(ChatOptions? options)
                         jsonFormat.SchemaName ?? "json_schema",
                         BinaryData.FromBytes(
                             JsonSerializer.SerializeToUtf8Bytes(jsonSchema, OpenAIJsonContext.Default.JsonElement)),
-                        jsonFormat.SchemaDescription) :
+                        jsonFormat.SchemaDescription,
+                        jsonSchemaIsStrict: true) :
```
eiriktsarpalis, line 407: Hardcoding this value doesn't seem right given that this is just a mapper from a `ChatOptions` instance. Should it be sourced from there, e.g. using a value from `AdditionalProperties`?

danielwinkler: Making it configurable sounds like a reasonable approach.

However, adding it to the options increases the complexity of the library's API, making it less intuitive to achieve the intended behavior.

For instance, when calling:  
```csharp
public static async Task<ChatResponse<T>> GetResponseAsync<T>
```
`options.ResponseFormat` gets overridden. This means that setting the format in the options will not take effect, as the template parameter will dictate the response format instead. While this isn't necessarily problematic (since they should align in most cases), it could lead to some unexpected behavior for users.

Additionally, if `useNativeJsonSchema` is not explicitly set, the response won't use the native JSON schema format but will instead include an extra message prompting a schema. Once this behavior is understood, users will also need to know that another flag must be included in the options to enable actual "structured output."

In most cases, developers using an OpenAI model that supports structured output and calling the templated version of `GetResponseAsync` would likely expect the "json_schema" version of the OpenAI API to be used under the hood. Making this behavior more intuitive or clearly documented could improve the developer experience.

eiriktsarpalis: > However, adding it to the options increases the complexity of the library's API, making it less intuitive to achieve the intended behavior.

I have no strong opinion on this, @SteveSandersonMS might be able to weight in more authoritatively here. I'm just pointing that this is a mapping function being used by multiple components. It might make sense to hardcode for `GetResponseAsync<T>`, but it might be surprising in others.

eiriktsarpalis, line 407: Does this require any changes for the inverse `FromOpenAIOptions` mapper so that everything roundtrips properly?

stephentoub:
What are the downsides to just always setting strict? I think it's "just" that if a schema is provided the OpenAI doesn't understand, it'll fail? Should we just always set strict and then optionally use something in additional properties as an override?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
SteveSandersonMS: OpenAI seems to have 3 levels of structured output:

A. **Prompting alone**

  * equivalent to what we have with `useNativeJsonSchema: false`, or equivalently, `options.ResponseFormat = ChatResponseFormat.Json`

B. **JSON schema provided but strict=false**

  * for OpenAIChatClient, seemingly what we do today for `useNativeJsonSchema: true` or equivalently, `options.ResponseFormat = ChatResponseFormat.ForJsonSchema(...)`

C. **JSON schema provided with strict=true**

  * for OpenAIChatClient, not equivalent to anything we have today, at least prior to this PR

The whole thing is a bit awkward because other providers (at least Ollama) have only A and C. The weird middle-ground B appears to be OpenAI-only.

I'm not totally sure what the point of B is, especially given that it's not widespread across other providers. For MEAI I would be happy enough with simplifying it to:

 * `useNativeJsonSchema:false`: does prompt-based structured output (A above)
 * `useNativeJsonSchema:true`: does strict native structured output (C above)

As such we would indeed set `jsonSchemaIsStrict: true` whenever we have a nonnull `jsonFormat.Schema`, exactly as done in this PR. Anyone who doesn't want that should use `useNativeJsonSchema:false` or perhaps use some OpenAI-specific config to force it into the weird middle-ground B.

---

Separately, I think now is the time for us to flip the default on `useNativeJsonSchema`. At the time we introduced it, most OpenAI models didn't support it, so we defaulted to `false` so that `GetResponseAsync<T>` would work by default. But now, most OpenAI models do support it, and Ollama supports it for all models. It's now becoming niche to want to do non-native structured output, and models/providers are only going to get better over time. So before MEAI ships with a weird old-fashioned default, I think we should change this to default to `true`.

Any provider that doesn't support native structured output at all won't fail, because they will already have to translate the request into something that provider supports. And any provider that supports it for everything (e.g., Ollama) won't fail either. The only ones that might incur extra issues if we change this default are ones that support it only for certain models, and for those it's reasonable for the developer to have to set `useNativeJsonSchema:false`.

@eiriktsarpalis @stephentoub What do you think?

SteveSandersonMS:
> What are the downsides to just always setting strict? I think it's "just" that if a schema is provided the OpenAI doesn't understand, it'll fail? Should we just always set strict and then optionally use something in additional properties as an override?

There's nothing in [OpenAI's announcement](https://openai.com/index/introducing-structured-outputs-in-the-api/) that clarifies the drawbacks. Maybe they just didn't know whether it would always work.

As per [this longer explanation](https://github.com/dotnet/extensions/pull/6064#discussion_r1989349126), I'd support always setting strict if there's a schema. And like you say, if OpenAI really needs to support some non-strict mode, that's very OpenAI-specific and could be handled by something in additional properties to override this.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
stephentoub: That all makes sense to me.

danielwinkler: Thank you for the investigation and clarification on the types of structured output, I'd also agree with your reasoning

> Separately, I think now is the time for us to flip the default on `useNativeJsonSchema`. At the time we introduced it, most OpenAI models didn't support it, so we defaulted to `false` so that `GetResponseAsync<T>` would work by default. But now, most OpenAI models do support it, and Ollama supports it for all models. It's now becoming niche to want to do non-native structured output, and models/providers are only going to get better over time. So before MEAI ships with a weird old-fashioned default, I think we should change this to default to `true`.

From a users perspective, I support this move.

Another consideration when changing the behavior is regarding `options.ResponseFormat`, maybe only set it to the provided template parameter if it is not yet set in the options; or report an argument exception if it is set already.

Do you need anything from my side on this PR?

stephentoub:
I pushed an update to your branch to fix up the remaining strict handling.

stephentoub (approved):
Thanks.

Issue #6034, opened 2025-03-05
[AI Evaluation] Introduce tags / metadata concept for executions


shyamnamboodiripad:
Currently, the evaluation reporting APIs have support for specifying a name for a particular execution / evaluation run. However, there is no way to associate additional metadata for an execution (such as the model used, settings like temperature etc.) and we have heard some feedback that this could be useful.

A workaround could be to include this metadata as part of the execution name(e.g. `Execution 123 (Model: gpt-4o, Temperature: 1.0, ...`). However, this may not be very flexible and can get unwieldy as the number of tags increases.
 
The proposal is to introduce a way to specify a set of `string` tags / metadata for each execution that can then be included at the top of the report html (alongside the execution name). For this we would need to do the following -

- [ ] Update the generated report to display execution name at the top.
- [ ] Introduce API for specifying execution tags / metadata. This could probably be specified when constructing the `ReportingConfiguration`. We would probably also need a new API on `IResultStore` to store this metadata in the appropriate storage provider (i.e., Disk, Azure Storage etc.).
- [ ] Update the generated report to display the metadata at the top.

PR #6070, opened 2025-03-10
[AI Evaluation] Add versioning to reporting and caching

peterwald:
These are basically independent escape mechanisms to handle future changes in formats.

`ReportFormatVersion` - Will record which version of the format was used to persist the scenario run, as well as exposing which version the current library will write. This should allow us to convert/adapt to reporting format changes in the future if needed.

`CacheVersion` - Will change all the caching keys causing the existing cache entries to be no longer valid. @stephentoub @SteveSandersonMS 

Hopefully neither of these will ever be needed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6070)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
@@ -30,4 +30,16 @@ public static class Defaults
     /// in the <see cref="IResponseCacheProvider"/>'s cache before they are considered expired and evicted.
     /// </summary>
     public static TimeSpan DefaultTimeToLiveForCacheEntries { get; } = TimeSpan.FromDays(14);
+
+    /// <summary>
+    /// Defines the version number for the reporting format. If and when the serialized format undergoes breaking changes, this number
+    /// will be incremented to indicate the version a report file was created with compared to the version this library will produce.
+    /// </summary>
+    public const int ReportingFormatVersion = 1;
+
+    /// <summary>
+    /// Defines an version number for the cache format. Changing this value will break the cache key and force requests
+    /// to retrieve fresh data from the source until the version is updated again.
+    /// </summary>
+    public const int CacheVersion = 1;
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
@@ -32,8 +32,8 @@ public static class Defaults
     public static TimeSpan DefaultTimeToLiveForCacheEntries { get; } = TimeSpan.FromDays(14);
 
     /// <summary>
-    /// Defines the version number for the reporting format. If and when the serialized format undergoes breaking changes, this number
-    /// will be incremented to indicate the version a report file was created with compared to the version this library will produce.
+    /// Defines the version number for the reporting format. If and when the serialized format undergoes
+    /// breaking changes, this number will be incremented.
     /// </summary>
-    public const int ReportingFormatVersion = 1;
+    internal const int ReportingFormatVersion = 1;
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
@@ -37,9 +37,4 @@ public static class Defaults
     /// </summary>
     public const int ReportingFormatVersion = 1;
 
-    /// <summary>
-    /// Defines an version number for the cache format. Changing this value will break the cache key and force requests
-    /// to retrieve fresh data from the source until the version is updated again.
-    /// </summary>
-    public const int CacheVersion = 1;
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
@@ -36,5 +36,4 @@ public static class Defaults
     /// will be incremented to indicate the version a report file was created with compared to the version this library will produce.
     /// </summary>
     public const int ReportingFormatVersion = 1;
-
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
@@ -86,7 +86,7 @@ public sealed class ScenarioRunResult(
     /// <summary>
     /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
     /// </summary>
-    public int FormatVersion { get; init; } = formatVersion;
+    public int FormatVersion { get; set; } = formatVersion;
 
     /// <summary>
     /// Gets or sets the <see cref="ScenarioRun.ScenarioName"/>.
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
@@ -84,9 +84,9 @@ public sealed class ScenarioRunResult(
     }
 
     /// <summary>
-    /// Gets or sets the version of the format used to persist the current <see cref="ScenarioRunResult"/>.
+    /// Gets the version of the format used to persist the current <see cref="ScenarioRunResult"/>.
     /// </summary>
-    public int? FormatVersion { get; set; } = formatVersion ?? Defaults.ReportingFormatVersion;
+    public int? FormatVersion { get; } = formatVersion ?? Defaults.ReportingFormatVersion;
 
     /// <summary>
     /// Gets or sets the <see cref="ScenarioRun.ScenarioName"/>.
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Extensions.AI.Evaluation.Reporting;
 /// The <see cref="Evaluation.EvaluationResult"/> for the <see cref="ScenarioRun"/> corresponding to the
 /// <see cref="ScenarioRunResult"/> being constructed.
 /// </param>
-/// <param name="formatVersion">The version of the format used to persist this scenario run.</param>
+/// <param name="formatVersion">The version of the format used to persist the current <see cref="ScenarioRunResult"/>.</param>
 [method: JsonConstructor]
 public sealed class ScenarioRunResult(
     string scenarioName,
@@ -47,7 +47,7 @@ public sealed class ScenarioRunResult(
     IList<ChatMessage> messages,
     ChatResponse modelResponse,
     EvaluationResult evaluationResult,
-    int formatVersion = Defaults.ReportingFormatVersion)
+    int? formatVersion = null)
 {
     /// <summary>
     /// Initializes a new instance of the <see cref="ScenarioRunResult"/> class.
@@ -84,9 +84,9 @@ public sealed class ScenarioRunResult(
     }
 
     /// <summary>
-    /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
+    /// Gets or sets the version of the format used to persist the current <see cref="ScenarioRunResult"/>.
     /// </summary>
-    public int FormatVersion { get; set; } = formatVersion;
+    public int? FormatVersion { get; set; } = formatVersion ?? Defaults.ReportingFormatVersion;
 
     /// <summary>
     /// Gets or sets the <see cref="ScenarioRun.ScenarioName"/>.
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
@@ -86,7 +86,7 @@ public sealed class ScenarioRunResult(
     /// <summary>
     /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
     /// </summary>
-    public int FormatVersion { get; set; } = formatVersion;
+    public int FormatVersion { get; init; } = formatVersion;
 
     /// <summary>
     /// Gets or sets the <see cref="ScenarioRun.ScenarioName"/>.
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
@@ -37,6 +37,7 @@ namespace Microsoft.Extensions.AI.Evaluation.Reporting;
 /// The <see cref="Evaluation.EvaluationResult"/> for the <see cref="ScenarioRun"/> corresponding to the
 /// <see cref="ScenarioRunResult"/> being constructed.
 /// </param>
+/// <param name="formatVersion">The version of the format used to persist this scenario run.</param>
 [method: JsonConstructor]
 public sealed class ScenarioRunResult(
     string scenarioName,
@@ -45,7 +46,8 @@ public sealed class ScenarioRunResult(
     DateTime creationTime,
     IList<ChatMessage> messages,
     ChatResponse modelResponse,
-    EvaluationResult evaluationResult)
+    EvaluationResult evaluationResult,
+    int formatVersion = Defaults.ReportingFormatVersion)
 {
     /// <summary>
     /// Initializes a new instance of the <see cref="ScenarioRunResult"/> class.
@@ -80,6 +82,11 @@ public sealed class ScenarioRunResult(
                 evaluationResult)
     {
     }
+
+    /// <summary>
+    /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
+    /// </summary>
+    public int FormatVersion { get; set; } = formatVersion;
 
     /// <summary>
     /// Gets or sets the <see cref="ScenarioRun.ScenarioName"/>.
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
@@ -107,7 +107,7 @@ public class DistributedCachingChatClient : CachingChatClient
     /// The generated cache key is not guaranteed to be stable across releases of the library.
     /// </para>
     /// </remarks>
-    protected override string GetCacheKey(params ReadOnlySpan<object?> values) 
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values)
     {
         // Bump the cache version to invalidate existing caches if the serialization format changes in a breaking way.
         const int CacheVersion = 1;
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
@@ -107,6 +107,11 @@ public class DistributedCachingChatClient : CachingChatClient
     /// The generated cache key is not guaranteed to be stable across releases of the library.
     /// </para>
     /// </remarks>
-    protected override string GetCacheKey(params ReadOnlySpan<object?> values) =>
-        AIJsonUtilities.HashDataToString(values, _jsonSerializerOptions);
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values) 
+    {
+        // Bump the cache version to invalidate existing caches if the serialization format changes in a breaking way.
+        const int CacheVersion = 1;
+
+        return AIJsonUtilities.HashDataToString([CacheVersion, .. values], _jsonSerializerOptions);
+    }
 }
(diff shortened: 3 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
```diff
@@ -81,6 +83,11 @@ public ScenarioRunResult(
     {
     }
 
+    /// <summary>
+    /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
+    /// </summary>
+    public int FormatVersion { get; set; } = formatVersion;
```
shyamnamboodiripad, line 89: A couple of questions / suggestions -

* How would we use this version within lower / inner layers of the report? (For example, if there has been a change to `EvaluationMetric` or `EvaluationResult` between versions and we need to change the deserialization accordingly...)

* Can we somehow put this at a higher level in the report (and access it in the lower level / inner objects)? The change for the following issue will introduce at least one additional top-level artifact for run-level metadata in report (which will sit outside / above the `ScenarioRunResult` level). So was wondering if there is some way to make a single version that applies to all the data. #6034 

* Alternatively, perhaps the version should be part of the JSON that is generated by the `aieval` tool and we can make the tool responsible for managing all data transformations between versions (and keep the result objects themselves agnostic)?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -107,6 +107,11 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
     /// The generated cache key is not guaranteed to be stable across releases of the library.
     /// </para>
     /// </remarks>
-    protected override string GetCacheKey(params ReadOnlySpan<object?> values) =>
-        AIJsonUtilities.HashDataToString(values, _jsonSerializerOptions);
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values) 
+    {
+        // Bump the cache version to invalidate existing caches if the serialization format changes in a breaking way.
+        const int CacheVersion = 1;
```
stephentoub, line 113: Hopefully we never need to use this, but having it as a failsafe just-in-case seems reasonable.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -107,6 +107,11 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
     /// The generated cache key is not guaranteed to be stable across releases of the library.
     /// </para>
     /// </remarks>
-    protected override string GetCacheKey(params ReadOnlySpan<object?> values) =>
-        AIJsonUtilities.HashDataToString(values, _jsonSerializerOptions);
+    protected override string GetCacheKey(params ReadOnlySpan<object?> values)
```
stephentoub, line 110:
```suggestion
    protected override string GetCacheKey(params ReadOnlySpan<object?> values)
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
peterwald: > * How would we use this version within lower / inner layers of the report? (For example, if there has been a change to `EvaluationMetric` or `EvaluationResult` between versions and we need to change the deserialization accordingly...)

There isn't currently a need to version at every level. Data is persisted only at the `ScenarioRunResult` level today, which is why I put it there. This a blunt tool, and only intended to make managing the results easier.

> * Can we somehow put this at a higher level in the report (and access it in the lower level / inner objects)? The change for the following issue will introduce at least one additional top-level artifact for run-level metadata in report (which will sit outside / above the `ScenarioRunResult` level). So was wondering if there is some way to make a single version that applies to all the data. [[AI Evaluation] Introduce tags / metadata concept for executions #6034](https://github.com/dotnet/extensions/issues/6034)

There is no persistence level above `ScenarioRunResult` today. When/If we add more data at another level, then we would need to consider how versioning information is kept there. It might be necessary to add version to whatever artifact is used to store tags/metadata when it is implemented.

> * Alternatively, perhaps the version should be part of the JSON that is generated by the `aieval` tool and we can make the tool responsible for managing all data transformations between versions (and keep the result objects themselves agnostic)?

The `aieval` tool should be version aware, certainly. And it's the logical choice for any migration functionality we might build. However, `aieval` doesn't generate the reporting results, it only consumes them (at least today). Versioning needs to be part of the report generation code as well.

shyamnamboodiripad (approved):

peterwald:
@dotnet/dotnet-extensions-ai This change is ready to merge if you approve.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
RussKie, line 89: Should this be readonly?
```suggestion
    public int FormatVersion { get; init; } = formatVersion;
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
```diff
@@ -30,4 +30,11 @@ public static class Defaults
     /// in the <see cref="IResponseCacheProvider"/>'s cache before they are considered expired and evicted.
     /// </summary>
     public static TimeSpan DefaultTimeToLiveForCacheEntries { get; } = TimeSpan.FromDays(14);
+
+    /// <summary>
+    /// Defines the version number for the reporting format. If and when the serialized format undergoes breaking changes, this number
+    /// will be incremented to indicate the version a report file was created with compared to the version this library will produce.
+    /// </summary>
+    public const int ReportingFormatVersion = 1;
+
```
RussKie, line 39: [nit]
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
```diff
@@ -81,6 +83,11 @@ public ScenarioRunResult(
     {
     }
 
+    /// <summary>
+    /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
+    /// </summary>
+    public int FormatVersion { get; init; } = formatVersion;
```
shyamnamboodiripad, line 89: We previously made the call to make the reporting types completely mutable. I think it would be best to stick to that and make this property mutable too to match everything else.

peterwald: I don't have a strong opinion either way, but happy to roll it back.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -107,6 +107,11 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
...
+    {
+        // Bump the cache version to invalidate existing caches if the serialization format changes in a breaking way.
+        const int CacheVersion = 1;
+
+        return AIJsonUtilities.HashDataToString([CacheVersion, .. values], _jsonSerializerOptions);
```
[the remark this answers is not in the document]
peterwald: Agreed, but I won't be able to bring this change in unless we ship M.E.AI (which I don't think is in the plan currently).

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
```diff
@@ -30,4 +30,10 @@ public static class Defaults
     /// in the <see cref="IResponseCacheProvider"/>'s cache before they are considered expired and evicted.
     /// </summary>
     public static TimeSpan DefaultTimeToLiveForCacheEntries { get; } = TimeSpan.FromDays(14);
+
+    /// <summary>
+    /// Defines the version number for the reporting format. If and when the serialized format undergoes breaking changes, this number
+    /// will be incremented to indicate the version a report file was created with compared to the version this library will produce.
+    /// </summary>
+    public const int ReportingFormatVersion = 1;
```
stephentoub, line 38: What's the expected use of this public const?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
```diff
@@ -45,7 +46,8 @@ public sealed class ScenarioRunResult(
     DateTime creationTime,
     IList<ChatMessage> messages,
     ChatResponse modelResponse,
-    EvaluationResult evaluationResult)
+    EvaluationResult evaluationResult,
+    int formatVersion = Defaults.ReportingFormatVersion)
```
stephentoub, line 50: We actually want the version to be part of the signature like this?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
```diff
@@ -81,6 +83,11 @@ public ScenarioRunResult(
     {
     }
 
+    /// <summary>
+    /// Gets or sets the Version of the format used to persist this scenario run. You should not modify this value from the default.
```
stephentoub, line 87: > You should not modify this value from the default.

Then why is it settable?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
shyamnamboodiripad: Good catch - I think this should be internal.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
shyamnamboodiripad: The reporting types are all completely mutable - we made this call earlier during API reviews as that just makes things simpler for serialization (and to avoid the need for special constructors etc.). We also decided to keep things simple and prefer `set` over `init` in all cases. I don't think we should revisit that call or deviate from it solely for this property :)

stephentoub: Then why say it shouldn't be changed?

The comment is just at odds with it being mutable.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
shyamnamboodiripad: It is a valid question if we could somehow omit the version from appearing in the public API. However, there doesn't seem to be an easy way to avoid this since the version needs to be included as part of the serialization of the results (so that we can know what the original serialized version was in the case where we need to do any special translation during deserialization to the 'current' version).

Our report storage is basically a collection of `ScenarioRunResult`s - it is currently the root object in the tree of result types, and we don't have a more root / wrapping concept - so this seemed like the best place to store this version for now.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
shyamnamboodiripad: Ah but I guess it has to be public so that it can be used as the default value in the constructor 🤔

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
shyamnamboodiripad: Ah good catch - sorry I missed the quoted text in your original comment. I agree, I think we should remove the guidance.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
```diff
@@ -37,6 +37,7 @@ namespace Microsoft.Extensions.AI.Evaluation.Reporting;
 /// The <see cref="Evaluation.EvaluationResult"/> for the <see cref="ScenarioRun"/> corresponding to the
 /// <see cref="ScenarioRunResult"/> being constructed.
 /// </param>
+/// <param name="formatVersion">The version of the format used to persist this scenario run.</param>
```
shyamnamboodiripad, line 40:
```suggestion
/// <param name="formatVersion">The version of the format used to persist the current <see cref="ScenarioRunResult"/>.</param>
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
shyamnamboodiripad, line 87:
```suggestion
    /// Gets or sets the version of the format used to persist the current <see cref="ScenarioRunResult"/>.
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
stephentoub: It couldn't be `int? formatVersion = null`?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
```diff
@@ -30,4 +30,10 @@ public static class Defaults
     /// in the <see cref="IResponseCacheProvider"/>'s cache before they are considered expired and evicted.
     /// </summary>
     public static TimeSpan DefaultTimeToLiveForCacheEntries { get; } = TimeSpan.FromDays(14);
+
+    /// <summary>
+    /// Defines the version number for the reporting format. If and when the serialized format undergoes breaking changes, this number
+    /// will be incremented to indicate the version a report file was created with compared to the version this library will produce.
```
shyamnamboodiripad, line 36:
```suggestion
    /// Defines the version number for the reporting format. If and when the serialized format undergoes breaking
    /// changes, this number will be incremented.
```

I was trying to reword the deleted part above to be a little less confusing but couldn't come up with a good enough replacement. Perhaps we can just omit that part?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Defaults.cs
peterwald: The thought was that someone using the library might want to check to determine which version of the reporting format is supported, but I'm not opposed to making it internal.

peterwald:
Updated based on your feedback @shyamnamboodiripad @stephentoub

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
```diff
@@ -81,6 +83,11 @@ public ScenarioRunResult(
     {
     }
 
+    /// <summary>
+    /// Gets or sets the version of the format used to persist the current <see cref="ScenarioRunResult"/>.
+    /// </summary>
+    public int? FormatVersion { get; set; } = formatVersion ?? Defaults.ReportingFormatVersion;
```
stephentoub, line 89: I suggest this should instead be:
```suggestion
    public int? FormatVersion { get; set; } = formatVersion;
```
or:
```suggestion
public int FormatVersion { get; set; } = formatVersion ?? Defaults.ReportingFormatVersion;
```

test/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Tests/ScenarioRunResultTests.cs
```diff
@@ -38,6 +38,7 @@ public void SerializeScenarioRunResult()
             messages: [new ChatMessage(ChatRole.User, "prompt")],
             modelResponse: new ChatResponse(new ChatMessage(ChatRole.Assistant, "response")),
             evaluationResult: new EvaluationResult(booleanMetric, numericMetric, stringMetric, metricWithNoValue));
+        Assert.Equal(Defaults.ReportingFormatVersion, entry.FormatVersion);
```
stephentoub, line 41: If it remains nullable, this should probably also validate what behavior you expect when FormatVersion is set to null and then read again. Does it get null? Does it get the default?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
stephentoub, line 89: Should FormatVersion be incorporated into the cache key? Otherwise, who's consuming this?

peterwald: I was under the impression that the type in the JSON constructor must be the same as the property type for the serialization logic to work. But I could be wrong about that. I'll test it.

test/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Tests/ScenarioRunResultTests.cs
peterwald: If FormatVersion is set to null, I'd expect it to remain null. We don't expect users to change the value, but if they did then it would be whatever they set it to.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ScenarioRunResult.cs
eiriktsarpalis: Types need to match with the parameter.

peterwald: > I was under the impression that the type in the JSON constructor must be the same as the property type for the serialization logic to work

Yes, confirmed that if the constructor is `int?` then the property type must match it. So the second option won't work.

I don't understand the reason for the first suggestion. We want to have the version set to the default in the library, so we don't have to remember to set it every time we use `ScenarioRunResult`. The only reason the property is settable, is because it is a JSON serializable object. I suppose we could make it readonly.

stephentoub: > We are consuming this from previously persisted copies of our ScenarioRunResults (i.e. stored on disk or Azure storage, and created by potentially earlier versions of our library).

Sorry, I'm not following. Nothing appears to be using this property at all. Am I missing it? And if nothing's using it, why does the ctor parameter / property need to exist?

peterwald: It is for future proofing. JSON persistence writes the value of this property to disk. In future versions, we will check the value of the property and determine whether we need to modify the data before using it. 

We have run into this issue already with the ChatResponse change, and this is intended to help us solve the problem in the future.

stephentoub: >  JSON persistence writes the value of this property to disk.

Where does that happen?

stephentoub: Oh... is it that instances of this object are being JSON serialized, and so the property comes along for the ride? Got it.

stephentoub: > Types need to match with the parameter.

Only if the parameter is used for roundtripping, right? There's no mechanism to say "yes, I see there's a property with the same name as this parameter, but the types don't match and that property's setter should be used instead of the constructor"?

peterwald: It happens any time a `ScenarioRunResult` is saved to an `IResultStore`. For example, [here](https://github.com/dotnet/extensions/blob/e351c2506f2af13be2645fbd554c89fb9bd32abf/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Storage/DiskBasedResultStore.cs#L101) in the DiskBased version.

stephentoub: If you're happy with it as is it, that's fine, don't let my feedback stop you.

It's just weird to me that the property is nullable but defaults to a non-null value, can then be set to null, but really shouldn't be, and if it is set to null, then it behaves differently than if null is provided to the constructor, as if null is provided to the constructor, it'll then default to the default value.

peterwald: > Only if the parameter is used for roundtripping, right? There's no mechanism to say "yes, I see there's a property with the same name as this parameter, but the types don't match and that property's setter should be used instead of the constructor"?

I think there is such a mechanism. When I try it the Deserialize method throws this exception...

> System.InvalidOperationException : Each parameter in the deserialization constructor on type 'Microsoft.Extensions.AI.Evaluation.Reporting.ScenarioRunResult' must bind to an object property or field on deserialization. Each parameter name must match with a property or field on the object. Fields are only considered when 'JsonSerializerOptions.IncludeFields' is enabled. The match can be case-insensitive.

What if we make the property readonly, but settable through the json constructor? I believe that works as expected.

peterwald: > If you're happy with it as is it, that's fine, don't let my feedback stop you.

I do still need an approval from @dotnet/dotnet-extensions-ai, due to the change in the CacheHelper.

eiriktsarpalis: > > Types need to match with the parameter.
> 
> Only if the parameter is used for roundtripping, right? There's no mechanism to say "yes, I see there's a property with the same name as this parameter, but the types don't match and that property's setter should be used instead of the constructor"?

There's no such mechanism present today, unfortunately. We've received a lot of feedback about this kind of enforcement being needlessly strict but we just haven't gotten around to relaxing it or offering any degree of configurability.

PR #6074, opened 2025-03-11
Fix grouping of ChatResponseUpdate into ChatMessage

SteveSandersonMS:
For Ollama, it was broken because it set a new ResponseId for every chunk. It now correctly sets per-response IDs.

For OpenAI, it had a minor issue with handling empty-string ResponseId values. It now treats this case as equivalent to null ResponseId values, and thereby avoids introducing extra blank messages.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6074)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
@@ -171,7 +171,7 @@ public static class ChatResponseExtensions
         // response ID than the newest update, create a new message.
         ChatMessage message;
         if (response.Messages.Count == 0 ||
-            (update.ResponseId is string updateId && response.ResponseId is string responseId && updateId != responseId))
+            (update.ResponseId is { Length: > 0 } updateId && response.ResponseId is string responseId && updateId != responseId))
         {
             message = new ChatMessage(ChatRole.Assistant, []);
             response.Messages.Add(message);
@@ -213,7 +213,7 @@ public static class ChatResponseExtensions
         // Other members on a ChatResponseUpdate map to members of the ChatResponse.
         // Update the response object with those, preferring the values from later updates.
 
-        if (update.ResponseId is not null)
+        if (update.ResponseId is { Length: > 0 })
         {
             // Note that this must come after the message checks earlier, as they depend
             // on this value for change detection.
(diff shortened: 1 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -171,7 +171,7 @@ private static void ProcessUpdate(ChatResponseUpdate update, ChatResponse respon
         // response ID than the newest update, create a new message.
         ChatMessage message;
         if (response.Messages.Count == 0 ||
-            (update.ResponseId is string updateId && response.ResponseId is string responseId && updateId != responseId))
+            (update.ResponseId is { Length: > 0 } updateId && response.ResponseId is string responseId && updateId != responseId))
```
stephentoub, line 174: Ah, was the issue that it had an ID but it was an empty string?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
```diff
@@ -99,6 +99,12 @@ public IList<AIContent> Contents
     public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
 
     /// <summary>Gets or sets the ID of the response of which this update is a part.</summary>
+    /// <remarks>
+    /// This value is used when <see cref="ChatResponseExtensions.ToChatResponseAsync(IAsyncEnumerable{ChatResponseUpdate}, System.Threading.CancellationToken)"/>
+    /// groups <see cref="ChatResponseUpdate"/> instances into <see cref="ChatMessage"/> instances.
+    /// The value must be unique to each call to the underlying provider, and must be shared by
+    /// all updates that are part of the same response.
```
stephentoub, line 106: It's ok for some updates to have an empty ID, but at least one update needs a non empty one, and once an update is non empty, all further updates in the same message must have the same non empty ID. Right?

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
SteveSandersonMS: I agree that's true in practice, but I don't think we have to document that as the contract. We can state that the contract is simply that you must put a common ID on every update.

The fact that the OpenAI client is not completely conforming, and `ToChatResponseAsync` is lenient enough to make it work regardless, can be an implementation detail.

SteveSandersonMS: I'm going to resolve and merge this now in case we're short on time, but if you want this XML doc to be updated further in another update, let me know.

PR #6083, opened 2025-03-11
DRAFT: Adding support for an AI + .NET Aspire template

jmatthiesen:
Currently still in a draft state - This initial version creates a second "AI + .NET Aspire Chat App" template which can be manually installed and works with Ollama.

Still TODO:
* Add support for the other providers from the existing AI Chat Web App template
* Expand with additional vector stores
* Update README
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6083)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/AspireAIChat/.template.config/template.json
```diff
@@ -0,0 +1,349 @@
...
+  "$schema": "http://json.schemastore.org/template",
+  "author": "Microsoft",
+  "classifications": [ "Common", "AI", "Web", "Aspire" ],
+  "identity": "Microsoft.Extensions.AI.Templates.AspireChat.CSharp",
+  "name": "AI + .NET Aspire Chat  App",
+  "description": "A project template for creating an AI chat application using .NET Aspire, which uses retrieval-augmented generation (RAG) to chat with your own data.",
+  "shortName": "aiaspirechat",
+  "defaultName": "AspireChatApp",
+  "preferNameDirectory": false,
+  // The placeholder sourceName needs to contain a dash to ensure the CSS bundle asset uses the assembly identity name
```
jmatthiesen, line 11: Left this comment in for now until we know that we've addressed the issue raised here. I haven't tested it yet.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/AspireAIChat/.template.config/template.json
```diff
@@ -0,0 +1,349 @@
+{
```
jmatthiesen, line 1: Note: I'm going with a separate template, instead of this being an "Add Aspire Support" for the existing AI Chat Web App Template. This is because a) there's a different approach to configuration and project structure for Aspire support, b) we likely will have additional options for Aspire that you wouldn't see in a non-Aspire scenario, and c) it's a lot easier to merge 2 down to 1 later than to unravel a lot of #if conditional logic.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/AspireAIChat/AspireAIChat.Web/Components/Pages/Chat/Chat.razor.cs
```diff
@@ -0,0 +1,109 @@
+using AspireAIChat.Web.Services;
```
jmatthiesen, line 1: This is a bit of feature creep - this PR splits out code behind from the .razor files, so there are new .razor.cs files. I based this update off of some source sent to me, but there has been general feedback about best practices should be to separate the code behind from the Razor file. I agree with that approach and it does seem to work better in the Razor editor in VS Code if your C# code is in a .cs file.

MackinnonBuck:
Looks like the `.csproj` files didn't get committed, and that's because they're .gitignored. We'll want to add `.csproj.in` files that generate `.csproj` files at build time at some point. We'll also want this new template to participate in JS dependency update automation (I've set that up locally but can add it to this PR if necessary).

jmatthiesen:
Replaced by PR #6102

PR #6084, opened 2025-03-11
Bump Package Validation

joperezr:
cc: @RussKie
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6084)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,7 +6,7 @@
     <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
     <PreReleaseVersionIteration>1</PreReleaseVersionIteration>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <ApiCompatBaselineVersion>9.2.0</ApiCompatBaselineVersion>
+    <ApiCompatBaselineVersion>9.3.0</ApiCompatBaselineVersion>
     <ApiCompatBaselineSuffix>-preview.1.25105.6</ApiCompatBaselineSuffix>
     <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
     <!--
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -7,7 +7,6 @@
     <PreReleaseVersionIteration>1</PreReleaseVersionIteration>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <ApiCompatBaselineVersion>9.3.0</ApiCompatBaselineVersion>
-    <ApiCompatBaselineSuffix>-preview.1.25105.6</ApiCompatBaselineSuffix>
     <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
     <!--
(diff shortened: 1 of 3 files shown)


RussKie (approved):

eng/Versions.props
```diff
@@ -6,7 +6,7 @@
     <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
     <PreReleaseVersionIteration>1</PreReleaseVersionIteration>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <ApiCompatBaselineVersion>9.2.0</ApiCompatBaselineVersion>
+    <ApiCompatBaselineVersion>9.3.0</ApiCompatBaselineVersion>
     <ApiCompatBaselineSuffix>-preview.1.25105.6</ApiCompatBaselineSuffix>
```
RussKie, line 10: This needs to be updated as well

RussKie (changes requested):
ApiCompatBaselineSuffix needs to be updated as well

eng/Versions.props
joperezr: Good catch, that said, why do we even check api compat for preview packages, that seems wrong. The main point of api compat is ensure we don't break API that has shipped stable.

RussKie:
This needs to be removed as well:
https://github.com/dotnet/extensions/blob/64108d7d82a243f4ed8e1a983618a77c1294da1f/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj#L32

This was added to allow M.E.C.H to go stable.

/cc: @mgravell

eng/Versions.props
RussKie: I see it as a guide to package authors that there are breaking changes. A "preview" package doesn't necessarily means breaking changes could be made as will.

joperezr: We don't necessarily want to break the build in that case though, right? I guess my point is that the Build shouldn't stop devs from changing APIs in a preview package, that is not an unnatural thing to happen. For now I've disabled it for non-stable packages but let me know if you feel strongly about this and I'm happy to discuss.

RussKie: The example is M.A.Testing - this is shipped as "stable" though it is still in a "dev" stage.

The M.E.AI packages are opted out of the compat validation.
https://github.com/dotnet/extensions/blob/64108d7d82a243f4ed8e1a983618a77c1294da1f/src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.csproj#L12

RussKie (approved):

eng/Versions.props
joperezr: > The example is M.A.Testing - this is shipped as "stable" though it is still in a "dev" stage.

Good catch, we should just fix that by marking the API as stabled in that case.

> The M.E.AI packages are opted out of the compat validation.

This is precisely what I want to avoid. That line was likely there to just disable Baseline validation (validate against an old version of the package) but that line is also opting out for in-package validation (as-in, making sure that frameworks shipped within the package are still validated and compatible).

PR #6096, opened 2025-03-12
Use just-built packages in the AI chat template by default

MackinnonBuck:
This PR makes the following changes to the chat template:
1. By default, reference just-built MEAI package versions
2. Use the `Microsoft.EntityFrameworkCore.Sqlite` version defined in `Versions.props` instead of being hard-coded
3. Add new MSBuild properties that can be specified to utilize public versions of template dependencies instead of just-built versions
4. Generate development-only files that allow the templates to utilize locally produced NuGet packages
5. Adapt to the recent breaking changes in MEAI.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6096)


src/ProjectTemplates/GenerateTemplateContent/GenerateTemplateContent.csproj
```diff
@@ -1,9 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="$(MSBuildThisFileDirectory)..\GeneratedContent.props" />
-
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
RussKie, line 5: This won't be needed with #6041

RussKie (approved):

src/ProjectTemplates/GeneratedContent.targets
```diff
@@ -0,0 +1,72 @@
...
+
+      <!-- Specify package version variables used in template content. -->
+      <GeneratedContentProperties>
+        $(GeneratedContentProperties);
+
+        <!-- Repo properties -->
+        ArtifactsShippingPackagesDir=$(ArtifactsShippingPackagesDir);
+
+        <!-- Package version properties -->
+        OllamaSharpVersion=$(OllamaSharpVersion);
```
stephentoub, line 35: Where is this pulled from? Presumably it'll need to be updated to whatever version of OllamaSharp includes https://github.com/awaescher/OllamaSharp/pull/200

MackinnonBuck: Just updated to the latest version. Validating now that everything works.

src/ProjectTemplates/GenerateTemplateContent/GenerateTemplateContent.csproj
MackinnonBuck: Do you have a sense of when that PR will be merged? If that PR merges before this one, I'll remove this line.

RussKie: Once @joperezr approves it.

This is a non-blocking comment, if this PR goes in first, we just need to remember to undo this later.

eng/Versions.props
```diff
@@ -135,7 +135,7 @@
     <AzureSearchDocumentsVersion>11.6.0</AzureSearchDocumentsVersion>
     <MicrosoftSemanticKernelConnectorsAzureAISearchVersion>1.37.0-preview</MicrosoftSemanticKernelConnectorsAzureAISearchVersion>
     <MicrosoftSemanticKernelCoreVersion>1.37.0</MicrosoftSemanticKernelCoreVersion>
-    <OllamaSharpVersion>5.0.7</OllamaSharpVersion>
+    <OllamaSharpVersion>5.1.5</OllamaSharpVersion>
```
RussKie, line 138: Please note this will result in a conflict after https://github.com/dotnet/extensions/pull/6108

jeffhandley (commented):
@MackinnonBuck I merged from main, including the snapshot tests, and tried to work through getting everything green again. I've hit a couple of snags that we'll have to figure out though.

1. The snapshot is specific to whether we're using the pinned version or the just-built version, with the `NuGet.config` and `Directory.Build.Targets` files either included or excluded.
2. The snapshot bakes in the just-built version number, which is different between local development and CI.
3. Question: Is there something in place already where for the official build pipeline we would be able to toggle which mode to build with, or would we just update the targets file to toggle which is the default based on the next release's approach?

To get around the snapshot issues in CI and since we expect our next template release to use the MEAI and Sqlite packages released earlier this week, I swapped it to use the pinned versions by default, and I updated to the versions released this week. This got tests passing locally and in CI.

I hope you don't mind I pushed those changes on into your branch. Feel free to rewind the history and force-push if you want to discard my commits.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
```diff
@@ -67,23 +67,25 @@
...
+
+        // Store responses in the conversation, and begin getting suggestions
+        messages.AddMessages(response);
 #else*@
         // Stream and display a new response from the IChatClient
         var responseText = new TextContent("");
         currentResponseMessage = new ChatMessage(ChatRole.Assistant, [responseText]);
         currentResponseCancellation = new();
-        await foreach (var chunk in ChatClient.GetStreamingResponseAsync(messages, chatOptions, currentResponseCancellation.Token))
+        await foreach (var chunk in ChatClient.GetStreamingResponseAsync([.. messages], chatOptions, currentResponseCancellation.Token))
```
SteveSandersonMS, line 78: Can you clarify the need to clone `messages` here? Sorry if this seems fussy. It's just because this becomes code that every template user needs to deal with, so I want to be totally sure we haven't somehow made the usage experience harder with recent MEAI changes.

SteveSandersonMS: I understand the `messages` list might get mutated by L61 even after the call to `GetStreamingResponseAsync`, but with this being Blazor it won't happen concurrently across threads. So the only reason it would be a problem is if the chat client chooses to enumerate the messages asynchronously (and then the collection may be modified during the enumeration).

However I'm not sure it's legitimate for a chat client to enumerate the messages asynchronously since nothing in the API contract indicates that the collection can't be modified after the call is made. If the chat client does need to enumerate asynchronously, it seems like it should be responsibile for cloning that collection synchronously.

SteveSandersonMS: @stephentoub Do you consider it the responsibility of app code to pass in non-mutating message lists (e.g., by cloning before calling `GenerateAsync`), or do you consider it the responsibility of middleware to avoid yielding the thread during enumeration over the message list?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
```diff
@@ -108,6 +110,24 @@
...
+            {
+                case FunctionCallContent fcc:
+                    messages.Add(new ChatMessage(ChatRole.Assistant, [fcc]));
+                    break;
+                case FunctionResultContent frc:
+                    messages.Add(new ChatMessage(ChatRole.Tool, [frc]));
+                    break;
+            }
+        }
+    }
```
SteveSandersonMS, line 128: I see why this is necessary currently. Our existing `AddMessages`/`AddMessagesAsync` extension methods don't help because they only work on they complete set of output, whereas for an interactive UI app it's necessary to process the updates as they arrive incrementally.

I wonder if we could improve MEAI to make this nicer. Right now it's unfortunate that everyone would have to replicate this `AddNonTextContentToConversation` logic, especially given how it would need to be more complex still if they wanted to receive other content types or to retain other information from the updates, such as `AuthorName`, `AdditionalProperties`, `RawRepresentation`).

One option would be adding to MEAI's `ChatResponseExtensions` another extension method like this:

```cs
public static void AddMessages(this IList<ChatMessage> messages, ChatResponseUpdate update, Func<AIContent, bool>? filter = null)
{
    var contentsList = (filter is null ? update.Contents : update.Contents.Where(filter)).ToList();
    if (contentsList.Any())
    {
        messages.Add(new ChatMessage(update.Role ?? ChatRole.Assistant, contentsList)
        {
            AuthorName = update.AuthorName,
            RawRepresentation = update.RawRepresentation,
            AdditionalProperties = update.AdditionalProperties,
        });
    }
}
```

This is more general since it works with any content type and preserves the other metadata properties.

Then in application code, the call to `AddNonTextContentToConversation` would be replaced by a one-liner inside the `await foreach` above:

```cs
messages.AddMessages(chunk, filter: c => c is not TextContent);
```

(separately: we should rename `chunk` to `update` in this tempate code)

What do you think?

 * @MackinnonBuck If you think this is better, would you be interested in adding this extension method to MEAI as part of this change? It's quite nice that the template code can now directly use MEAI updates in the same PR!
 * @stephentoub Do you have any opinions on this proposed extension method? If you have concerns like "it's too specific to this one scenario" please say so. In my view it is pretty specific to this one scenario, but also, UI applications are a critical scenario we need to have clean patterns for.

src/ProjectTemplates/GeneratedContent.targets
SteveSandersonMS: Does the latest OllamaSharp allow is to eliminate the streaming/nonstreaming distinctions? As in, it works with streaming function calls now, doesn't it?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
MackinnonBuck: > So the only reason it would be a problem is if the chat client chooses to enumerate the messages asynchronously (and then the collection may be modified during the enumeration).

> However I'm not sure it's legitimate for a chat client to enumerate the messages asynchronously since nothing in the API contract indicates that the collection can't be modified after the call is made. If the chat client does need to enumerate asynchronously, it seems like it should be responsibile for cloning that collection synchronously.

Right, it was the asynchronous reading of `messages` by the chat client that I was primarily concerned about. I took a look at our own chat client implementations, and I see that the `FunctionInvokingChatClient` enumerates messages asynchronously [here](https://github.com/dotnet/extensions/blob/0e167021401fc051d056e8161dd8d1bb1ec3c11e/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L318), although that might just be a bug. Perhaps we should be cloning the originally passed messages [here](https://github.com/dotnet/extensions/blob/0e167021401fc051d056e8161dd8d1bb1ec3c11e/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L278).

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
MackinnonBuck: > One option would be adding to MEAI's `ChatResponseExtensions` another extension method like this...

That looks like it could work. Wouldn't it have the issue where if a message was split into multiple updates, this method would construct a separate message for each update? I'd imagine this is primarily an issue when the filter allows `TextContent`. The existing `ToChatResponse{Async}` extension methods handle this correctly by combining sequential messages with the same `ResponseId`.

Maybe we could have that method augment the last chat message in the list if it has the same `ResponseId` as the new `ChatResponseUpdate` (similar to the existing logic in the private `ProcessUpdate` helper).

Or maybe we add an `AddMessage(ChatResponse response, ChatResponseUpdate update)` extension method that just calls through to the existing logic in `ProcessUpdate` and augments the caller-provided `response`. So, rather than building your own `List<ChatMessage>` and converting it to a `ChatResponse` when the updates complete, you construct the `ChatResponse` yourself and augment it with each update.

MackinnonBuck:
@jeffhandley,

> 1. The snapshot is specific to whether we're using the pinned version or the just-built version, with the `NuGet.config` and `Directory.Build.Targets` files either included or excluded.

Ideally the snapshot would ignore files that don't get included in the actual `.nupkg`, but the tests just treat the source files in the repo as the template content and ignore what gets specified as `<Content>` in `Microsoft.Extensions.AI.Templates`. We could try to figure out an alternative that somehow takes the build output of `M.E.AI.Templates` into account.

> 2. The snapshot bakes in the just-built version number, which is different between local development and CI.

Yeah, also unfortunate. If we did my suggestion above, maybe the tests could build `M.E.AI.Templates` and specify MSBuild props that normalize that version number somehow (maybe even using a fake one).

> 3. Question: Is there something in place already where for the official build pipeline we would be able to toggle which mode to build with, or would we just update the targets file to toggle which is the default based on the next release's approach?

If the official build pipeline allows us to customize the build command on the fly, we could specify `/p:TemplateUsePinnedMicrosoftExtensionsAIVersion` or any of the other properties in `GeneratedContent.targets`. Otherwise we'd have to edit the `.targets` file manually. Maybe someone more familiar with that area of the build infrastructure would know better than I do. @RussKie?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
jeffhandley: I recommend we split this out into a separate issue, but I agree it's a change to MEAI we'd want to make in the next libraries release. I will file an issue for it and then we can put a TODO into the template that links to the issue.

src/ProjectTemplates/GeneratedContent.targets
jeffhandley: I tried to remove this special-cashing logic, and it was not working. Every response I got back had text of `""`.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Snapshots/aichatweb.Basic.verified/aichatweb/Components/Pages/Chat/Chat.razor
```diff
@@ -99,6 +100,22 @@
         await chatInput!.FocusAsync();
     }
 
+    private void AddNonTextContentToConversation(IList<AIContent> contents)
```
jeffhandley, line 103:
```suggestion

    // TODO: This helper is needed until https://github.com/dotnet/extensions/issues/6114 is resolved
    private void AddNonTextContentToConversation(IList<AIContent> contents)
```

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
stephentoub: > @stephentoub Do you consider it the responsibility of app code to pass in non-mutating message lists (e.g., by cloning before calling GenerateAsync), or do you consider it the responsibility of middleware/leaf-clients to avoid yielding the thread during enumeration over the message list?

In general if you pass an argument to an operation, whether sync or async, you're transferring ownership for the duration of the operation.

That said, it's hard to imagine what a valid IChatClient would look like that would need to read the input enumerable concurrently with the caller mutating it. It should only be doing work as part of the MoveNextAsync calls that a caller would be paused for. And if the caller is not immediately joining with MoveNextAsync and instead purposefully doing work concurrently with the MoveNextAsync, that's the caller's error. (It's more problematic if the IChatClient were mutating, as that could legitimately introduce deterministic problems.) 

So, I think we can just document the contract, that a valid IChatClient.GetStreamingResponseAsync implementation should only read the messages/options as part of the normal control flow of the async enumerable and not, for example, spawn a separate work item with Task.Run to do so during yield returns. We may not even need to do it. Doing something like that is so counterintuitive that we might cause more problems just by trying to describe the problem.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AichatwebTemplatesTests.cs
```diff
@@ -34,16 +33,28 @@ public async Task BasicTest()
...
-
         TemplateVerifierOptions options = new TemplateVerifierOptions(templateName: templateShortName)
         {
             TemplatePath = templateLocation,
             SnapshotsDirectory = "Snapshots",
             OutputDirectory = workingDir,
             DoNotPrependCallerMethodNameToScenarioName = true,
             ScenarioName = "Basic",
+
+            // Keep the exclude patterns below in sync with those in Microsoft.Extensions.AI.Templates.csproj.
```
jeffhandley, line 44: Thanks for the bidirectional comments for this dual-maintenance!

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AichatwebTemplatesTests.cs
```diff
@@ -34,16 +33,28 @@ public async Task BasicTest()
         // Get the template location
         string templateLocation = Path.Combine(TemplateFeedLocation, "Microsoft.Extensions.AI.Templates", "src", "ChatWithCustomData");
 
-        // Treat *.in files as text, see https://github.com/VerifyTests/EmptyFiles#istext
-        FileExtensions.AddTextExtension(".in");
```
jeffhandley, line 38: Excellent; great that this works. I think you're right that this will yield more intuitive snapshots with less risk of us leaking something into the actual template package that wasn't expected.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
SteveSandersonMS: OK great, that sounds straightforwards. So we can simplify the template code not to clone the message list, and don't need to make a rule that everyone must always clone message lists before calling `Get[Streaming]ResponseAsync`.

stephentoub: > and I see that the FunctionInvokingChatClient enumerates messages asynchronously [here](https://github.com/dotnet/extensions/blob/0e167021401fc051d056e8161dd8d1bb1ec3c11e/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L318)

Depends on the definition of asynchronous. It should be part of the caller's doing `await enumerator.MoveNextAsync`, so it shouldn't actually be executing in parallel with other mutations the caller might be making. If the caller does `ValueTask<bool> moveNext = enumerator.MoveNextAsync(); /* do mutations to the state over which the enumerator is operating */; await moveNext;`, then that could be an issue, but I think we can reasonably say that's a bad idea.

MackinnonBuck: > If the caller does `ValueTask<bool> moveNext = enumerator.MoveNextAsync(); /* do mutations to the state over which the enumerator is operating */; await moveNext;`, then that could be an issue, but I think we can reasonably say that's a bad idea.

I believe something similar could happen in this `Chat` component if, for example, the user sends a message as the chat client is streaming a response. This would call `AddUserMessageAsync()`, which immediately mutates the `messages` list, before the previous call to `GetStreamingResponseAsync()` completes. Would this break the contract and force us to do the cloning of the `messages` list as we do currently?

stephentoub: > This would call AddUserMessageAsync(), which immediately mutates the messages list, before the previous call to GetStreamingResponseAsync() completes

And that AddUserMessageAsync can happen in parallel with another logical thread of operation doing the `await foreach (...)`. That is a problem. Even if this all happens in a serialized fashion (which is not guaranteed to be the case), you'd still be mutating a List while there's an active enumerator for it, which will trip that enumerator's version checks and cause it to throw.

MackinnonBuck:
Updates:
* Snapshots now remove the `-ci` and `-dev` version suffixes
* Snapshots now ignore content that doesn't get included in the `M.E.AI.Templates` NuGet package
* The templates now utilize a new `AddMessages()` API

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
SteveSandersonMS: OK. So we definitely need to produce clear guidance around this and ensure it's robust.

We can simply clone the list in the code here, as was done before. However we do have the following mitigations:

 * These calls *are* guaranteed to happen in a serialized fashion, assuming by that you mean that there's only ever one active thread processing the objects in this file. This is a core part of the Blazor programming model.
 * The method starts by tripping the cancellation token that was passed into the previous call to `GetStreamingResponseAsync`. It's pretty reasonable to say that once cancellation is requested, the chat clients should stop what they are doing and not continue trying to enumerate the list.

Are there grounds for saying these mitigations aren't enough?

MackinnonBuck: Just tested the chat template without the cloning to see what would happen. After sending a single message, I got:
```text
GetStreamingResponseAsync failed.
      System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: )
      Parameter: messages.[4].role

      An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_80p6PFKOecFoo5BoUhoAkAGC
         at Azure.AI.OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
         at Azure.AI.OpenAI.Chat.AzureChatClient.CompleteChatAsync(BinaryContent content, RequestOptions options)
         at OpenAI.Chat.ChatClient.<>c__DisplayClass10_0.<<CompleteChatStreamingAsync>g__sendRequestAsync|0>d.MoveNext()
      --- End of stack trace from previous location ---
         at OpenAI.Chat.InternalAsyncStreamingChatCompletionUpdateCollection.GetRawPagesAsync()+MoveNext()
         at OpenAI.Chat.InternalAsyncStreamingChatCompletionUpdateCollection.GetRawPagesAsync()+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at System.ClientModel.AsyncCollectionResult`1.GetAsyncEnumerator(CancellationToken cancellationToken)+MoveNext()
         at System.ClientModel.AsyncCollectionResult`1.GetAsyncEnumerator(CancellationToken cancellationToken)+MoveNext()
         at System.ClientModel.AsyncCollectionResult`1.GetAsyncEnumerator(CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at Microsoft.Extensions.AI.OpenAIModelMappers.FromOpenAIStreamingChatCompletionAsync(IAsyncEnumerable`1 updates, CancellationToken cancellationToken)+MoveNext()
         at Microsoft.Extensions.AI.OpenAIModelMappers.FromOpenAIStreamingChatCompletionAsync(IAsyncEnumerable`1 updates, CancellationToken cancellationToken)+MoveNext()
         at Microsoft.Extensions.AI.OpenAIModelMappers.FromOpenAIStreamingChatCompletionAsync(IAsyncEnumerable`1 updates, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at Microsoft.Extensions.AI.LoggingChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
```

And I think this is due to the fact that the mutations to `messages` inside the `await foreach` loop are getting picked up again by the `FunctionInvokingChatClient`. In general, do we expect chat clients to be well-behaved when the message list gets mutated during a streaming response? This would basically require that chat clients do what @SteveSandersonMS suggested [previously](https://github.com/dotnet/extensions/pull/6096#discussion_r1995432001) (don't enumerate chat messages asynchronously, instead creating a clone). And if that's our stance, then `FunctionInvokingChatClient` needs to do such cloning [here](https://github.com/dotnet/extensions/blob/0e167021401fc051d056e8161dd8d1bb1ec3c11e/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L278), right?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
jeffhandley: @SteveSandersonMS We otherwise can get a template preview 1 refresh published **today** so that all the versions line up across all the MEAI* packages and without bumping the MEAI package itself again to reset that. @stephentoub and I would like to proceed with that refresh release without holding it up.

Effectively, it puts the helper into Templates Preview 1, and we'll remove it from Templates Preview 2. I'll make it as tidy as possible though.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData.Web-CSharp/Components/Pages/Chat/Chat.razor
stephentoub: > And I think this is due to the fact that the mutations to messages inside the await foreach loop are getting picked up again by the FunctionInvokingChatClient

This is a good point. If the caller might augment the list at an arbitrary point, that could easily mess up the requirements of middleware. In this case, the message must be getting added after a function call content but before the function result content.

Just to make sure I understand what you did Mackinnon, you sent a message that kicked off a GetStreamingResponseAsync, and then while that was still in flight, you added another message, yes?

I'm still really skeptical of encouraging patterns that mutate state passed to a method while that method is still using it. Saying that method needs to clone it feels bad to me. And are we talking shallow clone? If it's ok to mutate the top-level list, is it also ok to mutate the objects that are a part of that list, and if not, what's the distinction? If yes, well, now we're talking about components needing to fully serialize and deserialize the inputs, which is not something we should require.

I'd rather tell the caller it should not mutate the state that it's handed off to the caller. It's transferred ownership until the call completes. If it needs to update that state for subsequent processing, it should either store that state elsewhere, or it should pass a copy off whatever it might mutate so that it can transfer ownership of the copy.

That, to me, is the simplest and cleanest answer. For this case, that means either changing how the mutation is done or passing off a copy as was being done.

MackinnonBuck: > Just to make sure I understand what you did Mackinnon, you sent a message that kicked off a GetStreamingResponseAsync, and then while that was still in flight, you added another message, yes?

Yes, but just to be clear: The second message that got added was a `FunctionCallContent` that came from the chat client itself. I didn't have to send another user-submitted message while the response was being generated. The exception happens after the user sends a single chat message. I just had to change `[.. messages]` to `messages`.

> I'd rather tell the caller it should not mutate the state that it's handed off to the caller. It's transferred ownership until the call completes. If it needs to update that state for subsequent processing, it should either store that state elsewhere, or it should pass a copy off whatever it might mutate so that it can transfer ownership of the copy. You wouldn't tell a Stream.WriteAsync method that it needs to copy the input buffer to a temporary location because the caller might concurrently mutate it: the caller needs to not concurrently mutate it.

Agreed.

> For this case, that means either changing how the mutation is done or passing off a copy as was being done.

I guess the alternative is for the app to pass in `messages` directly instead of creating a copy, and have a separate list that stores the "in-progress" messages. The UI would render both message lists. When streaming completes, we flush the list of in-progress messages to the running message history. From a perf perspective, this might be slightly better because we don't have a cloning operation that gets increasingly expensive as the message history gets longer.

stephentoub: > The second message that got added was a FunctionCallContent that came from the chat client itself. I didn't have to send another user-submitted message while the response was being generated.

Ah, yes, that makes sense. FICC doesn't currently clone the list initially because there's a good chance no function calls will be made and it will have cloned it unnecessarily. It only needs to then clone it when it'll need to add to the list for the next turn of the crank. So after it's received all the updates, if it received any function call content, it then clones the list, adds the received content in, and iterates with the new list.

But if the caller is adding the received content into the list as the caller receives the content, then by the time the FICC goes to clone the content, it will have already been updated with the same content, such that the content will appear twice in the list, and the ones added by the caller will not have been paired with function result content, resulting in the failure.

This has nothing to do with concurrency.

Separate from what we decide to do for this particular mutation case, we probably should just have FunctionInvokingChatClient clone the state at the beginning, so as to avoid iterating the incoming list twice (once as part of initially sending the content and another as part of cloning it). Now that an enumerable is passed in, while it's most likely an in-memory collection, it could actually be a LINQ query or something doing arbitrary amounts of work, and we shouldn't be reiterating through that source multiple times. We can't know in advance whether a function call would require us to enumerate it again, so our best bet is to just enumerate it once to copy it and then never touch the source again.

SteveSandersonMS: Alright, so it seems that even after we bypass this problem in FICC (#6119) there remains an issue that any middleware or leaf client can legitimately `yield return` a `ChatMessageUpdate` while they have a pending iterator on the `messages` enumerable. This can return control flow to the caller, which in this case mutates the `message` list, and then boom regardless of concurrency or the CancellationToken. So it's wrong for the caller to mutate the supplied list inside the `await foreach`.

So cloning the list in `Chat.razor` actually necessary regardless of the fix to FICC, and hence @MackinnonBuck's code in this PR is ideal as-is.

If anyone disagrees I'd be interested to hear the details, but otherwise I think this PR could go in now!

SteveSandersonMS (approved):

src/ProjectTemplates/GenerateTemplateContent/GenerateTemplateContent.csproj
MackinnonBuck: Let's address this in a follow-up. I'm going to go ahead and merge this now.

RussKie:
> @MackinnonBuck I merged from main, including the snapshot tests, and tried to work through getting everything green again. I've hit a couple of snags that we'll have to figure out though.
> 
> 1. The snapshot is specific to whether we're using the pinned version or the just-built version, with the `NuGet.config` and `Directory.Build.Targets` files either included or excluded.
> 2. The snapshot bakes in the just-built version number, which is different between local development and CI.
> 3. Question: Is there something in place already where for the official build pipeline we would be able to toggle which mode to build with, or would we just update the targets file to toggle which is the default based on the next release's approach?
> 
> To get around the snapshot issues in CI and since we expect our next template release to use the MEAI and Sqlite packages released earlier this week, I swapped it to use the pinned versions by default, and I updated to the versions released this week. This got tests passing locally and in CI.
> 
> I hope you don't mind I pushed those changes on into your branch. Feel free to rewind the history and force-push if you want to discard my commits.



> Thanks for adding those updates, @jeffhandley!
> 
> > 1. The snapshot is specific to whether we're using the pinned version or the just-built version, with the `NuGet.config` and `Directory.Build.Targets` files either included or excluded.
> 
> Ideally the snapshot would ignore files that don't get included in the actual `.nupkg`, but the tests just treat the source files in the repo as the template content and ignore what gets specified as `<Content>` in `Microsoft.Extensions.AI.Templates`. We could try to figure out an alternative that somehow takes the build output of `M.E.AI.Templates` into account.
> 
> > 2. The snapshot bakes in the just-built version number, which is different between local development and CI.
> 
> Yeah, also unfortunate. If we did my suggestion above, maybe the tests could build `M.E.AI.Templates` and specify MSBuild props that normalize that version number somehow (maybe even using a fake one). Or, I think there's a way for the tests to do some post-processing on the output ([here](https://github.com/dotnet/extensions/blob/0e167021401fc051d056e8161dd8d1bb1ec3c11e/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AichatwebTemplatesTests.cs#L52-L62)); maybe we could do the normalization there. In fact, maybe the template testing infrastructure allows us to completely remove files from the snapshot (like `NuGet.config` and `Directory.Build.targets`), which would also solve 1.).
> 
> > 3. Question: Is there something in place already where for the official build pipeline we would be able to toggle which mode to build with, or would we just update the targets file to toggle which is the default based on the next release's approach?
> 
> If the official build pipeline allows us to customize the build command on the fly, we could specify `/p:TemplateUsePinnedMicrosoftExtensionsAIVersion` or any of the other properties in `GeneratedContent.targets`. Otherwise we'd have to edit the `.targets` file manually. Maybe someone more familiar with that area of the build infrastructure would know better than I do. @RussKie?

I think this discussion directly relates and is the cause of https://github.com/dotnet/extensions/issues/6128.

Can you help me understand why the versions are different between local development and CI - or more specifically between the public and the internal CI?

MackinnonBuck:
> Can you help me understand why the versions are different between local development and CI - or more specifically between the public and the internal CI?

@RussKie, the reason is because the version suffix when building locally is `-dev`, but when building in CI it's `-ci`. I had some changes in this PR that removed the suffix from the snapshots as an attempt to "normalize" the version number. But it looks the internal branches use yet another suffix (`-preview.x.x.x`).

We could probably update the test to also scrub versions matching that suffix as well. Or we could do something possibly more robust and find the exact `<PackageReference />`s for packages defined within the repo and completely scrub the `Version`.

Issue #6091, opened 2025-03-12
Add a HostedWebSearchTool


stephentoub:
For Google Gemini, this can map to its google_search_tool tool.

For OpenAI, this can map to its web_search tool.

Issue #6094, opened 2025-03-12
Remove from M.E.AI.OpenAI public surface area we don't expect to be made available


stephentoub:
The expectation at this point is the AsChatClient implementations will move into the target libraries. On our way to that, we should remove all the other public surface area; then it's just a matter of which library provides those AsChatClient methods (and the internal implementations that back them).

PR #6103, opened 2025-03-13
Add AsChatClient for OpenAIResponseClient

stephentoub:
Closes https://github.com/dotnet/extensions/issues/6090
Closes https://github.com/dotnet/extensions/issues/6091

cc: @SteveSandersonMS, @eiriktsarpalis, @trrwilson
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6103)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs
@@ -5,7 +5,7 @@ namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a hosted tool that can be specified to an AI service to enable it to execute code it generates.</summary>
 /// <remarks>
-/// This tool does not itself implement code interpration. It is a marker that can be used to inform a service
+/// This tool does not itself implement code interpretation. It is a marker that can be used to inform a service
 /// that the service is allowed to execute its generated code if the service is capable of doing so.
 /// </remarks>
 public class HostedCodeInterpreterTool : AITool
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs
@@ -3,15 +3,15 @@
 
 namespace Microsoft.Extensions.AI;
 
-/// <summary>Represents a tool that can be specified to an AI service to enable it to execute code it generates.</summary>
+/// <summary>Represents a hosted tool that can be specified to an AI service to enable it to execute code it generates.</summary>
 /// <remarks>
 /// This tool does not itself implement code interpration. It is a marker that can be used to inform a service
 /// that the service is allowed to execute its generated code if the service is capable of doing so.
 /// </remarks>
-public class CodeInterpreterTool : AITool
+public class HostedCodeInterpreterTool : AITool
 {
-    /// <summary>Initializes a new instance of the <see cref="CodeInterpreterTool"/> class.</summary>
-    public CodeInterpreterTool()
+    /// <summary>Initializes a new instance of the <see cref="HostedCodeInterpreterTool"/> class.</summary>
+    public HostedCodeInterpreterTool()
     {
     }
 }
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
@@ -633,6 +633,9 @@ internal static partial class OpenAIModelMappers
 
         [JsonPropertyName("properties")]
         public Dictionary<string, JsonElement> Properties { get; set; } = [];
+
+        [JsonPropertyName("additionalProperties")]
+        public bool AdditionalProperties { get; set; } = false;
     }
 
     /// <summary>POCO representing function calling info. Used to concatenation information for a single function call from across multiple streaming updates.</summary>
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Globalization;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
@@ -620,7 +619,7 @@ internal static partial class OpenAIModelMappers
     private static T? GetValueOrDefault<T>(this AdditionalPropertiesDictionary? dict, string key) =>
         dict?.TryGetValue(key, out T? value) is true ? value : default;
 
-    private static string CreateCompletionId() => $"chatcmpl-{Guid.NewGuid().ToString("N", CultureInfo.InvariantCulture)}";
+    private static string CreateCompletionId() => $"chatcmpl-{Guid.NewGuid():N}";
 
     /// <summary>Used to create the JSON payload for an OpenAI chat tool description.</summary>
     public sealed class OpenAIChatToolJson
@@ -635,7 +634,7 @@ internal static partial class OpenAIModelMappers
         public Dictionary<string, JsonElement> Properties { get; set; } = [];
 
         [JsonPropertyName("additionalProperties")]
-        public bool AdditionalProperties { get; set; } = false;
+        public bool AdditionalProperties { get; set; }
     }
 
     /// <summary>POCO representing function calling info. Used to concatenation information for a single function call from across multiple streaming updates.</summary>
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
@@ -1,0 +1,582 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+using OpenAI.Responses;
+
+#pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable SA1204 // Static elements should appear before instance elements
+#pragma warning disable SA1108 // Block statements should not contain embedded comments
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents an <see cref="IChatClient"/> for an <see cref="OpenAIResponseClient"/>.</summary>
+internal sealed class OpenAIResponseChatClient : IChatClient
+{
+    /// <summary>Gets the default OpenAI endpoint.</summary>
+    internal static Uri DefaultOpenAIEndpoint { get; } = new("https://api.openai.com/v1");
+
+    /// <summary>A <see cref="ChatRole"/> for "developer".</summary>
+    private static readonly ChatRole _chatRoleDeveloper = new("developer");
+
+    /// <summary>Cached <see cref="BinaryData"/> for the string "none".</summary>
+    private static readonly BinaryData _none = BinaryData.FromBytes("\"none\""u8.ToArray());
+
+    /// <summary>Cached <see cref="BinaryData"/> for the string "auto".</summary>
+    private static readonly BinaryData _auto = BinaryData.FromBytes("\"auto\""u8.ToArray());
+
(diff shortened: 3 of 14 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
```diff
@@ -633,6 +633,9 @@ public sealed class OpenAIChatToolJson
 
         [JsonPropertyName("properties")]
         public Dictionary<string, JsonElement> Properties { get; set; } = [];
+
+        [JsonPropertyName("additionalProperties")]
+        public bool AdditionalProperties { get; set; } = false;
```
RussKie, line 638: Isn't it false by default?

RussKie (approved):
✔️ from infra side

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIModelMapper.ChatCompletion.cs
stephentoub: Yes, the `= false;` can be removed

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs
```diff
@@ -3,15 +3,15 @@
 
 namespace Microsoft.Extensions.AI;
 
-/// <summary>Represents a tool that can be specified to an AI service to enable it to execute code it generates.</summary>
+/// <summary>Represents a hosted tool that can be specified to an AI service to enable it to execute code it generates.</summary>
 /// <remarks>
 /// This tool does not itself implement code interpration. It is a marker that can be used to inform a service
```
SteveSandersonMS, line 8:
```suggestion
/// This tool does not itself implement code interpretation. It is a marker that can be used to inform a service
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+
+#pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable SA1204 // Static elements should appear before instance elements
+#pragma warning disable SA1108 // Block statements should not contain embedded comments
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents an <see cref="IChatClient"/> for an <see cref="OpenAIResponseClient"/>.</summary>
+internal sealed class OpenAIResponseChatClient : IChatClient
```
SteveSandersonMS, line 23: Having this be `internal` makes sense. However `OpenAIChatClient` is `public` - should we update that to be `internal` or do you think there's a reason the two should differ? (Note: `OpenAIAssistantClient` is `internal`).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+        // Make the call to the OpenAIResponseClient.
+        var openAIResponse = (await _responseClient.CreateResponseAsync(openAIResponseItems, openAIOptions, cancellationToken).ConfigureAwait(false)).Value;
+
+        // Convert and return the results.
+        ChatResponse response = new()
+        {
+            ResponseId = openAIResponse.Id,
+            CreatedAt = openAIResponse.CreatedAt,
+            FinishReason = ToFinishReason(openAIResponse.IncompleteStatusDetails?.Reason),
+            Messages = [new(ChatRole.Assistant, (string?)null)],
```
SteveSandersonMS, line 109: Minor detail, and maybe you have some reason not to do this:

```suggestion
            Messages = [new(ChatRole.Assistant, [])],
```

I found the `(string?)null` approach a bit confusing as it seems like it's creating an empty `TextContent` which would be undesirable, but then I realised the `ChatMessage` constructor special-cases `null` and omits the content. It's a bit difficult to realise that's going to happen so may be clearer to pass `[]`, unless there's some other reason I didn't notice.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+                {
+                    case MessageResponseItem messageItem:
+                        message.RawRepresentation = messageItem;
+                        message.Role = ToChatRole(messageItem.Role);
+                        (message.AdditionalProperties ??= []).Add(nameof(messageItem.Id), messageItem.Id);
+                        message.Contents = ToAIContents(messageItem.Content);
+                        break;
+
+                    case FunctionCallResponseItem functionCall:
+                        response.FinishReason ??= ChatFinishReason.ToolCalls;
```
SteveSandersonMS, line 139: Out of interest, doesn't `ToFinishReason(openAIResponse.IncompleteStatusDetails?.Reason)` already give us the finish reason in the case of tool calls? What makes it necessary to set this finish reason explicitly instead of just keeping whatever finish reason was given (or none)?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+                        FunctionCallContent.CreateFromParsedArguments(
+                            fci.Arguments?.ToString() ?? string.Empty,
+                            fci.CallId ?? string.Empty,
+                            fci.Name!,
+                            static json => JsonSerializer.Deserialize(json, OpenAIJsonContext.Default.IDictionaryStringObject)!));
+                }
+            }
+        }
+
+        yield return update;
```
SteveSandersonMS, line 273: I expect this may be unavoidable, but it looks like it could change the order of function calls relative to text content within the response stream. All the text content will be yielded first, followed by all the function calls, even if they were interleaved in some way in the source data.

Is this because we can't know when a given function call is "complete" and could be yielded? Is there any possibility that a function call can be spread out and be interleaved with text contents or other function calls, or do we know that once we receive the first update for a given function call, all the consecutive ones will be for the same call until that call's data is fully consumed?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: > doesn't ToFinishReason(openAIResponse.IncompleteStatusDetails?.Reason) already give us the finish reason in the case of tool calls?

No, there's no tool calls reason. It's just null.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+                        jsonFormat.SchemaDescription) :
+                    ResponseTextFormat.CreateJsonObjectFormat();
+            }
+        }
+
+        return result;
+    }
+
+    /// <summary>Convert a sequence of <see cref="ChatMessage"/>s to <see cref="ResponseItem"/>s.</summary>
+    private static IEnumerable<ResponseItem> ToOpenAIResponseItems(
```
SteveSandersonMS, line 421: The naming of this API creates some incredibly confusing code :) The "response" items are actually the "request" content in this case, because it's a request to the "response" API...

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+                {
+                    switch (tool)
+                    {
+                        case AIFunction af:
+                            var oaitool = JsonSerializer.Deserialize(af.JsonSchema, OpenAIJsonContext.Default.OpenAIChatToolJson)!;
+                            var functionParameters = BinaryData.FromBytes(JsonSerializer.SerializeToUtf8Bytes(oaitool, OpenAIJsonContext.Default.OpenAIChatToolJson));
+                            result.Tools.Add(ResponseTool.CreateFunctionTool(af.Name, af.Description, functionParameters));
+                            break;
+
+                        case HostedWebSearchTool:
```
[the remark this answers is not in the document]
stephentoub: It doesn't currently support it. I believe it's coming.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: Yup. @trrwilson has been trying to figure out how to make that better, but the underlying naming makes it challenging.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+
+                            yield return ResponseItem.CreateFunctionCallOutputItem(resultContent.CallId, result ?? string.Empty);
+                            break;
+                    }
+                }
+
+                continue;
+            }
+
+            if (input.Role == ChatRole.Assistant)
```
SteveSandersonMS, line 474: What's the situation where the outbound data includes "assistant" messages, given the server statefulness?

I think this probably relates to the fact that even with server statefulness, you can freely prepopulate it with any conversation history you like. Is that right?

If so I see that we're only supporting `TextContent` and `FunctionCallContent` in an outbound "Assistant" message. Longer term should we expect to generalize this so that, for example, you can preseed the conversation with some `DataContent` that an assistant might previously have returned? Or perhaps more likely, citations, URLs, etc.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -0,0 +1,582 @@
...
+    private static List<AIContent> ToAIContents(IEnumerable<ResponseContentPart> contents)
+    {
+        List<AIContent> results = [];
+
+        foreach (ResponseContentPart part in contents)
+        {
+            if (part.Kind == ResponseContentPartKind.OutputText)
+            {
+                results.Add(new TextContent(part.Text));
+            }
```
SteveSandersonMS, line 538: Is the API inherently text-only right now, or are multimodal cases handled differently?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: > What's the situation where the outbound data includes "assistant" messages, given the server statefulness?

You don't have to use the server history. If you don't specify a previous response id, it behaves just like chat completion. And if you do, it's so just concatenating the provided messages into whatever it already has (which, if there's no previous ID, it's nothing). So you could have part of a conversion stored on the server and send another chunk of it.

> If so I see that we're only supporting TextContent and FunctionCallContent in an outbound "Assistant" message.

That's all the underlying API supports right now.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
```diff
@@ -0,0 +1,296 @@
...
+
+            event: response.completed
+            data: {"type":"response.completed","response":{"id":"resp_67d329fbc87c81919f8952fe71dafc96029dabe3ee19bb77","object":"response","created_at":1741892091,"status":"completed","error":null,"incomplete_details":null,"instructions":null,"max_output_tokens":20,"model":"gpt-4o-mini-2024-07-18","output":[{"type":"message","id":"msg_67d329fc0c0081919696b8ab36713a41029dabe3ee19bb77","status":"completed","role":"assistant","content":[{"type":"output_text","text":"Hello! How can I assist you today?","annotations":[]}]}],"parallel_tool_calls":true,"previous_response_id":null,"reasoning":{"effort":null,"generate_summary":null},"store":true,"temperature":0.5,"text":{"format":{"type":"text"}},"tool_choice":"auto","tools":[],"top_p":1.0,"truncation":"auto","usage":{"input_tokens":26,"input_tokens_details":{"cached_tokens":0},"output_tokens":10,"output_tokens_details":{"reasoning_tokens":0},"total_tokens":36},"user":null,"metadata":{}}}
+
```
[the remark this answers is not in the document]
stephentoub: There's a bug in the underlying APIs right now that makes deserializing function information fail. We can assignment the tests once that's fixed.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: Yup. I didn't want to do it as part of this PR, though. It's tracked by https://github.com/dotnet/extensions/issues/6094.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: > Is this because we can't know when a given function call is "complete" and could be yielded?

Yes. @trrwilson, is there any way through the API to determine when we get the last part of a function call? There wasn't with chat completion, and I didn't see anything obvious for responses, either.

PR #6105, opened 2025-03-14
Add assignments for @dotnet/dotnet-extensions-ai-templates

RussKie:


RussKie:
If/when there will be other (i.e., non AI) templates, then we can revisit the pattern.

jeffhandley (approved):
Thanks!

PR #6111, opened 2025-03-14
[Telemetry, Logging] Emit {OriginalFormat} as the last property of the log record

### LINKED ISSUE #5720: [ExtendedLogger] {OriginalFormat} is incorrectly exported by OpenTelemetry
### Description

If a logger emits the `{OriginalFormat}` property OpenTelemetry handles it in a special way (see https://github.com/open-telemetry/opentelemetry-dotnet/pull/4334, Changes section):

Image

Our [`ExtendedLogger`](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.cs) does emit the `{OriginalFormat}` property, but it is not always handled by OpenTelemetry as it is described above. For example, regardless of the value of the `IncludeFormattedMessage` property OpenTelemetry always exports **source generated logs** produced by `ExtendedLogger` as formatted messages, i.e. it replaces the arguments in `{OriginalFormat}` and exports the resulting value as the `LogRecord.Body` property.

The correct behavior must be the following:

* If `IncludeFormattedMessage=true` then `LogRecord.Body` must contain the formatted message.
* If `IncludeFormattedMessage=false` then `LogRecord.Body` must contain the message format, i.e. the value of the `{OriginalFormat}` property.

This bug affects how log records are exported to the telemetry backend.

The bug is caused by the fact that `ExtendedLogger` doesn't ensure that the `{OriginalFormat}` property is the last in the collection of log properties, which is required by OpenTelemetry. Hence, the fix would be to ensure that the property is always returned as the last one.

We need to fix the bug and ensure that the `{OriginaFormat}` property is handled correctly in the following use cases:

* source-generated logs
* regular logs emitted by methods like `LogInformation`
* the above two cases must be validated when enrichment, static enrichment, and redaction, are enabled and used, because these features could affect the order of the `{OriginalFormat}` property in the collection of log properties

#### Project file:

```xml
  <PropertyGroup>
    <LangVersion>13</LangVersion>
    <Nullable>enable</Nullable>
    <OutputType>Exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.1" />
    <PackageReference Include="Microsoft.Extensions.Telemetry" Version="8.10.0" />
    <PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.8.1" />
  </ItemGroup>
</Project>
```

#### C# code:

```csharp
using OpenTelemetry.Logs;

internal static class Program
{
    public static void Main()
    {
        using var loggerFactory = LoggerFactory.Create(builder =>
        {
            builder
                .EnableEnrichment()
                .AddOpenTelemetry(loggerOptions =>
                {
                    loggerOptions.IncludeFormattedMessage = false;
                    loggerOptions.AddConsoleExporter();
                });
        });

        var logger = loggerFactory.CreateLogger("Logger");

        logger.TestMessage("Hello", "world");
        logger.LogInformation("LogInformation: {arg1}, {arg2}", "Hello", "world");
    }
}

{
    [LoggerMessage(0, LogLevel.Information, "LoggerMessage: {arg1}, {arg2}")]
    public static partial void TestMessage(this ILogger logger, string arg1, string arg2);
}
```

### Expected behavior

Both log statements must emit a `LogRecord` that has the message format in the `LogRecord.Body` property, i.e.

```output
LogRecord.Body = "LoggerMessage: {arg1}, {arg2}"
```

### Actual behavior

When running the sample you'll get the following result:

Image

Note that the first message produced by the source generated log statement has the formatted message in the `LogRecord.Body` property, while the second message produced by the `LogInformation` method has the message format in the `LogRecord.Body`.

The second message has the correct value in the `LogRecord.Body` because the `LogInformation` method ensures that the `{OriginalFormat}` property is the last in the collection. But if you enable enrichment then the second log statement will also have incorrect `LogRecord.Body` value.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

iliar-turdushev:
Updates the source-generated logging and the `ExtendedLogger` to emit the `{OriginalFormat}` property as the last item of the log record.

Fixes #5720

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6111)


src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -318,7 +319,14 @@ void GenTagWrites(LoggingMethod lm, string stateName, out int numReservedUnclass
             {
                 OutLn();
                 OutLn($"_ = {stateName}.ReserveTagSpace({numReservedUnclassifiedTags});");
+
                 int count = numReservedUnclassifiedTags;
+
+                if (!string.IsNullOrEmpty(lm.Message))
```
iliar-turdushev, line 325: We now emit `{OriginalFormat}` as the last item of `TagArray`.

src/Generators/Microsoft.Gen.Logging/Emission/Emitter.Method.cs
```diff
@@ -318,7 +319,14 @@ void GenTagWrites(LoggingMethod lm, string stateName, out int numReservedUnclass
...
+                {
+                    OutLn($"{stateName}.TagArray[{--count}] = new(\"{{OriginalFormat}}\", {EscapeMessageString(lm.Message)});");
```
iliar-turdushev, line 327: We now emit `{OriginalFormat}` as the last item of `TagArray`.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.ModernTagJoiner.cs
```diff
@@ -56,21 +56,23 @@ public void SetIncomingTags(LoggerMessageState value)
         {
             get
             {
-                if (index < _incomingTagsCount)
+                // Keep the order of the if-else statements. It ensures that the {OriginalFormat}
```
evgenyfedorov2, line 59: would smth like below bring more clarity
```suggestion
                // Iterating over _incomingTags at the end, because it contains {OriginalFormat} property which needs to be the last.
```

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.LegacyTagJoiner.cs
```diff
@@ -50,17 +50,19 @@ public void SetIncomingTags(IReadOnlyList<KeyValuePair<string, object?>> value)
...
-                if (index < _incomingTagCount)
+                // Keep the order of the if-else statements. It ensures that the {OriginalFormat}
+                // property is always the last tag in the list.
+                if (index < StaticTags!.Length)
                 {
-                    return _incomingTags![index];
+                    return StaticTags[index];
                 }
-                else if (index < _incomingTagCount + _extraTags.Count)
+                else if (index < StaticTags.Length + _extraTags.Count)
```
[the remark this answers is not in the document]
iliar-turdushev: Thank you, done as you suggested.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.ModernTagJoiner.cs
```diff
@@ -56,21 +56,23 @@ public void SetIncomingTags(LoggerMessageState value)
...
+                // property is always the last tag in the list.
+                if (index < _redactedTagsCount)
                 {
-                    return _incomingTags![index];
+                    return _redactedTags![index];
                 }
-                else if (index < _incomingTagsCount + _redactedTagsCount)
+                else if (index < _redactedTagsCount + _extraTags.Count)
```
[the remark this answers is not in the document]
iliar-turdushev: Thank you, done as you suggested.

src/Libraries/Microsoft.Extensions.Telemetry/Logging/ExtendedLogger.ModernTagJoiner.cs
iliar-turdushev: Thank you, updated the comments.

evgenyfedorov2 (approved):

PR #6120, opened 2025-03-15
Simplify chat message streaming in chat template

MackinnonBuck:
This PR is a proposal for improved handling of handle chat message streaming in the chat template.

Recently, the `IChatClient` interface was updated to discourage chat client implementations from manipulating the provided chat history. This created two problems for the chat template:
1. The chat template was relying on messages containing `FunctionInvocationContent` being automatically added to the chat history by `FunctionInvokingChatClient`. This no longer happens, so the chat template now has to manually augment its chat history with those messages while reading the streaming response. Discussion [here](https://github.com/dotnet/extensions/pull/6096#discussion_r1995496400).
2. However, chat clients may not expect the caller to manipulate the message list while producing a streaming response. This requires the chat template to clone the message list before passing it to the chat client. Discussion [here](https://github.com/dotnet/extensions/pull/6096#discussion_r1995415962).

Problem 2 goes away if the chat template stores in-progress messages separately from the "committed" chat history. It just renders two message lists in the UI: the in-progress list and the committed list.

Problem 1 then goes away if you change the in-progress list to store `ChatResponseUpdate`s directly rather than mapping them to `ChatMessage`s. When streaming ends, the second list gets added to the first list in the form of chat messages. Separate UI logic can decide how to display content from `ChatResponseUpdate`s. This is the approach that this PR demonstrates.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6120)


SteveSandersonMS:
Thanks @MackinnonBuck!

This is certainly an improvement to clarity in some respects, however if I'm understanding it correctly, it also appears to come at a significant perf cost. Perhaps it's possible to resolve that while retaining some of the clarity improvements?

The previous approach used an admittedly unusual pub-sub mechanism so that, for each streaming chunk, only the single `ChatMessageItem` displaying that chunk would re-render. Everything else was left alone. In the approach here, every incoming chunk causes the whole existing conversation to re-render in full, producing an O(N^2) effect on the number of renders.

I experimented by starting with the message "what's in the kit" and then clicking the first suggestion until I produced a conversation with 10 user messages. I think this is a realistic length of conversation. In total this causes ~2000 chunks to arrive through the whole conversation.

 * With the previous approach, `ChatMessageItem` rendered 1284 times
 * With the proposed approach, `ChatMessageItem` rendered 25,696 times (and this grows quadratically as the conversation grows)

Maybe it would be possible to get some of the benefits of the new approach without the drawbacks by inlining `ChatAssistantContentItem` back into `ChatMessageItem` (so that there's no need for differentiating `Content` and `Text` parameters there) and retaining the pub-sub signalling from `Chat` to `ChatMessageItem`, but retaining the new mechanism of tracking `currentResponseUpdates` separately from the committed history.

SteveSandersonMS:
To be honest, even the previous approach's 1284 renders (for a 10-user-message conversation) is pushing at the boundaries. We may want to at least think through what approach we'd use if we were determined to minimize the expense of this. For example:

 * We could have the `<assistant-message>` part be an independent component and be the subscriber in the pub-sub system, so we literally only re-render that single element per-chunk, hence the rendering and diffing overhead drops to roughly zero and the whole cost is more or less the same as just sending a websocket message from server to client on each chunk
 * Or, we could bypass Blazor rendering entirely on a per-chunk basis, and have it use JS interop to trigger an update to the text inside the `<assistant-message>`.

MackinnonBuck:
Thanks for the insight, @SteveSandersonMS!

> This is certainly an improvement to clarity in some respects, however if I'm understanding it correctly, it also appears to come at a significant perf cost. Perhaps it's possible to resolve that while retaining some of the clarity improvements?

> The previous approach used an admittedly unusual pub-sub mechanism so that, for each streaming chunk, only the single `ChatMessageItem` displaying that chunk would re-render. Everything else was left alone. But in the approach in this PR, every incoming chunk causes the whole existing conversation to re-render in full, producing an O(N^2) effect on the number of renders.

Yeah, this was something that crossed my mind when making these changes, but it was getting a little late by the time I put out the PR and didn't get around to thinking about it further. Perhaps I should have called that out. I appreciate that you took some measurements to quantify the perf difference, and I agree this should be addressed before we take this change (if we decide to do so).

The JS interop approach sounds interesting - I might try out something like that, and if it turns out to be too sophisticated for template code, I'll add back the pub/sub mechanism that existed previously. However, I probably won't get to that for a couple days while addressing other high-priority items.

jeffhandley:
@MackinnonBuck / @jmatthiesen  - I am closing this PR as we're not actively pursuing the feedback about the performance trade-offs. We will keep this on the backlog for consideration later though.

PR #6125, opened 2025-03-16
Add AIJsonSchemaCreateOptions.IncludeParameter filter

stephentoub:
The CreateFunctionJsonSchema method includes all parameters in the schema. If someone wants to exclude a parameter, e.g. because the LLM shouldn't recognize that parameter and the value for the argument will instead come from somewhere else when the function is invoked, there's no good way to achieve that while still using this helper.

This PR adds an IncludeParameter filter to the create options, such that a developer can supply a delegate invoked for each ParameterInfo to declare whether that parameter should be included or not.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6125)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -2,7 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Reflection;
 using System.Text.Json.Nodes;
+using System.Threading;
 
 #pragma warning disable S1067 // Expressions should not be too complex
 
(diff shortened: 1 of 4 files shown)


stephentoub:
This is mostly extracted from the previous DI PR I'd closed, as supporting parameters filled from DI necessitates then not showing up in the schema. I wasn't planning on adding the DI support for now, but after this goes in, I think we may actually want to. It makes it cleaner to reuse the same AIFunction parameterized with local state, as that state can be fetched from DI with a scoped lifetime.

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -23,6 +25,18 @@ public sealed class AIJsonSchemaCreateOptions : IEquatable<AIJsonSchemaCreateOpt
...
     public Func<AIJsonSchemaCreateContext, JsonNode, JsonNode>? TransformSchemaNode { get; init; }
 
+    /// <summary>
+    /// Gets a callback that is invoked for every parameter in the <see cref="MethodBase"/> provided to
+    /// <see cref="AIJsonUtilities.CreateFunctionJsonSchema"/> in order to determine whether it should
+    /// be included in the generated schema.
+    /// </summary>
+    /// <remarks>
+    /// By default, when <see cref="IncludeParameter"/> is <see langword="null"/>, all parameters other
+    /// than those of type <see cref="CancellationToken"/> are included in the generated schema.
```
eiriktsarpalis, line 35: Nit: I thought the point of this delegate was to avoid hardcoding types getting filtered at the AIFunctionFactory layer. It can be reasonably argued that `CancellationToken` should always be filtered, but what if we want to exclude more parameter types in the future? (i.e. something like the context type we removed recently, which wouldn't be visible in this assembly)

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
eiriktsarpalis: Right. My point is that this creates two sources of truth about what types do get filtered.

Issue #5965, opened 2025-02-24
GetAssistantClient issue with vision/images


SteveSandersonMS:
I've been trying to use `GetAssistantClient` to work with an Open AI assistant and ran into difficulties. I couldn't make images work (well I could but it involved some unreasonable hackery and private reflection).

### Use of vision/images with the assistants endpoint appears to be different than with the chat endpoint.

Our implementation tries to set an imageUrl using a data URI, but in realistic cases this will fail with "URI too long". I don't know if it's even supported if you have a tiny image for which the URI wouldn't be too long.

Instead, for assistants, you have to use OpenAI's "file" client to upload the image using a `FileUploadPurpose.Assistants`, and then reference the image in a later message using `MessageContent.FromImageFileId(...)`. There isn't an MEAI abstraction corresponding to the file client, and it doesn't populate `ImageFileId` on anything we send to OpenAI, so it doesn't seem possible to achieve.

Sidenote: as far as I can tell, even if we did use the APIs correctly, it would only work on OpenAI Platform. Right now assistants on Azure OpenAI doesn't appear to support `ImageFileId` - they just ignore whatever value you give, even if you did successfully upload an image and got back that ID.

stephentoub:
> Our implementation tries to set an imageUrl using a data URI, but in realistic cases this will fail with "URI too long".

Yes, this is a known limitation of Uri (cc: @MihaZupan).
https://github.com/dotnet/runtime/issues/96544
@MihaZupan, @karelz, would it be possible to prioritize fixing this for .NET 10? I realize we're unlikely to backport such a change to previous core versions, and certainly not to netfx, but just enabling it to work moving forward would be a really good step. This is a big challenge in the AI space especially, where images and audio are now very often embedded into data uris.

stephentoub:
OpenAI has stated they'll be deprecating Assistants in favor of Responses. Given that, as part of https://github.com/dotnet/extensions/issues/6094 I suggest we just delete this AsChatClient for AssistantClient.

Issue #5964, opened 2025-02-24
GetAssistantClient issue with "thread has active run"


SteveSandersonMS:
I've been trying to use `GetAssistantClient` to work with an Open AI assistant and ran into difficulties. I'm often getting this error:

```
Thread <threadID> already has an active run <runID>.
```

This is based on the typical chat loop code that works with `IChatClient`, except instead of building up an `IList<ChatMessage>` I'm just sending a single message at a time and preserving the thread ID:

```cs
var response = await chatClient.GetResponseAsync(new ChatMessage(ChatRole.User, text), chatOptions, cancellationToken: _prevCallCts.Token);
chatOptions.ChatThreadId = response.ChatThreadId;
```

The error occurs if the user sends a new message while an old one was still in flight. It's not realistic to stop a user from doing that because sometimes the assistant takes a long time (20s, say). For usability there has to be a way to abort an existing run and replace it with another.

I was hoping that if the `CancellationToken` fires, it would abort the old run and queue up any subsequent runs behind that abort operation. But it doesn't appear to, hence the error. I'm not sure if there's even a way to abort it manually since the API doesn't expose the run ID.

stephentoub:
> I was hoping that if the CancellationToken fires, it would abort the old run and queue up any subsequent runs behind that abort operation. But it doesn't appear to, hence the error. I'm not sure if there's even a way to abort it manually since the API doesn't expose the run ID.

We could certain add that in, though it's a bit unusual to actually initiate new I/O in response to someone asking to stop all I/O.

Another options would be to have a built-in retry, where we special-case errors about a run already being in-flight, cancel it, and then try again one more time.

SteveSandersonMS:
> Another options would be to have a built-in retry, where we special-case errors about a run already being in-flight, cancel it, and then try again one more time.

That's a decent approach, since from the code I get the impression we don't even know the runID of the previous call until it starts producing output. So we can't necessarily cancel it by ID.

Detecting the error and cancelling+retrying avoids both the issue you mention and this one above.

It's possible there could be multiple attempts to cancel a previous run and so there could be some kind of race condition determining who takes over as the new "current" run. Ideally we'd guarantee that last-call-wins.

stephentoub:
OpenAI has stated they'll be deprecating Assistants in favor of Responses. Given that, as part of https://github.com/dotnet/extensions/issues/6094 I suggest we just delete this AsChatClient for AssistantClient.

PR #6138, opened 2025-03-17
Remove non-AsXx surface area from M.E.AI.OpenAI/AzureAIInference

stephentoub:
We expect the AsChatClient/AsEmbeddingGenerator extension methods from M.E.AI.OpenAI and M.E.AI.AzureAIInference to move into the OpenAI / Azure.AI.Inference libraries, respectively. To prepare for that, and the temporary M.E.AI.OpenAI/AzureAIInference libs then being deprecated, this PR removes the other surface area from these assemblies. For anything folks found useful, we should find another way to ship it, likely just as sample source somewhere.

Closes https://github.com/dotnet/extensions/issues/6094
Closes https://github.com/dotnet/extensions/issues/5965
Closes https://github.com/dotnet/extensions/issues/5964

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6138)


src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -321,7 +311,7 @@ private ChatCompletionsOptions ToAzureAIOptions(IEnumerable<ChatMessage> chatCon
                         default:
                             if (prop.Value is not null)
                             {
-                                byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, ToolCallJsonSerializerOptions.GetTypeInfo(typeof(object)));
+                                byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(object)));
```
eiriktsarpalis, line 314: I'd forgotten about `AdditionalProperties`. Hardcoding to the default options effectively forces known types only in AOT.

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
stephentoub: I suspect primitives is sufficient. We can always add an overload later that takes a JSO if it proves important. Or you think it should be done now?

eiriktsarpalis: I suspect there always will be a need for storing something less trivial than just primitives.

stephentoub: Quite possibly, but this in particular is about properties that should be sent up to the service as extra properties that a given service understands:
https://learn.microsoft.com/en-us/rest/api/aifoundry/modelinference/#extensibility
and those are generally going to be primitives, like float, int, or string.

(Also, the Azure.AI.Inference library itself is not yet AOT friendly.)

jeffhandley (approved):

SteveSandersonMS:
This is certainly all good stuff in general.

Within the `GetService` implementations, should we also remove the clauses that return `this` if  `serviceType.IsInstanceOfType(this)`? I'm not sure what use case those clauses would serve now you can't just reference the type directly and can only get it in obscure ways (e.g., by being the second-to-last middleware and passing `innerClient.GetType()`.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -1,140 +0,0 @@
...
-using System.Threading.Tasks;
-using Microsoft.Shared.Diagnostics;
-using OpenAI.RealtimeConversation;
-
-namespace Microsoft.Extensions.AI;
-
-/// <summary>
-/// Provides extension methods for working with <see cref="RealtimeConversationSession"/> and related types.
-/// </summary>
-public static class OpenAIRealtimeExtensions
```
SteveSandersonMS, line 19: Losing this is a bit unfortunate (for me anyway, I'm certainly using it from various projects and from https://github.com/SteveSandersonMS/dotnet-ai-workshop). 

Isn't there's still legitimacy in the provider-specific packages including helpers for making more of the provider-specific functionality work with MEAI, even if it's not purely an `IChatClient`?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
eiriktsarpalis: What about `float[]`, `int?`, `int?[]`, or `string[]`? If we're drawing a line in the sand we probably need to document it (and quite possibly reject values whose type is not supported at runtime). Otherwise we find ourselves in a situation where clients that work without issue in CoreCLR end up failing in Native AOT without a workaround.

stephentoub: ? What about float[], int?, int?[], or string[]?

`int?` boxes the same as `int`, but regardless it's already represented in the default options.

If we think `float[]`, `int?[]`, or `string[]` will be desirable, we can just add them to AIJsonUtilities.

stephentoub:
This should be ready to go.

One important note:
I've made two additional changes:
- Removed the AsChatClient/AsEmbeddingGenerator extensions for the OpenAIClient type, leaving just the ones for OpenAI.Chat.ChatClient and OpenAI.Embeddings.EmbeddingClient. Given what we've seen with OpenAI's move to push folks away from ChatClient and towards ResponseClient, it seems unwise for us to have AsChatClient on OpenAIClient that's bound to the chat completion implementation.
- Renamed AsChatClient to AsIChatClient and AsEmbeddingGenerator to AsIEmbeddingGenerator. This is to avoid confusion with naming, e.g. having an OpenAI `ChatClient` have an `AsChatClient` extension method on it.

cc: @jeffhandley, @SteveSandersonMS

shyamnamboodiripad (approved):
Approving for the single change in the eval test 👍🏾

jeffhandley (approved):

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
```diff
@@ -134,7 +129,6 @@ public async Task BasicRequestResponse_NonStreaming()
               "tool_choice": "auto",
               "tools": [],
               "top_p": 1.0,
-              "truncation": "auto",
```
SteveSandersonMS, line 137: Out of interest, why remove this? Is it something OpenAI has changed?

stephentoub: Yes, it was removed with the latest rev of responses.

stephentoub:
@SteveSandersonMS, @jeffhandley, per my discussion with Steve yesterday, I temporarily added back the AsChatClient(OpenAIClient) and AsEmbeddingGenerator(OpenAIClient) extensions, but as EBNever and Obsolete.

jeffhandley (approved):

PR #6141, opened 2025-03-18
Enable AIFunctionFactory.Create functions to get DI services

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6141)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Shared.Diagnostics;
@@ -63,11 +64,13 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
     /// </summary>
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
     /// <param name="logger">An <see cref="ILogger"/> to use for logging information about function invocation.</param>
-    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null)
+    /// <param name="services">An optional <see cref="IServiceProvider"/> to use for resolving services required by the <see cref="AIFunction"/> instances being invoked.</param>
+    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null, IServiceProvider? services = null)
         : base(innerClient)
     {
-        _logger = logger ?? NullLogger.Instance;
+        _logger = logger ?? (ILogger?)services?.GetService<ILogger<FunctionInvokingChatClient>>() ?? NullLogger.Instance;
         _activitySource = innerClient.GetService<ActivitySource>();
+        Services = services;
     }
 
     /// <summary>
@@ -81,6 +84,9 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
         get => _currentContext.Value;
         protected set => _currentContext.Value = value;
     }
+
+    /// <summary>Gets the <see cref="IServiceProvider"/> used for resolving services required by the <see cref="AIFunction"/> instances being invoked.</summary>
+    public IServiceProvider? Services { get; }
 
     /// <summary>
     /// Gets or sets a value indicating whether to handle exceptions that occur during function calls.
@@ -721,8 +727,14 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
         object? result = null;
         try
         {
+            IEnumerable<KeyValuePair<string, object?>>? arguments = context.CallContent.Arguments;
+            if (Services is not null)
+            {
+                arguments = new AIFunctionArguments(arguments) { Services = Services };
+            }
+
             CurrentContext = context;
-            result = await context.Function.InvokeAsync(context.CallContent.Arguments, cancellationToken).ConfigureAwait(false);
+            result = await context.Function.InvokeAsync(arguments, cancellationToken).ConfigureAwait(false);
         }
         catch (Exception e)
         {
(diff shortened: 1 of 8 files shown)


src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -248,6 +253,30 @@ public static ReflectionAIFunctionDescriptor GetOrCreate(MethodInfo method, AIFu
 
         private ReflectionAIFunctionDescriptor(DescriptorKey key, JsonSerializerOptions serializerOptions)
         {
+            AIJsonSchemaCreateOptions schemaOptions = new()
+            {
+                // This needs to be kept in sync with the shape of AIJsonSchemaCreateOptions.
```
stephentoub, line 258: Should AIJsonSchemaCreateOptions expose a Clone method? Or be a record if that won't introduce other issues?

src/Libraries/Microsoft.Extensions.AI/Functions/FromServicesAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+
+/// <summary>Indicates that a parameter to an <see cref="AIFunction"/> should be sourced from an associated <see cref="IServiceProvider"/>.</summary>
+/// <remarks>
+/// <see cref="AIFunctionFactory"/> uses this attribute to determine whether a parameter's value should be sourced from
+/// an <see cref="IServiceProvider"/> instead of from the nominal arguments passed to the function. The <see cref="IServiceProvider"/>
+/// is extracted from the <see cref="AIFunctionArguments.Services"/> property of an <see cref="AIFunctionArguments"/> passed
+/// as the arguments to a call to <see cref="AIFunction.InvokeAsync"/>.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Parameter)]
+public sealed class FromServicesAttribute : Attribute
```
stephentoub, line 16: I know there's some hesitancy about duplicating the same name MVC uses. I don't have a better idea, though.

SteveSandersonMS: Do we have further reasons to think it wouldn't suffice to support passing in an `IServiceProvider`, and having the function resolve services from there?

eiriktsarpalis: Would be worth starting an effort lowering all those attributes to the M.E.DI package? We could then take a dependency on the next preview package.

stephentoub: Alternate approaches:
https://github.com/dotnet/extensions/pull/6144
https://github.com/dotnet/extensions/pull/6145
https://github.com/dotnet/extensions/pull/6146

stephentoub: > Would be worth starting an effort lowering all those attributes to the M.E.DI package? We could then take a dependency on the next preview package.

Not for a GA release. We wouldn't be able to add that until November. And FromServices is in an ASP.NET MVC namespace.

halter73: I left a comment on #6146 explaining why I like that proposal the best. Is there any major pushback to making the `IServiceProvider` a top-level parameter like `CancellationToken` to `AIFunction.InvokeAsync`?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
halter73: Given that `AIJsonSchemaCreateOptions` has custom implementations for `Equals` and `GetHashCode` that seem to mirror what records would give us, I think that makes sense. Especially now if we're cloning in some places.

If we don't go the record route, it would probably be best to extract this clone into an internal testable method, so you could write a reflection-based test like [this](https://github.com/dotnet/aspnetcore/blob/18b3b4ac8dfc4498edc6bf26c7bde9a35ef840f8/src/Servers/Kestrel/Core/test/SniOptionsSelectorTests.cs#L807) to verify no one accidentally adds a property and forgets to update the clone method.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -63,11 +64,13 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
     /// </summary>
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
     /// <param name="logger">An <see cref="ILogger"/> to use for logging information about function invocation.</param>
-    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null)
+    /// <param name="services">An optional <see cref="IServiceProvider"/> to use for resolving services required by the <see cref="AIFunction"/> instances being invoked.</param>
+    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null, IServiceProvider? services = null)
```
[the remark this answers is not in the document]
stephentoub: > Will it be confusing to pass an IServiceProvider into the constructor and not have it get used at all by FunctionInvokingChatClient.GetService(Type serviceType, object? serviceKey = null)?

GetService has the same shape as with IServiceProvider, but it's not consulting any IServiceProvider in any of the implementations we have or have seen thus far. It's just a mechanism to ask the components in the pipeline for a particular object, e.g. if you have a pipeline of IChatClients that ends in an OnnxRuntimeGenAIClient, and for some reason you want to ask it for its Tokenizer, you can do so with GetService, and that call will be passed down through the pipeline until OnnxRuntimeGenAIClient says "yeah, I've got a Tokenizer, here". That also means it doesn't really work as a property.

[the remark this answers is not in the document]
stephentoub: > I think it might be more consistent to rely on the IServiceProvider implementation from the innerClient

Most inner clients aren't going to have one. e.g. the OpenAIChatClient doesn't have an IServiceProvider, the AzureAIInferenceChatClient doesn't have one, etc. In fact I don't think any implementations I've seen do. Some other middleware components might, but there's no guarantee FunctionInvocationChatClient will be wrapping one of those. I can't think of a better way than to store the IServiceProvider that's pass in from the builder, typically in UseFunctionInvocation().

PR #6148, opened 2025-03-18
Use only a single TFM for M.E.AI.Eval.Reporting.Console

peterwald:
Restrict MTFM builds of M.E.AI.Eval.Reporting.Console to only the minimum version of .NET Core. This is to prevent build breaks such as [this](https://github.com/dotnet/extensions/pull/6133#issuecomment-2731431023).

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Description>A dotnet tool for managing the evaluation data and generating reports.</Description>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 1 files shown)


stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Description>A dotnet tool for managing the evaluation data and generating reports.</Description>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(MinimumSupportedTfmForPackaging)</TargetFrameworks>
```
shyamnamboodiripad, line 6: It may be a good idea to leave a comment here linking the SDK bug and saying that we can revert back to building all TFMs once it is fixed.

shyamnamboodiripad (approved):

Issue #6139, opened 2025-03-17
Add support for Azure storage management to the `aieval` tool.


peterwald:
- [ ] Support generating reports directly from Azure storage.
- [ ] Support cleanup actions on Azure storage.

PR #6152, opened 2025-03-18
Extend M.E.AI.Evaluation.Console to support Azure Storage options

peterwald:
Specifying the `--endpoint` option alternatively to `--path` to allow reporting and management from Azure storage rather than disk. 

Closes https://github.com/dotnet/extensions/issues/6139

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6152)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
@@ -37,8 +37,7 @@ internal sealed class CleanCacheCommand(ILogger logger)
         }
         else
         {
-            logger.LogError("Either --path or --endpoint must be specified");
-            return 1;
+            throw new InvalidOperationException("Either --path or --endpoint must be specified");
         }
 
         await logger.ExecuteWithCatchAsync(
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
@@ -1,10 +1,14 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Identity;
+using Azure.Storage.Files.DataLake;
 using Microsoft.Extensions.AI.Evaluation.Console.Utilities;
+using Microsoft.Extensions.AI.Evaluation.Reporting;
 using Microsoft.Extensions.AI.Evaluation.Reporting.Storage;
 using Microsoft.Extensions.Logging;
 
@@ -12,13 +16,30 @@ namespace Microsoft.Extensions.AI.Evaluation.Console.Commands;
 
 internal sealed class CleanCacheCommand(ILogger logger)
 {
-    internal async Task<int> InvokeAsync(DirectoryInfo storageRootDir, CancellationToken cancellationToken = default)
+    internal async Task<int> InvokeAsync(DirectoryInfo? storageRootDir, Uri? endpointUri, CancellationToken cancellationToken = default)
     {
-        string storageRootPath = storageRootDir.FullName;
-        logger.LogInformation("Storage root path: {storageRootPath}", storageRootPath);
-        logger.LogInformation("Deleting expired cache entries...");
+        IResponseCacheProvider cacheProvider;
 
-        var cacheProvider = new DiskBasedResponseCacheProvider(storageRootPath);
+        if (storageRootDir is not null)
+        {
+            string storageRootPath = storageRootDir.FullName;
+            logger.LogInformation("Storage root path: {storageRootPath}", storageRootPath);
+            logger.LogInformation("Deleting expired cache entries...");
+
+            cacheProvider = new DiskBasedResponseCacheProvider(storageRootPath);
+        }
+        else if (endpointUri is not null)
+        {
+            logger.LogInformation("Azure Storage endpoint: {endpointUri}", endpointUri);
+
+            var fsClient = new DataLakeDirectoryClient(endpointUri, new DefaultAzureCredential());
+            cacheProvider = new AzureStorageResponseCacheProvider(fsClient);
+        }
+        else
+        {
+            logger.LogError("Either --path or --endpoint must be specified");
+            return 1;
+        }
 
         await logger.ExecuteWithCatchAsync(
             () => cacheProvider.DeleteExpiredCacheEntriesAsync(cancellationToken)).ConfigureAwait(false);
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
@@ -41,8 +41,7 @@ internal sealed class CleanResultsCommand(ILogger logger)
         }
         else
         {
-            logger.LogError("Either --path or --endpoint must be specified");
-            return 1;
+            throw new InvalidOperationException("Either --path or --endpoint must be specified");
         }
 
         await logger.ExecuteWithCatchAsync(
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
@@ -1,11 +1,15 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Identity;
+using Azure.Storage.Files.DataLake;
 using Microsoft.Extensions.AI.Evaluation.Console.Utilities;
+using Microsoft.Extensions.AI.Evaluation.Reporting;
 using Microsoft.Extensions.AI.Evaluation.Reporting.Storage;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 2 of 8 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
```diff
@@ -24,13 +24,16 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Azure.Storage.Files.DataLake" />
```
shyamnamboodiripad, line 28: Does the data lake package reference not flow from the Reporting.Azure project reference below?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Program.cs
```diff
@@ -33,19 +35,44 @@ private static async Task<int> Main(string[] args)
...
+                "Endpoint URL under which the cache and results are stored for Azure Data Lake Gen2 storage")
+            {
+                IsRequired = false
+            };
+
+        var openReportOpt =
+            new Option<bool>(
+                ["--open"],
+                getDefaultValue: () => true,
+                "Open the report in the default browser")
```
shyamnamboodiripad, line 60: Nice! 👍🏾

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Program.cs
```diff
@@ -105,5 +139,16 @@ private static async Task<int> Main(string[] args)
 #endif
 
         return await rootCmd.InvokeAsync(args).ConfigureAwait(false);
+
+        void requiresPathOrEndpoint(CommandResult cmd)
```
shyamnamboodiripad, line 143: minor: Can we change to `RequiresPathOrEndpoint()`? Iirc nested functions we have created in other places start with capitals. I was a bit confused at the use site since I expected the lambda variable to be declared before use but then realized it is a nested function :)

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
peterwald: We directly construct a DataLakeFileSystemClient, so we need the reference.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Program.cs
```diff
@@ -105,5 +139,16 @@ private static async Task<int> Main(string[] args)
...
+        {
+            bool hasPath = cmd.GetValueForOption(pathOpt) is not null;
+            bool hasEndpoint = cmd.GetValueForOption(endpointOpt) is not null;
+            if (!(hasPath ^ hasEndpoint))
```
shyamnamboodiripad, line 147: Consider changing to -

```suggestion
            bool hasPath = string.IsNullOrWhiteSpace(cmd.GetValueForOption(pathOpt));
            bool hasEndpoint = string.IsNullOrWhiteSpace(cmd.GetValueForOption(endpointOpt));
            if (!(hasPath ^ hasEndpoint))
```

peterwald: Ah, they are not string values. They are strongly typed. `DirectoryInfo` and `Uri` respectively.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
```diff
@@ -17,17 +20,37 @@ namespace Microsoft.Extensions.AI.Evaluation.Console.Commands;
...
+            string storageRootPath = storageRootDir.FullName;
+            logger.LogInformation("Storage root path: {storageRootPath}", storageRootPath);
+
+            resultStore = new DiskBasedResultStore(storageRootPath);
+        }
+        else if (endpointUri is not null)
+        {
+            logger.LogInformation("Azure Storage endpoint: {endpointUri}", endpointUri);
+
+            var fsClient = new DataLakeDirectoryClient(endpointUri, new DefaultAzureCredential());
```
shyamnamboodiripad, line 44: Wondering if this may be insufficient like we saw with the doc team's use case where they also had to supply Tennant Id... Not much we can do for now though - we can revisit if we hear feedback.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
```diff
@@ -17,17 +20,37 @@ namespace Microsoft.Extensions.AI.Evaluation.Console.Commands;
...
+            resultStore = new AzureStorageResultStore(fsClient);
+        }
+        else
+        {
+            logger.LogError("Either --path or --endpoint must be specified");
```
shyamnamboodiripad, line 49: Shouldn't this be caught by the validator? If yes, it may be better to throw an exception (InvalidOperationException) since we don't expect this block to be reachable.

peterwald:
Closes #6139

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
```diff
@@ -38,6 +61,8 @@ internal async Task<int> InvokeAsync(
                     cancellationToken: cancellationToken).ConfigureAwait(false))
             {
                 results.Add(result);
+
+                logger.LogInformation("Execution: {executionName} Scenario: {scenarioName} Iteration: {iterationName}", result.ExecutionName, result.ScenarioName, result.IterationName);
```
shyamnamboodiripad, line 65: minor: This could end up being quite verbose when there are 1000s of scenarios and iterations. Should we limit this logging to verbose mode (and introduce a flag for verbosity)?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
```diff
@@ -58,6 +83,24 @@ internal async Task<int> InvokeAsync(
         await reportWriter.WriteReportAsync(results, cancellationToken).ConfigureAwait(false);
         logger.LogInformation("Report: {outputFilePath} [{format}]", outputFilePath, format);
 
+        // See the following issues for reasoning behind this check. We want to avoid opening the report
+        // if this process is running as a service or in a CI pipeline.
+        // https://github.com/dotnet/runtime/issues/770#issuecomment-564700467
+        // https://github.com/dotnet/runtime/issues/66530#issuecomment-1065854289
+        bool isRedirected = System.Console.IsInputRedirected && System.Console.IsOutputRedirected && System.Console.IsErrorRedirected;
+        bool isInteractive = Environment.UserInteractive && (OperatingSystem.IsWindows() || !(isRedirected));
```
shyamnamboodiripad, line 91: Does this mean we won't open the browser if someone is running evals on linux / macOS?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Program.cs
```diff
@@ -105,5 +139,16 @@ private static async Task<int> Main(string[] args)
...
+            {
+                cmd.ErrorMessage = $"Either '{pathOpt.Name}' or '{endpointOpt.Name}' must be specified.";
```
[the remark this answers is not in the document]
peterwald: I don't think so. The command line will return an error with this message to the console when it fails. Having the message print out twice would just be confusing.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
peterwald: Fair enough. Updated.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
shyamnamboodiripad: Wondering if it would be better to keep the option off by default. Adding a `-open` / `-view` when running locally does not seem like it would be too cumbersome.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanCacheCommand.cs
```diff
@@ -1,24 +1,45 @@
...
+        else if (endpointUri is not null)
+        {
+            logger.LogInformation("Azure Storage endpoint: {endpointUri}", endpointUri);
+
+            var fsClient = new DataLakeDirectoryClient(endpointUri, new DefaultAzureCredential());
+            cacheProvider = new AzureStorageResponseCacheProvider(fsClient);
+        }
+        else
+        {
+            logger.LogError("Either --path or --endpoint must be specified");
```
shyamnamboodiripad: Same comment about throwing instead.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/CleanResultsCommand.cs
```diff
@@ -14,14 +18,32 @@ namespace Microsoft.Extensions.AI.Evaluation.Console.Commands;
...
+        else if (endpointUri is not null)
+        {
+            logger.LogInformation("Azure Storage endpoint: {endpointUri}", endpointUri);
+
+            var fsClient = new DataLakeDirectoryClient(endpointUri, new DefaultAzureCredential());
+            resultStore = new AzureStorageResultStore(fsClient);
+        }
+        else
+        {
+            logger.LogError("Either --path or --endpoint must be specified");
```
shyamnamboodiripad: Same comment about throwing instead.

shyamnamboodiripad: Same comment about throwing instead.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
peterwald: It doesn't seem that bad to me. I added it because there is no other indication of progress when generating the report and it can take more than a few seconds.

I really would like to replace the logger. Maybe it's just me, but the output is fairly unreadable I think. Would like to have something more like the new dotnet build logging, that can update interactively. I'm open to updates here, but I feel like this is at least an improvement for now.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
peterwald: I haven't tested it, but this logic is supposed to be cross-platform. So it should open a browser on linux or mac. Maybe I can give it a try later.

shyamnamboodiripad (approved):
Looks great modulo some (mostly minor) suggestions!

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
shyamnamboodiripad: Sounds good - interactive updates that don't scroll the output would be great for the future.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
shyamnamboodiripad: I lean towards keeping it simple and always requiring explicit opt in - however, I don't feel too strongly about this :)

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Program.cs
shyamnamboodiripad: Ah my bad 😄

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Microsoft.Extensions.AI.Evaluation.Console.csproj
shyamnamboodiripad: Interesting, my assumption was that we won't need to add it explicitly here since we didn't say `PrivateAssets` in the `PackageReference` in the referenced MEAI.E.Reporting.Azure project...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Console/Commands/ReportCommand.cs
shyamnamboodiripad: Looks like this particular command may have been missed - but the instances in the other commands have been fixed to throw.

RussKie:
> Closes #6139

This should be in the top message (and ideally in the commit message itself).

RussKie (approved):

PR #6158, opened 2025-03-19
Change AIFunction.InvokeAsync to accept AIFunctionArguments

stephentoub:
- Adds a new AIFunctionArguments type.
- Changes AIFunction.InvokeAsync to accept an AIFunctionArguments instead of an arbitrary enumerable.
- Changes FunctionInvokingChatClient to accept an IServiceProvider and expose it as a Services property, and to then pass that IServiceProvider into the AIFunction via AIFunctionArguments.Services.
- Augments FunctionInvocationContext with an AIFunctionArguments property.
- Changes AIFunctionFactory to special-case parameters of type IServiceProvider and AIFunctionArguments, sourcing from AIFunctionArguments.
- Makes AIJsonSchemaCreateOptions a record.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6158)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
@@ -1,0 +1,119 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+#pragma warning disable SA1111 // Closing parenthesis should be on line of last parameter
+#pragma warning disable SA1112 // Closing parenthesis should be on line of opening parenthesis
+#pragma warning disable SA1114 // Parameter list should follow declaration
+#pragma warning disable CA1710 // Identifiers should have correct suffix
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents arguments to be used with <see cref="AIFunction.InvokeAsync"/>.</summary>
+/// <remarks>
+/// <see cref="AIFunctionArguments"/> is a dictionary of name/value pairs that are used
+/// as inputs to an <see cref="AIFunction"/>. However, an instance carries additional non-nominal
+/// information, such as an optional <see cref="IServiceProvider"/> that can be used by
+/// an <see cref="AIFunction"/> if it needs to resolve any services from a dependency injection
+/// container.
+/// </remarks>
+public sealed class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
+{
+    /// <summary>The nominal arguments.</summary>
+    private readonly Dictionary<string, object?> _arguments;
+
+    /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
+    public AIFunctionArguments()
+    {
+        _arguments = [];
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AIFunctionArguments"/> class containing
+    /// the specified <paramref name="arguments"/>.
+    /// </summary>
+    /// <param name="arguments">The arguments represented by this instance.</param>
+    /// <remarks>
+    /// The <paramref name="arguments"/> reference will be stored if the instance is
+    /// already a <see cref="Dictionary{TKey, TValue}"/>, in which case all dictionary
+    /// operations on this instance will be routed directly to that instance. If <paramref name="arguments"/>
+    /// is not a dictionary, a shallow clone of its data will be used to populate this
+    /// instance. A <see langword="null"/> <paramref name="arguments"/> is treated as an
+    /// empty dictionary.
+    /// </remarks>
+    public AIFunctionArguments(IDictionary<string, object?>? arguments)
+    {
+        _arguments =
+            arguments is null ? [] :
+            arguments as Dictionary<string, object?> ??
+            new Dictionary<string, object?>(arguments);
+    }
+
+    /// <summary>Gets or sets services optionally associated with these arguments.</summary>
+    public IServiceProvider? Services { get; set; }
+
+    /// <inheritdoc />
+    public object? this[string key]
+    {
+        get => _arguments[key];
+        set => _arguments[key] = value;
+    }
+
+    /// <inheritdoc />
+    public ICollection<string> Keys => _arguments.Keys;
+
+    /// <inheritdoc />
+    public ICollection<object?> Values => _arguments.Values;
+
+    /// <inheritdoc />
+    public int Count => _arguments.Count;
+
+    /// <inheritdoc />
+    bool ICollection<KeyValuePair<string, object?>>.IsReadOnly => false;
+
+    /// <inheritdoc />
+    IEnumerable<string> IReadOnlyDictionary<string, object?>.Keys => Keys;
+
+    /// <inheritdoc />
+    IEnumerable<object?> IReadOnlyDictionary<string, object?>.Values => Values;
+
+    /// <inheritdoc />
+    public void Add(string key, object? value) => _arguments.Add(key, value);
+
+    /// <inheritdoc />
+    void ICollection<KeyValuePair<string, object?>>.Add(KeyValuePair<string, object?> item) =>
+        ((ICollection<KeyValuePair<string, object?>>)_arguments).Add(item);
+
+    /// <inheritdoc />
+    public void Clear() => _arguments.Clear();
+
+    /// <inheritdoc />
+    bool ICollection<KeyValuePair<string, object?>>.Contains(KeyValuePair<string, object?> item) =>
+        ((ICollection<KeyValuePair<string, object?>>)_arguments).Contains(item);
+
+    /// <inheritdoc />
+    public bool ContainsKey(string key) => _arguments.ContainsKey(key);
+
+    /// <inheritdoc />
+    public void CopyTo(KeyValuePair<string, object?>[] array, int arrayIndex) =>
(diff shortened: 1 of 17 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -0,0 +1,119 @@
...
+    public AIFunctionArguments(IDictionary<string, object?>? arguments)
+    {
+        _arguments =
+            arguments is null ? [] :
+            arguments as Dictionary<string, object?> ??
+            new Dictionary<string, object?>(arguments);
+    }
+
+    /// <summary>Gets or sets services optionally associated with these arguments.</summary>
+    public IServiceProvider? Services { get; set; }
```
eiriktsarpalis, line 56: I take it abstractions now has taken a dependency on DI?

stephentoub: No. IServiceProvider is in System.ComponentModel.

eiriktsarpalis (approved):

stephentoub:
@SteveSandersonMS, @halter73, as this is a fundamental change, I'd like a second review before merging. Thanks!

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -248,9 +245,36 @@ public static ReflectionAIFunctionDescriptor GetOrCreate(MethodInfo method, AIFu
 
         private ReflectionAIFunctionDescriptor(DescriptorKey key, JsonSerializerOptions serializerOptions)
         {
+            // Augment the schema options to exclude AIFunctionArguments/IServiceProvider from the schema,
+            // as it'll be satisfied from AIFunctionArguments.
+            static bool IncludeNonAIFunctionArgumentParameter(ParameterInfo parameterInfo) =>
+                parameterInfo.ParameterType != typeof(AIFunctionArguments) &&
```
eiriktsarpalis, line 251: Curious if we should consider using `IServiceProviderIsService` to filter out other parameter types, similar to how AspNet does it.

eiriktsarpalis: This would need to be passed into the factory somehow, although I'm not sure there's a clean way to do that.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
SteveSandersonMS, line 56: Is there any case for this being `set` instead of `init`?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
```diff
@@ -107,7 +107,7 @@ public static async Task HandleToolCallsAsync(
 
             try
             {
-                var result = await aiFunction.InvokeAsync(functionCallContent.Arguments, cancellationToken).ConfigureAwait(false);
+                var result = await aiFunction.InvokeAsync(new(functionCallContent.Arguments), cancellationToken).ConfigureAwait(false);
```
SteveSandersonMS, line 110: Could we pass in an `IServiceProvider` here too? I know it would have to be threaded through from `HandleToolCallsAsync` but otherwise it might seem like a missing capability.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -82,6 +84,9 @@ public static FunctionInvocationContext? CurrentContext
         protected set => _currentContext.Value = value;
     }
 
+    /// <summary>Gets the <see cref="IServiceProvider"/> used for resolving services required by the <see cref="AIFunction"/> instances being invoked.</summary>
+    public IServiceProvider? Services { get; }
```
SteveSandersonMS, line 88: I'm not certain why this is public API. Could you clarify?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
stephentoub: The whole object is mutable. Not clear to me why this property would be different. But if you think it should be, I can change it.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: It doesn't have to be. I could make it private if you'd prefer.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
SteveSandersonMS: The intent with the current design is that we don't automatically supply parameters from the `IServiceProvider` based on the type, but rather we only supply the `IServiceProvider` itself.

Not certain if that addresses your question though!

SteveSandersonMS, line 251: I understand why we're filtering out `IServiceProvider` here, but why also filter out `AIFunctionArguments`? Is it possible for an AI function to receive the `AIFunctionArguments` directly, and if so what's the use case?

SteveSandersonMS (approved):
Posted some minor questions but the overall thing look great.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
SteveSandersonMS: I'd default to private if we have the choice.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
SteveSandersonMS: OK that's fair

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
stephentoub: > Is it possible for an AI function to receive the AIFunctionArguments directly

This PR makes that possible

> and if so what's the use case?

I'm not sure there's a strong use case, other than it provides a means for the AIFunction to enumerate arbitrary arguments passed in by a caller and pull out the things it wants dynamically rather than declaring them all statically. But the reason I included it wasn't because of a strong use case, but because of consistency: InvokeAsync accepts an AIFunctionArguments and a CancellationToken, with the AIFunctionArguments exposing an IServiceProvider. It's inconsistent to me that two of those three are special-cased; this makes it so that all three are. And there doesn't seem to be a strong downside to it, as there's nothing meaningful you could do by having an AIFunctionArguments in your schema.

SteveSandersonMS: OK thanks. I have no problem with it. Just wanted to check if there was something missing in my understanding.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -62,12 +62,14 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
     /// Initializes a new instance of the <see cref="FunctionInvokingChatClient"/> class.
     /// </summary>
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
-    /// <param name="logger">An <see cref="ILogger"/> to use for logging information about function invocation.</param>
-    public FunctionInvokingChatClient(IChatClient innerClient, ILogger? logger = null)
+    /// <param name="loggerFactory">An <see cref="ILoggerFactory"/> to use for logging information about function invocation.</param>
+    /// <param name="services">An optional <see cref="IServiceProvider"/> to use for resolving services required by the <see cref="AIFunction"/> instances being invoked.</param>
+    public FunctionInvokingChatClient(IChatClient innerClient, ILoggerFactory? loggerFactory = null, IServiceProvider? services = null)
```
halter73, line 67: I'm still a little worried that people will misunderstand the purpose IServiceProvider passed into the constructor and think it will get used by `FunctionInvokingChatClient.GetService(Type serviceType, object? serviceKey = null)`.

I suggested merging these concepts [here](https://github.com/dotnet/extensions/pull/6141#discussion_r2002047651), but I understand that might not be feasible given all of existing inner IChatClients don't expose arbitrary services that could be supplied by the host. Furthermore, the way IChatClients can wrap each other to add services wouldn't play very nicely with IServiceProviderIsService in the event we ever want to utilize that.

Maybe renaming the parameter will help avoid some of this confusion.

```suggestion
    public FunctionInvokingChatClient(IChatClient innerClient, ILoggerFactory? loggerFactory = null, IServiceProvider? functionInvocationServices = null)
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIRealtimeExtensions.cs
halter73: I agree this feels like a gap. Is this and FunctionInvokingChatClient the only two places we call AIFunction.InvokeAsync internally?

Issue #5941, opened 2025-02-19
[API Proposal]: IChatClient - Default ChatOptions


azchohfi:
### Background and motivation

Different IChatClient implementations might have different defaults for ChatOptions, which can be sent to CompleteAsync/CompleteStreamingAsync. Right now, there is no default way to retrieve such defaults. Yes, those methods allow you to send null to those parameters, which would use their default values for ChatOptions, but there is no way to retrieve that information in a generic way.
The only way to achieve this today is to use the GetService method, maybe sending a specific serviceKey ("DefaultChatOptions"?), but it would be nice to have a standard way of retrieving these values.

### API Proposal

```csharp
namespace System.Collections.Generic;

public interface IChatClient : IDisposable
{
...
ChatOptions? DefaultChatOptions { get; }
...
}
```

### API Usage

```csharp
IChatClient chatClient = ...;
// Check upper bound for MaxOutputTokens
if (desiredMaxOutputTokens < chatClient.DefaultChatOptions.MaxOutputTokens)
{
....
}
```

### Alternative Designs

I can also see this being part of ChatClientMetadata, and maybe not even using ChatOptions directly, since it might have too many information that doesn't make sense for a default, but at least some default properties should be retrievable, somehow.

### Risks

Its just extra metadata, so I don't see many risks. It is more work on implementations, but it should be optional/nullable.

luisquintanilla:
cc: @eiriktsarpalis @SteveSandersonMS @stephentoub for visibility.

stephentoub:
> The only way to achieve this today is to use the GetService method, maybe sending a specific serviceKey ("DefaultChatOptions"?)

Why would it need a specific key? i.e. why not just ask for a ChatOptions? `.GetService<ChatOptions>()`?

stephentoub:
Note, though, that in many cases an implementation won't know what the "defaults" are. Some service provides default values, and a client can choose to override whatever those defaults may be by sending new ones. Such a client would have no way to know what the defaults are. I'd be hesitant to add something beyond just saying "an implementation that wants to provide options can do so via GetService, it's there for things like this".

azchohfi:
I agree that it wouldn't need one, just thought it might have a clearer intent.
I also understand the guidance, just thought it could help with some specific models, such as with ORTGenAI.

stephentoub:
[...]

PR #6159, opened 2025-03-19
Enable JSON-schema based structured output by default

### LINKED ISSUE #6126: Default useNativeJsonSchema to true
Per https://github.com/dotnet/extensions/pull/6064#discussion_r1989349126, flip useNativeJsonSchema to default to true.

In addition or alternatively, we should consider exposing a capabilities mechanism that IChatClients can implement. This could be:
- Augmenting ChatClientMetadata with a capabilities enum or struct, where the client can make claims about supported features, like function calling, structured output, etc.
- Adding a marker interface for the capability, such that it can be queried via GetService

useNativeJsonSchema could then default to false unless this flag is observed.

SteveSandersonMS:
**Update** Following discussion, this change has been reduced to:

 * Changing the default for `useJsonSchema` to true
 * Renaming `ModelId`/`Dimensions` to `DefaultModelId`/`DefaultModelDimensions`

Plus other minor related renames and integration test fixes.

---

### Old description

Fixes #6126 based on the design discussed there.

This introduces a new concept of per-model metadata. Usage:

```cs
var providerMetadata = chatClient.GetService<ChatClientMetadata>();
Console.WriteLine("Provider name: " + providerMetadata.ProviderName);

var defaultModelMetadata  = await providerMetadata.GetModelMetadataAsync();
Console.WriteLine("Supports native JSON schema: " + defaultModelMetadata.SupportsNativeJsonSchema);

var anotherModelMetadata = await providerMetadata.GetModelMetadataAsync("anotherModelId");
```

There's also the equivalent for embedding generators, and that's now how you can find the `Dimensions` value.

### Structured output

`GetResponse<T>` still allows the developer to specify `useNativeJsonSchema`, but if it's unspecified, the default is now taken from the model metadata.

### Providers

* Ollama always supports structured output
* OpenAI and Azure AI Inference vary based on the model

This PR bakes in knowledge about which model IDs currently support structured output. I'm on the fence about how much we really want to get into this business, but there's no way to get the info programmatically. Neither OpenAI or Azure AI Inference expose this metadata on the endpoints they have for describing models.

I've tried to make the logic as forgiving as possible if the external reality changes and new models appear, but having correct metadata here is a best-effort thing that developers would still have to override on a per-call basis if our default doesn't match their requirement.

The main alternative would be to remove the built-in knowledge of specific models and just default to "unknown" for the OpenAI/AIInference providers across all models. In effect that would retain the same behavior we have before this PR. However it doesn't move us forwards and cements an old-fashioned view of the world.

Yet another alternative would be defaulting to "true" for all models, since *most* OpenAI models do support it. However even some new-ish OpenAI models (e.g., o1-mini) don't support it, and of course many Azure AI Inference ones don't since it includes lots of small models.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6159)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatModelMetadata.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Provides metadata about a model used with an <see cref="IChatClient"/>.</summary>
+public class ChatModelMetadata
+{
+    /// <summary>
+    /// Gets a value indicating whether the model can produce structured output conforming to a JSON schema.
+    /// </summary>
+    public bool? SupportsNativeJsonSchema { get; init; }
```
SteveSandersonMS, line 12: Not sure about the naming any more. Is the word "native" doing anything useful here?

Could be "SupportsStrictJsonSchema" or just "SupportsJsonSchema".

Could be "SupportsStructuredOutput" but the only kind we support is JSON schema.

shyamnamboodiripad (approved):
Approving for the change under the Eval project 👍🏾

peterwald (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatModelMetadata.cs
stephentoub: I'd drop "native", both here and in the parameter name to `GetResponseAsync<T>`. SupportsJsonSchema, SupportsJsonSchemaResponseFormat, and SupportsStructuredOutput all sound fine to me.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorMetadata.cs
```diff
@@ -35,17 +36,19 @@ public EmbeddingGeneratorMetadata(string? providerName = null, Uri? providerUri
...
+    public string? DefaultModelId { get; }
+
+    /// <summary>
+    /// Gets metadata for the the default model or a specified model.
+    /// </summary>
+    /// <param name="modelId">The model identifier. If not set, uses <see cref="DefaultModelId"/>.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that completes with the model metadata.</returns>
+    public virtual Task<EmbeddingModelMetadata> GetModelMetadataAsync(string? modelId = null, CancellationToken cancellationToken = default)
+        => Task.FromResult(_emptyModelMetadata);
```
stephentoub, line 53: Can we cache the `Task<Metadata>` rather than just the `Metadata`?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    // See https://platform.openai.com/docs/guides/structured-outputs
+    private static readonly string[] _supportedStructuredOutputModelIdPrefixes =
+    [
+        "gpt-4.5",
+        "o3-mini",
+        "o1",
+        "gpt-4o",
+        "gpt-4o-mini",
+    ];
```
stephentoub, line 25: I realize it might shoot a big whole in this whole plan, but I don't think we should be hardcoding model IDs like this. It'll require constant maintenance, forces deploying new binaries along with new models, etc. The OpenAI client in general is also usable with anything that's OpenAI compatible, so there are many models that support structured output that aren't listed here. And in the case of Azure OpenAI, aren't we actually providing a deployment ID rather than a model ID? In which case there's no guarantee the ID has anything to do with the model it maps to.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            modelId = DefaultModelId;
+        }
+
+        // There could be a brief race here and we could add more than the intended max items to the cache.
+        // That wouldn't be a problem since it's an arbitrary limit to mitigate memory leaks if the application
+        // is doing something very strange (generating distinct model IDs per call).
+        var result = _metadataByModelId.Count > MaxMetadataCacheEntries
+            ? (_metadataByModelId.TryGetValue(modelId ?? string.Empty, out var entry) ? entry : CreateModelMetadata(modelId))
+            : _metadataByModelId.GetOrAdd(modelId ?? string.Empty, CreateModelMetadata);
```
stephentoub, line 46: Count on a ConcurrentDictionary is unfortunately not as cheap as would be ideal. It'd be better to only check it once we know we've missed the cache, e.g.
```C#
if (!_metadataByModelId.TryGetValue(modelId, out var result)
{
    result = CreateModelMetadata(modelId);
    if (_metadataByModelId.Count < MaxMetadataCacheEntries)
    {
        result = _metadataByModelId.GetOrAdd(modelId, result);
    }
}

return result;
```

Also, I'm unclear what all these string.Empty's are for. If the modelId is null, we still want to use the cache?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -223,6 +222,17 @@ public static async Task<ChatResponse<T>> GetResponseAsync<T>(
         return new ChatResponse<T>(result, serializerOptions) { IsWrappedInObject = isWrappedInObject };
     }
 
+    private static async Task<bool?> ModelSupportsNativeJsonSchemaAsync(IChatClient chatClient, string? modelId, CancellationToken cancellationToken)
```
stephentoub, line 225: Nit: Task => ValueTask

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGeneratorMetadata.cs
SteveSandersonMS: Good point. Done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatModelMetadata.cs
SteveSandersonMS: Sounds good.

Since the metadata property isn't something normal app code would use, I've gone with the most explicit phrasing (`SupportsJsonSchemaResponseFormat`), just in case other kinds of structured output emerge in the future.

For the `GetResponseAsync<T>` parameter, it's now `useJsonSchema`.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
SteveSandersonMS: > Also, I'm unclear what all these string.Empty's are for. If the modelId is null, we still want to use the cache?

I'm not sure the modelId can actually be null. `OpenAIChatClient` throws if you pass a null/empty modelId constructor param, but there's a second constructor that gets modelId using private reflection, so I was unusure if there's some case where a `null` value has a meaning like "use a global default". Based on your comment I've changed this to skip the cache and return empty metadata if the resolve modelId is null/empty.

> Count on a ConcurrentDictionary is unfortunately not as cheap as would be ideal

Thanks. Updated.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
SteveSandersonMS: Yeah, I know it's undesirable. I thought about the Azure OpenAI deployment name issue last night and was hoping it would be possible to use `openAIClient.GetOpenAIModelClient().GetModelAsync` to get the actual model name behind the deployment, but having tried it I see that Azure OpenAI doesn't support that. And even if it did, presumably you could still have custom finetuned models that genuinely have custom names with nothing recognizable about them. And even we could find out the underlying base model somehow, that still leaves the issue of us being responsible for the maintenance of this info.

Do you have a preference about how we should proceed?

If we're not baking in any knowledge about models, and don't have any API endpoint that gives useful capability metadata, it seems like our options are reduced to:

 * Have OpenAI/AzureAIInference assume that **all** models support JSON schema
 * Have OpenAI/AzureAIInference assume that **no** models support JSON schema

Of these options, I'm more attracted to assuming they do support it. Developers are only likely to use `GenerateAsync<T>` with models that do support structured output, so in most cases it won't be an issue. In the case they use `GenerateAsync<T>` with a non-JSON-schema-supporting model, they will have to pass `useJsonSchema: false`. This also sets us up to have a more useful default in the future when we can (I hope!) expect that the majority of emerging models will support this feature.

SteveSandersonMS: More broadly we could argue that this means there's no current situation where we are using real per-model capabilities, so why even have support for this concept in the API? We could skip almost all the changes in this PR and just flip the default in `ChatClientStructuredOutputExtensions`.

If we keep the per-model metadata feature, it would just be to enable expanding in this direction in the future, or for 3rd-party providers that can provide the metadata. It seems that Ollama does actually expose some of this info, e.g., you can [query to get the context length or number of dimensions on an embedding model](https://github.com/ollama/ollama/blob/main/docs/api.md#show-model-information), so maybe that's a good enough validation that this could be used.

luisquintanilla:
cc: @azchohfi @nmetulev for awareness and had similar asks as part of #5941

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
SteveSandersonMS: I've updated the PR to do as described above (remove the hardcoded model IDs; default JSON schema support to true on OpenAI/AzureAIInference), in case we all agee that's the desired outcome.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -488,4 +488,14 @@ private static OllamaTool ToOllamaTool(AIFunction function)
...
     }
+
+    private sealed class OllamaChatClientMetadata(string provider, Uri endpoint, string? modelId)
+        : ChatClientMetadata(provider, endpoint, modelId)
+    {
+        // Ollama has supported native JSON schema for a while, so we'll assume it's always available
+        private static readonly ChatModelMetadata _metadata = new ChatModelMetadata { SupportsJsonSchemaResponseFormat = true };
+
+        public override Task<ChatModelMetadata> GetModelMetadataAsync(string? modelId = null, CancellationToken cancellationToken = default)
+            => Task.FromResult(_metadata);
```
stephentoub, line 499: Cache the task instead?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
```diff
@@ -0,0 +1,27 @@
...
+internal sealed class OpenAIChatClientMetadata(string provider, Uri endpoint, string? modelId)
+    : ChatClientMetadata(provider, endpoint, modelId)
+{
+    // Since this provider represents an arbitrary OpenAI-compatible endpoint, we can't know about
+    // model-specific capabilities. We don't even know which base model a request will use, since
+    // the "modelId" might refer to a custom model or be the name of a deployment on Azure OpenAI.
+    private static readonly Task<ChatModelMetadata> _fixedModelMetadata = Task.FromResult(new ChatModelMetadata
+    {
+        // This is true for most OpenAI models. We consider it useful to default to true, since
+        // then GenerateAsync<T> will behave well when used with models that support structured
```
stephentoub, line 19: "GenerateAsync"... is that supposed to be "GetResponseAsync"?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIEmbeddingGenerator.cs
```diff
@@ -160,4 +160,18 @@ private static EmbeddingGeneratorMetadata CreateMetadata(string providerName, st
...
+        private readonly EmbeddingModelMetadata _modelMetadata;
+
+        public OpenAIEmbeddingGeneratorMetadata(string? providerName, Uri? providerUri, string? defaultModelId, int? dimensions)
+            : base(providerName, providerUri, defaultModelId)
+        {
+            _modelMetadata = new() { Dimensions = dimensions };
+        }
+
+        public override Task<EmbeddingModelMetadata> GetModelMetadataAsync(string? modelId = null, CancellationToken cancellationToken = default)
+            => Task.FromResult(_modelMetadata);
```
stephentoub, line 175: The ctor could also cache the task instead of the metadata.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIEmbeddingGenerator.cs
```diff
@@ -160,4 +160,18 @@ private static EmbeddingGeneratorMetadata CreateMetadata(string providerName, st
 
         return openAIOptions;
     }
+
+    private sealed class OpenAIEmbeddingGeneratorMetadata : EmbeddingGeneratorMetadata
+    {
+        private readonly EmbeddingModelMetadata _modelMetadata;
+
+        public OpenAIEmbeddingGeneratorMetadata(string? providerName, Uri? providerUri, string? defaultModelId, int? dimensions)
```
stephentoub, line 168: Is there an assumption here that `dimensions` corresponds to `defaultModelId` or to any model that might be used? GetModelMetadataAsync is assuming the latter.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
stephentoub: > More broadly we could argue that this means there's no current situation where we are using real per-model capabilities, so why even have support for this concept in the API? We could skip almost all the changes in this PR and just flip the default in `ChatClientStructuredOutputExtensions`.
> 
> If we keep the per-model metadata feature, it would just be to enable expanding in this direction in the future, or for 3rd-party providers that can provide the metadata. It seems that Ollama does actually expose some of this info, e.g., you can [query to get the context length or number of dimensions on an embedding model](https://github.com/ollama/ollama/blob/main/docs/api.md#show-model-information), so maybe that's a good enough validation that this could be used.

I'm torn. I'm not 100% sure this model will stand the test of time, and as you say, at present this is really just adding overhead and an escape hatch, with the actual impact being just the original proposal of flipping the default.

Do we think there's a minimal set of changes we could make now, like renaming "modelId" to "defaultModelId", and flipping the default, that wouldn't prevent us from adding the extensibility in the future?

SteveSandersonMS: Cool. I'm fine with stripping this back to just a couple of renames and changing the default for `useJsonSchema`.

As for whether it keeps the option to add this extensibility in the future, yes I think it does. The only thing that would potentially be weird is that `EmbeddingGeneratorMetadata` will have a `Dimensions` on it that should really be model-specific but won't be. Perhaps it's enough just to rename that to `DefaultModelDimensions`.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -186,16 +186,17 @@ public static async Task<ChatResponse<T>> GetResponseAsync<T>(
                 { "type", "object" },
                 { "properties", new JsonObject { { "data", JsonElementToJsonNode(schemaElement) } } },
                 { "additionalProperties", false },
+                { "required", new JsonArray("data") },
```
SteveSandersonMS, line 189: The absence of this previously was a bug. Native structured output for non-object types was failing on OpenAI due to it rejecting the schema.

test/Libraries/Microsoft.Extensions.AI.AzureAIInference.Tests/IntegrationTestHelpers.cs
```diff
@@ -21,12 +24,49 @@ internal static class IntegrationTestHelpers
...
+    {
+        var result = new AzureAIInferenceClientOptions();
+
+        // The API vesion set here corresponds to the value used by AzureOpenAIClientOptions
+        // if the AZURE_OPENAI_GA flag is set during its compilation. This API version is the
+        // minimum required for structured output with JSON schema.
+        result.AddPolicy(new OverrideApiVersionPolicy("2024-08-01-preview"), HttpPipelinePosition.PerCall);
+
+        return result;
+    }
```
SteveSandersonMS, line 46: This problem will go away when AzureAIInference client becomes GA. See https://github.com/Azure/azure-sdk-for-net/blob/81e59aaccaea4b2ab82aded2f6cd1b17fa3a2c3e/sdk/openai/Azure.AI.OpenAI/src/Custom/AzureOpenAIClientOptions.cs.

Until then structured output isn't enabled by default on AzureAIInference.

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
```diff
@@ -819,7 +820,7 @@ private class Person
 
     private enum JobType
     {
-        Surgeon,
+        Wombat,
```
SteveSandersonMS, line 823: `llama3.1` is absolutely obsessed with claiming that people are surgeons, even Taylor Swift. I had to remove that as one of the options otherwise it would always pick it.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
SteveSandersonMS: Done.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClientMetadata.cs
SteveSandersonMS: Code was removed.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIEmbeddingGenerator.cs
SteveSandersonMS: Code was removed.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIEmbeddingGenerator.cs
SteveSandersonMS: Renamed to `defaultModelDimensions` for clarity.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
SteveSandersonMS: Code was removed.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -43,18 +43,18 @@ public static Task<ChatResponse<T>> GetResponseAsync<T>(
...
     /// <summary>Sends a user chat text message, requesting a response matching the type <typeparamref name="T"/>.</summary>
     /// <param name="chatClient">The <see cref="IChatClient"/>.</param>
     /// <param name="chatMessage">The text content for the chat message to send.</param>
     /// <param name="options">The chat options to configure the request.</param>
-    /// <param name="useNativeJsonSchema">
+    /// <param name="useJsonSchema">
     /// Optionally specifies whether to set a JSON schema on the <see cref="ChatResponseFormat"/>.
     /// This improves reliability if the underlying model supports native structured output with a schema, but may cause an error if the model does not support it.
-    /// If not specified, the underlying provider's default will be used.
+    /// If not specified, the default value is <see langword="true" />.
```
stephentoub, line 57: I assume we're keeping it as a `bool?` so that in the future we have wiggle-room around this default, right? e.g. if we add a discovery mechanism in the future. Is there a different wording here would could use to make it clear that it's up to the system but it errs on the side of true?

test/Libraries/Microsoft.Extensions.AI.AzureAIInference.Tests/IntegrationTestHelpers.cs
stephentoub: That link is for Azure.AI.OpenAI rather than Azure.AI.Inference. This code is Azure.AI.Inference.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
SteveSandersonMS: I could change this back to something more like the previous phrasing if that adds legitimacy to updating this later, e.g.

```suggestion
    /// If not specified, the a default value may be determined by the provider or model.
```

stephentoub: Maybe just "If not specified, the default value is determined by the implementation; if a specific value must be guaranteed, it should be specified by the caller", or something like that?

I'm probably also overthinking it. Go with whatever you prefer.

test/Libraries/Microsoft.Extensions.AI.AzureAIInference.Tests/IntegrationTestHelpers.cs
SteveSandersonMS: Sorry, yes you're correct. The Azure.AI.Inference equivalent appears to be https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/ai/Azure.AI.Inference/src/Generated/AzureAIInferenceClientOptions.cs.

I took the overridden value here (`2024-08-01-preview`) from the error message that actually comes back from the Azure AI inference endpoint (specifically, an endpoint URL that looks like `https://my_resource_name.services.ai.azure.com`, which as far as I understand really is Azure AI Inference) if you don't specify it.

So it looks like we don't have any definite info about when the Azure.AI.Inference client will be updated.

PR #6160, opened 2025-03-19
Add license info to the Azure DevOps package

peterwald:


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/azure-devops-report/vss-extension.json
```diff
@@ -25,8 +25,26 @@
...
+        "license": {
+            "path": "LICENSE"
+        }
+    },
+    "links": {
+        "getstarted": {
+            "uri": "https://learn.microsoft.com/en-us/dotnet/ai/quickstarts/evaluate-ai-response"
+        },
+        "learn": {
+            "uri": "https://learn.microsoft.com/en-us/dotnet/ai/conceptual/evaluation-libraries"
```
shyamnamboodiripad, line 38: It may be nice to include the samples link in this pane here as well

shyamnamboodiripad (approved):

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/azure-devops-report/vss-extension.json
peterwald: I thought about that but these are not links we can specify. There are actually a set of predetermined keys that are accepted here. There is not one for samples.

PR #6162, opened 2025-03-20
Enable custom argument binding with AIFunctionFactory

stephentoub:
This in particular enables higher-level libraries that need to create AI-invocable functions to use AIFunctionFactory for the bulk of the logic, even if they need to specialize behavior for certain arguments.


SteveSandersonMS (approved):
I'm fine with this as-is but want to suggest a possible variation.

With the approach here, we re-query the custom binder on every call. So in the `[FromKeyedServices]` case, it repeats the same reflection calls for every parameter on every call (I think?).

An approach more similar to various bits of data binding in ASP.NET Core and Blazor would be to have the API be a "CanBindParameter" and "BindParameter" pair. We would then only call "CanBindParameter" once per parameter for the lifetime of the AIFunction. Potentially, if it returns `true`, we'd use that to exclude the parameter from the JSON schema instead of also needing an `IncludeParameter` API (though this depends on whether there are use cases for customizing the JSON-marshalling, in which case maybe you want to include it in schema even if there's a custom binder).

Perhaps this matters more to ASP.NET Core where a given binder may be reused millions of times across the lifetime of an application, whereas an `AIFunction` tends to be scoped more tightly to a unit of work and maybe only reused tens of times.

stephentoub:
That's reasonable. I'll switch it.

stephentoub:
@SteveSandersonMS, feedback addressed. I revised it to instead have a two-delegate model, ConfigureParameterBinding and BindParameter. It covers separately configuring whether parameter should be excluded from the schema from whether BindParameter is used, as well as passing additional information from ConfigureParameterBinding into BindParameter so as to avoid duplicative reflection.

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactoryOptions.cs
```diff
@@ -55,4 +56,91 @@ public AIFunctionFactoryOptions()
...
+        /// <see cref="AIFunctionArguments.Context"/>. However, it may be used even when <see cref="ExcludeFromSchema"/> is <see langword="false"/>,
+        /// in a situation where it's desirable for the AI service to generate a value for the parameter, which <see cref="BindParameter"/>
+        /// may then use as part of its logic.
+        /// </para>
+        /// <para>
+        /// It is an error for <see cref="ConfigureParameterBinding"/> to return a <see cref="ParameterBindingOptions"/> with
+        /// <see cref="UseBindParameter"/> set to <see langword="true"/> when <see cref="BindParameter"/> is <see langword="null"/>.
+        /// </para>
+        /// </remarks>
+        public bool UseBindParameter { get; init; }
```
eiriktsarpalis, line 111: Why not just specify an individual binding delegate for each parameter binding options? I'm assuming you could avoid allocating a new delegate for every parameter being examined.

stephentoub: Hmm, they would probably simplify a few things. Will try.

stephentoub: Done.

PR #6169, opened 2025-03-20
Simplify function calling error handling

### LINKED ISSUE #6029: Update function calling error handling logic
Today we have two built-in error handling modes:

 * `RetryOnError=false`
   * On exception, does one final iteration to get an answer, during which no tools are available
   * The point of this is to produce some "Sorry, it didn't work" response without the app developer needing to have caught the error in their `AIFunction` (it's difficult to return errors from an `AIFunction`, e.g., because it might be declared to return an `int`)
 * `RetryOnError=true`
   * On exception, just keeps going round the loop until `MaximumIterationsPerRequest` (which itself is unlimited by default)

Proposed changes:

 * Be more aggressive about terminating the loop, whether or not there's an error
   * Default `MaximumIterationsPerRequest` to 10, say (TBD)
   * Add a further `MaximumConsecutiveErrorsPerRequest` defaulting to 3, say
   * Collectively these should greatly limit the scope for resource wastage in the case of an LLM trying to call a problematic function infinitely, or some prompt injection attack persuading the LLM to call a successful function infinitely
* Replace `RetryOnError` with a `FunctionInvocationErrorHandlingMode` ~~enum~~ class hierarchy with possible values `Throw`, `Retry`
  * Throw would be new and just means don't even catch the exception (well, rethrow it). This is preferred for non-chat-loop cases.
  * Retry would be equivalent to today's RetryOnError=true
  * I'm no longer convinced we should have a built-in equivalent to RetryOnError=false (i.e., "one last try, but without tools") because that's so nonobvious and doesn't make any sense in structured output cases.
  * So the default would be `Retry`, and you'd hope the LLM is smart enough to stop retrying after failure, but if it doesn't it's limited by `MaximumConsecutiveErrorsPerRequest` anyway
* Extensibility:
  * Actually it's not really right to configure this on the `FunctionCallingChatClient`, since it should be set on a per-call basis.
  * So in the same way that `ChatOptions` has a `ToolMode`, it would be best if it also had a `ToolErrorHandlingMode`, which should even be a hierarchy of types in the same way as `ChatToolMode`. That way (1) it can be set per-call, and (2) you can create a custom subclass if you are also implementing custion function calling. Then if you really want to implement something equivalent to "one last try" you could do so, for example either by subclassing `FunctionCallingChatClient` or by implementing some other middleware that catches the throw and makes one last call but without tools.

SteveSandersonMS:
Fixes #6029, following the design described there

 * Remove the `RetryOnError` config option
 * Add a new config option `MaximumConsecutiveErrorsPerRequest` (default: 3)
 * Change `MaximumIterationsPerRequest` to be non-nullable and default to 10

This is sufficient to model various error-handling modes:

 * To rethrow exceptions in all cases, set `MaximumConsecutiveErrorsPerRequest` to 0
 * To retry almost-forever like we had before, set `MaximumConsecutiveErrorsPerRequest` and `MaximumIterationsPerRequest` both to `int.MaxValue` but **this is almost certainly a bad idea**
 * To get the "one last try but without tools" behavior, have `MaximumConsecutiveErrorsPerRequest > 0` and add custom middleware that observes `FunctionResultContent.Exception` being nonnull and uses that as a signal to change `chatOptions.Tools` to `[]`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6169)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -16,6 +16,7 @@ using static Microsoft.Extensions.AI.OpenTelemetryConsts.GenAI;
 
 #pragma warning disable CA2213 // Disposable fields should be disposed
 #pragma warning disable EA0002 // Use 'System.TimeProvider' to make the code easier to test
+#pragma warning disable SA1202 // 'protected' members should come before 'private' members
 
 namespace Microsoft.Extensions.AI;
 
@@ -86,23 +87,6 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
         get => _currentContext.Value;
         protected set => _currentContext.Value = value;
     }
-
-    /// <summary>
-    /// Gets or sets a value indicating whether to handle exceptions that occur during function calls.
-    /// </summary>
-    /// <value>
-    /// <see langword="false"/> if the
-    /// underlying <see cref="IChatClient"/> will be instructed to give a response without invoking
-    /// any further functions if a function call fails with an exception.
-    /// <see langword="true"/> if the underlying <see cref="IChatClient"/> is allowed
-    /// to continue attempting function calls until <see cref="MaximumIterationsPerRequest"/> is reached.
-    /// The default value is <see langword="false"/>.
-    /// </value>
-    /// <remarks>
-    /// Changing the value of this property while the client is in use might result in inconsistencies
-    /// as to whether errors are retried during an in-flight request.
-    /// </remarks>
-    public bool RetryOnError { get; set; }
 
     /// <summary>
     /// Gets or sets a value indicating whether detailed exception information should be included
@@ -260,7 +244,7 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
             var modeAndMessages = await ProcessFunctionCallsAsync(augmentedHistory, options!, functionCallContents!, iteration, cancellationToken).ConfigureAwait(false);
             responseMessages.AddRange(modeAndMessages.MessagesAdded);
 
-            if (UpdateOptionsForMode(modeAndMessages.Mode, ref options!, response.ChatThreadId))
+            if (UpdateOptionsForMode(modeAndMessages.ShouldContinue, ref options!, response.ChatThreadId))
             {
                 // Terminate
                 break;
@@ -362,7 +346,7 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
                 Activity.Current = activity; // workaround for https://github.com/dotnet/runtime/issues/47802
             }
 
-            if (UpdateOptionsForMode(modeAndMessages.Mode, ref options, response.ChatThreadId))
+            if (UpdateOptionsForMode(modeAndMessages.ShouldContinue, ref options, response.ChatThreadId))
             {
                 // Terminate
                 yield break;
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -184,6 +170,48 @@ public int? MaximumIterationsPerRequest
...
+        get => _maximumConsecutiveErrorsPerRequest;
+        set
+        {
+            if (value < 0)
+            {
+                Throw.ArgumentOutOfRangeException(nameof(value));
+            }
+
+            _maximumConsecutiveErrorsPerRequest = value;
+        }
```
stephentoub, line 212:
```suggestion
        set => _maximumConsecutiveErrorsPerRequest = Throw.IfLessThan(value, 0);
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -458,46 +492,23 @@ private static bool CopyFunctionCalls(
     }
 
     /// <summary>Updates <paramref name="options"/> for the response.</summary>
-    /// <returns>true if the function calling loop should terminate; otherwise, false.</returns>
-    private static bool UpdateOptionsForMode(ContinueMode mode, ref ChatOptions options, string? chatThreadId)
+    private static void UpdateOptionsForMode(ref ChatOptions options, string? chatThreadId)
```
stephentoub, line 495: The "Mode" in this name was about ContinueMode, which is now gone. Seems like this should be renamed.

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/FunctionInvokingChatClientTests.cs
```diff
@@ -201,6 +201,125 @@ public async Task ConcurrentInvocationOfParallelCallsDisabledByDefaultAsync()
...
+            // Any failure within an iteration causes the whole iteration to be treated as failed
+            ..CreateIterationPlan(ref callIndex, false, true, false),
+
+            // Now NumConsecutiveErrors = 1
+            // Even if several calls in the same iteration fail, that only counts as a single iteration having failed, so won't exceed the limit yet
+            ..CreateIterationPlan(ref callIndex, true, true, true),
+
+            // Now NumConsecutiveErrors = 2
+            // Any more failures will now exceed the limit
+            ..CreateIterationPlan(ref callIndex, true, true),
```
stephentoub, line 290: Nice use of spreads :)

stephentoub (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -184,6 +170,48 @@ public int? MaximumIterationsPerRequest
...
+    /// recover from errors by trying other function parameters that may succeed.
+    /// </para>
+    /// <para>
+    /// However, in case function invocations continue to produce exceptions, this property can be used to
+    /// limit the number of consecutive failing attempts. When the limit is reached, the exception will be
+    /// rethrown to the caller.
+    /// </para>
+    /// <para>
+    /// If the value is set to zero, all function calling exceptions immediately terminate the function
+    /// invocation loop and the exception will be rethrown to the caller.
```
stephentoub, line 194: "and the exception will be rethrown to the caller"

Do you want to say anything about there possibly being multiple exceptions and them all being wrapped in an aggregate?

Also, did you consider not using an aggregate when only a single exception is incurred? I know we always use an aggregate with Task.Wait and friends, but I'm not sure it was the right choice.

Also seems like when AllowConcurrentInvocation is false (the default), we might just want to let the exception propagate if the limit has been reached?

stephentoub: If a single exception propagates, mostly just implementation detail, albeit with an extra catch/throw in the mix and the resulting impact on diagnostics. Not a big deal, though.

SteveSandersonMS: I guess it would make more of a behavioral difference because if there are multiple function calls in the current iteration, propagating the first exception directly would stop subsequent ones from running at all.

This is *probably* desirable though it's quite a subtle distinction. I've implemented it.

PR #6172, opened 2025-03-20
Add MessageId to ChatMessage and ChatResponseUpdate

### LINKED ISSUE #6095: Consider adding MessageId to ChatMessage and ChatResponseUpdate
As tools and agents become more prevalent, it's more likely to have multiple messages that are part of a single overall operation, which is represented as a ChatResponse. We have a ResponseId on ChatResponse, but we lack a unique identifer on each invidual ChatMessage that makes up the ChatResponse. Similarly, ChatResponseUpdate has a ResponseId on it, but it doesn't have a MessageId. We end up using the ChatResponseUpdate.ResponseId logically as a message ID to differentiate where one message ends and another begins in a stream, but this could be made cleaner by making the grouping of updates into messages more explicit by having a MessageId on each ChatResponseUpdate in addition to the ResponseId.

SteveSandersonMS:
Fixes #6095

This is a pretty delicate change that impacts plumbing all over. I'm trying to be pretty systematic in checking that all the right things are updated. If anyone can think of anything I may have missed, please say so!

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6172)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
@@ -96,6 +96,7 @@ public sealed class AzureAIInferenceChatClient : IChatClient
         // Create the return message.
         ChatMessage message = new(ToChatRole(response.Role), response.Content)
         {
+            MessageId = response.Id, // There is no per-message ID, but there's only one message per response, so use the response ID
             RawRepresentation = response,
         };
 
@@ -157,7 +158,7 @@ public sealed class AzureAIInferenceChatClient : IChatClient
             // The role and finish reason may arrive during any update, but once they've arrived, the same value should be the same for all subsequent updates.
             streamedRole ??= chatCompletionUpdate.Role is global::Azure.AI.Inference.ChatRole role ? ToChatRole(role) : null;
             finishReason ??= chatCompletionUpdate.FinishReason is CompletionsFinishReason reason ? ToFinishReason(reason) : null;
-            responseId ??= chatCompletionUpdate.Id;
+            responseId ??= chatCompletionUpdate.Id; // While it's unclear from the name, this Id is documented to be the response ID, not the chunk ID
             createdAt ??= chatCompletionUpdate.Created;
             modelId ??= chatCompletionUpdate.Model;
 
@@ -168,7 +169,8 @@ public sealed class AzureAIInferenceChatClient : IChatClient
                 FinishReason = finishReason,
                 ModelId = modelId,
                 RawRepresentation = chatCompletionUpdate,
-                ResponseId = chatCompletionUpdate.Id,
+                ResponseId = responseId,
+                MessageId = responseId, // There is no per-message ID, but there's only one message per response, so use the response ID
                 Role = streamedRole,
             };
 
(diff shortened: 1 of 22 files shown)


src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -96,6 +96,7 @@ public async Task<ChatResponse> GetResponseAsync(
         // Create the return message.
         ChatMessage message = new(ToChatRole(response.Role), response.Content)
         {
+            MessageId = response.Id, // There is no per-message ID, but there's only one message per response, so use the response ID
```
stephentoub, line 99: I _think_ this is the right answer. The other would be to generate a unique ID for each message. Did you consider and dismiss that? Just want to make sure we've thought it through.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
```diff
@@ -271,7 +272,10 @@ private static ChatMessage FromOllamaMessage(OllamaChatResponseMessage message)
             contents.Insert(0, new TextContent(message.Content));
         }
 
-        return new ChatMessage(new(message.Role), contents);
+        // Ollama doesn't set a response ID on responses or messages, so generate one for consistency with streaming
+        var messageId = Guid.NewGuid().ToString("N");
+
+        return new ChatMessage(new(message.Role), contents) { MessageId = messageId };
```
stephentoub, line 278: Is this message ID then inherited to be the response ID as well?

Trying to rationalize this with the AzureAIInference one above, where it sets them to be the same.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -187,6 +189,8 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
 
             if (streamingUpdate is StreamingResponseItemUpdate itemUpdate)
             {
+                messageId = itemUpdate.Item.Id;
```
stephentoub, line 192: If you search this file for "msg_ignored", there are a few places these message id could be propagated back to the underlying API (though I think those might go away).

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
SteveSandersonMS: I did consider returning a new GUID or something but don't see who it would help. If we can return an ID that actually came from the service then at least it will mean something if people are searching through logs or telemetry data to coordinate the ID with real-world activity.

src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs
SteveSandersonMS: Good point. This was inconsistent. I've pushed an update to align it.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
SteveSandersonMS: Done

stephentoub (approved):

SteveSandersonMS:
I verified the E2E integration tests all pass on OpenAI, AzureAIInference, and Ollama.

PR #6177, opened 2025-03-21
Add "see also" links to conceptual docs

gewarren:
Fixes dotnet/docs#11000.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6177)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
@@ -10,7 +10,7 @@ using System.Text.Json.Serialization;
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a chat message used by an <see cref="IChatClient" />.</summary>
-/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET</related>
+/// <related type="Article" href="https://learn.microsoft.com/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET.</related>
 [DebuggerDisplay("[{Role}] {ContentForDebuggerDisplay}{EllipsesForDebuggerDisplay,nq}")]
 public class ChatMessage
 {
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
@@ -10,6 +10,7 @@ using System.Text.Json.Serialization;
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a chat message used by an <see cref="IChatClient" />.</summary>
+/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET</related>
 [DebuggerDisplay("[{Role}] {ContentForDebuggerDisplay}{EllipsesForDebuggerDisplay,nq}")]
 public class ChatMessage
 {
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Extensions.AI;
 /// provided with a callback that could mutate the supplied options argument, and that should be avoided if using a singleton options instance.
 /// </para>
 /// </remarks>
-/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET</related>
+/// <related type="Article" href="https://learn.microsoft.com/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET.</related>
 public interface IChatClient : IDisposable
 {
     /// <summary>Sends chat messages and returns the response.</summary>
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.AI;
 /// provided with a callback that could mutate the supplied options argument, and that should be avoided if using a singleton options instance.
 /// </para>
 /// </remarks>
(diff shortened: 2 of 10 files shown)


peterwald (approved):

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
```diff
@@ -162,6 +163,7 @@ public ReportingConfiguration(
     /// A new <see cref="ScenarioRun"/> with the specified <paramref name="scenarioName"/> and
     /// <paramref name="iterationName"/>.
     /// </returns>
+    /// <related type="Article" href="/dotnet/ai/tutorials/evaluate-with-reporting">Tutorial: Evaluate a model's response with response caching and reporting</related>
```
shyamnamboodiripad, line 166: Perhaps it may be better to add this on the interface IEvaluator (otherwise we should probably add it to all the other evaluators like FluencyEvaluator, RTCEvaluator etc. in this project)

src/Libraries/Microsoft.Extensions.AI.Evaluation/IEvaluator.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Extensions.AI.Evaluation;
 /// <summary>
 /// Evaluates responses produced by an AI model.
 /// </summary>
+/// <related type="Article" href="/dotnet/ai/tutorials/evaluate-with-reporting">Tutorial: Evaluate a model's response with response caching and reporting</related>
```
shyamnamboodiripad, line 13: Did you intend to use the other link here (i.e. evaluate-ai-response). IEvaluator does not depend on Reporting concepts -  so maybe better to link the other article here.

src/Libraries/Microsoft.Extensions.AI.Evaluation/NumericMetric.cs
```diff
@@ -24,5 +24,6 @@ namespace Microsoft.Extensions.AI.Evaluation;
 /// An optional string that can be used to provide some commentary around the result represented by this
 /// <see cref="NumericMetric"/>.
 /// </param>
+/// <related type="Article" href="/dotnet/ai/quickstarts/evaluate-ai-response">Evaluate a model's response</related>
```
shyamnamboodiripad, line 27: Should we add this to the other metrics too (like StringMetric and BooleanMetric) or perhaps we can remove it from NumericMetric and add this on the base EvaluationMetric type)?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
shyamnamboodiripad: Looks like the link is present on IEvaluator which is great - so then the question is just whether CoherenceEvaluator needs to be special v/s whether we should also add it to the other evaluators I mentioned above.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
```diff
@@ -10,6 +10,7 @@
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a chat message used by an <see cref="IChatClient" />.</summary>
+/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET</related>
```
RussKie, line 13:
```suggestion
/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET.</related>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.AI;
 /// provided with a callback that could mutate the supplied options argument, and that should be avoided if using a singleton options instance.
 /// </para>
 /// </remarks>
+/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET</related>
```
RussKie, line 26:
```suggestion
/// <related type="Article" href="/dotnet/ai/quickstarts/build-chat-app">Build an AI chat app with .NET.</related>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
gewarren: We don't generally put periods/full stops on See also links, so I'm going to resolve this. But let me know if you feel strongly.

shyamnamboodiripad: This repo uses some analyzers that report a warning when the full stop is missing for a comment. That's unfortunately causing the CI build to fail.

Ideally, the analyzer that reports this warning would ignore missing full stops in `<related>` tags. But absent that, the easiest fix is probably to add the full stops. Otherwise, you would need to add `#pragma` to suppress these warnings individually (or suppress them globally in the corresponding projects which could mean also ignoring legit missing full stop bugs in other tags in the future...).

gewarren: Okay, I'll add them. As any FYI, these become bullets in a list [like this](https://learn.microsoft.com/en-us/dotnet/api/system.linq.parallelquery-1.getenumerator?view=net-9.0#see-also).

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
gewarren: CoherenceEvaluator is the only evaluator that's used in [the quickstart](https://learn.microsoft.com/en-us/dotnet/ai/quickstarts/evaluate-ai-response). The RTC evaluator is used in the tutorial, so I'll add a link to that from the RTC evaluator.

src/Libraries/Microsoft.Extensions.AI.Evaluation/IEvaluator.cs
gewarren: I chose the tutorial since it includes a custom evaluator that implements IEvaluator, but I'll change it to the quickstart.

RussKie (approved):

RussKie:
Thank you!

Issue #6150, opened 2025-03-18
Chat template: The POST request does not specify which form is being submitted.


MarvinKlein1508:
### Description

Whenever I submit the text via enter I get this error message:
```
The POST request does not specify which form is being submitted. To fix this, ensure <form> elements have a @formname attribute with any unique value, or pass a FormName parameter if using <EditForm>.
```

### Reproduction Steps

I've created a blank project from scratch using th ollama provider. When I type something into the chatbox and submit it with enter I get this error message. Adding `FormName` to the `EditForm` doesn't fix this issue.

This issue is being thrown in JavaScript in `ChatInput.razor.js`
```js
elem.closest('form').dispatchEvent(new CustomEvent('submit', { bubbles: true }));
```

### Expected behavior

I expect the chat to work normally when hitting enter.

### Actual behavior

The app show the error message.

![Image](https://github.com/user-attachments/assets/86367f3c-c5fc-4466-b148-e0575995bee2)

### Known Workarounds

Removing the JS line makes it work but obviously submitting via enter isn't working anymore then.

### Configuration

Host:
  Version:      9.0.2
  Architecture: x64
  Commit:       80aa709f5d

.NET SDKs installed:
  9.0.200 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 9.0.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.36 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.14 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 9.0.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 6.0.36 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
[...]

PR #6188, opened 2025-03-24
Fix chat template input form submission on Firefox

MackinnonBuck:
The JavaScript-dispatched `'submit'` event for the chat input form was uncancellable on Firefox. This meant sending a chat message performed a full POST request that bypasses the interactive form handler, and this fails because the chat template doesn't use prerendering (and even if it did, the behavior would still be wrong).

The solution is to specify the `cancelable` property as `true` on the options object used to construct the custom `'submit'` event.

Fixes https://github.com/dotnet/extensions/issues/6150
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6188)


SteveSandersonMS (approved):
Interesting - thanks for tracking this down!

Issue #5970, opened 2025-02-25
[AI Evaluation] Include reporting for token counts and latency


shyamnamboodiripad:
The evaluation library currently includes evaluators that score AI responses for various metrics like `Coherence`, `Fluency` etc. as well as for metrics defined by custom `IEvaluator`s.

It would be great to also capture stats such as number of tokens present in the response, latency of the response and whether or not the response was fetched from cache v/s via an LLM call etc. in the report.

Additionally, it would also be a good idea to include similar cache hit, latency and token count measurements for each evaluation that is performed using an LLM.

FYI @peterwald

PR #6189, opened 2025-03-25
Support adding tags to scenarios

### LINKED ISSUE #6034: [AI Evaluation] Introduce tags / metadata concept for executions

shyamnamboodiripad:
Also includes
- Support for measuring and logging diagnostic data related to caching, latency and token usage for all LLM turns involved in each evaluation.
- Report generation changes to view and filter scenarios by the tags above, and to display the above usage statistics under each scenario in the tree.
- Changes to make some previously public types internal to avoid public API bloat.
- Other miscellaneous bug fixes and improvements in the report generation and display.

Fixes #6034 and #5970
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6189)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ChatDetails.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ChatDetails.cs
@@ -1,0 +1,70 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Extensions.AI.Evaluation.Reporting;
+
+/// <summary>
+/// A class that records details related to all LLM chat conversation turns involved in the execution of a particular
+/// <see cref="ScenarioRun"/>.
+/// </summary>
+public sealed class ChatDetails
+{
+    /// <summary>
+    /// Gets or sets the <see cref="ChatTurnDetails"/> for the LLM chat conversation turns recorded in this
+    /// <see cref="ChatDetails"/> object.
+    /// </summary>
+#pragma warning disable CA2227
+    // CA2227: Collection properties should be read only.
+    // We disable this warning because we want this type to be fully mutable for serialization purposes and for general
+    // convenience.
+    public IList<ChatTurnDetails> TurnDetails { get; set; }
+#pragma warning restore CA2227
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ChatDetails"/> class.
+    /// </summary>
+    /// <param name="turnDetails">
+    /// A list of <see cref="ChatTurnDetails"/> objects.
+    /// </param>
+    [JsonConstructor]
+    public ChatDetails(IList<ChatTurnDetails> turnDetails)
+    {
+        TurnDetails = turnDetails;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ChatDetails"/> class.
+    /// </summary>
+    /// <param name="turnDetails">
+    /// An enumeration of <see cref="ChatTurnDetails"/> objects.
+    /// </param>
+    public ChatDetails(IEnumerable<ChatTurnDetails> turnDetails)
+        : this(turnDetails.ToList())
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ChatDetails"/> class.
+    /// </summary>
+    /// <param name="turnDetails">
+    /// An array of <see cref="ChatTurnDetails"/> objects.
+    /// </param>
+    public ChatDetails(params ChatTurnDetails[] turnDetails)
+        : this(turnDetails as IEnumerable<ChatTurnDetails>)
+    {
+    }
+
+    /// <summary>
+    /// Adds <see cref="ChatTurnDetails"/> for a particular LLM chat conversation turn to the <see cref="TurnDetails"/>
+    /// collection.
+    /// </summary>
+    /// <param name="turnDetails">
+    /// The <see cref="ChatTurnDetails"/> for a particular LLM chat conversation turn.
+    /// </param>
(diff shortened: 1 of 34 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ChatDetails.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// <summary>
+    /// Adds <see cref="ChatTurnDetails"/> for a particular LLM chat conversation turn to the <see cref="TurnDetails"/>
+    /// collection.
+    /// </summary>
+    /// <param name="turnDetails">
+    /// The <see cref="ChatTurnDetails"/> for a particular LLM chat conversation turn.
+    /// </param>
+    public void AddTurnDetails(ChatTurnDetails turnDetails)
+        => TurnDetails.Add(turnDetails);
```
peterwald, line 69: Is this necessary? It is extra public API surface, and it doesn't seem to offer additional value over `chatDetails.TurnDetails.Add()`.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ReportingConfiguration.cs
```diff
@@ -168,41 +186,67 @@ public async ValueTask<ScenarioRun> CreateScenarioRunAsync(
...
+            tags = Tags;
+        }
+        else if (Tags is null)
+        {
+            tags = additionalTags;
+        }
+        else
+        {
+            tags = [.. Tags, .. additionalTags];
+        }
```
peterwald, line 207: This could be simplified to

`tags = [.. Tags ?? [], .. additionalTags ?? []];`

shyamnamboodiripad: Ah I was trying to get `tags` to be `null` when there are no tags. The change proposed above would create an empty collection instead (and at that point the `Tags` property above as well as `ScenarioRunResults.Tags` could also be non-nullable).

This is one of the tensions I was describing earlier where with collections deciding between empty v/s `null` for optional stuff can be tricky. In this case, I favored `null` (both on `ReportingConfiguration` and `ScenarioRunResult`) to avoid any impact on serialized data when tags are not being used.

`CachingKeys` above is in a similar boat - however, since we always add some default keys (from the `IChatClient` metadata), we favored a non-nullable property for that.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ChatDetails.cs
shyamnamboodiripad: That is true and is a side effect of our earlier call to keep the result types in the reporting layer completely mutable (for convenience and to simplify the JSON serialization).

From the perspective of the type and IntelliSense, I think the method is (marginally) more discoverable. We also have similar APIs in other places (like `metric.AddDiagnostics()` instead of `metric.Diagnostics.Add()`). 

I'll leave this be for now but will try and take a holistic look at all similar occurrences in a follow-up PR. We can either remove the instance methods completely or convert these to convenience extension methods that live outside the type...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ChatTurnDetails.cs
```diff
@@ -0,0 +1,82 @@
...
+    string? model = null,
+    UsageDetails? usage = null,
+    string? cacheKey = null,
+    bool? cacheHit = null)
+{
+    /// <summary>
+    /// Gets or sets the duration between the time when the request was sent to the LLM and the time when the response
+    /// was received for the chat conversation turn.
+    /// </summary>
+    public TimeSpan Latency { get; set; } = latency;
```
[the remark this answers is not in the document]
shyamnamboodiripad: A response would not be produced in the case where rate limiting happens. Theoretically, we could inspect exception objects thrown from within the (provider-specific) `IChatClient` implementations to detect this, but that can be error prone. We can revisit if MEAI adds support for some standard way to detect rate limiting across providers.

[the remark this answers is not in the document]
peterwald: > A response would not be produced in the case where rate limiting happens. 

This depends on the implementation. If the resilience layer intercepts it, then a response would be produced, but it would take a longer time for the response. I would not want to inspect exceptions. This can be handled via the HttpClient pipeline, in combination with reporting the signals out.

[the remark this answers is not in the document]
shyamnamboodiripad: > This can be handled via the HttpClient pipeline, in combination with reporting the signals out.

Agreed - although this may not always be possible to handle gracefully (for example, in cases where there is a limit on the number of tokens available to a user for a particular model each day). Regardless, I think we need this to be supported in the layers below us first...

PR #6193, opened 2025-03-25
Enable AIFunctionFactory.Create to use Activator{Utilities}.Create

stephentoub:
Adds a Create overload that takes a target type and a MethodInfo. Each invocation of the AIFunction will create/dispose of a new instance, from DI if an IServiceProvider is provided, or via new() otherwise.

@halter73, @eiriktsarpalis, @SteveSandersonMS, how do we feel about this approach? This is based on folks asking for a capability this enables in https://github.com/modelcontextprotocol/csharp-sdk.

If we want to proceed with it, I'll add tests.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6193)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.AI.Tests/Functions/AIFunctionFactoryTest.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Tests/Functions/AIFunctionFactoryTest.cs
@@ -366,21 +366,64 @@ public class AIFunctionFactoryTest
             typeof(MyFunctionTypeWithOneArg)));
     }
 
-    private sealed class MyFunctionTypeWithNoArgs
+    [Fact]
+    public async Task Create_NoInstance_DisposableInstanceCreatedDisposedEachInvocation()
     {
-        private string _value = "42";
+        AIFunction func = AIFunctionFactory.Create(
+            typeof(DisposableService).GetMethod(nameof(DisposableService.GetThis))!,
+            typeof(DisposableService),
+            new()
+            {
+                MarshalResult = (result, type, cancellationToken) => new ValueTask<object?>(result),
+            });
 
-        public static void StaticMethod() => throw new NotSupportedException();
+        var d1 = Assert.IsType<DisposableService>(await func.InvokeAsync());
+        var d2 = Assert.IsType<DisposableService>(await func.InvokeAsync());
+        Assert.NotSame(d1, d2);
 
-        public string InstanceMethod() => _value;
+        Assert.Equal(1, d1.Disposals);
+        Assert.Equal(1, d2.Disposals);
     }
 
-    private sealed class MyFunctionTypeWithOneArg(MyArgumentType arg)
+    [Fact]
+    public async Task Create_NoInstance_AsyncDisposableInstanceCreatedDisposedEachInvocation()
     {
-        public object InstanceMethod() => Tuple.Create(this, arg);
+        AIFunction func = AIFunctionFactory.Create(
+            typeof(AsyncDisposableService).GetMethod(nameof(AsyncDisposableService.GetThis))!,
+            typeof(AsyncDisposableService),
+            new()
+            {
+                MarshalResult = (result, type, cancellationToken) => new ValueTask<object?>(result),
+            });
+
+        var d1 = Assert.IsType<AsyncDisposableService>(await func.InvokeAsync());
+        var d2 = Assert.IsType<AsyncDisposableService>(await func.InvokeAsync());
+        Assert.NotSame(d1, d2);
+
+        Assert.Equal(1, d1.AsyncDisposals);
+        Assert.Equal(1, d2.AsyncDisposals);
     }
 
-    private sealed class MyArgumentType;
+    [Fact]
+    public async Task Create_NoInstance_DisposableAndAsyncDisposableInstanceCreatedDisposedEachInvocation()
+    {
+        AIFunction func = AIFunctionFactory.Create(
+            typeof(DisposableAndAsyncDisposableService).GetMethod(nameof(DisposableAndAsyncDisposableService.GetThis))!,
+            typeof(DisposableAndAsyncDisposableService),
+            new()
+            {
+                MarshalResult = (result, type, cancellationToken) => new ValueTask<object?>(result),
+            });
+
+        var d1 = Assert.IsType<DisposableAndAsyncDisposableService>(await func.InvokeAsync());
+        var d2 = Assert.IsType<DisposableAndAsyncDisposableService>(await func.InvokeAsync());
+        Assert.NotSame(d1, d2);
+
+        Assert.Equal(0, d1.Disposals);
+        Assert.Equal(0, d2.Disposals);
+        Assert.Equal(1, d1.AsyncDisposals);
+        Assert.Equal(1, d2.AsyncDisposals);
+    }
 
     [Fact]
     public async Task ConfigureParameterBinding_CanBeUsedToSupportFromKeyedServices()
@@ -696,6 +739,59 @@ public class AIFunctionFactoryTest
     {
         public int Value => value;
     }
+
+    private class DisposableService : IDisposable
+    {
+        public int Disposals { get; private set; }
+        public void Dispose() => Disposals++;
+
+        public object GetThis() => this;
+    }
+
+    private class AsyncDisposableService : IAsyncDisposable
+    {
+        public int AsyncDisposals { get; private set; }
+
+        public ValueTask DisposeAsync()
+        {
+            AsyncDisposals++;
+            return default;
+        }
+
+        public object GetThis() => this;
+    }
+
+    private class DisposableAndAsyncDisposableService : IDisposable, IAsyncDisposable
+    {
+        public int Disposals { get; private set; }
+        public int AsyncDisposals { get; private set; }
+
+        public void Dispose() => Disposals++;
+
+        public ValueTask DisposeAsync()
+        {
+            AsyncDisposals++;
+            return default;
+        }
+
+        public object GetThis() => this;
+    }
+
+    private sealed class MyFunctionTypeWithNoArgs
+    {
+        private string _value = "42";
+
+        public static void StaticMethod() => throw new NotSupportedException();
+
+        public string InstanceMethod() => _value;
+    }
+
+    private sealed class MyFunctionTypeWithOneArg(MyArgumentType arg)
+    {
+        public object InstanceMethod() => Tuple.Create(this, arg);
+    }
+
+    private sealed class MyArgumentType;
 
     private class A;
     private class B : A;


stephentoub:
Based on offline discussion with @halter73, I'm marking this as ready for review.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -112,9 +114,47 @@ public static AIFunction Create(Delegate method, string? name = null, string? de
...
+    /// marshaled to the expected type via JSON and using <paramref name="options"/>'s
+    /// <see cref="AIFunctionFactoryOptions.SerializerOptions"/>. If the argument is a <see cref="JsonElement"/>,
+    /// <see cref="JsonDocument"/>, or <see cref="JsonNode"/>, it is deserialized directly. If the argument is anything else unknown,
+    /// it is round-tripped through JSON, serializing the object as JSON and then deserializing it to the expected type.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="ArgumentNullException"><paramref name="method"/> is <see langword="null"/>.</exception>
+    public static AIFunction Create(
+        MethodInfo method,
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type targetType,
```
eiriktsarpalis, line 149: I interpret `targetType` as being inserted artificially to satisfy the linker. I wonder if there's a better incantation we can use here cc @vitek-karas @eerhardt

eiriktsarpalis: Also, why not call this "declaringType"?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -181,35 +221,104 @@ public static ReflectionAIFunction Build(MethodInfo method, object? target, AIFu
...
+            AIFunctionFactoryOptions options)
+        {
+            FunctionDescriptor = functionDescriptor;
+            TargetType = targetType;
+            AdditionalProperties = options.AdditionalProperties ?? EmptyReadOnlyDictionary<string, object?>.Instance;
+        }
+
         public ReflectionAIFunctionDescriptor FunctionDescriptor { get; }
         public object? Target { get; }
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]
```
eiriktsarpalis, line 269: I've never seen this attribute being applied to out parameters before. What is the purpose of that?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: Oh nevermind, I see now how this is being used.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -181,35 +221,104 @@ public static ReflectionAIFunction Build(MethodInfo method, object? target, AIFu
...
+            object? target = Target;
+            try
+            {
+                if (TargetType is { } targetType)
+                {
+                    Debug.Assert(target is null, "Expected target to be null when we have a non-null target type");
+                    Debug.Assert(!FunctionDescriptor.Method.IsStatic, "Expected an instance method");
 
-            for (int i = 0; i < args.Length; i++)
+                    target = arguments.Services is { } services ?
```
eiriktsarpalis, line 291: I'm curious what this entails for lifetime management of `AIFunction` instances. If one of the constructor dependencies is transient, then the `AIFunction` itself must be transient but we won't be able to rely on DI to clean it up but instead the users would have to dispose it themselves.

Maybe an E2E example might better illuminate how this is supposed to be used, but my gut tells me that an approach where AIFunctions are directly inherited from using classes that are then registered offers better mechanical sympathy to existing DI mechanisms.

stephentoub: >  If one of the constructor dependencies is transient, then the AIFunction itself must be transient

Why? This object is always new'd up fresh just for the invocation. It's effectively always transient.

That then has the same standard issue of not disposing of input from DI that might be singletons or the like, but that's nothing new.

According to @halter73, this is how MVC works as well.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: Wouldn't the constructor parameter annotation suffice in this case? I don't think the linker would be able to reason about the type of the returned property.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
eiriktsarpalis: Oh ok, I had missed this was being done on each invocation.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
stephentoub: It looks like it suffices. I think I'd added it before adding the ctor annotation.

eiriktsarpalis (approved):

stephentoub:
> It might be worth verifying that the target type gets disposed at the end of the invocation.

Done

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -181,35 +221,103 @@ public static ReflectionAIFunction Build(MethodInfo method, object? target, AIFu
...
+            AIFunctionFactoryOptions options)
+        {
+            FunctionDescriptor = functionDescriptor;
+            TargetType = targetType;
+            AdditionalProperties = options.AdditionalProperties ?? EmptyReadOnlyDictionary<string, object?>.Instance;
+        }
+
         public ReflectionAIFunctionDescriptor FunctionDescriptor { get; }
         public object? Target { get; }
+        public Type? TargetType { get; }
```
eerhardt, line 269: Why doesn't this have 

```suggestion
        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]
        public Type? TargetType { get; }
```

vitek-karas: I was just answering the previous question about this - I think we need the annotation here:
The trimmer analysis doesn't do "Cross method" analysis - it requires that the annotations are on all "input/output" to a single method. In this case the knowledge that something can be constructed through reflection come into the `Build` method, then it's transfered to the parameter of the ReflectionAIFunction method and then it's tranfered to the TargetType property which is then used by the InvokeCoreAsync method which is the place which actually uses the fact that the type can be constructed through reflection (by calling Activator.CreateInstance on it).

Because the analysis is per single method, the annotation needs to be on all the things which transfer the type value from the input (where we declare that it requires this) to the output (the code which uses the fact).

stephentoub: It did have it. @eiriktsarpalis asked me to remove it here:
https://github.com/dotnet/extensions/pull/6193#pullrequestreview-2716543497

stephentoub: > Because the analysis is per single method, the annotation needs to be on all the things which transfer the type value from the input (where we declare that it requires this) to the output (the code which uses the fact).

Does that mean there's a bug in the analyzer, as this isn't triggering a warning?

eerhardt: Looks like the AOT test caught this:

https://dev.azure.com/dnceng-public/public/_build/results?buildId=994461&view=logs&j=9db12f06-3bef-52cb-1e93-45aadd48533c&t=04f10100-7f53-527f-69e4-9793d0a279c7&l=90

```
  Generating native code
/mnt/vss/_work/1/s/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs(290): Trim analysis error IL2072: Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.<InvokeCoreAsync>d__26.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.TargetType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/mnt/vss/_work/1/s/test/Libraries/Microsoft.Extensions.AotCompatibility.TestApp/Microsoft.Extensions.AotCompatibility.TestApp.csproj]
/mnt/vss/_work/1/s/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs(290): Trim analysis error IL2072: Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.<InvokeCoreAsync>d__26.MoveNext(): 'instanceType' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider,Type,Object[])'. The return value of method 'Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.TargetType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/mnt/vss/_work/1/s/test/Libraries/Microsoft.Extensions.AotCompatibility.TestApp/Microsoft.Extensions.AotCompatibility.TestApp.csproj]
```

stephentoub: > Can you try to run it through a trimmer

We do that in CI, and it just failed.

```
/mnt/vss/_work/1/s/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs(290): Trim analysis error IL2072: Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.<InvokeCoreAsync>d__26.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.TargetType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/mnt/vss/_work/1/s/test/Libraries/Microsoft.Extensions.AotCompatibility.TestApp/Microsoft.Extensions.AotCompatibility.TestApp.csproj]
/mnt/vss/_work/1/s/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs(290): Trim analysis error IL2072: Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.<InvokeCoreAsync>d__26.MoveNext(): 'instanceType' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider,Type,Object[])'. The return value of method 'Microsoft.Extensions.AI.AIFunctionFactory.ReflectionAIFunction.TargetType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [/mnt/vss/_work/1/s/test/Libraries/Microsoft.Extensions.AotCompatibility.TestApp/Microsoft.Extensions.AotCompatibility.TestApp.csproj]
```

I'll revert the commit that removed the attribute.

We should also fix the analyzer :)

eerhardt: > why doesn't this produce any warnings during publish?

`public Repro? Create()` is getting trimmed becuase it is not getting called.

vitek-karas: Because the `Repro.Create` method is removed - nothing calls it. If you modify it like:
```csharp
Console.WriteLine(Create(typeof(HttpClient)).Create());
```

Then you get (publish with PublishTrimmed=true):
```
    F:\repro\illink\t2\Program.cs(20,13): Trim analysis warning IL2072: Repro.Create(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Activator.CreateInstance(Type)'. The return value of method 'Repro.TargetType.get' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

But the analyzer still doesn't report anything, which I think it definitely should in this case.

@sbomer - this is the problem I sent you an email about - or at least one version of it. My guess is that the analyzer doesn't understand pattern matching at all, and gives up on it. Here it's really only used to basically do a null check, but it contains a "hidden" value assignment which we probably don't track.

stephentoub: Face slap. I neglected to call Create after calling the ctor. Thanks.

stephentoub: https://github.com/dotnet/runtime/issues/113922

PR #6197, opened 2025-03-26
A couple more minor eval report fixes

shyamnamboodiripad:
* Without the background color on the header, scrolling the report was causing the header portion to become transparent when the report was scrolled.
* Also moved the filterable tags into the header per earlier feedback.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6197)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/App.tsx
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/App.tsx
@@ -24,7 +24,8 @@ const useStyles = makeStyles({
     position: 'sticky',
     top: 0,
     zIndex: 1,
-    paddingBottom: '12px'
+    paddingBottom: '12px',
+    backgroundColor: tokens.colorNeutralBackground1,
   },
   headerTop: {
     display: 'flex',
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/App.tsx
```diff
@@ -24,7 +24,8 @@ const useStyles = makeStyles({
     position: 'sticky',
     top: 0,
     zIndex: 1,
-    paddingBottom: '12px'
+    paddingBottom: '12px',
+    backgroundColor: tokens.colorNeutralBackground1,
```
peterwald, line 28: Ah, I had already fixed this in my PR. :smiley:

peterwald (approved):

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/App.tsx
shyamnamboodiripad: I needed the fix for the demo in our talk - so just wanted to make sure the fix is in the checked in sources too :)

PR #6204, opened 2025-03-26
Strip `default` keyword in generated schemas when RequireAllProperties is specified.

### LINKED ISSUE #6196: Error with MEAI 9.4.0-preview.1.25175.4
### Description

I tried out 9.4.0-preview.1.25175.4 and I got this error:

```
fail: Microsoft.Extensions.AI.LoggingChatClient[1784604714]
      GetStreamingResponseAsync failed.
      System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: invalid_function_parameters)
      Parameter: tools[3].function.parameters

      Invalid schema for function 'GetLogs': In context=('properties', 'name'), 'default' is not permitted.
         at OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
         at OpenAI.Chat.ChatClient.CompleteChatAsync(BinaryContent content, RequestOptions options)
         at OpenAI.Chat.ChatClient.<>c__DisplayClass10_0.<<CompleteChatStreamingAsync>b__0>d.MoveNext()
      --- End of stack trace from previous location ---
         at OpenAI.AsyncSseUpdateCollection`1.GetRawPagesAsync()+MoveNext()
         at OpenAI.AsyncSseUpdateCollection`1.GetRawPagesAsync()+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at System.ClientModel.AsyncCollectionResult`1.GetAsyncEnumerator(CancellationToken cancellationToken)+MoveNext()
         at System.ClientModel.AsyncCollectionResult`1.GetAsyncEnumerator(CancellationToken cancellationToken)+MoveNext()
         at System.ClientModel.AsyncCollectionResult`1.GetAsyncEnumerator(CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at Microsoft.Extensions.AI.OpenAIModelMappers.FromOpenAIStreamingChatCompletionAsync(IAsyncEnumerable`1 updates, CancellationToken cancellationToken)+MoveNext()
         at Microsoft.Extensions.AI.OpenAIModelMappers.FromOpenAIStreamingChatCompletionAsync(IAsyncEnumerable`1 updates, CancellationToken cancellationToken)+MoveNext()
         at Microsoft.Extensions.AI.OpenAIModelMappers.FromOpenAIStreamingChatCompletionAsync(IAsyncEnumerable`1 updates, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at Microsoft.Extensions.AI.LoggingChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
```

GetLogs is a function with a `name` parameter and it allows null.

```cs
[Description("Get logs.")]
private async Task<string?> GetLogs(
    [Description("The name. Optional.")]
    string? name = null,
{
}
```

I haven't made any changes to my code other than updating to 9.4.0-preview.1.25175.4.

Endpoint is `https://api.openai.com/v1`

### Reproduction Steps

Call GetStreamingResponseAsync

### Expected behavior

AI response returned

### Actual behavior

See error above.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

eiriktsarpalis:
Fix #6196
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6204)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -199,18 +199,31 @@ private static JsonNode CreateJsonSchemaCore(
                 (schemaObj = [])[SchemaPropertyName] = SchemaKeywordUri;
             }
 
-            if (description is not null)
+            if (hasDefaultValue)
             {
-                (schemaObj ??= [])[DescriptionPropertyName] = description;
+                if (inferenceOptions.RequireAllProperties)
+                {
+                    // Do not emit default values for schemas where all properties
```
stephentoub, line 206: Can we add a comment about why we're applying this heuristic?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -302,37 +323,33 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext schemaExporterContext, Js
...
             }
 
-            if (ctx.Path.IsEmpty)
+            if (ctx.Path.IsEmpty && hasDefaultValue)
             {
-                // We are at the root-level schema node, update/append parameter-specific metadata
-
-                if (!string.IsNullOrWhiteSpace(description))
+                // Add root-level default value metadata
+                if (inferenceOptions.RequireAllProperties)
```
stephentoub, line 329: Do we document RequireAllProperties as basically the OpenAI-compatible schema option?

eiriktsarpalis: We don't, although that's its sole raison d'etre is OpenAI compatibility. In hindsight we should just use a conformance mode enum instead of these ad hoc flags.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
eiriktsarpalis: The implied message is that `default` only makes sense for optional parameters, but of course that doesn't explain why we still embed it in the description.

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
```diff
@@ -259,6 +259,90 @@ public virtual async Task FunctionInvocation_AutomaticallyInvokeFunction_WithPar
         Assert.Contains("3528", sb.ToString());
     }
 
+    [ConditionalFact]
+    public virtual async Task FunctionInvocation_OptionalParameter()
```
stephentoub, line 263: Could we have a test that's along the lines of:
```C#
[ConditionalFact]
public virtual async Task AvailableTools_SchemasAreAccepted()
{
    ChatOptions options = new()
    {
        MaxOutputTokens = 100,
        Tools =
        [
            AIFunctionFactory.Create((int? i) => i, "Method1"),
            AIFunctionFactory.Create((string? s) => s, "Method2"),
            AIFunctionFactory.Create((int? i = null) => i, "Method3"),
            AIFunctionFactory.Create((bool b) => b, "Method4"),
            AIFunctionFactory.Create((double d) => d, "Method5"),
            AIFunctionFactory.Create((decimal d) => d, "Method6"),
            AIFunctionFactory.Create((float f) => f, "Method7"),
            AIFunctionFactory.Create((long l) => l, "Method8"),
            AIFunctionFactory.Create((char c) => c, "Method9"),
            AIFunctionFactory.Create((DateTime dt) => dt, "Method10"),
            AIFunctionFactory.Create((DateTime? dt) => dt, "Method11"),
            AIFunctionFactory.Create((Guid guid) => guid, "Method12"),
            AIFunctionFactory.Create((List<int> list) => list, "Method13"),
            AIFunctionFactory.Create((int[] arr) => arr, "Method14"),
            AIFunctionFactory.Create((Dictionary<string, int> dict) => dict, "Method15"),
            ... // any interesting things we can think of from a schema generation perspective
        ],
    };

    // We don't care about the response, only that we get one and that an exception isn't thrown due to unacceptable schema.
    var response = await chatClient.GetResponseAsync("Briefly, what is the most popular tower in Paris?", options);
    Assert.NotNull(response);
}
```
?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.cs
```diff
@@ -199,18 +199,31 @@ private static JsonNode CreateJsonSchemaCore(
...
-            if (description is not null)
+            if (hasDefaultValue)
             {
-                (schemaObj ??= [])[DescriptionPropertyName] = description;
+                if (inferenceOptions.RequireAllProperties)
+                {
+                    // Do not emit default values for schemas where all properties
+                    // are required and embed in the description instead.
+                    string defaultValueJson = defaultValue is not null
+                        ? JsonSerializer.Serialize(defaultValue, serializerOptions.GetTypeInfo(defaultValue.GetType()))
```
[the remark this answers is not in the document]
SteveSandersonMS: Seems like our options include:

 * Allowing this to throw if `defaultValue` isn't serializable, so developers realize they need not to do that
 * Or, omit the `"Default value: {...}"` annotation if `serializerOptions.GetTypeInfo` can't produce typeinfo, so that things work albeit without honoring the default

As for the scenario, in my opinion it doesn't make sense for developers to set any nontrivial value as the default in their schema. In just about all cases, there's no good reason to move the complexity of producing the nontrivial value into the LLM instead of having the default be `null` and then having the C# code coalesce this with some default.

So personally I'd be fine with this throwing if it can't serialize `defaultValue`, as that will just force developers to stop trying to use a nontrivial default value.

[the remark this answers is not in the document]
eiriktsarpalis: Agreed. This is also what the implementation is doing currently.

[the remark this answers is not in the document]
stephentoub: How does someone supply an arbitrary non-primitive default value?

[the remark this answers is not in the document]
SteveSandersonMS: If the only way to do it is by supplying a custom `MethodInfo`, that makes it a super extreme niche. I was commenting just based on what the API appears to permit, but it's a good point that people are not generally going to do that.

So I guess your original point is that there is a finite and in fact small set of possible default value types and we could handle them all?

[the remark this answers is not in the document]
eiriktsarpalis: No, this path is only entered via this API when the `Type` is set to null:
```
    public static JsonElement CreateJsonSchema(
        Type? type,
        string? description = null,
        bool hasDefaultValue = false,
        object? defaultValue = null,
```

Honestly, it is not clear to me why we need to keep accepting null `Type?` inputs since the schemas it does produce are rather trivial.

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
eiriktsarpalis: Sure, I can do that. It won't test if the LLM can use the schema to supply us with valid parameters though.

stephentoub: Right. I'm more interested right now in seeing whether we generate schema that it outrights fails when handed.

PR #6210, opened 2025-03-28
Added support for OTel semantic conventions for GenAI with Aspire

jmatthiesen:
This PR adds some additional meters/sources for OTel to include generative AI messages and properties, enabling the [semantic conventions for generative AI](https://opentelemetry.io/docs/specs/semconv/gen-ai/). With these updates we can more detailed traces for chat interactions (previously the different calls for AI models were not included):

image

As well as details such as message contents, token usage, etc.

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6210)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.ServiceDefaults/Extensions.cs
```diff
@@ -49,12 +49,16 @@ public static TBuilder ConfigureOpenTelemetry<TBuilder>(this TBuilder builder) w
             logging.IncludeScopes = true;
         });
 
+        // Enable the output of semantic conventions for generative AI
+        AppContext.SetSwitch("OpenAI.Experimental.EnableOpenTelemetry", true);
```
SteveSandersonMS, line 53: I suspect the changes in this file aren't required. Let's discuss separately.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/Program.Aspire.cs
```diff
@@ -27,7 +27,16 @@
...
+    .UseFunctionInvocation()
+    .UseOpenTelemetry(null, null, configure =>
+    {
+        // In development mode, enable the inclusion of sensitive data like
+        // prompt and user messages
+        if (builder.Environment.IsDevelopment())
+        {
+            configure.EnableSensitiveData = true;
+        }
+    });
```
SteveSandersonMS, line 39: It's a long Program.cs already so if we can shorten this that would be great, e.g.:

```suggestion
    .UseFunctionInvocation()
    .UseOpenTelemetry(null, null, configure =>
        configure.EnableSensitiveData = builder.Environment.IsDevelopment());
```

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.ServiceDefaults/Extensions.cs
```diff
@@ -49,12 +49,16 @@ public static TBuilder ConfigureOpenTelemetry<TBuilder>(this TBuilder builder) w
...
+
```
MackinnonBuck, line 54: Do you have an opinion about whether these changes shoule be conditional depending on whether OpenAI/Azure OpenAI/GitHub Models is used? i.e., should this be left out in the Ollama scenario? I recognize this would just no-op in the Ollama case, so maybe it's fine to leave it as-is.

MackinnonBuck (commented):
We'll also need to update the template snapshots before these changes hit `main`.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/Program.Aspire.cs
```diff
@@ -27,7 +27,9 @@
 #else // IsAzureOpenAI || IsOpenAI || IsGHModels
```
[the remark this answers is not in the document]
MackinnonBuck: It looks like opentelemetry is enabled by default in the Ollama case, so there's no need to add a call  to `UseOpenTelemetry()`:

https://github.com/CommunityToolkit/Aspire/blob/5a3c19b56a05d608ba81fc989e5ac20424b7f5b7/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaChatClientExtensions.cs#L61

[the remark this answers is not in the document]
stephentoub: > Should this middleware work fine with Ollama as well?

Yes

MackinnonBuck (approved):

Issue #5022, opened 2024-03-12
Microsoft.Extensions.Http.Resilience .NET Standard compatible


bdovaz:
Why is it not compatible? It makes it impossible for us to use it in Unity (.NET Standard 2.1 only)...

https://www.nuget.org/packages/Microsoft.Extensions.Http.Resilience#dependencies-body-tab

mike-ayres:
bump

joperezr:
Sorry for the lack of replies here. We haven't forgotten about this and we should be able to post an update here soon.

Abdragiz:
@joperezr, Is there a specific reason why many libraries in this repository don't target `netstandard2.0`? It seems logical to align with the same targeting policy used by other libraries in the `Microsoft.Extensions.*` family (e.g., `Logging`, `Configuration`, `DependencyInjection`).

bdovaz:
@joperezr I am surprised that you have never taken into account Unity which is why I opened this issue really. Unity does not support .NET Core CLR, it uses an [old mono fork](https://github.com/Unity-Technologies/mono) so it only supports .NET Standard 2.0 / 2.1 (depending Unity version):

https://docs.unity3d.com/6000.0/Documentation/Manual/dotnet-profile-support.html

All the Microsoft.Extensions.* libraries are compatible with .NET Standard 2.0 or 2.1 so we have always been able to use them without any problem, that's why I was surprised that this one in particular was not compatible...

bdovaz:
@joperezr I am encountering this problem again in Unity when I try to create tests with this package:

https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Diagnostics.Testing

Not compatible with .NET Standard 2.0 / 2.1:

https://www.nuget.org/packages/Microsoft.Extensions.Diagnostics.Testing/#dependencies-body-tab

To workaround this problem, I had to copy the source code and adapt it to C# 8 which is the Unity version:
* Remove the telemetry part (not compatible with .NET Standard 2.0 as I said)
* Remove the file scoped namespaces.
* Remove the collection expressions.

joperezr:
Thanks for sharing, you are right, I don't think we have any push back on adding NS2.0 to some of these libraries where it makes sense, so it is just mainly about prioritizing the work versus our other priorities. Since you mention that you have done at least some of this work already, would you be interested in contributing a PR with the NS support?

bdovaz:
[...]

PR #6218, opened 2025-03-31
Add netstandard2.0 compatibility to Microsoft.Extensions.Telemetry and dependencies

bdovaz:
Fixes https://github.com/dotnet/extensions/issues/5022

I have based on the format of this project to change the target frameworks:

https://github.com/dotnet/extensions/blob/56df0c4f4933909367536413dcf9c16520a37f82/src/Libraries/Microsoft.Extensions.Compliance.Abstractions/Microsoft.Extensions.Compliance.Abstractions.csproj#L4

### DIFF
--- a/src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/Microsoft.Extensions.DependencyInjection.AutoActivation.csproj
+++ b/src/Libraries/Microsoft.Extensions.DependencyInjection.AutoActivation/Microsoft.Extensions.DependencyInjection.AutoActivation.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>Microsoft.Extensions.DependencyInjection</RootNamespace>
+    <TargetFrameworks>$(NetCoreTargetFrameworks);netstandard2.0;</TargetFrameworks>
     <Description>Extensions to auto-activate registered singletons in the dependency injection system.</Description>
     <Workstream>Fundamentals</Workstream>
   </PropertyGroup>
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Microsoft.Extensions.Telemetry.Abstractions.csproj
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Microsoft.Extensions.Telemetry.Abstractions.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>Microsoft.Extensions.Telemetry</RootNamespace>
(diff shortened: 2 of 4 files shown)


RussKie:
The build is failing.

bdovaz:
@RussKie I don't understand the error if the only thing I have changed is the target frameworks. The errors are very similar to each other:

> .dotnet\sdk\9.0.104\Microsoft.Common.CurrentVersion.targets(2413,5): error MSB3277: (NETCORE_ENGINEERING_TELEMETRY=Build) Found conflicts between different versions of "Microsoft.Extensions.DependencyInjection.Abstractions" that could not be resolved.

src/Libraries/Microsoft.Extensions.AmbientMetadata.Application/Microsoft.Extensions.AmbientMetadata.Application.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>Microsoft.Extensions.AmbientMetadata</RootNamespace>
+    <TargetFrameworks>$(NetCoreTargetFrameworks);netstandard2.0;</TargetFrameworks>
```
evgenyfedorov2:
```suggestion
    <TargetFrameworks>$(NetCoreTargetFrameworks);netstandard2.0;net462</TargetFrameworks>
```

here and in other .csproj files

evgenyfedorov2:
Please generate the suppressions file with `dotnet pack <your sln file> /p:ApiCompatGenerateSuppressionFile=true` switch

src/Libraries/Microsoft.Extensions.AmbientMetadata.Application/Microsoft.Extensions.AmbientMetadata.Application.csproj
bdovaz: @evgenyfedorov2 But that's intentional because net462 supports netstandard2.0, right? https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0

evgenyfedorov2: yes, `net462` implements `nestandard2.0`, but also has something extra (as well as any other implementation such as `xamarin`, `mono`), `netstandard2.0` does not cover everything there is in `net462`

evgenyfedorov2: and if we want to align with this 
image
 we have to add net462 anyway, right?

evgenyfedorov2 (approved):

Issue #6213, opened 2025-03-31
MEAI - Handling Errors for Function Tool Calls with LLM


schaveyt:
What is the team's thoughts on handling functions that need error handling communicated back to the LLM?

```csharp
public static class FileSystemTools
{
        [Description("Request to read the contents of a text file at the specified path.")]
        public static string ReadFileAllText(
            [Description("The path of the file to read (relative to the current working directory)")]
            string path)
        {
            var filePath = Path.Combine(Directory.GetCurrentDirectory(), path);
            if (File.Exists(filePath))
            {
                return File.ReadAllText(filePath);
            }
            
            // 🆘 What should I do here? How can I report to the LLM there was an error.
            //
            return "ERROR: BAD READ"; // 🤷‍♂️
        }
}
```

SteveSandersonMS:
For methods that return a string, you can handle it exactly as you have done here.

I know it feels wrong to mix together different meanings for the return value (e.g., what if the file literally contained the text `ERROR: BAD READ` and this was not an error?) but as far as I'm aware, typical LLM backend services don't have any way to differentiate successful outputs from unsuccessful ones. For example, see [OpenAI's API docs](https://platform.openai.com/docs/api-reference/runs/submitToolOutputs) - there's only a single way to specify a result value, and it doesn't differentiate "successful" or "unsuccessful" ones. If anyone is aware that any AI service backends do differentiate this, please post a link.

For methods that return other types, it's technically possible to throw an exception and then Microsoft.Extensions.AI's function calling middleware will catch this and will build an error result. However I appreciate it's not good to have to use an exception for that. Right now your best bet is to make sure you're using a string-returning function, and if that feels wrong, remember that everything's going to be reduced to a string on the wire anyway so no fidelity is actually being lost.

@stephentoub Do you think we should consider adding some API like `string? FunctionInvocationContext.FailureDescription { get; set; }`? If nonnull, we'd ignore the delegate's return value and set the FRC value to something like `CALL FAILED: {description}`. This would just be so that people can emit failure descriptions from methods that return non-string values.

stephentoub:
> However I appreciate it's not good to have to use an exception for that.

Because? I get not wanting to use exceptions for normal control flow, but this case would be for errors, right?

SteveSandersonMS:
[...]

PR #6221, opened 2025-03-31
Add AIFunctionException intended to send exception messages to the model

jozkee:
Fixes https://github.com/dotnet/extensions/issues/6213.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6221)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
@@ -1,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Functions;
+
+/// <summary>
+/// Defines an exception raised in an <see cref="AIFunction"/> invocation containing a message intended for the model detailing the reason for the failure.
+/// </summary>
+#pragma warning disable CA1032 // Implement standard exception constructors
+public sealed class AIFunctionException : Exception
+#pragma warning restore CA1032 // Implement standard exception constructors
+{
+    /// <summary>
+    /// Gets the message intended for the model.
+    /// </summary>
+    public string MessageForModel { get; private init; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AIFunctionException"/> class.
+    /// </summary>
+    /// <param name="messageForModel">The message intended for the model.</param>
+    /// <param name="message">The message that can contain information not intended to share with the model.</param>
+    public AIFunctionException(string messageForModel, string? message = null)
+        : base(GetMessageForBase(messageForModel, message))
+    {
+        if (messageForModel is null)
+        {
+            Throw.ArgumentNullException(nameof(messageForModel));
+        }
+
+        MessageForModel = messageForModel;
+    }
+
+    private static string GetMessageForBase(string messageForModel, string? message)
+    {
+        if (message is null)
+        {
+            return messageForModel;
+        }
+
+        return message + " Message for model: " + messageForModel;
+    }
+}
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -10,6 +10,7 @@ using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.AI.Functions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Shared.Diagnostics;
@@ -716,7 +717,8 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
                 string message = result.Status switch
                 {
                     FunctionInvocationStatus.NotFound => $"Error: Requested function \"{result.CallContent.Name}\" not found.",
-                    FunctionInvocationStatus.Exception => "Error: Function failed.",
+                    FunctionInvocationStatus.Exception when result.Exception is AIFunctionException aife => aife.MessageForModel,
+                    FunctionInvocationStatus.Exception when result.Exception is not AIFunctionException => "Error: Function failed.",
                     _ => "Error: Unknown error.",
                 };
 
--- a/test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Microsoft.Extensions.AI.Functions;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
@@ -27,6 +28,7 @@ namespace Microsoft.Extensions.AI;
 
 public abstract class ChatClientIntegrationTests : IDisposable
 {
+    private const string FunctionExceptionMessageForModel = "The tool is disabled, please advice user to contact admin@contoso.com for assistance.";
     private readonly IChatClient? _chatClient;
 
     protected ChatClientIntegrationTests()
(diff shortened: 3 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
```diff
@@ -0,0 +1,46 @@
...
+/// Defines an exception raised in an <see cref="AIFunction"/> invocation containing a message intended for the model detailing the reason for the failure.
+/// </summary>
+#pragma warning disable CA1032 // Implement standard exception constructors
+public sealed class AIFunctionException : Exception
+#pragma warning restore CA1032 // Implement standard exception constructors
+{
+    /// <summary>
+    /// Gets the message intended for the model.
+    /// </summary>
+    public string MessageForModel { get; private init; }
```
stephentoub, line 19:
```suggestion
    public string MessageForModel { get; }
```

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AIFunctionException"/> class.
+    /// </summary>
+    /// <param name="messageForModel">The message intended for the model.</param>
+    /// <param name="message">The message that can contain information not intended to share with the model.</param>
+    public AIFunctionException(string messageForModel, string? message = null)
```
stephentoub, line 26: - There should be a constructor that accepts an inner exception as well.
- I don't think the message needs to be optional. It should probably be first, as well, so that it's consistent with other exception types.
- messageForModel should be allowed to be null, with consumers only special-casing the exception if it's AIFunctionException and contains a non-null messageForModel
- Maybe it should be messageForAI instead of messageForModel?

eiriktsarpalis: > There should be a constructor that accepts an inner exception as well.

I take it the inner exception wouldn't be surfaced to the LLM right, it would just be metadata for the stacktrace?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
eiriktsarpalis, line 19: Not a huge fan of preserving separate `Message` and `MessageForModel` properties. Couldn't we just document that `Message` will always be forwarded to the LLM?

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Functions;
+
+/// <summary>
+/// Defines an exception raised in an <see cref="AIFunction"/> invocation containing a message intended for the model detailing the reason for the failure.
+/// </summary>
+#pragma warning disable CA1032 // Implement standard exception constructors
+public sealed class AIFunctionException : Exception
```
eiriktsarpalis, line 13: With exception types it is generally a good idea to keep them unsealed.
```suggestion
public class AIFunctionException : Exception
```

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
SteveSandersonMS: @jozkee's example from the test is a good illustration - it contains the phrase "*Please advise the user to email admin@contoso.com*" which is almost certainly not something the app developer would want to have show up in their logs. In general `MessageForModel` is an LLM prompt not a .NET exception message; it serves a different purpose and will contain different information.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
SteveSandersonMS: All of @stephentoub's suggestions sound good to me.

Maybe just add "If messageForModel is nonnull, it must be nonempty" to avoid a really weird situation.

Agreed that `messageForAI` is a better name.

> I take it the inner exception wouldn't be surfaced to the LLM right (e.g. in the theoretical case where both outer and inner are IAFunctionExceptions), it would just be metadata for the stacktrace?

Agreed

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    private static string GetMessageForBase(string messageForModel, string? message)
+    {
+        if (message is null)
+        {
+            return messageForModel;
+        }
+
+        return message + " Message for model: " + messageForModel;
```
SteveSandersonMS, line 44: I don't think we need to append the messageForModel here. That might be a really long string since it's a form of LLM prompt (I could imagine people putting multiple paragraphs of explanation about how the LLM should attempt different solutions).

As long as we have the regular `Message` property this can serve its normal purpose as the .NET exception message for the application developer.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
eiriktsarpalis: > which is almost certainly not something the app developer would want to have show up in their logs.

Given that this is meant to be intercepted by the function invoking middleware, it showing up in the logs sounds like a bug. I suppose it might be possible if the AIFunction itself has set up exception logging, but to me at least it feels like we're inserting clunkiness just to cover for a niche scenario.

If we must keep it like this, I would recommend dropping the term "message" in favor of something else like "response", "instruction", "prompt", etc.

SteveSandersonMS: > Given that this is meant to be intercepted by the function invoking middleware, it showing up in the logs sounds like a bug.

That's a good point, and is an issue with use of exceptions for this in general.

If I'm reading the code correctly, `FunctionInvokingChatClient.InvokeFunctionAsync` currently logs all AIFunction exceptions at `Error` level. That ceases to be correct if those exceptions can now refer to user-initiated problems. At least under ASP.NET conventions, Error level should not be used for anything that an end user can trigger on purpose (assuming there are no errors in app code). For AIFunction it's extremely believable than an end user could intentionally prompt the LLM into making a function call with invalid params.

I think what we should do here is:

 * In `FunctionInvokingChatClient.InvokeFunctionAsync`, special-case `AIFunctionException` so that it produces log entries at the `Information` level. By throwing this type of exception, the developer is not signalling an error in application code, but rather signalling an anticipated scenario with potential for recovery.
 * ... but for any other type of exception, continue to log at `Error` level since by default we assume it to be a failing of application code (and if not, it's up to the developer to catch and handle in some way)

It's not a perfect situation since "this failure is the end user's fault" and "I want to explain to the LLM how to recover" are not 100% overlapping in the venn diagram of possibility but it's a better approximation than just using either `Information` or `Error` levels in all cases.

The alternative would be something like not using an exception to signal this at all, so we could define error handling semantics for AIFunction as being "they should never throw unless there's a defect in your code that we can log at Error level".

eiriktsarpalis: Fully agree with those points. This feature is definitely pulling us into exceptions-as-control-flow territory and it might make better sense to encourage users to model this explicitly. We could for example define a `Result<T>` type similar to what can be found in Rust or F# which defines a discriminated union between a success value and an error value. The function invoking middleware could recognize that and format it in an LLM-friendly manner.

eiriktsarpalis: One possible shape might be the following:
```C#
public readonly struct AIFunctionResult<TSuccess>
{
      public AIFunctionResult(TSuccess value) { Value = value; }
      public AIFunctionResult(string errorMessage) { ErrorMessage = Throw.IfNull(errorMessage); }

      public bool IsSuccess => ErrorMessage is null;
      public bool IsError => ErrorMessage is not null;

      public TSuccess? Value { get; }
      public string? ErrorMessage { get; }
}
```

stephentoub: How is it exceptions as control flow? Functions are defined to return a T, and because of an error, they're unable to. The only reason this feature exists is because of concerns about leaking details to the llm, otherwise this would be a regular exception as in any other regular method. I'm still not convinced it needs to exist, but if we're simply looking for a way to convey to the llm "actually, it's ok, you can share this one", then modeling it exactly the same as it would have been otherwise makes the most sense to me. 

We need to keep in mind that FICC is not the only consumer here. We need to avoid putting additional concept count onto other AIFunction consumers. Having it be an exception you can treat like any other exception but opt in to pulling additional info from it is a good way to do that. Changing the return type forces every consumer to be aware of how to deal with that.

eiriktsarpalis: > How is it exceptions as control flow?

The sole purpose of the exception type is to signal to the invoking middleware that it needs to emit a particular message in lieu of a result. I don't necessarily find that objectionable, but the fact that we need to reconcile two distinct modes (a stack unwinding exception vs. passing a string to the middleware) makes me want to hold my nose.

SteveSandersonMS: > I'm still not convinced it needs to exist

What would be your preferred alternative?

stephentoub: > The sole purpose of the exception type is to signal to the invoking middleware that it needs to emit a particular message in lieu of a result.

Not in lieu of a result. There was already going to be an exception, but that exception's message wouldn't be used. This one's will. The purpose from my perspective is to say that the exception's info should be allowed to be used in cases where an exception's normally isn't. It's not changing control flow. There would already be an exception; it's effectively just marking that exception to say you can use it when otherwise safety restrictions would say to substitute a different message. My expectation is the most common case will be wrapping the whole body of a function with try/catch that then propogates an instance of this type instead. A consumer can then choose to get info from one place in the exception vs another.

> What would be your preferred alternative?

A developer can change the return type of their function to anything they want, and include whatever they want to be sent back to the llm as the answer. It could be object, Json Element, string, whatever.

eiriktsarpalis: Anyway, I won't die on this hill. It seems to solve a real problem, but I would really prefer it if we collapsed the two messages into one for the sake of simplicity.

stephentoub: > Isn't that why we have two separate message fields in the type?

No. It propagates exactly the same. It's just when certain consumers catch it and go to extract information from it, they may choose to extract MessageForAI instead of Message (or instead of ignoring Message).

stephentoub: > the truth is this does reflect the reality of how the result gets delivered to the LLM.

Yup.

> What do you think?

My preference is to leave well enough alone and close this. If we discover we need to do something more in the future, it's additive.

SteveSandersonMS: @jozkee Sorry for the misdirection on this one. Looks like your PR has helped us understand this isn't the right path. OK to close?

RogerBarreto:
Should we consider:
1. Having a reference to the actual `AIFunction` that triggered the exception, 
2. or the `JsonSchema` of the function.
3. or the `AIToolException` with the `FunctionName` + `Description`.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionException.cs
jozkee: No worries, glad it helped.

Issue #6057, opened 2025-03-09
Standardize several additional forms of contetn


stephentoub:
It's possible for implementations to create their own custom AIContent-derived types, and that's valuable when there's something the service consumes or produces that's non-standard. However, there are a variety of kinds of content that are becoming more standard and that could be abstracted over if the right types were in the core abstractions library:
- Code content. We might be able to represent this as TextContent if we tagged TextContent with a mime type.
- Code execution results. When a code interpreter tool is used, some services send back a piece of content containing the result. We might want a Code Result Content, or a ServerToolResultContent. Or maybe we add a ToolResultContent that's a base for Function Result Content.
- Reasoning content. https://github.com/dotnet/extensions/issues/5977. This could be an extra property on AI Content, it could be a wrapper content type, it could be derived from TextContent if we think it'll only ever be text, etc. 
- Citation / reference content. https://github.com/dotnet/extensions/issues/5527
-....

We need to do a pass through all of the currently available services and determine what is common enough (and/or simple enough with little need for future variation) now to warrant standardization.

cc: @luisquintanilla, @SteveSandersonMS

jeffhandley:
@RogerBarreto Please take a look and report during Wednesday's sync on if we need this ahead of GA

RogerBarreto:
I would also like to add that similar to `CodeContent` we can have `HTMLContent`, `Markdown`, `Mermaid` and so on, which are pretty much accepted standards out there. 

> We might be able to represent this as TextContent if we tagged TextContent with a mime type

I think that would make sense if the plan is a `TextContent` to be a specialization of `DataContent` or move the `MediaType` up in the stack to the `AiContent`.

If not we can have `Format` or other name that align with text subtypes.

I would add the `ErrorContent` as part of this task, as I brought it up in the `ISpeechToText` impl.

stephentoub:
> I would also like to add that similar to CodeContent we can have HTMLContent, Markdown, Mermaid and so on, which are pretty much accepted standards out there.

Couldn't DataContent just be used, with a mime type indicating which it is?

>> We might be able to represent this as TextContent if we tagged TextContent with a mime type

> I think that would make sense if the plan is a TextContent to be a specialization of DataContent or move the MediaType up in the stack to the AiContent.

We seem to be landing in place where TextContent is a special-case for the super, super common case of human-readable text (having a mime type would muddy the waters), and all other content is DataContent.

stephentoub:
[...]

PR #6222, opened 2025-04-01
Add TextReasoningContent

stephentoub:
Contributes to https://github.com/dotnet/extensions/issues/6057
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6222)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/TextReasoningContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/TextReasoningContent.cs
@@ -1,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents text reasoning content in a chat.
+/// </summary>
+/// <remarks>
+/// <see cref="TextReasoningContent"/> is distinct from <see cref="TextContent"/>. <see cref="TextReasoningContent"/>
+/// represents "thinking" or "reasoning" performed by the model and is distinct from the actual output text from
+/// the model, which is represented by <see cref="TextContent"/>. Neither types derives from the other.
+/// </remarks>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+public sealed class TextReasoningContent : AIContent
+{
+    private string? _text;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TextReasoningContent"/> class.
+    /// </summary>
+    /// <param name="text">The text reasoning content.</param>
+    public TextReasoningContent(string? text)
+    {
+        _text = text;
+    }
+
+    /// <summary>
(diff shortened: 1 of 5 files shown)


stephentoub:
@tghamm, IIRC, your anthropic library has two custom types for reasoning. Am I correct that this will enable you to replace one or both of them? I think one was for redacted content and the other not, so at a minimum I'd hope this could be used in place of the non-redacted one?

stephentoub:
> Looks good.
> 
> Do you think there should be reasoningtext equivalents for the cases where we concat `TextContent`:
> 
> * `ChatMessage.TextReasoning` (or maybe better named `ChatMessage.ReasoningText` as that's what people actually say)
> * `ChatResponseExtensions.ToChatResponse[Async]` coalescing consecutive `TextReasoningContent` into a single `TextReasoningContent`
> 
> Of these, I'm more interested in the second one, since otherwise streamed output will produce a huge amount of overhead when added to conversation history, having separate content objects for every chunk of reasoning.

I will hold off on the first; I'd like us to get more experience with this first. I'll do the second.

tghamm:
@stephentoub You are correct about the two types, and I THINK using the `AdditionalProperties` dict would allow for storing the necessary `Signature` field and perhaps a delineator between Redacted/Normal Thinking Text.  

So In short, yes, I believe this would allow for the replacement of at least one, if not both types I had to create.

stephentoub:
Personally, I prefer "reasoning" over "thinking". These models are generally referred to as reasoning, as this content represents the performed reasoning. @SteveSandersonMS, naming opinion?

SteveSandersonMS:
Agreed.

Abstractly, "thinking" and "reasoning" are close enough to synonymous that I don't think there's any first-principles reason to say one is more accurate than the other.

Practically, "reasoning" is the word most commonly used for this type of LLM behavior, so I'd side with that. I know there are examples of people referring to "thinking models" and "thinking tokens" but it seems a lot less common than "reasoning models" and "reasoning tokens".

RogerBarreto:
Thanks for the input, brought this more by experience using the Ollama where they use the first token sharing the `Assistant.Content` within `<think/>` xml node.

image

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/TextReasoningContent.cs
```diff
@@ -0,0 +1,46 @@
...
+/// <summary>
+/// Represents text reasoning content in a chat.
+/// </summary>
+/// <remarks>
+/// <see cref="TextReasoningContent"/> is distinct from <see cref="TextContent"/>. <see cref="TextReasoningContent"/>
+/// represents "thinking" or "reasoning" performed by the model and is distinct from the actual output text from
+/// the model, which is represented by <see cref="TextContent"/>. Neither types derives from the other.
+/// </remarks>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+public sealed class TextReasoningContent : AIContent
```
eiriktsarpalis, line 18: Given that you're going for a bespoke derived type over it being a boolean on the base, I take it we believe reasoning is always going to be in text form?

eiriktsarpalis (approved):

RogerBarreto (approved):

PR #6230, opened 2025-04-02
Lower M.E.AI dependencies on .NET 8 and lower back to 8.0.x

stephentoub:
We had been relying on System.Text.Json 9.0.x. This removes that override to only use 8.x versions of packages until building for .NET 9.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6230)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
@@ -20,7 +20,12 @@ namespace Microsoft.Extensions.AI;
 /// </remarks>
 public class ChatResponse<T> : ChatResponse
 {
-    private static readonly JsonReaderOptions _allowMultipleValuesJsonReaderOptions = new() { AllowMultipleValues = true };
+    private static readonly JsonReaderOptions _allowMultipleValuesJsonReaderOptions = new()
+    {
+#if NET9_0_OR_GREATER
+        AllowMultipleValues = true
+#endif
+    };
     private readonly JsonSerializerOptions _serializerOptions;
 
     private T? _deserializedResult;
--- a/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/AssertExtensions.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/AssertExtensions.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json;
+using System.Text.Json.Nodes;
 using Xunit;
 using Xunit.Sdk;
 
@@ -57,7 +58,9 @@ internal static class AssertExtensions
         options ??= JsonSerializerOptions.Default;
         JsonElement expectedElement = NormalizeToElement(expected, options);
         JsonElement actualElement = NormalizeToElement(actual, options);
-        if (!JsonElement.DeepEquals(expectedElement, actualElement))
+        if (!JsonNode.DeepEquals(
+            JsonSerializer.SerializeToNode(expectedElement),
+            JsonSerializer.SerializeToNode(actualElement)))
         {
             string message = propertyName is null
                 ? $"Function result does not match expected JSON.\r\nExpected: {expectedElement.GetRawText()}\r\nActual:   {actualElement.GetRawText()}"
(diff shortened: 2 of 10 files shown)


test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/AssertExtensions.cs
```diff
@@ -57,7 +58,9 @@ private static void AreJsonEquivalentValues(object? expected, object? actual, Js
         options ??= JsonSerializerOptions.Default;
         JsonElement expectedElement = NormalizeToElement(expected, options);
         JsonElement actualElement = NormalizeToElement(actual, options);
-        if (!JsonElement.DeepEquals(expectedElement, actualElement))
+        if (!JsonNode.DeepEquals(
```
eiriktsarpalis, line 61: Note that the pre-.NET 9 version of that implementation is very rough and can result in false negative comparisons.

eiriktsarpalis: I think we could still up the STJ version to v9 for net8.0 tests without losing coverage.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
```diff
@@ -20,7 +20,12 @@ namespace Microsoft.Extensions.AI;
 /// </remarks>
 public class ChatResponse<T> : ChatResponse
 {
-    private static readonly JsonReaderOptions _allowMultipleValuesJsonReaderOptions = new() { AllowMultipleValues = true };
+    private static readonly JsonReaderOptions _allowMultipleValuesJsonReaderOptions = new()
+    {
+#if NET9_0_OR_GREATER
+        AllowMultipleValues = true
```
eiriktsarpalis, line 26: If we know that this is no longer necessary, why only remove it for net8.0 targets?

stephentoub: Because it seems like it can only help.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/AssertExtensions.cs
stephentoub: I actually see this as a benefit. All our testing was being done against STJ v9, so if there were any STJ v8 bugs that the implementation might hit, our testing on .NET 8 and .NET Framework wouldn't reveal it.

stephentoub: > Note that the pre-.NET 9 version of that implementation is very rough and can result in false negative comparisons.

We run all tests on .NET Framework, .NET 8, and .NET 9. We'll get coverage with both STJ v8 on the former two and STJ v9 on the latter.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
SteveSandersonMS: We don't know for sure that it's no longer necessary. There are still reports saying this affects GPT-4, for example: https://community.openai.com/t/gpt-4-in-json-mode-intermittently-returning-answer-twice/606104/3

It might be much less important than it used to be assuming that native structured output (with a JSON schema) avoids the issue. So arguably we could drop this workaround but if it doesn't cost us much to leave it in, that seems fine too.

eiriktsarpalis: If removing it is creating problems, putting it on specific targets will make more difficult to diagnose.

SteveSandersonMS (approved):

RogerBarreto:
Bringing this library to .Net 8, is a game changer for LTS based apps, thanks for doing it.

This going forward will revaluate the need for specific FunctionCalling impl on Semantic Kernel side.

RogerBarreto (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
eiriktsarpalis: It's common for newer versions to improve performance, just not intentionally baking in differences in functional behavior. I won't die on this hill though.

stephentoub: We fix bugs in new versions all the time. That is baking in differences in functional behavior.

PR #6231, opened 2025-04-02
Use ErrorContent in OpenAIResponseChatClient

stephentoub:
Also tweak ErrorContent a bit.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6231)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/ErrorContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/ErrorContent.cs
@@ -3,7 +3,6 @@
 
 using System.Diagnostics;
 using System.Text.Json.Serialization;
-using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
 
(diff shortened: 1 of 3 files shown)


RogerBarreto (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/ErrorContent.cs
```diff
@@ -18,31 +17,34 @@ public class ErrorContent : AIContent
...
+        // information about an error. So, rather than throwing if message is null, we just normalize to empty.
+        _message = message ?? string.Empty;
     }
 
     /// <summary>Gets or sets the error message.</summary>
     public string Message
     {
         get => _message;
-        set => _message = Throw.IfNull(value);
+        set => _message = value ?? string.Empty;
```
eiriktsarpalis, line 35: Should we add [AllowNull] to emphasize the change?

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/ErrorContent.cs
stephentoub: If we want to do that, we'd also want to add `?` to the ctor. I don't have a strong preference either way. TextContent explicitly allows nulls, so I guess it'd make sense to be consistent and do the same here.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -246,6 +251,24 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
...
+                        CreatedAt = createdAt,
+                        MessageId = lastMessageId,
+                        ModelId = modelId,
+                        ResponseId = responseId,
+                        Contents =
+                        [
+                            new ErrorContent(errorUpdate.Message)
+                            {
+                                ErrorCode = errorUpdate.Code,
+                                Details = errorUpdate.Param,
```
SteveSandersonMS, line 267: It's odd that streaming responses have `Details` for errors but nonstreaming doesn't. I expect that's just the reality of the OpenAI client rather than an oversight in this PR, but just mentioning it in case.

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: Yeah, I double-checked that one, but for whatever reason, the error on the Response object lacks param (https://platform.openai.com/docs/api-reference/responses/object) and the corresponding streaming object has param (https://platform.openai.com/docs/api-reference/responses-streaming/error).

PR #6236, opened 2025-04-03
Minimal change in AIJsonUtilities from SHA256 to SHA384

SteveSandersonMS:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6236)

This is a minimal change. If we want to consider changing slightly more, also see https://github.com/dotnet/extensions/pull/6237

PR #6237, opened 2025-04-03
Use SHA384 and make key more explicit

SteveSandersonMS:
This PR includes the changes from https://github.com/dotnet/extensions/pull/6236 and also changes the `GetCacheKey` method signature so that it explicitly takes `messages`/`options` params:

```diff
-    protected abstract string GetCacheKey(params ReadOnlySpan<object?> values);
+    protected abstract string GetCacheKey(IEnumerable<ChatMessage> messages, ChatOptions? options, params ReadOnlySpan<object?> additionalValues);
```

We don't strictly have to do this. However it can be beneficial for people who want to customize the cache key computation to include other data (e.g., from `AdditionalProperties` from within messages or options). Previously they would have to just somehow know that one of the incoming `values` will be the data they need, and would need to do runtime type checking to find the relevant one, and would have to hope that we don't change the incoming data shape in the future (e.g., pre-JSON-serializing the values).

The more explicit signature in this PR would be a guarantee that we will in fact pass in these specific `IEnumerable<ChatMessage>`/`ChatOptions?` types. I don't think there's much drawback to us offering that guarantee, since this is really baked into the `IChatClient` contract anyway - those are exactly the data items that middleware has access to.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6237)

## Hash algorithm

I considered also making the hash algorithm configurable (defaulting to SHA384). However that complicates things because currently we use a thread-static `IncrementalHashStream` that holds the `IncrementalHash` instance. We'd have to create new instances when different hash algorithms are specified.

Since we've been told that in this case, SHA384 is sufficient without being configurable, I'd be happy to leave it at that.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.cs
@@ -90,7 +90,7 @@ public static partial class AIJsonUtilities
 
         // For cases where the hash may be used as a cache key, we rely on collision resistance for security purposes.
         // If a collision occurs, we'd serve a cached LLM response for a potentially unrelated prompt, leading to information
-        // disclosure. Use of SHA256 is an implementation detail and can be easily swapped in the future if needed, albeit
+        // disclosure. Use of SHA384 is an implementation detail and can be easily swapped in the future if needed, albeit
         // invalidating any existing cache entries.
 #if NET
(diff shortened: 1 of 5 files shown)


stephentoub (approved):

stephentoub:
@shyamnamboodiripad, @peterwald, fyi

peterwald (approved):

PR #6238, opened 2025-04-03
Add default value for namedArg in Microsoft.Gen.Metrics.parser

BowenYang666:
Add default value for namedArg in Microsoft.Gen.Metrics.parser


dariusclay:
@BowenYang666 can you describe the situation where this change fixes an issue? I'm not against the change but would like more details in the PR.

BowenYang666:
> @BowenYang666 can you describe the situation where this change fixes an issue? I'm not against the change but would like more details in the PR.
@dariusclay 
This is to solve the build error: 'Use of unassigned local variable namedArg'. I saw this error when I open the project via visual studio. But seems building through build.cmd doesn't have this issue.

Besides that, it's my first time contributing this repo. So I'm finding some simple code that I can fix to get started.

evgenyfedorov2 (approved):

PR #6239, opened 2025-04-03
In telemetry, treat AdditionalProperties as sensitive

SteveSandersonMS:
It would be nice if `IChatClient` implementations had some way to signal which parts of `AdditionalProperties` are sensitive and which aren't, but without some means to do that we're limited to treating it all as potentially sensitive.

If this later becomes problematic, we can design some way for implementations to communicate this info. For example, we could define a `"telemetry"` entry in `AdditionalProperties` that is documented as always included in telemetry output, and then it would be up to `IChatClient` implementations to put the right data there.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6239)


stephentoub:
> we can design some way for implementations to communicate this info

Maybe some of the functionality from the M.E redaction library would be useful.

PR #6241, opened 2025-04-03
Disable STJ default reflection and fix a number of failing tests.

eiriktsarpalis:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6241)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
@@ -42,30 +42,18 @@ public static partial class AIJsonUtilities
     [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026", Justification = "DefaultJsonTypeInfoResolver is only used when reflection-based serialization is enabled")]
     private static JsonSerializerOptions CreateDefaultOptions()
     {
-        // If reflection-based serialization is enabled by default, use it, as it's the most permissive in terms of what it can serialize,
-        // and we want to be flexible in terms of what can be put into the various collections in the object model.
-        // Otherwise, use the source-generated options to enable trimming and Native AOT.
-        JsonSerializerOptions options;
+        // Copy configuration from the source generated context.
+        JsonSerializerOptions options = new(JsonContext.Default.Options)
+        {
+            Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+        };
 
         if (JsonSerializer.IsReflectionEnabledByDefault)
         {
-            // Keep in sync with the JsonSourceGenerationOptions attribute on JsonContext below.
-            options = new(JsonSerializerDefaults.Web)
-            {
-                TypeInfoResolver = new DefaultJsonTypeInfoResolver(),
-                Converters = { new JsonStringEnumConverter() },
-                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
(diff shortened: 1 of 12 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
```diff
@@ -42,30 +42,18 @@ public static partial class AIJsonUtilities
     [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026", Justification = "DefaultJsonTypeInfoResolver is only used when reflection-based serialization is enabled")]
     private static JsonSerializerOptions CreateDefaultOptions()
     {
-        // If reflection-based serialization is enabled by default, use it, as it's the most permissive in terms of what it can serialize,
-        // and we want to be flexible in terms of what can be put into the various collections in the object model.
-        // Otherwise, use the source-generated options to enable trimming and Native AOT.
-        JsonSerializerOptions options;
+        // Copy configuration from the source generated context.
```
eiriktsarpalis, line 45: This is not fixing any bugs, just simplifying and removing a bit of duplication in configuration. It also prefers source generated models for known types even in the case where reflection is enabled, which should theoretically result in better startup times.

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -487,9 +487,7 @@ static bool IsAsyncMethod(MethodInfo method)
                 Throw.ArgumentException(nameof(parameter), "Parameter is missing a name.");
             }
 
-            // Resolve the contract used to marshal the value from JSON -- can throw if not supported or not found.
             Type parameterType = parameter.ParameterType;
-            JsonTypeInfo typeInfo = serializerOptions.GetTypeInfo(parameterType);
```
eiriktsarpalis, line 492: Premature calculation of the JTI was causing errors because no contract could be resolved for `CancellationToken` or parameters using custom marshallers.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Microsoft.Extensions.AI.Abstractions.Tests.csproj
```diff
@@ -10,6 +10,10 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
```
eiriktsarpalis, line 13: The `JsonSerializerIsReflectionEnabledByDefault` feature switch does not work for .NET framework targets, which I found surprising. The explicit condition doubles down on this quirk so that we can get test coverage for reflection-based serialization as well.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
```diff
@@ -53,6 +53,18 @@ public static void DefaultOptions_UsesExpectedEscaping(string input, string expe
         Assert.Equal($@"""{expectedJsonString}""", json);
     }
 
+    [Fact]
+    public static void DefaultOptions_UsesReflectionWhenDefault()
```
stephentoub, line 57: Ah, looks like this is exactly what I was asking for.

stephentoub (approved):

PR #6250, opened 2025-04-04
Remove use of ConfigureAwait from Microsoft.Extensions.AI.dll for AIFunction invocations

stephentoub:
We try to use ConfigureAwait(false) throughout our libraries. However, we exempt ourselves from that in cases where user code is expected to be called back from within the async code, and there's a reasonable presumption that such code might care about the synchronization context. AIFunction fits that bill. And FunctionInvokingChatClient needs to invoke such functions, which means that we need to be able to successfully flow the context from where user code calls Get{Streaming}ResponseAsync through into wherever a FunctionInvokingChatClient is in the middleware pipeline. We could try to selectively avoid ConfigureAwait(false) on the path through middleware that could result in calls to FICC.Get{Streaming}ResponseAsync, but that's fairly brittle and hard to maintain. Instead, this PR just removes ConfigureAwait use from the M.E.AI library. It also fixes a few places where tasks were explicitly being created and queued to the thread pool.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousDelegatingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/AnonymousDelegatingChatClient.cs
@@ -4,7 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+#if !NET9_0_OR_GREATER
 using System.Runtime.CompilerServices;
+#endif
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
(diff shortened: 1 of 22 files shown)


eiriktsarpalis:
> However, we exempt ourselves from that in cases where user code is expected to be called back from within the async code, and there's a reasonable presumption that such code might care about the synchronization context.

I think I need to better understand this exemption and its implications. For example, one could argue that `JsonSerializer.SerializeAsync` does call back to user code by virtue of it accepting registrations of custom converters. That logic could be extended to pretty much any async component accepting abstraction.

stephentoub:
It's the "and there's a reasonable presumption that such code might care about the synchronization context" part.

eiriktsarpalis:
And what decides that presumption I guess is the fact that both here and in S.L.AsyncEnumerable is the fact that we're unambiguously passing a single delegate for execution?

stephentoub:
Yes. It's subjective, how likely is it that it'll really be an issue, that someone has a reasonable assumption their code will be invoked in a particular context.

eiriktsarpalis (approved):

Issue #6247, opened 2025-04-04
MEAI docs: Function calling


SteveSandersonMS:
An outcome from the TM review.

For function calling docs:

 * Emphasize that incoming args are untrusted data from the LLM
 * The exceptions are the parameter types special-cased by `AIFunctionFactory`:
    - `CancellationToken` is special and will receive a token that is chained onto the CT on the original call
    - `AIFunctionArgument` is special and describes all the incoming params as well as other context info (`IServiceProvider`)
    - `IServiceProvider` is special and will receive the SP provided when you set up the middleware chain originally
 * Beyond that, developers can also customize parameter binding with their own code using the `AIFunctionFactoryOptions.ConfigureParameterBinding` option. If they do that, it's up to them to know whether the data can be trusted or not based on how they have implemented the binding logic and where it gets its source data from.
 * Emphasize that in general, `AIFunctions` should be instance methods where possible. Then they can obtain trusted info about context from the instance they are a method on (e.g., to access the user's `ClaimsPrincipal`, other context about data being worked on, etc). This is always better than trying to round-trip information through the LLM, since anything that comes out of the LLM may be compromised.

PR #6255, opened 2025-04-07
Augment AIFunctionFactory.Create XML docs

stephentoub:
Closes https://github.com/dotnet/extensions/issues/6247
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6255)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
@@ -41,15 +41,67 @@ public static partial class AIFunctionFactory
     /// <returns>The created <see cref="AIFunction"/> for invoking <paramref name="method"/>.</returns>
     /// <remarks>
     /// <para>
-    /// Return values are serialized to <see cref="JsonElement"/> using <paramref name="options"/>'s
-    /// <see cref="AIFunctionFactoryOptions.SerializerOptions"/>. Arguments that are not already of the expected type are
-    /// marshaled to the expected type via JSON and using <paramref name="options"/>'s
-    /// <see cref="AIFunctionFactoryOptions.SerializerOptions"/>. If the argument is a <see cref="JsonElement"/>,
-    /// <see cref="JsonDocument"/>, or <see cref="JsonNode"/>, it is deserialized directly. If the argument is anything else unknown,
-    /// it is round-tripped through JSON, serializing the object as JSON and then deserializing it to the expected type.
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
```diff
@@ -41,15 +41,67 @@ public static partial class AIFunctionFactory
...
+    ///   </item>
+    /// </list>
+    /// All other parameter types are, by default, bound from the <see cref="AIFunctionArguments"/> dictionary passed into <see cref="AIFunction.InvokeAsync"/>
+    /// and are included in the generated JSON schema. This may be overridden by the <see cref="AIFunctionFactoryOptions.ConfigureParameterBinding"/> provided
+    /// via the <paramref name="options"/> parameter; for every parameter, the delegate is enabled to choose if the parameter should be included in the
+    /// generated schema and how its value should be bound, including handling of optionality (by default, required parameters that are not included in the
+    /// <see cref="AIFunctionArguments"/> dictionary will result in an exception being thrown). Loosely-typed additional context information may be passed
+    /// into <see cref="AIFunction.InvokeAsync"/> via the <see cref="AIFunctionArguments"/>'s <see cref="AIFunctionArguments.Context"/> dictionary; the default
+    /// binding ignores this collection, but a custom binding supplied via <see cref="AIFunctionFactoryOptions.ConfigureParameterBinding"/> may choose to
+    /// source arguments from this data.
```
SteveSandersonMS, line 82: All the info here looks good and correct, and I understand having it is an outcome from the TM review.

I wonder a bit if XML docs are the right place for this level of detail though. It's not common for us to describe so much of the workings of the whole system inside XML docs is it? I was expecting https://learn.microsoft.com/en-us/dotnet/ai/ai-extensions to cover the end-to-end system while leaving XML docs on specific .NET types to be more tightly scoped to those types alone.

stephentoub: > I wonder a bit if XML docs are the right place for this level of detail though

I think it is. Most of this is about what's permitted / how inputs/outputs are handled. That's a key thing devs need to understand when using AIFunctionFactory, and one of the main thing API docs cover in general (method signature-related stuff).

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
SteveSandersonMS: OK, fine with me

PR #6262, opened 2025-04-08
Add fuzzy filtering to the evaluation reports

peterwald:
The filter is OR'd with the tag filter, and is based on a fuzzy search of the scenario name, iteration name, and chat contents.

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6262)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/App.tsx
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/components/App.tsx
@@ -3,7 +3,7 @@
 
 import { useState } from 'react';
 import { Settings28Regular, FilterDismissRegular, DismissRegular, ArrowDownloadRegular } from '@fluentui/react-icons';
-import { Button, Drawer, DrawerBody, DrawerHeader, DrawerHeaderTitle, Switch, Tooltip } from '@fluentui/react-components';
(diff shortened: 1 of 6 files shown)


shyamnamboodiripad (approved):
The UX looks great. The only other feedback I have is that it would be nice to be able to search for text within metrics, diagnostics and metric metadata etc. This would enable scenarios like "which scenarios contain an evaluation result for metric xyz", "which other scenarios failed because of the same exception" etc.

There are also some more advanced cases that would likely need some form of search language / natural language search support. Examples include "filter down to all scenarios where metric xyz has value < 3", "which scenarios contain code vulnerability metric where code injection attack was detected in the metric metadata". We could wait for more feedback around the more advanced cases.

peterwald:
> The UX looks great. The only other feedback I have is that it would be nice to be able to search for text within metrics, diagnostics and metric metadata etc. This would enable scenarios like "which scenarios contain an evaluation result for metric xyz", "which other scenarios failed because of the same exception" etc.
> 
> There are also some more advanced cases that would likely need some form of search language / natural language search support. Examples include "filter down to all scenarios where metric xyz has value < 3", "which scenarios contain code vulnerability metric where code injection attack was detected in the metric metadata". We could wait for more feedback around the more advanced cases.

Agree that functionality would be valuable. This is intended to be an initial first step. We'll probably need to create a more advanced query function to address those more complex scenarios.

PR #6276, opened 2025-04-10
Adding reference to an unsupported built-in tool on OpenAI Chat API no longer throws

### LINKED ISSUE #6274: Using HostedWebSearchTool causes OpenAI's IChatClient to throws an exception when request sent
### Description

Hi, folks, I'm getting the following error when trying to use `HostedWebSearchTool`.
This is my min repro from my solution. I'm using the latest dev builds. 

While it works fine using a responses client, it should not throw in this case (after chatting with @stephentoub ) 

### Reproduction Steps

Sample code: 

```C#
using System.ClientModel;


var key = Environment.GetEnvironmentVariable("OPENAI_API_KEY") 
          ?? throw new InvalidOperationException("Missing environment variable: OPENAI_API_KEY.");
var openAIClient = new OpenAI.OpenAIClient(new ApiKeyCredential(key));

var chatClient = openAIClient.GetChatClient("gpt-4o-mini").AsIChatClient();


ChatOptions chatoptions = new (){
    Tools = [ new HostedWebSearchTool()],
};

while (true)
{

    var response = await chatClient.GetResponseAsync(Console.ReadLine()!, chatoptions, cancellationToken: default);
    if (response is not null)
    {
        Console.WriteLine(response);
    }
    Console.WriteLine();
}
```

### Expected behavior

The call to get response async should succeed even if the tool is inaccessible. 

### Actual behavior

```
Exception has occurred: CLR/System.ClientModel.ClientResultException
An exception of type 'System.ClientModel.ClientResultException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'HTTP 400 (invalid_request_error: )
Parameter: tool_choice

Invalid value for 'tool_choice': 'tool_choice' is only allowed when 'tools' are specified.'
   at OpenAI.ClientPipelineExtensions.<ProcessMessageAsync>d__0.MoveNext()
   at System.Runtime.CompilerServices.ConfiguredValueTaskAwaitable`1.ConfiguredValueTaskAwaiter.GetResult()
   at OpenAI.Chat.ChatClient.<CompleteChatAsync>d__16.MoveNext()
   at OpenAI.Chat.ChatClient.<CompleteChatAsync>d__6.MoveNext()
   at Microsoft.Extensions.AI.OpenAIChatClient.<GetResponseAsync>d__7.MoveNext()
   at Program.<<Main>$>d__0.MoveNext() in /Users/artleonard/src/minrepro/Program.cs:line 19
```

### Regression?

No. New scenario.

The theory is that it's using tool choice is using auto. 

### Known Workarounds

Use the Responses API. 

### Configuration

- .NET 9
- "Azure.AI.OpenAI" Version="2.2.0-beta.4"
- "Microsoft.Extensions.AI" Version="9.5.0-preview.1.25207.6"
- "Microsoft.Extensions.AI.OpenAI" Version="9.5.0-preview.1.25207.6"

### Other information

_No response_

artl93:
Fixes #6274 

- After removing non-AIFunction calls, we no longer set the result.ToolChoice to Auto, but leave it null. 
- Added a corresponding test to confirm that the function is dropped.

Previously asking for a response would result in: 

```
An exception of type 'System.ClientModel.ClientResultException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'HTTP 400 (invalid_request_error: )
Parameter: tool_choice

Invalid value for 'tool_choice': 'tool_choice' is only allowed when 'tools' are specified.'
```

Now it just carries on as if no tool was specified.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -582,22 +582,25 @@ internal sealed partial class OpenAIChatClient : IChatClient
                     }
                 }
 
-                switch (options.ToolMode)
+                if (result.Tools.Count > 0)
                 {
-                    case NoneChatToolMode:
-                        result.ToolChoice = ChatToolChoice.CreateNoneChoice();
-                        break;
+                    switch (options.ToolMode)
+                    {
+                        case NoneChatToolMode:
+                            result.ToolChoice = ChatToolChoice.CreateNoneChoice();
+                            break;
 
-                    case AutoChatToolMode:
-                    case null:
-                        result.ToolChoice = ChatToolChoice.CreateAutoChoice();
-                        break;
+                        case AutoChatToolMode:
+                        case null:
+                            result.ToolChoice = ChatToolChoice.CreateAutoChoice();
+                            break;
 
-                    case RequiredChatToolMode required:
-                        result.ToolChoice = required.RequiredFunctionName is null ?
-                            ChatToolChoice.CreateRequiredChoice() :
-                            ChatToolChoice.CreateFunctionChoice(required.RequiredFunctionName);
-                        break;
+                        case RequiredChatToolMode required:
+                            result.ToolChoice = required.RequiredFunctionName is null ?
+                                ChatToolChoice.CreateRequiredChoice() :
+                                ChatToolChoice.CreateFunctionChoice(required.RequiredFunctionName);
+                            break;
+                    }
                 }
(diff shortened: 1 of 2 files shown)


artl93:
@stephentoub  - there is the alternative here that we could throw when the requested tool / function counts do not match. 

Also note, the "CreateNoneChoice" does not work either - same result as auto. It appears that if there are no tools rendered, setting auto or none have the same effect.

artl93:
Verified having a valid and built-in tool will prefer the auto tool choice, which works properly.

stephentoub (approved):
Thanks!

Issue #6033, opened 2025-03-05
[AI Evaluation] Add EvaluationMetric property for contexts used in evaluation


shyamnamboodiripad:
When looking at scores for metrics like `Equivalence` and `Groundedness` in the report, there is currently no way to know the grounding context that was used for the evaluation. We have heard feedback that this makes it harder for anyone who is viewing the report to assess / debug why the corresponding score was low / high.

This issue tracks the following changes to address this -
- [ ] Introduce a first-class property with type `Dictionary<string, string>?` on `EvaluationMetric` to store the contexts (if any) that were used as part of the evaluation.
- [ ] Update `GroundednessEvaluator` and `EquivalenceEvaluator` to store their respective contexts in this property.
- [ ] Update the generated evaluation report to display the contexts. For now, we can display contexts if available when hovering on the card for a particular metric. But eventually, it should be possible to click on a card for a metric and view the contexts associated with this metric in a details section below the cards.

PR #6277, opened 2025-04-10
Support response caching for safety evaluators

### LINKED ISSUE #6260: [AI Evaluation] Support response caching for safety evaluators
Unlike the Quality evaluators (which talk to an LLM and cache LMM responses using `ResposeCachingChatClient`), Safety evaluators which talk to the Azure AI Content Safety service currently do not cache the evaluation responses from the service. This issue tracks the task to add response caching support for the Safety evaluators as well.

shyamnamboodiripad:
Introduce an `IChatClient` for Safety evaluators so that we can support response caching in the same way as we do for other LLM-based evaluators.

Also includes some changes to introduce a `GetContents` API on the `EvaluationContext` type. This aids with some of changes in this PR and should also help for #6033 in a subsequent PR.

Here's a screenshot that shows how this improves the run time of `SafetyEvaluatorTests` when cached responses are available.

image

TODO:
- [x] Include additional caching keys based on the service connection parameters so that the cache can be invalidated if something changes.

Fixes #6260 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6277)


src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluationContext.cs
```diff
@@ -1,10 +1,37 @@
...
+    /// <para>
+    /// For simple <see cref="EvaluationContext"/>s that only contain text, this function can return a single
+    /// <see cref="TextContent"/> object that includes the contained text.
+    /// </para>
+    /// </remarks>
+    /// <returns>
+    /// A list of <see cref="AIContent"/> objects that include all the information present in this
+    /// <see cref="EvaluationContext"/>.
+    /// </returns>
+    public abstract IReadOnlyList<AIContent> GetContents();
```
peterwald, line 36: Interesting idea to make context an AIContent. I like it.

peterwald (approved):
:shipit:

src/Libraries/Microsoft.Extensions.AI.Evaluation/EvaluationContext.cs
shyamnamboodiripad: Yes, an evaluator can still also look at any first-class properties available in the context and decide what / how they want to include as part of the evaluation request to the LLM. `GetContents` provides a convenient way to do this that is standard across multiple derived context types (and we use this to include context in the Safety evaluators).

However, the more interesting use case for `GetContents` is to provide a standard way to serialize context into the report for #6033. My plan was to take the returned `AIContent`s and include them as part of the `EvaluationMetric` so that we can show them in the report. We would only show text-based context in the report for now - but other modalities can also be supported in the future.

Issue #6266, opened 2025-04-09
Revert M.E.AI.OpenAI change that turns on strict by default


stephentoub:
While we now ourselves generate schema that we _think_ conforms to what OpenAI allows in strict mode, not all AIFunctions come from AIFunctionFactory. Any other AIFunction that gets its schema from somewhere else doesn't necessarily conform to the strict rules OpenAI puts in place, and using it via an abstraction makes it really hard to have such hard constraints there by default. We see this with MCP, for example, where MCP tools advertise valid schema that OpenAI's strict mode rejects, e.g. using `default`.

stephentoub:
It was already possible by setting a value into Additional Properties.

eiriktsarpalis:
Should we disable some of the other defaults imposed by OpenAI restrictions? Thinking of `IncludeTypeInEnumSchemas`, `DisallowAdditionalProperties`?

PR #6285, opened 2025-04-11
Disable default required property schema generation and OpenAI strict mode.

eiriktsarpalis:
Fix https://github.com/dotnet/extensions/issues/6266.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6285)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
@@ -30,10 +30,12 @@ public class ChatResponseFormat
     /// <param name="schema">The JSON schema.</param>
     /// <param name="schemaName">An optional name of the schema. For example, if the schema represents a particular class, this could be the name of the class.</param>
     /// <param name="schemaDescription">An optional description of the schema.</param>
+    /// <param name="schemaIsStrict">Whether the response should strictly adhere to the schema.</param>
     /// <returns>The <see cref="ChatResponseFormatJson"/> instance.</returns>
     public static ChatResponseFormatJson ForJsonSchema(
-        JsonElement schema, string? schemaName = null, string? schemaDescription = null) =>
+        JsonElement schema, string? schemaName = null, string? schemaDescription = null, bool schemaIsStrict = false) =>
         new(schema,
             schemaName,
-            schemaDescription);
+            schemaDescription,
+            schemaIsStrict);
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
@@ -33,7 +33,7 @@ public class ChatResponseFormat
     /// <param name="schemaIsStrict">Whether the response should strictly adhere to the schema.</param>
     /// <returns>The <see cref="ChatResponseFormatJson"/> instance.</returns>
     public static ChatResponseFormatJson ForJsonSchema(
-        JsonElement schema, string? schemaName = null, string? schemaDescription = null, bool schemaIsStrict = false) =>
+        JsonElement schema, string? schemaName = null, string? schemaDescription = null, bool? schemaIsStrict = false) =>
         new(schema,
             schemaName,
             schemaDescription,
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
@@ -30,12 +30,10 @@ public class ChatResponseFormat
     /// <param name="schema">The JSON schema.</param>
     /// <param name="schemaName">An optional name of the schema. For example, if the schema represents a particular class, this could be the name of the class.</param>
     /// <param name="schemaDescription">An optional description of the schema.</param>
-    /// <param name="schemaIsStrict">Whether the response should strictly adhere to the schema.</param>
     /// <returns>The <see cref="ChatResponseFormatJson"/> instance.</returns>
     public static ChatResponseFormatJson ForJsonSchema(
-        JsonElement schema, string? schemaName = null, string? schemaDescription = null, bool? schemaIsStrict = false) =>
+        JsonElement schema, string? schemaName = null, string? schemaDescription = null) =>
         new(schema,
             schemaName,
-            schemaDescription,
-            schemaIsStrict);
+            schemaDescription);
 }
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -611,8 +611,7 @@ internal sealed partial class OpenAIChatClient : IChatClient
                         jsonFormat.SchemaName ?? "json_schema",
                         BinaryData.FromBytes(
                             JsonSerializer.SerializeToUtf8Bytes(jsonSchema, ChatClientJsonContext.Default.JsonElement)),
-                        jsonFormat.SchemaDescription,
-                        jsonSchemaIsStrict: true) :
+                        jsonFormat.SchemaDescription) :
                     OpenAI.Chat.ChatResponseFormat.CreateJsonObjectFormat();
             }
         }
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -611,8 +611,7 @@ internal sealed partial class OpenAIChatClient : IChatClient
                         jsonFormat.SchemaName ?? "json_schema",
                         BinaryData.FromBytes(
                             JsonSerializer.SerializeToUtf8Bytes(jsonSchema, ChatClientJsonContext.Default.JsonElement)),
-                        jsonFormat.SchemaDescription,
-                        jsonFormat.SchemaIsStrict) :
+                        jsonFormat.SchemaDescription) :
                     OpenAI.Chat.ChatResponseFormat.CreateJsonObjectFormat();
             }
         }
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -623,10 +623,10 @@ internal sealed partial class OpenAIChatClient : IChatClient
     /// <summary>Converts an Extensions function to an OpenAI chat tool.</summary>
     private static ChatTool ToOpenAIChatTool(AIFunction aiFunction)
     {
-        // Default strict to true, but allow to be overridden by an additional Strict property.
+        // Default strict to false, but allow to be overridden by an additional Strict property.
         bool strict =
-            !aiFunction.AdditionalProperties.TryGetValue("Strict", out object? strictObj) ||
-            strictObj is not bool strictValue ||
+            aiFunction.AdditionalProperties.TryGetValue("Strict", out object? strictObj) &&
+            strictObj is bool strictValue &&
             strictValue;
 
         // Map to an intermediate model so that redundant properties are skipped.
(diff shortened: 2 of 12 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -627,10 +627,10 @@ private static ChatCompletionOptions ToOpenAIOptions(ChatOptions? options)
     /// <summary>Converts an Extensions function to an OpenAI chat tool.</summary>
     private static ChatTool ToOpenAIChatTool(AIFunction aiFunction)
     {
-        // Default strict to true, but allow to be overridden by an additional Strict property.
+        // Default strict to false, but allow to be overridden by an additional Strict property.
```
stephentoub, line 630: I think we should default to null like we previously did:
https://github.com/dotnet/extensions/pull/6064/files#diff-5c09d9da8396d3bee35a8fd4dac021577c046401c7b4e1ee8797fbae4bd8d686L446
such that we end up using whatever is the default behavior of the underlying OpenAI library.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -24,7 +24,8 @@ public static class ChatClientStructuredOutputExtensions
     {
         IncludeSchemaKeyword = true,
         DisallowAdditionalProperties = true,
-        IncludeTypeInEnumSchemas = true
+        IncludeTypeInEnumSchemas = true,
+        RequireAllProperties = true,
```
stephentoub, line 28: With AIFunction, someone who wants to turn on strict can do so by adding "Strict"=true to AdditionalProperties.

Do we have a way to do that for response format? Maybe we need to add an AdditionalProperties to ForJsonSchema / ChatResponseFormat?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -627,11 +627,10 @@ private static ChatCompletionOptions ToOpenAIOptions(ChatOptions? options)
...
     {
-        // Default strict to true, but allow to be overridden by an additional Strict property.
-        bool strict =
-            !aiFunction.AdditionalProperties.TryGetValue("Strict", out object? strictObj) ||
-            strictObj is not bool strictValue ||
-            strictValue;
+        bool? strict =
+            aiFunction.AdditionalProperties.TryGetValue("Strict", out object? strictObj) &&
+            strictObj is bool strictValue ?
+            strictValue : null;
```
stephentoub, line 633: There are some other places that turn on strict as well:
https://github.com/search?q=repo%3Adotnet%2Fextensions+%22strict%3A+true%22&type=code

stephentoub: See https://github.com/dotnet/extensions/pull/6064 for the changes that were made to turn it on

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
eiriktsarpalis: I think we need to expose an AIJsonSchemaCreateOptions parameter to these methods so that the default can be overridden

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
eiriktsarpalis: Incorporated.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -615,8 +615,7 @@ private static ChatCompletionOptions ToOpenAIOptions(ChatOptions? options)
                         jsonFormat.SchemaName ?? "json_schema",
                         BinaryData.FromBytes(
                             JsonSerializer.SerializeToUtf8Bytes(jsonSchema, ChatClientJsonContext.Default.JsonElement)),
-                        jsonFormat.SchemaDescription,
-                        jsonSchemaIsStrict: true) :
+                        jsonFormat.SchemaDescription) :
```
SteveSandersonMS, line 618: Am I right to think this doesn't only affect tool calling, but also affects structured output? If so, do we really have to do that or can we leave strict mode on for structured output?

My guess right now about how "strict" mode works on OpenAI is that:

 * When a JSON schema is provided by the client, OpenAI somehow includes that in the prompt so that the LLM tends to produce conformant output. This happens whether or not "strict" mode is enabled. It doesn't matter if the JSON schema includes unknown rules since this is just LLM prompting.
 * If "strict" mode is enabled, OpenAI does the above **plus** it changes the token sampler so that it can't pick tokens that would violate the schema. This is only possible if their token sampling logic understands the JSON schema completely, hence it will fail if you use any JSON schema rules they don't support.

I don't know if this is actually true; it's just a guess. The [docs](https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat) use `strict` in many examples but never define what it means.

If the guess is true, then it seems like strict mode is strictly better than nonstrict in all cases where you know for sure that the JSON schema is supported. And we do know that is the case for `GetResponseAsync<T>`, right?

SteveSandersonMS: BTW let's not block on this. If there isn't an easy way to know exactly when strict mode is safe to turn on, I'm fine with defaulting it to off. Developers can still control this manually, and it's up to OpenAI to make both strict and nonstrict modes work as well as possible (for example, they could inspect the JSON schema and detect whether or not they can handle it in strict mode automatically - don't know why they don't already do that).

eiriktsarpalis: If the OpenAI SDK API docs are anything to go by, this flag specifically governs strictness of structured outputs only. I had initially left this change out, but then @stephentoub requested we [revert all strict mode related changes](https://github.com/dotnet/extensions/pull/6285#discussion_r2039539442). 

Note that apart from reverting strict mode, this PR also disables marking all properties as required in generated schemas by default since we've received feedback on this creating problems in non-OpenAI contexts (particularly MCP). As such, hardcoding strict mode in OpenAI components is bound to clash with the new schema generation defaults, so turning it off by default seems advisable.

Thoughts?

SteveSandersonMS: I'm fine with simplifying the change so that we just don't override OpenAI's default strictness at all, based on the reasoning that whatever default they pick should be right for them.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
eiriktsarpalis: I've reverted this change and exposed an additional parameter in the `ForJsonSchema` factory.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -623,11 +623,10 @@ private static ChatCompletionOptions ToOpenAIOptions(ChatOptions? options)
     /// <summary>Converts an Extensions function to an OpenAI chat tool.</summary>
     private static ChatTool ToOpenAIChatTool(AIFunction aiFunction)
     {
-        // Default strict to true, but allow to be overridden by an additional Strict property.
-        bool strict =
-            !aiFunction.AdditionalProperties.TryGetValue("Strict", out object? strictObj) ||
-            strictObj is not bool strictValue ||
-            strictValue;
+        bool? strict =
+            aiFunction.AdditionalProperties.TryGetValue("Strict", out object? strictObj) &&
```
SteveSandersonMS, line 627: Do you think we should have a more specific name for this? It's not obvious what it means for the function itself to be "strict", whereas if we called it `StrictJsonSchema` there would be no ambiguity.

eiriktsarpalis: Possibly, but wouldn't this break folks relying on the current name to set it?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
```diff
@@ -30,10 +30,12 @@ private protected ChatResponseFormat()
     /// <param name="schema">The JSON schema.</param>
     /// <param name="schemaName">An optional name of the schema. For example, if the schema represents a particular class, this could be the name of the class.</param>
     /// <param name="schemaDescription">An optional description of the schema.</param>
+    /// <param name="schemaIsStrict">Whether the response should strictly adhere to the schema.</param>
     /// <returns>The <see cref="ChatResponseFormatJson"/> instance.</returns>
     public static ChatResponseFormatJson ForJsonSchema(
-        JsonElement schema, string? schemaName = null, string? schemaDescription = null) =>
+        JsonElement schema, string? schemaName = null, string? schemaDescription = null, bool schemaIsStrict = false) =>
```
SteveSandersonMS, line 36: I like the idea of having control over this, but I think more is needed for this to be useful end-to-end. The most common usage pattern is calling `GetResponse<T>` in `ChatResponseStructuredOutputExtensions`, and that doesn't have a way to specify a value for this flag.

One option is to change `GetResponseAsync<T>'s` `useJsonSchema` flag into some enum like `JsonSchemaMode? jsonSchemaMode` with values `None` (equivalent to `null`), `Normal` and `Strict`. It could be a `[Flags]` enum if we want it to be combinable like `JsonSchemaMode.Strict | JsonSchemaMode.Gemini`. Or maybe we would think some combinations are meaningless so instead of `[Flags]`, just have explicit combinations like `None`, `OpenAI`, `OpenAIStrict`, `Gemini`, etc.

Altogether this would force us to make quite a few more design decisions.

Another whole option would be to say that "strictness" is not a general concept and is OpenAI-specific, so this concept shouldn't appear in MEAI's API surface and instead the OpenAI client can look at something in `ChatOptions.AdditionalProperties` to decide whether to use strict mode or not. That's a safer option in the short term and would still allow us to create an MEAI-level "strictness" concept later non-breakingly if we wanted.

What do you think?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
SteveSandersonMS: We can make breaking changes before GA :) Not saying it's not something to think carefully about, but it is allowed.

Personally I'd prefer taking a breaking change on this relatively niche API in preview if that gets us the API shape we want to support forever.

SteveSandersonMS: BTW we are doing breaking changes for loads of stuff related to `AdditionalProperties` in other places because we're greatly reducing its usage for configuring requests and getting custom properties from responses. This would be a minor change by comparison.

eiriktsarpalis: Fair. I'll rename to `stringJsonSchema` then.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
eiriktsarpalis: I agree that this probably forces a redesign of the current `useJsonSchema` API shape into something different, but it felt like something going beyond the scope of this particular PR, which is reverting strict to being turned off (without possibility of turning it back on in the context of structured responses).

I think a `JsonSchemaMode` or `JsonSchemaCompatibilityMode` enum is the way to go, but I think we need to go back to the drawing board and incorporate this design into the `AIJsonUtilities` APIs rather than just using it in this context.

As an interim workaround, I have turned on `RequireAllProperties` specifically for structured responses since OpenAI complains about that even with strict mode turned off.

Let's try to get this merged today so that we can unblock the MCP use case.

SteveSandersonMS: Shall we remove the `schemaIsStrict` and `ChatResponseFormatJson.SchemaIsStrict` for now then, and leave control over strictness to the OpenAI client (which can read something from `AdditionalProperties` if it wants, but doesn't have to immediately)?

That would ensure we don't create any API surface that we don't want to retain long-term.

eiriktsarpalis: Sure, can do that.

eiriktsarpalis: Done.

SteveSandersonMS (approved):
Excellent!

PR #6296, opened 2025-04-15
Augment FunctionInvokingChatClient's span with token counts

stephentoub:
@lmolkova suggested this would be valuable.

Alternatively, we could consider removing the activity use entirely from FunctionInvokingChatClient, and just tell folks that want such a span to add an OpenTelemetryChatClient before the FICC in the pipeline, even if they also have one after it:
```C#
.UseOpenTelemetry()
.UseFunctionInvocation()
.UseOpenTelemetry()
```
I _think_ it would just naturally do the "right thing". Though if we were to do that, we might want to make the meter optional, so that metrics weren't counted twice.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6296)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -294,6 +294,8 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
         response.Messages = responseMessages!;
         response.Usage = totalUsage;
 
+        AddUsageTags(activity, totalUsage);
+
         return response;
     }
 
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Makes sense. I think counting the usage within FICC is completely fine as an extension of the existing design decision to make FICC aware of telemetry as a concept.

I agree that making FICC unaware of telemetry would be another valid design approach and would arguably be a cleaner separation of concerns, but the drawback of needing two copies of `UseOpenTelemetry` in the pipeline (to provide spans representing both aggregate and per-iteration views of function invocation) would be quite bad for developer experience as very few people would understand why.

In general it makes sense that *sometimes* middleware should emit its own telemetry data since it might be doing something custom. Given that general rule I'm perfectly happy with FICC doing so, and it simplifies the experience so seems like a good tradeoff.

Issue #6312, opened 2025-04-17
Extend DI support for AIFunctionFactory


jeffhandley:
We should support the existing `[FromKeyedServices]` attribute. This scenario has surfaced in MCP scenarios, and this attribute is currently ignored.

For non-keyed services, we could use the `IServiceProvider` to see if it can resolve the type. This would require accepting an `IServiceProvider` on `AIFunctionFactoryOptions`.

stephentoub:
https://github.com/dotnet/extensions/pull/6310

PR #6317, opened 2025-04-17
Utilize IServiceProviderIsService in AIFunctionFactory

stephentoub:
Add AIFunctionFactoryOptions.Services, and use it when examining function parameters to determine whether they should be resolved by default from DI.

Closes https://github.com/dotnet/extensions/issues/6312, whether or not we decide to merge this.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6317)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/Functions/AIFunctionFactory.cs
@@ -80,6 +80,16 @@ public static partial class AIFunctionFactory
     ///       The handling of such parameters may be overridden via <see cref="AIFunctionFactoryOptions.ConfigureParameterBinding"/>.
     ///     </description>
     ///   </item>
+    ///   <item>
+    ///     <description>
+    ///       When the <see cref="AIFunction"/> is constructed, it may be passed an <see cref="IServiceProvider"/> via 
+    ///       <see cref="AIFunctionFactoryOptions.Services"/>. Any parameter that can be satisfied by that <see cref="IServiceProvider"/>
+    ///       according to <see cref="IServiceProviderIsService"/> will not be included in the generated JSON schema and will be resolved 
+    ///       from the <see cref="IServiceProvider"/> provided to <see cref="AIFunction.InvokeAsync"/> via <see cref="AIFunctionArguments.Services"/>,
+    ///       rather than from the argument collection. The handling of such parameters may be overridden via
+    ///       <see cref="AIFunctionFactoryOptions.ConfigureParameterBinding"/>.
+    ///     </description>
+    ///   </item>
     /// </list>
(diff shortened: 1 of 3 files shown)


jeffhandley (approved):
I'm in favor of supporting this.

Do I understand correctly though that the potential drawback is that we take the liberty of excluding parameters that will be resolved from the service provider from the schema exchanged with the model, therefore preventing the model from being able to include a value for the parameter itself, and there might be scenarios where that would be unexpected/undesired?

*After typing that, I asked Copilot what the potential drawbacks were, and it included that reasoning and also included other drawbacks, pasted below.*

I'm still in favor myself though, and the implementation plugs in very surgically, so I'm comfortable with the risk at this stage too, but this could also be serviced in later as a behavioral breaking change that I think would be acceptable.

I ultimately defer to you two though, @stephentoub and @SteveSandersonMS.

> The potential drawback in integrating the feature outlined in this pull request includes:
> 
> 1. **Increased Complexity:**
>    - The addition of `IServiceProvider`-specific logic introduces more complexity to the `AIFunctionFactory` and its associated classes. This could make the code harder to understand, maintain, and debug, especially for developers unfamiliar with the dependency injection flow.
> 
> 2. **Dependency on `IServiceProvider`:**
>    - Tight coupling with `IServiceProvider` and `IServiceProviderIsService` may make the feature less flexible. This dependency assumes that the `IServiceProvider` provided at construction matches the one used during invocation, which could lead to issues if mismatched.
> 
> 3. **Behavioral Changes:**
>    - Parameters that can be resolved via `IServiceProvider` are excluded from the JSON schema. This change in behavior may surprise users who expect all parameters to be explicitly listed in the schema. It may also create discrepancies between the schema and runtime behavior.
> 
> 4. **Error Handling Overhead:**
>    - The feature introduces additional error handling for cases where `IServiceProvider` is missing or fails to resolve a parameter. This might lead to more runtime exceptions, adding to the risk of failure if users do not carefully configure the `Services`.
> 
> 5. **Testing Complexity:**
>    - Unit testing becomes more challenging because the behavior of the functions now depends on external services and their availability in the provided `IServiceProvider`. Mocking or setting up these services for tests could increase the burden on test authors.
> 
> 6. **Potential for Misconfiguration:**
>    - If the `Services` provided during construction and invocation are not aligned or do not satisfy the required dependencies, the feature could fail at runtime. This may make it harder for developers to debug and resolve configuration issues.
> 
> 7. **Backward Compatibility:**
>    - Existing users of the library who rely on the explicit schema for all parameters or who do not use `IServiceProvider` may need to adjust their code to accommodate the new behavior.
> 
> 8. **Performance Overhead:**
>    - Resolving parameters dynamically from `IServiceProvider` may introduce slight performance overhead compared to binding all parameters from the arguments dictionary.
> 
> These drawbacks should be weighed against the benefits of making the library more dependency-injection-friendly and flexible for users relying on service resolution.

PR #6319, opened 2025-04-17
Add ChatOptions.RawRepresentation

jozkee:
And stop looking for individually specific additional properties in top-level types
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6319)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -461,164 +419,91 @@ private static ChatResponse FromOpenAIChatCompletion(ChatCompletion openAIComple
...
+        if (options is null)
         {
-            (response.AdditionalProperties ??= [])[nameof(openAICompletion.Refusal)] = refusal;
+            return new ChatCompletionOptions();
         }
 
-        if (openAICompletion.RefusalTokenLogProbabilities is { Count: > 0 } refusalTokenLogProbs)
+        ChatCompletionOptions result;
+        if (options.AdditionalProperties is { Count: > 0 } additionalProperties &&
+            additionalProperties.TryGetValue(nameof(ChatCompletionOptions), out ChatCompletionOptions? optionsFromAdditionalProperties))
```
jozkee, line 435: nameof(ChatCompletionOptions) is just a placeholder.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -362,19 +346,6 @@ private static async IAsyncEnumerable<ChatResponseUpdate> FromOpenAIStreamingCha
                 }
             }
 
-            // Refusals are about the model not following the schema for tool calls. As such, if we have any refusal,
-            // add it to this function calling item.
-            if (refusal is not null)
```
stephentoub, line 367: If we're no longer using this, we no longer need the local or appending to it?

jozkee: Right, I also removed `fingerprint`.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
jozkee: Do I understand correctly that this approach will also save us from affecting how AdditionalProperties are handled in other levels e.g. AIContent?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,109 +273,99 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
+        {
+            result = CreateAzureAIOptions(chatContents, options);
+        }
+
+        result.FrequencyPenalty = options.FrequencyPenalty;
+        result.MaxTokens = options.MaxOutputTokens;
+        result.NucleusSamplingFactor = options.TopP;
+        result.PresencePenalty = options.PresencePenalty;
+        result.Temperature = options.Temperature;
+        result.Seed = options.Seed;
```
stephentoub, line 307: Am I remembering correctly that all of these ChatCompletionOptions properties are nullable? If so, we should only set these now if the destination is null and the source is non-null, since they may have been set on the underlying object.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,109 +273,99 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-            if (options.StopSequences is { Count: > 0 } stopSequences)
-            {
-                foreach (string stopSequence in stopSequences)
-                {
-                    result.StopSequences.Add(stopSequence);
-                }
-            }
+            result = azureAIOptions;
+            result.Messages = ToAzureAIInferenceChatMessages(chatContents).ToList();
+            result.Model = options.ModelId ?? _metadata.DefaultModelId ?? throw new InvalidOperationException("No model id was provided when either constructing the client or in the chat options.");
```
stephentoub, line 295: As with below, I think we should only overwrite the Model if it's null.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,109 +273,99 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-            result.Seed = options.Seed;
-
-            if (options.StopSequences is { Count: > 0 } stopSequences)
-            {
-                foreach (string stopSequence in stopSequences)
-                {
-                    result.StopSequences.Add(stopSequence);
-                }
-            }
+            result = azureAIOptions;
```
stephentoub, line 293: You shouldn't need two separate variables here, and could just use:
```C#
if (options.RawRepresentation is ChatCompletionOptions result)
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -118,6 +118,17 @@ public string? ChatThreadId
     [JsonIgnore]
     public IList<AITool>? Tools { get; set; }
 
+    /// <summary>
+    /// Gets or sets the raw representation of the chat options from an underlying implementation.
+    /// </summary>
+    /// <remarks>
+    /// If a <see cref="ChatOptions"/> is created to represent some underlying object from another object
+    /// model, this property can be used to store that original object. This can be useful for debugging or
+    /// for enabling a consumer to access the underlying object model if needed.
```
stephentoub, line 127:
```suggestion
    /// The underlying <see cref="IChatClient" /> implementation may have its own representation of options.
    /// When <see cref="IChatClient.GetResponseAsync" /> or <see cref="IChatClient.GetStreamingResponseAsync" />
    /// is invoked with a <see cref="ChatOptions" />, that implementation may convert the provided options into
    /// its own representation in order to use it while performing the operation. For situations where a consumer knows
    /// which concrete <see cref="IChatClient" /> is being used and how it represents options, an instance of that
    /// implementation-specific options type may be stored into this <see cref="RawRepresentation" /> property, for
    /// the <see cref="IChatClient" /> implementation to use instead of creating a new instance. Such implementations
    /// may mutate the supplied options instance further based on other settings supplied on this <see cref="ChatOptions" /> instance or from other inputs, like the enumerable of <see cref="ChatMessage"/>s. This is typically used
    /// in order to set an implementation-specific setting that isn't otherwise exposed from the strongly-typed properties
    /// on <see cref="ChatOptions" />.
```

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,109 +273,99 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-                            }
-
-                            break;
-                    }
+                    result.Tools.Add(ToAzureAIChatTool(af));
                 }
             }
 
-            if (options.Tools is { Count: > 0 } tools)
+            switch (options.ToolMode)
```
stephentoub, line 327: Does result.ToolChoice default to null? Assuming yes, we should probably only do this switch if result.ToolChoice is null, so as not to overwrite an explicit setting the user providing the raw representation already made.

If, however, result.ToolChoice defaults to non-null, we should stick with what you have.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,109 +273,99 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
+                case RequiredChatToolMode required:
+                    result.ToolChoice = required.RequiredFunctionName is null ?
+                        ChatCompletionsToolChoice.Required :
+                        new ChatCompletionsToolChoice(new FunctionDefinition(required.RequiredFunctionName));
+                    break;
             }
+        }
 
-            if (options.ResponseFormat is ChatResponseFormatText)
+        if (options.ResponseFormat is ChatResponseFormatText)
```
stephentoub, line 346: Same comment/question here for result.ResponseFormat.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -465,159 +414,83 @@ private static ChatResponse FromOpenAIChatCompletion(ChatCompletion openAIComple
...
-            result.MaxOutputTokenCount = options.MaxOutputTokens;
-            result.TopP = options.TopP;
-            result.PresencePenalty = options.PresencePenalty;
-            result.Temperature = options.Temperature;
-            result.AllowParallelToolCalls = options.AllowMultipleToolCalls;
+            return new ChatCompletionOptions();
+        }
+
+        ChatCompletionOptions result = options.RawRepresentation is ChatCompletionOptions openAIOptions ?
+            openAIOptions : new ChatCompletionOptions();
```
stephentoub, line 429:
```suggestion
        ChatCompletionOptions result = options.RawRepresentation as ChatCompletionOptions ?? new();
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -465,159 +414,83 @@ private static ChatResponse FromOpenAIChatCompletion(ChatCompletion openAIComple
...
+
+        result.FrequencyPenalty = options.FrequencyPenalty;
+        result.MaxOutputTokenCount = options.MaxOutputTokens;
+        result.TopP = options.TopP;
+        result.PresencePenalty = options.PresencePenalty;
+        result.Temperature = options.Temperature;
+        result.AllowParallelToolCalls = options.AllowMultipleToolCalls;
```
stephentoub, line 436: Same question/comments here as on Azure.AI.Inference. Are these properties on ChatCompletionOptions all nullable?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
stephentoub: I think we still have the issue with AIContent, like image details.

jozkee: Are you saying we should remove AdditionalProperties lookups from other levels as well and pass the RawRepresentation if present? e.g: https://github.com/dotnet/extensions/blob/805dd0564f9d30093a1968b3b7a3bf2690fc1505/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs#L181-L186

would be 
```
foreach (var content in contents)
{
    if (content.RawRepresentation is ChatMessageContentPart part)
    {
        parts.Add(part);
        continue;
    }
    ...

Comment (stephentoub) [reply src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs]: I wasn't saying that, rather that having this on ChatOptions doesn't dove the problem at those layers. I'm not sure if we want to do what you suggest or not. On the one hand, it'd be fullest fidelity with regards to additional props, but on the other, I fear a) we'd lose any mutations made on the AIContent, b) we might duplicate content if we ever split an underlying content into multiple AIContent, and c) we'd end up with different behavior if using an AIContent that had roundtripped via serialization and lost its raw representation.

File: src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs [hunk 11]
```diff
@@ -273,109 +273,99 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-                        default:
-                            if (prop.Value is not null)
-                            {
-                                byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(object)));
-                                result.AdditionalProperties[prop.Key] = new BinaryData(data);
-                            }
-
-                            break;
-                    }
+                    result.Tools.Add(ToAzureAIChatTool(af));
```
jozkee, line 323: Tools and some other collections are initialized in the ctor, if the collection is not empty here, do we want to still `Add` elements specified in ChatOptions? I feel it would not align with the write-only-if-null semantic you described for the nullable properties of the underlying options, and it may lead to errors like duplicated elements if the user defined the tools in both the underlying and our options.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
SteveSandersonMS: > I mentioned this offline earlier today, but I think we should look at making ChatOptions symmetrical with the response types, by adding a RawRepresentation object property.

That sounds OK to me.

It does come with the drawback that it restricts dev to having only a single `RawRepresentation` value per `ChatOptions`, which loses the option to assign multiple provider-specific options objects if you want to do some custom think for multiple possible predetermined providers.

However:

1. It seems unlikely people will need this escape hatch for multiple things. Class libraries won't want to do so because it would force them to take dependencies on multiple provider client libraries. Applications could do so but they could always check which provider is in use and then assign the provider-specific object.
2. We could still add a further escape hatch using `AdditionalProperties` to override this in the future if we want.

Another approach that makes it feel easy and strongly-typed is having some provider-specific extension method like `chatOptions.SetOpenAIOptions(...)` that internally writes to `AdditionalProperties`. Arguably `RawRepresentation` has more obvious semantics for how to subsequently mutate it further.

But it seems fine either way.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
SteveSandersonMS: So something like this?

```cs
        result.Seed ??= options.Seed;
```

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
SteveSandersonMS: Merging the tools from `ChatOptions.Tools` and `ChatOptions.RawRepresentation.ToolsOrWhatever` in a single collection is what I'd expect to happen.

If the developer duplicates entries across both sources that's their mistake and their bug to fix, but I don't think people will commonly have any way to make a tool show up in both places since they would be represented by different object types.

stephentoub: > Merging the tools from ChatOptions.Tools and ChatOptions.RawRepresentation.ToolsOrWhatever in a single collection is what I'd expect to happen.

Me, too.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
stephentoub: > However

Yup. As another future extension, we could also say that RawRepresentation could be a `Dictionary<IChatClient, object>` or something like that, such that multiple raw representations could be passed in. But, as you say, this seems like a very uncommon need.

> Arguably RawRepresentation has more obvious semantics for how to subsequently mutate it further.

That's part of my thinking as well, with the same reasoning for why we have RawRepresentation on various response objects and AIContent types, rather than them all special-casing some slot in their AdditionalProperties.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,66 +273,57 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-            result.Seed = options.Seed;
-
-            if (options.StopSequences is { Count: > 0 } stopSequences)
-            {
-                foreach (string stopSequence in stopSequences)
-                {
-                    result.StopSequences.Add(stopSequence);
-                }
-            }
+            result.Messages = ToAzureAIInferenceChatMessages(chatContents).ToList();
```
SteveSandersonMS, line 292: This seems surprising. `ChatCompletionsOptions` holds the list of messages? That makes it not seem like an "options" object at all but rather some kind of "request" representation. I know we don't own that type and didn't make that choice.

Mutating it like this would be surprising to me as an app developer. I would have expected I could reuse the same `chatOptions.RawRepresentation` instance across multiple requests, since in my expectation it represents some options that may be applicable to multiple requests. But if each request mutates it, we risk bugs.

Is there any way to clone a `ChatCompletionsOptions` before we mutate and use it?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -465,145 +414,71 @@ private static ChatResponse FromOpenAIChatCompletion(ChatCompletion openAIComple
...
-
-                if (additionalProperties.TryGetValue(nameof(result.StoredOutputEnabled), out bool storeOutputEnabled))
-                {
-                    result.StoredOutputEnabled = storeOutputEnabled;
-                }
-
-                if (additionalProperties.TryGetValue(nameof(result.TopLogProbabilityCount), out int topLogProbabilityCountInt))
-                {
-                    result.TopLogProbabilityCount = topLogProbabilityCountInt;
+                    result.Tools.Add(ToOpenAIChatTool(af));
```
SteveSandersonMS, line 454: Does this risk adding more and more copies of the same tool every time the same `options` object is used?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
SteveSandersonMS: BTW I see this has been mentioned in the XML docs:

> Such implementations may mutate the supplied options instance further

However people won't necessarily understand that. It's pretty high-stakes since if a `ChatCompletionsOptions` was used across two different user sessions (which seems reasonable based on the naming), it could leak one user's messages into the other user's session.

stephentoub: > but rather some kind of "request" representation

That's what it is: all of the arguments to CompleteAsync method, wrapped up into one object. It's not a great design.

> It's pretty high-stakes since if a ChatCompletionsOptions was used across two different user sessions

That's the design of ChatCompletionsOptions. Though I guess your point is that if folks don't realize the messages are included in "options", they might just set the options with their few knobs set and then reuse the same ChatOptions over and over.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
jozkee: Yes, it does.

stephentoub: We can't just clear it out first, as someone could reasonably expect that they can add tools here, especially one's that the abstraction can't yet represent. And we can't just check first whether the tool is there, because if the same object is being reused, it'll include tools from the previous go around that were never removed, and thus this could include tools not intended for reuse. Seems like we're going to have to clone, which is unfortunate.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -273,66 +275,71 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-
-            if (options.StopSequences is { Count: > 0 } stopSequences)
-            {
-                foreach (string stopSequence in stopSequences)
-                {
-                    result.StopSequences.Add(stopSequence);
-                }
-            }
+            // Clone the options to avoid modifying the original.
+            ////TODO: ToolChoice is not being cloned.
```
jozkee, line 306: It may be a bug in AI.Inference because I don't observe this in OpenAI, I will look into it, just wanted to push changes to show the POC.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
stephentoub: Another option would be to have RawRepresentation be RawRepresentations, and have it be an `IList<object>?`. You could then put multiple different options into that if you were expecting to be working with possibly multiple different providers, and each leaf IChatClient could look through all the objects for the first it understands.

jozkee: since it's an `object`, I believe is desired to support both ways.

stephentoub:
Another idea... rather than trying to clone ourselves, what if we make RawRepresentation instead be a `Func<IChatClient, object>?`. It has the responsibility of creating a usable options object that's fresh. We might need a slightly different name, like RawRepresentationFunc. @SteveSandersonMS, @jozkee, thoughts?

jozkee:
A delegate still has the risk of returning a shared instance.  We can add a remark saying that we strongly recommend it returns new options.

I'm assuming the `IChatClient` param is for https://github.com/dotnet/extensions/pull/6319#discussion_r2071594321. this way, a user can choose what to return for each chat client impl.

As for the naming, Factory implies returning a new instance, so that would be my choice.
Could there be people interested in returning shared options? If we care about those cases, RawRepresentationProvider may work since it doesn't imply "always return new" but I don't think we want to encourage that due to all the issues already discussed here.

If we move to this Func approach, I'm assuming we can remove this serialize-deserialize cloning since we will consider the "errors by using shared options" case to be "under your own risk".

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -118,6 +119,26 @@ public string? ChatThreadId
...
 
+    /// <summary>
+    /// Gets or sets a callback responsible of creating the raw representation of the chat options from an underlying implementation.
+    /// </summary>
+    /// <remarks>
+    /// The underlying <see cref="IChatClient" /> implementation may have its own representation of options.
+    /// When <see cref="IChatClient.GetResponseAsync" /> or <see cref="IChatClient.GetStreamingResponseAsync" />
+    /// is invoked with a <see cref="ChatOptions" />, that implementation may convert the provided options into
+    /// its own representation in order to use it while performing the operation. For situations where a consumer knows
+    /// which concrete <see cref="IChatClient" /> is being used and how it represents options, an instance of that
```
stephentoub, line 130:
```suggestion
    /// which concrete <see cref="IChatClient" /> is being used and how it represents options, a new instance of that
```

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -281,66 +281,68 @@ private static ChatRole ToChatRole(global::Azure.AI.Inference.ChatRole role) =>
...
-            // These properties are strongly typed on ChatOptions but not on ChatCompletionsOptions.
-            if (options.TopK is int topK)
+        if (options.AdditionalProperties is { } props)
+        {
+            foreach (var prop in props)
             {
-                result.AdditionalProperties["top_k"] = new BinaryData(JsonSerializer.SerializeToUtf8Bytes(topK, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(int))));
+                byte[] data = JsonSerializer.SerializeToUtf8Bytes(prop.Value, AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(object)));
+                result.AdditionalProperties[prop.Key] = new BinaryData(data);
             }
```
stephentoub, line 313: We still need this one, don't we? This is ensuring that the strongly-typed property on ChatOptions is respected if the underlying provider supports it.

stephentoub (approved):
Thanks.

I assume you've run all of the integration tests and they pass?

jozkee:
> I assume you've run all of the integration tests and they pass?

I just did, no new failures after comparing test results from main vs this branch. Have to compare because some fail in main.

stephentoub:
> some fail in main.

Some are failing in main? That's a problem

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
jozkee: Yes, but we should write it only if it's not already set in the additional properties of raw representation, correct?

stephentoub: That sounds like a good idea.

stephentoub:
@jozkee, looks like the PR missed updating https://github.com/dotnet/extensions/blob/0313177a0f7deef3d0e13f25ff90c9e78c017878/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs#L320 ?

stephentoub:
Also embedding options? e.g.
https://github.com/dotnet/extensions/blob/0313177a0f7deef3d0e13f25ff90c9e78c017878/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIEmbeddingGenerator.cs#L112

PR #6320, opened 2025-04-18
Remove the abstraction for token counting from the main evaluation API

### LINKED ISSUE #6234: [AI Evaluation] Rationalize IEvaluationTokenCounter interface
Currently `IEvaluationTokenCounter` has a single `int CountTokens(string message)` function that takes a `string` and returns a token count. Ideally, this would also support other modalities like images, audio etc. We may also want to make the function asynchronous (to support cases where IO may be needed).

The proposal is to change the function to `CountTokensAsync(AIContent content, CancellationToken cancellationToken)` with extension method overloads for counting tokens in `IEnumerable<AIContent>` and `ChatMessage`.

`AIContent` already models all supported modalities so it seems like an ideal candidate.

shyamnamboodiripad:
This change is being made because there is still some uncertainty around what a general-purpose token counting abstraction (that supports all kinds of future models, and all kinds of input modalities) should look like at the moment. We do not want to bake in an API that only supports text-based inputs for the models and use cases that are prevalent today, since it would be a potential breaking change to change this API after we release a stable version of the evaluation APIs.

We can always reintroduce the token counting support in a non-breaking fashion in the future if and when there is more clarity on what a general-purpose token counting abstraction should look like, or if and when a general-purpose token counting abstraction is introduced in a lower layer (Microsoft.Extensions.AI) in the future.

In the meanwhile, callers should still be able to use the `Microsoft.ML.Tokenizers` library directly to count tokens in text-based content and trim down the conversation history before calling `EvaluateAsync()` if needed.

Fixes #6234 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6320)


src/Libraries/Microsoft.Extensions.AI.Evaluation/Microsoft.Extensions.AI.Evaluation.csproj
```diff
@@ -15,6 +15,10 @@
     <MinMutationScore>0</MinMutationScore>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <InjectExperimentalAttributeOnLegacy>true</InjectExperimentalAttributeOnLegacy>
+  </PropertyGroup>
```
peterwald, line 20: What is the purpose of this property?

src/Libraries/Microsoft.Extensions.AI.Evaluation/ChatMessageExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+                break;
+            }
+            else
+            {
+                remainingTokenBudget = newTokenBudget;
+                trimmedConversation.Add(message);
+            }
+        }
+
+        return (trimmedConversation, remainingTokenBudget);
```
peterwald, line 71: >trimmedConversation [](http://example.com/codeflow?start=16&length=19)

Should this be reversed again, so that the returned conversation is in the same order it was provided in?

test/Libraries/Microsoft.Extensions.AI.Evaluation.Tests/ConversationTrimmingTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        (IEnumerable<ChatMessage> trimmedConversation, int remainingTokenBudget) =
+            conversation.Trim(tokenizer, TokenBudget);
+
+        List<ChatMessage> trimmedConversationList = trimmedConversation.ToList();
+        Assert.Equal(2, trimmedConversationList.Count);
+        Assert.Equal(message3, trimmedConversationList[0]);
+        Assert.Equal(message2, trimmedConversationList[1]);
+        Assert.Equal(0, remainingTokenBudget);
+    }
+}
```
peterwald, line 198: These tests are great. Can we add a test that checks that the order of the trimmed messages is the same as the input order?

src/Libraries/Microsoft.Extensions.AI.Evaluation/ChatMessageExtensions.cs
shyamnamboodiripad: Ah good catch - will fix!

src/Libraries/Microsoft.Extensions.AI.Evaluation/Microsoft.Extensions.AI.Evaluation.csproj
shyamnamboodiripad: This was originally needed to make the `[Experimental]` attribute available in the legacy target frameworks. However, I have removed this now that we decided not to ship the helper on which the attribute was added before.

src/Libraries/Microsoft.Extensions.AI.Evaluation/ChatMessageExtensions.cs
shyamnamboodiripad: This no longer applies since we decided to remove this helper and avoid the dependency on `Microsoft.ML.Tokenizers` entirely.

test/Libraries/Microsoft.Extensions.AI.Evaluation.Tests/ConversationTrimmingTests.cs
shyamnamboodiripad: This no longer applies since we decided to remove the helper and the above tests to avoid the dependency on Microsoft.ML.Tokenizers entirely.

peterwald (approved):

PR #6321, opened 2025-04-21
Fix Process Metric calculation in CgroupsV2

nivebb8:
Missed updating the process metric in first check-in. Updating the same to use correct calculations behind the flag, and to ensure CpuUtilization() is not invoked at all.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6321)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
@@ -17,8 +17,8 @@ internal sealed class LinuxUtilizationProvider : ISnapshotProvider
     private const double CpuLimitThreshold110Percent = 1.1;
 
     // Meters to track CPU utilization threshold exceedances
-    private readonly Counter<long> _cpuUtilizationLimit100PercentExceededCounter;
-    private readonly Counter<long> _cpuUtilizationLimit110PercentExceededCounter;
+    private readonly Counter<long>? _cpuUtilizationLimit100PercentExceededCounter;
+    private readonly Counter<long>? _cpuUtilizationLimit110PercentExceededCounter;
 
     private readonly object _cpuLocker = new();
     private readonly object _memoryLocker = new();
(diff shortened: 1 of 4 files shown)


evgenyfedorov2 (changes requested):
Why do we need this change?
The `ResourceUtilizationInstruments.ProcessCpuUtilization` metric is deprecated (unofficially). `ResourceUtilizationInstruments.ContainerCpuRequestUtilization` is the replacement and it already has the right calculation logic

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -79,15 +79,16 @@ public LinuxUtilizationProvider(IOptions<ResourceMonitoringOptions> options, ILi
             cpuRequest = _parser.GetCgroupRequestCpuV2();
             _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ContainerCpuLimitUtilization, observeValue: () => CpuUtilizationWithoutHostDelta() / cpuLimit, unit: "1");
             _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ContainerCpuRequestUtilization, observeValue: () => CpuUtilizationWithoutHostDelta() / cpuRequest, unit: "1");
+            _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ProcessCpuUtilization, observeValue: () => CpuUtilizationWithoutHostDelta() / cpuRequest, unit: "1");
```
evgenyfedorov2, line 82: I think we can simply remove this line. For new customers / services, the recommended metric is `ResourceUtilizationInstruments.ContainerCpuRequestUtilization`

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -70,24 +77,28 @@ public LinuxUtilizationProvider(IOptions<ResourceMonitoringOptions> options, ILi
         var meter = meterFactory.Create(ResourceUtilizationInstruments.MeterName);
 #pragma warning restore CA2000 // Dispose objects before losing scope
 
+        // Initialize the counters
+        _cpuUtilizationLimit100PercentExceededCounter = meter.CreateCounter<long>("cpu_utilization_limit_100_percent_exceeded");
+        _cpuUtilizationLimit110PercentExceededCounter = meter.CreateCounter<long>("cpu_utilization_limit_110_percent_exceeded");
+
```
evgenyfedorov2, line 83: This should be inside the `if` condition starting on line 84

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/Log.cs
```diff
@@ -44,4 +44,11 @@ public static partial void CpuUsageDataV2(
         long previousCgroupCpuTime,
         double actualElapsedNanoseconds,
         double cpuCores);
+
+    [LoggerMessage(5, LogLevel.Debug,
+        "Counters for CgroupV2 {message}, Counter = {counterValue}")]
+    public static partial void CounterMessage(
+        ILogger logger,
+        string message,
```
evgenyfedorov2, line 52: We don't need the message parameter. The text in the LoggerMessage is sufficient.

evgenyfedorov2 (approved):

PR #6325, opened 2025-04-23
Enhance Function Invocation Extensibility for Microsoft.Extensions.AI

RogerBarreto:
## Proposal for MEAI Function Calling specialization

As my recent investigation in the Semantic Kernel prepareing it for doing function calling on top of MEAI abstractions, some changes would be required for seamless support. 

Semantic Kernel's implementation changes made below as part for the investigation.

- https://github.com/microsoft/semantic-kernel/pull/11678
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6325)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -547,51 +549,61 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
+            IList<ChatMessage> addedMessages = CreateResponseMessages([result]);
+            ThrowIfNoFunctionResultsAdded(addedMessages);
+            UpdateConsecutiveErrorCountOrThrow(addedMessages, ref consecutiveErrorCount);
 
-            messages.AddRange(added);
-            return (result.ShouldTerminate, consecutiveErrorCount, added);
+            foreach (var addedMessage in addedMessages)
+            {
+                messages.Add(addedMessage);
+            }
```
SteveSandersonMS, line 588: Why do this instead of `AddRange`?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -601,27 +613,37 @@ select ProcessFunctionCallAsync(
...
-            IList<ChatMessage> added = CreateResponseMessages(results);
-            ThrowIfNoFunctionResultsAdded(added);
-            UpdateConsecutiveErrorCountOrThrow(added, ref consecutiveErrorCount);
+            IList<ChatMessage> addedMessages = CreateResponseMessages(results);
+            ThrowIfNoFunctionResultsAdded(addedMessages);
+            UpdateConsecutiveErrorCountOrThrow(addedMessages, ref consecutiveErrorCount);
+            foreach (var addedMessage in addedMessages)
+            {
+                messages.Add(addedMessage);
+            }
```
SteveSandersonMS, line 628: Why do this instead of `AddRange`?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: As this method is not private anymore, `IList` doesn't have `AddRange`.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Answered above

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -601,27 +613,37 @@ select ProcessFunctionCallAsync(
...
 
-    private void UpdateConsecutiveErrorCountOrThrow(IList<ChatMessage> added, ref int consecutiveErrorCount)
+#pragma warning disable CA1851 // Possible multiple enumerations of 'IEnumerable' collection
+    /// <summary>
+    /// Updates the consecutive error count, and throws an exception if the count exceeds the maximum.
+    /// </summary>
+    /// <param name="added">Added messages.</param>
+    /// <param name="consecutiveErrorCount">Consecutive error count.</param>
+    /// <exception cref="AggregateException">Thrown if the maximum consecutive error count is exceeded.</exception>
+    protected void UpdateConsecutiveErrorCountOrThrow(IList<ChatMessage> added, ref int consecutiveErrorCount)
```
SteveSandersonMS, line 646: This doesn't feel like something we'd want to make part of the API surface because it's quite a low-level implementation detail.

Could you explain the usage scenario? Maybe we can find a way to enable the scenario without leaking out this aspect of the implementation.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -648,11 +670,12 @@ private void UpdateConsecutiveErrorCountOrThrow(IList<ChatMessage> added, ref in
             consecutiveErrorCount = 0;
         }
     }
+#pragma warning restore CA1851
 
     /// <summary>
     /// Throws an exception if <see cref="CreateResponseMessages"/> doesn't create any messages.
     /// </summary>
-    private void ThrowIfNoFunctionResultsAdded(IList<ChatMessage>? messages)
+    protected void ThrowIfNoFunctionResultsAdded(IList<ChatMessage>? messages)
```
SteveSandersonMS, line 678: Same point as for `UpdateConsecutiveErrorCountOrThrow`

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -840,6 +869,19 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
         return result;
     }
 
+    /// <summary>This method will invoke the fuction within the try block.</summary>
```
SteveSandersonMS, line 872: Is this true? I don't see any `try` block.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -840,6 +869,19 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+    /// <param name="context">The function invocation context.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>The function invocation context and result.</returns>
+    protected virtual async Task<(FunctionInvocationContext context, object? result)> TryInvokeFunctionAsync(FunctionInvocationContext context, CancellationToken cancellationToken)
```
SteveSandersonMS, line 876: It would be odd for both `InvokeFunctionAsync` and `TryInvokeFunctionAsync` to become part of the public API surface, as it's very unclear (from the naming) how they differ.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -405,7 +407,7 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
 
             if (modeAndMessages.ShouldTerminate)
             {
-                break;
+                yield break;
```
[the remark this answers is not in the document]
RogerBarreto: Not intended, the code i was working on was from this commit.

https://github.com/dotnet/extensions/blob/84d09b794d994435568adcbb85a981143d4f15cb/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L392

Some changes needs to be reverted to acommodate latest changes after I start the investigation.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -667,12 +690,17 @@ private void ThrowIfNoFunctionResultsAdded(IList<ChatMessage>? messages)
     /// <param name="iteration">The iteration number of how many roundtrips have been made to the inner client.</param>
     /// <param name="functionCallIndex">The 0-based index of the function being called out of <paramref name="callContents"/>.</param>
     /// <param name="captureExceptions">If true, handles function-invocation exceptions by returning a value with <see cref="FunctionInvocationStatus.Exception"/>. Otherwise, rethrows.</param>
+    /// <param name="isStreaming">Whether the function calls are being processed in a streaming context.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A value indicating how the caller should proceed.</returns>
-    private async Task<FunctionInvocationResult> ProcessFunctionCallAsync(
-        List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> callContents,
-        int iteration, int functionCallIndex, bool captureExceptions, CancellationToken cancellationToken)
+    protected virtual async Task<FunctionInvocationResult> ProcessFunctionCallAsync(
```
SteveSandersonMS, line 696: Does this definitely have to be made `public virtual`?

Looking at your implementation of `KernelFunctionInvokingChatClient`, it seems like you override `ProcessFunctionCallsAsync` and that calls `ProcessFunctionCallAsync`, which never calls `base.ProcessFunctionCallsAsync`. I might be missing something but it seems like this pattern doesn't require `ProcessFunctionCallAsync` to become protected or virtual.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: THis can be put back to hidden, the sole reason why this was exposed was because I need to change the behavior of `ProcessFunctionCallsAsync`, where I clarified in the Semantic Kernel PR here:

https://github.com/microsoft/semantic-kernel/pull/11678#discussion_r2055257722

@jeffhandley Had also the context.

Basically we terminate the function calling before the loop finishes when its not concurrent, if we can add a Policy to allow it in the Function Invoking client, this method would still be overriden but SK changes would be very small something like.

```csharp
protected override async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
        ChatResponse response, List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount, bool isStreaming, CancellationToken cancellationToken)
{
    // Prepare the options for the next auto function invocation iteration.
    UpdateOptionsForAutoFunctionInvocation(ref options!, response.Messages.Last().ToChatMessageContent(), isStreaming: false);

    var result = await base.ProcessFunctionCallsAsync(...)

    // Clear the auto function invocation options.
    ClearOptionsForAutoFunctionInvocation(ref options);

    return result;
}

```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
SteveSandersonMS: Maybe the right design is for us to make `InvokeFunctionAsync` *not* be protected/virtual and have only the inner thing (TryInvokeFunctionAsync as you've called it) be protected/virtual.

The other thing just adds logging (and sets context), so we could make that private and call it `LogAndInvokeFunction` or something, then the new method you've added could have the nicer name `InvokeFunctionAsync` and be protected and virtual.

I think what you've added here is a better thing to be protected/virtual since if someone overrides this, they likely *only* want to change the actual invocation and not change anything about logging.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: This can be gone if we can configure the behavior of terminating early function invocation while looping.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Line 831 happens within a `try` block, this description can be improved.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -547,51 +549,61 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
     /// <param name="functionCallContents">The function call contents representing the functions to be invoked.</param>
     /// <param name="iteration">The iteration number of how many roundtrips have been made to the inner client.</param>
     /// <param name="consecutiveErrorCount">The number of consecutive iterations, prior to this one, that were recorded as having function invocation errors.</param>
+    /// <param name="isStreaming">Whether the function calls are being processed in a streaming context.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A value indicating how the caller should proceed.</returns>
-    private async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
-        List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount, CancellationToken cancellationToken)
+    protected virtual async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
```
SteveSandersonMS, line 561: Looking at your override of `ProcessFunctionCallsAsync` in `KernelFunctionInvokingChatClient`, I see there's still a lot of duplicated code. Is there any way to split out the SK-specific aspect of it so that less duplication is needed? That might make it possible to avoid needing to put `ThrowIfNoFunctionResultsAdded` and `ThrowIfNoFunctionResultsAdded` into the public API.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
SteveSandersonMS: Oh, thanks. I interpreted the XML comment to mean that *this* method contains a try block. It's a bit odd for the XML docs here to refer to code in the caller, given that the method doesn't know who is going to call it.

SteveSandersonMS:
Thanks @RogerBarreto! This looks like a good direction to go in. While there are still some details to streamline, I'm hopeful we can figure it out pretty soon and make this change.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: > Maybe the right design is for us to make InvokeFunctionAsync not be protected/virtual and have only the inner thing (TryInvokeFunctionAsync as you've called it) be protected/virtual.

I agree here, this was the original method and I didn't want to introduce breaking changes, but, yes, if we can have the current `InvokeFunctionAsync` internalized and the `Try` become the new `InvokeFunctionAsync`, thats correct, in our specialized client we didn't need to override this (InvokeFunctionAsync).

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
SteveSandersonMS: > if we can add a Policy to allow it in the Function Invoking client

That sounds reasonable. Could you update the PR to propose a way to do that?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: That's a good point, I might be able to reuse the function calling entirely if a CreateContext override allow me to do the 

```
var context = new AutoFunctionInvocationContext(options)
        {
            AIFunction = function,
            Arguments = new KernelArguments(callContent.Arguments ?? new Dictionary<string, object?>()) { Services = this.FunctionInvocationServices },
            Messages = messages,
            CallContent = callContent,
            Iteration = iteration,
            FunctionCallIndex = functionCallIndex,
            FunctionCount = callContents.Count,
            IsStreaming = isStreaming
        };
```

All else is similar.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Absolutely, I had to override this method just so I could interrupt the function calling as soon I get the first `ShouldTerminate` from the for-loop when the function calling is performed in a **non concurrent invocation**.

Described here:

https://github.com/microsoft/semantic-kernel/pull/11678/files#diff-9ef864b76a36e09adf87b47254aa4be3902332f986f08289ce94ea930970febaR115-R117

If we can add a Policy/Flag into the invocation behavior that I don't need to rewrite the method, that implementation will be gone.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,36 +21,42 @@ namespace Microsoft.Extensions.AI;
...
-public sealed class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
+public class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
 {
     /// <summary>The nominal arguments.</summary>
     private readonly Dictionary<string, object?> _arguments;
 
     /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
-    public AIFunctionArguments()
+    /// <param name="comparer">The optional <see cref="IEqualityComparer{T}"/> to use for key comparisons.</param>
+    public AIFunctionArguments(IEqualityComparer<string>? comparer = null)
```
jeffhandley, line 31: This would be a binary breaking change. I recommend we add this by way of a new overload.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,36 +21,42 @@ namespace Microsoft.Extensions.AI;
...
-    /// instance. A <see langword="null"/> <paramref name="arguments"/> is treated as an
-    /// empty dictionary.
+    /// already a <see cref="Dictionary{TKey, TValue}"/> and no <paramref name="comparer"/> is specified,
+    /// in which case all dictionary operations on this instance will be routed directly to that instance.
+    /// If <paramref name="arguments"/> is not a dictionary or a <paramref name="comparer"/> is specified,
+    /// a shallow clone of its data will be used to populate this instance.
+    /// A <see langword="null"/> <paramref name="arguments"/> is treated as an empty dictionary.
     /// </remarks>
-    public AIFunctionArguments(IDictionary<string, object?>? arguments)
+    public AIFunctionArguments(IDictionary<string, object?>? arguments, IEqualityComparer<string>? comparer = null)
```
jeffhandley, line 50: This would be a binary breaking change. I recommend we add this by way of a new overload.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
jeffhandley: I also like the approach of adding a public property to `FunctionInvokingChatClient` to control this behavior. I don't like any of the property names I'm thinking of though:

1. `StopOnFirstTermination`
2. `MaxAllowedTerminationRequests`
3. `AllowEarlyTermination`

Since only one of `AllowConcurrentInvocation` and `AllowEarlyTermination` can be respected as `true`, I was tempted to suggest an enum that provides the allowed combinations, but that seems even worse.

With two bools, I think we'd just want the chat client to throw an `InvalidOperationException` if both are set to true.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,36 +21,42 @@ namespace Microsoft.Extensions.AI;
...
 /// </remarks>
-public sealed class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
+public class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
 {
     /// <summary>The nominal arguments.</summary>
     private readonly Dictionary<string, object?> _arguments;
 
     /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
-    public AIFunctionArguments()
+    /// <param name="comparer">The optional <see cref="IEqualityComparer{T}"/> to use for key comparisons.</param>
```
stephentoub, line 30: This should state that ordinal is used by default.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
stephentoub: What's driving the need to override the comparison? Is this based on wanting to use OrdinalIgnoreCase for SK?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,36 +21,42 @@ namespace Microsoft.Extensions.AI;
...
-            arguments as Dictionary<string, object?> ??
-            new Dictionary<string, object?>(arguments);
+        _arguments = comparer is null
+            ? arguments is null
+                ? []
+                : arguments as Dictionary<string, object?> ??
+                  new Dictionary<string, object?>(arguments)
+            : arguments is null
+                ? new Dictionary<string, object?>(comparer)
+                : new Dictionary<string, object?>(arguments, comparer);
```
stephentoub, line 59: What if the comparer matches the one already used by the dictionary?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -99,7 +106,7 @@ public object? this[string key]
     public void Clear() => _arguments.Clear();
 
     /// <inheritdoc />
-    bool ICollection<KeyValuePair<string, object?>>.Contains(KeyValuePair<string, object?> item) =>
+    public bool Contains(KeyValuePair<string, object?> item) =>
```
stephentoub, line 109: Ditto

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvocationContext.cs
```diff
@@ -94,4 +94,9 @@ public IList<ChatMessage> Messages
     /// more function call requests in responses.
     /// </remarks>
     public bool Terminate { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the context is happening in a streaming scenario.
```
stephentoub, line 99:
```suggestion
    /// Gets or sets a value indicating whether the function invocation is occurring as part of a <see cref="IChatClient.GetStreamingResponseAsync"/> call as opposed to a <see cref="IChatClient.GetResponseAsync"/> call.
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -17,6 +17,8 @@
 #pragma warning disable CA2213 // Disposable fields should be disposed
 #pragma warning disable EA0002 // Use 'System.TimeProvider' to make the code easier to test
 #pragma warning disable SA1202 // 'protected' members should come before 'private' members
+#pragma warning disable S107 // Methods should not have too many parameters
+#pragma warning disable CA1002 // Do not expose generic lists
```
stephentoub, line 21: Nit: we've tried to keep these grouped by prefix

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -49,8 +51,8 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
     /// <summary>The <see cref="FunctionInvocationContext"/> for the current function invocation.</summary>
     private static readonly AsyncLocal<FunctionInvocationContext?> _currentContext = new();
 
-    /// <summary>Optional services used for function invocation.</summary>
-    private readonly IServiceProvider? _functionInvocationServices;
+    /// <summary>Gets services used for function invocation.</summary>
```
stephentoub, line 54:
```suggestion
    /// <summary>Gets the <see cref="IServiceProvider"/> specified when constructing the <see cref="FunctionInvokingChatClient"/>, if any.</summary>
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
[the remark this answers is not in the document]
stephentoub: This is actually broken; it needs to be `break;` rather than `yield break;`, or else the `AddUsageTags` won't be called in this case.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: I'm not a fan of this signature. If we have to make a change here, I'd like to better understand what's needed and come up with a better method that doesn't require such a complicated and non-evolvable return tuple.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: The 99.9% case will be `List<T>`, and AddRange is significantly more efficient than adding elements one at a time. If we have to accept `IList`, we should still use AddRange if the list type tests as `List<T>`.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: This doesn't need to be protected regardless. It's just a throw helper; there's no need for it to be exposed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -80,7 +87,7 @@ public object? this[string key]
     public int Count => _arguments.Count;
 
     /// <inheritdoc />
-    bool ICollection<KeyValuePair<string, object?>>.IsReadOnly => false;
+    public bool IsReadOnly => false;
```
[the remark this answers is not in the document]
RogerBarreto: Was because of the waring, will ignore the warning.

NuGet.config
jeffhandley, line 1: This file will need to be reverted.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,15 +22,44 @@ namespace Microsoft.Extensions.AI;
...
+    /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
+    /// <param name="comparer">The <see cref="IEqualityComparer{T}"/> to use for key comparisons.</param>
+    public AIFunctionArguments(IEqualityComparer<string> comparer)
+    {
+        _ = Throw.IfNull(comparer);
+
+        _arguments = new Dictionary<string, object?>(comparer);
+    }
+
     /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
```
jeffhandley, line 39:
```suggestion
    /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class, and uses the default comparer for key comparisons.</summary>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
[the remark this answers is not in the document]
jeffhandley: Need to revert this and ignore the warning.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
jeffhandley: Need to revert this and ignore the warning.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -465,8 +467,8 @@ private static void FixupHistories(
         }
         else if (lastIterationHadConversationId)
         {
-            // In the very rare case where the inner client returned a response with a thread ID but then
-            // returned a subsequent response without one, we want to reconstitue the full history. To do that,
+            // In the very rare case where the inner client returned a response with a conversation ID but then
+            // returned a subsequent response without one, we want to reconstitute the full history. To do that,
```
jeffhandley, line 471: Thanks for cleaning up these comments

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
jeffhandley: I think we want to revert this back to private to not expose this signature.

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/FunctionInvokingChatClientTests.cs
```diff
@@ -744,7 +744,7 @@ void AssertInvocationContext(FunctionInvocationContext context, int iteration, b
     }
 
     [Fact]
-    public async Task CanResumeFunctionCallingAfterTermination()
+    public async Task HaltFunctionCallingAfterTermination()
```
jeffhandley, line 747: Oh, cool. It looks like this test was likely failing before being fixed; right?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
RogerBarreto: I don't have access to the comparer in the `IDictionary` interface to be able to compare or am I missing something here?

test/Libraries/Microsoft.Extensions.AI.Tests/ChatCompletion/FunctionInvokingChatClientTests.cs
RogerBarreto: Right, missed this one and adjusted for early termination.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: I still use this method in Semantic Kernel, very simplified now, where I need to perform ChatOptions operations before and after, I'm happy if we can expose a ChatOptions transformer that I can rely before and after the fact (`next()`), or just go with what is there.

```csharp
    /// <inheritdoc/>
    protected override async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
        ChatResponse response, IList<ChatMessage> messages, ChatOptions options, IList<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount, bool isStreaming, CancellationToken cancellationToken)
    {
        // Prepare the options for the next auto function invocation iteration.
        UpdateOptionsForAutoFunctionInvocation(ref options!, response.Messages.Last().ToChatMessageContent(), isStreaming: false);

        // We must add a response for every tool call, regardless of whether we successfully executed it or not.
        // If we successfully execute it, we'll add the result. If we don't, we'll add an error.
        var result = await base.ProcessFunctionCallsAsync(response, messages, options, functionCallContents, iteration, consecutiveErrorCount, isStreaming, cancellationToken).ConfigureAwait(false);

        // Clear the auto function invocation options.
        ClearOptionsForAutoFunctionInvocation(ref options);

        return result;
    }
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
RogerBarreto: Done

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: I'm completely open to rethink this signature and anything we can come up with and I agree, tuple is not the best for a overridable method 👍🏻

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -4,10 +4,13 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using Microsoft.Shared.Diagnostics;
 
 #pragma warning disable SA1111 // Closing parenthesis should be on line of last parameter
 #pragma warning disable SA1112 // Closing parenthesis should be on line of opening parenthesis
 #pragma warning disable SA1114 // Parameter list should follow declaration
+#pragma warning disable S3358 // Extract this nested ternary operation into an independent statement.
+#pragma warning disable S1698 // Consider using 'Equals' if value comparison is intended.
```
jeffhandley, line 13: This is for the comparer equality check. I'd like to either move this suppression down around that line of code only. Would `.Equals(comparer)` actually be better though, @stephentoub?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
jeffhandley: @stephentoub - I presume a struct would be preferred here; right? Something like this?

```csharp

public partial class FunctionInvokingChatClient
{
    protected struct FunctionCallProcessState
    {
        public bool ShouldTerminate { get; }
        public int ConsecutiveErrorCount { get; }
        public IList<ChatMessage> MessagesAdded { get; }
    }

    protected virtual async Task<FunctionCallProcessState>(
        ChatResponse response, IList<ChatMessage> messages, ChatOptions options, IList<FunctionCallContent> 
        functionCallContents, int iteration, int consecutiveErrorCount,
        bool isStreaming, CancellationToken cancellationToken);
}
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,15 +23,50 @@ namespace Microsoft.Extensions.AI;
...
+    /// The <paramref name="arguments"/> reference will be stored if the instance is already a
+    /// <see cref="Dictionary{TKey, TValue}"/> with the same <see cref="IEqualityComparer{T}"/>,
+    /// in which case all dictionary operations on this instance will be routed directly to that instance.
+    /// A shallow clone of the provided <paramref name="arguments"/> will be used to populate this instance.
+    /// A <see langword="null"/> <paramref name="arguments"/> is treated as an empty dictionary.
+    /// </remarks>
+    public AIFunctionArguments(IDictionary<string, object?>? arguments, IEqualityComparer<string> comparer)
+    {
+        _ = Throw.IfNull(comparer);
+
```
stephentoub, line 63:
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,15 +23,50 @@ namespace Microsoft.Extensions.AI;
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AIFunctionArguments"/> class containing
+    /// the specified <paramref name="arguments"/>.
+    /// </summary>
+    /// <param name="arguments">The arguments represented by this instance.</param>
+    /// <param name="comparer">The <see cref="IEqualityComparer{T}"/> to be used.</param>
+    /// <remarks>
+    /// The <paramref name="arguments"/> reference will be stored if the instance is already a
+    /// <see cref="Dictionary{TKey, TValue}"/> with the same <see cref="IEqualityComparer{T}"/>,
```
stephentoub, line 55:
```suggestion
    /// <see cref="Dictionary{TKey, TValue}"/> with the same <see cref="IEqualityComparer{T}"/> or if <paramref name="arguments"/> is <see langword="null" />,
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -79,8 +117,12 @@ public object? this[string key]
     /// <inheritdoc />
     public int Count => _arguments.Count;
 
+#pragma warning disable CA1033 // Make 'AIFunctionArguments' sealed
+#pragma warning disable S4039 // Make 'AIFunctionArguments' sealed
```
stephentoub, line 121: You can just suppress these at the top of the file

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: What do UpdateOptionsForAutoFunctionInvocation and ClearOptionsForAutoFunctionInvocation do? Could they instead be done in an override of InvokeFunctionAsync, i.e. around each function invocation rather than around each group of function invocations?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -547,80 +549,118 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A value indicating how the caller should proceed.</returns>
-    private async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
-        List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount, CancellationToken cancellationToken)
+    protected virtual async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
+        ChatResponse response, IList<ChatMessage> messages, ChatOptions options, IList<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount,
+        bool isStreaming, CancellationToken cancellationToken)
     {
+        _ = Throw.IfNull(messages);
+        _ = Throw.IfNull(functionCallContents);
```
stephentoub, line 566: (I'd like to avoid exposing this method as protected, but in case we must...)

If we're validating inputs because the it's now protected, why only these two parameters? i.e. don't we also need to validate response and options?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -547,80 +549,118 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
+                messagesList.AddRange(addedMessages);
+            }
+            else
             {
-                shouldTerminate = shouldTerminate || fir.ShouldTerminate;
+                foreach (var addedMessage in addedMessages)
+                {
+                    messages.Add(addedMessage);
+                }
             }
```
stephentoub, line 651: If we don't already have one somewhere that does this (I think we might), we should move this into a helper and dedup the occurrences so that its just `messagesList.AddRange(addedMessages)`, with that AddRange extension method doing the type test and the if/else.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -667,12 +708,17 @@ private void ThrowIfNoFunctionResultsAdded(IList<ChatMessage>? messages)
     /// <param name="iteration">The iteration number of how many roundtrips have been made to the inner client.</param>
     /// <param name="functionCallIndex">The 0-based index of the function being called out of <paramref name="callContents"/>.</param>
     /// <param name="captureExceptions">If true, handles function-invocation exceptions by returning a value with <see cref="FunctionInvocationStatus.Exception"/>. Otherwise, rethrows.</param>
+    /// <param name="isStreaming">Whether the function calls are being processed in a streaming context.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A value indicating how the caller should proceed.</returns>
-    private async Task<FunctionInvocationResult> ProcessFunctionCallAsync(
-        List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> callContents,
-        int iteration, int functionCallIndex, bool captureExceptions, CancellationToken cancellationToken)
+    protected virtual async Task<FunctionInvocationResult> ProcessFunctionCallAsync(
```
stephentoub, line 714: I thought we were reverting this to be private again?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -840,6 +887,19 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+    /// <param name="context">The function invocation context.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>The function invocation context and result.</returns>
+    protected virtual async Task<(FunctionInvocationContext context, object? result)> InvokeFunctionCoreAsync(FunctionInvocationContext context, CancellationToken cancellationToken)
+    {
+        _ = Throw.IfNull(context);
+
+        var result = await context.Function.InvokeAsync(context.Arguments, cancellationToken);
+
+        return (context, result);
```
stephentoub, line 900: Why does this need to return a FunctionInvocationContext?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -775,7 +822,7 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The result of the function invocation, or <see langword="null"/> if the function invocation returned <see langword="null"/>.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="context"/> is <see langword="null"/>.</exception>
-    protected virtual async Task<object?> InvokeFunctionAsync(FunctionInvocationContext context, CancellationToken cancellationToken)
+    protected async Task<object?> InvokeFunctionAsync(FunctionInvocationContext context, CancellationToken cancellationToken)
```
stephentoub, line 825: You're moving the virtualness to another method in order to preserve the diagnostics? Does this still need to be protected at that point, or could this be private again, and then remove "Core" from the other method's name?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -868,7 +928,15 @@ private static TimeSpan GetElapsedTime(long startingTimestamp) =>
...
-        internal FunctionInvocationResult(bool shouldTerminate, FunctionInvocationStatus status, FunctionCallContent callContent, object? result, Exception? exception)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FunctionInvocationResult"/> class.
+        /// </summary>
+        /// <param name="shouldTerminate">Indicates whether the caller should terminate the processing loop.</param>
+        /// <param name="status">Indicates the status of the function invocation.</param>
+        /// <param name="callContent">Contains information about the function call.</param>
+        /// <param name="result">The result of the function call.</param>
+        /// <param name="exception">The exception thrown by the function call, if any.</param>
+        public FunctionInvocationResult(bool shouldTerminate, FunctionInvocationStatus status, FunctionCallContent callContent, object? result, Exception? exception)
```
stephentoub, line 939: Why does this need to be public? I don't want to expose this ctor. Something else needs to construct these? If so, I'd rather just have the default parameterless ctor and make all of the properties settable.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -890,7 +958,7 @@ internal FunctionInvocationResult(bool shouldTerminate, FunctionInvocationStatus
         public Exception? Exception { get; }
 
         /// <summary>Gets a value indicating whether the caller should terminate the processing loop.</summary>
-        internal bool ShouldTerminate { get; }
+        public bool ShouldTerminate { get; }
```
stephentoub, line 961: Remove the word "Should"? We just call this `Terminate` on FunctionInvocationContext.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
jeffhandley: `UpdateOptionsForAutoFunctionInvocation` adds `AdditionalProperties` for `IsStreaming` and the `ChatMessageContent`, and `ClearOptionsForAutoFunctionInvocation` removes those from the dictionary. Making this method virtual allows for a clean override that calls the base method in the middle.

https://github.com/microsoft/semantic-kernel/blob/b62b51bb9a88370a212e07076a260e87939447fa/dotnet/src/SemanticKernel.Abstractions/AI/ChatClient/KernelFunctionInvokingChatClient.cs#L519-L548

I suspect you're right, @stephentoub, that there's another way to approach this that doesn't result in needing to expose that protected state.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,15 +23,50 @@ namespace Microsoft.Extensions.AI;
...
 /// </remarks>
-public sealed class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
+public class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
 {
     /// <summary>The nominal arguments.</summary>
     private readonly Dictionary<string, object?> _arguments;
 
     /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
+    /// <param name="comparer">The <see cref="IEqualityComparer{T}"/> to use for key comparisons.</param>
+    public AIFunctionArguments(IEqualityComparer<string> comparer)
```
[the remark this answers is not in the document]
RogerBarreto: I intentionally made this method not expecting a null because we have the empty constructor, so this wound't be a binary breaking change, wondering what's the point of having a "required" nullable comparer.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: I thought about this, but didn't want to introduce more code for such a unit need, I'm happy adding an extension method though.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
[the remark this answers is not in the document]
stephentoub: It's composable. If you're in a method that receives an nullable comparer, a non-nullable design forces the caller to have to accommodate and call the right constructor. Note that Dictionary, for example, has a parameterless ctor, and a ctor that accepts a comparer, and the comparer is allowed to be null.

This needs to be nullable.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Since context is modified by reference we don't need to send it back.

As the caller `CurrentContext = context;` right before calling this.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: It'll be less code, as you have multiple copies of this pattern now.

It doesn't need to be public, nor does it even need to be an extension... it could just be a private helper.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
stephentoub: I don't understand. Why does this method need to return the context at all?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: This is a must need for our specialization as we construct the `context` and make the `AutoInvocationFilter` work.

https://github.com/microsoft/semantic-kernel/pull/11678/files#diff-9ef864b76a36e09adf87b47254aa4be3902332f986f08289ce94ea930970febaR291

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: When overriding the `ProcessFunctionCallAsync` I need a way to return a `FunctionInvocationResult`, can be a factory, static `Create`, any other option.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Will do.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: we don't.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: With latest changes this method was moved to private.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: moved to `private`

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Returning only the `result`, in latest changes.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Reverted with latest changes.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -552,75 +554,91 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
+    /// <param name="isStreaming">Whether the function calls are being processed in a streaming context.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A value indicating how the caller should proceed.</returns>
     private async Task<(bool ShouldTerminate, int NewConsecutiveErrorCount, IList<ChatMessage> MessagesAdded)> ProcessFunctionCallsAsync(
-        List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount, CancellationToken cancellationToken)
+        List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> functionCallContents, int iteration, int consecutiveErrorCount,
+        bool isStreaming, CancellationToken cancellationToken)
     {
+        _ = Throw.IfNull(messages);
+        _ = Throw.IfNull(functionCallContents);
```
stephentoub, line 565: Do we still need this arg validation now? Seems like these lines can be reverted

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvocationContext.cs
```diff
@@ -94,4 +94,10 @@ public IList<ChatMessage> Messages
     /// more function call requests in responses.
```
stephentoub, line 94: We should update this to also mention the sequential behavior, e.g. adding to the end
```
If multiple function call requests are issued as part of a single iteration (a single response from the inner <see cref="IChatClient"/>), setting  <see cref="Terminate" /> to <see langword="true" /> may also prevent subsequent requests within that same iteration from being processed.
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -667,39 +686,43 @@ private void ThrowIfNoFunctionResultsAdded(IList<ChatMessage>? messages)
...
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A value indicating how the caller should proceed.</returns>
     private async Task<FunctionInvocationResult> ProcessFunctionCallAsync(
         List<ChatMessage> messages, ChatOptions options, List<FunctionCallContent> callContents,
-        int iteration, int functionCallIndex, bool captureExceptions, CancellationToken cancellationToken)
+        int iteration, int functionCallIndex, bool captureExceptions, bool isStreaming, CancellationToken cancellationToken)
     {
+        _ = Throw.IfNull(messages);
+        _ = Throw.IfNull(options);
+        _ = Throw.IfNull(callContents);
```
stephentoub, line 698: Can we revert this arg validation?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -840,6 +863,17 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
         return result;
     }
 
+    /// <summary>This method will invoke the function within the try block.</summary>
+    /// <param name="context">The function invocation context.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>The function result.</returns>
+    protected virtual async Task<object?> InvokeFunctionAsync(FunctionInvocationContext context, CancellationToken cancellationToken)
```
stephentoub, line 870: Does this need to be `async`? Could we instead remove the `async` and the `await` from the impl?

stephentoub, line 870: While we're making changes here, probably good to change this to return ValueTask.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
stephentoub: Which line is this referring to? (In general, I'm not a fan of many of these "S" rules; they're not from the .NET SDK but some other analyzer library that's being pulled into extensions, and there are _lots_ of false positives.)

stephentoub: If this is about https://github.com/dotnet/extensions/pull/6325/files#diff-7089a7412a27a3cecae56c30dbc91b04277c83d6aa5db5302d3616eb94e6aab7R65, == is fine, but it could also be ReferenceEquals if that would make folks happier.

stephentoub (commented):
Looking much better! Almost there.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: Yep, removing from all non nullable argument from private methods.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -552,75 +554,91 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
+            messages.AddRange(addedMessages);
 
-            messages.AddRange(added);
-            return (result.ShouldTerminate, consecutiveErrorCount, added);
+            return (result.Terminate, consecutiveErrorCount, addedMessages);
         }
         else
         {
-            FunctionInvocationResult[] results;
+            List<FunctionInvocationResult> results = [];
```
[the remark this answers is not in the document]
RogerBarreto: Ending the loop early I have a smaller result, before was take for granted the array size as number of function results.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
RogerBarreto: I thought about that and wasn't sure what would be the impact of `Task` for `ValueTask` implementers.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -20,15 +24,48 @@ namespace Microsoft.Extensions.AI;
...
-public sealed class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
+public class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictionary<string, object?>
 {
     /// <summary>The nominal arguments.</summary>
     private readonly Dictionary<string, object?> _arguments;
 
-    /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class.</summary>
+    /// <summary>Initializes a new instance of the <see cref="AIFunctionArguments"/> class, and uses the default comparer for key comparisons.</summary>
+    /// <remarks>The <see cref="IEqualityComparer{T}"/> is ordinal by default.</remarks>
     public AIFunctionArguments()
```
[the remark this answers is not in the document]
RogerBarreto: User `this(null, null)` to avoid named parameter and clash with the `this(arguments)`

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
```diff
@@ -32,8 +32,27 @@ public class AIFunctionArguments : IDictionary<string, object?>, IReadOnlyDictio
...
+    /// The <paramref name="arguments"/> reference will be stored if the instance is
+    /// already a <see cref="Dictionary{TKey, TValue}"/>, in which case all dictionary
+    /// operations on this instance will be routed directly to that instance. If <paramref name="arguments"/>
+    /// is not a dictionary, a shallow clone of its data will be used to populate this
+    /// instance. A <see langword="null"/> <paramref name="arguments"/> is treated as an
+    /// empty parameters dictionary.
+    /// The <see cref="IEqualityComparer{T}"/> is ordinal by default.
+    /// </remarks>
+    public AIFunctionArguments(IDictionary<string, object?>? arguments)
+        : this(arguments, null)
```
stephentoub, line 54: GH won't let me comment on it, but the below method can also delegate, e.g.
```C#
public AIFunctionArguments(IEqualityComparer<string>? comparer) : this(null, comparer)
{
}
```

RogerBarreto: Absolutely, missed 😊

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionArguments.cs
RogerBarreto: Leaving as is for now.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -552,75 +554,88 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions options, strin
...
+                    if (functionResult.Terminate)
+                    {
+                        shouldTerminate = true;
+                        break;
+                    }
+                }
             }
 
-            return (shouldTerminate, consecutiveErrorCount, added);
+            IList<ChatMessage> addedMessages = CreateResponseMessages(results.ToArray());
```
stephentoub, line 623: We don't need to block this PR on it. For the future, this ToArray could be replaced with CollectionsMarshal.AsSpan(results) under `#if NET`

stephentoub (approved):
Thanks!

PR #6328, opened 2025-04-24
[WIP] Update prompts for quality evaluators

shyamnamboodiripad:
Also includes changes to carve out some new public helpers to see if we can remove the base evaluator classes in Quality.dll. The refactoring is work in progress - but creating a draft PR for offline preview and discussion.

Currently, `RetrievalEvaluator` (newly added in this PR) uses the new helpers while all other evaluators still use the base types.


src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Utilities/ParsingUtilities.cs
```diff
@@ -76,4 +78,76 @@ await chatConfig.ChatClient.GetResponseAsync(
 
         return response.Text.Trim();
     }
+
+    internal static bool TryParseNumericScore(
```
peterwald, line 82: nit: It might be better to name this something like `TryParseTaggedResponseText` or something similar, since there are other forms of numeric output that we parse as well.

src/Libraries/Microsoft.Extensions.AI.Evaluation/ChatMessageExtensions.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Linq;
+using System.Text;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Evaluation;
+
+/// <summary>
+/// Extension methods for <see cref="ChatMessage"/>.
+/// </summary>
+public static class ChatMessageExtensions
```
peterwald, line 15: Do you think these helpers will be stable enough to be made public when we go to GA? Should we mark them as experimental for the first release? or can we make them internal?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/Utilities/ParsingUtilities.cs
shyamnamboodiripad: Sure - this one is internal, but happy to rename it. Some of the new evaluators from the python SDK (which we haven't added yet) also use the same format for Boolean metrics - so perhaps we will need to say something like `TryParseTaggedResponseTextWithNumericScore` :)

src/Libraries/Microsoft.Extensions.AI.Evaluation/ChatMessageExtensions.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// Renders the supplied <paramref name="message"/> to a <see langword="string"/>.
+    /// </summary>
+    /// <remarks>
+    /// This function only considers the <see cref="ChatMessage.Text"/> and ignores any <see cref="AIContent"/>s
+    /// present within the <see cref="ChatMessage.Contents"/> of the <paramref name="message"/> that are not
+    /// <see cref="TextContent"/>s.
+    /// </remarks>
+    /// <param name="message">The <see cref="ChatMessage"/> that is to be rendered.</param>
+    /// <returns>A <see langword="string"/> containing the rendered <paramref name="message"/>.</returns>
+    public static string Render(this ChatMessage message)
```
peterwald, line 101: Maybe we should call this RenderText? In the future we may want to add something like RenderImage or RenderAudio

shyamnamboodiripad: The purpose of this helper is currently to return one string for the supplied `ChatMessage`. We then embed the returned string within the evaluation prompt (as part of embedding the evaluated conversation history).

If we end up needing to include non-text `AIContent`s as part of evaluation prompts in the future, the evaluation prompt construction flow would change. At that point, we can no longer return a single string per message. Instead, I suspect we would construct the evaluation prompt by embedding each `AIContent` (present inside the `ChatMessage`s for the conversation that is being evaluated) directly into the `ChatMessage` that we construct for the evaluation prompt.

We would probably still want to add a leading `TextContent` at the start of each embedded message to identify the role for that message. So, the equivalent helper (for rendering a single `ChatMessage` that is part of the conversation history being embedded in an evaluation prompt) might look something like below -

```
private static IEnumerable<AIContent> Render(ChatMessage message)
{
    string? author = message.AuthorName;
    string role = message.Role.Value;
    yield return string.IsNullOrWhiteSpace(author)
        ? new TextContent($"[{role}] {content}\n")
        : new TextContent($"[{author} ({role})] {content}\n");

    foreach (AIContent content in message.Contents)
    {
        yield return content;
    }
}
```

As you propose, I'll rename the above text-only helper to `RenderText`. We could reserve the name `Render` for the future modality-agnostic rendering scenario. I think the `RenderText` helper would still remain relevant for authoring simple text-only evaluators even if we end up switching all the evaluators that ship as part of the library over to the new helper to support other modalities...

peterwald (approved):

PR #6334, opened 2025-04-26
Allow outer IChatClient middleware like OTel to expose features to inner middleware

halter73:
I've already talked a bit to @BrennanConroy, @MackinnonBuck, @stephentoub and @SteveSandersonMS about how the current IChatClient middleware is a bit awkward because there's no good way for outer middleware like UseOpenTelemetry to expose features like ActivitySource to inner middleware like UseFunctionInvocation. Today, it uses IChatClient.GetService to expose the ActivitySource, but that only works for middleware wrapping the OTel middleware, not the inner middleware the OTel middleware is tracking. And the natural place to add the OTel middleware is at the very beginning of the middleware pipeline so it can time and track the complete progress of overall middleware pipeline.

Developers might also want to add the OTel middleware underneath other middleware to observe things like function-invocation that might otherwise be hidden to the outer-layers of the middleware pipeline, but that's something that would usually be done in addition to rather than instead of the overall tracking of the middleware pipeline. 

And now that the OTel middleware exposes the ActivitySource to the inner pipeline, it could be used to track internal activities without necessarily be forced to reapply the OTel middleware between each layer. The function-invoking middleware was already doing effectively this but for the opposite reason. It was trying to work in scenarios where the function-invoking middleware wrapped the OTel middleware which was awkward. But now, that very same logic can be used to track internal calls.

And because of the two-phase approach of first building the client factory instead of directly creating the client with the middleware pipeline, we can make the ActivitySource feature available to both inner and outer middleware if we choose to. This is how we can now get rid of the GetService-wrapping to expose the ActivitySource, and still make it available to any outer function-invoking middleware while it's being constructed.

This does add a Microsoft.Extensions.Features dependency to Microsoft.Extensions.AI, but makes no changes to the core IChatClient interface or any other type in Microsoft.Extensions.AI.Abstractions. I think it's minimally invasive to existing code while giving us the functionality we desire. We also don't have to use IFeatureCollection to go with this approach. We could also use a plain `Dictionary<Type, object>` if we desire.

You'll notice that in the FunctionInvocationTrackedWithActivity we now captures an extra event because we can more conveniently wrap the OTel middleware both underneath and around the function-invoking middleware. But if we stuck to only putting the OTel middleware after the function-invoking middleware, we wouldn't have had to change the test.

```csharp
Func<ChatClientBuilder, ChatClientBuilder> configure = b =>
    b.UseOpenTelemetry(sourceName: sourceName).UseFunctionInvocation().UseOpenTelemetry(sourceName: sourceName);
```

Thanks @BrennanConroy for pointing me to how unintuitive it was that [adding UseOpenTelemetry() to the beginning of the pipeline didn't fully work here](https://github.com/ReubenBond/AccedeConcierge/blob/cafd824db85ddf74ec39b6fb6429d66a2d6497a6/src/System.Distributed.AI.Agents/Utilities/ChatClientExtensions.cs#L102-L107), and instead it had to be added later in the pipeline.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6334)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
@@ -12,10 +13,12 @@ namespace Microsoft.Extensions.AI;
 /// <summary>A builder for creating pipelines of <see cref="IChatClient"/>.</summary>
 public sealed class ChatClientBuilder
 {
+    private delegate IChatClient ClientFactory(IFeatureCollection features, IServiceProvider services);
+
     private readonly Func<IServiceProvider, IChatClient> _innerClientFactory;
 
     /// <summary>The registered client factory instances.</summary>
-    private List<Func<IChatClient, IServiceProvider, IChatClient>>? _clientFactories;
+    private List<Func<ClientFactory, ClientFactory>>? _clientMiddleware;
 
     /// <summary>Initializes a new instance of the <see cref="ChatClientBuilder"/> class.</summary>
     /// <param name="innerClient">The inner <see cref="IChatClient"/> that represents the underlying backend.</param>
@@ -30,6 +33,8 @@ public sealed class ChatClientBuilder
     /// <param name="innerClientFactory">A callback that produces the inner <see cref="IChatClient"/> that represents the underlying backend.</param>
     public ChatClientBuilder(Func<IServiceProvider, IChatClient> innerClientFactory)
     {
+        // NOTE: We would probably also add ChatClientBuilder overload that takes a ClientFactory delegate with an IFeatureCollection
+        // instead of only making it accessible through middleware
         _innerClientFactory = Throw.IfNull(innerClientFactory);
     }
 
@@ -41,25 +46,59 @@ public sealed class ChatClientBuilder
     /// <returns>An instance of <see cref="IChatClient"/> that represents the entire pipeline.</returns>
     public IChatClient Build(IServiceProvider? services = null)
     {
-        services ??= EmptyServiceProvider.Instance;
-        var chatClient = _innerClientFactory(services);
+        ClientFactory chatClientFactory = (f, s) => _innerClientFactory(s);
 
         // To match intuitive expectations, apply the factories in reverse order, so that the first factory added is the outermost.
-        if (_clientFactories is not null)
+        if (_clientMiddleware is not null)
         {
-            for (var i = _clientFactories.Count - 1; i >= 0; i--)
+            for (var i = _clientMiddleware.Count - 1; i >= 0; i--)
             {
-                chatClient = _clientFactories[i](chatClient, services);
-                if (chatClient is null)
+                chatClientFactory = _clientMiddleware[i](chatClientFactory);
+
+                if (chatClientFactory is null)
                 {
                     Throw.InvalidOperationException(
-                        $"The {nameof(ChatClientBuilder)} entry at index {i} returned null. " +
-                        $"Ensure that the callbacks passed to {nameof(Use)} return non-null {nameof(IChatClient)} instances.");
+                        $"The {nameof(ChatClientBuilder)} middleware entry at index {i} returned null. " +
+                        $"Ensure that the callbacks passed to {nameof(Use)} return non-null {nameof(Func<IFeatureCollection, IChatClient>)} instances.");
                 }
             }
         }
 
+        var features = new FeatureCollection();
+        services ??= EmptyServiceProvider.Instance;
+        var chatClient = chatClientFactory(features, services);
+
+        if (chatClient is null)
+        {
+            Throw.InvalidOperationException(
+                $"The {nameof(ChatClientBuilder)} middleware pipeline returned null." +
+                $"Ensure that the callbacks passed to {nameof(Use)} return non-null {nameof(IChatClient)} instances.");
+        }
+
         return chatClient;
+    }
+
+    /// <summary>Adds a factory for an intermediate chat client to the chat client pipeline.</summary>
+    /// <param name="clientMiddleware">The client middleware function.</param>
+    /// <returns>The updated <see cref="ChatClientBuilder"/> instance.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="clientMiddleware"/> is <see langword="null"/>.</exception>
+    public ChatClientBuilder Use(Func<Func<IFeatureCollection, IChatClient>, IServiceProvider, Func<IFeatureCollection, IChatClient>> clientMiddleware)
+    {
+        _ = Throw.IfNull(clientMiddleware);
+
+        (_clientMiddleware ??= []).Add(nextFactory =>
+        {
+            // Pass the services around the user-defined middleware since we don't allow them to pass the IServiceProvider to the next() middleware.
+            // It'd be simpler to make passing the IServiceProvider to the next middleware explicit, but as before, it's probably best to
+            // not allow IServiceProvider to be changed from one middleware to the next.
+            return (features, services) =>
+            {
+                var appliedMiddleware = clientMiddleware(fc => nextFactory(fc, services), services);
+                return appliedMiddleware(features);
+            };
+        });
+
+        return this;
     }
 
     /// <summary>Adds a factory for an intermediate chat client to the chat client pipeline.</summary>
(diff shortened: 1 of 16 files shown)


halter73:
> With this change, do you think it leads to any drawbacks for people who really only way to have OpenTelemetryChatClient at the innermost position (directly wrapping the underlying client), and would it force us to change the Aspire integration somehow?

Given that this makes the ActivitySource to both inner and outer middleware using the same mechanism, I don't think there are any real drawbacks other than maybe the small additional complexity of adding another Use overload. I don't even think it's that much more complicated given it makes the middleware pipeline a more conventional `List<Func<T..., TResult>, Func<T..., TResult>>` and gets rid of the need to expose the ActivitySource via GetService.

That's why I opened the PR instead of just filing the issue. I was planning on filing the issue first, but I wanted to show at least a little bit of sample code to demonstrate what I was talking about, and then I saw how well it worked without any real drawbacks and just finished the change.

halter73:
> It might be advantageous to do so and call the value pipelineProperties or something. Just thinking we're on the border of it getting pretty confusing (having 3 bags-o-stuff: IServiceCollection, GetService<T>, IFeatureCollection, without obvious differences in role from the naming alone).

There might be some value in naming the IFeatureCollection, but it's only visible at all via the new `Use(Func<Func<IFeatureCollection, IChatClient>, IServiceProvider, Func<IFeatureCollection, IChatClient>> clientMiddleware)` overload, so I doubt hardly anyone will see it unless they have an advanced scenario like picking up the ActivitySource from the OTel middleware.

However, this does remind me to one downside of this change as-is which I should have brought up in the PR description. You'll notice that I had to change the UsesEmptyServiceProviderWhenNoServicesProvided test to specify parameters. I don't think we should have two overloads of Use take a Func of two parameters. It's too hard for the compiler to reason about. Even when it otherwise could infer the right overload, one typo in the callback leads to terrible error messages now that the compiler doesn't know which overload your calling and therefore the types of your parameters.

I've also considered renaming the new `Use` to something different like `UseMiddleware` even though that's semantically redundant just to help the compiler. I'm open to suggestions on that. Maybe `UseWithPipelineProperties()`? 😆

RussKie (approved):
Approving the infra side

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
```diff
@@ -41,27 +46,61 @@ public ChatClientBuilder(Func<IServiceProvider, IChatClient> innerClientFactory)
...
-                chatClient = _clientFactories[i](chatClient, services);
-                if (chatClient is null)
+                chatClientFactory = _clientMiddleware[i](chatClientFactory);
+
+                if (chatClientFactory is null)
                 {
                     Throw.InvalidOperationException(
-                        $"The {nameof(ChatClientBuilder)} entry at index {i} returned null. " +
-                        $"Ensure that the callbacks passed to {nameof(Use)} return non-null {nameof(IChatClient)} instances.");
+                        $"The {nameof(ChatClientBuilder)} middleware entry at index {i} returned null. " +
```
stephentoub, line 61: I'd like to avoid "middleware" showing up publicly.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
```diff
@@ -41,27 +46,61 @@ public ChatClientBuilder(Func<IServiceProvider, IChatClient> innerClientFactory)
...
+                        $"Ensure that the callbacks passed to {nameof(Use)} return non-null {nameof(Func<IFeatureCollection, IChatClient>)} instances.");
                 }
             }
         }
 
+        var features = new FeatureCollection();
```
stephentoub, line 67: What would this be without M.E.Features... `Dictionary<Type, object?>`?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
```diff
@@ -41,27 +46,61 @@ public ChatClientBuilder(Func<IServiceProvider, IChatClient> innerClientFactory)
...
+
+        (_clientMiddleware ??= []).Add(nextFactory =>
+        {
+            // Pass the services around the user-defined middleware since we don't allow them to pass the IServiceProvider to the next() middleware.
+            // It'd be simpler to make passing the IServiceProvider to the next middleware explicit, but as before, it's probably best to
+            // not allow IServiceProvider to be changed from one middleware to the next.
+            return (features, services) =>
+            {
+                var appliedMiddleware = clientMiddleware(fc => nextFactory(fc, services), services);
+                return appliedMiddleware(features);
```
stephentoub, line 97: What guarantees appliedMiddleware is not null?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -71,11 +72,12 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
     /// <param name="loggerFactory">An <see cref="ILoggerFactory"/> to use for logging information about function invocation.</param>
     /// <param name="functionInvocationServices">An optional <see cref="IServiceProvider"/> to use for resolving services required by the <see cref="AIFunction"/> instances being invoked.</param>
-    public FunctionInvokingChatClient(IChatClient innerClient, ILoggerFactory? loggerFactory = null, IServiceProvider? functionInvocationServices = null)
+    /// <param name="features">An <see cref="IFeatureCollection"/> for acquiring the <see cref="ActivitySource"/> possibly set by earlier <see cref="OpenTelemetryChatClient"/> middleware.</param>
```
stephentoub, line 75: I'll repeat my feedback previously: this feels like a lot of effort and complication just for the ActivitySource published by OpenTelemetryChatClient. We only did it this way because it happened to work for the needed cases. I'm not seeing the value in generalizing this capability. Where else is it going to be used?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
halter73: I think so. I might make the object non-nullable, but you're always going to have to cast anyway.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
halter73: I think the check on line 58 still guards against the "appliedMiddleware" ClientFactory being null. What we currently don't check until the end of the pipeline is whether the returned IChatClient itself is null. I can go back to checking this at every stage.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
```diff
@@ -5,17 +5,20 @@
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
```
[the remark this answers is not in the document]
halter73: You might already know this, but IFeatureCollection, while being in the Microsoft.AspNetCore namespace, is in the [Microsoft.Extensions.Features nuget package](https://www.nuget.org/packages/Microsoft.Extensions.Features) which only defines 5 types and itself has no dependencies.

https://github.com/dotnet/aspnetcore/tree/main/src/Extensions/Features/src

I do understand not wanting to rely on anything in the Microsoft.AspNetCore namespace though, so I'll change it to a `IDictionary<string, object>`.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
halter73: I suspect that there will be more cases where earlier "client factories" or "middleware" want to expose state to factories added later in the pipeline. I think it's really odd to make any kind of wrapping pipeline without the ability to pass context from the outer wrappers to the inner wrappers during pipeline construction whether or not you call it middleware.

I think it'd be worth it if all this ever did was make it so you could all UseOpenTelemetry at the start of the ChatClientBuilder pipeline and expose its ActivitySource to all FunctionInvokingChatClient added later in the same pipeline. I also think this simplifies the pipeline by following the typical `List<Func<T, T>> where T : Delegate` pattern.

But if you don't think it's worth it, I'll close the PR.

stephentoub: @SteveSandersonMS ?

SteveSandersonMS: Looking at the implementation, it seems this is something we could add non-breakingly in the future if we get further use cases.

If:

  * it would only be used for otel
  * and there's some alternative in that case (e.g., people can put `UseOpenTelemetry` manually as the outer middleware if they want to measure spans both inside and outside function calling)
  * and unless we also figure out how this would work with Aspire integration

...I'm fine with leaving things as-is.

halter73: I admit I probably had a too server-centric view of things to think that the otel middleware naturally belongs at the start of the pipeline. And as you say, we could add it later.

PR #6181, opened 2025-03-23
Add Windows disk I/O metrics to ResourceMonitoring

makazeu:
Closes #5842

Add the below new Windows metrics to the ResourceMonitoring library:
- system.disk.io
- system.disk.operations

Will try to support below metrics in the future:
- system.disk.io_time
- system.disk.operation_time
- system.disk.pending_operations

Also, will support Linux in the future

### Tested in Grafana + Prometheus with [OpenTelemetry.Exporter.Prometheus.HttpListener](https://www.nuget.org/packages/OpenTelemetry.Exporter.Prometheus.HttpListener)
image

### Tested in Geneva with [OpenTelemetry.Exporter.Geneva](https://www.nuget.org/packages/OpenTelemetry.Exporter.Geneva)
Screenshot 2025-04-26 042010

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6181)

Issue #5842, opened 2025-02-04
Add disk IO metrics to Resource Monitoring


evgenyfedorov2:
In addition to already provided network, CPU, and memory utilization metrics, please provide disk IO metrics.

PR #6338, opened 2025-04-28
Add Disk IO time metric for Windows in ResourceMonitoring

makazeu:
Add Disk IO time metric (`system.disk.io_time`) for Windows in ResourceMonitoring.

This is a recommended metric [according to OpenTelemetry specs](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemdiskio_time).
image

The metric value represents the real elapsed time (“wall clock”) used in the I/O path (time from operations running in parallel are not counted). Measured as:

- Linux: Field 13 from [procfs-diskstats](https://www.kernel.org/doc/Documentation/ABI/testing/procfs-diskstats)
- Windows: The complement of [“Disk% Idle Time”](https://learn.microsoft.com/archive/blogs/askcore/windows-performance-monitor-disk-counters-explained#windows-performance-monitor-disk-counters-explained) performance counter: uptime * (100 - "Disk\\% Idle Time") / 100

This PR provides the Windows implementation for the metric.

Tested in Grafana + Prometheus
438507941-f943d201-b218-413f-af41-fbda7cb7d2d9

And tested in Geneva
image

Linked: #6181 & #5842

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6338)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/Network/LinuxNetworkUtilizationParser.cs
+++ b/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/Network/LinuxNetworkUtilizationParser.cs
@@ -18,12 +18,12 @@ internal sealed class LinuxNetworkUtilizationParser
     private static readonly ObjectPool<BufferWriter<char>> _sharedBufferWriterPool = BufferWriterPool.CreateBufferWriterPool<char>();
 
     /// <remarks>
-    /// File that provide information about currently active TCP_IPv4 connections.
+    /// File that provides information about currently active TCP_IPv4 connections.
     /// </remarks>
     private static readonly FileInfo _tcp = new("/proc/net/tcp");
 
     /// <remarks>
-    /// File that provide information about currently active TCP_IPv6 connections.
+    /// File that provides information about currently active TCP_IPv6 connections.
     /// </remarks>
     private static readonly FileInfo _tcp6 = new("/proc/net/tcp6");
 


src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Disk/WindowsDiskIoRatePerfCounter.cs
```diff
@@ -43,15 +43,9 @@ internal void InitializeDiskCounters()
     {
         foreach (string instanceName in _instanceNames)
         {
-            // Skip the total instance
```
makazeu, line 46: In `InitializeDiskCounters` method of `WindowsDiskMetrics.cs`, I filter out the `_Total` from `instanceNames`, so here it's unnecessary to filter out `_Total` again.

makazeu:
cc @evgenyfedorov2 thanks

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Disk/WindowsDiskIoTimePerfCounter.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            _ = counter.NextValue();
+        }
+
+        _lastTimestamp = _timeProvider.GetUtcNow().ToUnixTimeMilliseconds();
+    }
+
+    internal void UpdateDiskCounters()
+    {
+        long currentTimestamp = _timeProvider.GetUtcNow().ToUnixTimeMilliseconds();
```
evgenyfedorov2, line 60:
```suggestion
        long currentTimestamp = _timeProvider.GetUtcNow().ToUnixTimeSeconds();
```

and avoid converting to seconds later because it is already seconds?

makazeu: I still prefer using `ToUnixTimeMilliseconds`, because if we use `ToUnixTimeSeconds`, the calculated time interval can only be accurate to the second. However, we cannot guarantee that the time difference between two samples is exactly an integer number of seconds, which may lead to precision issues.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Disk/WindowsDiskMetrics.cs
```diff
@@ -42,39 +44,69 @@ public WindowsDiskMetrics(
...
+                WindowsDiskPerfCounterNames.DiskIdleTimeCounter,
+                instanceNames);
+            ioTimePerfCounter.InitializeDiskCounters();
+            _diskIoTimePerfCounter = ioTimePerfCounter;
+        }
+#pragma warning disable CA1031
+        catch (Exception ex)
+#pragma warning restore CA1031
+        {
+            Debug.WriteLine("Error initializing disk io time perf counter: " + ex.Message);
```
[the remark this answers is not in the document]
makazeu: done, thanks

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/Disk/WindowsDiskIoTimePerfCounter.cs
evgenyfedorov2: Ok, I see your point, thank you!

However, it has just occurred to me, why don't we use ticks? For instance, we do use ticks here https://github.com/dotnet/extensions/blob/9652f798f2de4e40c088075d00d185628c8e3dbb/src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs#L235. We can convert to seconds only at the very last moment before saving it to the `TotalSeconds` dictionary

evgenyfedorov2: .

evgenyfedorov2 (approved):

PR #6344, opened 2025-04-29
Add PDF support to OpenAI AsIChatClient

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6344)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -209,6 +209,10 @@ internal sealed partial class OpenAIChatClient : IChatClient
                     }
 
                     break;
+
+                case DataContent dataContent when dataContent.MediaType.StartsWith("application/pdf", StringComparison.OrdinalIgnoreCase):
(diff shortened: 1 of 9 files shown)


RussKie (approved):
Approving the infra side

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -209,6 +209,10 @@ private static List<ChatMessageContentPart> ToOpenAIChatContent(IList<AIContent>
                     }
 
                     break;
+
+                case DataContent dataContent when dataContent.MediaType.StartsWith("application/pdf", StringComparison.OrdinalIgnoreCase):
+                    parts.Add(ChatMessageContentPart.CreateFilePart(BinaryData.FromBytes(dataContent.Data), dataContent.MediaType, $"{Guid.NewGuid():N}.pdf"));
+                    break;
```
SteveSandersonMS, line 215: Would there be a drawback to just mapping all other `DataContent` objects, not limiting it to PDF? Are there cases where someone would add a `DataContent` but not want it to be sent in the request?

stephentoub: Just that, at least today, anything else will result in the request failing.

PR #6355, opened 2025-04-30
Add project template build tests + CG reporting

### LINKED ISSUE #6322: Ensure templates are built via a pipeline to support CG
The below templates need to be built using a pipeline so that CG is run on them

AI Chat Web Project

MackinnonBuck:
Adds project template create/restore/build tests and CG reporting.

Summary of changes:
* Added infrastructure for invoking .NET commands from test code
* Added a `TemplateSandbox` folder that gets used to build (and in the future, run) tests in an environment isolated from the rest of the repo
* Added a script that configures the local PowerShell environment to match that of template execution tests
  * This makes debugging generated tests locally a lot easier than it would otherwise have been
* Added tests that create, restore, and build variations of the AI Chat Web project template
  * **Note:** Test output does not get cleared until the next test run. This enables easy debugging of generated templates and allows template dependencies to get discovered for CG.
* Added a README explaining how to debug template execution tests
* Made the `Microsoft.Extensions.AI.Templates.Test` project an "integration test" so it runs separately from existing unit tests
  * This is required because template execution tests can only be run after the repo gets packed
* Updated pipelines to include a "Run integration tests" step that runs after packing the repo

Fixes #6322

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6355)


test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
```diff
@@ -0,0 +1,29 @@
+# Template test sandbox
+
+This folder exists to serve as an isolated environment for template execution tests.
+
+## Debugging template execution tests
+
+Before running template execution tests, make sure that packages defined in this solution have been packed by running the following commands from the repo root:
+```sh
+./build.cmd -build
+./build.cmd -pack
```
RussKie, line 10: Just to confirm these could be executed together - i.e., `./build.cmd -build -pack`, right?

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
```diff
@@ -0,0 +1,29 @@
+# Template test sandbox
```
RussKie, line 1: The readme isn't very discoverable, unfortunately. I would suggest creating a new doc in https://github.com/dotnet/extensions/tree/main/docs (say, call it "testing.md"), adding a paragraph on how to test the templates and then link to this doc.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Infrastructure/TemplateExecutionTestBase.cs
```diff
@@ -0,0 +1,67 @@
...
+        if (disposing)
+        {
+            Fixture.SetCurrentTestOutputHelper(null);
+        }
+    }
+
+    public void Dispose()
+    {
+        Dispose(true);
+        GC.SuppressFinalize(this);
```
RussKie, line 50: Since there's no finaliser, is this really needed? Just curious.

RussKie (approved):
🚀

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
MackinnonBuck: The one time I tried that, the dynamically generated template content (like the `.csproj` file) didn't get included in the NuGet package. So the build currently has to be run before the pack. It does seem broken, but I think it's been this way since the templates were introduced. Would be nice to fix soon 🙂

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
MackinnonBuck: Good idea!

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Infrastructure/TemplateExecutionTestBase.cs
MackinnonBuck: Theoretically a derived class could have a finalizer? But in reality that's probably never ever going to happen. So I agree this can probably be removed.

MackinnonBuck: Ah, the compiler adds a warning if I remove it :/

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
MackinnonBuck: Hm, `building.md` already has general testing instructions, and having a `testing.md` might make those original instructions less discoverable. I could move all testing-related stuff to `testing.md`, but then the testing info for this project could get buried in other, unrelated details.

I've instead added a new README at the root of the project folder. I think that should be reasonably discoverable. Does that sound fine?

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
RussKie: May be worth adding a comment to mention that the commands should be run separately (until this is fixed). Because I'd be tempted to run those as one.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/README.md
RussKie: I'd add the link to the new docs/testing.md to the main readme, so that contributors have a clear(er) direction. 

But I leave it with you to decide. :)

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/Directory.Build.props
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <!--
+    This file exists to detach generated projects from other Director.Build.props
```
SteveSandersonMS, line 3:
```suggestion
    This file exists to detach generated projects from other Directory.Build.props
```

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/Directory.Build.targets
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <!--
+    This file exists to detach generated projects from other Director.Build.targets
```
SteveSandersonMS, line 3:
```suggestion
    This file exists to detach generated projects from other Directory.Build.targets
```

SteveSandersonMS (approved):
Cool!

This is all pretty sophisticated. Did you take much of the infrastructure code from our other repos or is this all custom for `dotnet/extensions`? If it's custom, do you think we should look into ways to share it with `dotnet/aspnetcore` etc or is this unique to the scenarios in this repo?

Presumably this could be used as the basis for Playwright tests in the future, though that would also force us to provide working AI service endpoints in CI which is a whole other challenge. Right?

MackinnonBuck:
> Did you take much of the infrastructure code from our other repos or is this all custom for `dotnet/extensions`?

It's mostly custom, but I did look into the approaches taken in the `aspnetcore`, `sdk`, and `aspire` repos to see if there was anything that could be borrowed. Thankfully, it's quite a lot simpler to get these kinds of tests working in the `extensions` repo because we don't need to mutate the local .NET SDK installation to execute template projects (unlike e.g., `aspnetcore`, which needs to make a copy of the SDK with the just-built shared framework swapped in, or `aspire`, which has to install a workload). In the `extensions` case, it's enough to specify an alternate NuGet packages location and a custom template install directory. Therefore, it didn't seem worth it to do a direct copy/paste when the complexity isn't warranted.

The main candidate for shared logic seems to be the ability to reliably run the .NET CLI from tests. The `sdk` repo has the most sophisticated solution for this out of the repos I looked at ([Microsoft.DotNet.Cli.Utils](https://github.com/dotnet/sdk/tree/2fccb09820906c70d141e8df3221760769055ef7/src/Cli/Microsoft.DotNet.Cli.Utils) + [Microsoft.NET.TestFramework](https://github.com/dotnet/sdk/tree/2fccb09820906c70d141e8df3221760769055ef7/test/Microsoft.NET.TestFramework)), and I effectively used a very simplified version of that (similar to `aspire`, but simpler still). And `aspnetcore` sort of does its own separate thing.

That said, it's likely possible to create a shared utility that provides the most common functionality required by each repo (maybe something like what the SDK already has). I'd be curious to see what others think about this.

radical:
This is what we have - https://github.com/dotnet/aspire/blob/main/tests/Aspire.Templates.Tests/NewUpAndBuildStandaloneTemplateTests.cs

It doesn't depend on a workload now, so doesn't need to modify local dotnet installs. It supports testing against multiple sdks.

radical:
Playwright tests - https://github.com/dotnet/aspire/blob/main/tests/Aspire.Templates.Tests/TemplateTestsBase.cs#L173

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/nuget.template_test.config
```diff
@@ -0,0 +1,26 @@
...
+
+  <fallbackPackageFolders>
+    <!-- Clear to avoid inheriting from the repo root NuGet.config -->
+    <clear />
+  </fallbackPackageFolders>
+  <packageSources>
+    <clear />
+    <!-- 'LOCAL_SHIPPING_PATH' is an environment variable that gets set by the template tests. -->
+    <add key="local-shipping" value="%LOCAL_SHIPPING_PATH%" />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
```
jeffhandley, line 16: For scenarios where AI packages and/or non-AI packages from this repo are set to use pinned version from nuget.org, will this setup ensure the pinned versions are resolved instead of the local-shipping versions?

For my own local testing, I've been deleting the locally built packages from the folder for anything that I want to ensure gets resolved from a pinned version to be certain the pinned version is resolved and used.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        "--vector-store", "azureaisearch")]
+    [InlineData("GitHubModelsWithQdrant",
+        "--provider", "githubmodels",
+        "--vector-store", "qdrant")]
+    [InlineData("OllamaWithLocalStore",
+        "--provider", "ollama",
+        "--vector-store", "local")]
+    [InlineData("OpenAIWithLocalStore",
+        "--provider", "ollama",
+        "--vector-store", "local")]
```
jeffhandley, line 63: I think there are other configuration combinations possible/supported that aren't covered; is that right? Is this set of scenarios giving us the distinct set of possible dependencies?

I'd be inclined to generate every possible/supported configuration and doing so here in a way where when new combinations are added, it's less likely we'll forget to add the relevant coverage. That could be achieved by using a MemberData approach that generates all possible combinations of providers, vector stores, and the other switches, when the test having logic to identify unsupported combinations and skip the scenario.

jeffhandley (commented):
With this now in place, what are your thoughts on the existing snapshot tests? Changes to the snapshot tests should be a separate PR if you suggest any though.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/activate.ps1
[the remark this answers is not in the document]
MackinnonBuck: Yes, definitely would be nice for those who aren't debugging on a Windows machine (or don't have PowerShell installed on their Mac/Linux machine).

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/nuget.template_test.config
MackinnonBuck: From my testing, if I switch to using the pinned versions of the template, then the pinned versions get resolved over the local ones (even if the local version is "higher" than the pinned one).

However, I just tried testing what would happen if the specified pinned version could not be found (causing a restore warning), and while the warning gets logged to the output, it doesn't get logged to standard error and the test still passes. Going to push a fix for that since restore warnings should be considered failures.

MackinnonBuck: Fixed now!

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
MackinnonBuck: Generating all valid combinations sounds good 🙂 I initially avoided that to prevent redundant checks from taking up too much time on CI. But since each check is only ~10 seconds, I think this should be fine. I'll go ahead and implement that approach.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/activate.ps1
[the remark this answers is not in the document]
jeffhandley: I'm glad you realized that I totally missed the operative word in my comment: "We _don't_ need it now..." 😆

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -0,0 +1,148 @@
...
+        await Fixture.RestoreProjectAsync(project);
+        await Fixture.BuildProjectAsync(project);
+    }
+
+    private static readonly (string name, string[] values)[] _templateOptions = [
+        ("--provider",          ["azureopenai", "githubmodels", "ollama", "openai"]),
+        ("--vector-store",      ["azureaisearch", "local", "qdrant"]),
+        ("--managed-identity",  ["true", "false"]),
+        ("--aspire",            ["true", "false"]),
+    ];
```
jeffhandley, line 66: Nice! I like the way you put this together.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<Project>
+  <!--
+    This file exists to detach generated projects from other Directory.Build.props
+    files in this repo.
+  -->
+
+  <PropertyGroup>
+    <!-- Restore/build warnings are treated as errors to be easily detectable by test code. -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
jeffhandley, line 9: 👍

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/TemplateSandbox/.editorconfig
```diff
@@ -0,0 +1,2 @@
+# Don't apply the repo's editorconfig settings to generated templates.
```
jeffhandley, line 1: Thanks for the comment on the editorconfig

jeffhandley (approved):
This is great, @MackinnonBuck. The only comment I have left that I'd like you to address is that I don't know what specific side-effect of generating and building the template combinations gives component governance visibility into the dependencies.

Can you add a comment in the place or places where there are any subtle details that make that wire up correctly? Is it _how_ the projects are built, the fact that they simply perform a restore, the output location, and/or other aspects that result in CG visibility?

jeffhandley (approved):
Excellent comments and clarity; thanks!

PR #6377, opened 2025-05-05
Add some additional documentation around usage of cache, and CSP properties on report

peterwald:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6377)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/IResponseCacheProvider.cs
```diff
@@ -26,6 +26,12 @@ public interface IResponseCacheProvider
     /// Returns an <see cref="IDistributedCache"/> that caches the AI responses associated with a particular
     /// <see cref="ScenarioRun"/>.
     /// </summary>
+    /// <remarks>
+    /// The <see cref="IDistributedCache"/> returned from this method is specifically designed to support
+    /// the caching of responses for evaluation purposes. It is not intended for use as a general-purpose
+    /// distributed cache. For general purposes, consider using a different caching provider such as an
```
shyamnamboodiripad, line 32: This guidance is a bit confusing / vague imho. `IDistributedCache` is a simple interface with Get / Add / Remove / Update functions. An `IDistributedCache` implementer should not need to consider how the cache will end up being used, and it should be completely valid for an `IResponseCacheProvider` to return any `IDistributedCache` here (so long as the implementation is able to Get / Add / Removed / Update bytes for any key that is supplied to it).

The `IResponseCacheProvider` itself is the only part that is eval-specific (since it considers things like the `scenarioName` and `iterationName`). And I think this aspect should already be clear based on usage / based on the existing docs. So, I am not sure whether we should include these remarks here...

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/html-report/index.html
```diff
@@ -8,8 +8,7 @@
 
 <head>
   <meta charset="UTF-8" />
-  <link rel="stylesheet"
-    href="https://res-1.cdn.office.net/files/fabric-cdn-prod_20230815.002/office-ui-fabric-core/11.1.0/css/fabric.min.css" />
```
shyamnamboodiripad, line 12: Just curious - what was this style sheet exactly doing and how different was it from what we are doing now with the new meta tag below?

shyamnamboodiripad, line 12: Should we additionally HTML encode the report JSON before embedding it into the HTML (for an additional layer of protection)?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/IResponseCacheProvider.cs
```diff
@@ -26,6 +26,12 @@ public interface IResponseCacheProvider
     /// Returns an <see cref="IDistributedCache"/> that caches the AI responses associated with a particular
     /// <see cref="ScenarioRun"/>.
     /// </summary>
+    /// <remarks>
```
shyamnamboodiripad, line 29: Thinking aloud mostly - but looks like naming wise we have `Evaluation` in the name of most of our abstractions (`IEvaluator`, `EvaluationResult`, `EvaluationMetric`, `EvaluationDiagnostic`, `EvaluationContext`, `IEvaluationReportWriter` etc.). We also have some counter examples - `ScenarioRun`, `ReportingConfiguration` - but I think those names make sense as is.

I am wondering whether we should make the following name changes to clarify that the below reporting abstractions are not for general-purpose response caches / result stores - but rather specifically for evaluation purposes instead?

`IResponseCacheProvider` -> `IEvaluationResponseCacheProvider`
`IResultStore` -> `IEvaluationResultStore`

The types are already in namespaces that contain the word `Evaluation` - but perhaps it wouldn't hurt to make this explicit in the type names as well?

The implementing types could probably stay unchanged - those names seem fine as is. Or we could change them as follows -

`DiskBasedResponseCache*` -> `DiskBasedEvaluationResponseCache*`
`DiskBasedResultStore*` -> `DiskBasedEvaluationResultStore*`
(and same for `AzureStorage*)

Thoughts?

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/html-report/index.html
peterwald: It was left behind from earlier edits. It is redundant with the packaged CSS and no longer needed.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/IResponseCacheProvider.cs
peterwald: Removed after online discussion.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/IResponseCacheProvider.cs
```diff
@@ -26,6 +26,7 @@ public interface IResponseCacheProvider
...
```
shyamnamboodiripad, line 29: Looks like the start tag survived. This will probably cause a build warning...

```suggestion
```

peterwald: Good catch. Thanks.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/TypeScript/html-report/index.html
peterwald: We added the CSP policy to avoid having to do this. I would suggest not doing it for now, as with CSP there is nothing accessible to the page anyway, and the HTML encoding is likely to be a source of bugs and fragility, not to mention slow performance of loading.

PR #6386, opened 2025-05-07
Add security comments for GetResponseAsync/GetStreamingResponseAsync

### LINKED ISSUE #6245: MEAI docs: GetResponseAsync/GetStreamingResponseAsync
An outcome from the TM review.

The docs for GetResponseAsync/GetStreamingResponseAsync should be clear that:

 * We are just passing through any messages and returning whatever results come back.
 * MEAI doesn't do any defence against prompt injection; it's up to application code to do that
 * MEAI doesn't impose any limits on data sizes, either outbound or returned. Apps that need to limit this must do it themselves.
 * MEAI doesn't impose any limits on the number of calls (per user or in total, over any period of time) since it's stateless. Apps that need to impose limits (e.g., using rate limiting) must do it themselves

jeffhandley:
Fixes #6245

An outcome from the threat model review. The docs for GetResponseAsync/GetStreamingResponseAsync should be clear that:

1. We are just passing through any messages and returning whatever results come back.
2. MEAI doesn't do any defence against prompt injection; it's up to application code to do that
3. MEAI doesn't impose any limits on data sizes, either outbound or returned. Apps that need to limit this must do it themselves.
4. MEAI doesn't impose any limits on the number of calls (per user or in total, over any period of time) since it's stateless. Apps that need to impose limits (e.g., using rate limiting) must do it themselves
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6386)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -33,6 +33,22 @@ public interface IChatClient : IDisposable
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The response messages generated by the client.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="messages"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// Prompt <paramref name="messages"/> are passed through to the underlying provider, and the <see cref="ChatResponse.Messages"/>
+    /// in the <see cref="ChatResponse"/> are returned without modification or filtering.
```
stephentoub, line 38: I understand the intent, but I think this is misleading. One of the strengths of the middleware approach is exactly that IChatClient implementations in the pipeline _can_ manipulate these objects, in both directions. FunctionInvokingChatClient, for example, is augmenting the messages with its own. Someone could write a SummarizeHIstoryChatClient that replaced the whole contents with a summarized version that was then forwarded along. Etc.

jeffhandley: Good point; thanks. We could remove the opening sentence entirely, change "Callers are expected..." to "Applications are expected..." and the declarations to include "by default."

gewarren (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -33,6 +33,20 @@ public interface IChatClient : IDisposable
...
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on data sizes sent to the underlying provider or returned from it
+    /// by default. Applications are expected to apply data size limits as needed.
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on the number of messages sent to the underlying provider or returned from
+    /// it by default. Applications are expected to apply rate limits on a per-user or per-session basis as needed.
+    /// </para>
+    /// </remarks>
```
stephentoub, line 49: These are interfaces. We don't own the implementations (or at least not most of them). I don't know that we can make any claims about what implementations do or don't do by default with regards to limits.

jeffhandley:
@gewarren @stephentoub I used GitHub Copilot in VS to coach me through how to address the problem of needing to set expectations on the interface API, while then providing clear statements on the implementations. I'm happy with the results; what do you think?

Also, @gewarren, is it OK to use `<inheritdoc />` and then add additional `<remarks>`?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceChatClient.cs
```diff
@@ -73,6 +73,20 @@ public AzureAIInferenceChatClient(ChatCompletionsClient chatCompletionsClient, s
...
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on data sizes sent to the underlying provider or returned from it
+    /// by default. Applications are expected to apply data size limits as needed.
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on the number of messages sent to the underlying provider or returned from
+    /// it by default. Applications are expected to apply rate limits on a per-user or per-session basis as needed.
+    /// </para>
+    /// </remarks>
```
stephentoub, line 89: This type is internal. These docs won't show up anywhere.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -68,6 +68,20 @@ public OpenAIChatClient(ChatClient chatClient)
...
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on data sizes sent to the underlying provider or returned from it
+    /// by default. Applications are expected to apply data size limits as needed.
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on the number of messages sent to the underlying provider or returned from
+    /// it by default. Applications are expected to apply rate limits on a per-user or per-session basis as needed.
+    /// </para>
+    /// </remarks>
```
stephentoub, line 84: This type is internal. These docs won't show up anywhere.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs
```diff
@@ -33,6 +33,11 @@ public interface IChatClient : IDisposable
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>The response messages generated by the client.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="messages"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// Applications must consider the risks of prompt injection attacks, data sizes, and the number of messages
+    /// sent to the underlying provider or returned from it. Unless a specific <see cref="IChatClient"/> implementation
+    /// explicitly documents safeguards for these concerns, the application is expected to implement appropriate protections.
+    /// </remarks>
```
stephentoub, line 40: How about adding it as a paragraph to the remarks on the type rather than on each method?

jeffhandley: Great idea. Thank you.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceExtensions.cs
```diff
@@ -9,6 +9,20 @@ namespace Microsoft.Extensions.AI;
...
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on data sizes sent to the underlying provider or returned from it
+    /// by default. Applications are expected to apply data size limits as needed.
+    /// </para>
+    /// <para>
+    /// There are no limits or restrictions imposed on the number of messages sent to the underlying provider or returned from
+    /// it by default. Applications are expected to apply rate limits on a per-user or per-session basis as needed.
+    /// </para>
+    /// </remarks>
```
SteveSandersonMS, line 25: This is fine though I wonder if it's really necessary to have it on the `AsIChatClient` methods. It seems like the remark you've added to `IChatClient` already covers it fully, since that's the critical place where the calls occur.

I also like how you phrased it concisely on `IChatClient` better than the longer explanation on each of the `AsIChatClient` methods.

SteveSandersonMS (approved):

gewarren:
> Also, @gewarren, is it OK to use `<inheritdoc />` and then add additional `<remarks>`?

Yes, that just means that no remarks will be inherited.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceExtensions.cs
jeffhandley: Thanks, folks. You've convinced me that the succinct remarks on `IChatClient` are sufficient. I'll keep those and drop the remarks on the implementations.

PR #6383, opened 2025-05-06
Add JSON schema transformation functionality to `AIJsonUtilities`

eiriktsarpalis:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6383)

PR #6394, opened 2025-05-08
Rename useJsonSchema parameter

stephentoub:
The `GetResponseAsync<T>` methods accept a `bool?` parameter currently called `useJsonSchema`. This is confusing, because the whole point of the method is to create and use a JSON schema from the `T`. The parameter actually controls _how_ that schema is used, whether it's included as part of the messages (false), as part of a ChatResponseFormat in the ChatOptions (true), or up to the system to decide.

I've clarified it by renaming it from `useJsonSchema` to `useJsonSchemaResponseFormat`. It's wordier, but it disambiguates the intent.


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -192,8 +190,8 @@ public static async Task<ChatResponse<T>> GetResponseAsync<T>(
 
         // We default to assuming that models support JSON schema because developers will normally use
         // GetResponseAsync<T> only with models that do. If the model doesn't support JSON schema, it may
-        // throw or it may ignore the schema. In these cases developers should pass useJsonSchema: false.
-        if (useJsonSchema.GetValueOrDefault(true))
+        // throw or it may ignore the schema. In these cases developers should pass useJsonSchemaResponseFormat: false.
+        if (useJsonSchemaResponseFormat.GetValueOrDefault(true))
```
eiriktsarpalis, line 194:
```suggestion
        if (useJsonSchemaResponseFormat ?? true)
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
```diff
@@ -149,7 +147,7 @@ public static async Task<ChatResponse<T>> GetResponseAsync<T>(
         IEnumerable<ChatMessage> messages,
         JsonSerializerOptions serializerOptions,
         ChatOptions? options = null,
-        bool? useJsonSchema = null,
+        bool? useJsonSchemaResponseFormat = null,
```
eiriktsarpalis, line 150: Couldn't we just make this a regular boolean in all methods?
```suggestion
        bool useJsonSchemaResponseFormat = true,
```

eiriktsarpalis (approved):
I had a conversation with @SteveSandersonMS about potentially replacing `useJsonSchema` with

stephentoub:
> I had a conversation with @SteveSandersonMS about potentially replacing `useJsonSchema` with

With...?

eiriktsarpalis:
> With...?

Please disregard. I was going to [quote this conversation](https://github.com/dotnet/extensions/pull/6285#discussion_r2058659587) but this is now being superseded by https://github.com/dotnet/extensions/pull/6383. Somehow my partially completed sentence still ended up in the review :-)

stephentoub:
> but this is now being superseded by https://github.com/dotnet/extensions/pull/6383

Hmm. How does that help with `GetResponseAsync<T>`?

eiriktsarpalis:
We debated whether we should replace the boolean flag with a schema conformance mode enum, however that requirement is rendered obsolete by the fact that individual leaf clients apply their own transformations on the response format being constructed by this method.

stephentoub:
> leaf clients apply their own transformations

Ah, of course, great

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientStructuredOutputExtensions.cs
stephentoub: I believe Steve 's intent was we might be able to detect in the future whether the underlying provider supports structured output or not

PR #6395, opened 2025-05-08
Remove CacheOptions from DiskBasedResponseCache

### LINKED ISSUE #6387: [AI Evaluation] Remove CacheOptions from DiskBasedResponseCache
`DiskBasedResponseCacheProvider` supports an options file that can be used to tweak the default time-to-live for cache entries and a cache mode setting which can be used to completely turn off caching, or to completely block calls to the LLM when there is a cache miss.

`AzureStorageResponseCacheProvider` does not support similar options file or cache mode functionality - instead it supports tweaking the default time-to-live in code via a constructor parameter.

The cache modes functionality as well as the options file above is pretty specialized and should not be required in normal usage scenarios. We ideally don't want to support this functionality out of the box in the eval library. For advanced usage scenarios that require the special mode above to block LLM calls on cache misses, it should be possible (and relatively easy) to implement a custom `IResponseCacheProvider`s with that behavior.

Logging this issue to remove the options file and modes functionality, and to align the constructors for `DiskBasedResponseCacheProvider` and `AzureStorageResponseCacheProvider` such that both support tweaking the default time-to-live for cache entries via code.

shyamnamboodiripad:
Details for this change are available in #6387.

Fixes #6387 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6395)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/Storage/DiskBasedResponseCacheProvider.cs
```diff
@@ -20,7 +20,13 @@ namespace Microsoft.Extensions.AI.Evaluation.Reporting.Storage;
...
 /// The path to a directory on disk under which the cached AI responses should be stored.
 /// </param>
-public sealed class DiskBasedResponseCacheProvider(string storageRootPath) : IResponseCacheProvider
+/// <param name="timeToLiveForCacheEntries">
+/// An optional <see cref="TimeSpan"/> that specifies the maximum amount of time that cached AI responses should
+/// survive in the cache before they are considered expired and evicted.
+/// </param>
+public sealed class DiskBasedResponseCacheProvider(
+    string storageRootPath,
+    TimeSpan? timeToLiveForCacheEntries = null) : IResponseCacheProvider
```
shyamnamboodiripad, line 29: @peterwald The constructors for `DiskBasedResponseCacheProvider` and `AzureStorageResponseCacheProvider` are now aligned (both take an optional parameter for time-to-live). Do you think we should wrap this setting in `*CacheOptions` classes and change the constructors to use those instead?

* I can't foresee needing other settings beyond the time-to-live at the moment so `*CacheOptions` classes may be overkill for the single setting we have at the moment.
* We'll probably want to create separate `*CacheOptions` classes for `AzureStrogeResponseCacheProvider` and `DiskBasedResponseCacheProvider` since we would want to reserve the option for the set of settings needed for each provider to diverge in the future. Again, this seems like overkill for the single setting that we have at the moment.
* If end up needing more settings in the future, and if any of those settings don't have reasonable defaults (i.e. if any of them require caller supplied values), then it would likely be a breaking change regardless of whether or not we introduce `*CacheOptions` classes now.
* If we end up needing more settings in the future, and all of those settings have reasonable defaults, we will have the option to introduce `*CacheOptions` classes (and additional constructor overloads for `DiskBasedResponseCacheProvider` and `AzureResponseCacheProvider`) at that point without breaking callers.

So, I think it may be better to leave the current API shape in place and punt on the `*CacheOptions` classes for now. Please let me know if you feel otherwise.

peterwald (approved):

PR #6398, opened 2025-05-08
Add BinaryEmbedding

stephentoub:
Also:
- Renames the polymorphic discriminators to conform with typical lingo for these types.
- Adds an Embedding.Dimensions virtual property.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6398)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        _vector = Throw.IfNull(vector);
+    }
+
+    /// <summary>Gets or sets the embedding vector this embedding represents.</summary>
+    [JsonConverter(typeof(VectorConverter))]
+    public BitArray Vector
+    {
+        get => _vector;
+        set => _vector = Throw.IfNull(value);
```
eiriktsarpalis, line 33: I presume we're OK with making the embedding mutable?

stephentoub: Yes, the base properties and those on `Embedding<T>` are as well, e.g.
https://github.com/dotnet/extensions/blob/8271d4e7a3fba362915660eb81c92f39ab3364f6/src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/Embedding%7BT%7D.cs#L21

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <inheritdoc/>
+        public override void Write(Utf8JsonWriter writer, BitArray value, JsonSerializerOptions options)
+        {
+            _ = Throw.IfNull(writer);
+            _ = Throw.IfNull(value);
+            _ = Throw.IfNull(options);
+
+            writer.WriteStartObject();
+
+            writer.WritePropertyName("length");
```
eiriktsarpalis, line 106: Is this implementing a schema that is standard for binary embeddings or is it just for convenience in roundtripping within .NET applications?

eiriktsarpalis: Given that we're using base64 I think transmitting the value modulo 8 would suffice.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
```diff
@@ -0,0 +1,121 @@
...
+        set => _vector = Throw.IfNull(value);
+    }
+
+    /// <inheritdoc />
+    [JsonIgnore]
+    public override int Dimensions => _vector.Length;
+
+    /// <summary>Provides a <see cref="JsonConverter{BitArray}"/> for serializing <see cref="BitArray"/> instances.</summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class VectorConverter : JsonConverter<BitArray>
```
eiriktsarpalis, line 42: A fully serialized embedding would look as follows:
```json
{
   "vector": { "length": xxx, "bits": xxx }
}
```
Not sure if I'm a fan of the nestedness here. If it's possible to make the embedding value immutable, perhaps one alternative would be to

1. Hide the BitArray behind a read-only indexer and
2. Expose a pair of `Length` and `Bits` properties on the embedding itself (perhaps with EB.Never annotation) that map from the input `BitArray`.

roji: Should we just have a bitstring for the JSON representation, i.e. `"vector": "01101010111"`, rather than length + base64?

roji (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
stephentoub: > Is this implementing a schema that is standard for binary embeddings or is it just for convenience in roundtripping within .NET applications?

The latter.

> Given that we're using base64 I think transmitting the value modulo 8 would suffice.

Can you elaborate?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
stephentoub: > If it's possible to make the embedding value immutable

The embedding types are mutable. I don't see us changing that now.

> Should we just have a bitstring for the JSON representation, i.e. "vector": "01101010111", rather than length + base64?

I went with base64 to keep it smaller, e.g. a 1024 bit vector would be ~170 characters with base64 vs 1024 characters with such a character-per-bit encoding. But if we don't care about that, sure, we could do "0110..." or "[0, 1, 1, 0, ...]" to match other types.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
eiriktsarpalis: > Can you elaborate?

Not super important. I was simply pointing out that transmitting the value of `length % 8` would suffice since the base64 representation already incorporates the number of bytes that fit the embedding.

stephentoub:
Updated it to serialize as "0101011..."

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
```diff
@@ -0,0 +1,118 @@
...
+            else
+            {
+                utf8 = reader.ValueSpan;
+            }
+
+            BitArray result = new(utf8.Length);
+
+            for (int i = 0; i < utf8.Length; i++)
+            {
+                result[i] = utf8[i] switch
```
eiriktsarpalis, line 80: This does not account for potential escaping that might exist in the underlying raw string, e.g. the string "01" can be rendered equivalently as `"\u0030\u0031"`. You should consider using the `reader.CopyString` which copies strings to a destination buffer in unescaped form.

Here's an example of how it's being used in one of the built-in STJ converters: https://github.com/dotnet/runtime/blob/0a290e814abaf99f5aaeaacebe76d2e71175db0f/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Value/TimeSpanConverter.cs#L42-L52

stephentoub:
@eiriktsarpalis, better?

ladeak (approved):

jeffhandley (approved):
Looks good to me. I turned off auto-merge so that this can wait for a re-review by @eiriktsarpalis.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            _ = Throw.IfNull(typeToConvert);
+            _ = Throw.IfNull(options);
+
+            if (reader.TokenType != JsonTokenType.String)
+            {
+                throw new JsonException("Expected string property.");
+            }
+
+            ReadOnlySpan<byte> utf8;
```
eiriktsarpalis, line 55: Nit: much of this branching can be simplified by writing
```C#
int length = reader.HasValueSequence ? checked((int)reader.ValueSequence.Length) : reader.ValueSpan.Length;
```
In practice, it's vanishingly likely for the underlying sequence to be larger than `int.MaxValue`. What's more, `JsonSerializer` never uses `ReadOnlySpan<byte>` readers and always goes with spans.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                if (reader.HasValueSequence)
+                {
+                    long sequenceLength = reader.ValueSequence.Length;
+                    if (sequenceLength > int.MaxValue)
+                    {
+                        throw new JsonException("Value is too long.");
+                    }
+
+                    tmpArray = ArrayPool<byte>.Shared.Rent((int)sequenceLength);
```
eiriktsarpalis, line 71: Consider using a stack allocated buffer for small lengths, e.g.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Value/EnumConverter.cs#L244-L246

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
stephentoub: Are you saying I shouldn't validate the sequence length? The structure is as it is in order to avoid accessing ValueSequence twice.

eiriktsarpalis: You shouldn't need to access `ValueSequence` twice, just get the length to allocate an appropriate buffer and then pass that to `CopyString`.

stephentoub: If the length overflows, then it's relying on either ArrayPool renting or CopyString failing, which feels weak. Can that not happen for some reason?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
stephentoub: Is it worth it? The 99.999% case should be deserializing strings we created, which means they shouldn't have any escape sequences in them, and it sounds like it'll basically never be a `ReadOnlySequence<byte>` if I understood you correctly.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/BinaryEmbedding.cs
stephentoub: > while avoiding the duplicated code in the Span and Sequence cases.

The duplicated code being the Rent call?

Issue #6030, opened 2025-03-04
Update CachingChatClient with callbacks for if caching should happen


stephentoub:
The caching chat client doesn't work great when dealing with an OpenAI Assistant-like entity, where the ChatThreadId is non-null and remains the same across all operations. In such cases, the provided chat messages and options are insufficient to uniquely identify state in the cache.

We should:
- Disable caching by default when ChatOptions.ChatThreadId is set
- Provide either virtual methods or delegate properties that allow determining on a per-call basis whether to check the cache or cache a result

stephentoub:
The first bullet is handled by https://github.com/dotnet/extensions/pull/6400.

The second bullet is still open.

PR #6400, opened 2025-05-08
Avoid caching in CachingChatClient when ConversationId is set

stephentoub:
Contributes to https://github.com/dotnet/extensions/issues/6030
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6400)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
 
 #pragma warning disable S127 // "for" loop stop conditions should be invariant
+#pragma warning disable SA1202 // Elements should be ordered by access
 
 namespace Microsoft.Extensions.AI;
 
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/CachingChatClient.cs
```diff
@@ -178,4 +195,13 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
     /// <exception cref="ArgumentNullException"><paramref name="key"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langword="null"/>.</exception>
     protected abstract Task WriteCacheStreamingAsync(string key, IReadOnlyList<ChatResponseUpdate> value, CancellationToken cancellationToken);
+
+    /// <summary>Determine whether to use caching with the request.</summary>
+    private static bool UseCaching(ChatOptions? options)
+    {
+        // We want to skip caching if options.ConversationId is set. If it's set, that implies there's
+        // some state that will impact the response and that's not represented in the messages. Since
+        // that state could change even with the same ID, we have to assume caching isn't valid.
```
shyamnamboodiripad, line 204: Interesting - this would be something to keep in mind if we see any reports around response caching not working in the context of eval (cc @peterwald).

Evaluators themselves are unlikely to specify ConversationId, so this is unlikely to impact caching for the evaluation / scoring responses. But this could impact caching of the model responses being evaluated (since it is possible that the ChatOptions.ConversationId can be specified somewhere in the product / agent code, and this would turn off the caching even if the test code were passing down a caching IChatClient when fetching the agent response that is to be evaluated).

Just noting as something to double-check / keep in mind in case we hear any reports around eval caching not working as intended :)

ladeak (approved):

jeffhandley (approved):
Thanks for tackling this, @stephentoub!

PR #6408, opened 2025-05-09
Fix streaming chat response example

gewarren:
This was noticed in https://github.com/dotnet/docs/issues/46054. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6408)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
@@ -97,14 +97,12 @@ while (true)
     Console.Write("Q: ");
     history.Add(new(ChatRole.User, Console.ReadLine()));
 
-    List<ChatResponseUpdate> updates = [];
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/README.md
```diff
@@ -97,14 +97,12 @@ while (true)
     Console.Write("Q: ");
     history.Add(new(ChatRole.User, Console.ReadLine()));
 
-    List<ChatResponseUpdate> updates = [];
     await foreach (var update in client.GetStreamingResponseAsync(history))
     {
         Console.Write(update);
+        history.AddMessages(update);
```
stephentoub, line 103: This isn't the right fix. The right fix is to keep exactly what was there before and add here as L103 `updates.Add(update);`.

stephentoub (approved):
Thanks.

PR #6412, opened 2025-05-10
Move AIFunctionFactory down to M.E.AI.Abstractions

stephentoub:
We've debated doing this for a long time. Now's the time.

The first commit reverts some recently added capabilities that rely on interfaces from M.E.DI.Abstractions, which isn't available in M.E.AI.Abstractions. The features can be layered on top via extensibility.

The second commit moves the files down to M.E.AI.Abstractions.

The third commit adds one more extensibility point to AIFunctionFactoryOptions, which is necessary to replace the use of ActivatorUtilities.CreateInstance.

No meaningful code changes were made to AIFunctionFactory, other than the cited third commit.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6412)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactoryOptions.cs
```diff
@@ -107,14 +107,22 @@ public AIFunctionFactoryOptions()
...
+    /// <see cref="AIFunctionFactory.Create(MethodInfo, Type, AIFunctionFactoryOptions?)"/> creates <see cref="AIFunction"/> instances that invoke an
+    /// instance method on the specified <see cref="Type"/>. This delegate is used to create the instance of the type that will be used to invoke the method.
+    /// By default if <see cref="CreateInstance"/> is <see langword="null"/>, <see cref="Activator.CreateInstance(Type)"/> is used. If
+    /// <see cref="CreateInstance"/> is non-<see langword="null"/>, the delegate is invoked with the <see cref="Type"/> to be instantiated and the
+    /// <see cref="AIFunctionArguments"/> provided to the <see cref="AIFunction.InvokeAsync"/> method.
+    /// </para>
+    /// <para>
+    /// Each created instance will be used for a single invocation. If the object is <see cref="IAsyncDisposable"/> or <see cref="IDisposable"/>, it will
+    /// be disposed of after the invocation completes.
+    /// </para>
```
stephentoub, line 123: @halter73, @eiriktsarpalis, one thing we can decide after this merges for a follow-up... should this delegate actually apply for all instance methods and not just ones created without an instance?

jeffhandley (approved):

eiriktsarpalis:
Nice. This change should open up the possibility of us marking some of the schema related `AIJsonUtilities` methods as internal, slightly reducing concept duplication. Is this something we should consider or are we too late?

stephentoub:
> Nice. This change should open up the possibility of us marking some of the schema related `AIJsonUtilities` methods as internal, slightly reducing concept duplication. Is this something we should consider or are we too late?

@eiriktsarpalis, I think we still can, especially if they're more esoteric.

PR #6416, opened 2025-05-10
Update CHANGELOGs for M.E.AI

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6416)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Release History
 
+## 9.4.4-preview.1.25259.16
+
+- Added `AIJsonUtilities.TransformSchema` and supporting types.
+- Added `BinaryEmbedding` for bit embeddings.
+- Added `ChatOptions.RawRepresentationFactory` to make it easier to pass options to the underlying service.
+- Added `Base64Data` property to `DataContent`.
+- Moved `AIFunctionFactory` to `Microsoft.Extensions.AI.Abstractions`.
+- Fixed `AIFunctionFactory` handling of default struct arguments.
+
 ## 9.4.3-preview.1.25230.7
 
 - Renamed `ChatThreadId` to `ConversationId` on `ChatResponse`, `ChatResponseUpdate`, and `ChatOptions`.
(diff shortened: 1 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Release History
 
+## 9.4.4-preview.1.25259.16
+
+- Added `AIJsonUtilities.TransformSchema` and supporting types.
+- Added `BinaryEmbedding` for bit embeddings.
```
SteveSandersonMS, line 6: Is `BinaryEmbedding` exclusively for one-bit-per-dimension embeddings (i.e., ones where hamming distance is a sensible similarity metric), or does it mean "any binary blob that somehow represents an embedding"? Makes me wonder if the naming is too loose, or if the description here is too specific.

src/Libraries/Microsoft.Extensions.AI/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 9.4.4-preview.1.25259.16
+
+- Fixed `CachingChatClient` to avoid caching when `ConversationId` is set.
+- Renamved `useJsonSchema` parameter in `GetResponseAsync<T>` to `useJsonSchemaResponseFormat`.
```
SteveSandersonMS, line 6:
```suggestion
- Renamed `useJsonSchema` parameter in `GetResponseAsync<T>` to `useJsonSchemaResponseFormat`.
```

SteveSandersonMS (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
stephentoub: It's intended for 1-bit embeddings. It's backed by BitArray.

SteveSandersonMS: So would the name `BitEmbedding` be more applicable?

stephentoub: We could of you think it's important, but binary embedding seems like the much more common term. When I search for bit embedding, the titles of the articles that come back are all about binary embeddings.

cc: @roji

SteveSandersonMS: OK, that seems totally reasonable. I was focused on .NET conventions but agree that the favoring the industry-standard term for this makes sense.

PR #6419, opened 2025-05-11
Add WriteAsync overrides to stream helper in AIFunctionFactory

stephentoub:
We use JsonSerializer.SerializeAsync but were missing the async overrides. As with MemoryStream, these don't need to queue.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6419)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -26,6 +26,7 @@ using Microsoft.Shared.Diagnostics;
 #pragma warning disable CA1031 // Do not catch general exception types
 #pragma warning disable S2333 // Redundant modifiers should not be used
 #pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable SA1202 // Public members should come before private members
 
 namespace Microsoft.Extensions.AI;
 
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -1105,6 +1106,34 @@ public override void Flush()
...
+#if NET
+        public override
+#else
+        private
+#endif
+        ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
+        {
+            EnsureNotDisposed();
+
+            if (cancellationToken.IsCancellationRequested)
```
ladeak, line 1124: Why not `ThrowIfCancellationRequested`? due to perf of exception?

stephentoub: This method isn't `async`. The cancellation should be propagated as part of the task.

eiriktsarpalis:
> As with MemoryStream, these don't need to queue.

Can you clarify what you mean by this? Does the default async method implementation work differently?

stephentoub:
> Does the default async method implementation work differently?

The base implementations effectively do Task.Run(() => Write(...))

PR #6427, opened 2025-05-12
Add an AIJsonSchemaTransformOptions property inside AIJsonSchemaCreateOptions and mark redundant properties in the latter as obsolete.

eiriktsarpalis:
Apart from obsoletions, I decided to be a bit more aggressive and additionally remove the implementations of the obsoleted properties.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6427)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -43,7 +43,7 @@ public sealed record class AIJsonSchemaCreateOptions
     public AIJsonSchemaTransformOptions? TransformOptions { get; init; }
 
     /// <summary>
-    /// Gets a value indicating whether to include the type keyword in inferred schemas for .NET enums.
+    /// Gets a value indicating whether to include the type keyword in created schemas for .NET enums.
     /// </summary>
     [Obsolete("This property has been deprecated.")]
     [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
@@ -57,7 +57,7 @@ public sealed record class AIJsonSchemaCreateOptions
     public bool DisallowAdditionalProperties { get; init; }
 
     /// <summary>
-    /// Gets a value indicating whether to include the $schema keyword in inferred schemas.
+    /// Gets a value indicating whether to include the $schema keyword in created schemas.
     /// </summary>
     public bool IncludeSchemaKeyword { get; init; }
 
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
@@ -38,14 +38,23 @@ public sealed record class AIJsonSchemaCreateOptions
     public Func<ParameterInfo, bool>? IncludeParameter { get; init; }
 
     /// <summary>
+    /// Gets a <see cref="AIJsonSchemaTransformOptions"/> governing transformations on the JSON schema after it has been generated.
+    /// </summary>
+    public AIJsonSchemaTransformOptions? TransformOptions { get; init; }
+
+    /// <summary>
     /// Gets a value indicating whether to include the type keyword in inferred schemas for .NET enums.
     /// </summary>
+    [Obsolete("This property has been deprecated.")]
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
     public bool IncludeTypeInEnumSchemas { get; init; } = true;
 
     /// <summary>
     /// Gets a value indicating whether to generate schemas with the additionalProperties set to false for .NET objects.
     /// </summary>
-    public bool DisallowAdditionalProperties { get; init; } = true;
+    [Obsolete("This property has been deprecated. Use the equivalent property in TransformOptions instead.")]
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+    public bool DisallowAdditionalProperties { get; init; }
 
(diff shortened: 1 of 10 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -37,15 +37,24 @@ public sealed record class AIJsonSchemaCreateOptions
...
 
+    /// <summary>
+    /// Gets a <see cref="AIJsonSchemaTransformOptions"/> governing transformations on the JSON schema after it has been generated.
+    /// </summary>
+    public AIJsonSchemaTransformOptions? TransformOptions { get; init; }
+
     /// <summary>
     /// Gets a value indicating whether to include the type keyword in inferred schemas for .NET enums.
     /// </summary>
+    [Obsolete("This property has been deprecated.")]
```
eiriktsarpalis, line 48: Note that this property has no equivalent in the the transform options because it is now hardcoded as the default behavior.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaTransformOptions.cs
```diff
@@ -38,6 +38,11 @@ public sealed record class AIJsonSchemaTransformOptions
     /// </summary>
     public bool UseNullableKeyword { get; init; }
 
+    /// <summary>
+    /// Gets a value indicating whether to move the default keyword to the description field in the schema.
+    /// </summary>
+    public bool MoveDefaultKeywordToDescription { get; init; }
```
eiriktsarpalis, line 44: This behavior was originally coupled to `RequireAllProperties` but I've now factored it into a standalone setting since they are logically independent.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -37,15 +37,24 @@ public sealed record class AIJsonSchemaCreateOptions
...
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
     public bool IncludeTypeInEnumSchemas { get; init; } = true;
```
stephentoub, line 50: I don't see an equivalent for IncludeTypeInEnumSchemas in transform options. I see it's defaulting to true, so presumably we're emitting the information, and then if someone didn't want it they'd write a transform to remove it? It's not common enough to need being exposed on the transform options?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
```diff
@@ -55,5 +64,7 @@ public sealed record class AIJsonSchemaCreateOptions
     /// <summary>
```
stephentoub, line 64: IncludeSchemaKeyword isn't being obsoleted, I see. We don't need something on the transform options to remove a schema keyword if it's there but problematic?

Also, what does "in inferred schemas" mean in the comment. Are there kinds of schemas AIJsonSchemaCreateOptions creates that aren't "inferred"? I assume not?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
eiriktsarpalis: https://github.com/dotnet/extensions/pull/6427#discussion_r2085159460

eiriktsarpalis: Adding the `type` keyword to enums is redundant, but it's also benign. There is no reason why we'd want to switch this off.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaCreateOptions.cs
eiriktsarpalis: It's a holdover from when the `CreateJsonSchema` method was called `InferSchema`. We also use the term `inferenceOptions` in a few places as well.

eiriktsarpalis: > IncludeSchemaKeyword isn't being obsoleted, I see. We don't need something on the transform options to remove a schema keyword if it's there but problematic?

I have no evidence of this causing problems. We can add it if we get feedback but it should be fairly trivial to express using the transformer delegate.

eiriktsarpalis: Maybe? It seems we need three states: no change, ensure that it exists, and ensure that it's removed. We might also want to consider users customizing the schema url.

eiriktsarpalis: IOW it might make sense to hold off until a requirement arises.

eiriktsarpalis: The only use case is structured responses. @SteveSandersonMS requested at the time because it was needed as a hint for certain LLMs to understand that they're dealing with JSON schema (presumably in the case where we're inlining the schema into the prompt). 

A `bool IncludeSchemaKeyword` option makes sense when you're generating a schema from scratch, but not so much when you're transforming potentially arbitrary schema documents that may already contain a schema uri which may or may not match the schema uri we want to use.

stephentoub (approved):

PR #6428, opened 2025-05-12
Eliminate ingestion cache from AI Chat Web template

SteveSandersonMS:
This makes use of MEVD's new `GetAsync(...)` overload that can retrieve records using a search expression without doing nearest-neighbour search. It means we can eliminate the SQLite and EF dependencies at least in the qdrant and Azure AI Search cases, and eliminates cases where ingestion tracking can get out of sync with the chunk storage.

Because of updating to newer MEVD, many APIs changed so I had to do a lot of tangential updates, including reimplementing `JsonVectorStore` almost entirely, hence the PR diff looking complex there. But conceptually I haven't changed how that works. I just had to implement a different interface.

I did take the opportunity to do various renames and cleanups. It's still far from perfect but is a significant step forwards, and we can make it a whole lot better still when we can eliminate `JsonVectorStore`.

@roji I looked into using the InMemoryVectorStore's JSON dumping capability as discussed, but it didn't work out. It's not a good match for what we're doing here because it works on a per-collection basis and doesn't auto-write updates to disk when the vector store is changes, so it leaks out of the `IVectorStore` abstraction. While there may be ways to make it work well enough, since we're planning to eliminate it in favour of SQLite anyway, I concluded that it was a distraction for now.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6428)


MackinnonBuck (approved):
Looks great!

We'll need to update template snapshots before the template tests start passing again. Let me know if you'd like any help with that.

jeffhandley (commented):
Thanks for tackling this, @SteveSandersonMS. I'll update the integration test snapshots and push to your branch to get the tests passing.

joperezr (approved):
Changes to eng/Versions.props LGTM

PR #6433, opened 2025-05-13
Add RawRepresentationFactory to other options types

jozkee:
Continuation of https://github.com/dotnet/extensions/pull/6319
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6433)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGenerationOptions.cs
```diff
@@ -31,6 +33,25 @@ public int? Dimensions
     /// <summary>Gets or sets additional properties for the embedding generation request.</summary>
     public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
 
+    /// <summary>
+    /// Gets or sets a callback responsible of creating the raw representation of the embedding generation options from an underlying implementation.
```
SteveSandersonMS, line 37:
```suggestion
    /// Gets or sets a callback responsible for creating the raw representation of the embedding generation options from an underlying implementation.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGenerationOptions.cs
```diff
@@ -31,6 +33,25 @@ public int? Dimensions
...
+    /// <remarks>
+    /// The underlying <see cref="IEmbeddingGenerator" /> implementation may have its own representation of options.
+    /// When <see cref="IEmbeddingGenerator{TInput, TEmbedding}.GenerateAsync" /> 
+    /// is invoked with an <see cref="EmbeddingGenerationOptions" />, that implementation may convert the provided options into
+    /// its own representation in order to use it while performing the operation. For situations where a consumer knows
+    /// which concrete <see cref="IEmbeddingGenerator" /> is being used and how it represents options, a new instance of that
+    /// implementation-specific options type may be returned by this callback, for the <see cref="IEmbeddingGenerator" />
+    /// implementation to use instead of creating a new instance. Such implementations may mutate the supplied options
+    /// instance further based on other settings supplied on this <see cref="EmbeddingGenerationOptions" /> instance or from other inputs,
+    /// therefore, its <b>strongly recommended</b> to not return shared instances and instead make the callback return a new instance per each call.
```
SteveSandersonMS, line 48:
```suggestion
    /// therefore, it is <b>strongly recommended</b> to not return shared instances and instead make the callback return a new instance per each call.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/SpeechToText/SpeechToTextOptions.cs
```diff
@@ -24,6 +26,25 @@ public class SpeechToTextOptions
...
+    /// <remarks>
+    /// The underlying <see cref="ISpeechToTextClient" /> implementation may have its own representation of options.
+    /// When <see cref="ISpeechToTextClient.GetTextAsync" /> or <see cref="ISpeechToTextClient.GetStreamingTextAsync"/>
+    /// is invoked with an <see cref="SpeechToTextOptions" />, that implementation may convert the provided options into
+    /// its own representation in order to use it while performing the operation. For situations where a consumer knows
+    /// which concrete <see cref="ISpeechToTextClient" /> is being used and how it represents options, a new instance of that
+    /// implementation-specific options type may be returned by this callback, for the <see cref="ISpeechToTextClient" />
+    /// implementation to use instead of creating a new instance. Such implementations may mutate the supplied options
+    /// instance further based on other settings supplied on this <see cref="SpeechToTextOptions" /> instance or from other inputs,
+    /// therefore, its <b>strongly recommended</b> to not return shared instances and instead make the callback return a new instance per each call.
```
SteveSandersonMS, line 41:
```suggestion
    /// therefore, it is <b>strongly recommended</b> to not return shared instances and instead make the callback return a new instance per each call.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -124,7 +124,7 @@ public class ChatOptions
     /// implementation-specific options type may be returned by this callback, for the <see cref="IChatClient" />
     /// implementation to use instead of creating a new instance. Such implementations may mutate the supplied options
     /// instance further based on other settings supplied on this <see cref="ChatOptions" /> instance or from other inputs,
-    /// like the enumerable of <see cref="ChatMessage"/>s, therefore, its **strongly recommended** to not return shared instances
+    /// like the enumerable of <see cref="ChatMessage"/>s, therefore, its <b>strongly recommended</b> to not return shared instances
     /// and instead make the callback return a new instance per each call.
```
SteveSandersonMS, line 128:
```suggestion
    /// and instead make the callback return a new instance on each call.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGenerationOptions.cs
```diff
@@ -31,6 +33,25 @@ public int? Dimensions
...
+    /// This is typically used to set an implementation-specific setting that isn't otherwise exposed from the strongly-typed
+    /// properties on <see cref="EmbeddingGenerationOptions" />.
+    /// </remarks>
+    [JsonIgnore]
+    public Func<IEmbeddingGenerator, object?>? RawRepresentationFactory { get; set; }
```
SteveSandersonMS, line 53: I find it a bit surprising that we pass only the `IEmbeddingGenerator` in and not the `EmbeddingGenerationOptions`. Is the idea that any information on `EmbeddingGenerationOptions` is already being transferred into the raw representation by MEAI code and that there aren't any cases where developers would want to write custom logic to populate the raw representation based on the `EmbeddingGenerationOptions`?

stephentoub: That's the thinking, and that whoever is creating the ChatOptions is also creating the callback, and can populate either or both however they like.

You think this and the other options should be updated to accept the options as an additional argument?

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceEmbeddingGenerator.cs
```diff
@@ -164,16 +164,35 @@ static void ThrowInvalidData() =>
...
+                EncodingFormat = EmbeddingEncodingFormat.Base64,
+            };
+        }
+
+        if (options.RawRepresentationFactory?.Invoke(this) is EmbeddingsOptions result)
+        {
+            // Input doesn't have a setter, so use reflection to set it.
+            typeof(EmbeddingsOptions)
+                .GetField("<Input>k__BackingField", BindingFlags.NonPublic | BindingFlags.Instance)!
+                .SetValue(result, inputs.ToList());
```
jozkee, line 184: Heads up on this reflection.

stephentoub, line 184: This is unfortunate. I think the right answer is going to be to just ignore RawRepresentationFactory in this implementation for now, i.e. revert this file. There's really nothing meaningful on the EmbeddingOptions that someone might provide, and something like EncodingFormat if set could actually mess up our parsing of the response.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceImageEmbeddingGenerator.cs
```diff
@@ -117,16 +117,37 @@ void IDisposable.Dispose()
...
+                EncodingFormat = EmbeddingEncodingFormat.Base64,
+            };
+        }
+
+        if (options.RawRepresentationFactory?.Invoke(this) is ImageEmbeddingsOptions result)
+        {
+            // Input doesn't have a setter, so use reflection to set it.
+            typeof(ImageEmbeddingsOptions)
+                .GetField("<Input>k__BackingField", BindingFlags.NonPublic | BindingFlags.Instance)!
+                .SetValue(result, imageEmbeddingInputs.ToList());
```
stephentoub, line 139: Ditto

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceEmbeddingGenerator.cs
jozkee: I choose this approach considering that we can get rid of reflection on https://github.com/jozkee/azure-sdk-for-net/pull/1.

stephentoub: I see. That's reasonable, the trouble is if the dependency revs and changes this field, it'll break.

Can we keep this one as it was but do the non-reflection / internals access in your other branch where it's possible?

jozkee: OK, will do.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Embeddings/EmbeddingGenerationOptions.cs
SteveSandersonMS: > You think this and the other options should be updated to accept the options as an additional argument?

I was imagining what is needed to enable factoring out the mapping into a central, reusable method, e.g.:

```cs
var chatOptions = new ChatOptions()
{
    RawRepresentationFactory = MyApp.Utils.MapChatOptions,
    AdditionalProperties = { "ReasoningLevel": "high" }
};
```

with

```cs
public static class Utils
{
    public object MapChatOptions(IChatClient chatClient, ChatOptions options)
    {
        if (chatClient is OpenAIChatClient)
        {
            var reasoningLevel = options.AdditionalProperties.TryGetValue("ReasoningLevel") ? ... ;
            return new OpenAI.ChatOptions() { ReasoningLevel = reasoningLevel };
        }
        else if (chatClient is ...) {  ... }
    }
}
```

... so that you don't have to duplicate logic for constructing provider-specific options at all the call sites.

But I guess an alternative is to do it with nested callbacks, e.g.:

```cs
var chatOptions = new ChatOptions()
{
    RawRepresentationFactory = MyApp.Utils.CreateChatOptions(reasoningLevel: "high")
};
```

with

```cs
public static class Utils
{
    public Func<IChatClient, object> CreateChatOptions(string reasoningLevel)
    {
        return chatClient =>
        {
            if (chatClient is OpenAIChatClient)
            {
                return new OpenAI.ChatOptions() { ReasoningLevel = reasoningLevel };
            }
            else if (chatClient is ...) {  ... }
        };
    }
}
```

Or even just:

```cs
var chatOptions = new ChatOptions()
{
    RawRepresentationFactory = chatClient => MyApp.Utils.CreateChatOptions(chatClient, reasoningLevel: "high")
};
```

Maybe that's actually nicer. So yeah I think it's fine as-is.

src/Libraries/Microsoft.Extensions.AI.AzureAIInference/AzureAIInferenceEmbeddingGenerator.cs
stephentoub: > would it be appropriate to go with RawRepresentationFactory here or does the EncodingFormat concern is enough reason to revert?

We could write the code to successfully handle EmbeddingEncodingFormat.Float (and .Base64 is already handled), which should be trivially handled via deserialization. We would want to throw if it's set to other values, as they don't make sense with an `Embedding<float>` output. We can subsequently add overloads of `AsIEmbeddingGenerator` that produce other embedding types (e.g. `BinaryEmbedding`) if that's deemed valuable, but by far the most common use today is with float.

jozkee: Can we address all of this, including enabling RawRepresentationFactory in AI.Inference in a follow up PR to unblock this one?

stephentoub: Yes

jozkee: Then, if there's no more feedback, please approve.

stephentoub (approved):

PR #6442, opened 2025-05-14
[9.5] Remove unused API in Safety package (#6439)

shyamnamboodiripad:
This change ensures that we will have just one way to create a ChatConfiguration for the Safety evaluators.

The removed API was also confusing because it could mislead callers into thinking that the returned IChatClient can be used like any other general purpose LLM endpoint when in fact it is pretty specialized and only intended for internal use within the Safety evaluators.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6442)


shyamnamboodiripad:
FYI @joperezr @jeffhandley Cherry-picking this simple change to release/9.5 for inclusion in the upcoming release

peterwald (approved):

PR #6444, opened 2025-05-15
Couple of fixes for safety evaluators

shyamnamboodiripad:
Includes following changes -

**Add a convenience overload:**
  Currently, if a caller wants to run evaluations that involve both safety and quality evals, they need to create one ChatConfiguration for the LLM interactions and then wrap it with another ChatConfiguration for the AI Foundry service (by calling the existing overload for ToChatConfiguration).
  
  The new overload makes it so that caller only needs to create a IChatClient for the LLM interactions needed for the quality evals. They can then directly call ToChatConfiguration with this IChatClient to get a single Chatconfiguration (which supports both safety and quality evals) and supply this ChatConfiguration as part of the ReportingConfiguration.

**Improve an error message:**
  The NotSupportedException being thrown before was pretty opaque. Improve the error so that the reason is a bit clearer.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6444)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyChatClient.cs
@@ -9,6 +9,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyChatClient.cs
```diff
@@ -171,4 +168,20 @@ private static void ValidateSingleMessage(IEnumerable<ChatMessage> messages)
...
+    {
+        if (_originalChatClient is null)
+        {
+            string errorMessage =
+                $"""
+                Failed to invoke '{nameof(IChatClient)}.{callerMemberName}()'.
+                Did you forget to specify the argument value for 'originalChatClient' or 'originalChatConfiguration' when calling '{nameof(ContentSafetyServiceConfiguration)}.ToChatConfiguration()'?
+                """;
+
+            Throw.ArgumentNullException(nameof(_originalChatClient), errorMessage);
```
shyamnamboodiripad, line 184: @gewarren I am hoping this improves the error for the issue you ran into.

peterwald (approved):

PR #6445, opened 2025-05-15
Use RawRepresentationFactory on AzureAIInference embedding generators

jozkee:
As per offline discussion, we currently don't have to support other encoding formats other than base64, therefore, we hardcode it on both cases 1) when there's appropriate underlying options returned by RawRepresentationFactory 2) when there's not.

https://github.com/dotnet/extensions/pull/6433#discussion_r2087236788
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6445)


jeffhandley (approved):

SteveSandersonMS (approved):
Approving on the assumption that the "overwrite `EncodingFormat`" behavior here matches the offline discussion you refer to (with @stephentoub?).

In the earlier PR #6433 there was a statement that:

> We would want to throw if it's set to other values

... which is different to the behavior here, but I'm fine with it either way if this matches some consensus that has already been reached.

stephentoub (approved):

stephentoub:
> Approving on the assumption that the "overwrite EncodingFormat" behavior here matches the offline discussion you refer to (with @stephentoub?).

Yeah, I realized afterwards that overwriting is reasonable and simpler. First, the chances that someone is setting this is low, and the chances that they're setting it to something incompatible with the type of the embedding generator (e.g. `Embedding<float>`) is even lower. But regardless of what they set it to, we're the ones parsing the output and converting it into the format required to satisfy the type of the return type, so even if someone requested something else, we could be quantizing the results into the expected format. The simplest thing, then, is to just always use the format we know we're parsing and always give the user back the thing the type is supposed to give back.

PR #6447, opened 2025-05-15
Add `AIFunction.ReturnJsonSchema`

eiriktsarpalis:


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -38,6 +38,9 @@ public abstract class AIFunction : AITool
     /// </remarks>
     public virtual JsonElement JsonSchema => AIJsonUtilities.DefaultJsonSchema;
 
+    /// <summary>Gets a JSON Schema describing the function's return value.</summary>
+    public virtual JsonElement ReturnJsonSchema => AIJsonUtilities.DefaultJsonSchema;
```
stephentoub, line 42: Should the property be nullable instead of returning a "{}" schema? If someone had an object return type, that would logically be a "{}" schema as well, right? How does someone determine whether the schema being returned just happens to match an empty object or is unknown?

eiriktsarpalis: > If someone had an object return type, that would logically be a "{}" schema as well, right?

That would have been `{ "type" : "object" }`. `{}` represents _any_ JSON value.

stephentoub: What would our recommendation be then for determining whether the ReturnJsonSchema can be trusted or not?
```C#
if (!JsonElement.DeepEquals(func.ReturnJsonSchema, AIJsonUtilities.DefaultJsonSchema) { ... }
```
is much more complex than:
```C#
if (func.ReturnJsonSchema is JsonElement schema) { ... }
```

eiriktsarpalis: Not trusted in what sense? It should still correctly specify the output in a trivial sense.

stephentoub: As in someone is deriving from AIFunction today and hasn't overridden the property, so it's return "{}" even though the function returns meaningful data.

stephentoub: I'm saying, this is using "{}" as a sentinel to indicate that no data is being returned, yes? In which case, what is the code the consumer needs to write to know whether they want to use the schema (because the API chose to return something meaningful) vs not (because the API didn't provide anything and is just getting back the base behavior)?

eiriktsarpalis: > I'm saying, this is using "{}" as a sentinel to indicate that no data is being returned, yes?

I wouldn't say so. `{}` could be returned for a multitude of reasons: the return type of the underlying method is `object?` or `JsonElement`, or that a custom converter is being used preventing us from reasoning about the shape. `{}` does not denote that the schema is unusable or that no data is being returned; it is appropriate to use even though it makes a trivial assertion about the returned value.

stephentoub: For example, if an McpServerTool is constructed from AIFunction, I would only want to populate it a outputSchema if the developer that built the AIFunction explicitly populated the property, rather than it being a default. How do figure that out if the default value of {} is valid?

eiriktsarpalis: We _could_ choose to make it nullable but I think this would put it at odds with the existing `JsonSchema` property which is not nullable. One might similarly require a way to determine if the input schema has been explicitly overridden.

> How do figure that out if the default value of {} is valid?

Rather than thinking of this in terms of valid vs invalid schemas, I prefer a distinction of trivial vs. meaningful schemas. In the context of MCP, I think it makes sense to advertise output schemas if and only if the schema is nontrivial.

stephentoub: > One might similarly require a way to determine if the input schema has been explicitly overridden.

Possibly, but there's a) a difference between promises about what you accept vs what you return, and b) virtuals added before you created your derived type and after.

stephentoub: For example, if I wanted to implement a piece of middleware that augmented the prompt sent to the LLM with return value details, how should I be deciding to send it? Is your recommendation that I should do:
```C#
if (!JsonElement.DeepEquals(func.ReturnJsonSchema, AIJsonUtilities.DefaultJsonSchema) { ... }
```
to make that determination? To me, even if the method is returning JsonElement, there's a distinction between the developer having explicitly chosen for that to be the return signature and it happening because the AIFunction hasn't been updated to provide anything better. I don't want to consume a bunch of tokens and muddy up the request if the developer wasn't explicitly choosing to give me information that was deemed useful.

If you really don't think there's a reason to make the distinction, then we can go with what you have. I'm just nervous it's throwing away meaningful information.

eiriktsarpalis: > Is your recommendation that I should do ...

That would be my thinking, since from my perspective it should be equivalent to specifying nothing and therefore a `{}` return schema is redundant.

> I'm just nervous it's throwing away meaningful information.

It might be meaningful from the perspective of how the AIFunction is defined, but I just don't see how it could be from the perspective of the AIFunction contract.

If there is an argument to be made about making the return schema nullable, I think it might be to represent functions whose return type is void/Task/ValueTask. Today all canonical `AIFunction` implementations return `JsonElement` if they return a value, but `null` (not JSON null) if they are void. I think an argument could be made about the need to match the semantics of `InvokeAsync` to the returned schema that way, but this concern is completely orthogonal to whether the virtual has been explicitly implemented.

eiriktsarpalis: @stephentoub thoughts on this bit?

> If there is an argument to be made about making the return schema nullable, I think it might be to represent functions whose return type is void/Task/ValueTask. Today all canonical AIFunction implementations return JsonElement if they return a value, but null (not JSON null) if they are void. I think an argument could be made about the need to match the semantics of InvokeAsync to the returned schema that way, but this concern is completely orthogonal to whether the virtual has been explicitly implemented.

eiriktsarpalis: I updated the property to be nullable and made the void returning functions report a null schema.

PR #6453, opened 2025-05-16
Add a script to diff the contents of folders that match a specific pattern across two branches

shyamnamboodiripad:
Over recent releases, there have been multiple occasions where I found myself needing to compare the contents of two branches to figure out what is different for a subset of libraries (such as *Evaluation*).

I'd like to check in the script that I have been using for this since this may prove useful again in the future.

### DIFF
--- a/scripts/DiffBranches.ps1
+++ b/scripts/DiffBranches.ps1
@@ -1,0 +1,76 @@
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory=$false)]
+    [string]$baseline = "main",
+    
+    [Parameter(Mandatory=$true)]
+    [string]$target,
+    
+    [Parameter(Mandatory=$false)]
+    [string]$folderPattern = "*.AI.*",
+    
+    [Parameter(Mandatory=$false)]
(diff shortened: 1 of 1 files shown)


scripts/DiffBranches.ps1
```diff
@@ -0,0 +1,76 @@
+[CmdletBinding()]
```
RussKie, line 1: It would be good to add some docs explaining how to use the script.
Ideally, something like in https://github.com/dotnet/extensions/blob/main/scripts/Slngen.ps1 or https://github.com/dotnet/extensions/blob/main/eng/scripts/ValidateProjectCoverage.ps1

RussKie (approved):
With suggestion.

PS: You can also use [Git Extensions](https://github.com/gitextensions/gitextensions) :)

jeffhandley (approved):
Cool; thanks! I'll use this too.

scripts/DiffBranches.ps1
shyamnamboodiripad: Thanks - added

PR #6454, opened 2025-05-16
Fix issues in the MEAI Templates

jeffhandley:
A couple template issues were found during testing of the MEAI 9.5.0 stable release:

1. The vector index name replacement token wasn't matching in the configuration anymore
    - This regressed in #6428. While reviewing that PR, I had the fleeting thought that maybe the replacement logic would be affected, but I didn't verify it.
2. GitHub now documents that the personal access token needs to have `models:read` permissions
    - Our README for using GitHub Models was misleading and needed to be updated.

Additionally, we had to make changes in the `release/9.5` branch to get the template builds working correctly against the stable package versions for MEAI and MEAI.Abstractions while continuing to work with preview versions for the other packages and the `-dev` version locally. Those changes were merged back into main in this PR, but different incompatibilities arose with that in place. A new approach is taken in this PR where the versions of packages referenced by the template are now all defined directly in the `GeneratedContents.targets` file, and instead of toggling pinned vs. unpinned, versions are just specified either as hard-coded versions or as `$(Version)`/`$(VersionPrefix)`. We will still have to modify these values in release branches, but it should be simplified and easier to manage.

After this is merged into `main`, I will port it into the `release/9.5` branch and we will be able to publish an updated version of the Microsoft.Extensions.AI.Templates package that references the 9.5.0 stable version of Microsoft.Extensions.AI/Abstractions.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6454)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/README.Aspire.md
```diff
@@ -125,7 +125,7 @@ To use Azure AI Search, you will need an Azure account and an Azure AI Search re
 ### 1. Create an Azure AI Search Resource
 Follow the instructions in the [Azure portal](https://portal.azure.com/) to create an Azure AI Search resource. Note that there is a free tier for the service but it is not currently the default setting on the portal.
 
-Note that if you previously used the same Azure AI Search resource with different model using this project name, you may need to delete your `data-ChatWithCustomData-CSharp.Web-ingestion` index using the [Azure portal](https://portal.azure.com/) first before continuing; otherwise, data ingestion may fail due to a vector dimension mismatch.
+Note that if you previously used the same Azure AI Search resource with different model using this project name, you may need to delete your `data-ChatWithCustomData-CSharp.Web-chunks` and `data-ChatWithCustomData-CSharp.Web-chunks` indexes using the [Azure portal](https://portal.azure.com/) first before continuing; otherwise, data ingestion may fail due to a vector dimension mismatch.
```
SteveSandersonMS, line 128:
```suggestion
Note that if you previously used the same Azure AI Search resource with different model using this project name, you may need to delete your `data-ChatWithCustomData-CSharp.Web-chunks` and `data-ChatWithCustomData-CSharp.Web-documents` indexes using the [Azure portal](https://portal.azure.com/) first before continuing; otherwise, data ingestion may fail due to a vector dimension mismatch.
```

src/ProjectTemplates/GeneratedContent.targets
MackinnonBuck, line 1: This new approach looks good to me.

Since predicting these package versions accurately seems to be a recurring challenge, I wonder if we should consider dynamically determining the versions of just-built packages rather than hard-coding them. We do this in the `aspnetcore` repo [here](https://github.com/dotnet/aspnetcore/blob/302f094c992a653acc6378bd2bc19444587fac2b/src/ProjectTemplates/GenerateContent.targets#L16-L35), although I was initially hesitant to implement that solution here because of the added complexity to the build. However, I'm starting to think it might be warranted. Not something we should do in this PR, but the sooner we do it, the sooner we can (hopefully) stop worrying about manually managing these versions.

MackinnonBuck (approved):

joperezr (approved):

PR #6465, opened 2025-05-20
Remove `title` and `description` keywords from root-level schemas in AIFunctionFactory.

eiriktsarpalis:
As an alternative solution to https://github.com/modelcontextprotocol/csharp-sdk/pull/436, this PR removes the `title` and `description` keywords from root-level schemas in AIFunctionFactory. This is due to the fact both keywords are already represented in the resultant `AIFunction` via the `Name` and `Description` properties respectively, and because keywords beyond `type`, `properties`, and `required` is known to create problems with certain clients in the context of MCP.

Because both the `name` and `description` keywords are optional in the schema, I do not consider this to be a breaking change.

Fix https://github.com/modelcontextprotocol/csharp-sdk/issues/342. Supersedes https://github.com/modelcontextprotocol/csharp-sdk/pull/436

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -692,10 +692,10 @@ public static partial class AIFunctionFactory
             JsonSerializerOptions = serializerOptions;
             JsonSchema = AIJsonUtilities.CreateFunctionJsonSchema(
                 key.Method,
-                Name,
-                Description,
-                serializerOptions,
-                schemaOptions);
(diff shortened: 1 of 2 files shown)


stephentoub (approved):

stephentoub:
I'm happy with the change, but what changed your opinion since https://github.com/dotnet/extensions/pull/5826#discussion_r1934397818... that it breaks some consumers?

eiriktsarpalis:
> what changed your opinion since [#5826 (comment)](https://github.com/dotnet/extensions/pull/5826#discussion_r1934397818)... that it breaks some consumers?

We had no reports of this creating issues with consumers until https://github.com/modelcontextprotocol/csharp-sdk/issues/342.

Issue #5624, opened 2024-11-11
Microsoft.Extensions.Caching.Hybrid is not trim/AOT compatible


eerhardt:
When analyzing the Microsoft.Extensions.Caching.Hybrid library, we get the following warnings

```
D:\git\extensions\src\Libraries\Microsoft.Extensions.Caching.Hybrid\Internal\DefaultJsonSerializerFactory.cs(26): Trim analysis error IL2026: Microsoft.Extensions.Caching.Hybrid.Internal.DefaultJsonSerializerFactory.DefaultJsonSerializer`1.Microsoft.Extensions.Caching.Hybrid.IHybridCacheSerializer<T>.Deserialize(ReadOnlySequence`1<Byte>): Using member 'System.Text.Json.JsonSerializer.Deserialize<T>(Utf8JsonReader&,JsonSerializerOptions)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. JSON serialization and deserialization might require types that cannot be statically analyzed. Use the overload that takes a JsonTypeInfo or JsonSerializerContext, or make sure all of the required types are preserved. [D:\git\extensions\test\Libraries\Microsoft.Extensions.AotCompatibility.TestApp\Microsoft.Extensions.AotCompatibility.TestApp.csproj]
    D:\git\extensions\src\Libraries\Microsoft.Extensions.Caching.Hybrid\Internal\DefaultJsonSerializerFactory.cs(26): AOT analysis error IL3050: Microsoft.Extensions.Caching.Hybrid.Internal.DefaultJsonSerializerFactory.DefaultJsonSerializer`1.Microsoft.Extensions.Caching.Hybrid.IHybridCacheSerializer<T>.Deserialize(ReadOnlySequence`1<Byte>): Using member 'System.Text.Json.JsonSerializer.Deserialize<T>(Utf8JsonReader&,JsonSerializerOptions)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. JSON serialization and deserialization might require types that cannot be statically analyzed and might need runtime code generation. Use System.Text.Json source generation for native AOT applications. [D:\git\extensions\test\Libraries\Microsoft.Extensions.AotCompatibility.TestApp\Microsoft.Extensions.AotCompatibility.TestApp.csproj]
[...]

PR #6475, opened 2025-05-21
HybridCache : simplify AOT usage with default JSON serialization

mgravell:
1. don't touch `JsonSerializerOptions.Default` in AOT mode
2. add new extension methods to register JsonSerializerOptions
3. offer a suitable error message instead, if none supplied in AOT mode
4. include `[UnconditionalSuppressMessage]` to prevent build-time consumer warnings

This has been tested successfully with an AOT published project; it now works correctly at runtime with zero build-time warnings, test code shown below; without the `WithJsonSerializerOptions` line, exception at runtime:

>  When using AOT, JsonSerializerOptions with TypeInfoResolver specified must be provided via IHybridCacheBuilder.WithJsonSerializerOptions.

Without this work, it is unclear how to register AOT JSON options, and the consumer must use `<NoWarn>$(NoWarn);IL2026;IL2104;IL3053</NoWarn>` or similar.

open questions:

- naming: With* or Add* ?
- which API review process should this follow?

Sample:

``` c#
// Licensed to the .NET Foundation under one or more agreements.

using System.Text.Json;
using Microsoft.Extensions.Caching.Hybrid;

var jsonOptions = new JsonSerializerOptions() { TypeInfoResolver = Bar.MyJsonSerializationContext.Default };
services.AddHybridCache().WithJsonSerializerOptions(jsonOptions);
services.AddStackExchangeRedisCache(options => options.Configuration = "localhost:6379");
using var provider = services.BuildServiceProvider();

var cache = provider.GetRequiredService<HybridCache>();

int id = 42;
var obj = await cache.GetOrCreateAsync($"/ob/{id}", ct => GetTheThingAsync(id));
Console.WriteLine(obj.Id);

static ValueTask<Bar.Foo> GetTheThingAsync(int id) => new(new Bar.Foo(id));

namespace Bar
{
    [JsonSourceGenerationOptions(WriteIndented = true)]
    [JsonSerializable(typeof(Foo))]
    internal partial class MyJsonSerializationContext : JsonSerializerContext
    {
    }

    internal sealed class Foo(int id)
    {
        public int Id => id;
    }
}
```

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheBuilderExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheBuilderExtensions.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
 using Microsoft.Extensions.Caching.Hybrid;
 using Microsoft.Shared.Diagnostics;
 
@@ -57,6 +58,27 @@ public static class HybridCacheBuilderExtensions
         where TImplementation : class, IHybridCacheSerializerFactory
     {
         _ = Throw.IfNull(builder).Services.AddSingleton<IHybridCacheSerializerFactory, TImplementation>();
+        return builder;
+    }
+
+    /// <summary>
+    /// Register a default <see cref="JsonSerializerOptions"/> for use with JSON serialization.
+    /// </summary>
+    /// <returns>The <see cref="IHybridCacheBuilder"/> instance.</returns>
+    public static IHybridCacheBuilder WithJsonSerializerOptions(this IHybridCacheBuilder builder, JsonSerializerOptions options)
+    {
+        _ = Throw.IfNull(builder).Services.AddKeyedSingleton<JsonSerializerOptions>(typeof(IHybridCacheSerializer<>), Throw.IfNull(options));
+        return builder;
+    }
+
+    /// <summary>
+    /// Register a <see cref="JsonSerializerOptions"/> for use with JSON serialization of type <typeparamref name="T"/>.
+    /// </summary>
+    /// <typeparam name="T">The type being serialized.</typeparam>
+    /// <returns>The <see cref="IHybridCacheBuilder"/> instance.</returns>
+    public static IHybridCacheBuilder WithJsonSerializerOptions<T>(this IHybridCacheBuilder builder, JsonSerializerOptions options)
+    {
+        _ = Throw.IfNull(builder).Services.AddKeyedSingleton<JsonSerializerOptions>(typeof(IHybridCacheSerializer<T>), Throw.IfNull(options));
         return builder;
     }
 }
--- a/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
@@ -10,6 +10,10 @@ using System.Reflection;
 using System.Text.Json;
 using Microsoft.Extensions.DependencyInjection;
 
+[assembly: UnconditionalSuppressMessage("AOT", "IL2026", Justification = "Checked at runtime, guidance issued")]
+[assembly: UnconditionalSuppressMessage("AOT", "IL2070", Justification = "Checked at runtime, guidance issued")]
+[assembly: UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Checked at runtime, guidance issued")]
+
 namespace Microsoft.Extensions.Caching.Hybrid.Internal;
 
 internal sealed class DefaultJsonSerializerFactory : IHybridCacheSerializerFactory
@@ -34,7 +38,7 @@ internal sealed class DefaultJsonSerializerFactory : IHybridCacheSerializerFacto
 
         // see if there is a per-type options registered (keyed by the **closed** generic type), otherwise use the default
         JsonSerializerOptions options = _serviceProvider.GetKeyedService<JsonSerializerOptions>(typeof(IHybridCacheSerializer<T>)) ?? Options;
-        if (!options.IncludeFields && ReferenceEquals(options, SystemDefaultJsonOptions) && IsFieldOnlyType(typeof(T)))
+        if (!options.IncludeFields && IsDefaultJsonOptions(options) && IsFieldOnlyType(typeof(T)))
         {
             // value-tuples expose fields, not properties; special-case this as a common scenario
             options = FieldEnabledJsonOptions;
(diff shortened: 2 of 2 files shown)


mgravell:
CI build fail is because of API delta:

> error LA0003: (NETCORE_ENGINEERING_TELEMETRY=Build) Newly added symbol 'Microsoft.Extensions.DependencyInjection.HybridCacheBuilderExtensions.WithJsonSerializerOptions(Microsoft.Extensions.Caching.Hybrid.IHybridCacheBuilder, System.Text.Json.JsonSerializerOptions)' must be marked as experimental

This in in turn tied into the open "which API review process should this follow?" question.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -50,11 +54,42 @@ internal static bool IsFieldOnlyType(Type type)
         return IsFieldOnlyType(type, ref state) == FieldOnlyResult.FieldOnly;
     }
 
+    private static bool IsDefaultJsonOptions(JsonSerializerOptions options)
+    {
+#if NETCOREAPP3_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+        if (!System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported)
```
eerhardt, line 60: Instead of IsDynamicCodeSupported, these checks should be based on `JsonSerializer.IsReflectionEnabledByDefault`. That's a feature switch that is disabled in both trimmed and AOT'd apps.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -10,6 +10,10 @@
 using System.Text.Json;
 using Microsoft.Extensions.DependencyInjection;
 
+[assembly: UnconditionalSuppressMessage("AOT", "IL2026", Justification = "Checked at runtime, guidance issued")]
+[assembly: UnconditionalSuppressMessage("AOT", "IL2070", Justification = "Checked at runtime, guidance issued")]
+[assembly: UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Checked at runtime, guidance issued")]
```
eerhardt, line 15:
```suggestion
```

We shouldn't be suppressing these warnings for the whole assembly. Instead, we should just add the suppressions in the exact places we know are OK to suppress.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheBuilderExtensions.cs
```diff
@@ -59,4 +61,27 @@ public static IHybridCacheBuilder AddSerializerFactory<
         _ = Throw.IfNull(builder).Services.AddSingleton<IHybridCacheSerializerFactory, TImplementation>();
         return builder;
     }
+
+    /// <summary>
+    /// Register a default <see cref="JsonSerializerOptions"/> for use with JSON serialization.
+    /// </summary>
+    /// <returns>The <see cref="IHybridCacheBuilder"/> instance.</returns>
+    [Experimental(DiagnosticIds.Experiments.HybridCache, UrlFormat = DiagnosticIds.UrlFormat)]
+    public static IHybridCacheBuilder WithJsonSerializerOptions(this IHybridCacheBuilder builder, JsonSerializerOptions options)
```
eerhardt, line 70: This API and design is a little different from how we do it elsewhere (minimal APIs and SignalR). Instead of giving a full JsonSerializerOptions, the developer "configures" a single JsonSerializerOptions. I guess the case the HybridCache covers is if you want a completely different set of options for a specific type.

See 

https://github.com/dotnet/aspnetcore/blob/aebd07af9854541970d0c05d99dc2e6e8bd2fdd4/src/ProjectTemplates/Web.ProjectTemplates/content/WebApiAot-CSharp/Program.cs#L6-L9

for an example at how people configure the JsonSerializerOptions for minimal APIs.

cc @halter73 @BrennanConroy

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
mgravell: Great feedback, will tweak

src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheBuilderExtensions.cs
mgravell: Happy to refactor to any preferred pattern

eerhardt: >  if you want a completely different set of options for a specific type

Is that a hard requirement? It doesn't seem to be possible minimal APIs and SignalR today, AFAICT.

Not that I'm against this design, if it is necessary. But I guess I would have thought we would follow the same design/pattern as the other 2 places that do this.

cc @captainsafia as well

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -34,7 +38,7 @@ public bool TryCreateSerializer<T>([NotNullWhen(true)] out IHybridCacheSerialize
 
         // see if there is a per-type options registered (keyed by the **closed** generic type), otherwise use the default
         JsonSerializerOptions options = _serviceProvider.GetKeyedService<JsonSerializerOptions>(typeof(IHybridCacheSerializer<T>)) ?? Options;
```
eerhardt, line 40: Do we need to change this code for the open generic `typeof(IHybridCacheSerializer<>)`?

src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheBuilderExtensions.cs
mgravell: It looks like the overall aim could be met by inserting multiple type handlers to a single options, but: I'll check

eerhardt: Talking it over with @captainsafia, since you could return an entirely different SerializerFactory per Type (i.e. use Json for some types, and XML for others), having this design / API allowing for different JsonSerializerOptions per Type seems OK.

eerhardt:
This should be able to resolve [Microsoft.Extensions.Caching.Hybrid is not trim/AOT compatible (dotnet/extensions#5624)](https://github.com/dotnet/extensions/issues/5624).

eerhardt:
We should remove 

https://github.com/dotnet/extensions/blob/6f29ab505709370fc9b72bf4bb7e004a7662f69f/test/Libraries/Microsoft.Extensions.AotCompatibility.TestApp/Microsoft.Extensions.AotCompatibility.TestApp.csproj#L20-L21

as well. We should make the library trimming and AOT compatible.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Microsoft.Extensions.Caching.Hybrid.csproj
```diff
@@ -23,6 +23,11 @@
     <DisableMicrosoftExtensionsLoggingSourceGenerator>false</DisableMicrosoftExtensionsLoggingSourceGenerator>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">
+    <IsTrimmable>true</IsTrimmable>
+    <IsAotCompatible>true</IsAotCompatible>
```
eerhardt, line 28:
```suggestion
    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```

You only need to set `IsAotCompatible`. It will default `IsTrimmable`. See https://github.com/dotnet/sdk/blob/fd301ec19c3190b8129b28847d6faed9d26bc0d3/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets#L24

src/Libraries/Microsoft.Extensions.Caching.Hybrid/Internal/DefaultJsonSerializerFactory.cs
```diff
@@ -12,6 +12,9 @@
 
 namespace Microsoft.Extensions.Caching.Hybrid.Internal;
 
+[UnconditionalSuppressMessage("AOT", "IL2026", Justification = "Checked at runtime, guidance issued")]
+[UnconditionalSuppressMessage("AOT", "IL2070", Justification = "Checked at runtime, guidance issued")]
+[UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Checked at runtime, guidance issued")]
```
eerhardt, line 17: These should be moved even more local to only suppress around the lines of code that need it. Doing it the current way will suppress it for all new code written in this class, which isn't correct.

Also, is there a better Justifcation that can be written here? A little more of a reason why this is OK.

KLuuKer:
@eerhardt @sebastienros maybe assign a different person? (maybe one that still is employed by MS, heck why not give copilot a shot at it?)

mgravell:
>  sebastienros unassigned mgravell 3 hours ago
>
> dotnet-policy-service dotnet-policy-service bot assigned mgravell 3 hours ago

🎵 Welcome to the repo California... 
🎶 you can `checkout` any time you like, but you can never leave

sebastienros:
@mgravell I think you need to delete this account and create a new one

mjgaux:
(deletes `dotnet`, as requested)

mgravell:
(considers reporting `dotnet-policy-service` for... harassment? TOS violations? just "being a total jerk"?)

sebastienros:
FYI we hoped [this PR](https://github.com/dotnet/extensions/pull/7083) would fix it

stephentoub:
@joperezr what should happen with this PR?

PR #6500, opened 2025-06-07
Improve handling of RawRepresentation in OpenAI{Response}ChatClient

stephentoub:
We weren't always including the original object as the RawRepresentation, and we weren't returning every streaming update raw object. Make sure we do. Also improves the ability to break glass by checking arbitrary AIContent instances for a RawRepresentation that's then used as-is.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -261,6 +261,9 @@ internal sealed partial class OpenAIChatClient : IChatClient
 
             case DataContent dataContent when dataContent.MediaType.StartsWith("application/pdf", StringComparison.OrdinalIgnoreCase):
                 return ChatMessageContentPart.CreateFilePart(BinaryData.FromBytes(dataContent.Data), dataContent.MediaType, $"{Guid.NewGuid():N}.pdf");
+
+            case AIContent when content.RawRepresentation is ChatMessageContentPart rawContentPart:
+                return rawContentPart;
         }
 
         return null;
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
@@ -15,6 +15,7 @@ using Microsoft.Shared.Diagnostics;
 using OpenAI.Responses;
 using static Microsoft.Extensions.AI.OpenAIChatClient;
 
+#pragma warning disable S907 // "goto" statement should not be used
 #pragma warning disable S1067 // Expressions should not be too complex
 #pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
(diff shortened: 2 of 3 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -261,6 +261,9 @@ private static List<ChatMessageContentPart> ToOpenAIChatContent(IList<AIContent>
 
             case DataContent dataContent when dataContent.MediaType.StartsWith("application/pdf", StringComparison.OrdinalIgnoreCase):
                 return ChatMessageContentPart.CreateFilePart(BinaryData.FromBytes(dataContent.Data), dataContent.MediaType, $"{Guid.NewGuid():N}.pdf");
+
+            case AIContent when content.RawRepresentation is ChatMessageContentPart rawContentPart:
```
rogerbarreto, line 265: Shouldn't this be the first test and avoid all other transformations otherwise?

```suggestion
            case AIContent when content.RawRepresentation is ChatMessageContentPart rawContentPart:
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -570,6 +618,10 @@ private static List<ResponseContentPart> ToOpenAIResponsesContent(IList<AIConten
                 case ErrorContent errorContent when errorContent.ErrorCode == nameof(ResponseContentPartKind.Refusal):
                     parts.Add(ResponseContentPart.CreateRefusalPart(errorContent.Message));
                     break;
+
+                case AIContent when content.RawRepresentation is ResponseContentPart rawRep:
```
rogerbarreto, line 622: Ditto.

```suggestion
                case AIContent when content.RawRepresentation is ResponseContentPart rawRep:
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
stephentoub: I purposefully made it last so that it doesn't override changes made to the content by the developer. Otherwise, if someone did something like:
```C#
foreach (var m in chatResponse.Messages)
{
    foreach (var c in m.Content.OfType<TextContent>())
    {
        c.Text = Summarize(c.Text);
    }
}
```
and then sent those messages back as part of a subsequent request, the modifications would end up being ignored, as the original unmodified raw representation was used.

Instead, this is saying "If the type is known, use it, otherwise, defer to a raw representation if one exists". The main purpose here isn't roundtripping but breaking glass. If there's a type of content the developer wants to send as input and that the abstraction can't represent, they can do `new AIContent() { RawRepresentation = ... }`, creating the kind of content and sending it in.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
stephentoub: Same.

PR #6501, opened 2025-06-08
Bring back AsIChatClient for OpenAI AssistantClient

stephentoub:
cc: @markwallace-microsoft
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6501)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
@@ -42,21 +42,22 @@ internal sealed partial class OpenAIAssistantChatClient : IChatClient
     private readonly string _assistantId;
 
     /// <summary>The thread ID to use if none is supplied in <see cref="ChatOptions.ConversationId"/>.</summary>
-    private readonly string? _threadId;
+    private readonly string? _defaultThreadId;
 
     /// <summary>Initializes a new instance of the <see cref="OpenAIAssistantChatClient"/> class for the specified <see cref="AssistantClient"/>.</summary>
-    public OpenAIAssistantChatClient(AssistantClient client, string assistantId, string? threadId)
+    public OpenAIAssistantChatClient(AssistantClient assistantClient, string assistantId, string? defaultThreadId)
     {
-        _client = Throw.IfNull(client);
+        _client = Throw.IfNull(assistantClient);
         _assistantId = Throw.IfNullOrWhitespace(assistantId);
-        _threadId = threadId;
+
+        _defaultThreadId = defaultThreadId;
 
         // https://github.com/openai/openai-dotnet/issues/215
         // The endpoint isn't currently exposed, so use reflection to get at it, temporarily. Once packages
         // implement the abstractions directly rather than providing adapters on top of the public APIs,
         // the package can provide such implementations separate from what's exposed in the public API.
         Uri providerUrl = typeof(AssistantClient).GetField("_endpoint", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
-            ?.GetValue(client) as Uri ?? OpenAIResponseChatClient.DefaultOpenAIEndpoint;
+            ?.GetValue(assistantClient) as Uri ?? OpenAIResponseChatClient.DefaultOpenAIEndpoint;
 
         _metadata = new("openai", providerUrl);
     }
@@ -85,7 +86,7 @@ internal sealed partial class OpenAIAssistantChatClient : IChatClient
         (RunCreationOptions runOptions, List<FunctionResultContent>? toolResults) = CreateRunOptions(messages, options);
 
         // Get the thread ID.
-        string? threadId = options?.ConversationId ?? _threadId;
+        string? threadId = options?.ConversationId ?? _defaultThreadId;
         if (threadId is null && toolResults is not null)
         {
             Throw.ArgumentException(nameof(messages), "No thread ID was provided, but chat messages includes tool results.");
@@ -327,17 +328,22 @@ internal sealed partial class OpenAIAssistantChatClient : IChatClient
             }
         }
 
-        // Process ChatMessages. System messages are turned into additional instructions.
-        // All other messages are added 1:1, treating assistant messages as agent messages
-        // and everything else as user messages.
+        // Process ChatMessages.
         StringBuilder? instructions = null;
         List<FunctionResultContent>? functionResults = null;
         foreach (var chatMessage in messages)
         {
             List<MessageContent> messageContents = [];
 
+            // Assistants doesn't support system/developer messages directly. It does support transient per-request instructions,
+            // so we can use the system/developer messages to build up a set of instructions that will be passed to the assistant
+            // as part of this request. However, in doing so, on a subsequent request that information will be lost, as there's no
+            // way to store per-thread instructions in the OpenAI Assistants API. We don't want to convert these to user messages,
+            // however, as that would then expose the system/developer messages in a way that might make the model more likely
+            // to include that information in its responses. System messages should ideally be instead done as instructions to
+            // the assistant when the assistant is created.
             if (chatMessage.Role == ChatRole.System ||
-                chatMessage.Role == OpenAIChatClient.ChatRoleDeveloper)
+                chatMessage.Role == OpenAIResponseChatClient.ChatRoleDeveloper)
             {
                 instructions ??= new();
                 foreach (var textContent in chatMessage.Contents.OfType<TextContent>())
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
@@ -1,0 +1,440 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+using OpenAI.Assistants;
+
+#pragma warning disable CA1031 // Do not catch general exception types
+#pragma warning disable SA1005 // Single line comments should begin with single space
+#pragma warning disable SA1204 // Static elements should appear before instance elements
+#pragma warning disable S125 // Sections of code should not be commented out
+#pragma warning disable S907 // "goto" statement should not be used
+#pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S1751 // Loops with at most one iteration should be refactored
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable S4456 // Parameter validation in yielding methods should be wrapped
+#pragma warning disable S4457 // Parameter validation in "async"/"await" methods should be wrapped
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents an <see cref="IChatClient"/> for an Azure.AI.Agents.Persistent <see cref="AssistantClient"/>.</summary>
+[Experimental("OPENAI001")]
+internal sealed partial class OpenAIAssistantChatClient : IChatClient
+{
+    /// <summary>The underlying <see cref="AssistantClient" />.</summary>
+    private readonly AssistantClient _client;
+
+    /// <summary>Metadata for the client.</summary>
+    private readonly ChatClientMetadata _metadata;
+
+    /// <summary>The ID of the agent to use.</summary>
+    private readonly string _assistantId;
+
+    /// <summary>The thread ID to use if none is supplied in <see cref="ChatOptions.ConversationId"/>.</summary>
+    private readonly string? _threadId;
+
+    /// <summary>Initializes a new instance of the <see cref="OpenAIAssistantChatClient"/> class for the specified <see cref="AssistantClient"/>.</summary>
+    public OpenAIAssistantChatClient(AssistantClient client, string assistantId, string? threadId)
+    {
+        _client = Throw.IfNull(client);
+        _assistantId = Throw.IfNullOrWhitespace(assistantId);
+        _threadId = threadId;
+
+        // https://github.com/openai/openai-dotnet/issues/215
+        // The endpoint isn't currently exposed, so use reflection to get at it, temporarily. Once packages
+        // implement the abstractions directly rather than providing adapters on top of the public APIs,
+        // the package can provide such implementations separate from what's exposed in the public API.
+        Uri providerUrl = typeof(AssistantClient).GetField("_endpoint", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
+            ?.GetValue(client) as Uri ?? OpenAIResponseChatClient.DefaultOpenAIEndpoint;
+
+        _metadata = new("openai", providerUrl);
+    }
+
+    /// <inheritdoc />
+    public object? GetService(Type serviceType, object? serviceKey = null) =>
+        serviceType is null ? throw new ArgumentNullException(nameof(serviceType)) :
+        serviceKey is not null ? null :
+        serviceType == typeof(ChatClientMetadata) ? _metadata :
+        serviceType == typeof(AssistantClient) ? _client :
+        serviceType.IsInstanceOfType(this) ? this :
+        null;
+
+    /// <inheritdoc />
+    public Task<ChatResponse> GetResponseAsync(
+        IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default) =>
+        GetStreamingResponseAsync(messages, options, cancellationToken).ToChatResponseAsync(cancellationToken);
+
+    /// <inheritdoc />
+    public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
+        IEnumerable<ChatMessage> messages, ChatOptions? options = null, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(messages);
+
+        // Extract necessary state from messages and options.
+        (RunCreationOptions runOptions, List<FunctionResultContent>? toolResults) = CreateRunOptions(messages, options);
+
+        // Get the thread ID.
+        string? threadId = options?.ConversationId ?? _threadId;
+        if (threadId is null && toolResults is not null)
+        {
+            Throw.ArgumentException(nameof(messages), "No thread ID was provided, but chat messages includes tool results.");
+        }
+
+        // Get any active run ID for this thread. This is necessary in case a thread has been left with an
+        // active run, in which all attempts other than submitting tools will fail. We thus need to cancel
+        // any active run on the thread.
+        ThreadRun? threadRun = null;
+        if (threadId is not null)
+        {
+            await foreach (var run in _client.GetRunsAsync(
+                threadId,
+                new RunCollectionOptions { Order = RunCollectionOrder.Descending, PageSizeLimit = 1 },
+                cancellationToken: cancellationToken).ConfigureAwait(false))
+            {
+                if (run.Status != RunStatus.Completed && run.Status != RunStatus.Cancelled && run.Status != RunStatus.Failed && run.Status != RunStatus.Expired)
+                {
+                    threadRun = run;
+                }
+
+                break;
+            }
+        }
+
+        // Submit the request.
+        IAsyncEnumerable<StreamingUpdate> updates;
+        if (threadRun is not null &&
+            ConvertFunctionResultsToToolOutput(toolResults, out List<ToolOutput>? toolOutputs) is { } toolRunId &&
+            toolRunId == threadRun.Id)
+        {
+            // There's an active run and we have tool results to submit, so submit the results and continue streaming.
+            // This is going to ignore any additional messages in the run options, as we are only submitting tool outputs,
+            // but there doesn't appear to be a way to submit additional messages, and having such additional messages is rare.
+            updates = _client.SubmitToolOutputsToRunStreamingAsync(threadRun.ThreadId, threadRun.Id, toolOutputs, cancellationToken);
+        }
+        else
+        {
+            if (threadId is null)
+            {
+                // No thread ID was provided, so create a new thread.
+                ThreadCreationOptions threadCreationOptions = new();
+                foreach (var message in runOptions.AdditionalMessages)
+                {
+                    threadCreationOptions.InitialMessages.Add(message);
+                }
+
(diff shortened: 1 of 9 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
```diff
@@ -0,0 +1,440 @@
...
+#pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S1751 // Loops with at most one iteration should be refactored
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable S4456 // Parameter validation in yielding methods should be wrapped
+#pragma warning disable S4457 // Parameter validation in "async"/"await" methods should be wrapped
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents an <see cref="IChatClient"/> for an Azure.AI.Agents.Persistent <see cref="AssistantClient"/>.</summary>
+[Experimental("OPENAI001")]
```
rogerbarreto, line 32: Do we want to add Experimental to a `preview` package?

```suggestion
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
```diff
@@ -0,0 +1,440 @@
...
+internal sealed partial class OpenAIAssistantChatClient : IChatClient
```
rogerbarreto, line 33: I see a full blown implementation for the `OpenAIAssistantChatClient` could this be a specialization of `OpenAIChatClient`?

```suggestion
internal sealed partial class OpenAIAssistantChatClient : OpenAIChatClient
```

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
```diff
@@ -618,9 +618,9 @@ public virtual async Task Caching_AfterFunctionInvocation_FunctionOutputUnchange
 
         // Second time, the calls to the LLM don't happen, but the function is called again
         var secondResponse = await chatClient.GetResponseAsync([message]);
-        Assert.Equal(response.Text, secondResponse.Text);
         Assert.Equal(2, functionCallCount);
         Assert.Equal(FunctionInvokingChatClientSetsConversationId ? 3 : 2, llmCallCount!.CallCount);
+        Assert.Equal(response.Text, secondResponse.Text);
```
rogerbarreto, line 623: nit.
Is this change of position necessary?

```suggestion
        Assert.Equal(response.Text, secondResponse.Text);
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -25,6 +26,19 @@ public static IChatClient AsIChatClient(this ChatClient chatClient) =>
...
+    /// <param name="client">The <see cref="AssistantClient"/> instance to be accessed as an <see cref="IChatClient"/>.</param>
+    /// <param name="assistantId">The unique identifier of the assistant with which to interact.</param>
+    /// <param name="threadId">
+    /// An optional existing thread identifier for the chat session. This serves as a default, and may be overridden per call to
+    /// <see cref="IChatClient.GetResponseAsync"/> or <see cref="IChatClient.GetStreamingResponseAsync"/> via the <see cref="ChatOptions.ConversationId"/>
+    /// property. If no thread ID is provided via either mechanism, a new thread will be created for the request.
+    /// </param>
+    /// <returns>An <see cref="IChatClient"/> instance configured to interact with the specified agent and thread.</returns>
+    [Experimental("OPENAI001")]
+    public static IChatClient AsIChatClient(this AssistantClient client, string assistantId, string? threadId = null) =>
```
rogerbarreto: Following the same pattern for `defaultModelId` and `defaultDimensions`, I think all default settings for clients should also have `default` prefix for consistency.

```suggestion
    public static IChatClient AsIChatClient(this AssistantClient client, string assistantId, string? defaultThreadId = null) =>
```

rogerbarreto: Do we want to be sticky to the `assistantId` for the whole life of the client? If the `modelId` can be the default for normal chat clients which is the main target, I would suggest the `assistantId` to also be a `defaultAssistantId` which can be override per request basis.

```suggestion
    public static IChatClient AsIChatClient(this AssistantClient client, string defaultAssistantId, string? threadId = null) =>
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponseChatClient.cs
```diff
@@ -26,10 +26,7 @@ namespace Microsoft.Extensions.AI;
 internal sealed partial class OpenAIResponseChatClient : IChatClient
 {
     /// <summary>Gets the default OpenAI endpoint.</summary>
-    private static Uri DefaultOpenAIEndpoint { get; } = new("https://api.openai.com/v1");
-
-    /// <summary>A <see cref="ChatRole"/> for "developer".</summary>
-    private static readonly ChatRole _chatRoleDeveloper = new("developer");
+    internal static Uri DefaultOpenAIEndpoint { get; } = new("https://api.openai.com/v1");
```
rogerbarreto, line 30: `OpenAIChatClient` gets the DefaultOpenAIUri from Responses but responses get the `ChatRoleDeveloper` from the `OpenAIChatCLient`.

I would rather keep all in `OpenAIChatClient` and setting this as internal there.

```suggestion
```

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIAssistantChatClientIntegrationTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            AutomaticDecompression = System.Net.DecompressionMethods.GZip,
+        });
+
+        client.DefaultRequestHeaders.Add("accept", "*/*");
+        client.DefaultRequestHeaders.Add("accept-encoding", "gzip");
+        client.DefaultRequestHeaders.Add("accept-language", "en-US,en;q=0.9");
+        client.DefaultRequestHeaders.Add("openai-beta", "assistants=v2");
+        client.DefaultRequestHeaders.Add("origin", "https://platform.openai.com");
+        client.DefaultRequestHeaders.Add("user-agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36 Edg/137.0.0.0");
```
rogerbarreto: Is the user agent necessary here?

```suggestion
```

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIAssistantChatClientIntegrationTests.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        // These values need to be filled in. The bearer token needs to be sniffed from a browser
+        // session interacting with the dashboard (e.g. F12 networking tools).
+        client.DefaultRequestHeaders.Add("authorization", $"Bearer TODO-SESSION-TOKEN");
+        client.DefaultRequestHeaders.Add("openai-organization", "TODO");
+        client.DefaultRequestHeaders.Add("openai-project", "TODO");
```
rogerbarreto: Some TODO's in code, is this intentional?
Wonder if we have a tracking issue for the forthcoming changes to this commented `Fact`
```suggestion
        // These values need to be filled in. The bearer token needs to be sniffed from a browser
        // session interacting with the dashboard (e.g. F12 networking tools).
        client.DefaultRequestHeaders.Add("authorization", $"Bearer TODO-SESSION-TOKEN");
        client.DefaultRequestHeaders.Add("openai-organization", "TODO");
        client.DefaultRequestHeaders.Add("openai-project", "TODO");
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
stephentoub: The OpenAI assistant APIs themselves are [Experimental("OPENAI001")]. A developer using AssistantClient is going to get these errors and need to suppress them, and if I didn't have this here, I would also need to suppress it. Seemed better to just glob them together.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
stephentoub: > I see a full blown implementation for the OpenAIAssistantChatClient could this be a specialization of OpenAIChatClient?

The assistants API and the chat completion API are completely different. They wouldn't be able to share practically anything.

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
stephentoub: This is moving the text validation to after the numerical validation. So it's now first validating the structure is correct (e.g. how many calls were made) and only then validating that the content is right.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
stephentoub: That's reasonable.

stephentoub: > I would suggest the assistantId to also be a defaultAssistantId which can be override per request basis.

Overridden with what?

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIAssistantChatClientIntegrationTests.cs
stephentoub: They're "TODO" in that a developer that wants to use this needs to fill these in. OpenAI doesn't expose a list-threads API, so it's very challenging to clean up. This is just a bit of a helper code someone can manually execute to forcibly clean up threads; otherwise, they'll sit around until they're automatically deleted after some period of time.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIAssistantChatClientIntegrationTests.cs
stephentoub: From experimentation I thought it was, but it appears to work without it.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
rogerbarreto: `chatOptions.AdditionalProperties["assistantId"]` or a dedicated property in the abstraction for it or a specialized `ChatOptions`.

stephentoub: I don't think there should be a dedicated property in the abstraction, at least not until it's used with more endpoints that have the notion of an agent/assistant/participant ID. I'd also like to avoid encouraging derivations from ChatOptions, as it starts to get confusing where state should be provided, especially with the escape hatch of ChatOptions.RawRepresentationFactory, which lets you fully configure the underlying provider's options bag. And assistant ID isn't part of the assistant's object bag, because it's a required part of the underlying API.

We've used "default" on things that are optional. I think we'd be better off keeping it not "default" and non-optional. If we ever want to allow overriding it, we could make it nullable.

I'm also very cognizant of the OpenAI assistants API a) being beta at the OpenAI service level and experimental in the .NET client API, and b) already being deprecated by OpenAI.

asinghca:
I just wanted to thank you all so much for this update! 

I have recently discovered the two AI packages :

https://github.com/openai/openai-dotnet/blob/main/src/Custom/Assistants/AssistantClient.cs
https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs

Leveraging your updated chat client for assistant and the Open AI's assistant API client, I was able to replace a whole bunch of code with just this wrapper to meet my app level interface ILargeLanguageModelService. Prior to this, I had written my own implementation for all of it.

I am curious to know if I am using, the OpenAIAssistantChatClient : IChatClient correctly? I would appreciate any feedback as I have not able to find much documentation on this particular chat client other than the source code for it.

```
using UseCases.Models.Chat;

namespace UseCases.Interfaces.Chat;

/// <summary>
/// Defines a vendor-neutral contract for interacting with large language model (LLM) providers
/// using application-managed run identifiers. Abstracts the full lifecycle of response generation.
/// </summary>
public interface ILargeLanguageModelService
{
    /// <summary>
    /// Initiates a background response generation task for the specified chat thread and run ID.
    /// Implementations may use <see cref="IChatMessageStore"/> internally to read message context.
    /// </summary>
    /// <param name="chatId">The ID of the chat thread.</param>
    /// <param name="runId">The application-generated identifier for this response generation task.</param>
    /// <param name="ct">Optional cancellation token.</param>
    Task SendMessageAsync(string chatId, string runId, CancellationToken ct = default);

    /// <summary>
    /// Returns the current lifecycle status of the generation task.
    /// </summary>
    /// <param name="chatId">The chat thread ID.</param>
    /// <param name="runId">The application-assigned run ID.</param>
    /// <param name="ct">Optional cancellation token.</param>
    /// <returns>The current status of the generation process.</returns>
    Task<ResponseGenerationStatus> GetStatusAsync(string chatId, string runId, CancellationToken ct = default);

    /// <summary>
    /// Returns the most recent response content generated so far for the given task.
    /// If the task is still InProgress, this may return partial content or an empty string.
    /// If the task is Completed, the content is expected to be final.
    /// Returns null if no content has been generated yet or if the task failed.
    /// </summary>
    /// <param name="chatId">The chat thread ID.</param>
    /// <param name="runId">The application-assigned run ID.</param>
    /// <param name="ct">Optional cancellation token.</param>
    /// <returns>Partial or complete response content, or null if none is available.</returns>
    Task<string?> GetResponseAsync(string chatId, string runId, CancellationToken ct = default);

    /// <summary>
    /// Cancels a background response generation task if it is still in progress.
    /// </summary>
    /// <param name="chatId">The chat thread ID.</param>
    /// <param name="runId">The application-assigned run ID.</param>
    /// <param name="ct">Optional cancellation token.</param>
    Task CancelAsync(string chatId, string runId, CancellationToken ct = default);
}


```


```

using Infrastructure.ApiClient.OpenAI.Assistant;
using UseCases.Interfaces.Auth;
using UseCases.Interfaces.Chat;
using UseCases.Models.Chat;

using AIChatMessage = Microsoft.Extensions.AI.ChatMessage;
using ChatRole = Microsoft.Extensions.AI.ChatRole;

namespace Infrastructure.Services.LargeLanguageModel;

/// <summary>
/// Modern implementation of <see cref="ILargeLanguageModelService"/> using Microsoft.Extensions.AI abstractions.
/// Leverages <see cref="IChatClient"/> for OpenAI Assistant interactions with built-in thread management.
/// </summary>
public class OpenAIAssistantService : ILargeLanguageModelService
{
    private readonly IChatClient _chatClient;
    private readonly IAssistantThreadResponseStore _responseStore;
    private readonly IUserContext _user; // We might need this later
    private readonly IChatMessageStore _messageStore;
    private readonly ILogger<OpenAIAssistantService> _logger;
    /// <summary>
    /// Initializes a new instance of the <see cref="OpenAIAssistantService"/> class.
    /// </summary>
    /// <param name="serviceProvider">Service provider to resolve keyed IChatClient.</param>
    /// <param name="responseStore">Backing store for tracking OpenAI chat response state.</param>
    /// <param name="user">Context for the currently authenticated user.</param>
    /// <param name="messageStore">Chat message repository scoped by user and chat ID.</param>
    public OpenAIAssistantService(
        IChatClient chatClient,
        IAssistantThreadResponseStore responseStore,
        IUserContext user,
        IChatMessageStore messageStore,
        ILogger<OpenAIAssistantService> logger)
    {
        _chatClient = chatClient;
        _responseStore = responseStore;
        _user = user;
        _messageStore = messageStore;
        _logger = logger;
    }

    /// <inheritdoc />
    /// <remarks>
    /// 1. Only the latest user message is sent to the AI client.
    /// 2. It's a turn based conversation, so the AI client will respond to the latest user's message.
    /// 3. Since this is an Open AI Assistant client, the AI client is stateful and will remember the conversation.
    /// 4. There is no need to send the entire conversation history to the AI client.
    /// 5. Since we are also wrapping the <see cref="IChatClient"/> with a <see cref="FunctionInvokingChatClient"/>,
    /// the function calling is handled by the <see cref="FunctionInvokingChatClient"/>, so we don't need to handle it ourself.
    /// <see cref="ChatResponse.ConversationId"/> is used to track the created conversation state.
    /// <see cref="ChatOptions.ConversationId"/> is used to track the existing conversation state.
    /// </remarks>
    public async Task SendMessageAsync(string appChatId, string appRunId, CancellationToken ct = default)
    {
        var latest = await Task.Run(() =>
            _messageStore.GetLatestUserMessageForChat(appChatId), ct);
            

        if (latest == null)
            throw new InvalidOperationException("No user message found to respond to.");

        // Convert chat history to Microsoft.Extensions.AI format
        AIChatMessage? chatMessages = ConvertToChatMessages(latest);
        string? openAiThreadId = null;
        _responseStore.TryGetThreadId(appChatId, out openAiThreadId);

        // Ensure empty strings are treated as null to allow OpenAI to create new threads
        if (string.IsNullOrWhiteSpace(openAiThreadId))
        {
            openAiThreadId = null;
        }

        ChatOptions chatOptions = new ChatOptions
        {
            ConversationId = openAiThreadId
        };

        _logger.LogInformation("Starting response generation for chat {AppChatId} and run {AppRunId} with thread ID {ThreadId}", 
            appChatId, appRunId, openAiThreadId ?? "new thread");

        // Create cancellation token source for this specific task
        var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(ct);
        
        try
        {
            // Create and store the task execution state
            var executionState = new AiGenerationTaskState
            {
                Status = ResponseGenerationStatus.InProgress,
                CancellationTokenSource = cancellationTokenSource,
                StartedAtUtc = DateTime.UtcNow
            };

            // Start the assistant conversation asynchronously
            var backgroundTask = Task.Factory.StartNew(async () =>
            {
                try
                {
                    ChatResponse response = await _chatClient.GetResponseAsync(chatMessages, chatOptions, cancellationToken: cancellationTokenSource.Token);
                    
                    _responseStore.SaveChatResponse(appChatId, appRunId, response);

                    // Save thread ID if present in response (indicates new thread was created or existing thread was used)
                    if(response.ConversationId != null)
                    {
                        _responseStore.SaveThreadId(appChatId, response.ConversationId);
                        _logger.LogInformation("Saved thread ID {ThreadId} for chat {AppChatId}", response.ConversationId, appChatId);
                    }
                    
                    _responseStore.UpdateTaskStatus(appChatId, appRunId, ResponseGenerationStatus.Completed);
                }
                catch (OperationCanceledException)
                {
                    _logger.LogWarning("Response generation was cancelled for chat {AppChatId} and run {AppRunId}", appChatId, appRunId);
                    _responseStore.UpdateTaskStatus(appChatId, appRunId, ResponseGenerationStatus.Cancelled, "Operation was cancelled");
                }
                catch (Exception ex)
                {
                    _logger.LogError("Response generation failed for chat {AppChatId} and run {AppRunId}: {ErrorMessage}", appChatId, appRunId, ex.Message);
                    _responseStore.UpdateTaskStatus(appChatId, appRunId, ResponseGenerationStatus.Failed, ex.Message);
                }
            },
            cancellationTokenSource.Token,
            TaskCreationOptions.LongRunning | TaskCreationOptions.DenyChildAttach,
            TaskScheduler.Default);

            // Store the task execution state with the actual task
            executionState.ExecutingTask = backgroundTask;
            _responseStore.SaveTaskExecutionState(appChatId, appRunId, executionState);
        }
        catch (Exception ex)
        {
            _logger.LogError("Failed to start response generation for chat {AppChatId} and run {AppRunId}: {ErrorMessage}", appChatId, appRunId, ex.Message);
            _responseStore.UpdateTaskStatus(appChatId, appRunId, ResponseGenerationStatus.Failed, "Failed to start generation task");
            cancellationTokenSource.Dispose();
            throw;
        }
    }

    /// <inheritdoc />
    public async Task<ResponseGenerationStatus> GetStatusAsync(string appChatId, string appRunId, CancellationToken ct = default)
    {
        if (!_responseStore.TryGetTaskExecutionState(appChatId, appRunId, out AiGenerationTaskState? executionState))
            return ResponseGenerationStatus.Unknown;

        return await Task.FromResult(executionState!.Status);
    }

    /// <inheritdoc />
    public async Task<string?> GetResponseAsync(string appChatId, string appRunId, CancellationToken ct = default)
    {
        string? responseContent = null;

        if (_responseStore.TryGetChatResponse(appChatId, appRunId, out ChatResponse? response) 
        && response != null
        && response.Messages != null
        && response.Messages.Count > 0) {
            responseContent = response.Messages.Last().ToString();
        }

        return await Task.FromResult(responseContent);
    }

    /// <inheritdoc />
    public async Task CancelAsync(string appChatId, string appRunId, CancellationToken ct = default)
    {
        if (!_responseStore.TryGetTaskExecutionState(appChatId, appRunId, out AiGenerationTaskState? executionState))
            return;

        // Cancel the task if it's still running
        if (executionState!.CancellationTokenSource != null && !executionState.CancellationTokenSource.Token.IsCancellationRequested)
        {
            executionState.CancellationTokenSource.Cancel();
        }

        _responseStore.UpdateTaskStatus(appChatId, appRunId, ResponseGenerationStatus.Cancelled, "Cancelled by user request");
        
        await Task.CompletedTask;
    }

    /// <summary>
    /// Converts application chat message to Microsoft.Extensions.AI ChatMessage format.
    /// </summary>
    /// <param name="message">Application-specific chat message.</param>
    /// <returns>ChatMessage object for the AI client.</returns>
    private static AIChatMessage ConvertToChatMessages(ChatMessageBase message)
    {
        var content = message switch
            {
                UserPrompt userPrompt => userPrompt.Content,
                BotResponse botResponse => botResponse.Content,
                _ => null
            };

            if (string.IsNullOrEmpty(content)){
                throw new InvalidOperationException("User's message content is empty.");
            }

            var role = message.Role switch
            {
                ChatMessageRole.User => ChatRole.User,
                ChatMessageRole.Bot => ChatRole.Assistant,
                ChatMessageRole.System => ChatRole.System,
                _ => ChatRole.User
            };

        return new AIChatMessage(role, content);
    }
} 
```

stephentoub:
> I just wanted to thank you all so much for this update!

You're welcome :)

> I am curious to know if I am using, the OpenAIAssistantChatClient : IChatClient correctly?

It should be usable like any other IChatClient. The main thing to ensure, which it looks like you are, is that you're providing the thread ID as the ConversationId, as outlined in https://learn.microsoft.com/en-us/dotnet/ai/microsoft-extensions-ai#stateless-vs-stateful-clients.

PR #6505, opened 2025-06-09
Add ChatOptions.Instructions

stephentoub:
More services (especially those focused on agents) are starting to support the notion of per-request instructions, effectively system messages that aren't stored as part of a persisted chat history. For existing stateless services that don't have their own notion of separate instructions, we can just translate instructions into a system message.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
@@ -15,6 +15,9 @@ public class ChatOptions
     /// <related type="Article" href="https://learn.microsoft.com/dotnet/ai/microsoft-extensions-ai#stateless-vs-stateful-clients">Stateless vs. stateful clients.</related>
     public string? ConversationId { get; set; }
 
+    /// <summary>Gets or sets additional per-request instructions to be provided to the <see cref="IChatClient"/>.</summary>
+    public string? Instructions { get; set; }
+
     /// <summary>Gets or sets the temperature for generating chat responses.</summary>
     /// <remarks>
     /// This value controls the randomness of predictions made by the model. Use a lower value to decrease randomness in the response.
@@ -146,20 +149,21 @@ public class ChatOptions
     {
         ChatOptions options = new()
         {
-            ConversationId = ConversationId,
-            Temperature = Temperature,
-            MaxOutputTokens = MaxOutputTokens,
-            TopP = TopP,
-            TopK = TopK,
-            FrequencyPenalty = FrequencyPenalty,
-            PresencePenalty = PresencePenalty,
-            Seed = Seed,
-            ResponseFormat = ResponseFormat,
-            ModelId = ModelId,
-            AllowMultipleToolCalls = AllowMultipleToolCalls,
-            ToolMode = ToolMode,
-            RawRepresentationFactory = RawRepresentationFactory,
             AdditionalProperties = AdditionalProperties?.Clone(),
+            AllowMultipleToolCalls = AllowMultipleToolCalls,
+            ConversationId = ConversationId,
+            FrequencyPenalty = FrequencyPenalty,
+            Instructions = Instructions,
+            MaxOutputTokens = MaxOutputTokens,
+            ModelId = ModelId,
+            PresencePenalty = PresencePenalty,
+            RawRepresentationFactory = RawRepresentationFactory,
+            ResponseFormat = ResponseFormat,
+            Seed = Seed,
+            Temperature = Temperature,
+            ToolMode = ToolMode,
+            TopK = TopK,
+            TopP = TopP,
         };
 
         if (StopSequences is not null)
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -80,7 +80,7 @@ internal sealed partial class OpenAIChatClient : IChatClient
     {
         _ = Throw.IfNull(messages);
 
-        var openAIChatMessages = ToOpenAIChatMessages(messages, AIJsonUtilities.DefaultOptions);
+        var openAIChatMessages = ToOpenAIChatMessages(messages, options, AIJsonUtilities.DefaultOptions);
         var openAIOptions = ToOpenAIOptions(options);
 
         // Make the call to OpenAI.
@@ -95,7 +95,7 @@ internal sealed partial class OpenAIChatClient : IChatClient
     {
         _ = Throw.IfNull(messages);
 
-        var openAIChatMessages = ToOpenAIChatMessages(messages, AIJsonUtilities.DefaultOptions);
+        var openAIChatMessages = ToOpenAIChatMessages(messages, options, AIJsonUtilities.DefaultOptions);
         var openAIOptions = ToOpenAIOptions(options);
 
         // Make the call to OpenAI.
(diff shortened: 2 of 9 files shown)


stephentoub:
cc: @RogerBarreto, @westey-m, @markwallace-microsoft

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -15,6 +15,9 @@ public class ChatOptions
     /// <related type="Article" href="https://learn.microsoft.com/dotnet/ai/microsoft-extensions-ai#stateless-vs-stateful-clients">Stateless vs. stateful clients.</related>
     public string? ConversationId { get; set; }
 
+    /// <summary>Gets or sets additional per-request instructions to be provided to the <see cref="IChatClient"/>.</summary>
+    public string? Instructions { get; set; }
```
westey-m, line 19: Should we differentiate this name further if we want to also support `OverrideInstructions`?

I suppose Override and Additional are a bit strained outside of an Agentic scenario, so it's tricky.  It's not like you can override instructions for a regular model unless you deleted earlier system messages in the supplied chat history.

I did a quick survey of what different APIs have named additional instructions and override instructions anyway:
[OpenAIAssistant](https://platform.openai.com/docs/api-reference/runs/createRun): additional_instructions, instructions
[AzureAIAgents .net SDK](https://learn.microsoft.com/en-us/dotnet/api/azure.ai.agents.persistent.threadruns.createrunasync?view=azure-dotnet#azure-ai-agents-persistent-threadruns-createrunasync(system-string-system-string-system-string-system-string-system-string-system-collections-generic-ienumerable((azure-ai-agents-persistent-threadmessageoptions))-system-collections-generic-ienumerable((azure-ai-agents-persistent-tooldefinition))-system-nullable((system-boolean))-system-nullable((system-single))-system-nullable((system-single))-system-nullable((system-int32))-system-nullable((system-int32))-azure-ai-agents-persistent-truncation-system-binarydata-system-binarydata-system-nullable((system-boolean))-system-collections-generic-ireadonlydictionary((system-string-system-string))-system-collections-generic-ienumerable((azure-ai-agents-persistent-runadditionalfieldlist))-system-threading-cancellationtoken)): additionalInstructions, overrideInstructions
Bedrock Agents: [PromptSessionAttributes](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-session-state.html#session-state-attributes), [PromptOverrideConfiguration](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_PromptOverrideConfiguration.html)

stephentoub: My hesitancy here is that's baking the concept of "the thing I'm talking to already has instructions" into ChatOptions, and it's not clear to me how that can be supported generally. For example, if I use responses and my server-side chat history contains a system message, there is no way to implement instructions override as far as I can tell. And unlike other options, just ignoring it is problematic.

So I don't think we should add OverrideInstructions now. Is your request to just rename Instructions to AdditionalInstructions? I don't personally see much of a difference between Instructions and AdditionalInstructions, and if we decide in the future we need override and the lack of fidelity is acceptable, I would be ok in the future if we had both Instructions and OverrideInstructions.

Thoughts?

westey-m: Yeah, makes sense.

rogerbarreto: Should we bring this property as `Experimental` for now as we evolve the experimenting?

stephentoub: > Should we bring this property as `Experimental` for now as we evolve the experimenting?

Can you elaborate on why?

[Experimental] doesn't work well on an API like this. It effectively means that any stable libraries can't depend on it, yet they would need to reference it to correctly handle the user's provided options.

I don't see a lot of risk in this API. Just looking across various APIs:
- OpenAI Responses has per-request Instructions in options. Instructions => Instructions.
- OpenAI Chat Completion has any number of system / developer messages that aren't persisted. Instructions => a system message.
- OpenAI Assistants has per-request AdditionalInstructions in options. Instructions => AdditionalInstructions.
- Ollama has any number of system / developer messages that aren't persisted. Instructions => a system message.
- Azure.AI.Inference has any number of system / developer messages that aren't persisted. Instructions => a system message.
- AWS Bedrock's Runtime APIs support a collection of any number of system messages. Instructions => a system message.
- Anthropic's Messaging API supports a top-level system parameter. Instructions => added to that system parameter (along with other system messages)
- Google's Gemini API supports a system_instruction. Instructions => added to that system_instructions (along with other system messages)
- Mistral's API supports system messages that aren't persisted. Instructions => a system message.
- ...

If it's about the naming, in general the APIs already on ChatOptions that provide collections are additive, not overrides. Stop sequences doesn't change what the service already pays attention to, it just adds more. Tools doesn't change any tools built-in to the service, it just adds more. Etc. Instructions would be the same. If we ever did want an override (and I'm not convinced we will, especially since as noted I'm not sure how feasible it is), it'd be fine I think to have OverrideInstructions.

RogerBarreto (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
rogerbarreto: > I don't see a lot of risk in this API. Just looking across various APIs:

Ok, I brought this more in case we think on expanding the `Instructions` to be multi modal capable, possibly a `List<AIContent>`'s, starting with `string` seems quite limited, but if the ecosystem is all Text-Only, seems reasonable to go GA with it.

stephentoub: I see, thanks. Everyone right now is text (as far as I can tell, even in system/developer messages that are part of otherwise multimodal message lists). I suspect it'll be that way for a while. But if you were ever able to provide other kinds of content for instructions, we could add an additional InstructionsContent property, or something like that.

markwallace-microsoft (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -111,11 +111,16 @@ void IDisposable.Dispose()
...
     /// <summary>Converts an Extensions chat message enumerable to an OpenAI chat message enumerable.</summary>
-    private static IEnumerable<OpenAI.Chat.ChatMessage> ToOpenAIChatMessages(IEnumerable<ChatMessage> inputs, JsonSerializerOptions options)
+    private static IEnumerable<OpenAI.Chat.ChatMessage> ToOpenAIChatMessages(IEnumerable<ChatMessage> inputs, ChatOptions? chatOptions, JsonSerializerOptions jsonOptions)
     {
         // Maps all of the M.E.AI types to the corresponding OpenAI types.
         // Unrecognized or non-processable content is ignored.
 
+        if (chatOptions?.Instructions is { } instructions && !string.IsNullOrWhiteSpace(instructions))
+        {
+            yield return new SystemChatMessage(instructions);
```
markwallace-microsoft, line 121: Should the role of this message be configurable? Would a developer message be more appropriate given that is where these instructions come from?

stephentoub: https://community.openai.com/t/how-is-developer-message-better-than-system-prompt/1062784/2

My understanding then is that there's no good reason to specify developer over system, because openai will translate whichever you provide to the one and only one that model understands. Which means from an api point of view introducing developer was pointless beyond creating an alias because they didn't like the "system" name for this purpose.

And if you specify developer and you're talking to something else, it will likely break. So just always using system is correct.

PR #6512, opened 2025-06-11
Workaround OpenAI assistant's RunCreationOption's tools override

stephentoub:
In OpenAIAssistantChatClient, as part of providing ChatOptions.Tools to the service, the only API available to do so is ToolsOverride, which ends up replacing rather than augmenting any tools defined at the assistant level. To work around that, this changes the implementation to retrieve the assistant's tools and send them as part of the override, effectively implementing a merge rather than override.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6512)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
@@ -44,6 +44,9 @@ internal sealed partial class OpenAIAssistantChatClient : IChatClient
     /// <summary>The thread ID to use if none is supplied in <see cref="ChatOptions.ConversationId"/>.</summary>
     private readonly string? _defaultThreadId;
 
+    /// <summary>List of tools associated with the assistant.</summary>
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
```diff
@@ -290,7 +311,6 @@ void IDisposable.Dispose()
                         runOptions.ToolConstraint = ToolConstraint.None;
                         break;
 
-                    case null:
```
eiriktsarpalis, line 293: Is this related to the core change?

stephentoub: Tangentially. I was validating what other areas we were overwriting a default with a different default. If this is null, then the user hasn't explicitly chosen a setting, so we should just leave it to whatever the service's default is.

PR #6514, opened 2025-06-11
Introduce evaluators for agentic workflows

### LINKED ISSUE #6350: [AI Evaluation] Introduce IntentResolution, ToolCallAcuracy and TaskAdherence evaluators for agentic workflows
Corresponding evaluators are available in the Azure AI Python SDK. These evaluators need tool definitions and tool calls to be included as part of the evaluation prompts (in a JSON format). We should be able to leverage the existing abstractions in `Microsoft.Extensions.AI.Abstractions` such as `AIFunction` and `FunctionCallContent` for this.

### LINKED ISSUE #6370: [AI Evaluation] Tool messages (or anything without 'Text' contents) are not rendered in reports
In Microsoft.Extensions.AI.Evaluations.Reporting, the report created does not render tool calls with FunctionCallContents or FunctionResponseContents because the content type is not 'text'. I've worked around this by using the messages with FunctionCallContents to get the response, and changing the tool messages contents for the EvaluateAsync method, but it was frustrating to debug the library to figure this out.

shyamnamboodiripad:
Introduces the following new evaluators as part of the Quality package: `ToolCallAccuracyEvaluator`, `TaskRelevanceEvaluator`

TODO: 
- [ ] Introduce `IntentResolutionEvaluator` - will add this before the PR is merged

Also includes following changes:
* Fixes a regex bug that was causing reasoning and chain of thought outputs present in the evaluation response to not be parsed correctly into the corresponding metrics.
* Adds support for displaying tool calls and tool results in the conversation displayed in the report. This fixes #6370
* Adds support for displaying JSON content both in the conversation as well as in context (along with a new settings toggle for controlling pretty printing for the displayed JSON).
* Adds tests for the new evaluators.

Fixes #6350
Fixes #6370


src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/ChatResponseExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+                string toolCallOrResultJsonString =
+                    JsonSerializer.Serialize(
+                        content,
+                        AIJsonUtilities.DefaultOptions.GetTypeInfo(contentType));
+
+                JsonNode? toolCallOrResultJsonNode = JsonNode.Parse(toolCallOrResultJsonString);
+                if (toolCallOrResultJsonNode is not null)
+                {
+                    toolCallsAndResultsJsonArray.Add(toolCallOrResultJsonNode);
+                }
```
peterwald, line 44: What is the purpose of re-parsing the serialized JSON?

shyamnamboodiripad: Perhaps there is a better way to do this that I missed - but I am adding the serialized JSON to a `JsonArray`, and I could not figure out a way to do this without parsing it into a `JsonNode`.

peterwald: I guess that's the only route to get from a .NET object to a JsonNode?

peterwald: I believe [JsonSerializer.SerializeToNode](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.serializetonode?view=net-9.0) will do it for you.

shyamnamboodiripad: Looks like there is a [`JsonSerializer.SerializeToNode()`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.serializetonode?view=net-9.0) API that I could use to combine the two steps into one. I'll look into this.

PR #6519, opened 2025-06-16
CompletenessEvaluator: Remove Seed parameter from ChatOptions

planetis-m:
Azure AI does not accept that parameter, which causes this evaluator to error out.

```
Diagnostics
Severity
Message
Error
Azure.RequestFailedException: {"detail":[{"type":"extra_forbidden","loc":["body","seed"],"msg":"Extra inputs are not permitted","input":123,"url":"https://errors.pydantic.dev/2.10/v/extra_forbidden"}]}
Status: 422 (Unprocessable Entity)
ErrorCode: Invalid input

Content:
{"error":{"code":"Invalid input","message":"{\"detail\":[{\"type\":\"extra_forbidden\",\"loc\":[\"body\",\"seed\"],\"msg\":\"Extra inputs are not permitted\",\"input\":123,\"url\":\"https://errors.pydantic.dev/2.10/v/extra_forbidden\"}]}","status":422}}
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6519)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/CompletenessEvaluator.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/CompletenessEvaluator.cs
@@ -54,7 +54,6 @@ public sealed class CompletenessEvaluator : IEvaluator
             Temperature = 0.0f,
             MaxOutputTokens = 800,
             TopP = 1.0f,
-            Seed = 123,
             PresencePenalty = 0.0f,
             FrequencyPenalty = 0.0f,
             ResponseFormat = ChatResponseFormat.Text


peterwald:
@planetis-m Can you say which model version you are using that doesn't support `seed`?

planetis-m:
I can't reproduce anymore with gpt4.1. Must have been one of the low rate limit tier LLMs I used for testing. Can't remember really.

shyamnamboodiripad:
@planetis-m we generally want to keep the evaluators in this repo aligned with the ones in the Azure Python Evaluation SDK. The seed setting used here matches the one that is used in the [corresponding python evaluator](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/evaluation/azure-ai-evaluation/azure/ai/evaluation/_evaluators/_response_completeness/response_completeness.prompty#L10).

I would be surprised if this doesn't work for the [models that are recommended for use with these evaluators (such as gpt-4o and newer)](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/CompletenessEvaluator.cs#L37). Please feel free to ping back / log an issue if you happen to run into this again. Thanks!

--

Also, it was my understanding that `ChatOptions` should generally be portable across models / model providers and that model / provider specific `IChatClient`s would simply ignore options (like `Seed`) that they don't support (yet). However, specific clients may have reasons to diverge on this. Tagging @stephentoub and @eiriktsarpalis in case this is an incorrect assumption / in case you have any additional guidance around this.

PR #6521, opened 2025-06-16
Log and redact query string params

rainsxng:
This feature allows to log and redact request query string parameters
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6521)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
+++ b/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
@@ -14,6 +14,7 @@ using Microsoft.Extensions.Http.Diagnostics;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Telemetry.Internal;
 using Microsoft.Shared.Diagnostics;
+using Microsoft.Shared.Pools;
 
 namespace Microsoft.Extensions.Http.Logging.Internal;
 
@@ -156,57 +157,64 @@ internal sealed class HttpRequestReader : IHttpRequestReader
 
     private string ExtractAndRedactQueryParameters(string query)
     {
-        var sb = new System.Text.StringBuilder();
-        ReadOnlySpan<char> querySpan = query.AsSpan();
-        int length = querySpan.Length;
-        int start = 0;
-
-        // Remove leading '?'
-        if (length > 0 && querySpan[0] == '?')
+        var stringBuilder = PoolFactory.SharedStringBuilderPool.Get();
+        try
         {
-            start = 1;
-        }
+            ReadOnlySpan<char> querySpan = query.AsSpan();
+            int length = querySpan.Length;
+            int start = 0;
 
-        while (start < length)
-        {
-            int amp = querySpan.Slice(start).IndexOf('&');
-            int end = amp == -1 ? length : start + amp;
-
-            int eq = querySpan.Slice(start, end - start).IndexOf('=');
-            if (eq >= 0)
+            // Remove leading '?'
+            if (length > 0 && querySpan[0] == '?')
             {
-                var keySpan = querySpan.Slice(start, eq);
-                var valueSpan = querySpan.Slice(start + eq + 1, end - (start + eq + 1));
+                start = 1;
+            }
 
-                string key = UnescapeDataString(keySpan);
-                string value = UnescapeDataString(valueSpan);
+            while (start < length)
+            {
+                int amp = querySpan.Slice(start).IndexOf('&');
+                int end = amp == -1 ? length : start + amp;
 
-                // Only process if the key is in the classification dictionary and value is not empty
-                if (!string.IsNullOrEmpty(value) && _queryParameterDataClasses.TryGetValue(key, out var classification))
+                int eq = querySpan.Slice(start, end - start).IndexOf('=');
+                if (eq >= 0)
                 {
-                    string redacted = _httpHeadersReader.RedactValue(value, classification);
+                    var keySpan = querySpan.Slice(start, eq);
+                    var valueSpan = querySpan.Slice(start + eq + 1, end - (start + eq + 1));
 
-                    // Append to string builder directly with proper encoding
-                    if (sb.Length > 0)
+                    string key = UnescapeDataString(keySpan);
+                    string value = UnescapeDataString(valueSpan);
+
+                    // Only process if the key is in the classification dictionary and value is not empty
+                    if (!string.IsNullOrEmpty(value) && _queryParameterDataClasses.TryGetValue(key, out var classification))
                     {
-                        _ = sb.Append('&');
+                        string redacted = _httpHeadersReader.RedactValue(value, classification);
+
+                        // Append to string builder directly with proper encoding
+                        if (stringBuilder.Length > 0)
+                        {
+                            _ = stringBuilder.Append('&');
+                        }
+
+                        _ = stringBuilder.Append(Uri.EscapeDataString(key))
+                            .Append('=')
+                            .Append(Uri.EscapeDataString(redacted));
                     }
-
-                    _ = sb.Append(Uri.EscapeDataString(key))
-                        .Append('=')
-                        .Append(Uri.EscapeDataString(redacted));
                 }
+
+                if (amp == -1)
+                {
+                    break;
+                }
+
+                start = end + 1;
             }
 
-            if (amp == -1)
-            {
-                break;
-            }
-
-            start = end + 1;
+            return stringBuilder.ToString();
         }
-
-        return sb.ToString();
+        finally
+        {
+            PoolFactory.SharedStringBuilderPool.Return(stringBuilder);
+        }
     }
 
     private void GetRedactedPathAndParameters(HttpRequestMessage request, LogRecord logRecord)


src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -86,15 +99,34 @@ internal HttpRequestReader(
 
         _logRequestHeaders = options.RequestHeadersDataClasses.Count > 0;
         _logResponseHeaders = options.ResponseHeadersDataClasses.Count > 0;
+        _logRequestQueryParameters = options.LogQueryParameters;
```
dariusclay: Instead of a boolean flag we could check if `options.RequestQueryParametersDataClasses.Count > 0`, this keeps it consistent with how enabling logging headers works.

rainsxng: I had it like that originally, but it was interfering in one of the use cases. I'll try that again to see whether it can be done that way

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
```diff
@@ -235,6 +236,17 @@ private void FillLogRecord(
             }
         }
 
+        if (logRecord.QueryParameters is { Length: > 0 })
+        {
+            if (loggerMessageState is IEnrichmentTagCollector collector)
+            {
+                foreach (var param in logRecord.QueryParameters)
+                {
+                    collector.Add($"Query_{param.Key}", param.Value ?? string.Empty);
```
dariusclay: I understand wanting to avoid ambiguity, but is the prefix really needed?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,26 +143,59 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
         }
 
-        logRecord.StatusCode = (int)response.StatusCode;
-    }
+        var parsed = HttpUtility.ParseQueryString(query);
+        var result = new List<KeyValuePair<string, string?>>();
+
+        foreach (var kvp in _queryParameterDataClasses)
+        {
+            var value = parsed[kvp.Key];
```
dariusclay: Couldn't this throw a KeyNotFoundException if the configured query parameter doesn't actually exist?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,26 +143,59 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+    {
+        if (string.IsNullOrEmpty(query) || _queryParameterDataClasses.Count == 0)
         {
-            logRecord.ResponseBody = await _httpResponseBodyReader.ReadAsync(response, cancellationToken).ConfigureAwait(false);
+            return ([], 0);
         }
 
-        logRecord.StatusCode = (int)response.StatusCode;
-    }
+        var parsed = HttpUtility.ParseQueryString(query);
```
dariusclay: Can we include support for NetFW ? Possibly just parsing the query string ourselves, it's not that difficult.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: No, it won't since it should return when key doesn't exist. I check it later with `string.IsNullOrEmpty(value)`, so we iterate only through existing keys

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done, but I combined both of them. Please let me know what you think of it

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: I'm afraid that it would introduce a lot of the overhead there. But we can to add it there

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
dariusclay: IMO I think just adding some query parameters to the dictionary "enables" it and removing them would disable it. This is consistent with how other configurations are done here.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
dariusclay: Where is the overhead exactly? We parse the route segments already.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
dariusclay: I still think this can throw. Let's check if the key exists before trying to get the value out of the dictionary. In .NET FW it would be done with ContainsKey + Lookup. In .NET Core there is a "TryGetValue" method that is optimized.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
```diff
@@ -75,6 +75,16 @@ internal sealed class LogRecord : IResettable
     /// </summary>
     public int PathParametersCount { get; set; }
 
+    /// <summary>
+    /// Gets or sets request query string parameters.
+    /// </summary>
+    public KeyValuePair<string, string?>[]? QueryParameters { get; set; } = [];
```
evgenyfedorov2:
```suggestion
    public KeyValuePair<string, string?>[]? QueryParameters { get; set; };
```

Why does it need a non-null value by default when it is nullable anyway?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
```diff
@@ -75,6 +75,16 @@ internal sealed class LogRecord : IResettable
...
+
+    /// <summary>
+    /// Gets or sets request query string parameters count.
+    /// </summary>
+    public int QueryParametersCount { get; set; }
```
evgenyfedorov2: Do we really need this property?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
```diff
@@ -75,6 +75,16 @@ internal sealed class LogRecord : IResettable
...
+
     public bool TryReset()
```
evgenyfedorov2, line 84: Don't two new properties need to be reset in this method?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,99 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+            return ExtractAndRedactQueryParameters(request.RequestUri.Query);
         }
 
-        if (_logResponseBody)
+        return ([], 0);
+    }
+
+    private (KeyValuePair<string, string?>[] queryParameters, int count) ExtractAndRedactQueryParameters(string query)
+    {
+        if (string.IsNullOrEmpty(query) || _queryParameterDataClasses.Count == 0)
```
evgenyfedorov2: we have already checked _logRequestQueryParameters on line 144 and we know that  RequestQueryParametersDataClasses > 0, why do we need to check it again with _queryParameterDataClasses.Count == 0 ?

evgenyfedorov2: can this condition be merged with the one on line 144?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,99 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+                var keySpan = querySpan.Slice(start, end - start);
+                key = Uri.UnescapeDataString(keySpan.ToString());
+                value = null;
+            }
+
+            // Only add the first occurrence of a key
+            if (!dict.ContainsKey(key))
+            {
+                dict[key] = value;
+            }
```
evgenyfedorov2: can we use `.TryAdd()` instead?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,99 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+                var keySpan = querySpan.Slice(start, eq);
+                var valueSpan = querySpan.Slice(start + eq + 1, end - (start + eq + 1));
+                key = Uri.UnescapeDataString(keySpan.ToString());
+                value = Uri.UnescapeDataString(valueSpan.ToString());
+            }
+            else
+            {
+                var keySpan = querySpan.Slice(start, end - start);
+                key = Uri.UnescapeDataString(keySpan.ToString());
+                value = null;
```
evgenyfedorov2: why do we add `null` value to the dictionary only to discard it later on line 210?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,99 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+        }
+
+        var result = new List<KeyValuePair<string, string?>>();
+
+        foreach (var kvp in _queryParameterDataClasses)
+        {
+            if (dict.TryGetValue(kvp.Key, out var value) && !string.IsNullOrEmpty(value))
+            {
+                string redacted;
+                if (_httpHeadersReader is HttpHeadersReader realReader)
```
evgenyfedorov2: please help me understand why we need to do this type checking in the foreach loop?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
```diff
@@ -235,6 +236,17 @@ private void FillLogRecord(
             }
         }
 
+        if (logRecord.QueryParameters is { Length: > 0 })
+        {
+            if (loggerMessageState is IEnrichmentTagCollector collector)
```
evgenyfedorov2: why do we omit cases when `loggerMessageState` is not of `IEnrichmentTagCollector` type?

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
```diff
@@ -573,6 +581,226 @@ public async Task ReadAsync_MetadataWithoutRequestRouteOrNameUsesConstants_Retur
         actualRecord.Should().BeEquivalentTo(expectedRecord);
     }
 
+    [Fact]
+    public async Task ReadAsync_LogsQueryParameters_WhenEnabled()
```
evgenyfedorov2: This and other tests below do not actually log anything. They just test if LogRecord contains query parameters.

1. I think they need to be renamed to reflect reality.
2. Also, please add proper acceptance tests

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,99 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+
+            int eq = querySpan.Slice(start, end - start).IndexOf('=');
+            string key;
+            string? value;
+            if (eq >= 0)
+            {
+                var keySpan = querySpan.Slice(start, eq);
+                var valueSpan = querySpan.Slice(start + eq + 1, end - (start + eq + 1));
+                key = Uri.UnescapeDataString(keySpan.ToString());
+                value = Uri.UnescapeDataString(valueSpan.ToString());
```
evgenyfedorov2: why .ToString(), given that there is an overload for ReadOnlySpan https://learn.microsoft.com/en-us/dotnet/api/system.uri.unescapedatastring?view=net-9.0#system-uri-unescapedatastring(system-readonlyspan((system-char)))

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,99 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+            }
+            else
+            {
+                var keySpan = querySpan.Slice(start, end - start);
+                key = Uri.UnescapeDataString(keySpan.ToString());
```
evgenyfedorov2: use the ReadOnlySpan overload https://learn.microsoft.com/en-us/dotnet/api/system.uri.unescapedatastring?view=net-9.0#system-uri-unescapedatastring(system-readonlyspan((system-char)))

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
```diff
@@ -235,6 +236,17 @@ private void FillLogRecord(
...
+            {
+                foreach (var param in logRecord.QueryParameters)
+                {
+                    collector.Add(param.Key, param.Value ?? string.Empty);
```
evgenyfedorov2: why don't just use classified tags https://github.com/dotnet/extensions/blob/48f95ca63e0b02aba43eb99893d50c10d65ccc57/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LoggerMessageState.TagCollector.cs#L20 ?

evgenyfedorov2 (changes requested):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
rainsxng: Removed that property

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: Essentially, it was made to be in check with the headers enrichment. But should we try to enrich anything else that doesn't implement this interface?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: This part is unified for .net 462 and net core, so the latest doesn't have this method and therefore cannot be used. I can try to add a conditional compilation there if needed. WDYT?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: To simplify the parsing logic. We basically add all the keys and then just filter ones without a value. We can stop doing that and add just non-null values instead

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Doesn't have to be in the foreach loop. Added a new method to the interface so we can omit this check

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
evgenyfedorov2: Yes, let's use the conditional compilation

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
evgenyfedorov2: hmm, unless I am missing something, it does not make sense? If we stop adding nulls, then on line 205 `dict.TryGetValue()` will be just returning `false` for those cases where it was previously returning null values and we won't need to check `string.IsNullOrEmpty` on the same line, right?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: Also, had to use it so I can access the correct Add() overload, but it checks for the ITagCollector interface instead. Please let me know what you think of it

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -49,8 +51,14 @@ public static class HttpClientLoggingTagNames
     /// <summary>
     /// HTTP Response Headers prefix.
     /// </summary>
+    [Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
```
evgenyfedorov2: It should not be marked experimental?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
```diff
@@ -36,6 +36,22 @@ public class LoggingOptions
     /// </remarks>
     public bool LogRequestStart { get; set; }
 
+    /// <summary>
+    /// Gets or sets the set of HTTP request query parameters to log and their respective data classifications to use for redaction.
+    /// </summary>
+    /// <value>
+    /// The default value is <see cref="HashSet{T}"/>.
```
evgenyfedorov2: Default should be some Dictionary, not HashSet, right?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -86,15 +89,34 @@ internal HttpRequestReader(
...
+        {
+            logRecord.ResponseBody = await _httpResponseBodyReader.ReadAsync(response, cancellationToken).ConfigureAwait(false);
+        }
+
+        logRecord.StatusCode = (int)response.StatusCode;
+    }
+
     public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage request,
-        List<KeyValuePair<string, string>>? requestHeadersBuffer, CancellationToken cancellationToken)
+    List<KeyValuePair<string, string>>? requestHeadersBuffer, CancellationToken cancellationToken)
```
evgenyfedorov2:
```suggestion
        List<KeyValuePair<string, string>>? requestHeadersBuffer, CancellationToken cancellationToken)
```
formatting is broken

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,94 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+#else
+        return Uri.UnescapeDataString(value.ToString());
+#endif
+    }
+
+    private List<KeyValuePair<string, string>> GetQueryParameters(HttpRequestMessage request)
+    {
+        if (_logRequestQueryParameters &&
+            request.RequestUri is not null &&
+            !string.IsNullOrEmpty(request.RequestUri.Query))
```
evgenyfedorov2:
```suggestion
        if (_logRequestQueryParameters &&
            !string.IsNullOrEmpty(request.RequestUri?.Query))
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
```diff
@@ -75,6 +75,11 @@ internal sealed class LogRecord : IResettable
     /// </summary>
     public int PathParametersCount { get; set; }
 
+    /// <summary>
+    /// Gets or sets request query string parameters.
+    /// </summary>
+    public List<KeyValuePair<string, string>>? QueryParameters { get; set; }
```
evgenyfedorov2: can this be array instead of list?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,94 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
         {
-            _httpHeadersReader.ReadResponseHeaders(response, responseHeadersBuffer);
-            logRecord.ResponseHeaders = responseHeadersBuffer;
+            return ExtractAndRedactQueryParameters(request.RequestUri.Query);
         }
 
-        if (_logResponseBody)
+        return new List<KeyValuePair<string, string>>();
```
evgenyfedorov2: to minimize number of allocations we should return empty list:


```suggestion
        return Array.Empty<KeyValuePair<string, string>>();
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
```diff
@@ -235,6 +238,24 @@ private void FillLogRecord(
...
 
+        if (logRecord.QueryParameters is { Count: > 0 })
+        {
+            if (loggerMessageState is ITagCollector collector)
+            {
+                foreach (var param in logRecord.QueryParameters)
+                {
+                    if (_options.RequestQueryParametersDataClasses.TryGetValue(param.Key, out var classification))
+                    {
+                        collector.Add(param.Key, param.Value ?? string.Empty, new DataClassificationSet(classification));
```
evgenyfedorov2: We know that the value is not null or empty from the `ExtractAndRedactQueryParameters` method, probably we should do just param.Value ?


```suggestion
                        collector.Add(param.Key, param.Value, new DataClassificationSet(classification));
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
```diff
@@ -235,6 +238,24 @@ private void FillLogRecord(
...
+                    }
+                    else
+                    {
+                        collector.Add(param.Key, param.Value ?? string.Empty);
+                    }
```
evgenyfedorov2: do I understand it right, that 

- if data classification is None, we log the value as is
- if there is any other data classification, we redact the value accordingly
- if there is no data classification, we don't log the query parameter
?

If it is correct, we don't need the else {} block here

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
evgenyfedorov2: Also, I am not exactly sure if we need to use the overload with data classification? If a user uses ExtendedLogger, won't it cause double redaction, because ExtendedLogger redacts classified tags?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
evgenyfedorov2: Ok, I see, LoggerMessageState explicitly implements ITagCollector, therefore we must use this type check

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +133,94 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+
+                // Only add if value is not null or empty
+                if (!string.IsNullOrEmpty(value))
+                {
+#if NET8_0_OR_GREATER
+                    dict.TryAdd(key, value);
+#else
+                    if (!dict.ContainsKey(key))
+                    {
+                        dict[key] = value;
```
evgenyfedorov2: We add key/value pairs to the `dict` only to throw it away when this method returns, so it is not terribly useful, but causing trouble in the form of extra memory allocations.

Can we somehow minimize the number of memory allocations? One way to achieve that is check if each query parameter key exists in `_queryParameterDataClasses` right here in place, and if so, redact it also right here. 

Then, for the `result` list, please use list pooling, one example of it is [here](https://github.com/dotnet/extensions/blob/fd810cf900e4be4c3a35bd96385dcc87a7c92927/src/Libraries/Microsoft.Extensions.Telemetry/Buffering/GlobalBuffer.cs#L23), but there are more in this repo, of course

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
evgenyfedorov2: Actually there is `AcceptanceTests` file in this project, so please use it - such tests should be able to test the feature end to end

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Pipeline fails as it forces me to mark it experimental

iliar-turdushev: I think the pipeline meant that the `QueryParametersPrefix` field had to be marked as `Experimental`.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -49,8 +51,14 @@ public static class HttpClientLoggingTagNames
...
     public const string ResponseHeaderPrefix = "http.response.header.";
 
+    /// <summary>
+    /// HTTP query parameters prefix.
+    /// </summary>
+    public const string QueryParametersPrefix = "http.query.";
```
iliar-turdushev: There is no such attribute in [OTel's semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/0c4e76a5a604b23de6f0b3dae17b0e8d164b09db/docs/http/http-spans.md#http-client-span). According to OTel, we can have the `url.full` attribute that includes a query string.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
```diff
@@ -52,6 +52,25 @@ public static void AddResponseHeaders(this LoggerMessageState state, List<KeyVal
...
+    /// </summary>
+    /// <param name="state">A <see cref="LoggerMessageState"/> to be filled.</param>
+    /// <param name="items">A list with query parameters.</param>
+    /// <param name="index">Represents an index to be used when writing tags into <paramref name="state"/>.</param>
+    /// <remarks><paramref name="index"/> will be mutated to point to the next <paramref name="state"/> item.</remarks>
+    public static void AddQueryParameters(this LoggerMessageState state, List<KeyValuePair<string, string>> items, ref int index)
+    {
+        foreach (var t in items)
+        {
+            var key = _responsePrefixedNamesCache.GetOrAdd(
```
iliar-turdushev: I think we should use a separate cache for prefixed query parameter names. Otherwise, we risk to end up emitting a query parameter with a header name prefix, or vice versa.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
```diff
@@ -136,13 +137,17 @@ private static void OutgoingRequest(
             loggerMessageState.TagArray[index] = new(HttpClientLoggingTagNames.ResponseBody, record.ResponseBody);
         }
 
+        if (queryParametersCount > 0)
```
iliar-turdushev: Why are we emitting each query parameter separately instead of using [OTel's recommendation to emit a query string as part of the `url.full` attribute](https://github.com/open-telemetry/semantic-conventions/blob/0c4e76a5a604b23de6f0b3dae17b0e8d164b09db/docs/http/http-spans.md)?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
rainsxng: Done, switched it to the array

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: But shouldn't we use it as per this comment - https://github.com/dotnet/extensions/pull/6521#discussion_r2189760282? I'm fine with removing this overload and using the regular one instead (just with the key and a value)

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Good point, I'll look into it

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
rainsxng: Makes sense, I'll update it to correspond to the otel semantics

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/Log.cs
rainsxng: Done, but I'm not sure about the prefixes, would we really need them here?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Done

evgenyfedorov2: Is this used now or not? I don't see where it is used

dariusclay: Shouldn't QueryParametersPrefix be removed now, and `url.full` added as tag name?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
```diff
@@ -52,6 +53,23 @@ public static void AddResponseHeaders(this LoggerMessageState state, List<KeyVal
...
+    /// <param name="fullUri">The full URI to log with redacted query parameters.</param>
+    /// <param name="index">
+    /// Represents an index to be used when writing tags into <paramref name="state"/>.
+    /// <para>This parameter will be mutated to point to the next <paramref name="state"/> item.</para>
+    /// </param>
+    public static void AddFullUrl(this LoggerMessageState state, Uri? fullUri, ref int index)
+    {
+        if (fullUri is not null)
+        {
+            state.TagArray[index++] = new("url.full", fullUri.ToString());
```
dariusclay: Update to use constant defined in HttpClientLoggingTagNames for `url.full`

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpHeadersReader.cs
```diff
@@ -70,6 +70,9 @@ public void ReadResponseHeaders(HttpResponseMessage response, List<KeyValuePair<
         }
     }
 
+    public string RedactValue(string value, DataClassification classification)
+        => _redactor.Redact([value], classification);
```
evgenyfedorov2: this might cause an unnecessary allocation of an array, right?
can we do just `_redactor.Redact(value, classification)` (and create the respective method for that)?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +141,101 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
                 .ConfigureAwait(false);
         }
+
+        logRecord.QueryParameters = GetQueryParameters(request);
+
+        // Build redacted query string
+        var redactedQueryString = string.Join("&", logRecord.QueryParameters.Select(kvp => $"{kvp.Key}={kvp.Value}"));
+
+        // Build full URI with redacted query string
+        if (request.RequestUri != null && logRecord.QueryParameters.Length > 0)
```
dariusclay: I think we should emit `url.full` regardless of whether log record contains query parameters, for consistency. From a customer perspective, it seems odd that it would appear suddenly for requests with query params vs. none.
What do others think?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +141,101 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+            return ExtractAndRedactQueryParameters(request.RequestUri!.Query);
         }
 
-        if (_logResponseBody)
+        return [];
+    }
+
+    private KeyValuePair<string, string>[] ExtractAndRedactQueryParameters(string query)
+    {
+        var result = _queryParamListPool.Get();
```
evgenyfedorov2: by the way, for this repo, we have to use explicit types for cases where the type is not obvious, and this is the case here

```suggestion
        List<KeyValuePair<string, string>> result = _queryParamListPool.Get();
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +141,101 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+                    var keySpan = querySpan.Slice(start, eq);
+                    var valueSpan = querySpan.Slice(start + eq + 1, end - (start + eq + 1));
+
+                    string key = UnescapeDataString(keySpan);
+                    string value = UnescapeDataString(valueSpan);
+
+                    // Only process if the key is in the classification dictionary and value is not empty
+                    if (!string.IsNullOrEmpty(value) && _queryParameterDataClasses.TryGetValue(key, out var classification))
+                    {
+                        string redacted = _httpHeadersReader != null
```
evgenyfedorov2: I don't see how can `_httpHeadersReader` be really `null`?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
evgenyfedorov2: should `url.full` be a `const string` of the `HttpClientLoggingTagNames` class?

evgenyfedorov2: I don't understand this part `fullUri.ToString()`, because we just built the `Uri` object in the previous method, why do we allocate and convert it back to a `string` again? 

I think a new string should be created with StringBuilder instead, don't need `new Uri()` and `ToString()`

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +141,101 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
             logRecord.RequestBody = await _httpRequestBodyReader.ReadAsync(request, cancellationToken)
                 .ConfigureAwait(false);
         }
+
+        logRecord.QueryParameters = GetQueryParameters(request);
+
+        // Build redacted query string
+        var redactedQueryString = string.Join("&", logRecord.QueryParameters.Select(kvp => $"{kvp.Key}={kvp.Value}"));
```
evgenyfedorov2: 1. We can't afford LINQ it is too expensive
2. If I see it right, we duplicate information because we log each key/value query parameter separately and then altogether in `url.full` ?

I think this should be logged only once, IMO as `url.full`

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: It's not used for now as I'm just not really sure that we need a prefix here. My current approach uses url.full as a tag and it writes a pre-redacted URL there

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
rainsxng: done

rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Removed the unnecessary check

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Not sure about that one, but theoretically we can

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LoggerMessageStateExtensions.cs
rainsxng: Removed the cache completely, don't think we really need it for a single tag. But I'm open to suggestions if we decide that we need it

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Updated the prefix names and this part in general, so should be good now

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -51,6 +53,12 @@ public static class HttpClientLoggingTagNames
     /// </summary>
     public const string ResponseHeaderPrefix = "http.response.header.";
 
+    /// <summary>
+    /// HTTP full URI prefix.
+    /// </summary>
+    [Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
+    public const string FullUriPrefix = "url.full";
```
dariusclay: Please name it "FullUri", it's not a prefix, so we are consistent with the intention this value has.

rainsxng: Done

dariusclay:
It would be good to see actual output difference between how it exists before, and then with this PR change.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -51,6 +53,12 @@ public static class HttpClientLoggingTagNames
     /// </summary>
     public const string ResponseHeaderPrefix = "http.response.header.";
 
+    /// <summary>
+    /// HTTP full URI prefix.
+    /// </summary>
+    [Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
+    public const string FullUri = "url.full";
```
iliar-turdushev:
```suggestion
    public const string FullUrl = "url.full";
```

Please, adjust the xml-comment accordingly.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +141,98 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+                _ = sb.Append('&');
+            }
+
+            _ = sb.Append(kvp.Key).Append('=').Append(kvp.Value);
+        }
+
+        var queryString = sb.ToString();
+
+        logRecord.FullUri = string.IsNullOrEmpty(queryString)
+            ? null
```
iliar-turdushev:
```suggestion
            ? $"{logRecord.Host}/{logRecord.Path}"
```

Shouldn't we include here a request path without a query string when the latter is null?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
iliar-turdushev: I initially suggested to include the `url.full` tag because it is part of the [OTel's semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span). But we already have `url.path` which includes the path part of an HTTP request. If we now include `url.full` then the path part will be duplicated in two tags. What if we consider including [the `url.query` tag](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span) instead of `url.full`?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
```diff
@@ -235,6 +236,17 @@ private void FillLogRecord(
             }
         }
 
+        if (logRecord.QueryParameters is { Length: > 0 })
+        {
+            if (loggerMessageState is IEnrichmentTagCollector collector)
+            {
+                foreach (var param in logRecord.QueryParameters)
+                {
+                    collector.Add(param.Key, param.Value);
```
iliar-turdushev: Do we need to log each query parameter separately if we log them as part of the `url.full` or `url.query` tag?

A query parameter could be met several times in a query string, e.g. `?value=1&value=2`. Would this cause an issue if we log each query parameter separately?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
iliar-turdushev: Can you, please, use `url` instead of `uri` everywhere in this PR?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
```diff
@@ -36,6 +36,22 @@ public class LoggingOptions
     /// </remarks>
     public bool LogRequestStart { get; set; }
 
+    /// <summary>
+    /// Gets or sets the set of HTTP request query parameters to log and their respective data classifications to use for redaction.
+    /// </summary>
+    /// <value>
+    /// The default value is <see cref="System.Collections.Generic.IDictionary{TKey, TValue}"/>.
```
iliar-turdushev:
```suggestion
    /// The default value is <see cref="System.Collections.Generic.Dictionary{TKey, TValue}"/>.
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
```diff
@@ -36,6 +36,22 @@ public class LoggingOptions
...
+    /// </value>
+    /// <remarks>
+    /// If empty, no HTTP request query parameters will be logged.
+    /// If the data class is <see cref="DataClassification.None"/>, no redaction will be done.
+    /// </remarks>
+    [SuppressMessage("Usage", "CA2227:Collection properties should be read only",
+        Justification = "Options pattern.")]
+    [Required]
+    [Experimental(diagnosticId: DiagnosticIds.Experiments.Telemetry, UrlFormat = DiagnosticIds.UrlFormat)]
+    public IDictionary<string, DataClassification> RequestQueryParametersDataClasses { get; set; } = new Dictionary<string, DataClassification>();
```
iliar-turdushev, line 53: Since this is a new API we need to describe an API proposal in the issue that requested it, discuss it and approve. What do you think?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: We can, don't see much of issue here. Although it's just a combination of 2 existing metrics,  so I'm not sure that it's really useful

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Actually, we can. Downside of `url.full` is that it forces us to pass fragments if available and also to redact any credentials such as: `https://username:password@www.example.com/` -> `https://REDACTED:REDACTED@www.example.com/`

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Alright, since there is another suggestion like that I'll emit it all the time

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: I'll emit it, since there was another suggestion for it

rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
rainsxng: Yeah, that's a good idea. I'll write it in the issue

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/LoggingOptions.cs
evgenyfedorov2: > Since this is a new API we need to describe an API proposal in the issue that requested it, discuss it and approve.

we do, but this does not stop us from merging this PR, experimental and small API can be merged without review.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpClientLogger.cs
rainsxng: It was, removed this part completely

dariusclay (approved):

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Updated it to use the url.query tag

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpHeadersReader.cs
rainsxng: updated, should be good now

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
```diff
@@ -51,6 +53,12 @@ public static class HttpClientLoggingTagNames
     /// </summary>
     public const string ResponseHeaderPrefix = "http.response.header.";
 
+    /// <summary>
+    /// URL query parameters prefix.
```
iliar-turdushev, line 57:
```suggestion
    /// URL query parameters.
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/LogRecord.cs
```diff
@@ -75,6 +76,16 @@ internal sealed class LogRecord : IResettable
     /// </summary>
     public int PathParametersCount { get; set; }
 
+    /// <summary>
+    /// Gets or sets request query string parameters.
+    /// </summary>
+    public KeyValuePair<string, string>[]? QueryParameters { get; set; }
+
+    /// <summary>
+    /// Gets or sets formatted query string parameters>.
```
iliar-turdushev:
```suggestion
    /// Gets or sets formatted query string parameters.
```

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/HttpClientLoggingTagNames.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
```diff
@@ -573,6 +590,299 @@ public async Task ReadAsync_MetadataWithoutRequestRouteOrNameUsesConstants_Retur
...
+        logRecord.QueryString.Should().BeNullOrEmpty();
+    }
+
+    [Fact]
+    public async Task ReadAsync_SetsEmptyQueryParameters_WhenNoMatchingClassification()
+    {
+        var options = new LoggingOptions
+        {
+            RequestQueryParametersDataClasses = new Dictionary<string, DataClassification>
+        {
```
iliar-turdushev: Please, fix formatting.

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
```diff
@@ -573,6 +590,299 @@ public async Task ReadAsync_MetadataWithoutRequestRouteOrNameUsesConstants_Retur
...
+        logRecord.QueryParameters.Should().BeEmpty();
+    }
+
+    [Fact]
+    public async Task ReadAsync_SetsMultipleQueryParameters_WhenMultipleClassifications()
+    {
+        var options = new LoggingOptions
+        {
+            RequestQueryParametersDataClasses = new Dictionary<string, DataClassification>
+        {
```
iliar-turdushev: Please, fix formatting.

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
```diff
@@ -573,6 +590,299 @@ public async Task ReadAsync_MetadataWithoutRequestRouteOrNameUsesConstants_Retur
...
+            tag => tag.Key == "url.query" && (tag.Value!).Contains("userId=REDACTED"));
+    }
+
+    [Fact]
+    public async Task ReadAsync_DoesntSetQueryString_WhenQueryValueEmpty()
+    {
+        var options = new LoggingOptions
+        {
+            RequestQueryParametersDataClasses = new Dictionary<string, DataClassification>
+        {
```
iliar-turdushev: Please, fix formatting.

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
```diff
@@ -573,6 +590,299 @@ public async Task ReadAsync_MetadataWithoutRequestRouteOrNameUsesConstants_Retur
...
+        var requestContent = _fixture.Create<string>();
+        var queryParamName = "userId";
+        var queryParamValue = "12345";
+        var pathParamName = "orderId";
+        var pathParamValue = "789";
+
+        var options = new LoggingOptions
+        {
+            RequestQueryParametersDataClasses = new Dictionary<string, DataClassification>
+        {
```
iliar-turdushev: Please, fix formatting.

iliar-turdushev (approved):

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
rainsxng: Done, strange that it wasn't picked by the formatter

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
```diff
@@ -573,6 +590,299 @@ public async Task ReadAsync_MetadataWithoutRequestRouteOrNameUsesConstants_Retur
...
+        });
+
+        var logRecord = new LogRecord();
+        var requestHeadersBuffer = new List<KeyValuePair<string, string>>();
+        await reader.ReadRequestAsync(logRecord, httpRequestMessage, requestHeadersBuffer, CancellationToken.None);
+
+        // Assert: path parameter is redacted in the path
+        logRecord.Path.Should().NotContain(pathParamValue);
+        logRecord.Path.Should().Contain(Redacted);
+    }
```
evgenyfedorov2, line 884: Probably the assertion for redacted query parameters is missing?

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +140,103 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+                {
+                    _ = sb.Append('&');
+                }
+
+                _ = sb.Append(Uri.EscapeDataString(logRecord.QueryParameters[i].Key))
+                    .Append('=')
+                    .Append(Uri.EscapeDataString(logRecord.QueryParameters[i].Value));
+            }
+
+            logRecord.QueryString = sb.ToString();
```
evgenyfedorov2: Maybe this was already discussed, but do we really need to log both query string and query parameters? It is duplicated information, isn't it?

rainsxng: Query parameters are not being logged , they're just stored in the log record. I'm fine removing it, it was originally kept there so it can be reused for something else

evgenyfedorov2: I see, ok, I am afraid we have to remove it if it is not logged. It is just wasting CPU cycles for nothing. That means line 144 not needed and we can optimize the `ExtractAndRedactQueryParameters` method - put key/value pairs and `&` to a `StringBuilder` directly instead of the list of key/value pairs

rainsxng: Done

test/Libraries/Microsoft.Extensions.Http.Diagnostics.Tests/Logging/HttpRequestReaderTest.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Logging/Internal/HttpRequestReader.cs
```diff
@@ -111,24 +134,79 @@ public async Task ReadRequestAsync(LogRecord logRecord, HttpRequestMessage reque
...
+#if NET9_0_OR_GREATER
+        return Uri.UnescapeDataString(value);
+#else
+        return Uri.UnescapeDataString(value.ToString());
+#endif
+    }
+
+    private string ExtractAndRedactQueryParameters(string query)
+    {
+        var sb = new System.Text.StringBuilder();
```
evgenyfedorov2: consider using StringBuilderPool https://github.com/dotnet/extensions/blob/efe30a38adfa89d7b6b81df41ea30af0996269ae/src/Shared/Pools/PoolFactory.cs#L204

usage example https://github.com/dotnet/extensions/blob/efe30a38adfa89d7b6b81df41ea30af0996269ae/src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteFormatter.cs#L54-L113

rainsxng: Done

evgenyfedorov2 (approved):

PR #6522, opened 2025-06-16
Add Chat/EmbeddingGeneration/SpeechToTextOptions.Merge

stephentoub:
Provides a built-in way to combine one options instance into another. As with Clone, having this built-in enables us to keep them in-sync with additions we add over time. To create a new instance that combines two other instances, Clone+Merge may be used together.

This also fixes a bug where EmbeddingGenerationOptions.Clone and SpeechToTextOptions.Clone were not copying over RawRepresentationFactory.

cc: @jeffhandley, @eiriktsarpalis, @RogerBarreto, @westey-m
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6522)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -168,14 +168,108 @@ public virtual ChatOptions Clone()
...
         if (Tools is not null)
         {
-            options.Tools = new List<AITool>(Tools);
+            options.Tools = [.. Tools];
         }
 
         return options;
     }
+
+    /// <summary>Merges the options specified by <paramref name="other"/> into this instance.</summary>
```
rogerbarreto, line 182: Given that this accepts a nullable, this is an attempt (when applicable not null `other` is provided).

```suggestion
    /// <summary>Attempt to merge the options specified by <paramref name="other"/> into this instance.</summary>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -168,14 +168,108 @@ public virtual ChatOptions Clone()
...
+                    StopSequences.Add(sequence);
+                }
+            }
+        }
+
+        if (other.Tools is not null)
+        {
+            if (Tools is null)
+            {
+                Tools = [.. other.Tools];
```
rogerbarreto, line 315: As `tool` is a class and potentially could at some specialized form contains state, should we consider having a `virtual AITool Clone()` in it?

```suggestion
                Tools = [.. other.Tools.Select(t => t.Clone())];
```

stephentoub: Hmm. I'm not sure how feasible that would actually be. If this were an AIFunction created by AIFunctionFactory, for example, the thing containing state would be the instance some supplied instance method was hanging off, and we don't have a great way to deep clone said instance. I'm also not sure what the benefits would be. The primary goal here is creating a shallow clone just so that mutations to the ChatOptions itself don't affect the other; we're not, for example, deep cloning values in the AdditionalProperties dictionary.

Can you elaborate on your concerns / share an example?

stephentoub:
I'm going to close this until we have more signal it's needed.

Issue #6483, opened 2025-05-30
[M.E.AI] Tool call format problems


ForkBug:
### Description

When starting a tool call request. there are some issue in the request json.
The following AIFunctionOption is used:

```
                var fo = new AIFunctionFactoryOptions()
                {
                    Name = "get_current_weather",
                    Description = "Get the current weather for a given location",
                    MarshalResult = async (o, _, _) => { return o; },
                    SerializerOptions = new System.Text.Json.JsonSerializerOptions(System.Text.Json.JsonSerializerDefaults.Web)
                    {
                        WriteIndented = false,
                        TypeInfoResolver = new System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver()
                    },
                    JsonSchemaCreateOptions = new AIJsonSchemaCreateOptions()
                    {
                        TransformOptions = new AIJsonSchemaTransformOptions()
                        {
                            DisallowAdditionalProperties = true,
                        }
                    }
                };
```

But the following request json is post:
```
HTTP Request: 49fa56aa-415c-4983-84d4-6ab8059e0372
Method: POST
Uri: https://api.openai.com/chat/completions
Headers:
  Accept: application/json
  User-Agent: OpenAI/2.2.0-beta.4, (.NET 9.0.5; Microsoft Windows 10.0.19045)
  Authorization: Bearer sk-xxxxxxxxxxxxxxxxxxxxxxx
  Content-Type: application/json
Content:
{"tools":[{"function":{"description":"Get the current weather for a given location","name":"get_current_weather","parameters":{
  "type": "object",
  "required": [
    "location",
    "unit"
  ],
  "properties": {
    "location": {
      "type": "string"
    },
    "unit": {
      "type": "string"
    }
  },
  "additionalProperties": false
},"type":"function"}],"messages":[{"role":"user","content":"Please say: Hello, AI World! This is a test."}],"model":"gpt-4o","stream":true,"stream_options":{"include_usage":true},"tool_choice":"auto"}

```

Beautified version:

```
{
    "tools": [
        {
            "function": {
                "description": "Get the current weather for a given location",
                "name": "get_current_weather",
                "parameters": {
                    "type": "object",
                    "required": [
                        "location",
                        "unit"
                    ],
                    "properties": {
                        "location": {
                            "type": "string"
                        },
                        "unit": {
                            "type": "string"
                        }
                    },
                    "additionalProperties": false
                },
                "type": "function"
            }
        ],
        "messages": [
            {
[...]

PR #6523, opened 2025-06-17
Tweak OpenAI JSON schema transforms

stephentoub:
Only require all properties and force additionalProperties: false when strict is set or when structured output is used.

Also update the OpenAI assistant chat client to apply the transforms.

Closes https://github.com/dotnet/extensions/issues/6483
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6523)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
@@ -288,9 +288,7 @@ internal sealed partial class OpenAIAssistantChatClient : IChatClient
                                 strictBool :
                                 null;
 
-                            JsonElement jsonSchema = strict is true ?
-                                OpenAIClientExtensions.StrictSchemaTransformCache.GetOrCreateTransformedSchema(aiFunction) :
-                                OpenAIClientExtensions.NonStrictSchemaTransformCache.GetOrCreateTransformedSchema(aiFunction);
+                            JsonElement jsonSchema = OpenAIClientExtensions.GetSchema(aiFunction, strict);
 
                             runOptions.ToolsOverride.Add(new FunctionToolDefinition(aiFunction.Name)
                             {
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
 using OpenAI;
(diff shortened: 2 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
```diff
@@ -284,13 +284,18 @@ void IDisposable.Dispose()
                     switch (tool)
                     {
                         case AIFunction aiFunction:
-                            bool? strict = aiFunction.AdditionalProperties.TryGetValue(nameof(strict), out var strictValue) && strictValue is bool strictBool ?
+                            bool? strict = aiFunction.AdditionalProperties.TryGetValue(OpenAIClientExtensions.StrictKey, out var strictValue) && strictValue is bool strictBool ?
                                 strictBool :
                                 null;
+
+                            JsonElement jsonSchema = strict is true ?
```
eiriktsarpalis: Consider hiding this logic behind a method in `OpenAIClientExtensions` that accepts `strict` as a parameter.

stephentoub: Done.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -14,6 +16,26 @@ namespace Microsoft.Extensions.AI;
...
+    /// <summary>Gets the default OpenAI endpoint.</summary>
+    internal static Uri DefaultOpenAIEndpoint { get; } = new("https://api.openai.com/v1");
+
+    /// <summary>Gets a <see cref="ChatRole"/> for "developer".</summary>
+    internal static ChatRole ChatRoleDeveloper { get; } = new ChatRole("developer");
+
+    /// <summary>
+    /// Gets the JSON schema transformer cache conforming to OpenAI <b>strict</b> restrictions per https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#supported-schemas.
+    /// </summary>
+    internal static AIJsonSchemaTransformCache StrictSchemaTransformCache { get; } = new(new()
```
eiriktsarpalis, line 31: Can these be marked private now?

stephentoub:
> Thanks. I take it our OpenAI integration tests are passing with these changes?

Yup.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
stephentoub: It's still used directly by the places that use it with structured output / response format, where we always apply it.

PR #6489, opened 2025-06-03
Add resiliency to Resource Monitoring in Linux

amadeuszl:
Partially fixes #6478 

This PR address Linux RM for Disk and Networking. To make PR smaller CPU and Memory will be addressed in separated ones. Later I will check what we can do for Windows.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6489)

PR #6528, opened 2025-06-20
Add resiliency mechanism to CPU and memory utilization checks

amadeuszl:
Complementary PR to #6489 to cover CPU and memory


src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -288,4 +321,34 @@ public Snapshot GetSnapshot()
             userTimeSinceStart: TimeSpan.FromTicks((long)(cgroupTime / Hundred * _scaleRelativeToCpuRequestForTrackerApi)),
             memoryUsageInBytes: memoryUsed);
     }
+
+    private IEnumerable<Measurement<double>> GetMeasurementWithRetry(Func<double> func)
```
evgenyfedorov2, line 325: Why is it `IEnumerable<Measurement<double>>` instead of just `Measurement<double>` ? Or simply `double`?

amadeuszl: Good question, that's because empty collection of `Measurement` won't report any metric. With `double` it can't be nullable in that API, thus would have to be 0, but 0 would be reported as legit value, which is not something we want in that scenario. Same applies to Measurement object, can't be nullable.

Issue #6536, opened 2025-06-23
Add Support for DataAnnotations in JSON Schema Generation


rwoodley:
### Summary: 
Add support for System.ComponentModel.DataAnnotations attributes in  JSON schema generation to enable proper validation constraints in tool schemas. (Originally opened in the https://github.com/modelcontextprotocol
[csharp-sdk](https://github.com/modelcontextprotocol/csharp-sdk) repo as this issue: https://github.com/modelcontextprotocol/csharp-sdk/issues/535 ).

### Problem Statement: 
Currently, the Microsoft.Extensions.AI's AIFunction infrastructure for JSON schema generation doesn't honor standard .NET validation attributes like [Range], [Required], [StringLength], etc. This means that tool parameters cannot have proper validation constraints expressed in their JSON schemas, forcing developers to implement runtime validation manually.

Motivation: LLMs respect JSON schema validation attributes and if available will generate correctly formed JSON in one shot. This is preferable to having the MCP server throw validation exceptions which force the LLM to iterate to generate valid inputs.

#### Current Behavior: 
When a record parameter has the following attributes:
```
public record SimulationSettings(
    [Range(-1, 10, ErrorMessage = "Must be between -1 and 10")]
    decimal MarketROR
);
```

The generated JSON schema only includes:
```
{
  "marketROR": {
    "type": "number"
  }
}
```

 #### Expected Behavior: 

The generated JSON schema should include validation constraints:

```
{
  "marketROR": {
    "type": "number",
    "minimum": -1,
    "maximum": 10
  }
}
```

### Attributes to be supported:

[Range(min, max)] → "minimum" and "maximum" properties
[Required] → "required" array inclusion
[StringLength(max)] → "maxLength" property
[RegularExpression(pattern)] → "pattern" property
[EmailAddress] → "format": "email"

### Benefits:

Standards Compliance: Leverages existing .NET validation patterns
Better DX: Developers can use familiar validation attributes
Client Validation: MCP clients can validate inputs before making tool calls
Documentation: Schema serves as self-documenting API contracts
Consistency: Aligns with ASP.NET Core, Minimal APIs, and other .NET frameworks

Thank you.

stephentoub:
This seems like a reasonable thing to do in AIJsonUtilities. If we ever pushed it down further into STJ, then we could remove the relevant support in AIJsonUtilities and just rely on the lower layer handling it.

PR #6540, opened 2025-06-25
Augment AIJsonUtilities.CreateJsonSchema for more types and annotations

stephentoub:
Closes https://github.com/dotnet/extensions/issues/6536
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6540)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
@@ -3,6 +3,9 @@
 
 using System;
 using System.ComponentModel;
+#if NET
+using System.ComponentModel.DataAnnotations;
+#endif
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
@@ -14,11 +17,11 @@ using System.Text.Json.Serialization;
 using System.Threading;
 using Microsoft.Shared.Diagnostics;
 
-#pragma warning disable S1121 // Assignments should not be made from within sub-expressions
 #pragma warning disable S107 // Methods should not have too many parameters
-#pragma warning disable S1075 // URIs should not be hardcoded
-#pragma warning disable SA1118 // Parameter should not span multiple lines
 #pragma warning disable S109 // Magic numbers should not be used
+#pragma warning disable S1075 // URIs should not be hardcoded
+#pragma warning disable S1121 // Assignments should not be made from within sub-expressions
+#pragma warning disable SA1118 // Parameter should not span multiple lines
 
 namespace Microsoft.Extensions.AI;
 
@@ -38,6 +41,19 @@ public static partial class AIJsonUtilities
     private const string AdditionalPropertiesPropertyName = "additionalProperties";
     private const string DefaultPropertyName = "default";
     private const string RefPropertyName = "$ref";
+    private const string FormatPropertyName = "format";
+#if NET
+    private const string ContentEncodingPropertyName = "contentEncoding";
+    private const string ContentMediaTypePropertyName = "contentMediaType";
+    private const string MinLengthStringPropertyName = "minLength";
+    private const string MaxLengthStringPropertyName = "maxLength";
+    private const string MinLengthCollectionPropertyName = "minItems";
+    private const string MaxLengthCollectionPropertyName = "maxItems";
+    private const string MinRangePropertyName = "minimum";
+    private const string MaxRangePropertyName = "maximum";
+    private const string MinExclusiveRangePropertyName = "exclusiveMinimum";
+    private const string MaxExclusiveRangePropertyName = "exclusiveMaximum";
+#endif
 
     /// <summary>The uri used when populating the $schema keyword in created schemas.</summary>
     private const string SchemaKeywordUri = "https://json-schema.org/draft/2020-12/schema";
@@ -318,6 +334,9 @@ public static partial class AIJsonUtilities
                 ConvertSchemaToObject(ref schema).InsertAtStart(SchemaPropertyName, (JsonNode)SchemaKeywordUri);
             }
 
+            ApplyDataTypeFormats(parameterName, ref schema, ctx);
+            ApplyDataAnnotations(parameterName, ref schema, ctx);
+
             // Finally, apply any user-defined transformations if specified.
             if (inferenceOptions.TransformSchemaNode is { } transformer)
             {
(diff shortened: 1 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -38,6 +41,19 @@ public static partial class AIJsonUtilities
     private const string AdditionalPropertiesPropertyName = "additionalProperties";
     private const string DefaultPropertyName = "default";
     private const string RefPropertyName = "$ref";
+    private const string FormatPropertyName = "format";
```
eiriktsarpalis, line 46: Note that some of these keywords are explicitly noted as not supported by OpenAI:

https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat#some-type-specific-keywords-are-not-yet-supported

So we might want to make sure they don't break existing OpenAI clients.

stephentoub: > explicitly noted as not supported by OpenAI

When `strict` is opted into.

> So we might want to make sure they don't break existing OpenAI clients.

You can see that's already being handled in this PR in the OpenAI client.
https://github.com/dotnet/extensions/pull/6540/commits/e838066b6f1dd1df40661608e7e2d832f1dca2af#diff-cafca5cb037f3a61fd2013d87230747c733a131b08bf79d147c3112af5b15769R42

But yes, it's possible other OpenAI IChatClient implementations that enable `strict` will need to compensate.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -345,6 +353,234 @@ static JsonObject ConvertSchemaToObject(ref JsonNode schema)
                         return obj;
                 }
             }
+
+            void ApplyDataAnnotations(string? parameterName, ref JsonNode schema, AIJsonSchemaCreateContext ctx)
```
eiriktsarpalis, line 360: Looks good to me. Tagging @captainsafia in case she wants to weigh in from her OpenAPI experience.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -34,6 +39,61 @@ public static class OpenAIClientExtensions
...
+            // Move content from common but unsupported properties to description. In particular, we focus on properties that
+            // the AIJsonUtilities schema generator might produce.
+            // Based on guidance at:
+            // https://platform.openai.com/docs/guides/structured-outputs#supported-properties
+
+            if (node is JsonObject schemaObj)
+            {
+                StringBuilder? additionalDescription = null;
+
+                foreach (string propName in (ReadOnlySpan<string>)["contentEncoding", "contentMediaType", "minLength", "maxLength", "not"])
```
eiriktsarpalis: Should we populate this with the full list of banned keywords as documented by OpenAI? In this location it is not guaranteed that the schema has been generated by `AIJsonUtilities` so other keywords could be present.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -34,6 +39,61 @@ public static class OpenAIClientExtensions
...
+                    if (schemaObj[propName] is { } propNode)
+                    {
+                        _ = schemaObj.Remove(propName);
+                        AppendLine(ref additionalDescription, propName, propNode);
+                    }
+                }
+
+                if (schemaObj["format"] is { } formatNode)
+                {
+                    if (formatNode.GetValueKind() != JsonValueKind.String ||
```
eiriktsarpalis: Why are we making this conditional on the keyword value? I thought `format` is always disallowed.

stephentoub: > I thought format is always disallowed

Nope.
https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#supported-schemas
> format — Predefined formats for strings. Currently supported:
> date-time
> time
> date
> duration
> email
> hostname
> ipv4
> ipv6
> uuid

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
```diff
@@ -480,6 +448,363 @@ public static void CreateJsonSchema_NullableEnum_IncludesTypeKeyword()
...
+                        "items": {
+                            "type": "integer"
+                        },
+                        "minItems": 1,
+                        "maxItems": 4
+                    }
+                }
+            }
+            """,
+#else
```
eiriktsarpalis: Hard to tell what the diff is exactly between targets, why are they different?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
stephentoub: > Should we populate this with the full list of banned keywords as documented by OpenAI?

They don't provide a block list, only an allow list. What is the full set of possible things that could show up in a schema that's not in that allow list?

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
eiriktsarpalis: I assume it's because of the .NET 8+ data annotation attributes. Could we factor these in a standalone test for easier readbility?

stephentoub: > Hard to tell what the diff is exactly between targets, why are they different?

The data annotation attributes aren't recognized in netstandard because that'd require an additional package reference.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
eiriktsarpalis: There's also this reference further down the doc:
```
For fine-tuned models, we additionally do not support the following:

For strings: minLength, maxLength, pattern, format
For numbers: minimum, maximum, multipleOf
For objects: patternProperties
For arrays: minItems, maxItems
```
How can we detect if a model is fine tuned in the OpenAI client?

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
stephentoub: > Could we factor these in a standalone test for easier readbility?

Ok

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
eiriktsarpalis: https://platform.openai.com/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
stephentoub: > There's also this reference further down the doc

Well, yuck. Ok. I'll just strip all of these.

Quite unfortunate that this distinction exists. I'd venture that 99% of use is not with such models.

PR #6551, opened 2025-06-27
AIFunctionFactory: add a flag for disabling return schema generation.

eiriktsarpalis:
Needed by certain MCP sdk applications.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6551)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -690,7 +690,7 @@ public static partial class AIFunctionFactory
             Name = key.Name ?? GetFunctionName(key.Method);
             Description = key.Description ?? key.Method.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description ?? string.Empty;
             JsonSerializerOptions = serializerOptions;
-            ReturnJsonSchema = returnType is null || key.ExcludeResultFromSchema ? null : AIJsonUtilities.CreateJsonSchema(
+            ReturnJsonSchema = returnType is null || key.ExcludeResultSchema ? null : AIJsonUtilities.CreateJsonSchema(
                 returnType,
                 description: key.Method.ReturnParameter.GetCustomAttribute<DescriptionAttribute>(inherit: true)?.Description,
                 serializerOptions: serializerOptions,
@@ -1036,7 +1036,7 @@ public static partial class AIFunctionFactory
             string? Description,
             Func<ParameterInfo, AIFunctionFactoryOptions.ParameterBindingOptions>? GetBindParameterOptions,
             Func<object?, Type?, CancellationToken, ValueTask<object?>>? MarshalResult,
-            bool ExcludeResultFromSchema,
+            bool ExcludeResultSchema,
             AIJsonSchemaCreateOptions SchemaOptions);
     }
 


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactoryOptions.cs
```diff
@@ -106,6 +106,19 @@ public AIFunctionFactoryOptions()
...
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The default value is <see langword="false"/>.
+    /// </para>
+    /// <para>
+    /// When set to <see langword="true"/>, results in the produced <see cref="AIFunction.ReturnJsonSchema"/> to always be <see langword="null"/>.
+    /// </para>
+    /// </remarks>
+    public bool ExcludeResultFromSchema { get; set; }
```
stephentoub: I have trouble with double negatives. Could we make this CreateResultSchema and have it default to true?

(If we decide to leave it as a double negative, I'd still prefer a different name. It's not excluding a result _from_ a schema, which suggests there's still a schema being created; rather, it's not generating the schema at all. If the result schema were part of the input schema, then this would make sense.)

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactoryOptions.cs
```diff
@@ -106,6 +106,19 @@ public AIFunctionFactoryOptions()
     /// </remarks>
     public Func<object?, Type?, CancellationToken, ValueTask<object?>>? MarshalResult { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the result of the function should be excluded from the generated schema.
```
stephentoub:
```suggestion
    /// Gets or sets a value indicating whether a schema should be created for the function's result type, if possible, and included as <see cref="AIFunction.ReturnJsonSchema" />.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactoryOptions.cs
stephentoub: Also, can we reconsider whether this is necessary at all?

If we can create a schema, we include it; if we can't, we don't.

eiriktsarpalis: > I have trouble with double negatives. Could we make this CreateResultSchema and have it default to true?

It's mirroring a similar flag on parameter-level configuration, `ExcludeFromSchema`.

eiriktsarpalis: > If we can create a schema, we include it; if we can't, we don't.

That would require eating exceptions, which would presumably include cases where a return schema is desirable but the serializer wasn't configured correctly. We can do that, but it goes against serialization-related convention everywhere else.

stephentoub: > That would require eating exceptions

From what? Don't we have a TryGetTypeInfo?

eiriktsarpalis: `TryGetTypeInfo` only accounts for a single failure mode: there not being a contract for a given type. Any other configuration errors (property name conflicts, bad converter factories, etc) invariably surface as exceptions.

If you're suggesting we skip the return schema only for that particular failure mode, wouldn't we want to do the same to parameters for consistency?

stephentoub: > If you're suggesting we skip the return schema only for that particular failure mode, wouldn't we want to do the same to parameters for consistency?

Return schema is optional. Input schema is not.

eiriktsarpalis: Understood, but I'm not sure I agree that we should silently drop it if we fail to generate it. Assuming we don't use a custom result marshaller, calling the function would also trigger the same exception so from my perspective it makes sense to fail fast in the case.

stephentoub: > Assuming we don't use a custom result marshaller

The motivating use cases cited in the PR description _do_ use the custom result marshaler, don't they?

stephentoub: > It's mirroring a similar flag on parameter-level configuration, ExcludeFromSchema.

Ok, then maybe just drop the word From (ExcludeResultSchema)? I don't remember why I named it ExcludeFromSchema; looking at it now, I wish it were IncludeInSchema, but, c'est la vie.

If you really believe we have to have this as a property, fine.

eiriktsarpalis: Done

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -1018,6 +1018,7 @@ private record struct DescriptorKey(
             string? Description,
             Func<ParameterInfo, AIFunctionFactoryOptions.ParameterBindingOptions>? GetBindParameterOptions,
             Func<object?, Type?, CancellationToken, ValueTask<object?>>? MarshalResult,
+            bool ExcludeResultFromSchema,
```
stephentoub: ExcludeResultFromSchema => ExcludeResultSchema

stephentoub (approved):

Issue #6535, opened 2025-06-23
MEAI: Not possible to set strict mode for OpenAI chat clients.


NiceWaffel:
### Description

With pull request #6285 the default value for `jsonSchemaIsStrict` was set to `null` (left at default). This reverted the changes from #6064. @SteveSandersonMS and @eiriktsarpalis [had a discussion on this](https://github.com/dotnet/extensions/pull/6285/files#r2041908238) and decided to remove the flag altogether.

While I do agree that turning on `strict` should not be the default, developers should have the option to set it for the `ChatCompletionOptions`, using an additional property on `ChatOptions` or `ChatResponseFormat`.

I think my comment on #6266 was misunderstood, as I don't see an option to turn strict mode on for structured outputs, [only for tools](https://github.com/eiriktsarpalis/extensions/blob/0c8bc3ea381bb44c00daec2d7f02c4e0413962ec/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs#L491). I also don't see an option how "Developers can still control this manually", as Steve stated in the abovementioned discussion. Please correct me if this is somehow user error.

### Reproduction Steps

-

### Expected behavior

The following chat options should set the `strict` flag in the request to OpenAI.
```cs
var options = new ChatOptions()
{
  AdditionalProperties = new AdditionalPropertiesDictionary
  {
    { "Strict", true }
  }
}
```

### Actual behavior

The flag is left at the default value in the API call, with no obvious way to override this.

### Known Workarounds

None, other than re-implementing the `OpenAIChatClient`.

### Other information

This code in `ToOpenAIOptions(ChatOptions? options)` in `OpenAIChatClient.cs`:
```cs
[... 477 lines of other code ...]
else if (options.ResponseFormat is ChatResponseFormatJson jsonFormat)
{
    result.ResponseFormat = jsonFormat.Schema is string jsonSchema ?
        OpenAI.Chat.ChatResponseFormat.CreateJsonSchemaFormat(jsonFormat.SchemaName ?? "json_schema", BinaryData.FromString(jsonSchema), jsonFormat.SchemaDescription) :
        OpenAI.Chat.ChatResponseFormat.CreateJsonObjectFormat();
}
[...]
```
should be changed to (something like):
```cs
[...]
else if (options.ResponseFormat is ChatResponseFormatJson jsonFormat)
{
    bool? strict = null;
[...]

PR #6552, opened 2025-06-27
Enable specifying "strict" for OpenAI clients via ChatOptions

stephentoub:
Closes https://github.com/dotnet/extensions/issues/6535

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
@@ -21,6 +21,7 @@ using OpenAI.Responses;
 #pragma warning disable S1067 // Expressions should not be too complex
 #pragma warning disable SA1515 // Single-line comment should be preceded by blank line
 #pragma warning disable CA1305 // Specify IFormatProvider
+#pragma warning disable S1135 // Track uses of "TODO" tags
 
 namespace Microsoft.Extensions.AI;
 
@@ -28,7 +29,7 @@ namespace Microsoft.Extensions.AI;
 public static class OpenAIClientExtensions
 {
     /// <summary>Key into AdditionalProperties used to store a strict option.</summary>
-    private const string StrictKey = "strictJsonSchema";
+    private const string StrictKey = "strict";
 
     /// <summary>Gets the default OpenAI endpoint.</summary>
     internal static Uri DefaultOpenAIEndpoint { get; } = new("https://api.openai.com/v1");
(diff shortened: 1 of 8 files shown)


stephentoub:
@eiriktsarpalis, can you help review this? Thanks.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -21,14 +21,15 @@
...
+#pragma warning disable S1135 // Track uses of "TODO" tags
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Provides extension methods for working with <see cref="OpenAIClient"/>s.</summary>
 public static class OpenAIClientExtensions
 {
     /// <summary>Key into AdditionalProperties used to store a strict option.</summary>
-    private const string StrictKey = "strictJsonSchema";
+    private const string StrictKey = "strict";
```
eiriktsarpalis: @SteveSandersonMS had recommended renaming to `strict` to `strictJsonSchema` so that we have a more descriptive identifier. Wouldn't changing this back now be a breaking change?

stephentoub: > renaming to strict to strictJsonSchema so that we have a more descriptive identifier

"strict" is easily inferred and is what someone would naturally think to use. To know "strictJsonSchema" should be used, you need to look at the implementation. It becomes less of an issue if/when we add an extension property, but I don't know when we'll do that.

> Wouldn't changing this back now be a breaking change?

Yes, in that folks that had been using "strictJsonSchema" will no longer be opting into strictness and would need to change it to "strict" to do so. But this whole library is still preview and continues to take breaking changes. I also can't find any uses of "strictJsonSchema" on GitHub (outside of our own use in M.E.AI):
https://grep.app/search?f.lang=C%23&q=%22strictJsonSchema%22
https://github.com/search?q=%5C%22strictJsonSchema%5C%22+language%3AC%23&type=code

I can change it back if you feel strongly about it. It's just very non-intuitive.

eiriktsarpalis: No strong opinions here, I defer to @SteveSandersonMS

PederHP:
@stephentoub 
Just had a case in some production code where o4-mini (as an Azure deployment) was not consistent in giving structured output (using GetResponseAsync < T > for request generation) and sometimes responded with:

```json
{"type":"object","properties":{"field1":"foo","field2":true}}
```

instead of :
```json
{"field1":"foo","field2":true}
```

which caused it to not deserialize as expected into the response type object.

When setting "strictJsonSchema" to true via ChatOptions the response is consistently returned correctly and GetResponseAsync < T > working as expected.

For developer convenience I think it should arguably default to strict, as figuring out that an AdditionalProperties needs to be set in ChatOptions is going to be difficult for a lot of developers. But I am not sure about the ergonomics of having these ChatOptions have defaults in such a manner. Just wanted to share the observations and fix. I realize this is a model issue and not a framework issue, but it might be useful.

stephentoub:
> I think it should arguably default to strict

We tried that and had to revert. OpenAI has strict rules for schemas that are allowed to be used with strict, and a bunch of schemas folks use (especially coming from arbitrary places with MCP) blow up. We've added in some transforms to help minimize the pain, such that we detect _some_ violations and reconfigure the schema, but it's only a bandaid.

PR #6554, opened 2025-06-30
Refactor Resource Monitoring

evgenyfedorov2:
1. Rename `CalculateCpuUsageWithoutHostDelta` to `UseLinuxCpuUsageV2`.
2. Remove `UseDeltaNrPeriodsForCpuCalculation` option and merge its logic with `UseLinuxCpuUsageV2`.
3. Remove debug logs and metrics.
4. Minor code massaging.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6554)


evgenyfedorov2:
@nivebb8 do you want to review this? We are removing debug logs, metrics, and adjusting code to use only one boolean flag instead of two.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -80,28 +65,23 @@ public LinuxUtilizationProvider(IOptions<ResourceMonitoringOptions> options, ILi
...
-
-            // Try to get the CPU request from cgroup
             cpuRequest = _parser.GetCgroupRequestCpuV2();
 
             // Get Cpu periods interval from cgroup
             _cpuPeriodsInterval = _parser.GetCgroupPeriodsIntervalInMicroSecondsV2();
             (_previousCgroupCpuTime, _previousCgroupCpuPeriodCounter) = _parser.GetCgroupCpuUsageInNanosecondsAndCpuPeriodsV2();
 
-            // Initialize the counters
-            _cpuUtilizationLimit100PercentExceededCounter = meter.CreateCounter<long>("cpu_utilization_limit_100_percent_exceeded");
```
amadeuszl, line 102: Why we decided to remove it?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringOptions.cs
```diff
@@ -100,23 +100,17 @@ public partial class ResourceMonitoringOptions
...
-    /// <summary>
-    /// Gets or sets a value indicating whether to use the number of periods in cpu.stat for cgroup CPU usage.
-    /// We use delta time for CPU usage calculation when this flag is not set.
-    /// <value>
-    /// The default value is <see langword="false"/>.
-    /// </value>
-    ///  </summary>
-    [Experimental(diagnosticId: DiagnosticIds.Experiments.ResourceMonitoring, UrlFormat = DiagnosticIds.UrlFormat)]
-    public bool UseDeltaNrPeriodsForCpuCalculation { get; set; }
+    public bool UseLinuxCpuUsageV2 { get; set; }
```
amadeuszl: Without reading the comments the property itself may for some users indicate it's about switching to cgroup v2, is given user knows that his system doesn't support cgroup v2 then it won't be set to false.
Maybe something like `UseLinuxCalculationV2Method`, `UseLinuxCalculationV2`, `UseCpuLimits`, or the other way around `UseLegacyLinuxCalculation`, basically to make sure it can't be interpreted by default as cgroup v2 without reading comments

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -127,79 +106,34 @@ public double CpuUtilizationWithoutHostDelta()
...
+            long deltaCpuPeriodInNanoseconds = deltaPeriodCount * _cpuPeriodsInterval * 1000;
+            double coresUsed = deltaCgroup / (double)deltaCpuPeriodInNanoseconds;
 
-        // Increment counter if utilization exceeds 110%
-        if (utilization > CpuLimitThreshold110Percent)
-        {
-            _cpuUtilizationLimit110PercentExceededCounter?.Add(1);
-            _cpuUtilizationLimit110PercentExceeded++;
-            _logger.CounterMessage110(_cpuUtilizationLimit110PercentExceeded);
+            _logger.CpuUsageDataV2(cpuUsageTime, _previousCgroupCpuTime, deltaCpuPeriodInNanoseconds, coresUsed);
```
amadeuszl, line 160: We intentionally want to log more than 1 (100%), here while in other places logging happens after `Math.Min(1, ..)`?

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/AcceptanceTest.cs
```diff
@@ -359,82 +359,6 @@ public Task ResourceUtilizationTracker_And_Metrics_Report_Same_Values_With_Cgrou
         return Task.CompletedTask;
     }
 
-    [ConditionalFact]
-    [CombinatorialData]
-    [OSSkipCondition(OperatingSystems.Windows | OperatingSystems.MacOSX, SkipReason = "Linux specific tests")]
-    public Task ResourceUtilizationTracker_And_Metrics_Report_Same_Values_With_Cgroupsv2_v2()
```
amadeuszl, line 365: What was problematic with that test?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -288,4 +226,7 @@ public Snapshot GetSnapshot()
             userTimeSinceStart: TimeSpan.FromTicks((long)(cgroupTime / Hundred * _scaleRelativeToCpuRequestForTrackerApi)),
             memoryUsageInBytes: memoryUsed);
     }
+
+    private double CpuUtilizationRequest(double cpuRequest) => Math.Min(One, CpuUtilizationV2() / cpuRequest);
```
amadeuszl, line 294: So it's our last line of defense to not report more than 100%? Do we know example scenario when it happens? We could add a note to why we're adding it for future generations

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
evgenyfedorov2: It was added for debugging purposes only. No business need anymore

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringOptions.cs
evgenyfedorov2: Yeah, happy to discuss it here and at .NET API review.

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/AcceptanceTest.cs
evgenyfedorov2: we had two boolean flags and this test was for a scenario when one flag is true, the other is false. Now, since we have combined both flags into one, this test does not make sense because you can only enable both or none

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
evgenyfedorov2: The value might be slightly over 1, such as 1.00053, due to rounding issues or due to the fact that the calculations are not performed atomically, they are performed when time still goes by. And this is the defense we have, it does not make sense to report such values

Previously, though, the values were way off 1, such as 1.1 and 1.2 due to calculation issues which were fixed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringOptions.cs
evgenyfedorov2: Updated to UseLinuxCalculationV2

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
evgenyfedorov2: We don't know the percentage at this point, we are logging the raw number of cores used (before division by request or limit).
Requests and limits are logged separately on line 95.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
amadeuszl: Makes sense, still though we could add some note in code why it's done this way, for me it's not obvious.

evgenyfedorov2: Sorry forgot to add in the first place. Added now!

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceMonitoringOptions.cs
```diff
@@ -100,23 +100,18 @@ public partial class ResourceMonitoringOptions
...
-    /// <summary>
-    /// Gets or sets a value indicating whether to use the number of periods in cpu.stat for cgroup CPU usage.
-    /// We use delta time for CPU usage calculation when this flag is not set.
-    /// <value>
-    /// The default value is <see langword="false"/>.
-    /// </value>
-    ///  </summary>
-    [Experimental(diagnosticId: DiagnosticIds.Experiments.ResourceMonitoring, UrlFormat = DiagnosticIds.UrlFormat)]
-    public bool UseDeltaNrPeriodsForCpuCalculation { get; set; }
+    public bool UseLinuxCalculationV2 { get; set; }
```
nivebb8, line 114: Can we use CalculateCpuUsageWithoutHostDelta as the option name?
If not, we will need to communicate this to the partners that have already adopted it.

amadeuszl: If we simply want to prevent change in the API I think we're good with the change proposed as it's still experimental.

evgenyfedorov2: > Can we use CalculateCpuUsageWithoutHostDelta as the option name? If not, we will need to communicate this to the partners that have already adopted it.

`UseDeltaNrPeriodsForCpuCalculation` is an experimental/temporal name and it is definitely not going to make it to stable with its current name. I am trying to make it better and prevent leaking out implementation details. I will also bring this name up to .NET API review and try to make it stable (drop `Experimental` attribute). So the expectation here is that this is the one time name change, no more changes are expected.

Yes, this will need to be communicated to partners once it passes API review (I don't expect the partners to upgrade to every single minor version release)

PR #6555, opened 2025-06-30
Implement GLEU and F1 NLP evaluators

peterwald:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6555)


src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
```diff
@@ -79,8 +80,8 @@ public ValueTask<EvaluationResult> EvaluateAsync(
 
         var (score, duration) = TimingHelper.ExecuteWithTiming(() =>
```
shyamnamboodiripad:
```suggestion
        (double score, TimeSpan duration) = TimingHelper.ExecuteWithTiming(() =>
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
```diff
@@ -79,8 +80,8 @@ public ValueTask<EvaluationResult> EvaluateAsync(
...
         {
-            var references = context.References.Select(reference => SimpleWordTokenizer.WordTokenize(reference));
-            var hypothesis = SimpleWordTokenizer.WordTokenize(modelResponse.Text);
+            var references = context.References.Select(reference => SimpleWordTokenizer.WordTokenize(reference).ToArray()).ToArray();
+            var hypothesis = SimpleWordTokenizer.WordTokenize(modelResponse.Text).ToArray();
```
shyamnamboodiripad:
```suggestion
            string[][] references = context.References.Select(reference => SimpleWordTokenizer.WordTokenize(reference).ToArray()).ToArray();
            string[] hypothesis = SimpleWordTokenizer.WordTokenize(modelResponse.Text).ToArray();
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluatorContext.cs
```diff
@@ -52,11 +52,12 @@ public BLEUEvaluatorContext(params string[] references)
     /// <param name="references">
     /// The reference responses against which the response that is being evaluated is compared.
     /// </param>
-    public BLEUEvaluatorContext(IEnumerable<string> references)
+    public BLEUEvaluatorContext(params string[] references)
```
shyamnamboodiripad, line 55: Previously, if an `IEnumerable` was passed to `BLEUEvaluatorContext` constructor, we would end up allocating two collections - one for the `contents` argument for the `base` constructor, and another for the `References` property.

Unless I am mistaken, we will now also allocate an intermediate array via the `ToArray` call above (in addition to allocating the above two collections).

Not sure there is much we could do here though since we want to avoid one extra enumeration for `references` in this case... I wish there were some way to declare a variable within the base constructor call somehow and then use this variable within the constructor body :)

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluatorContext.cs
```diff
@@ -41,8 +41,8 @@ public sealed class BLEUEvaluatorContext : EvaluationContext
     /// <param name="references">
     /// The reference responses against which the response that is being evaluated is compared.
     /// </param>
-    public BLEUEvaluatorContext(params string[] references)
-        : this(references as IEnumerable<string>)
+    public BLEUEvaluatorContext(IEnumerable<string> references)
+        : this(references.ToArray())
```
shyamnamboodiripad, line 45: Based on a quick check, the same pattern of enumerating the supplied `IEnumerable` twice in the constructor also appears in the following `*Context` types. It would be nice to also fix those up such that the base constructor is called in the `params` overload...

```
IntentResolutionEvaluatorContext
RetrievalEvaluatorContext
TaskAdherenceEvaluatorContext
ToolCallAccuracyEvaluatorContext 
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluatorContext.cs
```diff
@@ -52,11 +52,12 @@ public BLEUEvaluatorContext(params string[] references)
...
     /// </param>
-    public BLEUEvaluatorContext(IEnumerable<string> references)
+    public BLEUEvaluatorContext(params string[] references)
         : base(
             name: BLEUContextName,
             contents: [.. references.Select(c => new TextContent(c))])
     {
         References = [.. references];
     }
+
```
shyamnamboodiripad: nit
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/BLEUAlgorithm.cs
```diff
@@ -55,26 +55,26 @@ internal static double BrevityPenalty(int closestRefLength, int hypLength)
...
 
-        if (!references.Any() || !hypothesis.Any())
+        if (references.Length == 0 || hypothesis.Length == 0)
         {
             return RationalNumber.Zero;
         }
 
-        var hyp = hypothesis.CreateNGrams(n);
-        var hypCounts = new MatchCounter<NGram<string>>(hyp);
+        var hypGrams = hypothesis.AsSpan().CreateNGrams(n);
```
shyamnamboodiripad: Consider changing `CreateNGrams` to accept `ReadOnlySpan<T>` (instead of `IEnumerable<T>`) - looks like all call sites are calling `AsSpan()` at the moment (and that would also make `CreateNGrams` symmetric with `CreateNGram` (singular) which already accepts `ReadOnlySpan<T>`).

shyamnamboodiripad:
```suggestion
        List<NGram<string>> hypGrams = hypothesis.AsSpan().CreateNGrams(n);
```

shyamnamboodiripad: Btw, what is the reason to include `AsSpan()` here (and below)?

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/MatchCounter.cs
```diff
@@ -53,7 +53,22 @@ public void AddRange(IEnumerable<T> items)
         }
     }
 
-    public string ToDebugString() => string.Concat(_counts.Select(v => $"{v.Key}: {v.Value}, "));
+    public MatchCounter<T> Intersect(MatchCounter<T> other)
+    {
+        _ = Throw.IfNull(other, nameof(other));
+        var intersection = new MatchCounter<T>();
+        foreach (var kvp in _counts)
```
shyamnamboodiripad: minor: Consider iterating the smaller of the two collections to reduce the overall number of lookups required.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/BLEUAlgorithm.cs
shyamnamboodiripad: Ah never mind - looks like this is already the case.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/NGramExtensions.cs
```diff
@@ -14,28 +13,73 @@ internal static class NGramExtensions
     public static NGram<T> CreateNGram<T>(this ReadOnlySpan<T> values)
         where T : IEquatable<T> => new(values);
 
+    internal static List<NGram<T>> CreateNGrams<T>(this Span<T> input, int n)
```
shyamnamboodiripad: minor: Probably not a significant consideration since the code is internal - but should `CreateNGram` (singular) above also have a similar `Span<T>` overload for completeness?

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/BLEUAlgorithm.cs
shyamnamboodiripad: Never mind - I see that we now use `Slice()` inside the implementation.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/BLEUAlgorithm.cs
```diff
@@ -55,26 +55,26 @@ internal static double BrevityPenalty(int closestRefLength, int hypLength)
...
+        var hypCounts = new MatchCounter<NGram<string>>(hypGrams);
 
         Dictionary<NGram<string>, int> maxCounts = [];
 
         foreach (var rf in references)
         {
-            IEnumerable<NGram<string>> refGrams = rf.CreateNGrams(n);
+            IEnumerable<NGram<string>> refGrams = rf.AsSpan().CreateNGrams(n);
```
shyamnamboodiripad:
```suggestion
            List<NGram<string>> refGrams = rf.AsSpan().CreateNGrams(n);
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/NGramExtensions.cs
```diff
@@ -14,28 +13,73 @@ internal static class NGramExtensions
...
 
         while (output.Length == n)
         {
-            yield return new NGram<T>(output);
+            nGrams.Add(new NGram<T>(output));
 
-            input = input.Skip(1);
-            output = [.. input.Take(n)];
+            input = input.Slice(1);
+            output = input.Slice(0, Math.Min(n, input.Length));
```
shyamnamboodiripad: Looks like changing the while loop condition as follows could make it so that you only need to call `input.Slice(0, Math.Min(n, input.Length))` once (i.e., you could remove the repeat call on line 42 then)

```
while (
    input.Slice(0, Math.Min(n, input.Length)) is ReadOnlySpan<T> output &&
    output.Length == n)
{
    nGrams.Add(new NGram<T>(output));
    input = input.Slice(1);
}
```

shyamnamboodiripad: Consider renaming `output` to `next`.

shyamnamboodiripad: minor: Also consider extracting nested helper functions like below which could make things slightly clearer -

```
void Advance() => input = input.Slice(1);
ReadOnlySpan<T> GetNext() => input.Slice(0, Math.Min(n, input.Length));
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/F1Algorithm.cs
```diff
@@ -0,0 +1,41 @@
...
+        int numCommonTokens = commonTokens.Sum();
+
+        if (numCommonTokens == 0)
+        {
+            return 0.0; // F1 score is 0 if there are no common tokens
+        }
+        else
+        {
+            double precision = (double)numCommonTokens / response.Count();
+            double recall = (double)numCommonTokens / groundTruth.Count();
```
shyamnamboodiripad: Perhaps `groundTruth` and `response` should be `string[]` similar to the other algorithms? This would avoid the need to enumerate `IEnumerable`s that are not already realized collections multiple times...

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/GLEUAlgorithm.cs
```diff
@@ -0,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Evaluation.NLP.Common;
+
+internal static class GLEUAlgorithm
```
shyamnamboodiripad, line 14: Is there some paper or example implementation we could link to in doc comments for this class?

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/F1Algorithm.cs
```diff
@@ -0,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Evaluation.NLP.Common;
+
+internal static class F1Algorithm
```
shyamnamboodiripad, line 13: Is there some paper or example implementation we could link to in doc comments for this class?

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/GLEUAlgorithm.cs
```diff
@@ -0,0 +1,62 @@
...
+            Throw.ArgumentNullException(nameof(references), $"'{nameof(references)}' cannot be null or empty.");
+        }
+
+        if (hypothesis == null || hypothesis.Length == 0)
+        {
+            Throw.ArgumentNullException(nameof(hypothesis), $"'{nameof(hypothesis)}' cannot be null or empty.");
+        }
+
+        MatchCounter<NGram<string>> hypNGrams = new(hypothesis.AsSpan().CreateAllNGrams(minN, maxN));
+        int truePosFalsePos = hypNGrams.Sum();
```
shyamnamboodiripad, line 29: The variable names in this function (like truePosFalsePos, truePosFalseNeg, nAll etc.) are a bit confusing 😄  However, I am guessing this is well understood in the context of the algorithm - and even if we expanded these names, they wouldn't make much sense without knowing exactly how the algorithm works...

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/GLEUAlgorithm.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        List<(int, int)> hypCounts = [];
+        foreach (var reference in references)
+        {
+            MatchCounter<NGram<string>> refNGrams = new(reference.AsSpan().CreateAllNGrams(minN, maxN));
```
shyamnamboodiripad: This comment applies to all the `Create*NGrams` helpers in this PR which currently require the caller to include `.AsSpan()` in the call site.

Can we add overloads for `Create*NGrams` that are based on arrays / `IEnumerable` so that the call sites can omit the `AsSpan()` calls and simply say `reference.CreateNGrams(n)` / `reference.CreateAllNGrams(minN, maxN)`?

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1Evaluator.cs
```diff
@@ -0,0 +1,87 @@
...
+
+/// <summary>
+/// An <see cref="IEvaluator"/> that evaluates the quality of a response produced by an AI model by comparing
+/// it to a reference response using the F1 scoring algorithm. F1 score is the ratio of the number of shared
+/// words between the generated response and the reference response. 
+/// </summary>
+/// <remarks>
+/// <para>
+/// The <see cref="F1Evaluator"/> computes the F1 score of a response ("hypothesis") in relation to a ground-truth reference
+/// <see cref="F1EvaluatorContext.GroundTruth"/>. The score is returned in a <see cref="NumericMetric"/>
```
shyamnamboodiripad:
```suggestion
/// supplied via <see cref="F1EvaluatorContext.GroundTruth"/>. The score is returned in a <see cref="NumericMetric"/>
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1Evaluator.cs
```diff
@@ -0,0 +1,87 @@
...
+                EvaluationDiagnostic.Error(
+                    $"A value of type '{nameof(F1EvaluatorContext)}' was not found in the '{nameof(additionalContext)}' collection."));
+
+            return new ValueTask<EvaluationResult>(result);
+        }
+
+        var (score, duration) = TimingHelper.ExecuteWithTiming(() =>
+        {
+            var reference = SimpleWordTokenizer.WordTokenize(context.GroundTruth);
+            var hypothesis = SimpleWordTokenizer.WordTokenize(modelResponse.Text);
```
shyamnamboodiripad:
```suggestion
        (double score, TimeSpan duration) = TimingHelper.ExecuteWithTiming(() =>
        {
            string[] reference = SimpleWordTokenizer.WordTokenize(context.GroundTruth).ToArray();
            string[] hypothesis = SimpleWordTokenizer.WordTokenize(modelResponse.Text).ToArray();
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1Evaluator.cs
```diff
@@ -0,0 +1,87 @@
...
+            return F1Algorithm.CalculateF1Score(reference, hypothesis);
+        });
+
+        metric.Value = score;
+        string durationText = $"{duration.TotalSeconds.ToString("F2", CultureInfo.InvariantCulture)} s";
+        metric.AddOrUpdateMetadata(name: "evaluation-duration", value: durationText);
+        metric.AddOrUpdateContext(context);
+        metric.Interpretation = NLPScoreInterpretation.Interpret(metric);
```
shyamnamboodiripad: Since the current assembly only includes NLP scores, the static class name in the call site is a bit superfluous.

Consider making this an extension method so that we can simply say `metric.Interpret()` for all call sites (including the ones in `BLEUEvaluator` and `GLEUEvaluator`).

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1EvaluatorContext.cs
```diff
@@ -0,0 +1,48 @@
...
+// constructor syntax.
+
+namespace Microsoft.Extensions.AI.Evaluation.NLP;
+
+/// <summary>
+/// Contextual information that the <see cref="F1Evaluator"/> uses to compute the F1 score for a response.
+/// </summary>
+/// <remarks>
+/// <see cref="F1Evaluator"/> measures the F1 score of a response compared to a reference. F1 (Google-BLEU)
+/// is a metric used to evaluate the quality of machine-generated text.
```
shyamnamboodiripad:
```suggestion
/// <see cref="F1Evaluator"/> measures the F1 score of a response compared to a reference response that is supplied via
/// <see cref="GroundTruth" />. F1 is a metric used to valuate the quality of machine-generated text. It is the ratio
/// of the number of shared words between the generated response and the reference response.
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1EvaluatorContext.cs
```diff
@@ -0,0 +1,48 @@
...
+/// </remarks>
+public sealed class F1EvaluatorContext : EvaluationContext
+{
+    /// <summary>
+    /// Gets the unique <see cref="EvaluationContext.Name"/> that is used for
+    /// <see cref="F1EvaluatorContext"/>.
+    /// </summary>
+    public static string F1ContextName => "F1 Context";
```
shyamnamboodiripad:
```suggestion
    public static string GroundTruthContextName => "Ground Truth (F1)";
```

This would match the naming convention we use for GroundTruth in other contexts like [CompletenessEvaluatorContext ](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/CompletenessEvaluatorContext.cs#L30), [GroundednessProEvaluatorContext](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/GroundednessProEvaluatorContext.cs#L29) etc.

This would also allow us to refer to the context name as `F1EvaluatorContext.GroundTruthContextName` which would be slightly more informative than `F1EvaluatorContext.F1ContextName` :)

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluatorContext.cs
```diff
@@ -52,11 +52,12 @@ public BLEUEvaluatorContext(params string[] references)
     /// <param name="references">
     /// The reference responses against which the response that is being evaluated is compared.
     /// </param>
-    public BLEUEvaluatorContext(IEnumerable<string> references)
+    public BLEUEvaluatorContext(params string[] references)
         : base(
             name: BLEUContextName,
```
shyamnamboodiripad: Ah I missed this in the previous PR - but could we change this to `ReferencesContextName` instead to match the convention we follow in other contexts like [CompletenessEvaluatorContext ](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Quality/CompletenessEvaluatorContext.cs#L30), [GroundednessProEvaluatorContext](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/GroundednessProEvaluatorContext.cs#L29) etc.

```
public static string ReferencesContextName => "References (BLEU)";
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluatorContext.cs
```diff
@@ -0,0 +1,62 @@
...
+/// <see cref="GLEUEvaluator"/> measures the GLEU score of a response compared to a reference. GLEU (Google-BLEU)
+/// is a metric used to evaluate the quality of machine-generated text.
+/// </remarks>
+public sealed class GLEUEvaluatorContext : EvaluationContext
+{
+    /// <summary>
+    /// Gets the unique <see cref="EvaluationContext.Name"/> that is used for
+    /// <see cref="GLEUEvaluatorContext"/>.
+    /// </summary>
+    public static string GLEUContextName => "GLEU Context";
```
shyamnamboodiripad:
```suggestion
    public static string ReferencesContextName => "References (GLEU)";
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1EvaluatorContext.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Gets the reference response against which the provided chat response will be scored.
```
shyamnamboodiripad:
```suggestion
    /// Gets the reference response against which the provided model response will be scored.
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1EvaluatorContext.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="F1EvaluatorContext"/> class.
+    /// </summary>
+    /// <param name="groundTruth">
+    /// The reference response against which the response that is being evaluated is compared.
+    /// </param>
+    public F1EvaluatorContext(string groundTruth)
+        : base(
+            name: F1ContextName,
+            contents: [new TextContent(groundTruth)])
```
shyamnamboodiripad:
```suggestion
            contents: groundTruth)
```

Iirc there should be an overload that allows to pass just the string.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluator.cs
```diff
@@ -0,0 +1,90 @@
...
+
+/// <summary>
+/// An <see cref="IEvaluator"/> that evaluates the quality of a response produced by an AI model by comparing
+/// it to a reference response using the GLEU (Google-BLEU) algorithm. The GLEU evaluator measures the similarity
+/// between the generated response and one or more reference responses using n-gram overlap.
+/// </summary>
+/// <remarks>
+/// <para>
+/// The <see cref="GLEUEvaluator"/> computes the GLEU score of a response ("hypothesis") compared to a reference
+/// <see cref="GLEUEvaluatorContext.References"/>. The score is returned in a <see cref="NumericMetric"/>
```
shyamnamboodiripad:
```suggestion
/// supplied via <see cref="GLEUEvaluatorContext.References"/>. The score is returned in a <see cref="NumericMetric"/>
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluator.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// <inheritdoc/>
+    public ValueTask<EvaluationResult> EvaluateAsync(
+        IEnumerable<ChatMessage> messages,
+        ChatResponse modelResponse,
+        ChatConfiguration? chatConfiguration = null,
+        IEnumerable<EvaluationContext>? additionalContext = null,
+        CancellationToken cancellationToken = default)
+    {
+        Stopwatch stopwatch = Stopwatch.StartNew();
```
shyamnamboodiripad:
```suggestion
```

This can probably be removed since you are using timing helper below.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluator.cs
```diff
@@ -0,0 +1,90 @@
...
+        if (additionalContext?.OfType<GLEUEvaluatorContext>().FirstOrDefault()
+                is not GLEUEvaluatorContext context)
+        {
+            metric.AddDiagnostics(
+                EvaluationDiagnostic.Error(
+                    $"A value of type '{nameof(GLEUEvaluatorContext)}' was not found in the '{nameof(additionalContext)}' collection."));
+
+            return new ValueTask<EvaluationResult>(result);
+        }
+
```
shyamnamboodiripad, line 71:
```suggestion

        if (context.References.Count is 0)
        {
            metric.AddDiagnostics(
                EvaluationDiagnostic.Error(
                    $"Supplied '{nameof(GLEUEvaluatorContext)}' did not contain any '{nameof(GLEUEvaluatorContext.References)}'."));
            return new ValueTask<EvaluationResult>(result);
        }
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluatorContext.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// <summary>
+    /// Gets the reference response against which the provided chat response will be scored.
```
shyamnamboodiripad:
```suggestion
    /// Gets the reference responses against which the provided model response will be scored.
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluatorContext.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="GLEUEvaluator"/> measures the degree to which the response being evaluated is similar to
+    /// the response supplied via <see cref="References"/>. The metric will be reported as a GLEU score.
```
shyamnamboodiripad, line 34:
```suggestion
    /// the responses supplied via <see cref="References"/>. The metric will be reported as a GLEU score.
```

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluatorContext.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Extensions.AI.Evaluation.NLP;
+
+/// <summary>
+/// Contextual information that the <see cref="GLEUEvaluator"/> uses to compute the GLEU score for a response.
+/// </summary>
+/// <remarks>
+/// <see cref="GLEUEvaluator"/> measures the GLEU score of a response compared to a reference. GLEU (Google-BLEU)
```
shyamnamboodiripad:
```suggestion
/// <see cref="GLEUEvaluator"/> measures the GLEU score of a response compared to one or more reference responses supplied via <see cref="References"/>. GLEU (Google-BLEU)
```

shyamnamboodiripad: Looks like a similar change may also be needed in `BLEUEvaluatorContext` at the same spot

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluatorContext.cs
shyamnamboodiripad: Looks like a similar change may also be needed in `BLEUEvaluatorContext` at the same spot

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/GLEUEvaluatorContext.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="GLEUEvaluatorContext"/> class.
+    /// </summary>
+    /// <param name="references">
+    /// The reference responses against which the response that is being evaluated is compared.
+    /// </param>
+    public GLEUEvaluatorContext(IEnumerable<string> references)
```
shyamnamboodiripad: Looks like this needs the same change as BLEUEvaluatorContext to call the base constructor from the params constructor and call the params constructor (with .ToArray) from the IEnumerable constructor.

test/Libraries/Microsoft.Extensions.AI.Evaluation.NLP.Tests/F1EvaluatorTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using Xunit;
+
+namespace Microsoft.Extensions.AI.Evaluation.NLP.Tests;
+
+#pragma warning disable AIEVAL001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+
+public class F1EvaluatorTests
+{
+    [Fact]
+    public async Task EvaluateAsync_ReturnsPerfectScoreForIdenticalText()
```
shyamnamboodiripad: Please remove the EvaluateAsync prefix to keep the test names shorter

test/Libraries/Microsoft.Extensions.AI.Evaluation.NLP.Tests/F1EvaluatorTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    [InlineData("It is a guide to action that ensures that the military will forever heed Party commands",
+        "It is a guide to action which ensures that the military always obeys the commands of the party", 0.70589)]
+    [InlineData("It is the practical guide for the army always to heed the directions of the party",
+        "It is to insure the troops forever hearing the activity guidebook that party direct", 0.4000)]
+    public async Task EvaluateAsync_SampleCases(string reference, string hypothesis, double score)
+    {
+        var evaluator = new F1Evaluator();
+        var context = new F1EvaluatorContext(reference);
+        var response = new ChatResponse(new ChatMessage(ChatRole.Assistant, hypothesis));
+        var result = await evaluator.EvaluateAsync([], response, null, [context]);
```
shyamnamboodiripad: As mentioned in the previous PR, we can just use call EvaluateAsync overload that takes a string for the response and omit the ChatResponse construction + the empty array for the first argument.

test/Libraries/Microsoft.Extensions.AI.Evaluation.NLP.Tests/GLEUAlgorithmTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            "It is a guide to action that ensures that the military will forever heed Party commands".Split(' '),
+        ];
+        string[] hypothesis = "It is a guide to action which ensures that the military always obeys the commands of the party".Split(' ');
+
+        double score = SentenceGLEU(references, hypothesis);
+        Assert.Equal(0.43939, score, 4);
+    }
+
+    [Fact]
+    public void TestSentenceGLEUMilitaryExampleB()
```
shyamnamboodiripad, line 60: The test names in this file are slightly odd :) I don't feel too strongly about it but one way to avoid this could be to use a theory with each example (and its expected result) becoming an inlinedata

test/Libraries/Microsoft.Extensions.AI.Evaluation.NLP.Tests/GLEUEvaluatorTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+public class GLEUEvaluatorTests
+{
+    [Fact]
+    public async Task EvaluateAsync_ReturnsPerfectScoreForIdenticalText()
+    {
+        var evaluator = new GLEUEvaluator();
+        var context = new GLEUEvaluatorContext("The quick brown fox jumps over the lazy dog.");
+        var response = new ChatResponse(new ChatMessage(ChatRole.Assistant, "The quick brown fox jumps over the lazy dog."));
+        var result = await evaluator.EvaluateAsync([], response, null, [context]);
```
shyamnamboodiripad: Similar comment about using simpler EvaluateAsync overload.

test/Libraries/Microsoft.Extensions.AI.Evaluation.NLP.Tests/GLEUEvaluatorTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using Xunit;
+
+namespace Microsoft.Extensions.AI.Evaluation.NLP.Tests;
+
+#pragma warning disable AIEVAL001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+
+public class GLEUEvaluatorTests
+{
+    [Fact]
+    public async Task EvaluateAsync_ReturnsPerfectScoreForIdenticalText()
```
shyamnamboodiripad: Same comment about removing the EvaluateAsync_ prefix.

shyamnamboodiripad (approved):
👍🏾 modulo comments :shipit:

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluatorContext.cs
peterwald: True. I'm not sure it is a high-priority area as this code should not normally be performance critical.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/MatchCounter.cs
peterwald: Good call.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/NGramExtensions.cs
peterwald: I only added the others because the compiler wouldn't auto-convert in our use cases. I think I'll leave it as is for now until we need it. I really wish we didn't need them.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/NGramExtensions.cs
peterwald: Putting the `Slice()` call in the while condition is a clever idea, but it causes analyzer warnings due to the expression always being true. Also, I feel like it is less readable given that it is relying on a side effect. 

I feel the same way about adding the suggested helper functions. It wouldn't be any less code, given that this is only two lines.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/F1Algorithm.cs
peterwald: Good catch. I was intending to make that change, but missed this spot.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/GLEUAlgorithm.cs
peterwald: Yes, they make sense in the context of the algorithm. `truePosFalsePos` is the number of true position matches plus the number of false positive matches. They all follow that pattern.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1EvaluatorContext.cs
peterwald: That makes sense, and I've made the change. However, this logic breaks down as soon as you have more than one property in a context type. The name is reflecting one property in the context, if there were more than one it would get very confusing. Maybe we will never get into that situation, but it's worth considering.

test/Libraries/Microsoft.Extensions.AI.Evaluation.NLP.Tests/GLEUAlgorithmTests.cs
peterwald: Yeah, I go back and forth. There are some benefits to using inline data, but when you have very large data to pass in, it becomes unwieldy and hard to grok. Also, running Theory tests in VS is maddening.

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/NGramExtensions.cs
shyamnamboodiripad: > it causes analyzer warnings due to the expression always being true

That is surprising - is that a bug in the analyzer?

I was a bit confused reading the code and thinking if we could make it a bit clearer. But perhaps renaming the variable already addresses this...

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/F1EvaluatorContext.cs
shyamnamboodiripad: The name is mainly used only for reporting as a label for "contents" passed to the base constructor.

If there is more than one property, we'll need to come up with a name that encompasses all of them. The name of the constant itself does not matter as much since it is not directly used anywhere else - however, since it is public, ideally the name can call out what the context is and can be distinct from the name of the class itself (which is mainly what we were trying to ensure with the existing convention).

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/Common/NGramExtensions.cs
peterwald: It's not a bug, as the expression would always be true. i.e. `input.Slice() is ReadOnlySpan<T>` must be true by definition.

shyamnamboodiripad: Ah because ReadOnlySpan<T> is a struct - if it were a class the `is` check also does `null` checking which means the expression cannot be statically flagged as always true.

I also just tried this in a simple project and don't see a similar warning. I guess it must be an analyzer that is turned on in this repo.

image

PR #6556, opened 2025-06-30
Efficiency updates to BLEU score processing

peterwald:
Use arrays in place of IEnumerable for internal processing, since they will be ultimately turned into arrays anyway, we might as well allocate them at the beginning and pass through the call stacks. This also saves multiple iterations over the same IEnumerator.
Add a few tweaks, like matching expressions in the tokenizer and Array.Fill
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6556)


peterwald:
Merging this with this PR #6555

Issue #6550, opened 2025-06-27
[API Proposal]: Add 'AdditionalCacheKeyItems' property to DistributedCachingChatClient and DistributedCachingEmbeddingGenerator


darthtrevino:
### Background and motivation

For our work on GraphRag internally, we normally use the model-name as part of our cache key in our Python-based codebases since different models will yield dramatically different results.

It appears that there's already a mechanism where the `DistributedCachingChatClient` and `DistributedCachingEmbeddingGenerator` can be configured by users: via the `configure` argument within `UseDistributedCache`. So this can just be a property on the client classes that's visible to developers.

### API Proposal

```csharp
class Test {
  public void test() {
    string modelName = "gpt-4o";
    ChatClientBuilder builder = createChatClientBuilder()
       .UseDistributedCache(configure: options =>
        {
            options.AdditionalCacheKeys = [modelName]
        });
  }
}
```

### API Usage

See example above

stephentoub:
We could certainly consider adding it as an extra property.

You shouldn't be blocked, though, as you can already derive from DistributedCachingChatClient/EmbeddingGenerator and override the GetCacheKey method, e.g.
```C#
protected override string GetCacheKey(IEnumerable<ChatMessage> messages, ChatOptions? options, params ReadOnlySpan<object?> additionalValues) =>
    base.GetCacheKey(messages, options, [.. additionalValues, modelName]);
```

For reference, the ResponseCachingChatClient in the Evaluations library does that here:
https://github.com/dotnet/extensions/blob/f99c625e3107c56ed94a5b0855378345e9923cf1/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/ResponseCachingChatClient.cs#L128-L129

stephentoub:
@shyamnamboodiripad / @peterwald , would it help or hurt or be a nop for you if there were exposed as an additional public property on DistributedXx?

peterwald:
I don't think it would make a difference for us. I can't think of any reason it would hurt.

shyamnamboodiripad:
Agreed - this should be a nop for `ResponseCachingChatClinet` specifically.

Only minor consideration I can think of is that having more than one mechanism could make debugging slightly more difficult since there would be more than one place to check to understand what things can impact the caching (e.g., when multiple nested `Distributed*Clients` are in play and / or when some of these clients are defined in 3P assemblies etc.).

[...]

PR #6558, opened 2025-07-01
Add DistributedCachingChatClient/EmbeddingGenerator.AdditionalCacheKeyValues

stephentoub:
GetCacheKey already enabled augmenting the key list, but doing so required deriving a custom caching client. It appears to be a reasonably common thing to want to configure, so exposing it as a property as well.

While doing this, I also removed additional per-call allocation coming from GetCacheKey.

Closes https://github.com/dotnet/extensions/issues/6550
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6558)


peterwald (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -56,6 +65,13 @@ public JsonSerializerOptions JsonSerializerOptions
         set => _jsonSerializerOptions = Throw.IfNull(value);
     }
 
+    /// <summary>Gets or sets additional cache key values used to inform the key employed for storing state.</summary>
+    public IReadOnlyList<object>? AdditionalCacheKeyValues
```
stephentoub: @eiriktsarpalis, are you good with this public API?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -122,9 +138,26 @@ protected override async Task WriteCacheStreamingAsync(string key, IReadOnlyList
...
 
-        return AIJsonUtilities.HashDataToString([CacheVersion, messages, options, .. additionalValues], _jsonSerializerOptions);
+        object?[] arr = ArrayPool<object?>.Shared.Rent(length);
+        try
+        {
+            arr[0] = _cacheVersion;
+            arr[1] = messages;
+            arr[2] = options;
+            additionalValues.CopyTo(arr.AsSpan(FixedValuesCount));
+            clientValues.CopyTo(arr, FixedValuesCount + additionalValues.Length);
```
eiriktsarpalis, line 154: Probably not important given how it's used, but I would have expected `_additionalCacheKeyValues` to go before keys scoped to the method call.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
eiriktsarpalis: Given that we have a method parameter named `additionalValues` calling this `AdditionalCacheKeyValues` seems somewhat inconsistent. How about calling it `DefaultAdditionalValues` akin to `HttpClient.DefaultRequestHeaders`?

stephentoub: >  method parameter named additionalValues calling this AdditionalCacheKeyValues seems somewhat inconsistent

But it's a parameter to GetCacheKey. Without the method context, what are these additional values to?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
stephentoub: I could go either way. I figured keeping all the parameters together was better, but if you think it'd be better to swap the spans, I don't have a strong preference and can do so.

eiriktsarpalis: Sounds good

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
eiriktsarpalis: How about `CacheKeyDefaultAdditionalValues`? Kind of long but it both identifies its purpose and communicates its relationship to the `additionalValues` parameter.

stephentoub: You're really keen to get "Default" in there? I'd be fine with CacheKeyAdditionalValues instead of my proposed AdditionalCacheKeyValues (i.e. reordering the words). I'm not sure that inserting "Default" adds much though?

eiriktsarpalis: If it's clear to users that this is meant to augment the `additionalValues` parameter, I'm fine with leaving `Default` out of the name.

Issue #6518, opened 2025-06-12
Create an MCP server template for package conventions and ModelContextProtocol starter


joelverhagen:
We've completed a proposal for setting some conventions around MCP NuGet packages:
https://github.com/NuGet/Home/pull/14344

We can ease the steps of setting up the .csproj by having a project template, perhaps shipping with https://www.nuget.org/packages/Microsoft.Extensions.AI.Templates. 

The .csproj will have:
- `McpServer` package type
- ModelContextProtocol dependency
- PackAsTool
- server.json file packed into the .nupkg

It should be a "preview" template.

joelverhagen:
I am interested in helping to contribute this if there is limited availability on your team.

stephentoub:
cc: @jeffhandley

jeffhandley:
Thanks, @joelverhagen. I will reach out to set up time to discuss this with you next week.

joelverhagen:
@jeffhandley, I have made a draft PR so I could try and understand the space a bit more. It seems to be working okay locally.
https://github.com/dotnet/extensions/pull/6547

I'd love to talk and answer some open questions. I'm sure I'm missing a lot :)

joelverhagen:
The first cut is done. Follow-up work can be tracked by other issues.

PR #6571, opened 2025-07-02
Use dnx instead of dotnet tool exec in MCP server template README

joelverhagen:
Related to https://github.com/dotnet/extensions/issues/6518.

I spoke to @baronfel about this. `dnx` is the preferred syntax and is reminiscent of `npx`. I believe this will also provide a preferred failure mode on an old .NET SDK (OS command not found vs. dotnet giving an unhandled `args` error).

We will provide the `dnx` syntax on NuGet.org for the copy-paste VS Code config and I will prescribe `dnx` while I add NuGet support to the MCP Registry.

Apologies for not getting this into the other PR: https://github.com/dotnet/extensions/pull/6547

I am open to feedback.


stephentoub (approved):

timheuer (approved):
sounds good -- i think since we didn't put the framework version option in here and locking to net10.0 this makes sense too.

PR #6572, opened 2025-07-02
AIFunctionFactory: tolerate JSON string function parameters.

### LINKED ISSUE #6544: Complex AIFunction Arguments
### Description

There seems to be a difference in serialization of function arguments between SemanticKernel's KernelFunctions and the AIFunction as an AIKernelFunction.

When a complex item like 
```csharp
public record TodoListItems(string[] Items);
``` 
is used as a function argument, it tries to call the function with a json string representation of that object.

While the jsonschema seems to be correct for both:

```json
{
  "type": "object",
  "properties": {
    "values": {
      "type": "object",
      "properties": {
        "items": {
          "type": "array",
          "items": {
            "type": "string"
          }
        }
      },
      "required": [
        "items"
      ]
    },
    "scope": {
      "type": "string"
    }
  },
  "required": [
    "values",
    "scope"
  ]
}
```

### Reproduction Steps

Let SemanticKernel via AzureOpenAI call these two methods:
```csharp
// this one works
            KernelFunctionFactory.CreateFromMethod((TodoListItems items, string scope)=> items.Items.Reverse().ToArray(),new(){
                FunctionName = "add_todos"
            }),
// this one fails
            AIFunctionFactory.Create((TodoListItems values, string scope) => "success", new()
            {
                Name = "add_special_todos",
                Description = "add todos, but only if the user explicitly asks for special todos"
            }).AsKernelFunction()
```

If you ask to "add some random todos", it will call the delegate.
If you ask to "add some special random todos", it will throw an error.

To debug the AIFunction, i wrapped it:
```csharp
public class DebugAiFunction(AIFunction inner) : AIFunction
{
    public override string Name => inner.Name;
    public override IReadOnlyDictionary<string, object?> AdditionalProperties => inner.AdditionalProperties;
    public override string Description => inner.Description;
    public override JsonElement JsonSchema => inner.JsonSchema;
    public override JsonSerializerOptions JsonSerializerOptions => inner.JsonSerializerOptions;
    public override JsonElement? ReturnJsonSchema => inner.ReturnJsonSchema;
    public override MethodInfo? UnderlyingMethod => inner.UnderlyingMethod;
    protected override ValueTask<object?> InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken)
    {
        return inner.InvokeAsync(arguments, cancellationToken);
    }
}
```

But the SemanticKernel's KernelFunction is internal, so that didn't work there so, I made a Matryoshka:
```csharp
    new DebugAiFunction(KernelFunctionFactory.CreateFromMethod((TodoListItems items, string scope)=> items.Items.Reverse().ToArray(),new(){
        FunctionName = "add_todos"
    }).AsAIFunction()).AsKernelFunction(),
```

Both debug functions see exactly the same argument (as a literal string):
```json
{"Items": ["Buy groceries", "Call mom", "Finish reading book"]}
```

### Expected behavior

The same behavior in a KernelFunction and an AIFunction.

### Actual behavior

System.ArgumentException: 'Object of type 'System.String' cannot be converted to type 

### Regression?

_No response_

### Known Workarounds

Writing a strongly typed plugin in SemanticKernel to wrap the MCP client, but then ... there goes the flexibility of the MCP...

### Configuration

```xml
    <TargetFramework>net9.0</TargetFramework>
    <PackageReference Include="ModelContextProtocol-SemanticKernel" Version="0.3.0-preview-01" />
    <PackageReference Include="Microsoft.SemanticKernel.Connectors.AzureOpenAI" Version="1.58.0" />
```

### Other information

The whole pipeline that went wrong here was something along the likes of:

https://github.com/microsoft/semantic-kernel/blob/fac316e7dc2cd7dabbaab0ee899c04187c301093/dotnet/samples/Demos/ModelContextProtocolClientServer/MCPClient/Samples/MCPToolsSample.cs#L42

With the MCP server expecting an array of arguments.

eiriktsarpalis:
Under certain circumstances, SK kernels can pass `KernelArguments` with parameters formatted as JSON strings to `AIFunction`s created by `AIFunctionFactory`, resulting in argument exceptions. This PR updates `AIFunctionFactory` so that JSON strings are tolerated as function parameters.

FYI @rogerbarreto @SergeyMenshykh

Fix #6544.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6572)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -825,6 +825,21 @@ public static partial class AIFunctionFactory
                     {
                         try
                         {
+                            if (value is string potentiallyJsonString)
+                            {
+                                // Account for the parameter potentially being a JSON string.
+                                // This is needed for compatibility with Semantic Kernel,
+                                // which may pass parameters as JSON strings.
+                                try
+                                {
+                                    return JsonSerializer.Deserialize(potentiallyJsonString, typeInfo);
+                                }
+                                catch (JsonException)
+                                {
+                                    // If the string is not valid JSON, fall through to the round-trip.
+                                }
+                            }
+
                             string json = JsonSerializer.Serialize(value, serializerOptions.GetTypeInfo(value.GetType()));
                             return JsonSerializer.Deserialize(json, typeInfo);
                         }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -827,6 +827,8 @@ public static partial class AIFunctionFactory
                         {
                             if (value is string potentiallyJsonString)
                             {
+                                Debug.Assert(typeInfo.Type != typeof(string), "string parameters should not enter this branch.");
+
                                 // Account for the parameter potentially being a JSON string.
                                 // This is needed for compatibility with Semantic Kernel,
                                 // which may pass parameters as JSON strings.
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -830,8 +830,8 @@ public static partial class AIFunctionFactory
                                 Debug.Assert(typeInfo.Type != typeof(string), "string parameters should not enter this branch.");
 
                                 // Account for the parameter potentially being a JSON string.
-                                // This is needed for compatibility with Semantic Kernel,
-                                // which may pass parameters as JSON strings.
+                                // The value is a string but the type is not. Try to deserialize it under the assumption that it's JSON.
+                                // If it's not, we'll fall through to the default path that makes it valid JSON and then tries to deserialize.
                                 try
                                 {
                                     return JsonSerializer.Deserialize(potentiallyJsonString, typeInfo);
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -26,6 +26,7 @@ using Microsoft.Shared.Diagnostics;
 #pragma warning disable S2333 // Redundant modifiers should not be used
 #pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
 #pragma warning disable SA1202 // Public members should come before private members
+#pragma warning disable SA1203 // Constants should appear before fields
 
 namespace Microsoft.Extensions.AI;
 
@@ -825,7 +826,7 @@ public static partial class AIFunctionFactory
                     {
                         try
                         {
-                            if (value is string potentiallyJsonString)
+                            if (value is string text && IsPotentiallyJson(text))
                             {
                                 Debug.Assert(typeInfo.Type != typeof(string), "string parameters should not enter this branch.");
 
@@ -834,7 +835,7 @@ public static partial class AIFunctionFactory
                                 // If it's not, we'll fall through to the default path that makes it valid JSON and then tries to deserialize.
                                 try
                                 {
-                                    return JsonSerializer.Deserialize(potentiallyJsonString, typeInfo);
+                                    return JsonSerializer.Deserialize(text, typeInfo);
                                 }
                                 catch (JsonException)
                                 {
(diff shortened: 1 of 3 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -825,6 +825,23 @@ static bool IsAsyncMethod(MethodInfo method)
                     {
                         try
                         {
+                            if (value is string potentiallyJsonString)
+                            {
+                                Debug.Assert(typeInfo.Type != typeof(string), "string parameters should not enter this branch.");
+
+                                // Account for the parameter potentially being a JSON string.
+                                // This is needed for compatibility with Semantic Kernel,
+                                // which may pass parameters as JSON strings.
```
stephentoub: Is this something that can/should be fixed at the SK layer instead?

Or if what SK is doing is appropriate, then should this comment be tweaked? At that point this wouldn't be about SK compat and would instead be about supporting such usage in general, regardless of the consumer who at that point is doing something reasonable by definition.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -825,6 +825,23 @@ static bool IsAsyncMethod(MethodInfo method)
...
+                                try
+                                {
+                                    return JsonSerializer.Deserialize(potentiallyJsonString, typeInfo);
+                                }
+                                catch (JsonException)
```
stephentoub, line 840: If we think this might be common, is there any screening we can do on potentiallyJsonString to avoid the cost of the exception as much as possible?

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
```diff
@@ -341,6 +341,39 @@ public virtual async Task FunctionInvocation_NestedParameters()
...
+
+        var sourceName = Guid.NewGuid().ToString();
+        var activities = new List<Activity>();
+        using var tracerProvider = OpenTelemetry.Sdk.CreateTracerProviderBuilder()
+            .AddSource(sourceName)
+            .AddInMemoryExporter(activities)
+            .Build();
+
+        using var chatClient = new FunctionInvokingChatClient(
+            new OpenTelemetryChatClient(_chatClient, sourceName: sourceName));
```
stephentoub, line 357: Is any of the otel stuff necessary for this test?

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
```diff
@@ -341,6 +341,39 @@ public virtual async Task FunctionInvocation_NestedParameters()
...
+        string? shopperName = null;
+        List<string> shoppingCart = [];
+        AIFunction func = AIFunctionFactory.Create((string[] items, string shopperId) => { shoppingCart.AddRange(items); shopperName = shopperId; }, "AddItemsToShoppingCart");
+        var response = await chatClient.GetResponseAsync(messages, new()
+        {
+            Tools = [func]
+        });
+
+        Assert.Equal("Peter", shopperName);
+        Assert.Equal(["bacon", "lettuce", "tomatoes"], shoppingCart);
```
stephentoub, line 373: Will this hit the path being updated? Or it's just validating arrays in general and not actually related to the fix?

eiriktsarpalis: It won't, like you said it's just validating the scenario as originally reported.

test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ChatClientIntegrationTests.cs
eiriktsarpalis: Probably not, I just copied it over from the related test above.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
eiriktsarpalis: If you're asking whether there's an efficient `IsValidJson` predicate we can apply here, there isn't one; this is the fastest way afforded by STJ.

Even though hitting this branch is expected to be common in the case of SK, I don't think the exceptional case is common enough for scenaria other than ones where the arguments are invalid.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
eiriktsarpalis: Possibly? I went for this approach because we already do something similar reconciling JsonElement and JsonNode inputs. The fact we're explicitly not entering this path for arguments of type string (something that SK is also doing) makes me less worried about ambiguity.

stephentoub: Yeah, at this point we're about to try to deserialize the input into a T; it's just a question of whether we serialize the data to a string before we do so or assume the data is already json.

I think I'd tweak the comment to be more like:
```suggestion
                                // The value is a string but the type is not. Try to deserialize it under the assumption that it's JSON.
                                // If it's not, we'll fall through to the default path that makes it valid JSON and then tries to deserialize.
```
or something along those lines, and not SK specific.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
stephentoub: > If you're asking whether there's an efficient IsValidJson predicate we can apply here

I was more asking if there's an efficient IsDefinitelyNotValidJson that would screen out the most common "this is obviously not JSON" cases.

eiriktsarpalis: The simplest check I can think of is a regex validating that the first non-whitespace character is a valid JSON start token. I don't think we can get better fidelity without escalating to full-blown parsing.

eiriktsarpalis: Added in https://github.com/dotnet/extensions/pull/6572/commits/77d324af616a4d7d2f20e0ead63af7082426d91e

PR #6574, opened 2025-07-02
Add reporting tests that show NLP results.

peterwald:
Also increase the digits in the timing results so it shows something better than `0.00` given that these tests run so fast.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6574)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
@@ -86,7 +86,7 @@ public sealed class BLEUEvaluator : IEvaluator
         });
 
         metric.Value = score;
-        string durationText = $"{duration.TotalSeconds.ToString("F2", CultureInfo.InvariantCulture)} s";
+        string durationText = $"{duration.TotalSeconds.ToString("F4", CultureInfo.InvariantCulture)} s";
         metric.AddOrUpdateMetadata(name: "evaluation-duration", value: durationText);
         metric.AddOrUpdateContext(context);
         metric.Interpretation = metric.Interpret();
--- a/test/Libraries/Microsoft.Extensions.AI.Evaluation.Integration.Tests/NLPEvaluatorTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Evaluation.Integration.Tests/NLPEvaluatorTests.cs
@@ -71,6 +71,33 @@ public class NLPEvaluatorTests
     }
 
     [ConditionalFact]
+    public async Task PartialMatch()
+    {
+        SkipIfNotConfigured();
+
+        await using ScenarioRun scenarioRun =
+            await _nlpReportingConfiguration.CreateScenarioRunAsync(
+                scenarioName: $"Microsoft.Extensions.AI.Evaluation.Integration.Tests.{nameof(NLPEvaluatorTests)}.{nameof(PartialMatch)}");
+
+        var referenceText = "The quick brown fox jumps over the lazy dog.";
+        var bleuContext = new BLEUEvaluatorContext(referenceText);
+        var gleuContext = new GLEUEvaluatorContext(referenceText);
+        var f1Context = new F1EvaluatorContext(referenceText);
+
+        var similarText = "The brown fox quickly jumps over a lazy dog.";
+        EvaluationResult result = await scenarioRun.EvaluateAsync(similarText, [bleuContext, gleuContext, f1Context]);
+
+        Assert.False(
+            result.ContainsDiagnostics(d => d.Severity >= EvaluationDiagnosticSeverity.Warning),
+            string.Join("\r\n\r\n", result.Metrics.Values.SelectMany(m => m.Diagnostics ?? []).Select(d => d.ToString())));
+
+        Assert.Equal(3, result.Metrics.Count);
+        Assert.True(result.TryGet(BLEUEvaluator.BLEUMetricName, out NumericMetric? _));
+        Assert.True(result.TryGet(GLEUEvaluator.GLEUMetricName, out NumericMetric? _));
+        Assert.True(result.TryGet(F1Evaluator.F1MetricName, out NumericMetric? _));
+    }
+
+    [ConditionalFact]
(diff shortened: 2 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
```diff
@@ -86,7 +86,7 @@ public ValueTask<EvaluationResult> EvaluateAsync(
         });
 
         metric.Value = score;
-        string durationText = $"{duration.TotalSeconds.ToString("F2", CultureInfo.InvariantCulture)} s";
+        string durationText = $"{duration.TotalSeconds.ToString("F4", CultureInfo.InvariantCulture)} s";
```
shyamnamboodiripad, line 89: All other evaluators use F2. Any reason to prefer F4 just for the NLP ones?

peterwald: Yes, these evaluators run as fast as 100usecs sometimes. So, without 4 digits, this will be `0.00`.

test/Libraries/Microsoft.Extensions.AI.Evaluation.Integration.Tests/NLPEvaluatorTests.cs
```diff
@@ -0,0 +1,134 @@
...
+            IEvaluator bleuEvaluator = new BLEUEvaluator();
+            IEvaluator gleuEvaluator = new GLEUEvaluator();
+            IEvaluator f1Evaluator = new F1Evaluator();
+
+            _nlpReportingConfiguration =
+                DiskBasedReportingConfiguration.Create(
+                    storageRootPath: Settings.Current.StorageRootPath,
+                    evaluators: [bleuEvaluator, gleuEvaluator, f1Evaluator],
+                    executionName: Constants.Version,
+                    tags: []);
```
shyamnamboodiripad: Could you please add some global tags similar to the other tests? Otherwise, some of the tags that used to be global in the report would become local and we would lose the global tags display in the report... I think there should be at least a few global ones that can be shared across all tests (besides the ones for model and provider which would only apply to the AI-based tests).

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
peterwald: I'm curious if any of the cached evaluators could possibly run so fast as to return zeroes here as well, but I have not run into that.

test/Libraries/Microsoft.Extensions.AI.Evaluation.Integration.Tests/NLPEvaluatorTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        await using ScenarioRun scenarioRun =
+            await _nlpReportingConfiguration.CreateScenarioRunAsync(
+                scenarioName: $"Microsoft.Extensions.AI.Evaluation.Integration.Tests.{nameof(NLPEvaluatorTests)}.{nameof(ExactMatch)}");
+
+        var referenceText = "The quick brown fox jumps over the lazy dog.";
+        var bleuContext = new BLEUEvaluatorContext(referenceText);
+        var gleuContext = new GLEUEvaluatorContext(referenceText);
+        var f1Context = new F1EvaluatorContext(referenceText);
+
+        var response = new ChatResponse(new ChatMessage(ChatRole.Assistant, "The quick brown fox jumps over the lazy dog."));
```
shyamnamboodiripad: minor: You could avoid creating the ChatResponse and ChatMessage and just directly pass the string to EvaluateAsync for these tests.

test/Libraries/Microsoft.Extensions.AI.Evaluation.Integration.Tests/NLPEvaluatorTests.cs
shyamnamboodiripad: Consider including a `NLP` tag that would allow quickly filtering down to the tests in this file.

shyamnamboodiripad: Looks like the other test classes include the following 3 global tags - 

image

shyamnamboodiripad: Looking closer - the other tests use `TestClass: <ClassName>` tag. So it would be nice to follow the pattern.

Also, the `Feature: Contex` tag would be nice to add.

image

src/Libraries/Microsoft.Extensions.AI.Evaluation.NLP/BLEUEvaluator.cs
shyamnamboodiripad: I think it may be best to use 4 decimal places for all of them. For example, the other ones currently report 0.00 when fulfilled from cache - perhaps including more 0s would be better if it reflects the cost.

I think there is one helper that all of those test's use (`AddChatMetadata`) so it probably would involve changing only one place. You could search for `evaluation-duration` to double check if any of the other evals are using similar custom strings.

test/Libraries/Microsoft.Extensions.AI.Evaluation.Integration.Tests/NLPEvaluatorTests.cs
```diff
@@ -24,6 +25,11 @@ static NLPEvaluatorTests()
     {
         if (Settings.Current.Configured)
         {
+            string version = $"Product Version: {Constants.Version}";
+            string date = $"Date: {DateTime.UtcNow:dddd, dd MMMM yyyy}";
+            string projectName = $"Project: Integration Tests";
+            string testClass = $"Test Class: {nameof(NLPEvaluatorTests)}";
```
shyamnamboodiripad, line 31: Could you also add `Feature: Context` (since all NLP evals require additional context)?

PR #6576, opened 2025-07-03
AIFunctionFactory: add test coverage for JSON comments.

eiriktsarpalis:
Following up on #6572 this adds test coverage for JSON comment handling.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6576)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -1058,7 +1058,7 @@ public static partial class AIFunctionFactory
            ( null   # null literal
            | false  # false literal
            | true   # true literal
-           | -?\d   # number
+           | -?[0-9]# number
            | "      # string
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -1058,8 +1058,7 @@ private record struct DescriptorKey(
            ( null   # null literal
            | false  # false literal
            | true   # true literal
-           | \d     # positive number
-           | -\d    # negative number
+           | -?\d   # number
```
stephentoub: This \d will be any Unicode digit. Is that expected? Or should it be `[0-9]`?

eiriktsarpalis:
```suggestion
           | -?[0-9]# number
```

PR #6577, opened 2025-07-03
Update M.E.AI.OpenAI for latest OpenAI release

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6577)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIEmbeddingGeneratorTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIEmbeddingGeneratorTests.cs
@@ -6,7 +6,6 @@ using System.ClientModel;
 using System.ClientModel.Primitives;
 using System.Net.Http;
 using System.Threading.Tasks;
-using Azure.AI.OpenAI;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Caching.Memory;
 using OpenAI;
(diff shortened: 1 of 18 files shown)


test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIEmbeddingGeneratorTests.cs
```diff
@@ -25,17 +24,13 @@ public void AsIEmbeddingGenerator_InvalidArgs_Throws()
...
-    [InlineData(false)]
-    [InlineData(true)]
-    public void AsIEmbeddingGenerator_OpenAIClient_ProducesExpectedMetadata(bool useAzureOpenAI)
+    [Fact]
+    public void AsIEmbeddingGenerator_OpenAIClient_ProducesExpectedMetadata()
     {
         Uri endpoint = new("http://localhost/some/endpoint");
         string model = "amazingModel";
 
-        var client = useAzureOpenAI ?
```
eiriktsarpalis, line 36: I take it the latest version of the sdk has consolidated the two client types?

eiriktsarpalis (approved):

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIEmbeddingGeneratorTests.cs
stephentoub: No, but there's no special logic anywhere in the library about the derived type, and we don't need to be testing derived instances, especially since we no longer have methods that operate on the client itself, but rather only in things retrieved from the client. More practically right now, though, it's there hasn't yet been a new release of the Azure library that's compatible with the new openai library.

PR #6596, opened 2025-07-10
Fix schema generation for Nullable<T> function parameters.

eiriktsarpalis:
Addresses https://github.com/modelcontextprotocol/csharp-sdk/issues/601
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6596)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -289,10 +289,25 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext schemaExporterContext, Js
...
+                if (Nullable.GetUnderlyingType(ctx.TypeInfo.Type) is Type nullableElement)
                 {
-                    objSchema.InsertAtStart(TypePropertyName, new JsonArray { (JsonNode)"string", (JsonNode)"null" });
+                    // Account for bug https://github.com/dotnet/runtime/issues/117493
+                    // null not inserted in the type keyword for root-level Nullable<T> types.
+                    if (objSchema.TryGetPropertyValue(TypePropertyName, out JsonNode? typeKeyWord) &&
+                        typeKeyWord?.GetValueKind() is JsonValueKind.String)
+                    {
+                        string typeValue = typeKeyWord.GetValue<string>()!;
+                        if (typeValue is not null)
```
eiriktsarpalis:
```suggestion
                        if (typeValue is not "null")
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -289,10 +289,25 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext schemaExporterContext, Js
                     objSchema.InsertAtStart(TypePropertyName, "string");
                 }
 
-                // Include the type keyword in nullable enum types
-                if (Nullable.GetUnderlyingType(ctx.TypeInfo.Type)?.IsEnum is true && objSchema.ContainsKey(EnumPropertyName) && !objSchema.ContainsKey(TypePropertyName))
+                if (Nullable.GetUnderlyingType(ctx.TypeInfo.Type) is Type nullableElement)
                 {
-                    objSchema.InsertAtStart(TypePropertyName, new JsonArray { (JsonNode)"string", (JsonNode)"null" });
+                    // Account for bug https://github.com/dotnet/runtime/issues/117493
```
stephentoub, line 316: I assume this means this change can be reverted once M.E.A.Abstractions references STJ 10.x? Can we augment the comment to say that?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -289,10 +289,25 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext schemaExporterContext, Js
...
+                    // null not inserted in the type keyword for root-level Nullable<T> types.
+                    if (objSchema.TryGetPropertyValue(TypePropertyName, out JsonNode? typeKeyWord) &&
+                        typeKeyWord?.GetValueKind() is JsonValueKind.String)
```
stephentoub, line 320: In what case will `typeKeyword` be `null` here (and thus the `?` on its dereference)?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -289,10 +289,25 @@ JsonNode TransformSchemaNode(JsonSchemaExporterContext schemaExporterContext, Js
...
+                    {
+                        string typeValue = typeKeyWord.GetValue<string>()!;
+                        if (typeValue is not "null")
+                        {
+                            objSchema[TypePropertyName] = new JsonArray { (JsonNode)typeValue, (JsonNode)"null" };
```
stephentoub, line 325: Instead of `(JsonNode)typeValue`, can we reuse `typeKeyword`?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
```diff
@@ -617,11 +632,23 @@ private static bool TypeIsIntegerWithStringNumberHandling(AIJsonSchemaCreateCont
                     switch (type)
                     {
                         case "integer" or "number":
+                            if (numericType is not null)
```
stephentoub, line 646: Is this change related, or is it just flagging an existing conflict that we were previously missing? Is it still a conflict if numericType == type?

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
```diff
@@ -452,20 +452,24 @@ JsonSchema CompleteSchema(ref GenerationState state, JsonSchema schema)
 
                 bool IsNullableSchema(ref GenerationState state)
                 {
-                    // A schema is marked as nullable if either
+                    // A schema is marked as nullable if either:
```
stephentoub, line 455: All of the changes in this file are just copies of changes made in runtime, right?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
eiriktsarpalis: When the type keyword is set to `null`. It's unlikely to happen with our generated schemas unless something has gone terribly wrong.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
eiriktsarpalis: Because of how `JsonNode` is designed, you cannot assign on instance to multiple parents. A fresh instance is necessary.

src/Shared/JsonSchemaExporter/JsonSchemaExporter.cs
eiriktsarpalis: Yes, with adaptations where STJ 9 APIs aren't available.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
eiriktsarpalis, line 316:
```suggestion
                    // Account for bug https://github.com/dotnet/runtime/issues/117493
                    // To be removed once System.Text.Json v10 becomes the lowest supported version.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Schema.Create.cs
eiriktsarpalis: It's something we were missing and got hit with the newly added tests. I don't think we need to account for duplicate numeric types since we don't generate that.

jeffhandley (approved):

eiriktsarpalis:
@jeffhandley we need a second approval from dotnet/dotnet-extensions-fundamentals because this PR is touching "shared" components.

jeffhandley:
I connected with @rosebyte to request a review from them or someone else on @dotnet/dotnet-extensions-fundamentals.

PR #6600, opened 2025-07-11
[Logging] Fixes LogProperties and LogPropertyIgnore attributes to work if an object being logged resides in a different assembly than the logging method

### LINKED ISSUE #6598: [Logging] LogPropertyIgnore and LogProperties attributes don't work if an object being logged resides in a different assembly than the logging method
### Description

If an object being logged resides in one assembly and the logging method in another then:

* its properties marked as `LogPropertyIgnore` are still being logged, although they must not be emitted
* for properties marked as `LogProperties` only their `ToString()` representation is logged, although each individual property must be logged separately

### Reproduction Steps

A sample application consists of two projects. ProjectA is a class library that includes an object to be logged, and ProjectB is a console application that logs the object defined in ProjectA.

#### ProjectA

**ProjectA.csproj**

```xml
  <PropertyGroup>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Telemetry.Abstractions" Version="9.6.0" />
  </ItemGroup>
</Project>
```

**ObjectToLog.cs**

```csharp

namespace ProjectA;

public class ObjectToLog
{
    // This property MUST NOT be logged.
    [LogPropertyIgnore]
    public string? PropertyToIgnore { get; set; }

    // This property MUST be logged.
    public string? PropertyToLog { get; set; }

    // This property MUST be logged. Each property of FieldToLog MUST be logged separately.
    [LogProperties]
    public Field? FieldToLog { get; set; }
}

public class Field
{
    public string? Name { get; set; }
    public string? Value { get; set; }
}
```

#### ProjectB

**ProjectB.csproj**

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Telemetry.Abstractions" Version="9.6.0" />
    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="9.0.5" />
  </ItemGroup>
  <ItemGroup>
    <ProjectReference Include="..\ProjectA\ProjectA.csproj" />
  </ItemGroup>
</Project>
```

**Log.cs**

```csharp
using ProjectA;

{
    [LoggerMessage(LogLevel.Information)]
    public static partial void LogObject(this ILogger logger, [LogProperties] ObjectToLog objectToLog);
}
```

**Program.cs**

```csharp
using ProjectA;

using ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
{
    builder.ClearProviders();
    builder.AddJsonConsole(options => options.JsonWriterOptions = new() { Indented = true });
});

ILogger logger = loggerFactory.CreateLogger("Demo");

logger.LogObject(new ObjectToLog
{
    PropertyToIgnore = "Foo",
    PropertyToLog = "Bar",
    FieldToLog = new Field { Name = "Fizz", Value = "Buzz" }
});
```

### Expected behavior

Image

Expected behavior:
* `objectToLog.PropertyToIgnore` wasn't emitted
* `objectToLog.FieldToLog.Name` and `objectToLog.FieldToLog.Value` were emitted as separate values

### Actual behavior

Image

Actual incorrect behavior:
* `objectToLog.PropertyToIgnore` was emitted
* `ToString()` representation of `objectToLog.FieldToLog` was emitted instead of its individual properties

### Regression?

_No response_

### Known Workarounds

Put the following line at the beginning of the file that contains a class to log whose properties are marked as `LogPropertyIgnore` or `LogProperties`:

```csharp
#define CODE_GENERATION_ATTRIBUTES

public class ObjectToLog
{
    ...
}
```

### Configuration

_No response_

### Other information

_No response_

iliar-turdushev:
Fixes #6598 by removing the `Conditional` attribute from the definition of the `LogProperties` and `LogPropertyIgnore` attributes.

The `Conditional` attribute is the root cause of the bug. When a property is marked with the `LogPropertyIgnore` attribute, the associated metadata is excluded from the compiled assembly. As a result, the logging method’s code generator, which relies on reflection to determine whether a property should be logged, lacks the necessary information and ends up emitting the property regardless.

When the logging method and the object being logged are located in the same assembly, the behavior differs. In this case, the code generator does not rely on the compiled assembly and can access the metadata associated with the `LogPropertyIgnore` attribute directly. As a result, it correctly identifies which properties should be excluded from logging.

While there is a workaround, i.e. defining the `CODE_GENERATION_ATTRIBUTES` compilation symbol required by the `Conditional` attribute, it is better to fix the bug by removing the `Conditional` attribute. If a user forget to apply the workaround he/she risks leaking privacy or security sensitive data which could be stored in properties marked as `LogPropertyIgnore`.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6600)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LogPropertiesAttribute.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LogPropertiesAttribute.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using Microsoft.Extensions.Logging;
 using Microsoft.Shared.DiagnosticIds;
@@ -14,7 +13,6 @@ namespace Microsoft.Extensions.Logging;
 /// </summary>
 /// <seealso cref="LoggerMessageAttribute"/>
 [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Property)]
-[Conditional("CODE_GENERATION_ATTRIBUTES")]
 public sealed class LogPropertiesAttribute : Attribute
 {
     /// <summary>
--- a/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LogPropertyIgnoreAttribute.cs
+++ b/src/Libraries/Microsoft.Extensions.Telemetry.Abstractions/Logging/LogPropertyIgnoreAttribute.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Diagnostics;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Extensions.Logging;
@@ -12,7 +11,6 @@ namespace Microsoft.Extensions.Logging;
 /// </summary>
 /// <seealso cref="LoggerMessageAttribute"/>.
 [AttributeUsage(AttributeTargets.Property)]
-[Conditional("CODE_GENERATION_ATTRIBUTES")]
 public sealed class LogPropertyIgnoreAttribute : Attribute
 {
 }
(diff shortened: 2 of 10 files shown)


iliar-turdushev:
> I wonder if similar behavior also affects to `TagNameAttribute` and `TagProviderAttribute`.

Yes, it is similar. For the `TagName` and `TagProvider` I'm going to create a separate issue, because they currently have a bug which doesn't allow to use them with properties. At the moment, they can be used only with log method parameters. We'll fix both bugs in the separate issue.

PR #6615, opened 2025-07-16
Add [Description] to DataContent.Uri

stephentoub:
So that if a DataContent has schema generated for it, it's obvious in the schema that the "uri" is specifically a data URI.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6615)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
@@ -5,6 +5,7 @@ using System;
 #if NET
 using System.Buffers;
 using System.Buffers.Text;
+using System.ComponentModel;
 #endif
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -142,6 +143,9 @@ public DataContent(ReadOnlyMemory<byte> data, string mediaType)
     /// or from a <see cref="System.Uri"/>.
     /// </remarks>
     [StringSyntax(StringSyntaxAttribute.Uri)]
+#if NET
```
rogerbarreto, line 146: Curious why this is wraped in `#if NET`, shoudn't the `using` be as well?

stephentoub: > why this is wraped in #if NET

Because data annotations requires a separate nuget package for netstandard2.0 and it's not worth the dependency for this.

>shoudn't the using be as well?

It is.

rogerbarreto: > It is.

Missed that, sure.

stephentoub:
@eiriktsarpalis, any concerns?

PR #6616, opened 2025-07-16
Add DataContent.Name property

stephentoub:
Enables a file name to optionally be associated with a DataContent. DataContent is often used to represent named entities, such as when uploading files.

I went back and forth on whether this should be FileName or just Name. @eiriktsarpalis, @rogerbarreto, do you have an opinion?
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6616)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
@@ -183,12 +183,12 @@ public class DataContent : AIContent
     [JsonIgnore]
     public string MediaType { get; }
 
-    /// <summary>Gets or sets an optional file name associated with the data.</summary>
+    /// <summary>Gets or sets an optional name associated with the data.</summary>
     /// <remarks>
     /// A service might use this name as part of citations or to help infer the type of data
-    /// being represented based on the file extension.
+    /// being represented based on a file extension.
     /// </remarks>
-    public string? FileName { get; set; }
+    public string? Name { get; set; }
 
     /// <summary>Gets the data represented by this instance.</summary>
     /// <remarks>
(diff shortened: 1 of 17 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/DataContent.cs
```diff
@@ -183,6 +183,13 @@ public string Uri
     [JsonIgnore]
     public string MediaType { get; }
 
+    /// <summary>Gets or sets an optional file name associated with the data.</summary>
+    /// <remarks>
+    /// A service might use this name as part of citations or to help infer the type of data
+    /// being represented based on the file extension.
+    /// </remarks>
+    public string? FileName { get; set; }
```
eiriktsarpalis: "FileName" sounds somewhat specific to local filesystems. Should we be calling it "Uri" instead?

eiriktsarpalis: I take it though that you're intending this contain just the name, i.e. `payments.csv` as opposed to a full path that might reveal more context i.e. `/contoso/payroll/payments.csv`?

stephentoub: > I take it though that you're intending this contain just the name, i.e. payments.csv

Yup.

We could just call it Name if that's desirable... I went back and forth on that.

Suggestions?

eiriktsarpalis: Name sounds good

eiriktsarpalis (approved):

PR #6619, opened 2025-07-16
Add TextContent.Annotations

stephentoub:
Adds support to M.E.AI.Abstractions for annotations.

Some design notes:
- Right now, I have the Annotations property on TextContent. We can push it down to the base later without it being a breaking change.
- Most services only support citations. OpenAI has a broader notion of annotations, and it looks like Anthropic will be adding other kinds of annotations as well. I modeled this by having a base AIAnnotation and a derived CitationAnnotation.
- Different services are very varied on what they support with citations. I went back and forth on whether to have CitationAnnotation just be a flat set of properties, or to try to model it as further derived types, e.g. UrlCitationAnnotation. I've gone with the former, but could be swayed to the latter if y'all think that's more wise. Note that there's a variety of things OpenAI doesn't support or expose that others do, e.g. Google, Anthropic, and AWS all support including snippets of the cited content; I've included such things here, as well.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
@@ -1,0 +1,55 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents an annotation on a span of content.
+/// </summary>
+[JsonDerivedType(typeof(CitationAnnotation), typeDiscriminator: "citation")]
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
+public class AIAnnotation
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AIAnnotation"/> class.
+    /// </summary>
+    public AIAnnotation()
+    {
+    }
+
+    /// <summary>
+    /// Gets or sets the start character index (inclusive) of the annotated span in the <see cref="AIContent"/>.
+    /// </summary>
+    public int? StartIndex { get; set; }
+
+    /// <summary>
+    /// Gets or sets the end character index (exclusive) of the annotated span in the <see cref="AIContent"/>.
+    /// </summary>
+    public int? EndIndex { get; set; }
+
+    /// <summary>
+    /// Gets or sets an optional placeholder or marker text that was present in the original content
+    /// to indicate the citation (e.g., "〖4:0+source.json〗" or "[1]").
+    /// </summary>
+    /// <remarks>
+    /// This is useful for UI rendering or when the original content contained inline citation markers
+    /// that need to be preserved or reconstructed.
+    /// </remarks>
+    public string? Placeholder { get; set; }
+
+    /// <summary>Gets or sets the raw representation of the annotation from an underlying implementation.</summary>
+    /// <remarks>
+    /// If an <see cref="AIAnnotation"/> is created to represent some underlying object from another object
+    /// model, this property can be used to store that original object. This can be useful for debugging or
+    /// for enabling a consumer to access the underlying object model, if needed.
+    /// </remarks>
+    [JsonIgnore]
+    public object? RawRepresentation { get; set; }
+
+    /// <summary>
+    /// Gets or sets additional metadata specific to the provider or source type.
+    /// </summary>
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+}
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
@@ -1,0 +1,26 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a location within source material used for an annotation.</summary>
+public class CitationSourceLocation
+{
+    /// <summary>Gets or sets the page at which the cited material begins.</summary>
+    public int? PageStart { get; set; }
+
+    /// <summary>Gets or sets the page at which the cited material ends.</summary>
+    public int? PageEnd { get; set; }
+
+    /// <summary>Gets or sets the block or chunk at which the cited material begins.</summary>
+    public int? BlockStart { get; set; }
+
+    /// <summary>Gets or sets the block or chunk at which the cited material ends.</summary>
+    public int? BlockEnd { get; set; }
+
+    /// <summary>Gets or sets the character at which the cited material begins.</summary>
+    public int? CharacterStart { get; set; }
+
+    /// <summary>Gets or sets the character at which the cited material ends.</summary>
+    public int? CharacterEnd { get; set; }
+}
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
@@ -203,7 +203,7 @@ internal sealed class OpenAIAssistantChatClient : IChatClient
                     break;
 
                 case MessageContentUpdate mcu:
-                    yield return new(mcu.Role == MessageRole.User ? ChatRole.User : ChatRole.Assistant, mcu.Text)
+                    ChatResponseUpdate textUpdate = new(mcu.Role == MessageRole.User ? ChatRole.User : ChatRole.Assistant, mcu.Text)
                     {
                         AuthorName = _assistantId,
                         ConversationId = threadId,
@@ -211,10 +211,45 @@ internal sealed class OpenAIAssistantChatClient : IChatClient
                         RawRepresentation = mcu,
                         ResponseId = responseId,
                     };
+
+                    // Add any annotations from the text update. The OpenAI Assistants API does not support passing these back
+                    // into the model (MessageContent.FromXx does not support providing annotations), so they end up being one way and are dropped
+                    // on subsequent requests.
+                    if (mcu.TextAnnotation is { } tau)
+                    {
+                        string? fileId = null;
+                        string? toolName = null;
+                        if (!string.IsNullOrWhiteSpace(tau.InputFileId))
+                        {
+                            fileId = tau.InputFileId;
+                            toolName = "file_search";
+                        }
+                        else if (!string.IsNullOrWhiteSpace(tau.OutputFileId))
+                        {
+                            fileId = tau.OutputFileId;
+                            toolName = "code_interpreter";
+                        }
+
+                        if (fileId is not null)
+                        {
+                            (((TextContent)textUpdate.Contents[0]).Annotations ??= []).Add(new CitationAnnotation
+                            {
+                                RawRepresentation = tau,
+                                StartIndex = tau.StartIndex,
+                                EndIndex = tau.EndIndex,
+                                Placeholder = tau.TextToReplace,
+
+                                FileId = fileId,
+                                ToolName = toolName,
+                            });
+                        }
+                    }
+
+                    yield return textUpdate;
                     break;
 
                 default:
-                    yield return new ChatResponseUpdate
+                    yield return new()
                     {
                         AuthorName = _assistantId,
                         ConversationId = threadId,
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
 using System.Threading;
@@ -196,14 +197,16 @@ public static class ChatResponseExtensions
             int start = 0;
             while (start < contents.Count - 1)
             {
-                // We need at least two TextContents in a row to be able to coalesce.
-                if (contents[start] is not TContent firstText)
+                // We need at least two TextContents in a row to be able to coalesce. We also avoid touching contents
+                // that have annotations, as we want to ensure the annotations (and in particular any start/end indices
+                // into the text content) remain accurate.
+                if (!TryAsCoalescable(contents[start], out var firstText))
                 {
                     start++;
                     continue;
                 }
 
-                if (contents[start + 1] is not TContent secondText)
+                if (!TryAsCoalescable(contents[start + 1], out var secondText))
                 {
                     start += 2;
                     continue;
@@ -215,7 +218,7 @@ public static class ChatResponseExtensions
                 _ = coalescedText.Clear().Append(firstText).Append(secondText);
                 contents[start + 1] = null!;
                 int i = start + 2;
-                for (; i < contents.Count && contents[i] is TContent next; i++)
+                for (; i < contents.Count && TryAsCoalescable(contents[i], out TContent? next); i++)
                 {
                     _ = coalescedText.Append(next);
                     contents[i] = null!;
@@ -229,6 +232,18 @@ public static class ChatResponseExtensions
                 newContent.AdditionalProperties = firstText.AdditionalProperties?.Clone();
 
                 start = i;
+
+                static bool TryAsCoalescable(AIContent content, [NotNullWhen(true)] out TContent? coalescable)
+                {
+                    if (content is TContent && (content is not TextContent tc || tc.Annotations is not { Count: > 0 }))
+                    {
+                        coalescable = (TContent)content;
+                        return true;
+                    }
+
+                    coalescable = null!;
+                    return false;
+                }
             }
 
             // Remove all of the null slots left over from the coalescing process.
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationAnnotation.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationAnnotation.cs
@@ -1,0 +1,61 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents an annotation that links content to source references,
+/// such as documents, URLs, files, or tool outputs.
+/// </summary>
+public class CitationAnnotation : AIAnnotation
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CitationAnnotation"/> class.
+    /// </summary>
+    public CitationAnnotation()
+    {
+    }
+
+    /// <summary>
+    /// Gets or sets the location within the source that contains the cited content.
+    /// </summary>
+    public CitationSourceLocation? Location { get; set; }
+
+    /// <summary>
+    /// Gets or sets the title or name of the source.
+    /// </summary>
+    /// <remarks>
+    /// This could be the title of a document, a title from a web page, a name of a file, or similarly descriptive text.
+    /// </remarks>
+    public string? Title { get; set; }
+
+    /// <summary>
+    /// Gets or sets a URI from which the source material was retrieved.
+    /// </summary>
+    public Uri? Url { get; set; }
+
+    /// <summary>Gets or sets a source identifier associated with the annotation.</summary>
+    /// <remarks>
+    /// This is a provider-specific identifier that can be used to reference the source material by
+    /// an ID. This may be a document ID, or a file ID, or some other identifier for the source material
+    /// that can be used to uniquely identify it with the provider.
+    /// </remarks>
+    public string? FileId { get; set; }
+
+    /// <summary>Gets or sets the name of any tool involved in the production of the associated content.</summary>
+    /// <remarks>
+    /// This might be a function name, such as one from <see cref="AITool.Name"/>, or the name of a built-in tool
+    /// from the provider, such as "code_interpreter" or "file_search".
+    /// </remarks>
+    public string? ToolName { get; set; }
+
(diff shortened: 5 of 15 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
```diff
@@ -0,0 +1,55 @@
...
+    public int? StartIndex { get; set; }
+
+    /// <summary>
+    /// Gets or sets the end character index (exclusive) of the annotated span in the <see cref="AIContent"/>.
+    /// </summary>
+    public int? EndIndex { get; set; }
+
+    /// <summary>
+    /// Gets or sets an optional placeholder or marker text that was present in the original content
+    /// to indicate the citation (e.g., "〖4:0+source.json〗" or "[1]").
```
SteveSandersonMS: Is the expectation that the marker text has already been removed from the original content by the `IChatClient` implementation? The word `was` in this text suggests it's been removed, but if so it could be more explicit.

Asking because, as somone seeing the approach that OpenAI uses to represent these annotations for the first time, it's a bit mysterious and strange - I can't quite tell what the point of the marker text is if the start/end indices duplicate that information. As an app developer I can't guess whether I need to write some logic that processes the strings to remove the marker texts or not.

Is this a new feature from OpenAI? Googling for `openai annotations example` and `openai annotations syntax` and `openai message annotations` I'm mostly just getting links to forum posts from a year ago. https://platform.openai.com/docs/api-reference/chat/create doesn't show any examples of annotations. https://platform.openai.com/docs/assistants/deep-dive#message-annotations just shows some Python code that processes annotations, without explicitly showing examples of the before/after strings.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
```diff
@@ -0,0 +1,26 @@
...
+public class CitationSourceLocation
+{
+    /// <summary>Gets or sets the page at which the cited material begins.</summary>
+    public int? PageStart { get; set; }
+
+    /// <summary>Gets or sets the page at which the cited material ends.</summary>
+    public int? PageEnd { get; set; }
+
+    /// <summary>Gets or sets the block or chunk at which the cited material begins.</summary>
+    public int? BlockStart { get; set; }
```
SteveSandersonMS: Are these relative to the page or absolute? Zero or one based?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    /// <summary>Gets or sets the block or chunk at which the cited material ends.</summary>
+    public int? BlockEnd { get; set; }
+
+    /// <summary>Gets or sets the character at which the cited material begins.</summary>
+    public int? CharacterStart { get; set; }
```
SteveSandersonMS: Are these relative to the block or absolute? Zero or one based?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
```diff
@@ -0,0 +1,26 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a location within source material used for an annotation.</summary>
+public class CitationSourceLocation
+{
+    /// <summary>Gets or sets the page at which the cited material begins.</summary>
+    public int? PageStart { get; set; }
```
SteveSandersonMS: Are these zero or one based?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    /// <summary>Gets or sets the block or chunk at which the cited material ends.</summary>
+    public int? BlockEnd { get; set; }
+
+    /// <summary>Gets or sets the character at which the cited material begins.</summary>
+    public int? CharacterStart { get; set; }
+
+    /// <summary>Gets or sets the character at which the cited material ends.</summary>
+    public int? CharacterEnd { get; set; }
+}
```
SteveSandersonMS: This is a tricky area. I know that if this is the data the underlying providers provide, we have to expose it in the same shape. But I feel for the poor app devs who have to somehow make a UI function on top of it - that's a lot of different combinations of data that might/might not be provided, and hard to interpret that info (e.g., what actually is a block, will the set of provided numbers change from call to call, etc.)

Do you think there will be enough consistency across providers to make this useful? If not I guess another option is to rely on people using `AIAnnotation.RawRepresentation` to get more provider-specific info.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
```diff
@@ -0,0 +1,26 @@
...
+    public int? PageStart { get; set; }
+
+    /// <summary>Gets or sets the page at which the cited material ends.</summary>
+    public int? PageEnd { get; set; }
+
+    /// <summary>Gets or sets the block or chunk at which the cited material begins.</summary>
+    public int? BlockStart { get; set; }
+
+    /// <summary>Gets or sets the block or chunk at which the cited material ends.</summary>
+    public int? BlockEnd { get; set; }
```
SteveSandersonMS: Inclusive or exclusive?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
```diff
@@ -203,18 +203,53 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
...
+                    // into the model (MessageContent.FromXx does not support providing annotations), so they end up being one way and are dropped
+                    // on subsequent requests.
+                    if (mcu.TextAnnotation is { } tau)
+                    {
+                        string? fileId = null;
+                        string? toolName = null;
+                        if (!string.IsNullOrWhiteSpace(tau.InputFileId))
+                        {
+                            fileId = tau.InputFileId;
+                            toolName = "file_search";
```
SteveSandersonMS: Just curious, what do you expect developers to do with this tool name information? If we chose not to set it in these cases, would there be any drawback?

Asking because it feels like annotations could have file IDs even if they came from other tools, but maybe that's a problem for the future and not for now.

SteveSandersonMS (approved):
Looks good!

I added some questions/comments about bits where I think the meaning may be hard to interpret but the broad concept makes sense and feels valuable.

> Note that there's a variety of things OpenAI doesn't support or expose that others do, e.g. Google, Anthropic, and AWS all support including snippets of the cited content; I've included such things here, as well.

That sounds wise to me. Including a "snippet" is extremely helpful to the app developer - maybe OpenAI (or its client library) will add that later. It's good we're not restricted to a pure lowest-common-denominator set of information and instead are modelling what's common across a broad range of providers.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
stephentoub: I think I'll just remove the placeholder. It can be added later if necessary, but I believe it's only relevant for assistants, not for chat completion nor responses, and I've not come across anyone else that uses this mechanism. And on further investigation, it appears to be duplicative, because the start/end indices just point to the start/end of that placeholder in the text.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CitationSourceLocation.cs
stephentoub: There's enough ambiguity here that I'll just remove the source location stuff for now. We can always add it later.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
stephentoub: Just to be able to distinguish the source of the file ID, if it matters. InputFileId is documented to come from file_search.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
rogerbarreto, line 1: I suggest removing the `StartIndex` and `EndIndex` from the `AIAnnotation` as this is very exclusive for annotations for text contents, maybe we could have a `TextAnnotation : AIAnnotation` for those.

This typically applies only to text or sequential data where positions are linear and indexable (e.g., character offsets in a string). For non-sequential or non-linear content, these properties are impractical or insufficient. 

For example:

- In images, you'd use coordinates (e.g., x/y positions) instead of indices.
- In videos, timestamps or frame numbers would be more appropriate.
- In binary data, byte offsets might work for low-level annotations, but higher-level locators (e.g., section identifiers in a file format) are often better.
- For URLs or references, no positional indices are needed at all; annotations might reference the entire entity or use anchors (e.g., HTML fragment identifiers like #section1).

Forcing StartIndex and EndIndex on non-text content could lead to awkward workarounds, like serializing everything into a string representation, which loses fidelity.

rogerbarreto: Might worth having an `string Id { get; set; }` for the `AIAnnotation` abstract level. 

i.e: "ann-123".

rogerbarreto:
Worth checking the [W3C Annotation Model](https://www.w3.org/TR/annotation-model/) 

Could be helpful on defining the specialized annotations which in this document are named like `Selectors`

- TextPositionSelector
- SvgSelector
- FragmentSelector
- CssSelector
- XPathSelector
- RangeSelector ...

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
stephentoub: > Might worth having an `string Id { get; set; }` for the `AIAnnotation` abstract level.
> 
> i.e: "ann-123".

Do the underlying services have anything similar? I'm not seeing that in any.

stephentoub: > I suggest removing the StartIndex and EndIndex from the AIAnnotation as this is very exclusive for annotations for text contents

I'm torn on this.

On the one hand, you're right, annotations could end up applying to other kinds of content, which would make these properties a bit out of place. Right now I have the Anntotations property on TextContent, so these are only applicable to text, but the idea is that it could be moved to the base type in the future, and that wouldn't be a breaking change.

On the other hand, the 100% case right now is that it applies to text, and separating out the indices into a type hierarchy makes them a bit harder to use. And if we did do that, it's weird for the property to the be on TextContent, because it's obviously intended to be applicable to all content, so then we would want to move the Annotations property down to the base now, and yet that makes it so that other content has things that right now are never applicable.

I'll think about it.

stephentoub: @SteveSandersonMS, any opinion on this?

stephentoub:
I've addressed the feedback.

@rogerbarreto and @SteveSandersonMS, could you both please take another look? Including at my naming choices as part of addressing Roger's feedback.

@eiriktsarpalis, can you also review please?

Thanks.

SteveSandersonMS:
@stephentoub Looks good to me. I see you've ended up going for the most general design. Given that it's a relatively advanced feature, that seems a reasonable tradeoff even if people have to do a bit of casting to access character indices. Moving annotations to `AIContent` sounds good too if you think that's where it would end up going in the long run anyway.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/TextSpanAnnotatedRegion.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Diagnostics;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Describes a location in the associated <see cref="AIContent"/> based on starting and ending character indices.</summary>
+/// <remarks>This <see cref="AnnotatedRegion"/> typically applies to <see cref="TextContent"/>.</remarks>
+[DebuggerDisplay("[{StartIndex}, {EndIndex})")]
+public sealed class TextSpanAnnotatedRegion : AnnotatedRegion
```
eiriktsarpalis, line 12: This feels like it could have been a readonly struct.

eiriktsarpalis: Nvm, just saw that this is part of a hierarchy.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AIAnnotation"/> class.
+    /// </summary>
+    public AIAnnotation()
+    {
+    }
+
+    /// <summary>Gets or sets the target region for the annotation, pointing to where in the associated <see cref="AIContent"/> this annotation applies.</summary>
+    /// <remarks>
+    /// The most common form of <see cref="AnnotatedRegion"/> is <see cref="TextSpanAnnotatedRegion"/>, which provides starting and ending character indices
```
eiriktsarpalis, line 25: What are some others that could be added in the future? Is there any chance that annotation kinds are tied to particular types of annotations, and if so, would we want to make these properties of the derived annotations instead?

stephentoub: > What are some others that could be added in the future?

This is all hypothetical, as to my knowledge no one does this. But you could imagine things like "this region of an image is a copy of the image at http://whatever" or "these timecodes in the audio are based on the file at blah".

> would we want to make these properties of the derived annotations instead?

For better or worse, this design is separating details about the source material from what's being annotated. The derived annotation (e.g. CitationAnnotation) provides details about the source material, whereas this AnnotatedRegion-derived type describes what portion of the generated content it applies to. It's a bit overengineered right now, because right now the AI services only apply annotations to text, which is why the original design just had these StartIndex/EndIndex properties directly on AIAnnotation. @rogerbarreto expressed concern that we'd regret that in the future, so I split it out into this AnnotatedRegion, where derived types can provide different data for different kinds of content. It's not ideal, but I also don't have a better answer... suggestions?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// for <see cref="TextContent"/>.
+    /// </remarks>
+    public AnnotatedRegion? AnnotatedRegion { get; set; }
```
rogerbarreto: I like this approach more, adds the multi modal flexibility for annotations.

Suggestions:

1. I believe we might cases where we have multiple selections for the same annotation, 80% > will be single, which I think we can have a helper `property` like we do for other abstractions. 

    i.e `TextPatternAnnotatedSpan` pointing to different places in the content where the given `term`, applies.

2. Maybe `Span` or `Selection` instead of `Region` ?
   My small concern with `Region` can be misleading to understand at the get go.

   IMHO (0-5 score)
   
| Content | Span | Selection | Region |
| -- | -- | -- | -- |
| Video | 5 ✅ | 2 | 3 |
| Audio | 5 ✅ | 2 | 1 |
| Text | 5 ✅ | 4 | 2 |
| Binary | 4 ✅ | 2 | 3 |

```suggestion
    public AnnotatedSpan? AnnotatedSpan { get; set; }
```

stephentoub: > I believe we might cases where we have multiple selections for the same annotation, 80% > will be single, which I think we can have a helper property like we do for other abstractions.

I'm worried we're continuing to make the API more and more complicated to handle future possibilities.

Are there APIs you're thinking of that do this?

> Maybe Span or Selection instead of Region ?

Here's o3's analysis in response to "I'm adding an API that describes where an annotation/citation applies to various kinds of content. I'm deciding between using the term "span", "selection", and "region". Content might be text, video, audio, images, etc. Which term would be best?":

---
<html>
<body>
<html><head></head><body><p><strong>Short answer:</strong> use <strong><code inline="">region</code></strong>.</p>
<hr>
<h3>Why “region” scales cleanly across media types</h3>

Candidate | Common associations | Where it feels awkward | Hidden baggage
-- | -- | -- | --
span | Text offsets (“character span”, NLP “entity span”) and sometimes “temporal span” | Spatial areas in images/video (“a bounding-box span” sounds odd) | Overlaps with the HTML <span> element and with tracing/span concepts in distributed-tracing APIs
selection | Any user-highlighted area, clipboard APIs | Objects that were never user-selected (e.g., server-side auto-generated citations) | Connotes something ephemeral and UI-driven rather than a persisted addressable range
region | “Region of interest” (images/CV), “audio region” (DAWs), “time region” (video editors), “text region” (editors) | Rarely—“region of text” is still intelligible | Minimal; seldom used as a reserved keyword in popular web specs


<p>Because <strong>region</strong> already means “a contiguous, addressable area inside larger content” in graphics, audio, and video tool-chains, it maps naturally to the non-text modalities you care about. Meanwhile, developers will still parse “region.start_char … region.end_char” for text without surprise.</p>
<h3>Additional tips</h3>
<ol>
<li>
<p><strong>Define modality-specific coordinate schemes.</strong></p>
<pre><code class="language-jsonc">// Example: one shape with multiple coordinate systems
{
  "region": {
    "text": { "startChar": 120, "endChar": 145 },
    "image": { "x": 0.25, "y": 0.40, "width": 0.15, "height": 0.10 },
    "video": { "startTime": 12.5, "endTime": 13.8 }
  }
}
</code></pre>
</li>
<li>
<p><strong>Keep the noun stable, vary the qualifier.</strong></p>
<ul>
<li>
<p><code inline="">region.coordinates</code>, <code inline="">region.mimeType</code>, <code inline="">region.confidence</code>, etc.<br>
This preserves discoverability in IDE auto-completion.</p>
</li>
</ul>
</li>
<li>
<p><strong>Document plural behavior.</strong><br>
If an annotation can apply to multiple disjoint areas, prefer an array (<code inline="">regions: [...]</code>) rather than overloading one word.</p>
</li>
<li>
<p><strong>Consider a helper alias.</strong><br>
If many text-only clients exist, you can expose <code inline="">span</code> as a <em>convenience</em> field that simply mirrors <code inline="">region.text</code>, without letting it leak into core concepts.</p>
</li>
</ol>
<hr>
<p>By choosing <strong>region</strong> you minimize mental context-switching when developers move between text, image, audio, and video workflows, and you avoid colliding with well-known UI or HTML semantics.</p></body></html>
</body>
</html>

rogerbarreto: Region it is then. 😊

rogerbarreto (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs
stephentoub: >> I believe we might cases where we have multiple selections for the same annotation, 80% > will be single

If we change it now, we'll have `List<AnnotatedRegion>? AnnotatedRegions { get; set; }` and devs will end up writing code to accomodate multiple.
```C#
if (annotation.AnnotatedRegions is { Count: > 0 } regions)
    foreach (var region in regions)
        Use(region);
```
but it's more future proofed.

If we don't, devs will end up writing code that only handles a single one:
```C#
if (annotation.AnnotatedRegion is { } region)
    Use(region);
```
Simpler, but if in the future we were to add support for multiple, they wouldn't realize they were dropping regions (not great, but not the end of the world, either). If we needed to later add support for multiple, it'd also be a bit awkward, in that we'd end up with something like:
```C#
public AnnotatedRegion? AnnotatedRegion
{
    get => AnnotatedRegions is { Count: > 0 } regions ? regions[0] : null;
    set
    {
        if (value is null)
        {
            AnnotatedRegions = null; // throw away everything??
        }
        else if (AnnotatedRegions is null)
        {
            AnnotatedRegions = [value];
        }
        else if (AnnotatedRegions.Count == 0)
        {
            AnnotatedRegions.Add(value);
        }
        else
        {
            AnnotatedRegions[0] = value;
        }
    }
}

public IList<AnnotatedRegion>? AnnotatedRegions { get; set; }
```
which "works" but feels very hacky.

I'm torn, but if we really think there will be multiple, and if we're not comfortable transforming such cases into duplicated AIAnnotation instances each with their region, it's probably better to make it more flexible now. If it's unlikely and/or we're comfortable with such a transformation, I'd rather not absorb the complexity.

@rogerbarreto, are you aware of any services that already or plan to support the same annotation/citation on multiple regions?

@SteveSandersonMS, any opinion?

rogerbarreto: @stephentoub Actually after looking better into the [W3C Annotation selectors](https://www.w3.org/TR/annotation-model/#text-quote-selector), I'm more convinced that we should consider something more like the `Selectors` approach.

For example, a `AnnotatedTextPosition` is a unique and fixed place in the target document while one `AnnotatedQuote` or `AnnotatedFragment` although being just one element by its own definition points to multiple places once parsed.

stephentoub: > I'm more convinced that we should consider something more like the Selectors approach.

Meaning:
```C#
public IList<AnnotatedRegion>? AnnotatedRegions { get; set; }
```
?

rogerbarreto: > Meaning

Yes, I think this make's sense as we can have multiple specialized AnnotatedRegion, to start with the `TextSpanAnnotatedRegion` as the proposal here.

Issue #6614, opened 2025-07-16
[API Proposal]: Add HostedCodeInterpreterTool Features


rogerbarreto:
### Background and motivation

Currently the existing `HostedCodeInterpreterTool` type is lacking the concept of resources/files.

As we see an increasing adoption of the Code Interpretation Tooling support most notably by Agentic API's from different providers like below, seems reasonable to bring this to the abstraction level.

- OpenAI Assistants (https://platform.openai.com/docs/assistants/tools/code-interpreter)
- Azure Persistent Agents (https://learn.microsoft.com/en-us/dotnet/api/overview/azure/ai.agents.persistent-readme?view=azure-dotnet#create-message-with-code-interpreter-attachment)
- Amazon Bedrock (https://docs.aws.amazon.com/bedrock/latest/userguide/agents-test-code-interpretation.html)

### API Proposal

Start as immutable files/resources for minimal API public exposure.

Using the therm "File" which seems to be common so far.

```csharp
public class HostedCodeInterpreterTool : AITool
{
    /// <summary>Initializes a new instance of the <see cref="HostedCodeInterpreterTool"/> class.</summary>
    public HostedCodeInterpreterTool(IReadOnlyCollection<AIContent> files)
    {
         Files = files;
    }

    public IReadOnlyCollection<AIContent> Files { get; }
}
```

Additional `HostedFile` Content
```
public class HostedFileContent : AIContent 
{
    // The Id is required for any hosted file
    public HostedFileContent (string id)
    {
        Id = id;
    }

    // Represents the fully qualified identifier for the hosted file to be found
    public string Id { get; set; }
}
```

### API Usage

End user usage.
```csharp
// Preparing to invoke a ChatClient 
var files = 
[
    new UrlContent("https://public-facing-resource"),
    new HostedFileContent("Hosted file UID"),
    new DataContent(fileByteArray, "text/csv")
];

chatOptions.Tools = [new HostedCodeInterpreterTool(files)];
```

Library implementers usage
```csharp
// Within a ChatClient GetResponse/GetResponseStream impl.

// Can be multiple code tools, but for sample brevity, using single
var hciTool = chatOptinos.Tools.Single(tool => tool is HostedCodeInterpreterTool);
foreach (AIContent file in hciTool.Files)
{
    switch (file) 
    {
        case UrlContent urlContent:
            // providers logic
            break;
        case HostedFileContent hostedFileContent:
            // providers logic
            break;
        case DataContent dataContent:
[...]

PR #6620, opened 2025-07-16
Add HostedFile/VectorStoreContent, HostedFileSearchTool, and HostedCodeInterpreterTool.Inputs

stephentoub:
Closes https://github.com/dotnet/extensions/issues/6614

cc: @rogerbarreto 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6620)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
```diff
@@ -300,7 +300,7 @@ internal static FunctionToolDefinition ToOpenAIAssistantsFunctionToolDefinition(
                             break;
 
                         case HostedCodeInterpreterTool:
-                            runOptions.ToolsOverride.Add(new CodeInterpreterToolDefinition());
+                            runOptions.ToolsOverride.Add(ToolDefinition.CreateCodeInterpreter());
                             break;
```
rogerbarreto: How the hostedFileContent would be used by the client.

```suggestion
                            var toolDefinition = ToolDefinition.CreateCodeInterpreter();
                            runOptions.ToolsOverride.Add(toolDefinition);
                            if (tool.Inputs.Count != 0)
                            {
                                var threadInitializationMessage = new ThreadInitializationMessage(OpenAI.Assistants.MessageRole.User, [OpenAI.Assistants.MessageContent.FromText("attachments")]);
                                foreach (var input in tool.Inputs)
                                {
                                    if (input is HostedFileContent hostedFile)
                                    {                                     
                                        threadInitializationMessage.Attachments.Add(new(hostedFile.Id, [toolDefinition]));

                                    }
                                }
                                runOptions.AdditionalMessages.Add(threadInitializationMessage);
                            }

                            break;
```

rogerbarreto:
# Proposal: CodeExecutionResultContent

## Context

The Microsoft.Extensions.AI.Abstractions library needed a unified `AIContent` subclass to represent code execution results from various AI providers. This content type is used in both `ChatMessage.Contents` (non-streaming) and `ChatResponseUpdate.Contents` (streaming) scenarios to provide a consistent abstraction across different provider implementations.

## Decision

Implemented `CodeExecutionResultContent` as a minimal, provider-agnostic abstraction that focuses only on properties proven to be common across multiple AI providers, while leveraging existing `AIContent` types for specialized use cases like file handling.

## Design Investigation & Rationale

### Provider Research Findings

Investigated code execution response formats from four major AI providers:

#### 1. Anthropic Claude
**Response Format:**
```json
{
  "role": "assistant",
  "container": {
    "id": "container_011CPR5CNjB747bTd36fQLFk",
    "expires_at": "2025-05-23T21:13:31.749448Z"
  },
  "content": [
    {
      "type": "text",
      "text": "I'll calculate the mean and standard deviation for you."
    },
    {
      "type": "server_tool_use",
      "id": "srvtoolu_01A2B3C4D5E6F7G8H9I0J1K2",
      "name": "code_execution",
      "input": {
        "code": "import numpy as np\ndata = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]\nmean = np.mean(data)\nstd = np.std(data)\nprint(f\"Mean: {mean}\")\nprint(f\"Standard deviation: {std}\")"
      }
    },
    {
      "type": "code_execution_tool_result",
      "tool_use_id": "srvtoolu_01A2B3C4D5E6F7G8H9I0J1K2",
      "content": {
        "type": "code_execution_result",
        "stdout": "Mean: 5.5\nStandard deviation: 2.8722813232690143\n",
        "stderr": "",
        "return_code": 0,
        "content": [
           {
              "file_id": "generated_file_id",
              "type": "code_execution_output"
           }
        ]
      }
    },
    {
      "type": "text",
      "text": "The mean of the dataset is 5.5 and the standard deviation is approximately 2.87."
    }
  ],
  "id": "msg_01BqK2v4FnRs4xTjgL8EuZxz",
  "model": "claude-opus-4-20250514",
  "stop_reason": "end_turn",
  "usage": {
    "input_tokens": 45,
    "output_tokens": 187,
  }
}
```
**Key Features:**
- ✅ Provides `stdout`, `stderr`, `return_code`
- ✅ Returns executed code
- ✅ Supports streaming via Server-Sent Events
- ✅ Supports built-in file generation

#### 2. Google Vertex AI/Gemini
**Response Format:**
```json
{
  "executableCode": {
    "language": "PYTHON",
    "code": "total = 0\nfor i in range(1, 11):\n    total += i\nprint(f'{total=}')\n"
  },
  "codeExecutionResult": {
    "outcome": "OUTCOME_OK",
    "output": "total=55\n"
  }
}
```
**Key Features:**
- ✅ Provides executed code and output
- ✅ Uses `outcome` enum instead of numeric exit codes
- ✅ Supports inline images: "The output files are returned as inline images in the response"
- ✅ Supports streaming via `streamGenerateContent`

#### 3. OpenAI Code Interpreter (Assistants API)
**Response Format:**
- Uses Assistants API pattern with tool calls and results
- ✅ Provides execution output
- ✅ Does NOT return executed code
- ✅ Supports generated files with file IDs
- ✅ Supports streaming for code execution

**Response Format:**
```json
{
  "id": "msg_abc123",
  "object": "thread.message",
  "created_at": 1698983503,
  "thread_id": "thread_abc123",
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": {
        "value": "Hi! How can I help you today?",
        "annotations": [
           {
              "type": "file_path",
              "end_index": "integer",
              "start_index": "integer",
              "file_path": 
              { 
                 "file_id": "string" 
              }
           }
        ]
      }
    }
  ],
  "assistant_id": "asst_abc123",
  "run_id": "run_abc123",
  "attachments": 
  [
     {
        "file_id", "string",
        "tools": [{ "type": "code_interpreter" }]
     }
  ],
  "metadata": {}
}
```

**Streaming Response Format:**
https://platform.openai.com/docs/api-reference/assistants-streaming/run-step-delta-object
```json
{
  "id": "step_123",
  "object": "thread.run.step.delta",
  "delta": {
    "step_details": {
      "type": "tool_calls",
      "tool_calls": [
        {
          "index": 0,
          "id": "call_123",
          "type": "code_interpreter",
          "code_interpreter": 
          { 
             "input": "string", 
             "outputs": 
             [
               { 
                  "type": "logs", 
                  "logs": "string" 
               },
               { 
                  "type": "image", 
                  "image": { "file_id", "string" }
               }
            ]
          }
        }
      ]
    }
  }
}

```

#### 3b. OpenAI Responses API (NEW)
**Response Format:**
```json
{
  "type": "code_interpreter_call",
  "outputs": [
    {
      "type": "output_text",
      "text": "Calculation result: 42"
    },
    {
      "type": "output_image",
      "image": "base64_encoded_image_data"
    }
  ]
}
```
**Streaming Events:**
```
response.code_interpreter_call.in_progress
response.code_interpreter_call_code.delta  // Streams executed code
response.code_interpreter_call_output.delta // Streams execution output
```
**Key Features:**
- ✅ Provides execution output via `output_text` and `output_image`
- ✅ **STREAMS executed code** via `code.delta` events
- ✅ **STREAMS execution output** via `output.delta` events
- ✅ Supports generated files and images
- ✅ **Full streaming support** for code execution

#### 4. Azure AI Foundry
**Response Format:**
- Follows OpenAI Responses API pattern (supports both Assistants and Responses APIs)
- ✅ Supports "file path annotations" and generated files
- ✅ **STREAMS executed code** (via Responses API)
- ✅ **STREAMS execution output** (via Responses API)
- ✅ **Full streaming support** for code execution (via Responses API)

### Decision-Making Process

Based on this analysis, applied the principle of **minimal proven abstractions**:

#### Properties INCLUDED (Multi-Provider Support):
- **`Stdout`**: Supported by Anthropic (`stdout`), Google Vertex AI (`output`), OpenAI/Azure (execution output)
- **`Stderr`**: Supported by Anthropic (`stderr`), others use similar error patterns
- **`GeneratedContents`**: Supported by Google Vertex AI (inline images), Azure AI Foundry (file annotations), OpenAI (file references)

#### Properties EXCLUDED (Not Proven Across Multiple Providers):
- **`Code`**: Only Google Vertex AI and OpenAI Responses API return executed code (2/5 provider implementations, but OpenAI has two different APIs)
- **`ExitCode`**: Only Anthropic Claude uses `return_code` (1/5 provider implementations)
- **`Language`**: Provider-specific metadata
- **`ExecutionTimeMs`**: Provider-specific metadata
- **`ErrorCode`/`ErrorMessage`**: Provider-specific error handling

**Note**: While OpenAI Responses API does stream executed code, the `Code` property was excluded because:
1. OpenAI has two different APIs (Assistants vs Responses) with different capabilities
2. The executed code is primarily useful for debugging/transparency rather than core functionality
3. The abstraction focuses on execution *results* rather than the code itself
4. Providers can include executed code in `GeneratedContents` if needed

### Streaming Investigation

**Providers with Streaming Support:**
- ✅ **Anthropic Claude**: Full streaming with fine-grained tool streaming
- ✅ **Google Vertex AI**: Real-time streaming with incremental processing
- ✅ **OpenAI Responses API**: Full streaming including executed code and output
- ✅ **Azure AI Foundry**: Full streaming support (via Responses API)
- ✅ **OpenAI Assistants API**: Full streaming support (via delta)

**Streaming Patterns Identified:**
1. **Incremental Output Streaming**: Real-time stdout/stderr updates
2. **Complete Result Streaming**: Single update with full execution results
3. **Mixed Content Streaming**: Code execution mixed with text explanations

## Final Design Implementation

### CodeExecutionResultContent Class

```csharp
// Licensed to the .NET Foundation under one or more agreements.

using System;
using System.Diagnostics;
using System.Diagnostics.CodeAnalysis;

/// <summary>
/// Represents the result of code execution in a chat.
/// </summary>
/// <remarks>
/// This content type represents the core output from executing code in various AI provider environments,
/// focusing on the essential properties: stdout, stderr, and generated content.
/// It provides a minimal unified format that accommodates the common execution result patterns
/// from providers such as Anthropic Claude, Azure AI Foundry, OpenAI Code Interpreter, and Google Vertex AI.
/// Generated files are represented using existing AIContent types like DataContent, UriContent, or HostedFileContent.
/// </remarks>
[DebuggerDisplay("{DebuggerDisplay,nq}")]
public sealed class CodeExecutionResultContent : AIContent
{
    /// <summary>The standard output from code execution.</summary>
    private string? _stdout;

    /// <summary>The standard error output from code execution.</summary>
    private string? _stderr;

    /// <summary>
    /// Initializes a new instance of the <see cref="CodeExecutionResultContent"/> class.
    /// </summary>
    /// <param name="stdout">The standard output from execution.</param>
    /// <param name="stderr">The standard error output from execution.</param>
    /// <param name="generatedContents">Collection of content generated during execution (images, files, etc.).</param>
    public CodeExecutionResultContent(string? stdout = null, string? stderr = null, IList<AIContent>? generatedContents = null)
    {
        _stdout = stdout;
        _stderr = stderr;
        GeneratedContents = generatedContents;
    }

    /// <summary>
    /// Initializes a new instance of the <see cref="CodeExecutionResultContent"/> class for JSON deserialization.
    /// </summary>
    /// <param name="stdout">The standard output from execution.</param>
    /// <param name="stderr">The standard error output from execution.</param>
    /// <param name="generatedContents">Collection of content generated during execution.</param>
    [JsonConstructor]
    public CodeExecutionResultContent(
        string? stdout = null,
        string? stderr = null,
        IList<AIContent>? generatedContents = null)
    {
        _stdout = stdout;
        _stderr = stderr;
        GeneratedContents = generatedContents;
    }

    /// <summary>
    /// Gets or sets the standard output from code execution.
    /// </summary>
    /// <remarks>
    /// This contains the text output that would normally be written to stdout during code execution,
    /// such as print statements, successful computation results, etc.
    /// </remarks>
    [AllowNull]
    public string Stdout
    {
        get => _stdout ?? string.Empty;
        set => _stdout = value;
    }

    /// <summary>
    /// Gets or sets the standard error output from code execution.
    /// </summary>
    /// <remarks>
    /// This contains error messages, warnings, and other diagnostic information that would
    /// normally be written to stderr during code execution.
    /// </remarks>
    [AllowNull]
    public string Stderr
    {
        get => _stderr ?? string.Empty;
        set => _stderr = value;
    }

    /// <summary>
    /// Gets or sets the collection of content generated during code execution.
    /// </summary>
    /// <remarks>
    /// This includes images, charts, data files, and other artifacts created by the executed code.
    /// Generated content is represented using existing AIContent types such as DataContent for inline data,
    /// UriContent for downloadable files, or HostedFileContent for provider-hosted files.
    /// This property is supported by multiple providers including Google Vertex AI (inline images),
    /// Azure AI Foundry (file path annotations), and OpenAI Code Interpreter (file references).
    /// </remarks>
    public IList<AIContent>? GeneratedContents { get; set; }

    /// <summary>
    /// Gets the combined text from code execution, including both stdout and stderr.
    /// </summary>
    /// <remarks>
    /// This property combines the standard output and standard error streams into a single string.
    /// If both stdout and stderr have content, stderr is appended after stdout.
    /// If only one stream has content, only that content is returned.
    /// </remarks>
    [JsonIgnore]
    public string Text
    {
        get
        {
            var hasStdout = !string.IsNullOrEmpty(Stdout);
            var hasStderr = !string.IsNullOrEmpty(Stderr);

            if (hasStdout && hasStderr)
            {
                return $"{Stdout}\n{Stderr}";
            }

            if (hasStdout)
            {
                return Stdout;
            }

            if (hasStderr)
            {
                return Stderr;
            }
            
            return string.Empty;
        }
    }

    /// <summary>
    /// Gets a value indicating whether the code execution was successful.
    /// </summary>
    /// <remarks>
    /// This is determined by checking if there are no errors in stderr.
    /// </remarks>
    [JsonIgnore]
    public bool IsSuccess => string.IsNullOrEmpty(Stderr);

    /// <inheritdoc/>
    public override string ToString() => Text;

    /// <summary>Gets a string representing this instance to display in the debugger.</summary>
    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
    private string DebuggerDisplay
    {
        get
        {
            var status = IsSuccess ? "Success" : "Failed";
            var output = !string.IsNullOrEmpty(Text) ? $", Text: \"{Text.Substring(0, Math.Min(50, Text.Length))}\"" : string.Empty;
            var generated = GeneratedContents?.Count > 0 ? $", Generated: {GeneratedContents.Count}" : string.Empty;

            return $"CodeExecution = {status}{output}{generated}";
        }
    }
}
```

### AIContent.cs JsonDerivedType Addition

```csharp
/// <summary>Represents content used by AI services.</summary>
[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
[JsonDerivedType(typeof(CodeExecutionResultContent), typeDiscriminator: "codeExecutionResult")]
[JsonDerivedType(typeof(DataContent), typeDiscriminator: "data")]
[JsonDerivedType(typeof(ErrorContent), typeDiscriminator: "error")]
// ... other existing types
public class AIContent
{
    // ... existing implementation
}
```

### Text Property Rationale

The `Text` property combines stdout and stderr into a single string because:
1. **Simplifies consumption**: `ToString()` can simply return `Text`
2. **Avoids naming confusion**: "Output" implied only stdout, "Text" is neutral
3. **Consistent with other AIContent types**: Similar to how `TextContent` has a `Text` property
4. **Streaming friendly**: Allows incremental text updates in streaming scenarios

## Consumer Usage Examples

### Non-Streaming Scenario

```csharp

IChatClient client = // ... initialize your provider client

// Request code execution
var response = await client.GetResponseAsync([
    new ChatMessage(ChatRole.User, "Calculate the mean of [1, 2, 3, 4, 5] and create a chart")
]);

// Process the response
foreach (var content in response.Message.Contents)
{
    switch (content)
    {
        case TextContent textContent:
            Console.WriteLine($"AI Response: {textContent.Text}");
            break;

        case CodeExecutionResultContent codeResult:
            Console.WriteLine($"Code Execution Output: {codeResult.Text}");
            Console.WriteLine($"Success: {codeResult.IsSuccess}");

            // Handle generated files
            if (codeResult.GeneratedContents != null)
            {
                foreach (var generated in codeResult.GeneratedContents)
                {
                    switch (generated)
                    {
                        case DataContent dataContent:
                            Console.WriteLine($"Generated inline file: {dataContent.MediaType}");
                            // Save dataContent.Data to file
                            break;

                        case HostedFileContent hostedFile:
                            Console.WriteLine($"Generated hosted file: {hostedFile.FileId}");
                            // Download file using provider's file API
                            break;

                        case UriContent uriContent:
                            Console.WriteLine($"Generated file available at: {uriContent.Uri}");
                            // Download from URI
                            break;
                    }
                }
            }
            break;
    }
}
```

### Streaming Scenario

```csharp

IChatClient client = // ... initialize your provider client

// Request streaming code execution
await foreach (var update in client.GetStreamingResponseAsync([
    new ChatMessage(ChatRole.User, "Run a data analysis and generate visualizations")
]))
{
    foreach (var content in update.Contents)
    {
        switch (content)
        {
            case TextContent textContent:
                Console.Write(textContent.Text); // Stream text as it arrives
                break;

            case CodeExecutionResultContent codeResult:
                // Handle incremental execution output
                if (!string.IsNullOrEmpty(codeResult.Stdout))
                {
                    Console.Write($"[STDOUT] {codeResult.Stdout}");
                }

                if (!string.IsNullOrEmpty(codeResult.Stderr))
                {
                    Console.Write($"[STDERR] {codeResult.Stderr}");
                }

                // Handle generated content as it becomes available
                if (codeResult.GeneratedContents?.Count > 0)
                {
                    Console.WriteLine($"\n[FILES] Generated {codeResult.GeneratedContents.Count} files");
                    foreach (var generated in codeResult.GeneratedContents)
                    {
                        Console.WriteLine($"  - {generated.GetType().Name}");
                    }
                }
                break;
        }
    }
}
```

### Different Content Combinations

```csharp
// Example 1: Successful execution with output
var successResult = new CodeExecutionResultContent(
    stdout: "Calculation complete!\nMean: 3.0\nStandard deviation: 1.58",
    stderr: ""
);
Console.WriteLine(successResult.Text); // Prints the stdout
Console.WriteLine(successResult.IsSuccess); // True

// Example 2: Failed execution with error
var errorResult = new CodeExecutionResultContent(
    stdout: "Starting calculation...",
    stderr: "NameError: name 'undefined_variable' is not defined"
);
Console.WriteLine(errorResult.Text); // Prints: "Starting calculation...\nNameError: name 'undefined_variable' is not defined"
Console.WriteLine(errorResult.IsSuccess); // False

// Example 3: Execution with generated files
var resultWithFiles = new CodeExecutionResultContent(
    stdout: "Chart generated successfully",
    generatedContents: [
        new DataContent(chartImageBytes, "image/png"),
        new DataContent(csvData, "text/csv")
    ]
);
```

## Streaming Scenarios

Based on this investigation, identified three streaming patterns that the design accommodates:

### Pattern 1: Incremental Output Streaming (Anthropic/Google/OpenAI Responses Style)

Real-time streaming of stdout/stderr as code executes. **Supported by**: Anthropic Claude, Google Vertex AI, OpenAI Responses API, Azure AI Foundry:

```csharp
// Provider implementation for incremental streaming
public async IAsyncEnumerable<ChatResponseUpdate> StreamCodeExecution(string code)
{
    // Stream 1: Execution start
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new CodeExecutionResultContent(stdout: "Starting execution...\n")
    ]);

    // Stream 2: Incremental output
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new CodeExecutionResultContent(stdout: "Processing data...\n")
    ]);

    // Stream 3: More output
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new CodeExecutionResultContent(stdout: "Generating visualization...\n")
    ]);

    // Stream 4: Final result with generated content
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new CodeExecutionResultContent(
            stdout: "Execution complete!\nResults saved.\n",
            generatedContents: [
                new DataContent(chartImageBytes, "image/png"),
                new DataContent(csvData, "text/csv")
            ]
        )
    ]);
}
```

### Pattern 2: Complete Result Streaming (OpenAI Assistants Style)

Single update with complete execution results. **Supported by**: OpenAI Assistants API (legacy):

```csharp
// Provider implementation for complete result streaming
public async IAsyncEnumerable<ChatResponseUpdate> StreamCodeExecution(string code)
{
    // Execute code completely, then stream result
    var executionResult = await ExecuteCodeAsync(code);

    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new CodeExecutionResultContent(
            stdout: executionResult.CompleteOutput,
            stderr: executionResult.ErrorOutput,
            generatedContents: executionResult.GeneratedFiles?.Select(f =>
                new HostedFileContent(f.FileId)).ToList()
        )
    ]);
}
```

### Pattern 3: Mixed Content Streaming

Code execution mixed with explanatory text:

```csharp
// Provider implementation for mixed content streaming
public async IAsyncEnumerable<ChatResponseUpdate> StreamCodeExecution(string code)
{
    // Stream 1: AI explanation
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new TextContent("I'll run this code to analyze your data:")
    ]);

    // Stream 2: Code execution result
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new CodeExecutionResultContent(
            stdout: "Data analysis complete\nMean: 42.5, Std: 12.3",
            generatedContents: [
                new DataContent(chartBytes, "image/png")
            ]
        )
    ]);

    // Stream 3: Follow-up explanation
    yield return new ChatResponseUpdate(ChatRole.Assistant, [
        new TextContent("The results show a normal distribution with...")
    ]);
}
```

### Consumer Streaming Handling

Consumers can handle all streaming patterns uniformly:

```csharp
var allExecutionResults = new List<CodeExecutionResultContent>();
var allText = new StringBuilder();

await foreach (var update in client.GetStreamingResponseAsync(messages))
{
    foreach (var content in update.Contents)
    {
        switch (content)
        {
            case CodeExecutionResultContent codeResult:
                allExecutionResults.Add(codeResult);

                // Handle incremental output
                if (!string.IsNullOrEmpty(codeResult.Text))
                {
                    Console.Write(codeResult.Text);
                    allText.Append(codeResult.Text);
                }

                // Handle generated content
                if (codeResult.GeneratedContents?.Count > 0)
                {
                    ProcessGeneratedContent(codeResult.GeneratedContents);
                }
                break;

            case TextContent textContent:
                Console.Write(textContent.Text);
                break;
        }
    }
}

// Final combined result
var combinedResult = new CodeExecutionResultContent(
    stdout: allText.ToString(),
    generatedContents: allExecutionResults
        .SelectMany(r => r.GeneratedContents ?? [])
        .ToList()
);
```

## Benefits of This Design

1. **Minimal Surface Area**: Only includes properties proven across multiple providers
2. **Leverages Existing Abstractions**: Uses `DataContent`, `UriContent`, `HostedFileContent` for files
3. **Streaming Compatible**: Works with both incremental and complete-result streaming (4/5 provider implementations support streaming)
4. **Provider Agnostic**: Accommodates different provider response formats including both OpenAI APIs
5. **Future-Proof**: Design accommodates the evolution from OpenAI Assistants API to Responses API
6. **Extensible**: Can add properties later if they prove common across providers
7. **Type Safe**: Full polymorphic JSON serialization support
8. **Developer Friendly**: Simple `Text` property and `ToString()` for easy consumption

## Alternatives Considered

1. **Comprehensive Design**: Including all provider-specific properties (rejected due to complexity)
2. **Provider-Specific Subclasses**: Separate classes per provider (rejected due to fragmentation)
3. **Generic Content with Metadata**: Using `AdditionalProperties` for everything (rejected due to lack of type safety)
4. **Custom File Type**: New `CodeExecutionFile` class (rejected in favor of existing `AIContent` types)

## Future Considerations

- Monitor provider evolution for new common patterns
- Consider adding properties if they become proven across multiple providers
- Evaluate streaming optimizations based on real-world usage
- Assess need for provider-specific extensions through `AdditionalProperties`
```

**Comment by stephentoub** [comment]
@rogerbarreto:

> Stderr: Supported by Anthropic (stderr), others use similar error patterns

I only see error information from anthropic. How is it represented in the others? I also find it a bit odd calling it stdout/stderr, as that presumes a particular mode of execution.

Could error information not be conveyed using ErrorContent in the AIContent collection?

And could textual output not just be TextContent in the AIContent collection, along with any other generated output? Then a Text property could just do what ChatResponse/ChatMessage/etc. do, which is have Text just concat any TextContent in their collection.

That breeds:
```C#
public sealed class CodeExecutionResultContent : AIContent
{
    public IList<AIContent>? GeneratedContents { get; set; }

    [JsonIgnore]
    public string Text { get; }
}
```

Is that sufficient / sufficiently flexible?

> Code: Only Google Vertex AI and OpenAI Responses API return executed code (2/5 provider implementations, but OpenAI has two different APIs)

Presumably we'd just store the code as DataContent (with an appropriate mime type for language) in GeneratedContents? How would we handle streaming? I believe the code itself can be streamed as it's generated... would we have distinct DataContent objects for each part, or would we hold onto that content until we can reassemble it, ala function call arguments?

rogerbarreto:
> Could error information not be conveyed using ErrorContent in the AIContent collection?

I think this is a good point, StdErr can use the `ErrorContent` when available in the list instead of `string` or it could be a standalone `ErrorContent? Error` property. 

> And could textual output not just be TextContent in the AIContent collection, along with any other generated output? Then a Text property could just do what ChatResponse/ChatMessage/etc. do, which is have Text just concat any TextContent in their collection.
Is that sufficient / sufficiently flexible?

IMHO, too much flexible, we definitely could, my initial thought was that, since we have a dedicated type that allow us to make clear distinction between what is a `CodeExecution` data from what are `CodeArtifacs`. 

This extra flexible approach (using just GeneratedArtifacs) can be problematic if you want to iterate in the contents and need to know what is a code execution block from what is a file/artifact generated information which can happen more often in streaming handling code.

As an alternative we may have try `TextContent? Output { get; }` for this, thoughts?

> I believe the code itself can be streamed as it's generated... would we have distinct DataContent objects for each part, or would we hold onto that content until we can reassemble it, ala function call arguments?

Ideally as far I see in the OpenAI Responses, we have the tool_id, which I probably missed in the design, we should consider having an identifier for the tool, this can be very useful in streaming multiple code executions.

stephentoub:
> since we have a dedicated type that allow us to make clear distinction between what is a CodeExecution data from what are CodeArtifacs

I was imagining that would be TextContent vs everything else, e.g. there's a distinction between TextContent and DataContent(..., "text/plain"). TextContent would be the tool output, and everything else is a DataContent with an appropriate media type.

stephentoub:
@rogerbarreto, I've started looking at this more, and I'm seeing what look like discrepancies from the AI generated analysis you shared. OpenAI Assistants appears to provide not only the code that was generated but also any inputs generated to that code. Anthropic _does_ provide the generated code. Can you double-check some of your analysis?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
```diff
@@ -209,6 +209,44 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
...
+                        (codeInterpreterContent.Inputs ??= []).Add(new TextContent(input));
+                    }
+
+                    if (details.CodeInterpreterOutputs is { Count: > 0 })
+                    {
+                        foreach (var codeInterpreterOutput in details.CodeInterpreterOutputs)
+                        {
+                            if (codeInterpreterOutput.ImageFileId is not null)
+                            {
+                                (codeInterpreterContent.Outputs ??= []).Add(new HostedFileContent(codeInterpreterOutput.ImageFileId));
```
rogerbarreto: We might consider adding an optional/informative `MediaType`

```suggestion
                                (codeInterpreterContent.Outputs ??= []).Add(new HostedFileContent(codeInterpreterOutput.ImageFileId, "image/*"));
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
```diff
@@ -307,6 +345,7 @@ internal static FunctionToolDefinition ToOpenAIAssistantsFunctionToolDefinition(
                             break;
 
                         case HostedCodeInterpreterTool:
+                            // There's currently no API for providing the tool resources so that it can access tool.Inputs.
```
rogerbarreto: You are right in a per-request basis, but it does for the thread creation, we might add this when `ThreadId` is NOT provided.

rogerbarreto:
> Can you double-check some of your analysis?

I double checked, missed the details in the doc, they are quite hidden in the expandables. Provided an updated information.

rogerbarreto:
> I was imagining that would be TextContent vs everything else, e.g. there's a distinction between TextContent and DataContent(..., "text/plain"). TextContent would be the tool output, and everything else is a DataContent with an appropriate media type.

We can start with this assumption, so any generated content is a `DataContent`, including plain text.

rogerbarreto:
One thing to note after having to deal with the `FileSearch` is that we have now other hosted references like `VectorSearch` and `File` and for everything that is hosted the `Id` will be a commonality, given that we might consider having an `abstract class HostedContent { string? Id { get; set; } }`.

This adds 3 level hierarchy to the repo (FileHostedContent -> HostedContent -> AIContent), not sure how I feel about it though or we just use the `HostedContent` for both, for now. `HostedContent : AIContent`

Note: VectorSearch actually is a HostedService not a "Content", but we might consider the reference to it the "content", thoughts?

stephentoub:
> we just use the HostedContent for both, for now. HostedContent : AIContent

How would an implementation distinguish whether an incoming ID was for a file or a store?

> we might consider having an abstract class HostedContent { string? Id { get; set; } }.

In what situation would someone want to use the base type directly (if the only benefit of the base type is avoiding duplicating an Id property, it's not worth it, it needs to have some polymorphic use)?

> VectorSearch actually is a HostedService not a "Content", but we might consider the reference to it the "content", thoughts?

I'm not understanding the distinction being made. VectorSearch as a tool wouldn't be an AIContent, it'd be an AITool. And VectorStore as content is because the content is the ID / reference / whatever data is being passed around.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
stephentoub: What would someone do with that?

stephentoub:
>> I double checked, missed the details in the doc, they are quite hidden in the expandables. Provided an updated information.

> 3. OpenAI Code Interpreter (Assistants API)
> ❌ Does NOT return executed code

It does, actually. In the response where it provides "input" to the code execution tool, that's the code.

rogerbarreto:
> It does, actually. In the response where it provides "input" to the code execution tool, that's the code.

It was not clear for me if this was the input message (Natural language) for generating the code or the code itself to be executed.

rogerbarreto:
> How would an implementation distinguish whether an incoming ID was for a file or a store?

> In what situation would someone want to use the base type directly (if the only benefit of the base type is avoiding duplicating an Id property, it's not worth it, it needs to have some polymorphic use)?

If you are the caller, when dealing with the `HostedContent`, you shouldn't what is a `File` or a `Vector`. But as you also made a good point, I don't think this is a common situation, most of the time you want to know directly the type, doesn't worth having that abstraction.

> I'm not understanding the distinction being made. VectorSearch as a tool wouldn't be an AIContent, it'd be an AITool. And VectorStore as content is because the content is the ID / reference / whatever data is being passed around.

I mispelled "VectorSeach" for "VectorStore", agree, happy to go forward as VectorStore as content.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
```diff
@@ -0,0 +1,43 @@
...
+/// Represents a file that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="DataContent"/> which contains the data for a file or blob, this class represents a file that is hosted
+/// by the AI service and referenced by an identifier. Such identifiers are specific to the provider.
+/// </remarks>
+[DebuggerDisplay("FileId = {FileId}")]
+public sealed class HostedFileContent : AIContent
+{
+    private string? _fileId;
```
rogerbarreto: If we are not allowing null, the field shouldn't be nullable.

```suggestion
    private string _fileId;
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /// <summary>
+    /// Gets or sets the ID of the hosted file.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="value"/> is empty or composed entirely of whitespace.</exception>
+    public string FileId
+    {
+        get => _fileId ?? string.Empty;
```
rogerbarreto:
```suggestion
        get => _fileId;
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
stephentoub: Your attachment workaround makes this no longer necessary, right?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs
stephentoub: Neat. Have you tested this?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
rogerbarreto: The overall idea here is to optimize the caller code, allowing it to know ahead of time without having to download the file to know what is the mediaType, which we don't make a distinction in the Abstraction, but the json response does.

i.e: If the caller code just want to process the images generated by the code it doesn't need to download the non-image content and vice-versa.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileStoreContent.cs
```diff
@@ -0,0 +1,43 @@
...
+
+/// <summary>
+/// Represents a file store (such as a vector store) that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="HostedFileContent"/> which represents a specific file that is hosted by the AI service,
+/// <see cref="HostedFileStoreContent"/> represents a file store that can contain multiple files.
+/// </remarks>
+[DebuggerDisplay("FileStoreId = {FileStoreId}")]
+public sealed class HostedFileStoreContent : AIContent
```
rogerbarreto: I vote for keeping `VectorStore` in the name as this is a very conventional name today in the AI ecosystem or maybe a generalistic name like [KnowledgeStore](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_SessionState.html#bedrock-Type-agent-runtime_SessionState-knowledgeBaseConfigurations) (inspired by Google's naming) , unless we have many different fileStore options (other than VectorStore) scenarios you are thinking one using as a `FileStore`?

This name can be a bit conflicting with the `HostedFile` vs `HostedFileStore`, at first I thought was a typo in the implementation, had to double check.

```suggestion
public sealed class HostedVectorStoreContent : AIContent
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CodeExecutionResultContent.cs
rogerbarreto, line 1: As you observed well here for Anthropic (Having WebSearch results)

- https://github.com/microsoft/agent-framework/pull/210#discussion_r2231528070

We might need an abstraction for the every tool result like `ToolResultContent` (Inputs + Outputs) having dedicated contents for each tool abstraction we add:  `WebSearchResultContent` && `FileSearchResultContent`.

Every tool (`ToolResultContent`) approach would be somewhat tricky as we already have `FunctionCallContent` and `FunctionResultContent` which are dedicated for the `AIFunction`

rogerbarreto: As the other contents can be used for either Config/Options/Tools and Result/Response, I think would make sense to consider removing `Result` from `CodeExecutionResultContent` name as this can eventually be used as a tool input,

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        _fileId = Throw.IfNullOrWhitespace(fileId);
+    }
+
+    /// <summary>
+    /// Gets or sets the ID of the hosted file.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="value"/> is empty or composed entirely of whitespace.</exception>
+    public string FileId
```
ericstj, line 38: This seems a useful holder, but I would expect to do anything useful with it we might also need abstractions for creation, access, removal.  Perhaps a "metadata" concept as well.   I guess without those it's just an abstraction for opaque remote data instance.

rogerbarreto: @ericstj This is a valid point, currently the API tools have only exposed the file identifier and indirectly what is the type "text" or "image", we might consider adding extra properties if we get multiple API's providing more data about the file during the tool config/response, for now the path for those would be using the `AdditionalProperties` dictionary.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/CodeExecutionResultContent.cs
stephentoub: > We might need an abstraction for the every tool result

I'd been imagining it wouldn't be every tool, but rather every hosted tool. We could have a HostedToolResultContent, and then as necessary derived types from that for each specialization.

> as this can eventually be used as a tool input

Sorry, not understanding, can you elaborate?

stephentoub:
@rogerbarreto, other than the design of the code interpreter content type, how are you feeling about everything else in this PR? Does the rest of it look good to you design-wise? I'm wondering if we should get everything else merged, and then work on how we want to represent output content from all of these tools.

rogerbarreto:
@stephentoub I'm happy with the current state representing the inputs/definition of the tooling. Happy to progress into the output content in later additions

> Sorry, not understanding, can you elaborate?

Thinking in a name for the contents that can be used both for input / output scenarios.
i.e: `Inputs.Add(new CodeExecutionContent)`

For example, a code I want to be executed server-side by the tool.

rogerbarreto (approved):

stephentoub:
I've removed the CodeInterpreterResultContent from this PR, though I left some of the support that will be necessary to add it back (e.g. changes in the coalescing code). Ready for review.

rogerbarreto (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a file that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="DataContent"/> which contains the data for a file or blob, this class represents a file that is hosted
+/// by the AI service and referenced by an identifier. Such identifiers are specific to the provider.
```
eiriktsarpalis, line 15: Would we also want to include a `ProviderId` property for easier debuggability in case there's cross contamination between services?

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
stephentoub: I think that's a good idea. Are you thinking it'd just be an arbitrary string that the IChatClient would populate for instances it creates? And presumably it'd just be for diagnostic purposes? If it is for diagnostics, maybe we just put a string into AdditionalProperties?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -402,22 +403,45 @@ private ResponseCreationOptions ToOpenAIResponseCreationOptions(ChatOptions? opt
...
+
+                    case HostedFileSearchTool fileSearchTool:
+                        result.Tools.Add(ResponseTool.CreateFileSearchTool(
+                            fileSearchTool.Inputs?.OfType<HostedVectorStoreContent>().Select(c => c.VectorStoreId) ?? [],
+                            fileSearchTool.MaximumResultCount));
+                        break;
+
+                    case HostedCodeInterpreterTool codeTool:
+                        string json;
+                        if (codeTool.Inputs is { Count: > 0 } inputs)
```
rogerbarreto, line 431: nit. As we are already in breaking glass territory here we could also consider the event of the `codeTool.AdditionalProperties` have a `"Container"` key with a string and use that instead. 

Noting that this would invalidate the configuration for files or throw.

stephentoub: We could... but at that point, wouldn't someone be better off just configuring the underlying ResponseTool via RawResponseFactory? Now that we have RawResponseFactory, the lines are a bit blurry as to when/how we should use AdditionalProperties in these situations. I realize we use it in the HostedWebSearchTool above, so, yeah.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
eiriktsarpalis: That could work as well, although I presume it might be less easy to discover.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
rogerbarreto: This is a very edge case scenario and AFAIK also very exclusive to OpenAI ATM. 

We might just skip for now...

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
rogerbarreto: Metadata uses `ProviderName`, is this the `Id` a different proposal in here?

stephentoub: Presumably it would end up being the same value as what's in ChatClientMetadata.ProviderName, yes. And, yeah, should probably be named the same, if we add it.

I'd like to get this PR in today.

@rogerbarreto  / @eiriktsarpalis, yay or nay on `public string? ProviderName { get; set; }` on HostedFileContent and HostedVectorStoreContent?

rogerbarreto: No fully sure on adding this, if this is for diagnostic purpose only I would have this as a hidden property that can be serialized/deserialized, something a bit similar how OpenAI SDK does with their Models.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Contents/HostedFileStoreContentTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        Assert.Throws<ArgumentNullException>(() => new HostedVectorStoreContent(null!));
+        Assert.Throws<ArgumentException>(() => new HostedVectorStoreContent(string.Empty));
+        Assert.Throws<ArgumentException>(() => new HostedVectorStoreContent(" "));
+    }
+
+    [Fact]
+    public void Constructor_String_PropsDefault()
+    {
+        string fileId = "id123";
```
eavanvalkenburg:
```suggestion
        string vectorStoreId= "id123";
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIContent.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.Extensions.AI;
 [JsonDerivedType(typeof(ErrorContent), typeDiscriminator: "error")]
 [JsonDerivedType(typeof(FunctionCallContent), typeDiscriminator: "functionCall")]
 [JsonDerivedType(typeof(FunctionResultContent), typeDiscriminator: "functionResult")]
+[JsonDerivedType(typeof(HostedFileContent), typeDiscriminator: "file")]
```
eavanvalkenburg: is 'file' specific enough, what if we want to add "LocalFileContent"?

stephentoub: Tell me more :) What would "LocalFileContent" be?

eavanvalkenburg: don't know, but file seems very generic...

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedFileContent.cs
ericstj: I see the usefulness if a process might be using multiple providers that both provide hosted file content and needing to know which provider to ask for that content. For that matter we might need to know "deployment name" or something of the sort, if I have multiple deployments of the same provider and might be dealing with files hosted on different instances/deployments. I need a fully descriptive "address/endpoint" to fetch the file from.

stephentoub: I'm going to leave this out for now. We can always add it later.

PR #6621, opened 2025-07-16
Add support for new Azure AI Foundry project type for Safety evals

### LINKED ISSUE #6592: [AI Evaluation] Support the new Azure AI Foundry project type for Safety evals
[Create a project for Azure AI Foundry](https://learn.microsoft.com/en-us/azure/ai-foundry/how-to/create-projects?tabs=ai-foundry&pivots=fdp-project) lists the steps for creating the new project type.

The new project type is easier to setup and configure and while the old (Hub-based) project type (represented by [`ContentSafetyServiceConfiguration`](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyServiceConfiguration.cs)) is still supported, the new project type is the preferred one going forward.

shyamnamboodiripad:
Fixes #6592
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6621)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
@@ -22,6 +22,9 @@ namespace Microsoft.Extensions.AI.Evaluation.Safety;
 
 internal sealed partial class ContentSafetyService(ContentSafetyServiceConfiguration serviceConfiguration)
 {
+    private const string APIVersionForServiceDiscoveryInHubBasedProjects = "?api-version=2023-08-01-preview";
+    private const string APIVersionForNonHubBasedProjects = "?api-version=2025-05-15-preview";
+
     private static HttpClient? _sharedHttpClient;
     private static HttpClient SharedHttpClient
     {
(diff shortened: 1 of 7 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
```diff
@@ -244,7 +254,10 @@ private async ValueTask EnsureServiceAvailabilityAsync(
         string evaluatorName,
         CancellationToken cancellationToken)
     {
-        string serviceAvailabilityUrl = $"{serviceUrl}/checkannotation";
+        string serviceAvailabilityUrl =
+            serviceConfiguration.IsHubBasedProject
+                ? $"{serviceUrl}/checkannotation"
+                : $"{serviceUrl}/checkannotation/{APIVersionForNonHubBasedProjects}";
```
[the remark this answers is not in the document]
shyamnamboodiripad: Good catch - fixed!

src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
```diff
@@ -196,7 +206,7 @@ private async ValueTask<string> GetServiceDiscoveryUrlAsync(
             $"https://management.azure.com/subscriptions/{serviceConfiguration.SubscriptionId}" +
             $"/resourceGroups/{serviceConfiguration.ResourceGroupName}" +
             $"/providers/Microsoft.MachineLearningServices/workspaces/{serviceConfiguration.ProjectName}" +
-            $"?api-version=2023-08-01-preview";
+            $"/{APIVersionForServiceDiscoveryInHubBasedProjects}";
```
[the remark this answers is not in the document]
shyamnamboodiripad: Good catch - fixed!

src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
```diff
@@ -297,7 +310,10 @@ private async ValueTask<string> SubmitAnnotationRequestAsync(
         string evaluatorName,
         CancellationToken cancellationToken)
     {
-        string annotationUrl = $"{serviceUrl}/submitannotation";
+        string annotationUrl =
+            serviceConfiguration.IsHubBasedProject
+                ? $"{serviceUrl}/submitannotation"
+                : $"{serviceUrl}/submitannotation/{APIVersionForNonHubBasedProjects}";
```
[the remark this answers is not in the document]
shyamnamboodiripad: Good catch - fixed!

src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyServiceConfiguration.cs
```diff
@@ -73,11 +70,151 @@ public sealed class ContentSafetyServiceConfiguration(
...
-    public int TimeoutInSecondsForRetries { get; } = timeoutInSecondsForRetries;
+    public int TimeoutInSecondsForRetries { get; }
+
+    [MemberNotNullWhen(true, nameof(SubscriptionId), nameof(ResourceGroupName), nameof(ProjectName))]
+    [MemberNotNullWhen(false, nameof(Endpoint))]
+    internal bool IsHubBasedProject =>
+        !string.IsNullOrWhiteSpace(SubscriptionId) &&
+        !string.IsNullOrWhiteSpace(ResourceGroupName) &&
+        !string.IsNullOrWhiteSpace(ProjectName) &&
+        Endpoint is null;
```
peterwald, line 88: For more consistency, does it make sense to switch `Endpoint is null` to `string.IsNullOrWhiteSpace(Endpoint)`. Or maybe switch the others so they match.

shyamnamboodiripad: Ah endpoint is a Uri while the others are strings. I wanted to also handle whitespace for the strings (since that is equivalent to null in this case).

I could probably add parameter validation for the strings in the constructor (in addition to the above) to better enforce this though...

peterwald: Ah, I missed the type difference.

shyamnamboodiripad: I have added the above null checks in the constructors. However, I realized that we are inconsistent in how we null check inputs to public APIs across the eval libraries :) We seem to be relying on nullable for the most part - but nullable checks won't guard against runtime nulls.

The repo has an analyzer that will report a warning if a (non-nullable) parameter on a public API is used in a context like for loop or function call where null is not allowed - but no warning is reported in other cases (for example, if we merely store the supplied argument value in a (non-nullable) field).

We probably need an audit of the public APIs to null check all inputs (not just in the cases where the analyzer reports the warning). Not a huge issue - but probably just something to keep in mind and fix up over time (using Copilot :))...

PR #6623, opened 2025-07-17
Add AmbientMetadata.Build component

evgenyfedorov2:
Adding a brand new component - Build metadata, which automatically grabs build information from the CI/CD pipelines, deserializes it into a strong type `BuildMetadata` and registers it in Dependency Injection container as `IOptions<BuildMetadata>`.
The component uses source generation to collect build information and immediately express it via C# code.

Initially, only GitHub Actions and Azure DevOps are supported
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6623)


dariusclay:
I love porting until I realize my commit history is lost. 🤣

src/Libraries/Microsoft.Extensions.AmbientMetadata.Build/BuildMetadata.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// Gets or sets the latest version control change that is included in this build, also known as the commit SHA.
+    /// </summary>
+    public string? SourceVersion { get; set; }
+
+    /// <summary>
+    /// Gets or sets the build time in sortable date/time pattern.
+    /// This is the time the BuildMetadataGenerator was run.
+    /// </summary>
+    public string? BuildDateTime { get; set; }
```
slang25: Should `BuildDateTime`,  `BuildNumber` and `BuildId` be removed or be optional, hear me out...

A good practice in dotnet is [reproducible builds](https://github.com/dotnet/reproducible-builds), having source info like repository url, source commit etc... is fine because that would be consistent and considered an input to the build (albeit implicitly), but certainly timestamps would break this practice.

evgenyfedorov2: Makes sense, thank you for the comment. Need to think about it

src/Generators/Microsoft.Gen.BuildMetadata/Model.cs
```diff
@@ -0,0 +1,32 @@
...
+#pragma warning disable S2223 // Non-constant static fields should not be visible - internal in order to be able to override this for tests
+
+namespace Microsoft.Gen.BuildMetadata;
+
+internal static class Model
+{
+    // Azure DevOps environment variables: https://learn.microsoft.com/azure/devops/pipelines/build/variables#build-variables-devops-services
+    internal static string? AzureBuildId = Environment.GetEnvironmentVariable("Build_BuildId");
+    internal static string? AzureBuildNumber = Environment.GetEnvironmentVariable("Build_BuildNumber");
+    internal static string? AzureSourceBranchName = Environment.GetEnvironmentVariable("Build_SourceBranchName");
```
slang25: Couldn't this alternatively be fetched by the `InitializeSourceControlInformation` and related targets in MSBuild that are used for SourceLink, that way this would work for local builds as well as CI.

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,154 @@
...
+                    OutLn("return builder.Add(new BuildMetadataSource(sectionName));");
+                OutCloseBrace();
+            OutCloseBrace();
+        OutCloseBrace();
+    }
+
+    [SuppressMessage("Format", "IDE0055", Justification = "For better visualization of how the generated code will look like.")]
+    private void OutNullGuards(bool checkBuilder = true)
+    {
+        OutPP("#if NETFRAMEWORK");
```
KalleOlaviNiemitalo: This should perhaps be `#if !NET` so that it doesn't try to use the throw helpers in a project that targets .NET Standard.

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
```diff
@@ -0,0 +1,66 @@
...
+        var e = new Emitter();
+        var result = e.EmitExtensions(context.CancellationToken);
+        context.AddSource("BuildMetadataExtensions.g.cs", SourceText.From(result, Encoding.UTF8));
+    }
+
+    private static void OverwriteModelValuesForTesting(ImmutableArray<AttributeData> attributes)
+    {
+        const int TestAttributeConstructorArgumentsLength = 5;
+
+        if (attributes.IsDefaultOrEmpty || attributes[0].ConstructorArguments.Length != TestAttributeConstructorArgumentsLength)
```
KalleOlaviNiemitalo: Should check that the name of the attribute type matches "Microsoft.Gen.BuildMetadata.GenerateBuildMetadataAttribute".

src/Generators/Microsoft.Gen.BuildMetadata/Microsoft.Gen.BuildMetadata.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <Description>Code generator to support Microsoft.Extensions.AmbientMetadata.Build</Description>
+    <Workstream>Fundamentals</Workstream>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AnalyzerLanguage>cs</AnalyzerLanguage>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <InjectExperimentalAttributeOnLegacy>true</InjectExperimentalAttributeOnLegacy>
+    <InjectSharedDiagnosticIds>true</InjectSharedDiagnosticIds>
+    <!-- CA1852 is for the internal shared type DiagDescriptiorsBase. It can't be sealed because there are child classes in other components-->
```
KalleOlaviNiemitalo: Spelling `DiagDescriptorsBase`.

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: Fixed, thanks!

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
evgenyfedorov2: Not relevant after the last updates

src/Generators/Microsoft.Gen.BuildMetadata/Microsoft.Gen.BuildMetadata.csproj
evgenyfedorov2: Fixed, thanks!

evgenyfedorov2:
> What's the purpose of the empty `src/Libraries/Microsoft.Extensions.AmbientMetadata.Build/Microsoft.Extensions.AmbientMetadata.Build.json` file?

Removed :)

src/Generators/Microsoft.Gen.BuildMetadata/Model.cs
evgenyfedorov2: updated to fetch data from MSBuild properties, which by default are initialized from environment variables.

src/Libraries/Microsoft.Extensions.AmbientMetadata.Build/BuildMetadata.cs
evgenyfedorov2: Removed BuildDateTime.
All other properties are fetched from MSBuild

test/Libraries/Microsoft.Extensions.AmbientMetadata.Build.Tests/ConfigurationBindingQuirkBehaviorTests.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace Microsoft.Extensions.AmbientMetadata.Test;
+
+public class ConfigurationBindingQuirkBehaviorTests
+{
+    [Theory]
+    [InlineData("ambientmetadata:build")]
+    [InlineData("customSection:ambientmetadata:build")]
+    public void GivenMetadata_RegistersOptions_HostBuilder(string sectionName)
+    {
+        // When configuration is not available, values are initilaized to the default value of string?, which is null
```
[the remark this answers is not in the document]
evgenyfedorov2: fixed

src/Libraries/Microsoft.Extensions.AmbientMetadata.Build/buildTransitive/Microsoft.Extensions.AmbientMetadata.Build.props
```diff
@@ -0,0 +1,28 @@
+﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Azure DevOps MSBuild properties - use Azure DevOps environment variables if available, otherwise use empty string -->
+    <BuildMetadataAzureBuildId Condition="'$(BuildMetadataAzureBuildId)' == ''">$(Build_BuildId)</BuildMetadataAzureBuildId>
+    <BuildMetadataAzureBuildNumber Condition="'$(BuildMetadataAzureBuildNumber)' == ''">$(Build_BuildNumber)</BuildMetadataAzureBuildNumber>
+    <BuildMetadataAzureSourceBranchName Condition="'$(BuildMetadataAzureSourceBranchName)' == ''">$(Build_SourceBranchName)</BuildMetadataAzureSourceBranchName>
+    <BuildMetadataAzureSourceVersion Condition="'$(BuildMetadataAzureSourceVersion)' == ''">$(Build_SourceVersion)</BuildMetadataAzureSourceVersion>
+    <BuildMetadataIsAzureDevOps Condition="'$(BuildMetadataIsAzureDevOps)' == ''">true</BuildMetadataIsAzureDevOps>
```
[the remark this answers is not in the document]
evgenyfedorov2: fixed

jaredpar:
@333fred, @jjonescz, @JoeRobich  PTAL

src/Generators/Microsoft.Gen.BuildMetadata/Microsoft.Gen.BuildMetadata.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Stage>normal</Stage>
+    <MinCodeCoverage>100</MinCodeCoverage>
+    <MinMutationScore>75</MinMutationScore>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" PrivateAssets="all" />
```
333fred: Generators should not reference this package; if you accidentally depend on anything from the workspaces layer, you will throw during `dotnet build`.
```suggestion
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" PrivateAssets="all" />
```

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
```diff
@@ -0,0 +1,76 @@
...
+        public string? SourceVersion { get; }
+
+        public BuildMetadata(string? buildId, string? buildNumber, string? sourceBranchName, string? sourceVersion)
+        {
+            BuildId = buildId;
+            BuildNumber = buildNumber;
+            SourceBranchName = sourceBranchName;
+            SourceVersion = sourceVersion;
+        }
+    }
```
333fred: Consider just making this a record struct. It provides a more efficient equality definition and is simpler to declare.

```suggestion
    private readonly record struct BuildMetadata(string? BuildId, string? BuildNumber, string? SourceBranchName, string? SourceVersion);
```

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,160 @@
...
+            OutOpenBrace();
+                OutLn("return new MemoryConfigurationProvider(new MemoryConfigurationSource())");
+                OutOpenBrace();
+                    OutLn($$"""{ $"{SectionName}:buildid", "{{_buildId}}" },""");
+                    OutLn($$"""{ $"{SectionName}:buildnumber", "{{_buildNumber}}" },""");
+                    OutLn($$"""{ $"{SectionName}:sourcebranchname", "{{_sourceBranchName}}" },""");
+                    OutLn($$"""{ $"{SectionName}:sourceversion", "{{_sourceVersion}}" },""");
+                OutCloseBraceWithExtra(";");
+            OutCloseBrace();
+        OutCloseBrace();
```
333fred: Consider putting actual braces in your code; it will allow you to get rid of the formatting suppression and make the formatter generally work with you.
```suggestion
        OutLn("private sealed class BuildMetadataSource : IConfigurationSource");
        OutOpenBrace();
        {
            OutLn("public string SectionName { get; }");
            OutLn();

            OutLn("public BuildMetadataSource(string sectionName)");
            OutOpenBrace();
            {
                OutNullGuards(checkBuilder: false);
                OutLn("SectionName = sectionName;");
            }
            OutCloseBrace();
            OutLn();

            OutLn("public IConfigurationProvider Build(IConfigurationBuilder builder)");
            OutOpenBrace();
            {
                OutLn("return new MemoryConfigurationProvider(new MemoryConfigurationSource())");
                OutOpenBrace();
                {
                    OutLn($$"""{ $"{SectionName}:buildid", "{{_buildId}}" },""");
                    OutLn($$"""{ $"{SectionName}:buildnumber", "{{_buildNumber}}" },""");
                    OutLn($$"""{ $"{SectionName}:sourcebranchname", "{{_sourceBranchName}}" },""");
                    OutLn($$"""{ $"{SectionName}:sourceversion", "{{_sourceVersion}}" },""");
                }
                OutCloseBraceWithExtra(";");
            }
            OutCloseBrace();
        }
        OutCloseBrace();
```

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,160 @@
...
+        cancellationToken.ThrowIfCancellationRequested();
+        GenerateBuildMetadataExtensions();
+        return Capture();
+    }
+
+    [SuppressMessage("Format", "IDE0055", Justification = "For better visualization of how the generated code will look like.")]
+    private void GenerateBuildMetadataSource()
+    {
+        OutGeneratedCodeAttribute();
+        OutLn("[EditorBrowsable(EditorBrowsableState.Never)]");
```
333fred: Consider using `global::Fully.Prefixed.Types`, rather than relying on `using`s. You don't know what `global using`s the user will have set up, and names they may have defined could potentially conflict with yours. Smaller generators can get away without fully-specifying, but for reliability we suggest that larger generators always use the fully-qualified form.

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,160 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+using Microsoft.Gen.Shared;
+
+namespace Microsoft.Gen.BuildMetadata;
+
+internal sealed class Emitter : EmitterBase
```
333fred, line 10: `GeneratorUtilies.cs` is not in this review, so I can't comment on it, but it links an issue that was closed in 2021. Consider removing the link and using a complete reason.

test/Generators/Microsoft.Gen.BuildMetadata/Unit/GeneratorTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            { "build_property.BuildMetadataGitHubRefName", "TEST_GITHUB_REFNAME" },
+            { "build_property.BuildMetadataGitHubSha", "TEST_GITHUB_SHA" }
+        });
+    }
+
+    private static async Task<(IReadOnlyList<Diagnostic> diagnostics, IReadOnlyList<GeneratedSourceResult> sources)> RunGenerator(
+        string source,
+        TestAnalyzerConfigOptionsProvider optionsProvider)
+    {
+        // Create a test project and compilation
```
333fred, line 83: Consider using Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing, rather than a homegrown framework. Specifically, `RoslynTestUtils` appears to have a number of anti-patterns, such as getting assemblies by using `typeof(object).Location`. The official testing library takes care of a lot of this for you, getting the real assemblies for different .NET versions.

src/Generators/Microsoft.Gen.BuildMetadata/Microsoft.Gen.BuildMetadata.csproj
evgenyfedorov2: Done, thank you!

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: Done, thank you!

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: Done, thank you!

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
evgenyfedorov2: Done, thank you!

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: That would affect all generators. Created an issue to handle it separately https://github.com/dotnet/extensions/issues/6856

test/Generators/Microsoft.Gen.BuildMetadata/Unit/GeneratorTests.cs
evgenyfedorov2: Thanks! Created an issue https://github.com/dotnet/extensions/issues/6857 to track this!

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.Gen.BuildMetadata;
+
+/// <summary>
+/// Source generator that creates build metadata extensions from MSBuild properties.
+/// Supports both Azure DevOps and GitHub Actions build environments.
+/// </summary>
+[Generator]
+public class BuildMetadataGenerator : IIncrementalGenerator
```
jjonescz:
```suggestion
public sealed class BuildMetadataGenerator : IIncrementalGenerator
```

src/Generators/Microsoft.Gen.BuildMetadata/AnalyzerConfigOptionsExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// Gets a boolean property value from MSBuild properties.
+    /// Supports "true"/"false" values (case-insensitive).
+    /// </summary>
+    /// <param name="options">The analyzer configuration options.</param>
+    /// <param name="propertyName">The property name (without "build_property." prefix).</param>
+    /// <returns>True if the property value is "true", false otherwise.</returns>
+    public static bool GetBooleanProperty(this AnalyzerConfigOptions options, string propertyName)
+    {
+        var value = GetProperty(options, propertyName);
+        return !string.IsNullOrEmpty(value) &&
```
jjonescz: nit: `!string.IsNullOrEmpty(value)` seems superfluous, the `string.Equals` below should be enough

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private static void Execute(SourceProductionContext context, BuildMetadata buildMetadata)
+    {
+        var emitter = new Emitter(buildMetadata.BuildId, buildMetadata.BuildNumber, buildMetadata.SourceBranchName, buildMetadata.SourceVersion);
+        var result = emitter.Emit(context.CancellationToken);
+        context.AddSource("BuildMetadataExtensions.g.cs", SourceText.From(result, Encoding.UTF8));
+    }
+
+    private readonly record struct BuildMetadata(string? BuildId, string? BuildNumber, string? SourceBranchName, string? SourceVersion);
```
jjonescz: Should "IsAzureDevOps" flag be also stored?

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,186 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+using Microsoft.Gen.Shared;
+
+namespace Microsoft.Gen.BuildMetadata;
+
+[SuppressMessage("Format", "S1199", Justification = "For better visualization of how the generated code will look like.")]
```
jjonescz, line 9: Is this suppression still needed? (I'm not sure what S1199 is about actually)

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
```diff
@@ -0,0 +1,59 @@
...
+        return new BuildMetadata(
+            isAzureDevOps ? azureBuildId : gitHubRunId,
+            isAzureDevOps ? azureBuildNumber : gitHubRunNumber,
+            isAzureDevOps ? azureSourceBranchName : gitHubRefName,
+            isAzureDevOps ? azureSourceVersion : gitHubSha);
+    }
+
+    private static void Execute(SourceProductionContext context, BuildMetadata buildMetadata)
+    {
+        var emitter = new Emitter(buildMetadata.BuildId, buildMetadata.BuildNumber, buildMetadata.SourceBranchName, buildMetadata.SourceVersion);
```
jjonescz: nit: perhaps it would be cleaner to make BuildMetadata struct internal and pass that instead of its properties separately.

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,186 @@
...
+    {
+        _buildId = buildId;
+        _buildNumber = buildNumber;
+        _sourceBranchName = sourceBranchName;
+        _sourceVersion = sourceVersion;
+    }
+
+    public string Emit(CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
```
jjonescz: Taking CancellationToken and checking it only at the beginning seems unnecessary? Did you intend to check it also elsewhere?

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
```diff
@@ -0,0 +1,186 @@
...
+        if (checkBuilder)
+        {
+            OutLn("if (builder is null)");
+            OutOpenBrace();
+            OutLn("throw new global::System.ArgumentNullException(nameof(builder));");
+            OutCloseBrace();
+            OutLn();
+        }
+
+        OutLn("if (global::System.string.IsNullOrWhiteSpace(sectionName))");
```
jjonescz: Since `string` is a keyword, no need to fully qualify it (and even if you want to fully qualify it, you seem to have the casing wrong currently)

```suggestion
        OutLn("if (string.IsNullOrWhiteSpace(sectionName))");
```

333fred: Note: not catching this issue in tests is a problem. The source generator testing framework would have caught it, but at the very least consider verifying that there are no diagnostics in the compilation that is returned from `generatorDriver.RunGeneratorsAndUpdateCompilation`

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: It is a Sonar rule https://sonarqube.inria.fr/sonarqube/coding_rules?open=csharpsquid%3AS1199&rule_key=csharpsquid%3AS1199, kicks in because we have extra curly braces

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
evgenyfedorov2: Done, thanks!

src/Generators/Microsoft.Gen.BuildMetadata/AnalyzerConfigOptionsExtensions.cs
evgenyfedorov2: Done, thanks!

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
evgenyfedorov2: Done, thanks!

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: Addressed both comments, thank you!

src/Generators/Microsoft.Gen.BuildMetadata/Emitter.cs
evgenyfedorov2: Removed it

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
evgenyfedorov2: As of today it is used for deciding azure devops vs. github actions, nothing else. It is not used as a public property either. So I don't see why we should store it?

jjonescz: If I understand correctly, this source generator is about embedding metadata like build ID into the compilation, right? But you only embed "build ID", and then when I look into the DLL, I will see an attribute which tells me the "build ID" but I won't know if it was actually built on GitHub Actions or AzDo. (I guess I should know where my product is built, but maybe it can be built on either of these...) So I was just wondering if that would also be a useful piece of information to embed. But I may be completely misunderstanding the big picture here :)

Also I guess IsAzDo + BuildId is still not a unique identifier (to be able to unambiguously find the original build), one would also need to know the azdo org etc. So that's probably not the goal and you can ignore me.

test/Generators/Microsoft.Gen.BuildMetadata/Unit/GeneratorTests.cs
```diff
@@ -89,12 +90,13 @@ private static TestAnalyzerConfigOptionsProvider CreateTestOptionsProvider(bool?
         // Create the generator driver with the options provider
         var driver = Microsoft.CodeAnalysis.CSharp.CSharpGeneratorDriver.Create(
             generators: new[] { new BuildMetadataGenerator().AsSourceGenerator() },
+            parseOptions: CSharpParseOptions.Default.WithLanguageVersion(LanguageVersion.Preview),
```
jjonescz, line 93: If the source generators are supposed to run across various user configurations, they should support different language versions. I'm not sure if that's something worth testing though.

test/Generators/Microsoft.Gen.BuildMetadata/Unit/GeneratorTests.cs
```diff
@@ -89,12 +90,13 @@ private static TestAnalyzerConfigOptionsProvider CreateTestOptionsProvider(bool?
...
             optionsProvider: optionsProvider);
 
-        var result = driver.RunGenerators(comp!);
+        var result = driver.RunGeneratorsAndUpdateCompilation(comp!, out var outputCompilation, out var diagnostics);
```
333fred, line 96: `diagnostics` is just the diagnostics your generator would have reported itself; the `global::System.string` issue would not have been caught by this. Please verify that `compilation.GetDiagnostics` either returns nothing, or the set of diagnostics you're expecting.

src/Generators/Microsoft.Gen.BuildMetadata/BuildMetadataGenerator.cs
evgenyfedorov2: > but I won't know if it was actually built on GitHub Actions or AzDo
this is intended, this component tries to abstract away that stuff. In theory, we can be adding support of some other CI system in future

> Also I guess IsAzDo + BuildId is still not a unique identifier 
This is not the goal either. If users find it useful, we will consider adding it, but right now we try to limit the public API surface and make the component simple - it just gives users build information.

src/Libraries/Microsoft.Extensions.AmbientMetadata.Build/README.md
```diff
@@ -0,0 +1,4 @@
+# About This Project
+
+This ambient metadata provider provides Build information at runtime.
+See https://learn.microsoft.com/azure/devops/pipelines/build/variables for details.
```
eduherminio: Should we also mention https://docs.github.com/en/actions/reference/variables-reference#default-environment-variables here?

test/Generators/Microsoft.Gen.BuildMetadata/Unit/GeneratorTests.cs
evgenyfedorov2: Done, thank you!

src/Libraries/Microsoft.Extensions.AmbientMetadata.Build/README.md
evgenyfedorov2: Done, thank you!

test/Generators/Microsoft.Gen.BuildMetadata/Unit/GeneratorTests.cs
evgenyfedorov2: I am not sure either, it has worked with Preview version and nobody complained :)

Issue #6604, opened 2025-07-14
The aichatweb project template references nonexistent preview version of M.E.AI


jeffhandley:
### Description

When creating a new project with the aichatweb template, the project file produces the wrong version for the `Microsoft.Extensions.AI` package.

```xml
    <PackageReference Include="Microsoft.Extensions.AI" Version="9.7.0-preview.1.25356.2" />
```

This seems to have been broken since Microsoft.Extensions.AI went stable with version 9.5.0. NuGet is still able to resolve the package, with the following warning:

```sh
warning NU1603: meai_20250713_test depends on Microsoft.Extensions.AI (>= 9.7.0-preview.1.25356.2) but Microsoft.Extensions.AI 9.7.0-preview.1.25356.2 was not found. Microsoft.Extensions.AI 9.7.0 was resolved instead.
```

### Reproduction Steps

```pwsh
dotnet new install Microsoft.Extensions.AI.Templates::9.7.0-preview.2.25356.2
mkdir meai_20250713_test
cd meai_20250713_test
dotnet new aichatweb
cat .\meai_20250713_test.csproj
```

### Expected behavior

```xml
    <PackageReference Include="Microsoft.Extensions.AI" Version="9.7.0" />
```

### Actual behavior

```xml
    <PackageReference Include="Microsoft.Extensions.AI" Version="9.7.0-preview.1.25356.2" />
```

### Other information

We still haven't fully cracked the nut for version management in `src\ProjectTemplates\GeneratedContent.targets`, as it doesn't handle referencing stable package versions for MEAI packages when producing official builds.

This also means something is incorrect with our integration tests (at least on official builds), as the build warning should cause the template integration test to fail when building every supported configuration of the template projects.

PR #6651, opened 2025-07-23
Fix internal package version detection for project templates

MackinnonBuck:
## Summary

Fixes an issue where project templates would sometimes reference nonexistent versions of packages defined internally within the `dotnet/extensions` repo.

Fixes https://github.com/dotnet/extensions/issues/6604

## Description

This fix works by executing a target to directly extract the package version from each internal project referenced by the project templates. This enables us to avoid having to "guess" the version of internal packages via the `$(Version)` property, which may differ from project to project.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6651)


evgenyfedorov2 (approved):

jeffhandley (approved):
I appreciate how surgical the changes are for consuming the versions in this new way.

PR #6654, opened 2025-07-24
Fix M.E.AI package refs

stephentoub:


src/Libraries/Microsoft.Extensions.AI.OpenAI/Microsoft.Extensions.AI.OpenAI.csproj
```diff
@@ -33,6 +33,9 @@
 
   <ItemGroup>
     <PackageReference Include="OpenAI" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
```
ericstj, line 38: Conditioning these out on all .NETCoreApp versions while still targeting `netstandard2.0` can create compatibility issues.

Consider if your `netstandard2.0` build uses System.Text.Json 9.0.x.  If you exclude the package reference when targeting `net8.0` you'll have a problem. 

So either 1) ensure the version of libraries packages you reference matches the major version of your lowest .NETCoreApp version or 2) only exclude these references on major versions >= your package major versions.

stephentoub: We're already doing this in other libs, e.g.
https://github.com/dotnet/extensions/blob/5d8efb684ac5a56965f6fb72037cbd1db51ad74e/src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.csproj#L33-L40

> Consider if your netstandard2.0 build uses System.Text.Json 9.0.x. If you exclude the package reference when targeting net8.0 you'll have a problem.

The netstandard2.0 build uses STJ 8:
https://github.com/dotnet/extensions/blob/5d8efb684ac5a56965f6fb72037cbd1db51ad74e/eng/packages/General-LTS.props#L5-L41
https://github.com/dotnet/extensions/blob/5d8efb684ac5a56965f6fb72037cbd1db51ad74e/eng/Versions.props#L125

ericstj: I see, so long as you keep these versions at 8.0 and the lowest .NETCoreApp version you support is 8.0 then it's fine.

ericstj (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/Microsoft.Extensions.AI.OpenAI.csproj
stephentoub: Once the repo updates for 10, then presumably the STJ package will bump to 10, at which point I assume we'll have an issue for the net8.0 and net9.0 builds. At that point presumably we'd need to then change the condition here (and in the existing case) to be < net10.0 rather than any .NET Core tfm? Or only if we actually need a 10.0 feature from the package?

sebastienros:
This PR seems to be the reason the main branch builds are broken on SourceIndexStage1 job.

https://dev.azure.com/dnceng/internal/_build/results?buildId=2757536&view=logs&j=316d5c15-0c50-544e-8051-e6b14a1ab674&t=4e01507e-10cc-507d-f6d0-adc59e92bcb8

stephentoub:
> This PR seems to be the reason the main branch builds are broken on SourceIndexStage1 job.
> 
> https://dev.azure.com/dnceng/internal/_build/results?buildId=2757536&view=logs&j=316d5c15-0c50-544e-8051-e6b14a1ab674&t=4e01507e-10cc-507d-f6d0-adc59e92bcb8

I'm having trouble deciphering the dependency output there. @ericstj, it looks like something is referencing STJ 6.0 and S.D.DiagnosticSource 6.0 packages and that's causing a conflict?

sebastienros:
@joperezr said he would take a look at it

ericstj:
Something weird about the build of https://github.com/dotnet/extensions/blob/main/src/Packages/Microsoft.Internal.Extensions.DotNetApiDocs.Transport/Microsoft.Internal.Extensions.DotNetApiDocs.Transport.proj

I would expect it to get the latest System.Diagnostics.DiagnosticSource which is brought in transitively by the projects it references, but instead it's getting an older version.

Something fishy with how that project is set up.  I see @joperezr already worked around a similar bug before https://github.com/dotnet/extensions/commit/424e9748b0a4815fd291f8a7e6a4fbfb76418bbd

stephentoub:
@sebastienros / @joperezr , did this get resolved?

sebastienros:
From what I see, [not resolved](https://dev.azure.com/dnceng/internal/_build/results?buildId=2760611&view=results)

PR #6657, opened 2025-07-24
Add ChatMessage.CreatedAt

stephentoub:
We currently have it at the ChatResponse{Update} level, but for more agentic scenarios, it's helpful to have a timestamp per message.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6657)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
@@ -53,6 +53,7 @@ public class ChatMessage
             AdditionalProperties = AdditionalProperties,
             _authorName = _authorName,
             _contents = _contents,
+            CreatedAt = CreatedAt,
             RawRepresentation = RawRepresentation,
             Role = Role,
             MessageId = MessageId,
@@ -64,6 +65,9 @@ public class ChatMessage
         get => _authorName;
         set => _authorName = string.IsNullOrWhiteSpace(value) ? null : value;
     }
+
+    /// <summary>Gets or sets a timestamp for the chat message.</summary>
+    public DateTimeOffset? CreatedAt { get; set; }
 
     /// <summary>Gets or sets the role of the author of the message.</summary>
     public ChatRole Role { get; set; } = ChatRole.User;
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
@@ -130,19 +130,19 @@ public class ChatResponse
             ChatMessage message = _messages![i];
             updates[i] = new ChatResponseUpdate
             {
-                ConversationId = ConversationId,
-
                 AdditionalProperties = message.AdditionalProperties,
                 AuthorName = message.AuthorName,
                 Contents = message.Contents,
+                MessageId = message.MessageId,
                 RawRepresentation = message.RawRepresentation,
                 Role = message.Role,
 
-                ResponseId = ResponseId,
-                MessageId = message.MessageId,
-                CreatedAt = CreatedAt,
+                ConversationId = ConversationId,
                 FinishReason = FinishReason,
-                ModelId = ModelId
+                ModelId = ModelId,
+                ResponseId = ResponseId,
+
+                CreatedAt = message.CreatedAt ?? CreatedAt,
             };
         }
 
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
@@ -162,6 +162,11 @@ internal sealed class OpenAIResponsesChatClient : IChatClient
             }
         }
 
+        foreach (var message in response.Messages)
+        {
+            message.CreatedAt = openAIResponse.CreatedAt;
+        }
+
         return response;
     }
 
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
@@ -84,9 +84,10 @@ public static class ChatResponseExtensions
         var contentsList = filter is null ? update.Contents : update.Contents.Where(filter).ToList();
         if (contentsList.Count > 0)
         {
-            list.Add(new ChatMessage(update.Role ?? ChatRole.Assistant, contentsList)
+            list.Add(new(update.Role ?? ChatRole.Assistant, contentsList)
             {
                 AuthorName = update.AuthorName,
+                CreatedAt = update.CreatedAt,
                 RawRepresentation = update.RawRepresentation,
                 AdditionalProperties = update.AdditionalProperties,
             });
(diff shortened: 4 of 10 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -130,19 +130,19 @@ public ChatResponseUpdate[] ToChatResponseUpdates()
...
-                ResponseId = ResponseId,
-                MessageId = message.MessageId,
-                CreatedAt = CreatedAt,
+                ConversationId = ConversationId,
                 FinishReason = FinishReason,
-                ModelId = ModelId
+                ModelId = ModelId,
+                ResponseId = ResponseId,
+
+                CreatedAt = message.CreatedAt ?? CreatedAt,
```
shyamnamboodiripad, line 145: Does this mean that it is possible to have a case where none of the ChatResponseUpdates have the same CreatedAt as the ChatResponse itself?

shyamnamboodiripad, line 145: Should the ChatResponse constructor on line 33 that takes a single ChatMessage also update the ChatResponse.CreatedAt?

Should the constructor on line 42 that takes a collection of ChatMessages update the ChatResponse.CreatedAt to the ChatMessage.CreatedAt for the first / oldest message?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
```diff
@@ -65,6 +66,9 @@ public string? AuthorName
         set => _authorName = string.IsNullOrWhiteSpace(value) ? null : value;
     }
 
+    /// <summary>Gets or sets a timestamp for the chat message.</summary>
+    public DateTimeOffset? CreatedAt { get; set; }
```
shyamnamboodiripad, line 70: Should the constructors initialize this to DateTime.Now perhaps? I guess we would have to think about whether it needs to be nullable then (and I guess that would raise a similar question about whether ChatResponse.CreatedAt should be nullable).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -162,6 +162,11 @@ internal static ChatResponse FromOpenAIResponse(OpenAIResponse openAIResponse, R
             }
         }
 
+        foreach (var message in response.Messages)
+        {
+            message.CreatedAt = openAIResponse.CreatedAt;
```
shyamnamboodiripad, line 167: Would similar changes be required in externally defined clients like the one in OllamaSharp?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
shyamnamboodiripad: Ah looking at the rest of this PR, looks like the intent is to only have this populated for model-generated messages (and not for user messages)?

Having it available for both user and model generated messages could make it easier to implement sorting of collection(s) of messages based on timestamps (like in the case we were discussing offline).

shyamnamboodiripad (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
stephentoub: That would be possible if the ChatResponse's messages had different CreatedAt's from the response, yes.

stephentoub: Basically should the ctor inherit properties from the ChatMessage? Seems reasonable.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
stephentoub: Desirable, probably. Required, no.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
stephentoub: Are you suggesting that in cases where we construct a ChatMessage to send whether we should populate its CreatedAt with DateTimeOffset.UtcNow? You of all people probably don't want that, as it'll completely break caching :smile:

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
stephentoub: Actually, I don't think we should do that. I don't think there should be a meaningful difference between:
```C#
ChatResponse response = new(message);
```
and
```C#
ChatResponse response = new([message]);
```
and
```C#
List<ChatMessage> messages = [];
ChatResponse response = new(messages);
messages.Add(message);
```
and
```C#
ChatResponse response = new();
response.Messages.Add(message);
```

The developer constructing the ChatResponse can set the properties they want to set.

That's different from this ToChatResponseUpdates method, which is just taking the already constructed response and distributing its data into ChatResponseUpdate instances.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatMessage.cs
shyamnamboodiripad: Ah good point - I hadn't considered the impact on caching! (Although if I didn't know how caching worked, I would have naively assumed that CreatedAt would be ignored when the cache key is computed :)) 

The CreatedAt could cause some confusion in the caching scenario. For example, if the conversation history includes some responses from cache and others not from cache, the timestamps between these response messages could end up being really far apart. However, the CreatedAt for the response messages in the history should still always be in the right temporal order even if the timestamps happen to be far apart (since once a cache miss happens, all subsequent messages for the same conversation should also result in cache misses)...

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
shyamnamboodiripad: In all of the cases above, it should be possible to update the ChatResponse.CreatedAt to match the CreatedAt for the oldest added message (i.e., if the message being added to the collection has an older CreatedAt than the ChatResponse.Created, then update the ChatResponse.CreatedAt to the older value). The caller would still have the option to update the property after the fact if this default behavior does not work for their scenario for some reason.

That said, I agree this may be complicated / confusing - leaving it completely up to the callers to decide how CreatedAt should be set for a ChatResponse (that is constructed from scratch) seems like a reasonable call.

stephentoub: > In all of the cases above, it should be possible to update the ChatResponse.CreatedAt to match the CreatedAt for the oldest added message (i.e., if the message being added to the collection has an older CreatedAt than the ChatResponse.Created, then update the ChatResponse.CreatedAt to the older value).

In the 2nd and 3rd cases, an arbitrary list is being provided to ChatResponse, and in the 3rd case, it's modified after construction. We have no way of tracking changes to that list to know when a message has been added, removed, or changed.

shyamnamboodiripad: Ah indeed you're right - my bad should have read the examples closer and realized that ChatResponse actually holds on to the list supplied by the caller.

Issue #6211, opened 2025-03-28
Use Aspire hosting integration packages for chat template


MackinnonBuck:
The Aspire variation of the AI Chat Web template currently relies on configuring connection strings for Azure OpenAI and Azure AI Search. This requires the customer to manually provision resources and obtain connection strings.

We should consider using `Aspire.Hosting.Azure.CognitiveServices` and `Aspire.Hosting.Azure.Search` to help streamline provisioning and configuration.

PR #6659, opened 2025-07-28
Use Azure hosting integrations in Aspire AI Chat Web template

MackinnonBuck:
Updates the "AI Chat Web" project template to use Aspire Azure hosting integrations.

Fixes https://github.com/dotnet/extensions/issues/6211
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6659)

### DIFF
--- a/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Snapshots/aichatweb.AzureOpenAI_Qdrant_Aspire.verified/aichatweb/aichatweb.Web/aichatweb.Web.csproj
+++ b/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Snapshots/aichatweb.AzureOpenAI_Qdrant_Aspire.verified/aichatweb/aichatweb.Web/aichatweb.Web.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.AI.OpenAI" Version="2.2.0-beta.5" />
+    <PackageReference Include="Azure.AI.OpenAI" Version="2.3.0-beta.1" />
     <PackageReference Include="Aspire.Azure.AI.OpenAI" Version="9.4.0-preview.1.25378.8" />
     <PackageReference Include="Microsoft.Extensions.AI.OpenAI" Version="9.9.0" />
     <PackageReference Include="Microsoft.Extensions.AI" Version="9.9.0" />
--- a/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Snapshots/aichatweb.BasicAspire.verified/aichatweb/aichatweb.Web/aichatweb.Web.csproj
+++ b/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/Snapshots/aichatweb.BasicAspire.verified/aichatweb/aichatweb.Web/aichatweb.Web.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.AI.OpenAI" Version="2.2.0-beta.5" />
+    <PackageReference Include="Azure.AI.OpenAI" Version="2.3.0-beta.1" />
     <PackageReference Include="Aspire.Azure.AI.OpenAI" Version="9.4.0-preview.1.25378.8" />
     <PackageReference Include="Microsoft.Extensions.AI.OpenAI" Version="9.9.0" />
     <PackageReference Include="Microsoft.Extensions.AI" Version="9.9.0" />


jeffhandley:
Looks clean to me, @MackinnonBuck. I presume you're seeking some feedback from folks with Aspire expertise; beyond that, what input/feedback/testing would be helpful for you on this?

MackinnonBuck:
> I presume you're seeking some feedback from folks with Aspire expertise; beyond that, what input/feedback/testing would be helpful for you on this?

Thanks @jeffhandley - If others are interesting in trying out the new Aspire + Azure OpenAI + Azure AI search template, that would be helpful to validate that I didn't miss any steps in the README. But I would say additional testing isn't a hard requirement before the PR merges; I was able to manually validate the changes myself 🙂

MackinnonBuck:
cc @eerhardt, @JamesNK, @davidfowl - I would appreciate getting some eyes on the Aspire-related changes 🙂

davidfowl:
Update to 9.4

davidfowl:
@sebastienros can you review this ?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
```diff
@@ -1,6 +1,6 @@
 var builder = DistributedApplication.CreateBuilder(args);
 #if (IsOllama) // ASPIRE PARAMETERS
-#else // IsAzureOpenAI || IsOpenAI || IsGHModels
+#elif (IsOpenAI || IsGHModels)
```
sebastienros, line 3: There is a GH Models integration in Aspire 9.4. Instead of adding a connection string you can call `builder.AddGitHubModel("openai", "openai/gpt-4o-mini");`

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
```diff
@@ -13,14 +13,27 @@
 //   dotnet user-secrets set ConnectionStrings:openai "Endpoint=https://YOUR-DEPLOYMENT-NAME.openai.azure.com;Key=YOUR-API-KEY"
 #endif
 var openai = builder.AddConnectionString("openai");
+#else // IsAzureOpenAI
+
+// See https://learn.microsoft.com/dotnet/aspire/azure/local-provisioning#configuration
+// for instructions providing configuration values
+var openai = builder.AddAzureOpenAI("openai");
```
sebastienros, line 20: It feels too specific to give the name `openai` to the resource. I would have called it `chat` or `somemodel`. From a consumer perspective it won't matter if it's an OpenAI model, or even if it's hosting on Azure, OpenAI ... somewhere else, since the programing model allows all of these combinations.

Unless the app is really meant for a model from OpenAI (even hosted on Azure or GitHub models).

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
```diff
@@ -13,14 +13,27 @@
...
+
+openai.AddDeployment(
+    name: "gpt-4o-mini",
+    modelName: "gpt-4o-mini",
```
sebastienros, line 24: If the consumer app resolving the "openai" resource or the "gpt-4o-mini" resource (resource names, not model names). I can see how the client can set the model name, but maybe it would make sense to name all deployments/models something similar to the consumer points to the same thing? If you have tried it and it works then it is probably fine.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/Program.Aspire.cs
```diff
@@ -3,8 +3,9 @@
 using ChatWithCustomData_CSharp.Web.Services;
 using ChatWithCustomData_CSharp.Web.Services.Ingestion;
 #if (IsOllama)
-#else // IsAzureOpenAI || IsOpenAI || IsGHModels
+#elif (IsOpenAI || IsGHModels)
 using OpenAI;
+#else // IsAzureOpenAI
```
sebastienros, line 8: There is an Azure Foundry resource in Aspire 9.4 that can host any type of model (OpenAI and others). Maybe that's a better choice (I understand there are too many options already ;) ).

FYI, I just opened a PR to model the actual OpenAI resource Not just an openai model on the Azure OpenAI offering, for Aspire 9.5 though.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
MackinnonBuck: In the client [`Program.cs`](https://github.com/dotnet/extensions/blob/ba4e6f9a739d1869c7c91c59077a31c03fa5ce60/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/Program.Aspire.cs#L25-L35), we call `AddAzureOpenAIClient()`, so there are still some aspects specific to Azure OpenAI even if the resource is named something else. I was mainly trying to follow the [docs](https://learn.microsoft.com/dotnet/aspire/azureai/azureai-openai-integration?tabs=dotnet-cli#add-an-azure-openai-resource) as closely as possible, and the docs use the "openai" naming. That said, I'm totally fine with picking a different name like "chat" 🙂

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
MackinnonBuck: Yeah, maybe a more general name like "chat" would be better. I'll change that.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/Program.Aspire.cs
MackinnonBuck: I wonder if Azure AI foundry should be another top-level option on the chat template. The current command-line option is `--provider azureopenai` so I think it makes sense to stick with Azure OpenAI hosting integration until we add `--provider azureaifoundry`. If you agree, I think this is something we'd do in a follow-up rather than in this PR.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
MackinnonBuck: Strangely, it looks like it's actually the _model_ name that the client has to use to identity the resource. So maybe it's best to keep them the same.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
MackinnonBuck: I'm not seeing support for embeddings in the client integration, and I'm not seeing any mention of embeddings in the [docs](https://learn.microsoft.com/dotnet/aspire/github/github-models-integration?tabs=dotnet-cli). I see `AddAzureChatCompletionsClient()`, but not an embeddings equivalent. Do you know if this is supported?

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
sebastienros: This shouldn't be the case, so it sounds like a bug. I am testing it rn.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.AppHost/Program.cs
sebastienros: Sounds like a gap, I will look into adding it.

Issue #6492, opened 2025-06-04
Create representations for hosted MCP servers and content


stephentoub:
AI services are starting to support server-side use of MCP servers. For example, Anthropic's service:
https://docs.anthropic.com/en/docs/agents-and-tools/mcp-connector
and OpenAI responses:
https://platform.openai.com/docs/guides/tools-remote-mcp

With the ModelContextProtocol library, it's easy to use any MCP server, including stdio ones, locally, treating every tool as an AIFunction, but there's not currently an abstraction for the hosted MCP server case, where the service uses the server directly.

- [ ] Rough sketch of how we could handle this well:
  1. [x] Add a `HostedMcpServer : AITool`. This would be configurable with all the common stuff: server url, optional list of allowed tool names, indication of whether to allow auto invocation, etc.
  2. [x] Add a `HostedMcpServerToolCall : AIContent` to represent the call the service makes to the server (including tool name and arguments) and similarly a `HostedMcpServerToolResult` to represent the result of the operation.
  3. [x] The MCP spec recommends human in the loop on tool calls, and the OpenAI design defaults to not automatically invoking tools; interestingly Anthropic's doesn't, but I suspect that's coming. We might also need an AIContent to represent approval/denial for the server invoking a server-side tool.
  4. [ ] For `IChatClient` that don't have this capability, we can enable it via an `McpServerChatClient : IChatClient` that itself uses MCP clients. It would translate a `HostedMcpServer` tool into creating an McpClient and replacing the tool in the tool collection with the appropriate McpClientTool instances. With a FunctionInvokingChatClient in the pipeline, it would enable then similar automatic handling of MCP Server interactions.
https://github.com/modelcontextprotocol/csharp-sdk/pull/1086.
  5. [x] Update FunctionInvokingChatClient to support automatic approval.

Additional details originally captured in Follow-ups for MCP tool (#6779):

- [ ] https://github.com/dotnet/extensions/pull/6664 adds the MCP tool and supporting types as experimental. Some things to follow-up on before marking them stable:
  1. [x] Do we need the base "user input" / "user output" types?
  2. [x] The OpenAI Responses implementation currently looks for the tool approval as part of a Tool message. Is that right? Should it be User instead? Should it be any kind of message?
  3. [x] We need to be stricter about which roles can contain approval responses, only user roles should be allowed, this applies to both MCP and FICC. We can reconsider supporting other roles in the future https://github.com/dotnet/extensions/pull/6881#discussion_r2419557556.
[...]

PR #6664, opened 2025-07-30
Add abstraction for remote MCP servers

jozkee:
Contributes to https://github.com/dotnet/extensions/issues/6492; addresses 1. and 2.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6664)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -0,0 +1,70 @@
...
+    private readonly string _name;
+    private readonly string _description;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerTool"/> class.
+    /// </summary>
+    /// <param name="name">The name of the remote MCP server.</param>
+    /// <param name="url">The URL of the remote MCP server.</param>
+    /// <param name="description">The description of the remote MCP server.</param>
+    public HostedMcpServerTool(string name, Uri url, string? description = null)
```
stephentoub: How is description used? If we need it, can it instead just be a get/set property?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerToolApprovalMode"/> class.
+    /// </summary>
+    /// <param name="require">The list of tools that require approval for.</param>
+    /// <param name="notRequire">The list of tools that do not require approval for.</param>
+    public HostedMcpServerToolApprovalMode(IList<string> require, IList<string> notRequire)
+    {
+        Require = Throw.IfNull(require);
+        NotRequire = Throw.IfNull(notRequire);
+    }
```
stephentoub: How do these map to what various services support?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: > (If this is only for implementing the base Description property

Yes, is only for that.  OpenAI describe the field as:
> `server_description string Optional`
Optional description of the MCP server, used to provide more context.

RE:
> This maps to inputs used by Anthropic, OpenAI, and anyone else today that has remote MCP support?

I'm not aware of other services that support remote MCP.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <summary>
+    /// Gets the list of tools that require user approval for calls to a hosted MCP server.
+    /// </summary>
+    public IList<string>? Require { get; }
+
+    /// <summary>
+    /// Gets the list of tools that do not require user approval for calls to a hosted MCP server.
+    /// </summary>
+    public IList<string>? NotRequire { get; }
```
stephentoub: I assume these properties will return null for Always/Never, and as such we don't need to be concerned about them being mutated?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    private sealed class AlwaysApprovalMode : HostedMcpServerToolApprovalMode;
+
+    private sealed class NeverApprovalMode : HostedMcpServerToolApprovalMode;
```
stephentoub: Why are these needed?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
jozkee: Correct.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a tool call request to a hosted MCP server.
+/// </summary>
+[Experimental("MEAI001")]
+public class HostedMcpServerToolCallContent : AIContent
```
stephentoub: It would be good to sync with @westey-m about the designs he's been working on for approval. This should likely fold in with that.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerToolCallContent"/> class.
+    /// </summary>
+    /// <param name="callId">The tool call ID.</param>
+    /// <param name="name">The tool name.</param>
+    /// <param name="serverName">The MCP server name.</param>
+    /// <param name="arguments">The arguments used for the tool call.</param>
+    public HostedMcpServerToolCallContent(string callId, string name, string serverName, IReadOnlyDictionary<string, object?>? arguments = null)
```
stephentoub: FunctionCallContent uses IDictionary rather than IReadOnlyDictionary. Should this match?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
jozkee: This maps to `require_approval` in OpenAI Responses. which accepts object or string. 
The string can be 'always' or 'never', to signal the handling for all tools.
The object can be `{"always": { "tool_names": [*the tools*] }, "never": { "tool_names": [*more tools*] } }`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a tool call request to a hosted MCP server.
+/// </summary>
+[Experimental("MEAI001")]
```
stephentoub: I'm not sure we can really use Experimental successfully here. These types are only viable if other libraries (the IChatClient implementations) depend on them, but if they depend on them, and we make a binary breaking change, whether or not it's Experimental won't actually help. Effectively this is viral, so only experimental IChatClient implementations would be able to use this.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
jozkee: I don't think this is meant to be mutated, this just represents what the AI service used as arguments when invoking the tool from the MCP server, that's already been done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
stephentoub: > Optional description of the MCP server, used to provide more context.

Ok, so it is sent to the service.

I think we should avoid having it in the ctor, though. I'm expecting in the future there will be other things we'll want to optionally support, and presumably for those we'll want to do so with get/set properties rather than adding more longer and longer ctors.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
jozkee: Is there a command I need to run when adding new types?  I was getting warnings (asking me to use `Experimental`).
I think I need to update https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
stephentoub: Ok. What happens with tools that aren't in either the always or never list?

jozkee: It requires approval, which makes sense because that's the default.

stephentoub: > because that's the default.

The default for who?

> Why are these needed?

My question was really why are these derived types needed. What's type checking against these private types?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
rogerbarreto: > I think we should avoid having it in the ctor, though

Have the same opinion here for the description and this IMO applies across for any optional properties.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
rogerbarreto: The current approach having a different property for listing Required / NotRequired also with the mode at the same time could be misused like: Providing Never or Always when also providing Never/Always Specific tools respectively. 

i.e:
```
tool.Required = ["tool-example"]
tool.ApprovalMode = AlwaysApprovalMode();

tool.NotRequired = ["tool-example"]
tool.ApprovalMode = NeverApprovalMode();
```

I would use an approach a bit similar to how `ChatToolMode` was conveyed moving the caller to have a similar experience, something I initially suggested here:

https://github.com/rogerbarreto/agent-framework/blob/issues/200-add-support-mcp-azure-foundry-agents/dotnet/samples/GettingStarted/External/MEAI.Abstractions/MCP/ApprovalMode.cs
 
```csharp
[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
[JsonDerivedType(typeof(NeverApprovalMode), typeDiscriminator: "never")]
[JsonDerivedType(typeof(AlwaysApprovalMode), typeDiscriminator: "always")]
#pragma warning disable CA1052 // Static holder types should be Static or NotInheritable
#pragma warning disable RCS1102 // Make class static
public class ApprovalMode
#pragma warning restore RCS1102 // Make class static
#pragma warning restore CA1052
{
    public static NeverApprovalMode Never { get; } = new NeverApprovalMode();
    public static AlwaysApprovalMode Always { get; } = new AlwaysApprovalMode();

    public static AlwaysApprovalMode AlwaysSpecific(IReadOnlyCollection<string> toolNames)
        => new(toolNames);

    public static NeverApprovalMode NeverSpecific(IReadOnlyCollection<string> toolNames)
        => new(toolNames);

    public static MixedApprovalMode Specific(IReadOnlyCollection<string> alwaysApproveToolNames, IReadOnlyCollection<string> neverApproveToolNames)
        => new(alwaysApproveToolNames, neverApproveToolNames);
}
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
rogerbarreto: I see we had the `attributes` provided in the ctor for the `FunctionCallContent` but similar to my [comment here](https://github.com/dotnet/extensions/pull/6664#discussion_r2242330869), since they are optional I would move the arguments to be a `IReadOnlyDictionary` setter and not in the constructor or define what is our path moving forward for consistency.

> I don't think this is meant to be mutated, this just represents what the AI service used as arguments when invoking the tool from the MCP server, that's already been done.

I think this is a valid point and I also wonder why the FunctionCallContent can be mutated, in theory it shouldn't after creation.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Gets or sets the output of the tool call.
+    /// </summary>
+    public string? Output { get; set; }
+
+    /// <summary>
+    /// Gets or sets the error of the tool call, if any.
+    /// </summary>
+    public string? Error { get; set; }
```
rogerbarreto: We do have `ErrorContent` abstraction, can't we use this here?

Anthropic also has [an array of contents](https://docs.anthropic.com/en/docs/agents-and-tools/mcp-connector#mcp-tool-result-block): `"contents" : [{"type":"text","text":"Hello"}]` 



Would suggest the following:

See: https://github.com/rogerbarreto/agent-framework/blob/issues/200-add-support-mcp-azure-foundry-agents/dotnet/samples/GettingStarted/External/MEAI.Abstractions/MCP/HostedMcpServerToolResult.cs

```suggestion
    /// <summary>
    /// Gets or sets the output of the tool call.
    /// </summary>
    public IList<AIContent>? Outputs { get; set; }

    /// <summary>
    /// Gets or sets the error of the tool call, if any.
    /// </summary>
    public ErrorContent? Error { get; set; }
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
stephentoub: > I also wonder why the FunctionCallContent can be mutated

With FCC, it's the client doing the invocation. Middleware might examine the arguments and choose to tweak them before they're passed back, such as to remove PII, to fill in any missing information, to augment  the dictionary with additional state it knows the function may optionally want to examine, etc

rogerbarreto: > Middleware might examine the arguments and choose to tweak them before they're passed back

Indeed, good point, for MCP tooling though as @jozkee well noted, this won't happen, as the content just represent a call already done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
rogerbarreto: I proposed initially we could use this type with an boolean flag, as all the other structure for approval (coming from OpenAI backend is pretty much identical with the call), Anthropic currently doesn't support MCP approval requests.

```csharp
    // Currently only OpenAI supports proxying the approval request from server to client, and
    // all other properties are very similar to an tool call request, except for this one, which may
    // work well without introducing a new MCP tool result type.
    /// <summary>
    /// Determines whether the MCP tool call requires approval before execution.
    /// </summary>
    public bool ApprovalRequest { get; set; } = false;
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: Our options are:
1. Add another property `ServerDescription { get; set; }` (and maybe ServerName for symmetry).
2. Hide the base property with `new`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
jozkee: > My question was really why are these derived types needed. What's type checking against these private types?

@stephentoub Nothing is type-checking against them, I thought of using them for value-checking `if (myMcpTool.ApprovalMode == HostedMcpServerToolApprovalMode.Always)  do smething;`, although that can be achieved in multiple ways, not necessarily with derived types.
I can change them to public classes to enable type checks.

jozkee: @rogerbarreto sorry, I don't follow, there's no Mode property.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
westey-m: Yeah, I think the types that are here are handling the part where the MCP tools are defined and the need for approvals are configured.  The work I'm doing will contribute types to represent any approval requests generated by the service and approval responses generated by the caller, so I think these should compliment each other.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolApprovalMode.cs
rogerbarreto: > sorry, I don't follow, there's no Mode property.

Edited, where you see `Mode` understand as `ApprovalMode`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
rogerbarreto: > 1. Add another property ServerDescription { get; set; } (and maybe ServerName for symmetry).
Hide the base property with new.

Like this approach more, is more clear to understand the distinction between what is the `Tool` description|name from what is the server description|name

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
jozkee: > We do have ErrorContent abstraction, can't we use this here?

Error should be an `IList<AIContent>` too.  I just tested this case and openai returns an object of type mcp_tool_execution_error with a content array:
```json
{
  "id": "mcp_123",
  "type": "mcp_call",
  "approval_request_id": null,
  "arguments": "{}",
  "error": {
    "type": "mcp_tool_execution_error",
    "content": [
      {
        "type": "text",
        "text": "An error occurred invoking 'test_error'.",
        "annotations": null,
        "meta": null
      }
    ]
  }
```
Also, Anthropic error response has the same structure as the success response, it just sets `"is_error": true` in the `mcp_tool_result`.

jozkee: Or we could follow Anthropic's suit and have a single `IList<AIContent>` with a bool IsError property.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    /// Gets or sets the list of tool names that always require approval.
+    /// </summary>
+    public IList<string>? AlwaysRequireApprovalToolNames { get; set; }
+
+    /// <summary>
+    /// Gets or sets the list of tool names that never require approval.
+    /// </summary>
+    public IList<string>? NeverRequireApprovalToolNames { get; set; }
```
jozkee, line 36: Should these be `HashSet<string>`?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
```diff
@@ -0,0 +1,33 @@
...
+public sealed class HostedMcpServerToolRequireSpecificApprovalMode : HostedMcpServerToolApprovalMode
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerToolRequireSpecificApprovalMode"/> class that specifies approval behavior for individual tool names.
+    /// </summary>
+    /// <param name="alwaysRequireApprovalToolNames">The list of tools names that always require approval.</param>
+    /// <param name="neverRequireApprovalToolNames">The list of tools names that never require approval.</param>
+    public HostedMcpServerToolRequireSpecificApprovalMode(IList<string>? alwaysRequireApprovalToolNames, IList<string>? neverRequireApprovalToolNames)
+    {
+        AlwaysRequireApprovalToolNames = alwaysRequireApprovalToolNames;
```
eavanvalkenburg, line 24: at least for OpenAI responses api, these are mutually exclusive, you either pass `never` with tool names or `always` with tool names...

jozkee: That's not what the docs say nor what I observed in my testing.  Specifying both objects works fine.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
rogerbarreto: I just wonder what else would they provide in an array of contents for an error, first time I see such structure, 

Haven't thought Error having contents, maybe an opportunity for `Error.Items`. Or just allow multiple `ErrorContent`s to exist in the `Outputs` and the `Error` => `Outputs?.OfType<ErrorContent>().FirstOrDefault()`, thoughts?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
rogerbarreto: I think this make sense for this scenario as we would not want multiple functions with the same name,

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
jozkee: how about using a virtual Success property:
```cs
public virtual bool Success => Output is not null && 
    Output.Count is not 0 && 
    !Output.OfType<ErrorContent>().Any();
```
And then Anthropic could lift their concrete `is_error` result:
```cs
class ClaudeMcpToolResult(string callId, bool isError) : HostedMcpServerToolResultContent(callId)
{
    public override bool Success => !isError;
}
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
jozkee: I'm having second thoughts about using `HashSet<string>` because it may set a precedent and make existing API inconsistent.  If we do it here, you could argue `HostedMcpServerTool.AllowedTools` should be a set too, then, what about  `ChatOptions.StopSequences`?

I also see we don't have a single `HashSet<T>` nor `ISet<T>` API, we have [plenty](https://github.com/search?q=repo%3Adotnet%2Fextensions+IList%3C+path%3Asrc%2FLibraries%2F*AI*%2F**%2F*.cs&type=code) of `IList<T>`s though.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
```diff
@@ -0,0 +1,78 @@
...
+        int sum = 0;
+        foreach (string item in set)
+        {
+            int hash = set.Comparer.GetHashCode(item);
+            xor ^= hash;
+            sum += hash;
+        }
+
+        return HashCode.Combine(set.Comparer, xor, sum);
+    }
```
jozkee: I provided custom `GetHashCode` and `Equals` implementations for `HostedMcpServerToolApprovalMode` since we are following ChatToolMode's suit. 

I'm pointing it out in here because it's quite cumbersome compared to its sibling types, as _I think_ we need to ensure order-insensitive equality between the two hashsets.

jozkee:
This should be ready for another round.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System.Collections.Generic;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a tool call request to a hosted MCP server.
+/// </summary>
+public class HostedMcpServerToolCallContent : AIContent
```
stephentoub: I don't think we need the "Hosted" prefix here. With the tool, the prefix is saying that the tool itself doesn't provide the implementation and it's just a marker, but here, this is actually conveying information about the tool call, regardless of where it came from.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerToolCallContent"/> class.
+    /// </summary>
+    /// <param name="callId">The tool call ID.</param>
+    /// <param name="toolName">The tool name.</param>
+    /// <param name="serverName">The MCP server name.</param>
```
stephentoub: Let's add the `<exception/>` tags for the exceptions that might be thrown, here and elsewhere in the PR.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents the result of a hosted MCP server tool call.
+/// </summary>
+public class HostedMcpServerToolResultContent : AIContent
```
stephentoub: Ditto

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
```diff
@@ -0,0 +1,46 @@
...
+    public HostedMcpServerToolCallContent(string callId, string toolName, string serverName)
```
stephentoub: How does this fit in with approvals?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Gets or sets the output of the tool call.
+    /// </summary>
+    public IList<AIContent>? Output { get; set; }
+
+    /// <summary>
+    /// Gets a value indicating whether the tool call was successful.
+    /// </summary>
+    public virtual bool Success => Output is not null && Output.Count is not 0 && !Output.OfType<ErrorContent>().Any();
```
stephentoub: Do we need this property?

And why is it virtual?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolAlwaysRequireApprovalMode.cs
```diff
@@ -0,0 +1,22 @@
...
+
+/// <summary>
+/// Indicates that approval is always required for tool calls to a hosted MCP server.
+/// </summary>
+/// <remarks>
+/// Use <see cref="HostedMcpServerToolApprovalMode.AlwaysRequire"/> to get an instance of <see cref="HostedMcpServerToolAlwaysRequireApprovalMode"/>.
+/// </remarks>
+[DebuggerDisplay(nameof(AlwaysRequire))]
+public sealed class HostedMcpServerToolAlwaysRequireApprovalMode : HostedMcpServerToolApprovalMode
+{
```
stephentoub, line 18: We should give this an explicit public parameterless ctor so that we can add XML comments to it. I believe elsewhere in such cases those XML comments discourage direct use of the constructor and instead point folks to use the singleton property.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
```diff
@@ -0,0 +1,78 @@
...
+/// Represents a mode where approval behavior is specified for individual tool names.
+/// </summary>
+public sealed class HostedMcpServerToolRequireSpecificApprovalMode : HostedMcpServerToolApprovalMode
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerToolRequireSpecificApprovalMode"/> class that specifies approval behavior for individual tool names.
+    /// </summary>
+    /// <param name="alwaysRequireApprovalToolNames">The list of tools names that always require approval.</param>
+    /// <param name="neverRequireApprovalToolNames">The list of tools names that never require approval.</param>
+    public HostedMcpServerToolRequireSpecificApprovalMode(HashSet<string>? alwaysRequireApprovalToolNames, HashSet<string>? neverRequireApprovalToolNames)
```
stephentoub: All of these `HashSet`s should be `ISet`s

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        AlwaysRequireApprovalToolNames = alwaysRequireApprovalToolNames;
+        NeverRequireApprovalToolNames = neverRequireApprovalToolNames;
+    }
+
+    /// <summary>
+    /// Gets or sets the list of tool names that always require approval.
```
stephentoub, line 29: list => set here and elsewhere sets are used

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
stephentoub: Is any of this needed? Can we just use reference equality on the instance itself? What's the scenario where deep value equality is required?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Name": "Microsoft.Extensions.AI.Abstractions, Version=9.6.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
+  "Name": "Microsoft.Extensions.AI.Abstractions, Version=9.9.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
```
stephentoub: Is this necessary?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -0,0 +1,62 @@
...
+/// Represents a hosted MCP server tool that can be specified to an AI service.
+/// </summary>
+public class HostedMcpServerTool : AITool
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerTool"/> class.
+    /// </summary>
+    /// <param name="serverName">The name of the remote MCP server.</param>
+    /// <param name="url">The URL of the remote MCP server.</param>
+    public HostedMcpServerTool(string serverName, Uri url)
```
stephentoub, line 36: We might want a convenience constructor that accepts a string url and just delegates to this

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -0,0 +1,62 @@
...
+    public Uri Url { get; }
+
+    /// <summary>
+    /// Gets or sets the description of the remote MCP server, used to provide more context to the AI service.
+    /// </summary>
+    public string? ServerDescription { get; set; }
+
+    /// <summary>
+    /// Gets or sets the list of tools allowed to be used by the AI service.
+    /// </summary>
```
stephentoub, line 59: The comment should specify behavior when the list is null. Presumably that's different from empty, where null means everything is allowed but empty means nothing is allowed?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -0,0 +1,62 @@
...
+    public IList<string>? AllowedTools { get; set; }
+
+    /// <summary>
+    /// Gets or sets the approval mode that indicates when the AI service should require user approval for tool calls to the remote MCP server.
+    /// </summary>
+    /// <remarks>
+    /// You can set this property to <see cref="HostedMcpServerToolApprovalMode.AlwaysRequire"/> to require approval for all tool calls, 
+    /// or to <see cref="HostedMcpServerToolApprovalMode.NeverRequire"/> to never require approval.
+    /// </remarks>
```
stephentoub, line 79: This should specify what the default is / what it means for this to be null.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
stephentoub: You need to augment AIContent's polymorphism attributes, too, right?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
stephentoub, line 1: There are other tools that you've not moved into the Tools folder, right? e.g. HostedFileSearchTool? What about AIFunction?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
jozkee: > Do we need this property?

I added it to provide a reliable way of determining if the result was successful.  Are you suggesting that we should not do that?

> why is it virtual?

I wanted to let users override this property to lift clearer indicators, like Claude's isError property, when possible.

OpenAI responds with an object of type error, Claude uses an isError property.  Both, success and error responses would put their content in the Output, presumably putting ErrorContents when the provider responds with an error, hence the virtual implementation.

Also see https://github.com/dotnet/extensions/pull/6664#discussion_r2250332780

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
jozkee: Do you have any thoughts on https://github.com/dotnet/extensions/pull/6664#discussion_r2283931629?

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
stephentoub: We shouldn't use `HashSet<T>`; we want to avoid using concrete types and make it flexible for arbitrary implementations to be used. I'd be fine with `IList<T>`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
jozkee: We could remove it.  I added it because `RequiredChatToolMode` asserts serialization roundtrip and does something similar in its overrides (less convoluted since it only has one string property), is it ok to assume this type here doesn't need to comply with that?

https://github.com/dotnet/extensions/blob/20349f88a5e8fb5a3dd0aa8a1f7920ebb5c75f9d/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatToolModeTests.cs#L85-L93

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
jozkee: No, was changed automatically by `MakeApiBaselines.ps1`, will undo.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: Good point, yes, empty means nothing is allowed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolResultContent.cs
stephentoub: > I added it to provide a reliable way of determining if the result was successful. Are you suggesting that we should not do that?

I don't think we need it, and we can always add it later if we do. Let's keep things as minimal as possible until we learn more.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
stephentoub: I responded there as well, but I'd be fine with either `IList<>` or `ISet<>`.

src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedMcpServerToolRequireSpecificApprovalMode.cs
stephentoub: I think reference equality is ok. I'd be surprised if anyone had a need for storing these as keys in a dictionary and looking them up based on these approval lists being configured. That said, if you switch to `IList<>` and wanted to encode a simple value-based equality of just doing a sequence equals on the two lists, that'd be ok.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
jozkee: I think we would have the request/response types very similar to FunctionApprovalRequest/ResponseContents from https://github.com/dotnet/extensions/pull/6745.

A provider like MEAI.OpenAI would generate the request based on the service's approval request and float it to the user where they will generate the approval response and then send it.  Then, the service will return the mcp tool call, and finally the provider library will wrap it with this type.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/HostedMcpServerToolCallContent.cs
jozkee: Yes, missed that.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: I would prefer keeping AIFuncion in the Functions folder, seems to me it deserves its own directory at this level, otherwise, we would have to have Tools/Functions.  Aside from that, I only see HostedFileSearchTool, which I prefer we move to Tools too.

 HostedFileSearchTool was merged after I opened this PR, and I must've missed them while rebasing.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientIntegrationTests.cs
```diff
@@ -49,4 +49,128 @@ public async Task UseWebSearch_AnnotationsReflectResults()
...
+
+                approvalsCount += approvalResponse.Contents.Count;
+                input.Add(approvalResponse);
+                response = streaming ?
+                    await client.GetStreamingResponseAsync(input, chatOptions).ToChatResponseAsync() :
+                    await client.GetResponseAsync(input, chatOptions);
+            }
+
+            // Validate final response
+            Assert.Equal(2, approvalsCount);
```
jozkee, line 172: I recall this not being reliably.  Sometimes the model will go straight to call ask_question.

stephentoub: We can change the test if it proves unstable. I ran it a bunch and it always did this, but we can remove the check if it sporadically fails. These integration tests, for better or worse (mostly worse), are not run automatically or anywhere in CI today.

jozkee:
Thanks, @stephentoub.

PR #6666, opened 2025-07-30
Add middleware for reducing chat history

### LINKED ISSUE #6647: [AI] Middleware for reducing chat history
The AI Chat Web template demonstrates a common scenario where a long-running chat can lead to the scenario of needing to reduce the chat history for continued exchanges with the model.

To address this common pitfall, we can implement a piece of middleware in Microsoft.Extensions.AI for reducing the chat history that is exchanged. We need to design the API for this middleware to allow for configuration of how the history is reduced.

Once we have the middleware API, we would add it into the AI Chat Web template and/or add it into complementary docs and samples.

MackinnonBuck:
## Summary

Adds chat client middleware to reduce chat history.

Fixes #6647

## Description

The design is based primarily on the existing `ReducingChatClient` defined in our integration tests.

It might also be worth implementing some chat reducers that we expect to be commonly used (e.g., one that truncates chat history, or one that uses a chat client to summarize chat history into a smaller number of messages). Even if we don't ship these, it would at least validate the design.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6666)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/IChatReducer.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/IChatReducer.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a reducer capable of shrinking the size of a list of chat messages.
+/// </summary>
+[Experimental("MEAI001")]
+public interface IChatReducer
+{
+    /// <summary>Reduces the size of a list of chat messages.</summary>
+    /// <param name="messages">The messages to reduce.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>The new list of messages.</returns>
+    Task<IList<ChatMessage>> ReduceAsync(IEnumerable<ChatMessage> messages, CancellationToken cancellationToken);
+}
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClient.cs
@@ -1,0 +1,50 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// A chat client that reduces the size of a message list.
+/// </summary>
+[Experimental("MEAI001")]
+public sealed class ReducingChatClient : DelegatingChatClient
+{
+    private readonly IChatReducer _reducer;
+
+    /// <summary>Initializes a new instance of the <see cref="ReducingChatClient"/> class.</summary>
+    /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
+    /// <param name="reducer">The reducer to be used by this instance.</param>
+    public ReducingChatClient(IChatClient innerClient, IChatReducer reducer)
+        : base(innerClient)
+    {
+        _reducer = Throw.IfNull(reducer);
+    }
+
+    /// <inheritdoc />
+    public override async Task<ChatResponse> GetResponseAsync(
+        IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        messages = await _reducer.ReduceAsync(messages, cancellationToken).ConfigureAwait(false);
+
+        return await base.GetResponseAsync(messages, options, cancellationToken).ConfigureAwait(false);
+    }
+
+    /// <inheritdoc />
+    public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
+        IEnumerable<ChatMessage> messages, ChatOptions? options = null, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        messages = await _reducer.ReduceAsync(messages, cancellationToken).ConfigureAwait(false);
+
+        await foreach (var update in base.GetStreamingResponseAsync(messages, options, cancellationToken).ConfigureAwait(false))
+        {
+            yield return update;
+        }
+    }
+}
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClientBuilderExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClientBuilderExtensions.cs
@@ -1,7 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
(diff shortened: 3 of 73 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClient.cs
```diff
@@ -0,0 +1,50 @@
...
+[Experimental("MEAI001")]
+public sealed class ReducingChatClient : DelegatingChatClient
+{
+    private readonly IChatReducer _reducer;
+
+    /// <summary>Initializes a new instance of the <see cref="ReducingChatClient"/> class.</summary>
+    /// <param name="innerClient">The underlying <see cref="IChatClient"/>, or the next instance in a chain of clients.</param>
+    /// <param name="reducer">The reducer to be used by this instance.</param>
+    public ReducingChatClient(IChatClient innerClient, IChatReducer reducer)
+        : base(innerClient)
```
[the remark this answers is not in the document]
MackinnonBuck: The base constructor actually does guarantee null validation.

stephentoub:
Thanks!

Can you please sync offline with @crickman and @westey-m? I want to make sure we're factoring in experiences with implementing similar concepts in SK and as part of various agentic libraries.

I also want to make sure we've thought through the various scenarios where this will be relevant. With a tokenizer-based reducer, or something that can be implemented entirely client side, this approach seems reasonable. I'm wondering about other needs, like submitting the conversation history or parts of it to an LLM for it to summarize and then replace that part of the conversation... that's not going to fit well into an IChatClient like this, as you'd need to do it as part of every request, rather than persisting the changes and maintaining a constantly summarized history or the like.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClientBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+public static class ReducingChatClientBuilderExtensions
+{
+    /// <summary>
+    /// Adds a <see cref="ReducingChatClient"/> to the chat pipeline.
+    /// </summary>
+    /// <param name="builder">The <see cref="ChatClientBuilder"/> being used to build the chat pipeline.</param>
+    /// <param name="reducer">An optional <see cref="IChatReducer"/> to apply to the chat client. If not supplied, an instance will be resolved from the service provider.</param>
+    /// <param name="configure">An optional callback that can be used to configure the <see cref="ReducingChatClient"/> instance.</param>
+    /// <returns>The configured <see cref="ChatClientBuilder"/> instance.</returns>
+    public static ChatClientBuilder UseChatReducer(
```
stephentoub, line 24: As written, UseChatReducer and ChatReducingChatClient don't actually add all that much over just using the existing Use method:
```C#
IChatReducer reducer = ...;
...
.Use(async (messages, options, nextAsync, cancellationToken) =>
{
    await nextAsync(await reducer.ReduceAsync(messages), options, cancellationToken);
})
```
Probably still makes sense to have them, but I wonder where we'd draw the line.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/IChatReducer.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a reducer capable of shrinking the size of a list of chat messages.
+/// </summary>
+[Experimental("MEAI001")]
+public interface IChatReducer
```
crickman: Are there any implementations of `IChatReducer` that I'm missing?

MackinnonBuck: There's a basic test implementation [here](https://github.com/dotnet/extensions/blob/f94902c64f76ba760b32c91c12b47847924ad47b/test/Libraries/Microsoft.Extensions.AI.Integration.Tests/ReducingChatClientTests.cs#L60) based on token counting, but I do plan to add other implementations to test the flexibility of the API.

I see that SK has a concept of an `IChatHistoryReducer` that can be used to reduce chat history directly and separately from normal chat completion. As @stephentoub pointed out, the current implementation is limiting in that it doesn't actually mutate the running chat history, so a summarizing reducer might not be easy to implement. We'd ideally want a way to augment the history directly, which means a middleware alone might not be enough.

crickman: Ok, sure, but this is test code.  No flavors of `IChatHistoryReducer` that work "out of the box" for customers then?

MackinnonBuck: Not yet, but it's something I've been considering, either as part of this PR or in a follow-up. I called it out in the PR description as an area worth exploring, but I wanted to start with a proposal based on an existing design (see deleted code in the PR diff) and then refine things based on feedback. Once the design solidifies a bit more, I do plan to include a few out-of-the-box implementations.

MackinnonBuck:
Looking at how Semantic Kernel handles chat history reduction, I think the API in this PR should support implementing summarization reducers that can persist summaries between requests by embedding them in the chat history itself. This appears to be similar to Semantic Kernel's [`ChatHistorySummarizationReducer`](https://github.com/microsoft/semantic-kernel/blob/59b4e8a77cfe89ede195bf91c2c1d8a05aa83670/dotnet/src/SemanticKernel.Core/AI/ChatCompletion/ChatHistorySummarizationReducer.cs).

We should also consider exposing the built-in reducers as standalone components (not just middleware) so applications can implement their own heuristics for when to reduce history, rather than doing it on every request.

I think the next step is to implement some concrete reducers to validate the API design.

stephentoub:
> Looking at how Semantic Kernel handles chat history reduction, I think the API in this PR should support implementing summarization reducers that can persist summaries between requests by embedding them in the chat history itself. This appears to be similar to Semantic Kernel's [ChatHistorySummarizationReducer](https://github.com/microsoft/semantic-kernel/blob/59b4e8a77cfe89ede195bf91c2c1d8a05aa83670/dotnet/src/SemanticKernel.Core/AI/ChatCompletion/ChatHistorySummarizationReducer.cs).

Note that there's a subtle but important difference here from SK. With IChatClient, the input messages list is an IEnumerable that's effectively immutable. With SK's IChatCompletionService, the input messages list is an IList that's mutable. We made a very conscious choice to switch to the immutable model, such that all new messages are returned as part of a ChatResponse, rather than adding some of the messages to the chat history and then only returning the last one. But it does make cases like this one harder.

MackinnonBuck:
> We made a very conscious choice to switch to the immutable model, such that all new messages are returned as part of a ChatResponse, rather than adding some of the messages to the chat history and then only returning the last one. But it does make cases like this one harder.

I was thinking that the summary would be appended to the chat list (e.g., via `ChatResponseExtensions.AddMessages()`) just like any other message, except it wouldn't be displayed in the app's UI. On the next call to `GetResponseAsync()`, the `ReducingChatClient` would construct a modified message list that excludes previously-summarized chat messages but keeps the latest summary.

stephentoub:
> > We made a very conscious choice to switch to the immutable model, such that all new messages are returned as part of a ChatResponse, rather than adding some of the messages to the chat history and then only returning the last one. But it does make cases like this one harder.
> 
> I was thinking that the summary would be appended to the chat list (e.g., via `ChatResponseExtensions.AddMessages()`) just like any other message, except it wouldn't be displayed in the app's UI. On the next call to `GetResponseAsync()`, the `ReducingChatClient` would construct a modified message list that excludes previously-summarized chat messages but keeps the latest summary.

Interesting. Worth pursuing. Devil will probably be in the details for how these things are represented in a way that makes them distinct. And will probably need some helpers that would enable a developer to clean up the resulting chat history.

crickman:
The original thinking in the SK reducer was that the reducer would provide an updated enumeration that was entirely consistent without mutating the original history.  Then the developer could determine if they desired replace or mutate the original chat-history.  Some utility extensions were provided to make opt-in mutation or replacement easy.

One advantage is that this avoided a large re-allocation during reduction (until the developer expressed their intent with the enumeration).

MackinnonBuck:
I just added a prototype implementation of a summarizing reducer that works by embedding summaries in the `AdditionalProperties` of chat messages. If a chat message has a summary in its `AdditionalProperties`, that summary encapsulates the entire conversation leading up to and including that message. This enables the following:
* Incremental summary generation: New summaries are able to be generated by finding the last generated summary and producing a new summary that incorporates newer chat messages.
* Conversation preservation: For apps that use the message list as the "source of truth" for a chat UI (as the chat template does), a destructive replacement of the chat history might not be ideal. The prototyped approach allows preserving chat history without sacrificing incremental summary generation.
* Summary trail: If a chat application supports editing and re-submitting a previous message (i.e., "forking" the conversation), the summarizer would automatically use the latest summary available from that point in the conversation.

The implementation I've put in this PR is a demonstration of this approach. Happy to consider alternatives. If we do pursue this implementation strategy, there's more I'll need to follow-up on:
* Helpers to clean up the chat history (replacing the existing chat history with a summarized one, removing summary metadata, etc.)
* Summarizer configuration, such as customizing the system prompt
* Integration tests
* Incorporation into the chat template, if we decide to do that

MackinnonBuck:
These are now done:
* Helpers to clean up the chat history
  * You can do this by using an `IChatReducer` "directly" instead of adding it to the middleware pipeline
* Customization of the summarization prompt in `SummarizingChatReducer`
* Integration tests

Follow-ups to consider:
* Integration with the AI Chat Web template. I've done this locally, but we need to decide if this is what we want by default.
* Which other `IChatReducer` implementations we should support out of the box.

MackinnonBuck:
> I want to make sure we're factoring in experiences with implementing similar concepts in SK and as part of various agentic libraries.

On the topic of agentic libraries: I looked into this, and the proposed abstractions and chat summarizer implementation in this PR seem like a natural fit. If multiple agents are participating in the same conversation, it may make more sense to use an `IChatReducer` directly to reduce the shared conversation, rather than having each agent use an `ReducingChatClient` that performs its own reduction within its middleware pipeline.

That said, if the goal is for each agent to handle its own reduction (or opt out entirely) the middleware approach should still work, since the message list isn't destructively mutated. The one caveat is with `SummarizingChatReducer`, which mutates `ChatMessage.AdditionalProperties` to store summaries between requests. If multiple chat clients each use their own `SummarizingChatReducer` with different configurations (e.g., different summarization instructions), they could end up stepping over each other, because the last summary in the message list becomes the basis for the next one.

If that scenario is a concern, we could consider making the `AdditionalProperties` key configurable in `SummarizingChatReducer`. That would allow multiple summaries to coexist in the same message list, where each one is tailored to the needs of a specific agent.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ReducingChatClientBuilderExtensions.cs
MackinnonBuck: Yeah, it doesn't add much currently. I'm guessing we could remove this in the future if we wanted to, given these APIs are currently experimental.

jeffhandley (approved):

PR #6668, opened 2025-07-31
Remove AdjustTime from experimental

### LINKED ISSUE #5072: [API Proposal]: FakeTimeProvider, option to allow going backwards

amadeuszl:
Fixes #5072

### DIFF
--- a/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
+++ b/src/Libraries/Microsoft.Extensions.TimeProvider.Testing/FakeTimeProvider.cs
@@ -137,7 +137,7 @@ public class FakeTimeProvider : TimeProvider
     }
 
     /// <summary>
-    /// Advances the date and time in the UTC time zone.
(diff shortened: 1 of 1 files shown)


eduherminio (approved):

dariusclay:
Looks like that fixed it

PR #6694, opened 2025-08-07
Support keyed HybridCache with keyed DistributedCaches and named options

ylt1534:
Closes dotnet/runtime#117976

Adds support for keyed `HybridCache` instances in DI, along with the ability to configure the `DistributedCacheServiceKey` for the hybrid cache to resolve a keyed `IDistributedCache` service. This opens up the scenario of having multiple hybrid caches within an application, each potentially using a separate backend cache. Those backend caches could be separate databases or entirely separate services. For example, having a Redis-backed cache alongside a SQL Server-backed cache.

```c#
    [Fact]
    public void CanCreateRedisAndSqlServerBackedHybridCaches()
    {
        services.AddKeyedSingleton<IDistributedCache, RedisCache>("Redis");

        services.AddKeyedSingleton<IDistributedCache, SqlServerCache>("SqlServer",
            (sp, key) => new SqlServerCache(new SqlServerCacheOptions
            {
                ConnectionString = "test",
                SchemaName = "test",
                TableName = "test"
            }));

        services.AddKeyedHybridCache("HybridWithRedis", options => options.DistributedCacheServiceKey = "Redis");
        services.AddKeyedHybridCache("HybridWithSqlServer", options => options.DistributedCacheServiceKey = "SqlServer");

        using ServiceProvider provider = services.BuildServiceProvider();
        var hybridWithRedis = Assert.IsType<DefaultHybridCache>(provider.GetRequiredKeyedService<HybridCache>("HybridWithRedis"));
        var hybridWithRedisBackend = Assert.IsType<RedisCache>(hybridWithRedis.BackendCache);
        Assert.Same(hybridWithRedisBackend, provider.GetRequiredKeyedService<IDistributedCache>("Redis"));

        var hybridWithSqlServer = Assert.IsType<DefaultHybridCache>(provider.GetRequiredKeyedService<HybridCache>("HybridWithSqlServer"));
        var hybridWithSqlServerBackend = Assert.IsType<SqlServerCache>(hybridWithSqlServer.BackendCache);
        Assert.Same(hybridWithSqlServerBackend, provider.GetRequiredKeyedService<IDistributedCache>("SqlServer"));
    }
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6694)


jeffhandley:
I've marked this PR as a draft as we need to get [[Support detect keyed dependency injected IDistributedCache as backend cache for DefaultHybridCache (#117976)](https://github.com/dotnet/runtime/issues/117976)](https://github.com/dotnet/runtime/issues/117976) through API review, and tests would need to be added to the PR as well.

I'm getting a .NET 10 RC1/RC2 bar check for this. I'll keep you posted, @ylt1534.

kelly-yinn:

kelly-yinn:
Thanks @jeffhandley  for the review. I just noticed that the DefaultHybridCache is designed as internal thus unable to DI is outside the lib. I added a new extension method and add some UTs. I want to test that the keyed instances (options, backend cache, memory cache, hybrid cache) are injected into each other correctly, but that might bring more changes like expose the _backendCache and _localCache properties from DefaultHybridCache.  For now I tried to make the PR small to be easy review. thanks.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheServiceExtensions.cs
```diff
@@ -41,4 +44,36 @@ public static IHybridCacheBuilder AddHybridCache(this IServiceCollection service
...
+        // Register keyed in-memory cache
+        services.TryAddKeyedSingleton<IMemoryCache, MemoryCache>(name);
+
+        // Register keyed HybridCache
+        services.TryAddKeyedSingleton<HybridCache>(name, (sp, _) =>
+            new DefaultHybridCache(
+                Options.Options.Create(sp.GetRequiredService<IOptionsMonitor<HybridCacheOptions>>().Get(name)),
+                sp,
+                sp.GetKeyedService<IMemoryCache>(name),
+                sp.GetKeyedService<IDistributedCache>(name)));
```
jeffhandley: Requiring that the `name` for this hybrid cache must match the name of both the `IMemoryCache` and `IDistributedCache` keyed services is somewhat limiting. I'm exploring a different approach here that would support each of the names being distinct from each other.

eerhardt:
Is this going to be an issue?

https://github.com/dotnet/aspnetcore/blob/6f91a7d7be082aa1246f41a884b182637942d5a4/src/Caching/StackExchangeRedis/src/RedisCacheImpl.cs#L16-L17

```C#
    internal override bool IsHybridCacheActive()
        => _services.GetService<HybridCache>() is not null;
```

this code assumes that if there is an unkeyed `HybridCache` in the service collection, then it should add the `HC` library name suffix:

https://github.com/dotnet/aspnetcore/blob/6f91a7d7be082aa1246f41a884b182637942d5a4/src/Caching/StackExchangeRedis/src/RedisCache.cs#L381-L383

jeffhandley:
@eerhardt and I discussed the `IsHybridCacheActive` scenario noted above and concluded it's not a blocker here. The implementation of that logging/telemetry is not compatible with having keyed HybridCache services registered, but it's also producing the incorrect result if an application uses a Redis cache and _also_ uses a `HybridCache` that _isn't_ backed by Redis. That design issue and bug should be addressed separately and in a way that is compatible with keyed hybrid caches.

src/Libraries/Microsoft.Extensions.Caching.Hybrid/HybridCacheServiceExtensions.cs
```diff
@@ -20,56 +18,107 @@ public static class HybridCacheServiceExtensions
...
+    /// Adds the services required for hybrid caching.
+    /// </summary>
+    /// <param name="services">The <see cref="IServiceCollection"/> to prepare with <see cref="HybridCache"/> prerequisites.</param>
+    /// <returns>A builder instance that allows further configuration of the <see cref="HybridCache"/> service.</returns>
+    private static HybridCacheBuilder PrepareServices(IServiceCollection services)
+    {
+        _ = Throw.IfNull(services);
+
+        services.TryAddSingleton(TimeProvider.System);
+        _ = services.AddOptions().AddMemoryCache();
```
kelly-yinn, line 121: Want to double confirm. So for local(memory) cache, no matter how many DistributedCache we injected, there will be only one local cache be injected right? And when we set local cache entry through diff Keyed HybridCache interface, the  LocalCacheExpiration and MaximumPayloadBytes in conrresponding Keyed options will set constraints on the cache entry separately?

For example, we injected 3 hybridCache instances: 

1. k1 hybridCache (with k1 options where MaximumPayloadBytes configured as 10KB)
2. k2 hybridCache (with k2 options where MaximumPayloadBytes configured as 100KB)
3. k3 hybridCache (with k3 options where MaximumPayloadBytes configured as 1000KB)

Now if i try to set cache entry with size as 500KB, i can succeed setting only through k3 hybridCache, right? Is that true, can we show the behavior in UT also?

jeffhandley: That's correct, @ylt1534. I added a test to illustrate keyed hybrid caches sharing a single memory cache. Thanks for the suggestion on that.

test/Libraries/Microsoft.Extensions.Caching.Hybrid.Tests/ServiceConstructionTests.cs
```diff
@@ -81,6 +296,107 @@ public void CanParseOptions_WithEntryOptions() // in particular, check we can pa
...
+
+        provider.GetRequiredKeyedService<HybridCache>(typeof(CustomMemoryDistributedCache2));
+        provider.GetRequiredKeyedService<HybridCache>(typeof(CustomMemoryDistributedCache2));
+        provider.GetRequiredKeyedService<HybridCache>(typeof(CustomMemoryDistributedCache2));
+    }
+
+    [Fact]
+    public void CanCreateKeyedServicesWithKeyedDistributedCaches_UsingSetupActions()
+    {
+        var source = new JsonConfigurationSource { Path = "BasicConfig.json" };
```
sebastienros: This made reading the test harder for me as I could not see where `"DistributedOne"` was set in the options. Maybe use an in-memory configuration and define it inline here. I don't think it makes it any easier to use a file based config.

For instance https://github.com/dotnet/aspire/blob/f528ede3c5ed78f73cf49328710dd4ec952e03a1/tests/Aspire.Azure.AI.Inference.Tests/AspireAzureAIInferenceExtensionTests.cs#L48-L50

eerhardt:
image

Why does this show up as a binary file?

jeffhandley:
> Why does this show up as a binary file?

It seems these ApiChief baseline JSON files, this one in particular, can match linguist's heuristic for long strings of content that cross a threshold for being identified as binary. I confirmed there's no BOM, no strange encoding or `\0` characters, or anything else out of the ordinary. I tried adding a `.gitattributes` entry for `*.json linguist-language=JSON` to [force an override](https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#summary), but that didn't have any effect. Viewing the file outside the diff view correctly shows it as JSON text. 🤷

jeffhandley (approved):

Issue #6688, opened 2025-08-05
Let not-found functions pass through FunctionInvokingChatClient


jozkee:
FICC today creates a FunctionResultContent with result "Error: Requested function \"{result.CallContent.Name}\" not found".
https://github.com/dotnet/extensions/blob/2fbb77643383db36bfd0e55e6923187491fcbfc1/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L812
For handling approvals on the McpChatClient (https://github.com/dotnet/extensions/issues/6492 (4)), we will need to modify FICC to allow FunctionCallContent to pass through if the function wasn't found.

This would be a breaking change.

cc @stephentoub

stephentoub:
> This would be a breaking change.

It'll be a functional change, but I think the current behavior is arguably a bug and it's unlikely folks are hitting this case.

PR #6695, opened 2025-08-07
Split AIFunction into a base class

stephentoub:
- Adds `AIFunctionDefinition` as a base class for `AIFunction`, where `AIFunction` layers on invocability over the metadata exposed by the base.
- Added `AIFunctionFactory.CreateDefinition` and `AIFunction.AsDefinitionOnly` for creating `AIFunctionDefinition`s that aren't invocable.
- Updated `FunctionInvokingChatClient` to allow non-`AIFunction` call requests to pass through.
- Added `FunctionInvokingChatClient.TerminateOnUnknownCalls` to control behavior when requests are made to unknown functions.

Closes https://github.com/dotnet/extensions/issues/6688

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6695)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
@@ -15,17 +15,17 @@ namespace Microsoft.Extensions.AI;
 public sealed class RequiredChatToolMode : ChatToolMode
 {
     /// <summary>
-    /// Gets the name of a specific <see cref="AIFunction"/> that must be called.
+    /// Gets the name of a specific tool that must be called.
     /// </summary>
     /// <remarks>
-    /// If the value is <see langword="null"/>, any available function can be selected (but at least one must be).
+    /// If the value is <see langword="null"/>, any available tool can be selected (but at least one must be).
     /// </remarks>
     public string? RequiredFunctionName { get; }
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="RequiredChatToolMode"/> class that requires a specific function to be called.
+    /// Initializes a new instance of the <see cref="RequiredChatToolMode"/> class that requires a specific tool to be called.
     /// </summary>
-    /// <param name="requiredFunctionName">The name of the function that must be called.</param>
+    /// <param name="requiredFunctionName">The name of the tool that must be called.</param>
     /// <exception cref="ArgumentException"><paramref name="requiredFunctionName"/> is empty or composed entirely of whitespace.</exception>
     /// <remarks>
     /// <paramref name="requiredFunctionName"/> can be <see langword="null"/>. However, it's preferable to use
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
@@ -6,44 +6,17 @@ using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 
+#pragma warning disable SA1202 // Elements should be ordered by access
+
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a function that can be described to an AI service and invoked.</summary>
-public abstract class AIFunction : AITool
+public abstract class AIFunction : AIFunctionDeclaration
 {
-    /// <summary>Gets a JSON Schema describing the function and its input parameters.</summary>
-    /// <remarks>
-    /// <para>
-    /// When specified, declares a self-contained JSON schema document that describes the function and its input parameters.
-    /// A simple example of a JSON schema for a function that adds two numbers together is shown below:
-    /// </para>
-    /// <code>
-    /// {
-    ///   "title" : "addNumbers",
-    ///   "description": "A simple function that adds two numbers together.",
-    ///   "type": "object",
-    ///   "properties": {
-    ///     "a" : { "type": "number" },
-    ///     "b" : { "type": "number", "default": 1 }
-    ///   }, 
-    ///   "required" : ["a"]
-    /// }
-    /// </code>
-    /// <para>
-    /// The metadata present in the schema document plays an important role in guiding AI function invocation.
-    /// </para>
-    /// <para>
-    /// When no schema is specified, consuming chat clients should assume the "{}" or "true" schema, indicating that any JSON input is admissible.
-    /// </para>
-    /// </remarks>
-    public virtual JsonElement JsonSchema => AIJsonUtilities.DefaultJsonSchema;
-
-    /// <summary>Gets a JSON Schema describing the function's return value.</summary>
-    /// <remarks>
-    /// A <see langword="null"/> typically reflects a function that doesn't specify a return schema
-    /// or a function that returns <see cref="void"/>, <see cref="Task"/>, or <see cref="ValueTask"/>.
-    /// </remarks>
-    public virtual JsonElement? ReturnJsonSchema => null;
+    /// <summary>Initializes a new instance of the <see cref="AIFunction"/> class.</summary>
+    protected AIFunction()
+    {
+    }
 
     /// <summary>
     /// Gets the underlying <see cref="MethodInfo"/> that this <see cref="AIFunction"/> might be wrapping.
@@ -72,4 +45,14 @@ public abstract class AIFunction : AITool
     protected abstract ValueTask<object?> InvokeCoreAsync(
         AIFunctionArguments arguments,
         CancellationToken cancellationToken);
+
+    /// <summary>Creates a <see cref="AIFunctionDeclaration"/> representation of this <see cref="AIFunction"/> that can't be invoked.</summary>
+    /// <returns>The created instance.</returns>
+    /// <remarks>
+    /// <see cref="AIFunction"/> derives from <see cref="AIFunctionDeclaration"/>, layering on the ability to invoke the function in addition
+    /// to describing it. <see cref="AsDeclarationOnly"/> creates a new object that describes the function but that can't be invoked.
+    /// </remarks>
+    public AIFunctionDeclaration AsDeclarationOnly() => new NonInvocableAIFunction(this);
+
+    private sealed class NonInvocableAIFunction(AIFunction function) : DelegatingAIFunctionDeclaration(function);
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -49,7 +49,7 @@ public static partial class AIFunctionFactory
     /// <para>
     /// By default, any parameters to <paramref name="method"/> are sourced from the <see cref="AIFunctionArguments"/>'s dictionary
     /// of key/value pairs and are represented in the JSON schema for the function, as exposed in the returned <see cref="AIFunction"/>'s
-    /// <see cref="AIFunction.JsonSchema"/>. There are a few exceptions to this:
+    /// <see cref="AIFunctionDeclaration.JsonSchema"/>. There are a few exceptions to this:
     /// <list type="bullet">
     ///   <item>
     ///     <description>
@@ -131,7 +131,7 @@ public static partial class AIFunctionFactory
     /// <para>
     /// Any parameters to <paramref name="method"/> are sourced from the <see cref="AIFunctionArguments"/>'s dictionary
     /// of key/value pairs and are represented in the JSON schema for the function, as exposed in the returned <see cref="AIFunction"/>'s
-    /// <see cref="AIFunction.JsonSchema"/>. There are a few exceptions to this:
+    /// <see cref="AIFunctionDeclaration.JsonSchema"/>. There are a few exceptions to this:
     /// <list type="bullet">
     ///   <item>
     ///     <description>
@@ -212,7 +212,7 @@ public static partial class AIFunctionFactory
     /// <para>
     /// By default, any parameters to <paramref name="method"/> are sourced from the <see cref="AIFunctionArguments"/>'s dictionary
     /// of key/value pairs and are represented in the JSON schema for the function, as exposed in the returned <see cref="AIFunction"/>'s
-    /// <see cref="AIFunction.JsonSchema"/>. There are a few exceptions to this:
+    /// <see cref="AIFunctionDeclaration.JsonSchema"/>. There are a few exceptions to this:
     /// <list type="bullet">
     ///   <item>
     ///     <description>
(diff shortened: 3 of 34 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Extensions.AI;
 public sealed class RequiredChatToolMode : ChatToolMode
 {
     /// <summary>
-    /// Gets the name of a specific <see cref="AIFunction"/> that must be called.
+    /// Gets the name of a specific function that must be called.
```
SteveSandersonMS: I don't feel 100% clear-minded on where we are with the tool/function distinction. If we wanted to be pedantically accurate, should this say "tool" here, or would it be incorrect for this to specify the name of a non-function tool?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -72,4 +45,10 @@ public abstract class AIFunction : AITool
     protected abstract ValueTask<object?> InvokeCoreAsync(
         AIFunctionArguments arguments,
         CancellationToken cancellationToken);
+
+    /// <summary>Creates a <see cref="AIFunctionDefinition"/> representation of this <see cref="AIFunction"/> that can't be invoked.</summary>
+    /// <returns>The created instance.</returns>
+    public AIFunctionDefinition AsDefinitionOnly() => new NonInvocableAIFunction(this);
```
SteveSandersonMS: Maybe I'll spot the reasoning later in the PR review, but I'm not certain at this point what the use case for this is.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/DelegatingAIFunctionDefinition.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Shared.Diagnostics;
+
+#pragma warning disable SA1202 // Elements should be ordered by access
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Provides an optional base class for an <see cref="AIFunctionDefinition"/> that passes through calls to another instance.
+/// </summary>
+public class DelegatingAIFunctionDefinition : AIFunctionDefinition
```
SteveSandersonMS: I'm not clear on the use case for this, except to support `AsDefinitionOnly`. Could it be `internal`?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaTransformCache.cs
```diff
@@ -57,7 +58,16 @@ public AIJsonSchemaTransformCache(AIJsonSchemaTransformOptions transformOptions)
     /// </summary>
     /// <param name="function">The function whose JSON schema we want to transform.</param>
     /// <returns>The transformed JSON schema corresponding to <see cref="TransformOptions"/>.</returns>
-    public JsonElement GetOrCreateTransformedSchema(AIFunction function)
+    [EditorBrowsable(EditorBrowsableState.Never)] // functionality for AIFunction is satisfied by AIFunctionDefinition overload
+    public JsonElement GetOrCreateTransformedSchema(AIFunction function) =>
+        GetOrCreateTransformedSchema((AIFunctionDefinition)function);
```
SteveSandersonMS: Is this just for binary back-compat?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
stephentoub: Yes, I think so.
```suggestion
    /// Gets the name of a specific tool that must be called.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
stephentoub: Primarily to enable skipping any function invoking middleware in the pipeline (I need to make a few tweaks to FunctionInvokingChatClient for this to actually work). I can put a non-AIFunction AIFunctionDefinition into ChatOptions.Tools, and then FunctionCallContent for it won't be automatically handled by the FICC but will instead pass through.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/DelegatingAIFunctionDefinition.cs
stephentoub: Yes, I had it public for completeness, but making it internal, at least for now, makes sense. It's much more applicable at the AIFunction level where it's more common to want to augment invocation behavior.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonSchemaTransformCache.cs
stephentoub: Yup.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
```diff
@@ -6,44 +6,17 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+#pragma warning disable SA1202 // Elements should be ordered by access
+
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents a function that can be described to an AI service and invoked.</summary>
-public abstract class AIFunction : AITool
+public abstract class AIFunction : AIFunctionDefinition
```
shyamnamboodiripad: I wish we could have implemented this using composition instead of inheritance. It feels a bit odd to say that every `AIFunction` _is an_ `AIFunctionDefinition` as opposed to saying that every `AIFunction` _has an_ `AIFunctionDefinition`. Having a dedicated (and concrete) `AIFunctionDefinition` (or `AIFunctionDescriptor`) class could also make it easier for consumers to instantiate definitions without needing to inherit and create their own sub classes.

However, the composition based approach would involve introducing a `AIFunction.Definition` (or `AIFunction.Descriptor`) property that returns an `AIFunctionDefiniton` with the same `Name`, `Description`, `JsonSchema` and `ReturnJsonSchema` properties as the `AIFunction` itself. This would also force API changes in the eval layer which currently requires consumers to describe the tools being used in terms of `AITool`s. We could have avoided both problems if this change were made pre-GA - but alas that ship has sailed :)

Anyways, just wanted to leave this note in case this sparks any further thoughts / ideas.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -398,7 +398,7 @@ public static AIFunction Create(MethodInfo method, object? target, string? name
     /// <para>
     /// By default, any parameters to <paramref name="method"/> are sourced from the <see cref="AIFunctionArguments"/>'s dictionary
     /// of key/value pairs and are represented in the JSON schema for the function, as exposed in the returned <see cref="AIFunction"/>'s
-    /// <see cref="AIFunction.JsonSchema"/>. There are a few exceptions to this:
+    /// <see cref="AIFunctionDefinition.JsonSchema"/>. There are a few exceptions to this:
     /// <list type="bullet">
     ///   <item>
     ///     <description>
```
shyamnamboodiripad, line 404: Can we provide a helper that can return an `AIFunctionDefintion` for supplied `name`, `description`, `jsonSchema` and `returnJsonSchema`? Alternatively, can we create a public `NonInvokableAIFunction` type with the same set of properties?

This could make things easier for consumers who  want to represent tool definitions sans implementation (as was the case with the eval-related scenarios / server-side tool definitions we discussed offline) without needing to inherit and define their own sub classes...

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionDefinition.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System.Text.Json;
+using System.Threading.Tasks;
+
+#pragma warning disable S1694 // An abstract class should have both abstract and concrete methods
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a function that can be described to an AI service.</summary>
+public abstract class AIFunctionDefinition : AITool
```
shyamnamboodiripad: Just thinking aloud - but would it be possible to make `AIFunctionDefinition` a concrete type (with the constructor parameters to instantiate the four virtual properties)?

Would someone want to inherit this abstract class for any reason other than providing custom values for these four properties (i.e., name, description and the two schemas)?

shyamnamboodiripad (commented):
Thanks much for adding! The only other thing that may be nice to include is a simple helper / API to construct definitions for non-invokable tools (such as server-side tools).

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunction.cs
stephentoub: I think the inheritance is desirable; maybe we need a better name for the base class, but really this is just the thing that provides the metadata, and then the derived implementation can also provide the ability to invoke. From the perspective of a left IChatClient implementation, it doesn't care about being able to invoke, it just needs to translate the function's metadata into something it can put on the wire, and with inheritance it can simply test whether it's the base class, using it if it is... the fact that it might also supply the ability to invoke doesn't matter, but it's not forced to check for two things and a difference it doesn't need to know about.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
stephentoub: I added AIFunctionFactory.CreateDefinition.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -513,6 +542,31 @@ private static void FixupHistories(
...
+    /// The lists of tools to combine into a single dictionary. Tools from later lists are preferred
+    /// over tools from earlier lists if they have the same name.
+    /// </param>
+    private static Dictionary<string, AITool>? CreateToolsDictionary(params ReadOnlySpan<IList<AITool>?> toolLists)
+    {
+        Dictionary<string, AITool>? tools = null;
+
+        foreach (var toolList in toolLists)
+        {
+            if (toolList?.Count is int count && count > 0)
```
jozkee, line 557:
```suggestion
            if (toolList?.Count > 0)
```

jozkee: I thought this was an optimization, didn't see below usage.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -571,19 +625,68 @@ private static void UpdateOptionsForNextIteration(ref ChatOptions? options, stri
...
+                }
+            }
+            else
+            {
+                // The tool couldn't be found. If we're configured to terminate on unknown call requests,
+                // break out of the loop now. Otherwise, ProcessFunctionCallsAsync will handle it by
+                // creating a NotFound response message.
+                if (TerminateOnUnknownCalls)
+                {
+                    return true;
```
jozkee, line 668: I think is good that termination occurs even when there's FCCs with known functions. Did you consider the alternative of signaling termination but invoking known functions and add the respective FRCs before returning to the caller?

jozkee: To be more specific, I'm talking about the case where the AI service returns FCCs with known and unknown functions.

stephentoub: > Did you consider the alternative of signaling termination but invoking known functions and add the respective FRCs before returning to the caller?

Adding the FRCs to what?

stephentoub:
@SteveSandersonMS, would you mind taking another look? In particular at the FICC logic changes. Thanks.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
jozkee: To the response, I meant to process the FCCs before breaking out.  This is a hacky sketch of what I am trying to convey: https://github.com/dotnet/extensions/commit/e905e9cd5a2b7b43649a0b1594370050f0e52828.

stephentoub: > To the response

If the conversation Id is set, the consumer won't send the response back under the assumption the server already has it.

jozkee (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionDefinition.cs
eiriktsarpalis: Nit: I'm not sure "definition" accurately conveys what this abstraction is representing. It took me a bit to understand its purpose since in my mind a "function definition" is something that defines an implementation. How about using terms like "signature" or "header"?

stephentoub: 1. AIFunctionDefinition
2. AIFunctionSignature
3. AIFunctionDeclaration
4. AIFunctionDescription
5. ... something else...

@eiriktsarpalis, @jeffhandley, @SteveSandersonMS, @shyamnamboodiripad... preferences?

SteveSandersonMS: I like "declaration" most, with "definition" as a 2nd preference.

eiriktsarpalis: I'd go for either "signature" or "declaration"

stephentoub: Sounds like Declaration is on everyone's ok-with list. I'll go with that.

shyamnamboodiripad: +1 "declaration" sounds good to me. "Definition" is also ok imho... "Description" would also have been fine if this were a composed type - but I feel that would be a bit confusing for a base type...

PR #6721, opened 2025-08-15
Fix GetResponseAsync<T> to look at the last message only

stephentoub:
It was looking at all messages in the whole response rather than just the last, which is where we expect the JSON to be.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6721)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
@@ -127,7 +127,7 @@ public class ChatResponse<T> : ChatResponse
             return _deserializedResult;
         }
 
-        var json = Text;
+        var json = Messages.Count > 0 ? Messages[Messages.Count - 1].Text : string.Empty;
         if (string.IsNullOrEmpty(json))
         {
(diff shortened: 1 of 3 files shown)


shyamnamboodiripad (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatResponse{T}.cs
```diff
@@ -127,7 +127,7 @@ public bool TryGetResult([NotNullWhen(true)] out T? result)
             return _deserializedResult;
         }
 
-        var json = Text;
+        var json = Messages.Count > 0 ? Messages[Messages.Count - 1].Text : string.Empty;
```
shyamnamboodiripad, line 130: Does this seem important enough to call out in the changelog for vNext?

shyamnamboodiripad:
@dotnet/dotnet-extensions-ai Could I please get one more approval on this fix from someone on the MEAI side? The bug was reported by an internal partner team, and it would be great to have the fix merged sooner rather than later. Thanks!

PR #6745, opened 2025-08-26
Add function approvals support

westey-m:
- Adding an ApprovalRequiredAIFunction that can be used to decorate an AIFunction
- Adding UserInputRequestContent and UserInputResponseContent base types to represent all user input
- Adding FunctionApprovalRequestContent and FunctionApprovalResponseContent for function approvals
- Updated FunctionInvokingChatClient with approval generation hand handling.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionApprovalRequestContent.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionApprovalRequestContent.cs
@@ -1,0 +1,38 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a request for user approval of a function call.
+/// </summary>
+[Experimental("MEAI001")]
+public sealed class FunctionApprovalRequestContent : UserInputRequestContent
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="FunctionApprovalRequestContent"/> class.
+    /// </summary>
+    /// <param name="id">The ID to uniquely identify the function approval request/response pair.</param>
+    /// <param name="functionCall">The function call that requires user approval.</param>
+    public FunctionApprovalRequestContent(string id, FunctionCallContent functionCall)
+        : base(id)
+    {
+        FunctionCall = Throw.IfNull(functionCall);
+    }
+
+    /// <summary>
+    /// Gets the function call that pre-invoke approval is required for.
+    /// </summary>
+    public FunctionCallContent FunctionCall { get; }
+
+    /// <summary>
+    /// Creates a <see cref="FunctionApprovalResponseContent"/> to indicate whether the function call is approved or rejected based on the value of <paramref name="approved"/>.
+    /// </summary>
+    /// <param name="approved"><see langword="true"/> if the function call is approved; otherwise, <see langword="false"/>.</param>
+    /// <returns>The <see cref="FunctionApprovalResponseContent"/> representing the approval response.</returns>
+    public FunctionApprovalResponseContent CreateResponse(bool approved)
+        => new(Id, approved, FunctionCall);
+}
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
@@ -247,6 +247,26 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
         bool lastIterationHadConversationId = false; // whether the last iteration's response had a ConversationId set
         int consecutiveErrorCount = 0;
 
+        // Process approval requests (remove from originalMessages) and rejected approval responses (re-create FCC and create failed FRC).
+        var (preDownstreamCallHistory, notInvokedApprovals) = ProcessFunctionApprovalResponses(
+            originalMessages, !string.IsNullOrWhiteSpace(options?.ConversationId), toolResponseId: null, functionCallContentFallbackMessageId: null);
+
+        // Invoke approved approval responses, which generates some additional FRC wrapped in ChatMessage.
+        (IList<ChatMessage>? invokedApprovedFunctionApprovalResponses, bool shouldTerminate, consecutiveErrorCount) =
+            await InvokeApprovedFunctionApprovalResponsesAsync(notInvokedApprovals, originalMessages, options, consecutiveErrorCount, isStreaming: false, cancellationToken);
+
+        if (invokedApprovedFunctionApprovalResponses is not null)
+        {
+            // We need to add the generated FRC to the list we'll return to callers as part of the next response.
+            preDownstreamCallHistory ??= [];
+            preDownstreamCallHistory.AddRange(invokedApprovedFunctionApprovalResponses);
+        }
+
+        if (shouldTerminate)
+        {
+            return new ChatResponse(preDownstreamCallHistory);
+        }
+
         for (int iteration = 0; ; iteration++)
         {
             functionCallContents?.Clear();
@@ -258,6 +278,10 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
                 Throw.InvalidOperationException($"The inner {nameof(IChatClient)} returned a null {nameof(ChatResponse)}.");
             }
 
+            // Before we do any function execution, make sure that any functions that require approval, have been turned into approval requests
+            // so that they don't get executed here.
+            response.Messages = await ReplaceFunctionCallsWithApprovalRequestsAsync(response.Messages, options?.Tools, AdditionalTools, cancellationToken);
+
             // Any function call work to do? If yes, ensure we're tracking that work in functionCallContents.
             bool requiresFunctionInvocation =
                 (options?.Tools is { Count: > 0 } || AdditionalTools is { Count: > 0 }) &&
@@ -268,6 +292,11 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
             // fast path out by just returning the original response.
             if (iteration == 0 && !requiresFunctionInvocation)
             {
+                // Insert any pre-invocation FCC and FRC that were converted from approval responses into the response here,
+                // so they are returned to the caller.
+                response.Messages = UpdateResponseMessagesWithPreDownstreamCallHistory(response.Messages, preDownstreamCallHistory);
+                preDownstreamCallHistory = null;
+
                 return response;
             }
 
@@ -342,10 +371,59 @@ public partial class FunctionInvokingChatClient : DelegatingChatClient
         List<ChatResponseUpdate> updates = []; // updates from the current response
         int consecutiveErrorCount = 0;
 
+        // This is a synthetic ID since we're generating the tool messages instead of getting them from
+        // the underlying provider. When emitting the streamed chunks, it's perfectly valid for us to
+        // use the same message ID for all of them within a given iteration, as this is a single logical
+        // message with multiple content items. We could also use different message IDs per tool content,
+        // but there's no benefit to doing so.
+        string toolResponseId = Guid.NewGuid().ToString("N");
+
+        // We also need a synthetic ID for the function call content for approved function calls
+        // where we don't know what the original message id of the function call was.
+        string functionCallContentFallbackMessageId = Guid.NewGuid().ToString("N");
+
+        ApprovalRequiredAIFunction[]? approvalRequiredFunctions = (options?.Tools ?? []).Concat(AdditionalTools ?? []).OfType<ApprovalRequiredAIFunction>().ToArray();
+        bool hasApprovalRequiringFunctions = approvalRequiredFunctions.Length > 0;
+
+        // Process approval requests (remove from original messages) and rejected approval responses (re-create FCC and create failed FRC).
+        var (preDownstreamCallHistory, notInvokedApprovals) = ProcessFunctionApprovalResponses(
+            originalMessages, !string.IsNullOrWhiteSpace(options?.ConversationId), toolResponseId, functionCallContentFallbackMessageId);
+        if (preDownstreamCallHistory is not null)
+        {
+            foreach (var message in preDownstreamCallHistory)
+            {
+                yield return ConvertToolResultMessageToUpdate(message, options?.ConversationId, message.MessageId);
+                Activity.Current = activity; // workaround for https://github.com/dotnet/runtime/issues/47802
+            }
+        }
+
+        // Invoke approved approval responses, which generates some additional FRC wrapped in ChatMessage.
+        (IList<ChatMessage>? invokedApprovedFunctionApprovalResponses, bool shouldTerminate, consecutiveErrorCount) =
+            await InvokeApprovedFunctionApprovalResponsesAsync(notInvokedApprovals, originalMessages, options, consecutiveErrorCount, isStreaming: true, cancellationToken);
+
+        if (invokedApprovedFunctionApprovalResponses is not null)
+        {
+            foreach (var message in invokedApprovedFunctionApprovalResponses)
+            {
+                message.MessageId = toolResponseId;
+                yield return ConvertToolResultMessageToUpdate(message, options?.ConversationId, message.MessageId);
+                Activity.Current = activity; // workaround for https://github.com/dotnet/runtime/issues/47802
+            }
+
+            if (shouldTerminate)
+            {
+                yield break;
+            }
+        }
+
         for (int iteration = 0; ; iteration++)
         {
             updates.Clear();
             functionCallContents?.Clear();
+
+            bool hasApprovalRequiringFcc = false;
+            int lastApprovalCheckedFCCIndex = 0;
+            int lastYieldedUpdateIndex = 0;
 
             await foreach (var update in base.GetStreamingResponseAsync(messages, options, cancellationToken))
             {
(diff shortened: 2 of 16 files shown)


westey-m:

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,470 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+    }
+
+    /// <summary>
+    /// Extracts the <see cref="FunctionCallContent"/> from the provided <see cref="FunctionApprovalResponseContent"/> to recreate the original function call messages.
+    /// The output messages tries to mimic the original messages that contained the <see cref="FunctionCallContent"/>, e.g. if the <see cref="FunctionCallContent"/>
+    /// had been split into separate messages, this method will recreate similarly split messages, each with their own <see cref="FunctionCallContent"/>.
+    /// </summary>
+#pragma warning disable CA1859 // Use concrete types when possible for improved performance
+    private static ICollection<ChatMessage>? ConvertToFunctionCallContentMessages(IEnumerable<ApprovalResultWithRequestMessage>? resultWithRequestMessages, string? fallbackMessageId)
+#pragma warning restore CA1859 // Use concrete types when possible for improved performance
```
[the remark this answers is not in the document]
westey-m: Changed code to remove warning

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,470 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+    {
+        if (resultWithRequestMessages is not null)
+        {
+            ChatMessage? currentMessage = null;
+            Dictionary<string, ChatMessage>? messagesById = null;
+
+            foreach (var resultWithRequestMessage in resultWithRequestMessages)
+            {
+                // Don't need to create a dictionary on the first iteration or if we alrady have one.
```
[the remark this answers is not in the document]
westey-m: Fixed.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,470 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
     private static TimeSpan GetElapsedTime(long startingTimestamp) =>
 #if NET
         Stopwatch.GetElapsedTime(startingTimestamp);
 #else
         new((long)((Stopwatch.GetTimestamp() - startingTimestamp) * ((double)TimeSpan.TicksPerSecond / Stopwatch.Frequency)));
 #endif
 
+    /// <summary>
+    /// Execute the provided <see cref="FunctionApprovalResponseContent"/> and return the resulting <see cref="FunctionCallContent"/>.
+    /// </summary>
```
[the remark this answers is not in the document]
westey-m: Updated comment.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+                        .ToArray();
+
+                    anyApprovalRequired |= approvalRequiredFunctions.FirstOrDefault(x => x.Name == functionCall.Name) is { } approvalFunction &&
+                        await approvalFunction.RequiresApprovalCallback(new(functionCall), cancellationToken);
+                }
+            }
+        }
+
+        // If any function calls were found, and any of them required approval, we should replace all of them with approval requests.
+        // This is because we do not have a way to deal with cases where some function calls require approval and others do not, so we just replace all of them.
```
jozkee, line 1549: What's blocking us from handling mixed cases?

jozkee: I believe is because FICC should not return FCCs that can be invoked, so we have to protect them with the approval request.  Have you considered any alternatives?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+        if (allFunctionCallContentIndices is not null && anyApprovalRequired)
+        {
+            // Clone the list so, we don't mutate the input.
+            outputMessages = [.. messages];
```
jozkee, line 1555:
```suggestion
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+            int lastMessageIndex = -1;
+
+            foreach (var (messageIndex, contentIndex) in allFunctionCallContentIndices)
+            {
+                // Clone the message if we didn't already clone it in a previous iteration.
+                var message = lastMessageIndex != messageIndex ? outputMessages[messageIndex].Clone() : outputMessages[messageIndex];
+                message.Contents = [.. message.Contents];
```
jozkee, line 1562:
```suggestion
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+        if (allFunctionCallContentIndices is not null && anyApprovalRequired)
+        {
+            // Clone the list so, we don't mutate the input.
+            outputMessages = [.. messages];
+            int lastMessageIndex = -1;
+
+            foreach (var (messageIndex, contentIndex) in allFunctionCallContentIndices)
+            {
+                // Clone the message if we didn't already clone it in a previous iteration.
+                var message = lastMessageIndex != messageIndex ? outputMessages[messageIndex].Clone() : outputMessages[messageIndex];
```
jozkee, line 1561: What's the point of these defensive copies?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -268,6 +292,11 @@ public override async Task<ChatResponse> GetResponseAsync(
             // fast path out by just returning the original response.
             if (iteration == 0 && !requiresFunctionInvocation)
             {
+                // Insert any pre-invocation FCC and FRC that were converted from approval responses into the response here,
+                // so they are returned to the caller.
+                response.Messages = UpdateResponseMessagesWithPreDownstreamCallHistory(response.Messages, preDownstreamCallHistory);
+                preDownstreamCallHistory = null;
+
                 return response;
```
jozkee, line 364: Is the code below reachable with approvals?  If so, do we have a test for that?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -342,11 +371,60 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
...
+        // message with multiple content items. We could also use different message IDs per tool content,
+        // but there's no benefit to doing so.
+        string toolResponseId = Guid.NewGuid().ToString("N");
+
+        // We also need a synthetic ID for the function call content for approved function calls
+        // where we don't know what the original message id of the function call was.
+        string functionCallContentFallbackMessageId = Guid.NewGuid().ToString("N");
+
+        ApprovalRequiredAIFunction[]? approvalRequiredFunctions = (options?.Tools ?? []).Concat(AdditionalTools ?? []).OfType<ApprovalRequiredAIFunction>().ToArray();
+        bool hasApprovalRequiringFunctions = approvalRequiredFunctions.Length > 0;
```
jozkee: This can be avoided until after the next 3 yields.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+        return new ChatMessage(ChatRole.Assistant, [resultWithRequestMessage.Response.FunctionCall]) { MessageId = fallbackMessageId };
+    }
+
+    /// <summary>
+    /// Check if any of the provided <paramref name="functionCallContents"/> require approval.
+    /// Supports checking from a provided index up to the end of the list, to allow efficient incremental checking
+    /// when streaming.
+    /// </summary>
+    private static async Task<(bool hasApprovalRequiringFcc, int lastApprovalCheckedFCCIndex)> CheckForApprovalRequiringFCCAsync(
+        List<FunctionCallContent>? functionCallContents,
```
jozkee, line 1455:
```suggestion
        List<FunctionCallContent> functionCallContents,
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+        ApprovalRequiredAIFunction[] approvalRequiredFunctions,
+        bool hasApprovalRequiringFcc,
+        int lastApprovalCheckedFCCIndex,
+        CancellationToken cancellationToken)
+    {
+        // If we already found an approval requiring FCC, we can skip checking the rest.
+        if (hasApprovalRequiringFcc)
+        {
+            return (true, functionCallContents?.Count ?? 0);
```
jozkee:
```suggestion
            return (true, functionCallContents.Count);
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+        }
+
+        for (; lastApprovalCheckedFCCIndex < (functionCallContents?.Count ?? 0); lastApprovalCheckedFCCIndex++)
```
jozkee:
```suggestion
        for (; lastApprovalCheckedFCCIndex < functionCallContents.Count; lastApprovalCheckedFCCIndex++)
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+    /// when streaming.
+    /// </summary>
+    private static async Task<(bool hasApprovalRequiringFcc, int lastApprovalCheckedFCCIndex)> CheckForApprovalRequiringFCCAsync(
+        List<FunctionCallContent>? functionCallContents,
+        ApprovalRequiredAIFunction[] approvalRequiredFunctions,
+        bool hasApprovalRequiringFcc,
+        int lastApprovalCheckedFCCIndex,
+        CancellationToken cancellationToken)
+    {
+        // If we already found an approval requiring FCC, we can skip checking the rest.
```
jozkee, line 1460:
```suggestion
        Debug.Assert(functionCallContents is not null);
        // If we already found an approval requiring FCC, we can skip checking the rest.
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -920,13 +1021,476 @@ FunctionResultContent CreateFunctionResultContent(FunctionInvocationResult resul
...
+            return (true, functionCallContents?.Count ?? 0);
+        }
+
+        for (; lastApprovalCheckedFCCIndex < (functionCallContents?.Count ?? 0); lastApprovalCheckedFCCIndex++)
+        {
+            var fcc = functionCallContents![lastApprovalCheckedFCCIndex];
+            if (approvalRequiredFunctions.FirstOrDefault(y => y.Name == fcc.Name) is ApprovalRequiredAIFunction approvalFunction &&
+                await approvalFunction.RequiresApprovalCallback(new(fcc), cancellationToken))
+            {
+                hasApprovalRequiringFcc |= true;
```
jozkee:
```suggestion
                hasApprovalRequiringFcc = true;
                break;
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
```diff
@@ -371,12 +449,55 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
...
+                    Activity.Current = activity; // workaround for https://github.com/dotnet/runtime/issues/47802
+                }
+                else
+                {
+                    // Check if any of the function call contents in this update requires approval.
+                    // Once we find the first one that requires approval, this method becomes a no-op.
+                    (hasApprovalRequiringFcc, lastApprovalCheckedFCCIndex) = await CheckForApprovalRequiringFCCAsync(
+                        functionCallContents, approvalRequiredFunctions, hasApprovalRequiringFcc, lastApprovalCheckedFCCIndex, cancellationToken);
+
+                    // We've encountered a function call content that requires approval (either in this update or ealier)
```
jozkee: Consider turning on the spell checker in VS, there's multiple typos in this file.
```suggestion
                    // We've encountered a function call content that requires approval (either in this update or earlier)
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionApprovalRequestContent.cs
```diff
@@ -0,0 +1,38 @@
...
+[Experimental("MEAI001")]
+public sealed class FunctionApprovalRequestContent : UserInputRequestContent
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="FunctionApprovalRequestContent"/> class.
+    /// </summary>
+    /// <param name="id">The ID to uniquely identify the function approval request/response pair.</param>
+    /// <param name="functionCall">The function call that requires user approval.</param>
+    public FunctionApprovalRequestContent(string id, FunctionCallContent functionCall)
+        : base(id)
```
jozkee, line 25: I think 99% of calls will be passing the FCC.CallId, it may be worth changing the ctor.
```suggestion
    public FunctionApprovalRequestContent(FunctionCallContent functionCall)
        : base(functionCall.CallId)
```

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: Correct, callers should not be returned un-invoked FCC, but more importantly, when in service thread mode, callers will not provide the uninvoked FCC back to the ChatClient after handling approvals, so the FCC that don't require approval will never get invoked.
FICC also cannot invoke the FCC that doesn't require approval (while keeping the FCC that does in reserve) and call the LLM with only the ones that don't require approval, since the services typically requires responses for all FCC and fail if not provided.
An alternative we considered was to wrap all approvals in one aggregated approval request, which also contains the FCC that doesn't require approval.  This introduces a lot of complexity though.
Another option to work around this limitation is to handle these cases automatically in the ChatClientAgent, since it can store the FCC that doesn't require approval in its thread temporarily.  This is something we can still do in addition to this work.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: If someone has a caching chat client lower down in the stack, we'd end up potentially modifying the cache.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: CopyFunctionCalls only creates the functionCallContents List if there are actually function calls, (to avoid an unnecessary allocation), so this argument can be null.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: This `if` is for cases where we have no functions, below is for cases where we have functions, both if they require approval and if not.  The code below is covered by both the existing unit tests and the new ones.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: functionCallContents may be null here if we have not encountered any yet.  It is not initialized until one is found to avoid unnecessary allocations.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: Thanks, moved down.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: Thanks, added Debug.Assert and updated.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: Thanks, added Debug.Assert

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: Makes sense, updated.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs
westey-m: Thanks, fixed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/FunctionApprovalRequestContent.cs
eavanvalkenburg: [OpenAI MCP integration](https://platform.openai.com/docs/guides/tools-connectors-mcp) use a separate ID for the approval request and for the executed call: 
Approval Request:
```json
{
    "id": "mcpr_68a619e1d82c8190b50c1ccba7ad18ef0d2d23a86136d339",
    "type": "mcp_approval_request",
    "arguments": "{\"diceRollExpression\":\"2d4 + 1\"}",
    "name": "roll",
    "server_label": "dmcp"
}
```
MCP call:
```json
{
    "id": "mcp_68a6102d8948819c9b1490d36d5ffa4a0679e572a900e618",
    "type": "mcp_call",
    "approval_request_id": null,
    "arguments": "{\"diceRollExpression\":\"2d4 + 1\"}",
    "error": null,
    "name": "roll",
    "output": "4",
    "server_label": "dmcp"
}
```
note the two id fields: `approval_request_id` and `id`

Potentially other implementations won't do this, but there is precedence that approval requests have other id's then the function calls they relate to.

stephentoub:
I'm working on this now to get this merged today...

Issue #6737, opened 2025-08-21
[API Proposal]: Make some Telemetry helper types public


evgenyfedorov2:
### Background and motivation

LAST UPDATED: 26.08.2025

This will allow consumers to provide and register in DI custom implementations of types specified below, replacing the default ones. For example, it can be used for working with telemetry (traces, logs) in incoming and outgoing HTTP requests and redacting PII information.

I propose to make following `internal` types `public`:

1. [DownstreamDependencyMetadataManager](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Http/DownstreamDependencyMetadataManager.cs) - it will allow consumers to use that type, for example, via constructor injection. There is already a public extension [method](https://github.com/dotnet/extensions/blob/77859cb87c33bcda7187871a683c5779557cb2ef/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Http/HttpDiagnosticsServiceCollectionExtensions.cs#L23) to register it in DI, but it is useless as of today - you can call the extension method, but you cannot inject anything to constructor because the class is `internal`.

1. [HttpRouteFormatter](https://github.com/dotnet/extensions/blob/77859cb87c33bcda7187871a683c5779557cb2ef/src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteFormatter.cs) - currently `internal`, but it would be nice to have it `public`. + we can add a `public` method for registering it in DI, similar to [this](https://github.com/dotnet/extensions/blob/77859cb87c33bcda7187871a683c5779557cb2ef/src/Libraries/Microsoft.Extensions.Telemetry/Http/TelemetryCommonExtensions.cs#L21).
 
1. [HttpRouteParser](https://github.com/dotnet/extensions/blob/77859cb87c33bcda7187871a683c5779557cb2ef/src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteParser.cs) - currently `internal`, but it would be nice to have it `public`. + we can add a `public` method for registering it in DI, similar to [this](https://github.com/dotnet/extensions/blob/77859cb87c33bcda7187871a683c5779557cb2ef/src/Libraries/Microsoft.Extensions.Telemetry/Http/TelemetryCommonExtensions.cs#L21).

[...]

PR #6747, opened 2025-08-26
Expose some Microsoft.Extensions.Http.Diagnostics API

evgenyfedorov2:
this is to back up the proposal https://github.com/dotnet/extensions/issues/6737
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6747)


dariusclay (approved):
LGTM

src/Libraries/Microsoft.Extensions.Telemetry/Http/HttpRouteParser.cs
```diff
@@ -38,61 +57,70 @@ public bool TryExtractParameters(
...
-    public ParsedRouteSegments ParseRoute(string httpRoute)
+    /// <summary>
+    /// Parses http route and breaks it into text and parameter segments.
+    /// </summary>
+    /// <param name="httpRoute">Http request's route template.</param>
+    /// <returns>Returns text and parameter segments of route.</returns>
+    public virtual ParsedRouteSegments ParseRoute(string httpRoute)
     {
-        return _routeTemplateSegmentsCache.GetOrAdd(httpRoute, httpRoute =>
+        return _routeTemplateSegmentsCache.GetOrAdd(httpRoute, _ =>
```
iliar-turdushev, line 123:
```suggestion
        return _routeTemplateSegmentsCache.GetOrAdd(httpRoute, httpRoute =>
```

If we pass `httpRoute` as a parameter rather than getting it from the closure we'll save some bytes, won't we?

iliar-turdushev (approved):

PR #6748, opened 2025-08-26
Add Kubernetes based Resource Monitoring

### LINKED ISSUE #6496: [API Proposal]: Kubernetes specialized Resource Monitoring for Windows
### Background and motivation

Current implementation of `Microsoft.Extensions.Diagnostics.ResourceMonitoring` for Windows container environment does not support `requests.cpu` and `requests.memory` in calculations. For `limits` it reads JobObject now, which is suitable for non-Kubernetes environments.

In Kubernetes, both requests and limits can be exposed via environment variables if cluster is configured to use Downward API. If the cluster is configured accordingly and assuming that engineer knows that and knows the names of the environment variables, ResourceMonitoring could be extended to use these values utilization calculations.
Example yaml with Downward API that injects values into pods' environment variables
```yaml
 env:
    - name: MY_CLUSTER_CPU_REQUEST
      valueFrom:
        resourceFieldRef:
          resource: requests.cpu
    - name: MY_CLUSTER_CPU_LIMIT
      valueFrom:
        resourceFieldRef:
          resource: limits.cpu
    - name: MY_CLUSTER_MEMORY_REQUEST
      valueFrom:
        resourceFieldRef:
          resource: requests.memory
    - name: MY_CLUSTER_MEMORY_LIMIT
      valueFrom:
        resourceFieldRef:
          resource: limits.memory
```
Linux is just fine, so that applies only for Windows.

### API Proposal

Configuration class extension of K8s config.
```csharp
public class ResourceMonitoringOptions
{
    ... other configs
    /// <summary>
    /// Configuration for reading CPU and memory limits and requests from Kubernetes environment variables.
    /// This is supported only on Windows containers.
    /// </summary>
    public KubernetesResourceConfig? KubernetesResourceConfig { get; set; }
}
```
K8s config details
```csharp
public sealed class KubernetesResourceConfig
{
    /// <summary>
    /// Enables reading CPU/memory utilization context from Kubernetes via environment variables.
    /// </summary>
    public bool Enabled { get; set; } = false;

    /// <summary>
    /// The environment variable name for CPU limit (e.g., "MY_CLUSTER_CPU_LIMIT").
    /// </summary>
    public string CpuLimitEnvVar { get; set; } = "MY_CLUSTER_CPU_LIMIT";

    /// <summary>
    /// The environment variable name for memory limit (e.g., "MY_CLUSTER_MEMORY_LIMIT").
    /// </summary>
    public string MemoryLimitEnvVar { get; set; } = "MY_CLUSTER_MEMORY_LIMIT";

    /// <summary>
    /// The environment variable name for CPU request (e.g., "MY_CLUSTER_CPU_REQUEST").
    /// </summary>
    public string CpuRequestEnvVar { get; set; } = "MY_CLUSTER_CPU_REQUEST";

    /// <summary>
    /// The environment variable name for memory request (e.g., "MY_CLUSTER_MEMORY_REQUEST").
   /// </summary>
   public string MemoryRequestEnvVar { get; set; } = "MY_CLUSTER_MEMORY_REQUEST";
}
```

### API Usage

How to configure Resource Monitoring with K8s config
```csharp
services.AddResourceMonitoring(options =>
{
    options.KubernetesResourceConfig = new KubernetesResourceConfig
    {
        Enabled = true,
        CpuLimitEnvVar = "MY_CLUSTER_CPU_LIMIT",
        MemoryLimitEnvVar = "MY_CLUSTER_MEMORY_LIMIT",
        CpuRequestEnvVar = "MY_CLUSTER_CPU_REQUEST",
        MemoryRequestEnvVar = "MY_CLUSTER_MEMORY_REQUEST"
    };
});
```

### Alternative Designs

_No response_

### Risks

If improperly configured it should fallback to existing behavior or throw on startup, although it seems extensively too conservative.
Env variables parsing needs to be done well to handle all possible options how K8s can be configured
It's Windows only => have a validation that throws on non-Windows

amadeuszl:
Fix #6496


src/Libraries/Microsoft.Extensions.ClusterMetadata.Kubernetes/KubernetesClusterMetadataExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        return builder.Add(new KubernetesClusterMetadataSource(sectionName, environmentVariablePrefix));
+    }
+
+    /// <summary>
+    /// Adds an instance of <see cref="KubernetesClusterMetadata"/> to the <see cref="IServiceCollection"/>.
+    /// </summary>
+    /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+    /// <param name="section">The configuration section to bind.</param>
+    /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+    public static IServiceCollection AddKubernetesClusterMetadata(this IServiceCollection services, IConfigurationSection section)
```
evgenyfedorov2: - `IServiceCollection` extension methods should be in the `KubernetesClusterMetadataServiceCollectionExtensions` class and in the `Microsoft.Extensions.DependencyInjection` namespace
- `IConfigurationBuilder` extension methods should be in the `KubernetesClusterMetadataConfigurationBuilderExtensions` class and in the `Microsoft.Extensions.Configuration` namespace

Use this component as an example - https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.AmbientMetadata.Application

src/Libraries/Microsoft.Extensions.ClusterMetadata.Kubernetes/Microsoft.Extensions.ClusterMetadata.Kubernetes.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>Microsoft.Extensions.ClusterMetadata.Kubernetes</RootNamespace>
+    <Description></Description>
+    <TargetFrameworks>$(NetCoreTargetFrameworks)$(ConditionalNet462)</TargetFrameworks>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
evgenyfedorov2: these two are controlled via https://github.com/dotnet/extensions/blob/da94b6045fef85f78a5eee183579221cee684e51/Directory.Build.props#L40-L41

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
```diff
@@ -119,6 +144,14 @@ public LinuxUtilizationProvider(IOptions<ResourceMonitoringOptions> options, ILi
             observeValues: () => GetMeasurementWithRetry(MemoryPercentage),
             unit: "1");
 
+        if (options.Value.UseKubernetesLimitsAndRequests)
```
evgenyfedorov2: while this specific change (new metric) is a good benefit of having K8s cluster metadata, it does not affect public API nor is it related to the K8s cluster metadata component itself, so could we introduce it in a separate PR later on, please?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
```diff
@@ -120,34 +133,20 @@ internal WindowsContainerSnapshotProvider(
...
+            _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ContainerCpuRequestUtilization, observeValue: () => CpuPercentageForRequests()); // todo implement
+            _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ContainerMemoryRequestUtilization, observeValue: () => MemoryPercentaForRequests()); // todo implement
+        }
 
-        _ = meter.CreateObservableGauge(
-            name: ResourceUtilizationInstruments.ProcessMemoryUtilization,
-            observeValue: ProcessMemoryPercentage);
+        // Process based metrics:
+        _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ProcessCpuUtilization, observeValue: CpuPercentage);
+        _ = meter.CreateObservableGauge(name: ResourceUtilizationInstruments.ProcessMemoryUtilization, observeValue: () => MemoryPercentage(() => _processInfo.GetCurrentProcessMemoryUsage()));
```
evgenyfedorov2: Doesn't having one method (`MemoryPercentage()`) for two metrics introduce again the old bug you spotted some time ago? `ProcessMemoryUtilization` and `ContainerMemoryLimitUtilization` metrics, if measured shortly enough one after another, will both get the same cached  `_memoryPercentage` value

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
amadeuszl: I'm adding it right away as it's a feature in 1P.

src/Libraries/Microsoft.Extensions.ClusterMetadata.Kubernetes/KubernetesClusterMetadataExtensions.cs
amadeuszl: Good point, fixed

src/Shared/Instruments/ResourceUtilizationInstruments.cs
```diff
@@ -58,6 +58,14 @@ internal static class ResourceUtilizationInstruments
...
 
+    /// <summary>
+    /// The name of an instrument to retrieve memory request consumption of all processes running inside a container or control group in range <c>[0, 1]</c>.
+    /// </summary>
+    /// <remarks>
+    /// The type of an instrument is <see cref="System.Diagnostics.Metrics.ObservableGauge{T}"/>.
+    /// </remarks>
+    public const string ContainerMemoryRequestUtilization = "container.memory.request.utilization";
+
     /// <summary>
```
evgenyfedorov2: can we consider introducing this metric in a separate PR? We may or may not introduce it later, but anyway it is a separate concern than the K8s layer.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/IResourceQuotasProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
+
+public interface IResourceQuotasProvider
+{
+    public ResourceQuotas GetResourceQuotas();
+}
+
```
evgenyfedorov2: Probably it should be singular `Quota` word everywhere.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceQuotas.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Gets or sets the resource memory limit the container is allowed to use.
+    /// </summary>
+    public ulong LimitsMemory { get; set; }
+
+    /// <summary>
+    /// Gets or sets the resource CPU limit the container is allowed to use.
+    /// </summary>
+    public ulong LimitsCpu { get; set; }
```
evgenyfedorov2: The public comment should mention the units, which is milli cores for CPU and bytes for memory, I guess?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/IResourceQuotasProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
+
+public interface IResourceQuotasProvider
```
evgenyfedorov2: This probably should be in a separate component `Microsoft.Extensions.Diagnostics.ResourceMonitoring.Abstractions` ?

- `Microsoft.Extensions.Diagnostics.ResourceMonitoring` will depend on the abstractions and will have internal implementations such as Windows and Linux Resource Quota Providers.
- `Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes` will depend on the abstractions and will have the internal implementation for Kubernetes Downward API.
- As you mentioned in comments, users may register the K8s implementation before calling the `AddResourceMonitoring()` method.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotasServiceCollectionsExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    /// <param name="services"></param>
+    /// <param name="configureOptions"></param>
+    /// <returns></returns>
+    public static IServiceCollection AddKubernetesResourceQuotas(
+        this IServiceCollection services,
+        Action<KubernetesMetadata>? configureOptions = null)
+    {
+        // Simple - just register before AddResourceMonitoring() is called
+        services.TryAddSingleton<IResourceQuotasProvider, KubernetesResourceQuotasProvider>();
```
evgenyfedorov2: The only remaining thing is to add support of `IConfigurationSource` too, similar to this https://learn.microsoft.com/en-us/dotnet/core/extensions/custom-configuration-provider#custom-configuration-provider

src/Libraries/Microsoft.Extensions.ClusterMetadata.Kubernetes/Microsoft.Extensions.ClusterMetadata.Kubernetes.csproj
amadeuszl: removed that code

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/LinuxUtilizationProvider.cs
amadeuszl: I think it's outdates after new changes

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Windows/WindowsContainerSnapshotProvider.cs
amadeuszl: Removed new metric, leaving it for separate PR

src/Shared/Instruments/ResourceUtilizationInstruments.cs
amadeuszl: Yes, removed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/IResourceQuotasProvider.cs
amadeuszl: Changed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceQuotas.cs
amadeuszl: Good point, added.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/IResourceQuotasProvider.cs
amadeuszl: Good idea, done.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotasServiceCollectionsExtensions.cs
amadeuszl: As agreed, we stay with plain object to not expose those values in `IConfiguration`

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/IResourceQuotasProvider.cs
amadeuszl: After API Review it was decided to not have new abstractions library just for 1 or 2 classes

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// </remarks>
+    public static IServiceCollection AddKubernetesResourceMonitoring(
+        this IServiceCollection services,
+        string? environmentVariablePrefix = default)
+    {
+        services.TryAddSingleton<KubernetesMetadata>(serviceProvider =>
+        {
+            var metadata = new KubernetesMetadata(environmentVariablePrefix ?? string.Empty);
+            return metadata.Build();
+        });
```
evgenyfedorov2: I think this should also support IConfiguration pattern?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <strong>Important:</strong> Do not call <see cref="ResourceMonitoringServiceCollectionExtensions.AddResourceMonitoring(IServiceCollection)"/> 
+    /// if you are using this method, as it already includes all necessary resource monitoring components and registers a Kubernetes-specific 
+    /// <see cref="ResourceQuotaProvider"/> implementation. Calling both methods may result in conflicting service registrations.
+    /// </para>
+    /// </remarks>
+    public static IServiceCollection AddKubernetesResourceMonitoring(
+        this IServiceCollection services,
+        string? environmentVariablePrefix = default)
+    {
+        services.TryAddSingleton<KubernetesMetadata>(serviceProvider =>
```
evgenyfedorov2: should this register `IOptions<KubernetesMetadata>` instead?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceQuota.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Extensions.Diagnostics.ResourceMonitoring;
+
+/// <summary>
+/// Represents resource quota information for a process or container, including CPU and memory constraints,
+/// that are used as denominators in CPU and memory utilization calculations.
+/// Maximum values define the upper limits of resource usage, while baseline values specify 
+/// the minimum assured resource allocations.
```
evgenyfedorov2: how does it translate into K8s request and limits?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            quota.BaselineCpuInCores = quota.MaxCpuInCores;
+        }
+
+        if (quota.BaselineMemoryInBytes == 0)
+        {
+            quota.BaselineMemoryInBytes = quota.MaxMemoryInBytes;
+        }
+
+        return quota;
```
evgenyfedorov2, line 39: Normally we would register `IOptions<KubernetesMetadata>` and validate those options with[ Validation API](https://learn.microsoft.com/en-us/dotnet/core/extensions/options-validation-generator), so that whatever gets injected via constructor is either `null` or valid (nothing in between). Is there a rationale to avoid that here?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
amadeuszl: We agreed to not expose it, as it would make those properties "public"

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceQuota.cs
amadeuszl: max are limits, and baselines are requests

amadeuszl: mapped in https://github.com/dotnet/extensions/pull/6748/files/296f89661ffd0ee92c057988f42e4d278676cad1#diff-09ef6b5054308faa24ecf97ff8280f22f013f30efa83039cda4b7df7a71f19c0R17

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
evgenyfedorov2: oh, I did not realize `KubernetesMetadata` is `internal`.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
amadeuszl: We don't use `IConfiguration` for it so it would bring a bit of complexity without much in return, especially if we don't use DI here too. 

But the usage here is a bit different. It's not validation logic, it's domain logic. It makes sure those values are not 0s and are changed to Max values, as they will be used as denominators during metering. And on the other hand, value `0` for `kubernetesMetadata.BaselineCpuInCores` or `kubernetesMetadata.BaselineMemoryInBytes` is correct value so failing validation for that would not be desired state.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceQuota.cs
evgenyfedorov2: do you think it makes sense to reflect that in XML documentation here?

test/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Tests/Linux/LinuxUtilizationProviderTests.cs
```diff
@@ -198,8 +200,8 @@ public Task Provider_EmitsLogRecord()
         });
 
         var parser = new LinuxUtilizationParserCgroupV2(fileSystem: fileSystem, new FakeUserHz(100));
-
-        var snapshotProvider = new LinuxUtilizationProvider(options, parser, meterFactoryMock.Object, logger, TimeProvider.System);
+        var resourceQuotaProvider = new LinuxResourceQuotaProvider(parser, options);
+        var snapshotProvider = new LinuxUtilizationProvider(options, parser, meterFactoryMock.Object, resourceQuotaProvider,logger, TimeProvider.System);
```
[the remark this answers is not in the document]
amadeuszl: Fixed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/Linux/ILinuxUtilizationParser.cs
```diff
@@ -95,4 +95,10 @@ internal interface ILinuxUtilizationParser
     /// </summary>
     /// <returns>The number of CPU periods.</returns>
     long GetCgroupPeriodsIntervalInMicroSecondsV2();
+
+    /// <summary>
+    /// For CgroupV2 only. Reads the file /sys/fs/cgroup/memory.min, if 0 reads the file /sys/fs/cgroup/memory.low.
+    /// </summary>
+    /// <returns>memory.</returns>
```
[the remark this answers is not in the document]
amadeuszl: Fixed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesMetadata.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </summary>
+    public ulong LimitsCpu { get; set; }
+
+    /// <summary>
+    /// Gets or sets the resource memory request the container is allowed to use in bytes.
+    /// </summary>
+    public ulong RequestsMemory { get; set; }
+
+    /// <summary>
+    /// Gets or sets the resource CPU request the container is allowed to use in milicores.
```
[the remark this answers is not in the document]
amadeuszl: Fixed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesMetadata.cs
```diff
@@ -0,0 +1,67 @@
...
+
+internal class KubernetesMetadata
+{
+    /// <summary>
+    /// Gets or sets the resource memory limit the container is allowed to use in bytes.
+    /// </summary>
+    public ulong LimitsMemory { get; set; }
+
+    /// <summary>
+    /// Gets or sets the resource CPU limit the container is allowed to use in milicores.
```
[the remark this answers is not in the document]
amadeuszl: Fixed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
+/// Lets you configure and register Kubernetes resource monitoring components.
+/// </summary>
+public static class KubernetesResourceQuotaServiceCollectionExtensions
+{
+    /// <summary>
+    /// Configures and adds an Kubernetes resource monitoring components to a service collection alltoghter with necessary basic resource monitoring components.
```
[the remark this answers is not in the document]
amadeuszl: Fixed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
evgenyfedorov2: ok, I see, thanks!

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// </para>
+    /// </remarks>
+    public static IServiceCollection AddKubernetesResourceMonitoring(
+        this IServiceCollection services,
+        string? environmentVariablePrefix = default)
+    {
+        services.TryAddSingleton<KubernetesMetadata>(serviceProvider =>
+        {
+            var metadata = new KubernetesMetadata(environmentVariablePrefix ?? string.Empty);
+            return metadata.Build();
```
evgenyfedorov2: is there a reason why we need the `Build()` method? If we create an object and then immediately call some initialization method on it, it is better to perform the initialization within the constructor

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+
+internal class KubernetesResourceQuotaProvider : ResourceQuotaProvider
+{
+    private const double MillicoresPerCore = 1000.0;
+    private KubernetesMetadata _kubernetesMetadata;
+
+    public KubernetesResourceQuotaProvider(KubernetesMetadata kubernetesMetadata)
+    {
+        _ = Throw.IfNull(kubernetesMetadata);
+        _kubernetesMetadata = kubernetesMetadata;
```
evgenyfedorov2, line 16: Given that `KubernetesMetadata` is `internal`, does `KubernetesResourceQuotaProvider` class really need to depend on it? I think we can safely get rid of it and move all its logic to the `KubernetesResourceQuotaProvider`

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <NoWarn Condition="'$(TargetFramework)' == 'net462'">$(NoWarn);CS0436</NoWarn>
+    <EnablePackageValidation>false</EnablePackageValidation>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <InjectSharedInstruments>true</InjectSharedInstruments>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Stage>normal</Stage>
```
evgenyfedorov2: I think it should be `dev` stage initially. After one release (= one month) we can change it to `normal`. And in this case we don't have to bother with EnablePackageValidation because it is not enabled for dev packages: https://github.com/dotnet/extensions/blob/7ddf87e16c92efd0d6d4b7da4a9ee7d185e8835f/eng/MSBuild/Packaging.targets#L40

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <MinCodeCoverage>99</MinCodeCoverage>
+    <MinMutationScore>90</MinMutationScore>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+    <Compile Remove="Linux\**\*.cs" />
+  </ItemGroup>
```
evgenyfedorov2: There is no Linux folder, so we can remove this

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes.csproj
amadeuszl: Let me verify that. I would prefer to release it already in December as normal.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
amadeuszl: I see, you are right. Although, my logic here was if we add new package for Kubernetes metadata in a future, then we can simply replace that. I don't have strong opinion on that, wdyt?

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
amadeuszl: Yep, here's proposal:
```csharp
    public static IServiceCollection AddKubernetesResourceMonitoring(
        string? environmentVariablePrefix = default)
    {
        services.TryAddSingleton<ResourceQuotaProvider>(sp =>
        {
            return new KubernetesResourceQuotaProvider(KubernetesMetadata.FromEnvironmentVariables(environmentVariablePrefix ?? string.Empty));
        });

        _ = services.AddResourceMonitoring();

        return services;
    }
```
And then `KubernetesMetadata` is:
```csharp
    public static KubernetesMetadata FromEnvironmentVariables(string environmentVariablePrefix)
    {
        return new KubernetesMetadata
        {
            LimitsMemory = GetEnvironmentVariableAsUInt64($"{environmentVariablePrefix}LIMITS_MEMORY"),
            LimitsCpu = GetEnvironmentVariableAsUInt64($"{environmentVariablePrefix}LIMITS_CPU"),
            RequestsMemory = GetEnvironmentVariableAsUInt64($"{environmentVariablePrefix}REQUESTS_MEMORY"),
            RequestsCpu = GetEnvironmentVariableAsUInt64($"{environmentVariablePrefix}REQUESTS_CPU"),
        };
    }
```

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
amadeuszl: I proposed alternative with static method, mainly I'm leaving `KubernetesMetadata` for readability

evgenyfedorov2: We might or might not add it, but as of today I'd prefer to keep code simpler and avoid having more classes than necessary as of today.

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaServiceCollectionExtensions.cs
evgenyfedorov2: I am fine the proposal too

amadeuszl: Changed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/KubernetesResourceQuotaProvider.cs
amadeuszl: Went with static method

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes.csproj
amadeuszl: removed

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes/Microsoft.Extensions.Diagnostics.ResourceMonitoring.Kubernetes.csproj
amadeuszl: Switched to dev

src/Libraries/Microsoft.Extensions.Diagnostics.ResourceMonitoring/ResourceQuota.cs
amadeuszl: Noted in the comment

evgenyfedorov2 (approved):

PR #6749, opened 2025-08-26
Image generation tool

ericstj:
This adds a prototype implementation of ImageGenerationTool with support in OpenAI provider.

To use this requires using the OpenAI Responses chat client with OpenAI provider (not Azure OpenAI).

Still required:
- [ ] Add tests for tool (currently testing with a sample)
- [ ] Polish and document API
- [ ] Refine implementation in OpenAI to ensure its minimally fragile (despite usage of reflection) - test with AOT
- [ ] Refine content update coalescing methods after testing in sample

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
@@ -13,6 +13,7 @@ using Microsoft.Shared.Diagnostics;
 
 #pragma warning disable S109 // Magic numbers should not be used
 #pragma warning disable S1121 // Assignments should not be made from within sub-expressions
+#pragma warning disable EXTAI0001 // Suppress experimental warnings for internal usage
 
 namespace Microsoft.Extensions.AI;
 
@@ -66,6 +67,32 @@ public static class ChatResponseExtensions
         }
 
         list.AddMessages(updates.ToChatResponse());
+    }
+
+    /// <summary>Converts the <paramref name="updates"/> into <see cref="ChatMessage"/> instances and adds them to <paramref name="list"/>.</summary>
+    /// <param name="list">The destination list to which the newly constructed messages should be added.</param>
+    /// <param name="updates">The <see cref="ChatResponseUpdate"/> instances to convert to messages and add to the list.</param>
+    /// <param name="options">Options for configuring how the updates are coalesced.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="list"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// As part of combining <paramref name="updates"/> into a series of <see cref="ChatMessage"/> instances, the
+    /// method may use <see cref="ChatResponseUpdate.MessageId"/> to determine message boundaries, as well as coalesce
+    /// contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static void AddMessages(this IList<ChatMessage> list, IEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options)
+    {
+        _ = Throw.IfNull(list);
+        _ = Throw.IfNull(updates);
+
+        if (updates is ICollection<ChatResponseUpdate> { Count: 0 })
+        {
+            return;
+        }
+
+        list.AddMessages(updates.ToChatResponse(options));
     }
 
     /// <summary>Converts the <paramref name="update"/> into a <see cref="ChatMessage"/> instance and adds it to <paramref name="list"/>.</summary>
@@ -122,6 +149,128 @@ public static class ChatResponseExtensions
             list.AddMessages(await updates.ToChatResponseAsync(cancellationToken).ConfigureAwait(false));
     }
 
+    /// <summary>Converts the <paramref name="updates"/> into <see cref="ChatMessage"/> instances and adds them to <paramref name="list"/>.</summary>
+    /// <param name="list">The list to which the newly constructed messages should be added.</param>
+    /// <param name="updates">The <see cref="ChatResponseUpdate"/> instances to convert to messages and add to the list.</param>
+    /// <param name="options">Options for configuring how the updates are coalesced.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="Task"/> representing the completion of the operation.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="list"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// As part of combining <paramref name="updates"/> into a series of <see cref="ChatMessage"/> instances, tne
+    /// method may use <see cref="ChatResponseUpdate.MessageId"/> to determine message boundaries, as well as coalesce
+    /// contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static Task AddMessagesAsync(
+        this IList<ChatMessage> list, IAsyncEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(list);
+        _ = Throw.IfNull(updates);
+
+        return AddMessagesAsync(list, updates, options, cancellationToken);
+
+        static async Task AddMessagesAsync(
+            IList<ChatMessage> list, IAsyncEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options, CancellationToken cancellationToken) =>
+            list.AddMessages(await updates.ToChatResponseAsync(options, cancellationToken).ConfigureAwait(false));
+    }
+
+    /// <summary>Applies a <see cref="ChatResponseUpdate"/> to an existing <see cref="ChatResponse"/>.</summary>
+    /// <param name="response">The response to which the update should be applied.</param>
+    /// <param name="update">The update to apply to the response.</param>
+    /// <param name="options">Options for configuring how the update is applied.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="response"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="update"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// This method modifies the existing <paramref name="response"/> by incorporating the content and metadata
+    /// from the <paramref name="update"/>. This includes using <see cref="ChatResponseUpdate.MessageId"/> to determine
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static void ApplyUpdate(this ChatResponse response, ChatResponseUpdate update, ChatResponseUpdateCoalescingOptions? options = null)
+    {
+        _ = Throw.IfNull(response);
+        _ = Throw.IfNull(update);
+
+        ProcessUpdate(update, response, options);
+        FinalizeResponse(response);
+    }
+
+    /// <summary>Applies <see cref="ChatResponseUpdate"/> instances to an existing <see cref="ChatResponse"/>.</summary>
+    /// <param name="response">The response to which the updates should be applied.</param>
+    /// <param name="updates">The updates to apply to the response.</param>
+    /// <param name="options">Options for configuring how the updates are applied.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="response"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// This method modifies the existing <paramref name="response"/> by incorporating the content and metadata
+    /// from the <paramref name="updates"/>. This includes using <see cref="ChatResponseUpdate.MessageId"/> to determine
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static void ApplyUpdates(this ChatResponse response, IEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options = null)
+    {
+        _ = Throw.IfNull(response);
+        _ = Throw.IfNull(updates);
+
+        if (updates is ICollection<ChatResponseUpdate> { Count: 0 })
+        {
+            return;
+        }
+
+        foreach (var update in updates)
+        {
+            ProcessUpdate(update, response, options);
+        }
+
+        FinalizeResponse(response);
+    }
+
+    /// <summary>Applies <see cref="ChatResponseUpdate"/> instances to an existing <see cref="ChatResponse"/> asynchronously.</summary>
+    /// <param name="response">The response to which the updates should be applied.</param>
+    /// <param name="updates">The updates to apply to the response.</param>
+    /// <param name="options">Options for configuring how the updates are applied.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="Task"/> representing the completion of the operation.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="response"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// This method modifies the existing <paramref name="response"/> by incorporating the content and metadata
+    /// from the <paramref name="updates"/>. This includes using <see cref="ChatResponseUpdate.MessageId"/> to determine
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static Task ApplyUpdatesAsync(
+        this ChatResponse response,
+        IAsyncEnumerable<ChatResponseUpdate> updates,
+        ChatResponseUpdateCoalescingOptions? options = null,
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(response);
+        _ = Throw.IfNull(updates);
+
+        return ApplyUpdatesAsync(response, updates, options, cancellationToken);
+
+        static async Task ApplyUpdatesAsync(
+            ChatResponse response,
+            IAsyncEnumerable<ChatResponseUpdate> updates,
+            ChatResponseUpdateCoalescingOptions? options,
+            CancellationToken cancellationToken)
+        {
+            await foreach (var update in updates.WithCancellation(cancellationToken).ConfigureAwait(false))
+            {
+                ProcessUpdate(update, response, options);
+            }
+
+            FinalizeResponse(response);
+        }
+    }
+
     /// <summary>Combines <see cref="ChatResponseUpdate"/> instances into a single <see cref="ChatResponse"/>.</summary>
     /// <param name="updates">The updates to be combined.</param>
     /// <returns>The combined <see cref="ChatResponse"/>.</returns>
@@ -142,6 +291,35 @@ public static class ChatResponseExtensions
         foreach (var update in updates)
         {
             ProcessUpdate(update, response);
+        }
+
+        FinalizeResponse(response);
+
+        return response;
+    }
+
+    /// <summary>Combines <see cref="ChatResponseUpdate"/> instances into a single <see cref="ChatResponse"/>.</summary>
+    /// <param name="updates">The updates to be combined.</param>
+    /// <param name="options">Options for configuring how the updates are coalesced.</param>
+    /// <returns>The combined <see cref="ChatResponse"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// As part of combining <paramref name="updates"/> into a single <see cref="ChatResponse"/>, the method will attempt to reconstruct
+    /// <see cref="ChatMessage"/> instances. This includes using <see cref="ChatResponseUpdate.MessageId"/> to determine
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static ChatResponse ToChatResponse(
+        this IEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options)
+    {
+        _ = Throw.IfNull(updates);
+
+        ChatResponse response = new();
+
+        foreach (var update in updates)
+        {
+            ProcessUpdate(update, response, options);
         }
 
         FinalizeResponse(response);
@@ -175,6 +353,42 @@ public static class ChatResponseExtensions
             await foreach (var update in updates.WithCancellation(cancellationToken).ConfigureAwait(false))
             {
                 ProcessUpdate(update, response);
+            }
+
+            FinalizeResponse(response);
+
+            return response;
+        }
+    }
+
+    /// <summary>Combines <see cref="ChatResponseUpdate"/> instances into a single <see cref="ChatResponse"/>.</summary>
+    /// <param name="updates">The updates to be combined.</param>
+    /// <param name="options">Options for configuring how the updates are coalesced.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The combined <see cref="ChatResponse"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// As part of combining <paramref name="updates"/> into a single <see cref="ChatResponse"/>, the method will attempt to reconstruct
+    /// <see cref="ChatMessage"/> instances. This includes using <see cref="ChatResponseUpdate.MessageId"/> to determine
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static Task<ChatResponse> ToChatResponseAsync(
+        this IAsyncEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(updates);
+
+        return ToChatResponseAsync(updates, options, cancellationToken);
+
+        static async Task<ChatResponse> ToChatResponseAsync(
+            IAsyncEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options, CancellationToken cancellationToken)
+        {
+            ChatResponse response = new();
+
+            await foreach (var update in updates.WithCancellation(cancellationToken).ConfigureAwait(false))
+            {
+                ProcessUpdate(update, response, options);
             }
 
             FinalizeResponse(response);
@@ -302,6 +516,15 @@ public static class ChatResponseExtensions
     /// <param name="response">The <see cref="ChatResponse"/> object that should be updated based on <paramref name="update"/>.</param>
     private static void ProcessUpdate(ChatResponseUpdate update, ChatResponse response)
     {
+        ProcessUpdate(update, response, null);
+    }
+
+    /// <summary>Processes the <see cref="ChatResponseUpdate"/>, incorporating its contents into <paramref name="response"/>.</summary>
+    /// <param name="update">The update to process.</param>
+    /// <param name="response">The <see cref="ChatResponse"/> object that should be updated based on <paramref name="update"/>.</param>
+    /// <param name="options">Options for configuring how the updates are coalesced.</param>
+    private static void ProcessUpdate(ChatResponseUpdate update, ChatResponse response, ChatResponseUpdateCoalescingOptions? options)
+    {
         // If there is no message created yet, or if the last update we saw had a different
         // message ID than the newest update, create a new message.
         ChatMessage message;
@@ -361,6 +584,34 @@ public static class ChatResponseExtensions
                 // Usage content is treated specially and propagated to the response's Usage.
                 case UsageContent usage:
                     (response.Usage ??= new()).Add(usage.Details);
+                    break;
+
+                case DataContent dataContent when options is not null &&
+                    options.ReplaceDataContentWithSameName &&
+                    !string.IsNullOrEmpty(dataContent.Name):
+                    // Check if there's an existing DataContent with the same name to replace
+                    int existingIndex = -1;
+                    for (int i = 0; i < message.Contents.Count; i++)
+                    {
+                        if (message.Contents[i] is DataContent existingDataContent &&
+                            string.Equals(existingDataContent.Name, dataContent.Name, StringComparison.Ordinal))
+                        {
+                            existingIndex = i;
+                            break;
+                        }
+                    }
+
+                    if (existingIndex >= 0)
+                    {
+                        // Replace the existing DataContent
+                        message.Contents[existingIndex] = dataContent;
+                    }
+                    else
+                    {
+                        // No existing DataContent with the same name, add it normally
+                        message.Contents.Add(dataContent);
+                    }
+
                     break;
 
                 default:
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Extensions.AI;
 /// </para>
 /// <para>
 /// The relationship between <see cref="ChatResponse"/> and <see cref="ChatResponseUpdate"/> is
-/// codified in the <see cref="ChatResponseExtensions.ToChatResponseAsync"/> and
+/// codified in the <see cref="ChatResponseExtensions.ToChatResponseAsync(IAsyncEnumerable{ChatResponseUpdate}, System.Threading.CancellationToken)"/> and
 /// <see cref="ChatResponse.ToChatResponseUpdates"/>, which enable bidirectional conversions
 /// between the two. Note, however, that the provided conversions may be lossy, for example if multiple
 /// updates all have different <see cref="RawRepresentation"/> objects whereas there's only one slot for
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
@@ -18,12 +18,12 @@ public class ImageGenerationTool : AITool
     /// Initializes a new instance of the <see cref="ImageGenerationTool"/> class with the specified options.
     /// </summary>
     /// <param name="options">The options to configure the image generation request. If <paramref name="options"/> is <see langword="null"/>, default options will be used.</param>
-    public ImageGenerationTool(TextToImageOptions? options = null)
+    public ImageGenerationTool(ImageGenerationOptions? options = null)
         : base()
     {
         AdditionalProperties = new AdditionalPropertiesDictionary(new Dictionary<string, object?>
         {
-            [nameof(TextToImageOptions)] = options
+            [nameof(ImageGenerationOptions)] = options
         });
     }
 
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
@@ -1,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a hosted tool that can be specified to an AI service to enable it to perform image generation.</summary>
+/// <remarks>
+/// This tool does not itself implement image generation. It is a marker that can be used to inform a service
+/// that the service is allowed to perform image generation if the service is capable of doing so.
+/// </remarks>
+[Experimental("MEAI001")]
+public class ImageGenerationTool : AITool
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ImageGenerationTool"/> class with the specified options.
+    /// </summary>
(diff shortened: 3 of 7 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a hosted tool that can be specified to an AI service to enable it to perform image generation.</summary>
+/// <remarks>
+/// This tool does not itself implement image generation. It is a marker that can be used to inform a service
+/// that the service is allowed to perform image generation if the service is capable of doing so.
+/// </remarks>
+[Experimental("MEAI001")]
+public class ImageGenerationTool : AITool
```
stephentoub: I'd expect this to be "HostedImageGenerationTool", to conform to the naming convention we've established where "Hosted" is used as a prefix when the tool instance itself doesn't actually provide the behaviors and instead is just a marker to signal to the service that some hosted tool should be used.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ImageGenerationTool"/> class with the specified options.
+    /// </summary>
+    /// <param name="options">The options to configure the image generation request. If <paramref name="options"/> is <see langword="null"/>, default options will be used.</param>
+    public ImageGenerationTool(ImageGenerationOptions? options = null)
+        : base()
+    {
+        AdditionalProperties = new AdditionalPropertiesDictionary(new Dictionary<string, object?>
```
stephentoub: Why put this in AdditionalProperties rather than just having an Options property on the tool?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -68,6 +69,32 @@ public static void AddMessages(this IList<ChatMessage> list, IEnumerable<ChatRes
...
+    /// <exception cref="ArgumentNullException"><paramref name="list"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// As part of combining <paramref name="updates"/> into a series of <see cref="ChatMessage"/> instances, the
+    /// method may use <see cref="ChatResponseUpdate.MessageId"/> to determine message boundaries, as well as coalesce
+    /// contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
+    public static void AddMessages(this IList<ChatMessage> list, IEnumerable<ChatResponseUpdate> updates, ChatResponseUpdateCoalescingOptions? options)
```
stephentoub: Why are these options needed? There are other policies encoded into our current coalescing logic that we don't expose for configuration; _if_ we had a type named ChatResponseUpdateCoalescingOptions, I'd expect those knobs to also show up here, but I'm more interested in understanding why we can't just choose a policy and only later add public knobs if folks really need them.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -68,6 +69,32 @@ public static void AddMessages(this IList<ChatMessage> list, IEnumerable<ChatRes
...
+    /// <param name="options">Options for configuring how the updates are coalesced.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="list"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="updates"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// As part of combining <paramref name="updates"/> into a series of <see cref="ChatMessage"/> instances, the
+    /// method may use <see cref="ChatResponseUpdate.MessageId"/> to determine message boundaries, as well as coalesce
+    /// contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("EXTAI0001")]
```
stephentoub: We currently use MEAI001

src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a hosted tool that can be specified to an AI service to enable it to perform image generation.</summary>
```
PederHP: There's other hosted tools, like OpenAI's Code Interpreter. Other providers have something similar. Anthropic for example has Web Search, Fetch and Code Interpreter as "Server Tools". Maybe out of scope for this change, but a generalized abstraction for these would be great. AdditionalProperties seems to be common for all of them - including the Anthropic ones. So I think this would fit beautifully as a more general abstraction than `ImageGenerationTool`. `ServerTool` or `HostedTool`?

stephentoub: We already have HostedWebSearchTool, HostedCodeInterpreterTool, HostedFileSearchTool, and HostedMcpServerTool. Having a HostedImageGenerationTool makes sense to me. Different providers have different ways of exposing the same fundamental information, so being able to write e.g. HostedWebSearchTool, and have that map to the right thing for Gemini and Anthropic and OpenAI makes sense to me. AdditionalProperties can be used in each when there's some setting that's not exposed in a strongly-typed fashion on the tool type.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: I came up with a new heuristic in how to coalesce DataContent.  That would break someone who might have expected otherwise.  I was adding this API to demonstrate how we might deal with that.  I'm fine leaving it out until folks ask for it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationTool.cs
ericstj: I thought I was following an existing convention here, but now I cannot find that.  I'll go ahead and make a property for it.

ericstj:
Forgot to mention that https://github.com/ericstj/imageGeneratorSample/tree/imageGeneratorTool is the sample that uses this.

ericstj:
I'm moving this to draft.  I have private changes will pick up a self-built copy of OpenAI that has these types public (change is in main of OpenAI, but not yet on NuGet), which we won't be able to merge.  I'm also still not happy with the middle-ware and need to experiment more with that.

ericstj:
I'll also add that I know this middle-ware is not thread safe at the moment and it needs to be.  I'm currently experimenting how to manage the state.  Once I settle on that I'll make it thread-safe since we document that all IChatClient's need to be.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -122,6 +122,95 @@ static async Task AddMessagesAsync(
...
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("MEAI0001")]
+    public static void ApplyUpdates(this ChatResponse response, IEnumerable<ChatResponseUpdate> updates)
+    {
+        _ = Throw.IfNull(response);
+        _ = Throw.IfNull(updates);
+
+        if (updates is ICollection<ChatResponseUpdate> { Count: 0 })
```
stephentoub: Is the existing ToChatResponse then just:
```csharp
ChatResponse response = new();
response.ApplyUpdates(updates);
return response;
```
or is there a meaningful behavioral or performance difference with that?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -375,6 +464,29 @@ private static void ProcessUpdate(ChatResponseUpdate update, ChatResponse respon
                     (response.Usage ??= new()).Add(usage.Details);
                     break;
 
+                case DataContent dataContent when
+                    !string.IsNullOrEmpty(dataContent.Name):
+                    // Check if there's an existing DataContent with the same name to replace
+                    for (int i = 0; i < message.Contents.Count; i++)
```
stephentoub: Does this change the algorithm to be O(N^2)?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ImageGeneratingChatClient.cs
```diff
@@ -0,0 +1,469 @@
...
+
+    private readonly IImageGenerator _imageGenerator;
+    private readonly AITool[] _aiTools;
+    private readonly HashSet<string> _functionNames;
+    private readonly DataContentHandling _dataContentHandling;
+
+    // the following fields all have scope per-request. They are cleared at the start of each request.
+    private readonly Dictionary<string, List<AIContent>> _imageContentByCallId = [];
+    private readonly Dictionary<string, AIContent> _imageContentById = new(StringComparer.OrdinalIgnoreCase);
+    private ImageGenerationOptions? _imageGenerationOptions;
```
stephentoub: I would expect instead you'd create an object in Get{Streaming}ResponseAsync to hold the state for that request, and whatever tools you want to send would reference that object (e.g. they'd be instance methods on it).

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ImageGeneratingChatClient.cs
```diff
@@ -0,0 +1,469 @@
...
+
+        List<ChatMessage>? newMessages = null;
+        int messageIndex = 0;
+        foreach (var message in messages)
+        {
+            List<AIContent>? newContents = null;
+            for (int contentIndex = 0; contentIndex < message.Contents.Count; contentIndex++)
+            {
+                var content = message.Contents[contentIndex];
+                if (content is DataContent dataContent && dataContent.MediaType.StartsWith("image/", StringComparison.OrdinalIgnoreCase))
```
stephentoub: DataContent.HasTopLevelMediaType is intended for this

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: Yeah, I can implement it that way.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: Yeah, let  me build a dictionary to make n*log(n).

ericstj: Actually since we call this for each update, and we might have many updates, each update could hit this, however only when adding a named image.  Still, if every update had a single named image, we'd still be at n^2.

In practice we'll only really see small numbers of named images.  For OpenAI streaming (the case I added this for) we have between 0 and 3 [partial_images](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools-image-generation-tool-partial_images).

Do we accept untrusted data from these?  If we need to fix this I think I would need to move the behavior into `FinalizeResponse`

ericstj:
Test failures are for templates due to my use of the private OpenAI build.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -219,6 +339,8 @@ internal static void CoalesceContent(IList<AIContent> contents)
                 return content;
             });
 
+        CoalesceImageDataContent(contents);
```
ericstj: I have replacement semantics for images, which may be non-contiguous.  For example, when streaming images, you have 2 iterations of 2 images, they will be interleaved.  The subsequent images are more detailed versions of prior ones with the same name.

I might be able to get this to work using Coalesce<DataContent> if I record the names I see along the way in the canMerge callback (and just ignore previous parameter).  Let me know if you would prefer that.

test/Libraries/Microsoft.Extensions.AI.OllamaSharp.Integration.Tests/OllamaSharpImageGeneratingChatClientIntegrationTests.cs
```diff
@@ -0,0 +1,19 @@
...
+
+/// <summary>
+/// OllamaSharp-specific integration tests for ImageGeneratingChatClient.
+/// Tests the ImageGeneratingChatClient with OllamaSharp chat client implementation.
+/// </summary>
+public class OllamaSharpImageGeneratingChatClientIntegrationTests : ImageGeneratingChatClientIntegrationTests
+{
+    protected override IChatClient? CreateChatClient() =>
+        IntegrationTestHelpers.GetOllamaUri() is Uri endpoint ?
+            new OllamaApiClient(endpoint, "llama3.2") :
```
stephentoub, line 17: llama3.2 supports image generation?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedImageGenerationTool.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// and how it represents this tool, a new instance of that implementation-specific tool type can be returned by this
+    /// callback for the <see cref="IChatClient" /> implementation to use instead of creating a new instance.
+    /// Such implementations might mutate the supplied options instance further based on other settings supplied on this
+    /// <see cref="HostedImageGenerationTool" /> instance or from other inputs, therefore, it is <b>strongly recommended</b> to not
+    /// return shared instances and instead make the callback return a new instance on each call.
+    /// This is typically used to set an implementation-specific setting that isn't otherwise exposed from the strongly typed
+    /// properties on <see cref="ImageGenerationOptions" />.
+    /// </remarks>
+    [JsonIgnore]
+    public Func<IChatClient?, object?>? RawRepresentationFactory { get; set; }
```
stephentoub: I'm not understanding why this needs to exist. The purpose of such a HostedXxTool is to be provider agnostic. If a developer can create the raw representation, then you can instead do the exact same thing with ChatOptions.RawRepresentationFactory, or in the case of Responses, use responseTool.AsAITool (and other providers can provide similar escape hatches). Let's delete this RawRepresentationFactory on the tool itself.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedImageGenerationTool.cs
```diff
@@ -0,0 +1,49 @@
...
+/// that the service is allowed to perform image generation if the service is capable of doing so.
+/// </remarks>
+[Experimental("MEAI001")]
+public class HostedImageGenerationTool : AITool
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedImageGenerationTool"/> class with the specified options.
+    /// </summary>
+    public HostedImageGenerationTool()
+        : base()
```
stephentoub:
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
```diff
@@ -88,6 +88,7 @@ private static JsonSerializerOptions CreateDefaultOptions()
 
     // Primitives
     [JsonSerializable(typeof(string))]
+    [JsonSerializable(typeof(string[]))]
```
stephentoub: What required this?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIJsonContext.cs
```diff
@@ -14,8 +14,10 @@ namespace Microsoft.Extensions.AI;
     WriteIndented = true)]
 [JsonSerializable(typeof(OpenAIClientExtensions.ToolJson))]
 [JsonSerializable(typeof(IDictionary<string, object?>))]
+[JsonSerializable(typeof(IDictionary<string, string?>))]
 [JsonSerializable(typeof(IReadOnlyDictionary<string, object?>))]
 [JsonSerializable(typeof(string[]))]
 [JsonSerializable(typeof(IEnumerable<string>))]
 [JsonSerializable(typeof(JsonElement))]
+[JsonSerializable(typeof(int))]
```
stephentoub: What requires these additional types?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -226,6 +230,34 @@ internal static IEnumerable<ChatMessage> ToChatMessages(IEnumerable<ResponseItem
...
+        {
+            RawRepresentation = update,
+            AdditionalProperties = new()
+            {
+                [nameof(update.ItemId)] = update.ItemId,
+                [nameof(update.OutputIndex)] = update.OutputIndex,
+                [nameof(update.PartialImageIndex)] = update.PartialImageIndex
+            }
+        };
+    }
```
stephentoub: We've begun adopting a pattern where the server tools have call and result types, e.g.
McpServerToolCallContent and McpServerToolResultContent
CodeInterpreterToolCallContent and CodeInterpreterToolResultContent

We'll want to do the same thing for HostedWebServerTool, so WebServerToolCallContent and WebServerToolResultContent.

We should do the same for image generation as well. That would mean adding an ImageGenerationToolCallContent and ImageGenerationToolResultContent, and these generated images would be available in a property hanging off of the ImageGenerationToolResultContent.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -226,6 +230,34 @@ internal static IEnumerable<ChatMessage> ToChatMessages(IEnumerable<ResponseItem
...
+        var outputType = imageGenTool?.OutputFileFormat?.ToString() ?? "png";
+
+        return new DataContent(update.PartialImageBytes, $"image/{outputType}")
+        {
+            RawRepresentation = update,
+            AdditionalProperties = new()
+            {
+                [nameof(update.ItemId)] = update.ItemId,
+                [nameof(update.OutputIndex)] = update.OutputIndex,
+                [nameof(update.PartialImageIndex)] = update.PartialImageIndex
```
stephentoub: Would this be helped by https://github.com/dotnet/extensions/issues/6965 ?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -546,6 +585,45 @@ internal static FunctionTool ToResponseTool(AIFunctionDeclaration aiFunction, Ch
...
+        // Model: Image generation model
+        if (imageGenerationOptions?.ModelId is not null && result.Model is null)
+        {
+            result.Model = imageGenerationOptions.ModelId;
+        }
+
+        // Size: Image dimensions (e.g., 1024x1024, 1024x1536)
+        if (imageGenerationOptions?.ImageSize is not null && result.Size is null)
+        {
+            // Use a custom type to ensure the size is formatted correctly.
```
stephentoub: I don't understand this comment. What is the "custom type" being used? Also, the constructor for ImageGenerationToolSize states:
```
    /// <b>Note:</b> arbitrary dimensions are not supported and a given model will only support a set of predefined
    /// sizes. If supported dimensions are not known, try using one of the static properties like <see cref="W1024xH1024"/>.
```
Is it going to be likely that a size the user provides won't work? Do we know what sizes will work and should be normalize to them? Or it's model-specific and we have no way of knowing what's valid?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -123,30 +123,114 @@ static async Task AddMessagesAsync(
...
+    /// <exception cref="ArgumentNullException"><paramref name="response"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="update"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// This method modifies the existing <paramref name="response"/> by incorporating the content and metadata
+    /// from the <paramref name="update"/>. This includes using <see cref="ChatResponseUpdate.MessageId"/> to determine
+    /// message boundaries, as well as coalescing contiguous <see cref="AIContent"/> items where applicable, e.g. multiple
+    /// <see cref="TextContent"/> instances in a row may be combined into a single <see cref="TextContent"/>.
+    /// </remarks>
+    [Experimental("MEAI0001")]
+    public static void ApplyUpdate(this ChatResponse response, ChatResponseUpdate update)
```
stephentoub: Why do these ApplyUpdate(s) methods need to be public? I'm not seeing anything in this PR that necessitates that.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ImageGeneratingChatClient.cs
```diff
@@ -0,0 +1,479 @@
...
+/// The provided implementation of <see cref="IChatClient"/> is thread-safe for concurrent use so long as the
+/// <see cref="IImageGenerator"/> employed is also thread-safe for concurrent use.
+/// </para>
+/// <para>
+/// This client automatically detects <see cref="HostedImageGenerationTool"/> instances in the <see cref="ChatOptions.Tools"/> collection
+/// and replaces them with equivalent function tools that the chat client can invoke to perform image generation and editing operations.
+/// </para>
+/// </remarks>
+[Experimental("MEAI001")]
+public sealed class ImageGeneratingChatClient : DelegatingChatClient
```
stephentoub, line 28: What is your informed opinion about how baked / robust this is? I'm wondering if we should "ship" it as a sample or the like for now.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
```diff
@@ -58,6 +58,30 @@ public ChatResponseUpdate(ChatRole? role, IList<AIContent>? contents)
...
 
+    /// <summary>
+    /// Creates a new ChatResponseUpdate instance that is a copy of the current object.
+    /// </summary>
+    /// <remarks>The cloned object is a shallow copy; reference-type properties will reference the same
+    /// objects as the original. Use this method to duplicate the response update for further modification without
+    /// affecting the original instance.</remarks>
+    /// <returns>A new ChatResponseUpdate object with the same property values as the current instance.</returns>
+    [Experimental("MEAI001")]
+    public ChatResponseUpdate Clone() =>
```
stephentoub: If we have a need for this, I don't think it needs to be experimental; this matches the shape we use elsewhere, so it's unlikely feedback is going to influence it. The most I could imagine us changing would be making it virtual, which we could do in the future (technically that can lead to issues where overrides aren't called from existing binaries that are using call rather than callvirt, but that's a very niche concern).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -546,6 +585,45 @@ internal static FunctionTool ToResponseTool(AIFunctionDeclaration aiFunction, Ch
             aiFunction.Description);
     }
 
+    internal static ImageGenerationTool ToImageResponseTool(HostedImageGenerationTool imageGenerationTool)
+    {
+        ImageGenerationOptions? imageGenerationOptions = imageGenerationTool.Options;
+
+        // Not every option is available on ImageGenerationOptions, so we allow the tool to define a factory for 
+        // the OpenAi ImageGenerationTool.  This is important to set partial_images, input_image_mask, etc.
+        var result = imageGenerationTool.RawRepresentationFactory?.Invoke(null) as ImageGenerationTool ?? new();
```
stephentoub: Do the options exposed on ImageGenerationTool at all change any opinions we held on what should be on ImageGenerationOptions?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: I created these to facilitate streaming scenarios in consumption of ImageGenerationTool.  As images streamed in I would call "ApplyUpdate" to incrementally apply the coalescing behavior where a more detailed image replaced a previous one.

If we don't have a method like this then the caller needs to do the image mapping and replacement themselves.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
ericstj: Ok, I was forced to make this experimental based on our analyzer, so I just suppress that?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedImageGenerationTool.cs
ericstj: The reason I created this was to allow the caller to specify `partial_images` which is important for streaming -- https://github.com/ericstj/imageGeneratorSample/blob/794e375f100c23726930f50fd9010d78d37140f0/imageGeneratorSample.Web/Components/Pages/Chat/Chat.razor#L68-L71  (this used a previous iteration)  It's the primary indicator that a caller wants partial images sent back as the model works on generating images.

The value needs to eventually be set on ImageGenerationTool we pass to OpenAI -- https://platform.openai.com/docs/api-reference/responses/create#responses_create-tools-image_generation_tool

> then you can instead do the exact same thing with ChatOptions.RawRepresentationFactory

The factory for the options doesn't work here, unless we expose a new property on the options type that OpenAI provider can map, since the property doesn't live on the provider-specific options, but on the tool.  Those don't accept an input parameter that would allow a callback to choose to return a tool.

Do you have a suggestion for how we might do this differently?  I could expose `partial_images` on the tool or the image generation options.  I didn't so far since it seemed to be very provider specific.

stephentoub: I'm not understanding. Why can't someone do:
```csharp
options.RawRepresentationFactory = _ => new ResponseCreationOptions
{
    Tools = { new ImageGenerationTool { PartialImageCount = 3 } }
};
```
or
```
options.Tools = [new ImageGenerationTool { PartialImageCount = 3 }.AsAITool()];
```
?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
stephentoub: I'm unclear why that's needed. If I'm consuming the streaming updates, rendering a partial update as I receive it, why do I need the partial coalescing behavior? I would assume I would just, e.g. update my UI with the new partial image as I receive it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
stephentoub: > so I just suppress that?

You mark it as stable in the json file. Basically we're shipping it as supported immediately.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIJsonContext.cs
ericstj: Probably this was from my earlier workarounds when I was using the serializer to create the openAI types that weren't exposed yet.  I'll remove them.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
ericstj: I see, I can add it to the JSON file as stable.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
ericstj: The `GetImagesForEdit` function in `ImageGeneratingChatClient`

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
ericstj: Yes, it would be helped by that. That scenario is what I was currently trying to do by coalescing image content on name.  It looks like I never applied it here, but here's what I was doing in the sample -- https://github.com/ericstj/imageGeneratorSample/blob/794e375f100c23726930f50fd9010d78d37140f0/imageGeneratorSample.Web/Components/Pages/Chat/Chat.razor#L121C77-L121C85

I'd pull out the `ItemId` and use that as the name of the image.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
ericstj: Discussed this above -- note that this is a factory on the tool itself, not the options.  Open AI doesn't have a different type to to represent tool options, they are all set directly on the `ImageGenerationTool`.  That's why I wanted to have some sort of factory for folks to create a pre-populated `ImageGenerationTool`.

The options for the tool are a subset of those available for image generation.  I do think it makes sense to have the same options type for a simpler API surface.

The one option that I really need the factory for is `partial_images` which is the count of streaming images to provide.  I might just go ahead and add that to options to avoid the need for the factory.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
stephentoub: You could avoid that just by changing its signature to return `IEnumerable<string>`. Then you could also avoid the `.ToArray()` it does.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
ericstj: This is a stale comment - previously I was creating a json object and passed the size into a another type just to format it correctly.  https://github.com/dotnet/extensions/commit/ffe9a92b156ed63926d7e0095f9c36357082b92a#diff-58f02424a0edcbbe3104602b2fa7b2f41437a03a249519c1ce8eb515bd666727R371-R375

I'll clean up the comment.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
ericstj: Got it.  I will fix that up.  I think that makes the ImageGeneratingChatClient a bit easier to implement too.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: I think this is related to the scenario you are describing here https://github.com/dotnet/extensions/issues/6965.  I think that would also benefit from these APIs.

>  I would assume I would just, e.g. update my UI with the new partial image as I receive it.

Sure, but you then have to _know_ which image replaces which other image.  A similar bit of logic is present for folks handling streaming text as well.  Essentially we say that we implement coalescing behavior for streaming but the only way we expose that is to coalesce the final result, which defeats the purpose of streaming.  If we can represent incremental coalescing behavior for the user it might be better.  

I don't feel too strongly about these APIs, they just seemed like they would eliminate boilerplate for anyone doing streaming - but if you aren't seeing the value here we can remove them.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ImageGeneratingChatClient.cs
ericstj: It's ok, not as baked as the other types in this PR - but that's because it relies on the model to make the right decision.  It works good for generate, mostly good for edit.  I wonder how useful this will be in the long run if chat clients are moving towards supporting more built-in modalities.  Maybe it is still useful for smaller self-hosted models.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedImageGenerationTool.cs
ericstj: Then they can't use `HostedImageGenerationTool` nor `ImageGenerationOptions` - so the moment they want to set some property that's not exposed they have to completely change how set up the client, vs with options they only need to use RawRepresentationFactory for the things which aren't exposed.

stephentoub: Yes, that's no different from any of the other hosted tools. I don't know why this one is special. Is the suggestion that we do this for all of the hosted tools?

ericstj: Perhaps, I was surprised we had this elsewhere but not for the tools.  I said the same in standup the other day.  The pass through API (factory/RawRepresentation) could be applied more consistently, but I don't want to force that here if you don't feel it's worth while.

I decided to remove it for now and expose a member on options instead to represent the parameter I needed.  It is supported by the backing OpenAI API on both the tool and the generator, it's just that the .NET image OpenAI ImageClient doesn't expose it (or streaming at all) right now.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
ericstj: This pattern isn't exactly helpful here.  

In the non-streaming case we don't get call/result - only [result](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-image_generation_call).

In the streaming case we get [in_progress](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/in_progress), [generating](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/generating), [partial_image](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/partial_image) (the actual image data), [complete](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/completed).

stephentoub: > In the non-streaming case we don't get call/result - only [result](https://platform.openai.com/docs/api-reference/responses/object#responses-object-output-image_generation_call).

That's fine; we can just have the Result half. I wouldn't be surprised if in the future prompt information were exposed, which would be the call half.

> In the streaming case we get [in_progress](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/in_progress), [generating](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/generating), [partial_image](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/partial_image) (the actual image data), [complete](https://platform.openai.com/docs/api-reference/responses-streaming/response/image_generation_call/completed).

Are in_progress and generating meaningful? I'd expect we could just output an XxResult instance for each updated image, including partials and the final version, no?

test/Libraries/Microsoft.Extensions.AI.OllamaSharp.Integration.Tests/OllamaSharpImageGeneratingChatClientIntegrationTests.cs
ericstj: No, that's why we can use the ImageGeneratingChatClient to add it.  We just need the chat model to support function calling.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
ericstj: > Are in_progress and generating meaningful?

I imagine they might be meaningful for visualizing some progress.  I'm thinking to surface the first as the `call` content.

>  I'd expect we could just output an XxResult instance for each updated image, including partials and the final version, no?

Yep, that's exactly what I'm doing. I was just listing out all the events sent in case you thought any were interesting or changed how you thought of the call/result pattern.  I think we're fine with call/result for now.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: I'll reimplement this now in terms of  `ImageGenerationToolResultContent` and have it coalesce on ID, combining partial images to the final one.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
ericstj: I opened https://github.com/dotnet/extensions/issues/7017 to track this idea.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ImageGeneratingChatClient.cs
```diff
@@ -0,0 +1,517 @@
...
+    /// <summary>
+    /// Specifies how image and other data content is handled when passing data to an inner client.
+    /// </summary>
+    /// <remarks>
+    /// Use this enumeration to control whether images in the data content are passed as-is, replaced
+    /// with unique identifiers, or only generated images are replaced. This setting affects how downstream clients
+    /// receive and process image data.
+    /// Reducing what's passed downstream can help manage the context window.
+    /// </remarks>
+    public enum DataContentHandling
```
ericstj, line 39: The purpose of this is to control which images get removed from messages (to be treated only candidates for edit) vs which are sent across (to be candidates both for edit and  vision / model understanding).

The reason for doing this is to avoid blowing out the cotnext window with very large images that are never meant to be consumed by the model.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/ImageGenerationToolCallContent.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Initializes a new instance of the <see cref="ImageGenerationToolCallContent"/> class.
+    /// </summary>
+    public ImageGenerationToolCallContent()
+    {
+    }
+
+    /// <summary>
+    /// Gets or sets the unique identifier of the image generation item.
+    /// </summary>
+    public string? ImageId { get; set; }
```
stephentoub, line 24: "ImageId" makes it sound to me like this is the ID of the resulting image. What is this? Is it the e equivalent of call id? Would RequestId be better?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Image/ImageGenerationOptions.cs
```diff
@@ -81,6 +81,11 @@ protected ImageGenerationOptions(ImageGenerationOptions? other)
     /// </summary>
     public ImageGenerationResponseFormat? ResponseFormat { get; set; }
 
+    /// <summary>
+    /// Gets or sets the number of intermediate streaming images to generate.
```
stephentoub, line 85: This is only relevant for streaming requests, right? Might be worth a comment. Should partial or intermediate be in the name? Otherwise it seems like it could be confused with Count, like Count applies to non-streaming and this is the number to generate for streaming. Also, is this per image? Like, if I set Count to 4 and this to 5, will I get 5 or 20?

ericstj: You get 20.  This is the number of intermediate images returned including the final image.  Of course the chat API also returns the complete image (presumably for people who don't handle partial messages, same as what happens with the text API).  OpenAI calls it partial images in their docs, but partial never resonated with me.  I felt streaming count was more consistent with how we exposed other API.  If you feel different I'm open to other names.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/ImageGenerationToolCallContent.cs
ericstj: OpenAI does represent an ItemId when it calls the tool.  These IDs can correlate request with response.  Sorry for not responding to this earlier. 

In the OpenAI API I mapped this to https://github.com/openai/openai-dotnet/blob/6846155f923357c0b55244a48308f3a10bc9a81a/src/Generated/Models/Responses/StreamingResponseImageGenerationCallInProgressUpdate.cs#L29C23-L29C30

In the middle-ware case I mapped it to the ID of the function call.

PR #6761, opened 2025-09-01
Add reasoning text content for OpenAI Responses ChatClient Streaming

rogerbarreto:
# Motivation and Context

Add support for text reasoning content while streaming in the `OpenAIResponsesChatClient`.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
@@ -9,6 +9,7 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
+using System.Text.Json.Serialization.Metadata;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
@@ -26,6 +27,9 @@ namespace Microsoft.Extensions.AI;
 /// <summary>Represents an <see cref="IChatClient"/> for an <see cref="OpenAIResponseClient"/>.</summary>
 internal sealed class OpenAIResponsesChatClient : IChatClient
 {
+    /// <summary>Type info for serializing and deserializing arbitrary JSON objects.</summary>
+    private static readonly JsonTypeInfo _jsonTypeInfo = AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(object));
+
     /// <summary>Metadata about the client.</summary>
     private readonly ChatClientMetadata _metadata;
 
@@ -324,8 +328,25 @@ internal sealed class OpenAIResponsesChatClient : IChatClient
                     break;
 
                 default:
+                {
+                    // Breaking-Glass capture of streaming reasoning contents
+                    if (streamingUpdate.GetType().Name == "InternalResponseReasoningSummaryTextDeltaEvent")
+                    {
+                        var responseDeltaElement = JsonSerializer.Deserialize(
+                            JsonSerializer.Serialize(streamingUpdate, _jsonTypeInfo),
+                            OpenAIJsonContext.Default.JsonElement);
+
+                        if (responseDeltaElement.TryGetProperty("delta", out var deltaProperty))
+                        {
+                            yield return CreateUpdate(new TextReasoningContent(deltaProperty.GetString()));
+                            break;
+                        }
+                    }
+
                     yield return CreateUpdate();
                     break;
+                }
+
             }
         }
     }
--- a/test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
@@ -169,6 +169,176 @@ public class OpenAIResponseClientTests
     }
 
     [Fact]
+    public async Task BasicReasoningResponse_Streaming()
+    {
+        const string Input = """
+            {
+              "input":[{
+                "type":"message",
+                "role":"user",
+                "content":[{
+                  "type":"input_text",
+                  "text":"Calculate the sum of the first 5 positive integers."
+                }]
+              }],
+              "reasoning": {
+                "summary": "detailed",
+                "effort": "low"
+              },
+              "model": "o4-mini",
+              "stream": true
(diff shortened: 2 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -324,8 +328,25 @@ ChatResponseUpdate CreateUpdate(AIContent? content = null) =>
...
+                    if (streamingUpdate.GetType().Name == "InternalResponseReasoningSummaryTextDeltaEvent")
+                    {
+                        var responseDeltaElement = JsonSerializer.Deserialize(
+                            JsonSerializer.Serialize(streamingUpdate, _jsonTypeInfo),
+                            OpenAIJsonContext.Default.JsonElement);
+
+                        if (responseDeltaElement.TryGetProperty("delta", out var deltaProperty))
+                        {
+                            yield return CreateUpdate(new TextReasoningContent(deltaProperty.GetString()));
+                            break;
```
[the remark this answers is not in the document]
rogerbarreto: I have considered it but that is not AOT friendly, ideally we should have this in the public API of the OpenAI SDK. For now, AFAIK this is the safest approach.

rogerbarreto:

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -324,8 +328,25 @@ ChatResponseUpdate CreateUpdate(AIContent? content = null) =>
                     break;
 
                 default:
+                {
+                    // Breaking-Glass capture of streaming reasoning contents
+                    if (streamingUpdate.GetType().Name == "InternalResponseReasoningSummaryTextDeltaEvent")
+                    {
+                        var responseDeltaElement = JsonSerializer.Deserialize(
+                            JsonSerializer.Serialize(streamingUpdate, _jsonTypeInfo),
+                            OpenAIJsonContext.Default.JsonElement);
```
stephentoub: You're already relying on internal implementation details at this point; why serialize/deserialize/extra property rather than just using reflection? I'm not excited at the reliance here on internal implementation details, but _if_ we're going to do it for the type name, we may as well do it for the property as well.

rogerbarreto: AFAIK, attempting to use reflection here will not be AOT friendly, I definitely can do the reflection if that's not an issue.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -26,6 +27,9 @@ namespace Microsoft.Extensions.AI;
 /// <summary>Represents an <see cref="IChatClient"/> for an <see cref="OpenAIResponseClient"/>.</summary>
 internal sealed class OpenAIResponsesChatClient : IChatClient
 {
+    /// <summary>Type info for serializing and deserializing arbitrary JSON objects.</summary>
```
eiriktsarpalis: In the case of AOT, this would really only work for types included in the source generated types behind `AIJsonUtilities.DefaultOptions`.
```suggestion
    /// <summary>Type info for serializing and deserializing JSON objects.</summary>
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
eiriktsarpalis: Per my remark in https://github.com/dotnet/extensions/pull/6761/files#r2320945912, this will almost certainly fail if running using the source generator because `AIJsonUtilities` has no access to `InternalResponseReasoningSummaryTextDeltaEvent`.

Using reflection to access the property seems safer, since we can assume that the property is being rooted by OpenAI components (until it does not).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
rogerbarreto: Removed the serialization approach, since reflection won't be a problem.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -26,6 +27,11 @@ namespace Microsoft.Extensions.AI;
 /// <summary>Represents an <see cref="IChatClient"/> for an <see cref="OpenAIResponseClient"/>.</summary>
 internal sealed class OpenAIResponsesChatClient : IChatClient
 {
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)]
+    private static readonly Type? _internalResponseReasoningSummaryTextDeltaEventType = Type.GetType("OpenAI.Responses.InternalResponseReasoningSummaryTextDeltaEvent, OpenAI");
+
+    private static readonly PropertyInfo? _summaryTextDeltaProperty = _internalResponseReasoningSummaryTextDeltaEventType?.GetProperty("Delta");
```
stephentoub:
```suggestion
    // Fix this to not use reflection once https://github.com/openai/openai-dotnet/issues/643 is addressed.
    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)]
    private static readonly Type? _internalResponseReasoningSummaryTextDeltaEventType = Type.GetType("OpenAI.Responses.InternalResponseReasoningSummaryTextDeltaEvent, OpenAI");
    private static readonly PropertyInfo? _summaryTextDeltaProperty = _internalResponseReasoningSummaryTextDeltaEventType?.GetProperty("Delta");
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -324,8 +330,18 @@ ChatResponseUpdate CreateUpdate(AIContent? content = null) =>
                     break;
 
                 default:
+                {
+                    // Breaking-Glass capture of streaming reasoning contents
```
stephentoub:
```suggestion
```

PR #6776, opened 2025-09-06
Change OpenTelemetryChatClient/EmbeddingGenerator to log raw additional properties

stephentoub:
For historical reasons (based on older versions of the genai convention), we were mangling the key names. Now just use the key names as sourced from the dictionary, enabling a developer to more easily augment the spans with data of their choice.

https://github.com/dotnet/extensions/discussions/6771 

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6776)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClient.cs
@@ -364,20 +364,13 @@ public sealed partial class OpenTelemetryChatClient : DelegatingChatClient
                         }
                     }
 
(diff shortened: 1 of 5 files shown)


jeffhandley (approved):
Could someone consider this to be a breaking change if they were doing post-processing on the keys or otherwise relying on the key names in the old structure?

stephentoub:
> Could someone consider this to be a breaking change if they were doing post-processing on the keys or otherwise relying on the key names in the old structure?

Yes, but a) these pale in comparison to the changes made for the 1.37 update, and b) the whole impl is under an Experimental otel namespace because the spec is still changing.

Issue #6670, opened 2025-07-31
Middleware for tool reduction


stephentoub:
Especially with MCP, but not limited to it, it's becoming more and more common for clients to have many tools. Too many tools can cause problems, both due to hitting explicit limits on the number of tools a service allows in a request but also due to implicit limits that stem from degraded capability by an LLM to choose the right tool when there are lots. Clients end up needing to implement custom schemes for tool reduction, such as embedding tool descriptions and choosing which subset of tools to incorporate based on an embedding of the supplied messages, or using virtual tools, where tools are grouped (possibly by an LLM) and that grouping is advertized as a single tool, which the LLM can choose to invoke in order to activate all of the tools in that group.

Such schemes can be implemented as middleware, such that a developer can just plug in a scheme into their IChatClient pipeline, e.g.
```C#
IChatClient client = ...;
client = client
    .UseToolReduction(new EmbeddingToolReductionStrategy(embeddingGenerator, 15))
    .UseFunctionInvocation()
    .UseOpenTelemetry()
    .Build();
```

Different strategies could be handled via parameterization of that IChatClient implementation, e.g.
```C#
public interface IToolReductionStrategy
{
    ValueTask<IEnumerable<AITool>> SelectToolsForRequest(IEnumerable<ChatMessage> messages, ChatOptions? options);
}
```
and implementations of that strategy, e.g.
```C#
public sealed class EmbeddingToolReductionStrategy(
    IEmbeddingGenerator<string, Embedding<float>> generator, int toolLimit) : IToolReductionStrategy
{
     public async ValueTask<IEnumerable<AITool>> SelectToolsForRequest(IEnumerable<ChatMessage> messages, ChatOptions? options)
    {
        IList<AITool>? tools = options?.Tools;
        if (tools is null) return [];

        var toolEmbeddings = await generator.GenerateAsync(tools.Select(t => $"{f.Name}\n{f.Description}"));

        Embedding<float> queryEmbedding = await generator.GenerateAsync(string.Concat(messages));
        return tools.Zip(toolEmbeddings)
            .OrderByDescending(f => TensorPrimitives.CosineSimilarity(queryEmbedding.Vector.Span, f.Item1.Vector.Span))
            .Take(toolLimit)
            .Select(f => f.Item1)
            .ToArray();
}
```
That's just a rough sketch, but something along those lines.

jeffhandley:
[...]

PR #6781, opened 2025-09-08
Tool reduction

MackinnonBuck:
### Summary
This draft introduces the tool reduction feature set:
- New abstractions (`IToolReductionStrategy`) for selecting a subset of tools per request.
- Middleware (`ToolReducingChatClient`) that applies a configured strategy automatically.
- An initial embedding-based implementation (`EmbeddingToolReductionStrategy`) as a first strategy.
- Supporting builder extensions and tests.

All components are marked experimental and may evolve. Additional strategies can be added in follow-up work.

### Next Steps
- Additional tool reduction strategy implementations.
- Refinements to abstractions as necessary.

Feedback welcome before expanding scope.

Fixes https://github.com/dotnet/extensions/issues/6670
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6781)

### DIFF
--- a/eng/packages/General.props
+++ b/eng/packages/General.props
@@ -29,6 +29,7 @@
     <PackageVersion Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageVersion Include="System.ComponentModel.Annotations" Version="5.0.0" />
     <PackageVersion Include="System.Memory" Version="4.5.5" />
+    <PackageVersion Include="System.Numerics.Tensors" Version="$(SystemNumericsTensorsVersion)" />
     <PackageVersion Include="System.Private.Uri" Version="4.3.2" />
     <PackageVersion Include="System.Runtime.Caching" Version="$(SystemRuntimeCachingVersion)" />
     <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.1.0" />
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -125,7 +125,7 @@
     <SystemSecurityCryptographyPkcsLTSVersion>8.0.1</SystemSecurityCryptographyPkcsLTSVersion>
     <SystemSecurityCryptographyXmlLTSVersion>8.0.2</SystemSecurityCryptographyXmlLTSVersion>
     <SystemTextEncodingsWebLTSVersion>8.0.0</SystemTextEncodingsWebLTSVersion>
-    <SystemNumericsTensorsLTSVersion>8.0.0</SystemNumericsTensorsLTSVersion>
+    <SystemNumericsTensorsLTSVersion>8.0.1</SystemNumericsTensorsLTSVersion>
     <SystemTextJsonLTSVersion>8.0.6</SystemTextJsonLTSVersion>
     <SystemThreadingChannelsLTSVersion>8.0.0</SystemThreadingChannelsLTSVersion>
     <!-- Dependencies from https://github.com/aspnet/AspNetCore -->
--- a/eng/packages/General-LTS.props
+++ b/eng/packages/General-LTS.props
@@ -38,7 +38,6 @@
     <PackageVersion Include="System.IO.Pipelines" Version="$(SystemIOPipelinesLTSVersion)" />
     <PackageVersion Include="System.Memory.Data" Version="$(SystemMemoryDataLTSVersion)" />
     <PackageVersion Include="System.Net.Http.Json" Version="$(SystemNetHttpJsonLTSVersion)" />
-    <PackageVersion Include="System.Numerics.Tensors" Version="$(SystemNumericsTensorsLTSVersion)" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="$(SystemTextEncodingsWebLTSVersion)" />
     <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonLTSVersion)" />
     <PackageVersion Include="System.Threading.Channels" Version="$(SystemThreadingChannelsLTSVersion)" />
--- a/eng/packages/General-net9.props
+++ b/eng/packages/General-net9.props
@@ -1,5 +1,5 @@
 <Project>
-
+  
   <!-- The items in this ItemGroup are packages coming from aspnetcore and runtime repos. When an asset is targeting net9.0 version
   of the framework, the following versions should be used. -->
   <ItemGroup Condition="'$(TargetFramework)' == 'net9.0' Or '$(ForceLatestDotnetVersions)' == 'true'">
@@ -38,7 +38,6 @@
     <PackageVersion Include="System.IO.Pipelines" Version="$(SystemIOPipelinesVersion)" />
     <PackageVersion Include="System.Memory.Data" Version="$(SystemMemoryDataVersion)" />
     <PackageVersion Include="System.Net.Http.Json" Version="$(SystemNetHttpJsonVersion)" />
-    <PackageVersion Include="System.Numerics.Tensors" Version="$(SystemNumericsTensorsVersion)" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="$(SystemTextEncodingsWebVersion)" />
     <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
     <PackageVersion Include="System.Threading.Channels" Version="$(SystemThreadingChannelsVersion)" />


eng/packages/TestOnly.props
```diff
@@ -22,7 +22,7 @@
     <PackageVersion Include="Polly.Testing" Version="8.4.2" />
     <PackageVersion Include="StrongNamer" Version="0.2.5" />
     <PackageVersion Include="System.Configuration.ConfigurationManager" Version="$(SystemConfigurationConfigurationManagerVersion)" />
-    <PackageVersion Include="System.Numerics.Tensors" Version="$(SystemNumericsTensorsVersion)" />
+    <PackageVersion Include="System.Linq.Async" Version="6.0.1" />
```
stephentoub: Can we use System.Linq.AsyncEnumerable instead? System.Linq.Async is now legacy.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+            return t.Name;
+        }
+
+        return t.Name + "\n" + t.Description;
+    };
+
+    private Func<IEnumerable<ChatMessage>, string> _messagesEmbeddingTextFactory = static messages =>
+    {
+        var messageTexts = messages.Select(m => m.Text).Where(s => !string.IsNullOrEmpty(s));
+        return string.Join("\n", messageTexts);
```
stephentoub: We have a helper that ChatMessage and ChatResponse use to do this a bit more efficiently. Maybe we should look at exposing that in a way this could use as well.

That said, I wonder if we'd want the default implementation here also including reasoning content, which this implementation doesn't currently, since ChatMessage.Text only includes TextContent and not TextReasoningContent.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+        IEmbeddingGenerator<string, Embedding<float>> embeddingGenerator,
+        int toolLimit)
+    {
+        _embeddingGenerator = Throw.IfNull(embeddingGenerator);
+        _toolLimit = Throw.IfLessThanOrEqual(toolLimit, min: 0);
+    }
+
+    /// <summary>
+    /// Gets or sets a delegate used to produce the text to embed for a tool.
+    /// Defaults to: <c>Name + "\n" + Description</c> (omitting empty parts).
```
stephentoub: Nit: move to remarks

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+    /// </summary>
+    public Func<AITool, string> ToolEmbeddingTextFactory
```
stephentoub: It's not wrong, but "Factory" feels odd to me personally as the name for these properties.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// <summary>
+    /// Gets or sets a value indicating whether tool embeddings are cached. Defaults to <see langword="true"/>.
+    /// </summary>
+    public bool EnableEmbeddingCaching { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to preserve original ordering of selected tools.
+    /// If <see langword="false"/> (default), tools are ordered by descending similarity.
+    /// If <see langword="true"/>, the top-N tools by similarity are re-emitted in their original order.
```
stephentoub, line 146: Curious where this ordering comes into play?

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+    public async Task<IEnumerable<AITool>> SelectToolsForRequestAsync(
+        IEnumerable<ChatMessage> messages,
+        ChatOptions? options,
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(messages);
+
+        if (options?.Tools is not { Count: > 0 } tools)
+        {
+            return options?.Tools ?? [];
```
stephentoub, line 163: Why not just `[]`?

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+            .Select((t, i) => (t.Tool, Index: i, Score: Similarity(queryVector, t.Embedding.Vector)))
+            .OrderByDescending(t => t.Score)
+            .Take(_toolLimit);
+
+        if (PreserveOriginalOrdering)
+        {
+            ranked = ranked.OrderBy(t => t.Index);
+        }
+
+        return ranked.Select(t => t.Tool);
```
stephentoub: We should probably tack a `.ToList()` onto this. Otherwise, a consumer that iterates through the enumerable multiple times will be executing a fairly expensive LINQ query multiple times.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+    {
+        get => _toolEmbeddingTextFactory;
+        set => _toolEmbeddingTextFactory = Throw.IfNull(value);
+    }
+
+    /// <summary>
+    /// Gets or sets the factory function used to generate a single text string from a collection of chat messages for
+    /// embedding purposes.
+    /// </summary>
+    public Func<IEnumerable<ChatMessage>, string> MessagesEmbeddingTextFactory
```
stephentoub: I wonder if we'd want/need as well another `Func<AITool, bool>`-like "override" that would ensure a particular tool is included even if based on similarity alone it wouldn't be included. Just speculating.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,212 @@
...
+            var embedding = uncachedEmbeddings[i];
+            result[cacheMisses[i].Index] = embedding;
+            _toolEmbeddingsCache.Add(cacheMisses[i].Tool, embedding);
+        }
+
+        return result;
+
+        async ValueTask<GeneratedEmbeddings<Embedding<float>>> ComputeEmbeddingsAsync(IEnumerable<string> texts, int expectedCount)
+        {
+            var embeddings = await _embeddingGenerator.GenerateAsync(texts, cancellationToken: cancellationToken).ConfigureAwait(false);
```
stephentoub: I'd need to double-check the IL, but IIRC, this will end up with an additional allocation in order to capture the cancellationToken because this is an async local function. Consider passing the CT in as a parameter.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/ToolReducingChatClient.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// A delegating chat client that applies a tool reduction strategy before invoking the inner client.
+/// </summary>
+/// <remarks>
+/// Insert this into a pipeline (typically before function invocation middleware) to automatically
```
stephentoub, line 19: I wonder how we'd implement a "virtual tools" strategy ala what VS Code uses. Tools come in, they're grouped by an LLM into buckets, and then fake tools are created for each bucket; those tools are passed along to the LLM, which can then invoke one of those tools in order to expand it out into the full set from that bucket. Probably requires some prototyping. It might be that we leverage AIFunctionDeclaration so that the FICC lets a virtual tool pass through back to the TRCC. Or maybe it's another feature in FICC that enables a tool to modify the options that'll be sent back on a subsequent request. Not sure.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/ToolReducingChatClient.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            return options;
+        }
+
+        IEnumerable<AITool> reduced;
+        try
+        {
+            reduced = await _strategy.SelectToolsForRequestAsync(messages, options, cancellationToken).ConfigureAwait(false);
+        }
+        catch (OperationCanceledException)
```
stephentoub: This doesn't look right. Why would cancellation not be allowed to propagate?

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
MackinnonBuck: I was thinking that consistent ordering might matter if there's some kind of downstream cache that uses the set of available tools as a key, or a chat client that expects a meaningful relative tool order. We can probably remove this though, as these are just hypothetical scenarios for now.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
MackinnonBuck: I agree it's a little weird. I'll think of a better name.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
MackinnonBuck: We could do that, but `ToolReducingChatClient` has an optimization where if the returned `IEnumerable<AITool>` is reference-equal to the original tools list, it just returns the original options rather than copying the result into a `List<AITool>` and using a cloned `ChatOptions`.

I can add a comment to clarify this.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
MackinnonBuck: Good idea.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/ToolReducingChatClient.cs
MackinnonBuck: You're right, it should propogate.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
PederHP: Ordering matters for caching yes. Caching, whether automatic (OpenAI) or explicit (Anthropic) is prefix-based, so it requries everything up to a certain point in the raw request to be the same. Stable tool ordering is important.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/ToolReducingChatClient.cs
PederHP: When I've used the "virtual tools" strategy I've had success with having a single tool that allows expansion. There's no need to have a tool per bucket. It can just be an argument to the expand and collapse tools, with the bucket descriptions passed in the meta-tool description or injected into system. This is greatly more token efficient and scales better in terms of tool count. This also makes it a well-known tool with some dynamic context rather than an entirely dynamic tool.

MackinnonBuck: @PederHP, that sounds like a promising approach. Just to clarify: is the LLM responsible for both expanding *and* collapsing tool groups, or does it only expand them, with collapsing happening automatically later (e.g. after a response completes)? In other words, how long does an expanded group remain open?

I also wonder if tool grouping might be more naturally handled through agents. Each agent has a limited toolset tailored to its purpose, so as work is handed off between agents, the available tools change without needing stateful expand/collapse logic in a single chat client. A lightweight "tool reduction" middleware in the chat client (such as what's prototyped in this PR) could still help narrow focus further, but in a more stateless way.

If we *do* want multiple tool groups in a single client, one option could be allowing only one group to be expanded at a time. This would mimic the agent handoff scenario I described above. All groups are collapsed at the start of a response, and the LLM can call a well-known, non-invocable "expand" function with a tool group name. A middleware earlier in the pipeline can then generate another response with the expanded set of tools. Further expansions close the already-expanded tool group. This approach reduces the responsibility of the LLM to decide when to collapse a tool group, and it allows tool reduction middleware to run after each expansion, if needed. That type of composition might be harder to achieve if functions can directly mutate the tools list inside `FunctionInvokingChatClient` iterations.

MackinnonBuck: Implemented a prototype in https://github.com/dotnet/extensions/pull/6865

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
```diff
@@ -0,0 +1,330 @@
...
+namespace Microsoft.Extensions.AI;
+
+#pragma warning disable IDE0032 // Use auto property, suppressed until repo updates to C# 14
+
+/// <summary>
+/// A tool reduction strategy that ranks tools by embedding similarity to the current conversation context.
+/// </summary>
+/// <remarks>
+/// The strategy embeds each tool (name + description by default) once (cached) and embeds the current
+/// conversation content each request. It then selects the top <c>toolLimit</c> tools by similarity.
```
ericstj, line 25: I wonder if there is a hybrid that can reduce tools, but then also expose a "get more tools" function where the model could describe the sort of tool it wanted and we could use that to give it back more tools that have embedding similarity to its request?  Don't want to feature creep this, just trying to think of alternatives.

I like that this one is hands free but but I can imagine that it can hit situations where it never gives the model the chance to call what might be the best tool.

I like that the grouping ensures that all tools are made available, but it requires explicit grouping by the caller.

MackinnonBuck: Maybe what we could do is change `MessagesEmbeddingTextSelector` to be a `Func<IEnumerable<ChatMessage>, Task<string>>`. Then if you wanted to have the model separately describe the tool it needs, you could asynchronously have a chat client analyze the conversation and generate a short string describing the type of tool that would be helpful. Then that string is what gets used when searching for "similar" tools.

MackinnonBuck: Just implemented the above ^

eng/packages/General.props
```diff
@@ -29,6 +29,7 @@
     <PackageVersion Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageVersion Include="System.ComponentModel.Annotations" Version="5.0.0" />
     <PackageVersion Include="System.Memory" Version="4.5.5" />
+    <PackageVersion Include="System.Numerics.Tensors" Version="$(SystemNumericsTensorsVersion)" />
```
ericstj: Confirming that the 9.0.10 version here is what you want.  We do have other runtime package references @8.0.

Numerics.Tesnsors may only have the API you need in 9.0 and later  though.  Luckily it won't bring other dependencies in.  It might be OK as a the only package we allow to come from 9.0 runtime.  cc @joperezr

MackinnonBuck: It looks like the required APIs are also present in version 8.0.0 of the package.

Just pushed a new commit to address this.

MackinnonBuck: Well, looks like I was wrong. The 9.0 dependency seems required after all.

src/Libraries/Microsoft.Extensions.AI/ToolReduction/EmbeddingToolReductionStrategy.cs
MackinnonBuck: The default implementation now includes `TextContent` and `TextReasoningContent`.

PR #6783, opened 2025-09-09
Add measures and tags with mediator object to the HttpClientLatencyLogEnricher

rainsxng:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6783)


src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpClientLatencyLogEnricher.cs
```diff
@@ -23,62 +23,151 @@ internal sealed class HttpClientLatencyLogEnricher : IHttpClientLogEnricher
...
+    private readonly HttpLatencyMediator _httpLatencyMediator;
+#endif
     private readonly CheckpointToken _enricherInvoked;
 
-    public HttpClientLatencyLogEnricher(HttpClientLatencyContext latencyContext, ILatencyContextTokenIssuer tokenIssuer)
+#if NET
+    public HttpClientLatencyLogEnricher(
+        HttpClientLatencyContext latencyContext,
+        ILatencyContextTokenIssuer tokenIssuer,
+        HttpLatencyMediator httpLatencyMediator)
```
evgenyfedorov2, line 32: we need the mediator regardless of .NET version (you correctly have both .NET and .NET FX versions of that class already), consider removing conditional compilation here.

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpClientLatencyLogEnricher.cs
```diff
@@ -23,62 +23,151 @@ internal sealed class HttpClientLatencyLogEnricher : IHttpClientLogEnricher
...
-    public void Enrich(IEnrichmentTagCollector collector, HttpRequestMessage request, HttpResponseMessage? response, Exception? exception)
+    public void Enrich(IEnrichmentTagCollector collector, HttpRequestMessage? request, HttpResponseMessage? response, Exception? exception)
     {
         if (response != null)
         {
             var lc = _latencyContext.Get();
-            lc?.AddCheckpoint(_enricherInvoked);
+            if (lc == null)
+            {
+                return;
```
evgenyfedorov2: We don't `return` in this case in the internal repo, right? Why doing it here? I mean, internally we still add enrichment stuff regardless of `lc` value

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/HttpClientLatencyTelemetryExtensions.cs
```diff
@@ -33,6 +33,9 @@ public static IServiceCollection AddHttpClientLatencyTelemetry(this IServiceColl
         _ = services.AddSingleton<HttpRequestLatencyListener>();
         _ = services.AddSingleton<HttpClientLatencyContext>();
         _ = services.AddTransient<HttpLatencyTelemetryHandler>();
+#if NET
```
evgenyfedorov2: don't we also need to do
```
        _ = services.RegisterMeasureNames(HttpMeasures.Measures);
        _ = services.RegisterTagNames(HttpTags.Tags);
```

?

rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpClientLatencyLogEnricher.cs
rainsxng: Done

src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpClientLatencyLogEnricher.cs
rainsxng: Done

evgenyfedorov2:
there should be changes in the https://github.com/dotnet/extensions/blob/cdb9ce445cf26ea3969284f0f9e53fdb2f7faab8/src/Libraries/Microsoft.Extensions.Http.Diagnostics/Latency/Internal/HttpRequestLatencyListener.cs file too

evgenyfedorov2:
LGTM 
@ananthsr could you please review this as SME?

evgenyfedorov2 (approved):

Issue #6510, opened 2025-06-11
Consider adding an EncryptedReasoningBlock


stephentoub:
We added TextReasoningContent, which is already supported by a variety of providers. Some AI services, however, also support encrypted reasonining content, where the reasoning information isn't sent back in plain text but rather in an encrypted form. It's provided not for the consumer but to be roundtripped back to the service, where it will be able to consume it. We should consider exposing an EncryptedReasoningBlock to abstract over such data.

jeffhandley:
Anthropic and OpenAI both have this concept as of now.

PR #6784, opened 2025-09-09
Add TextReasoningContent.ProtectedData

stephentoub:
Fixes https://github.com/dotnet/extensions/issues/6510

I considered multiple approaches, including having a separate content type dedicated to encrypted/redacted/protected data, but ended up just extending the existing TextReasoningContent with a new optional ProtectedData property.

Anthropic has thinking content, which is human-readable text that's accompanied by an opaque "signature" blob: that's representable with the thinking as Text and the signature as ProtectedData. It also has redacted thinking content, which is just an opaque data string, which we can represent with empty Text and data as ProtectedData.

OpenAI lets users of the Responses API opt-in to including an encrypted version of the reasoning trace in their output, which is most useful when the developer also opts-out of storing the responses in the service and thus needs to roundtrip the data back. By default, reasoning output only includes human-readable text, but if opted into encrypted reasoning, the output items will typically include both the summary human-readable text and the encrypted data. That can be modeled using Text for the summary (as we do today) and ProtectedData for the encrypted content.

AWS Bedrock exposes reasoning via a ReasoningContentBlock, which has both reasoningText (human readable) and reasoningContent (opaque encrypted content). However, it'll only ever use one of them in each instance, so we can end up with either Text being non-empty or ProtectedData being non-null.

Gemini has a Part that has both a thought (human readable) and thoughtSignature (opaque data), which we can represent as Text and ProtectedData, respectively.

The one thing gnawing at me is that we have Text right now as always returning a non-null string: if you give it null, it gives you back empty. That means the only way we have to distinguish "only ProtectedData" from "Text and ProtectedData" is if Text is empty. Other than bugs, I don't think there are real situations where one of these services sends back empty Text, but if we find that can happen, we might either need to take a (non-binary breaking) change to have Text be nullable, or add some sort of additional property to the instance that specifies what kind of data it carries.

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseExtensions.cs
@@ -186,17 +186,17 @@ public static class ChatResponseExtensions
     /// <summary>Coalesces sequential <see cref="AIContent"/> content elements.</summary>
     internal static void CoalesceTextContent(IList<AIContent> contents)
     {
-        Coalesce<TextContent>(contents, mergeSingle: false, static (contents, start, end) =>
-            new(MergeText(contents, start, end))
-            {
-                AdditionalProperties = contents[start].AdditionalProperties?.Clone()
-            });
+        Coalesce<TextContent>(
+            contents,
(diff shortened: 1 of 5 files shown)


stephentoub:
cc: @tghamm

eiriktsarpalis (approved):

Issue #6179, opened 2025-03-23
[API Proposal]: Add ChatResponseFormat.ForType


HavenDV:
### Background and motivation

The proposed `ChatResponseFormat.ForType` method automates generating JSON schemas directly from .NET types, reducing boilerplate and errors. This makes it easier and faster to build consistent and type-safe AI applications.

### API Proposal

```csharp
namespace Microsoft.Extensions.AI

public class ChatResponseFormat
{
  public static ChatResponseFormatJson ForType(
      Type type,
      string? schemaName = null,
      string? schemaDescription = null)
    {
      return ChatResponseFormat.ForJsonSchema(
                        JsonSerializerOptions.Default.GetJsonSchemaAsNode(type, new JsonSchemaExporterOptions
                        {
                            TreatNullObliviousAsNonNullable = true,
                        }).Deserialize<JsonElement>(), schemaName, schemaDescription);
    }
}
```

### API Usage

```csharp
var chatResponseFormat = ChatResponseFormat.ForType(typeof(CustomType));
```

### Alternative Designs

Not sure about ForType<T>, it can be alternative or also added as part of this

### Risks

Not sure

stephentoub:
Are you then deserializing the resulting data as a `T`? And if so, is `GetResponseAsync<T>` missing something?

HavenDV:
For some cases yes, for some no.

I would consider it more as refactoring some of the internal logic of `GetResponseAsync<T>` into a new `ChatResponseFormat.ForType` method to increase flexibility of use

PR #6786, opened 2025-09-09
Add ChatResponseFormat.ForJsonSchema<T>

stephentoub:
Moves the implementation used by `GetResponseAsync<T>` to be exposed as its own helper that can be used directly, for cases where folks want the formatting but don't want to use the generic `GetResponseAsync<T>`.

Fixes https://github.com/dotnet/extensions/issues/6179
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6786)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
@@ -1,11 +1,13 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.ComponentModel;
 using System.Reflection;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using System.Text.RegularExpressions;
+using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
 
@@ -21,12 +23,6 @@ public partial class ChatResponseFormat
     private static readonly AIJsonSchemaCreateOptions _inferenceOptions = new()
     {
         IncludeSchemaKeyword = true,
-        TransformOptions = new AIJsonSchemaTransformOptions
-        {
-            DisallowAdditionalProperties = true,
-            RequireAllProperties = true,
-            MoveDefaultKeywordToDescription = true,
-        },
     };
 
     /// <summary>Initializes a new instance of the <see cref="ChatResponseFormat"/> class.</summary>
@@ -50,7 +46,7 @@ public partial class ChatResponseFormat
         JsonElement schema, string? schemaName = null, string? schemaDescription = null) =>
         new(schema, schemaName, schemaDescription);
 
-    /// <summary>Creates a <see cref="ChatResponseFormatJson"/> representing structured JSON data with the specified schema.</summary>
+    /// <summary>Creates a <see cref="ChatResponseFormatJson"/> representing structured JSON data with a schema based on <typeparamref name="T"/>.</summary>
     /// <typeparam name="T">The type for which a schema should be exported and used as the response schema.</typeparam>
     /// <param name="serializerOptions">The JSON serialization options to use.</param>
     /// <param name="schemaName">An optional name of the schema. By default, this will be inferred from <typeparamref name="T"/>.</param>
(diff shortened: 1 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
```diff
@@ -1,19 +1,34 @@
...
 
+using System.ComponentModel;
+using System.Reflection;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.Extensions.AI;
 
+#pragma warning disable CA1052 // Static holder types should be Static or NotInheritable
```
eiriktsarpalis, line 14: NotInheritable? I haven't seen sealed types called that way before.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
```diff
@@ -33,7 +48,41 @@ private protected ChatResponseFormat()
...
+    /// <param name="schemaDescription">An optional description of the schema. By default, this will be inferred from <typeparamref name="T"/>.</param>
+    /// <returns>The <see cref="ChatResponseFormatJson"/> instance.</returns>
+    /// <remarks>
+    /// Many AI services that support structured output require that the JSON schema have a top-level 'type=object'.
+    /// If <typeparamref name="T"/> is a primitive type like <see cref="string"/>, <see cref="int"/>, or <see cref="bool"/>,
+    /// or if it's a type that serializes as a JSON array, attempting to use the resulting schema with such services may fail.
+    /// In such cases, consider instead using a <typeparamref name="T"/> that wraps the actual type in a class or struct so that
+    /// it serializes as a JSON object with the original type as a property of that object.
+    /// </remarks>
+    public static ChatResponseFormatJson ForJsonSchema<T>(
```
eiriktsarpalis, line 62: Does this need to be generic? A `Type` parameter should suffice.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
```diff
@@ -1,19 +1,34 @@
...
-public class ChatResponseFormat
-#pragma warning restore CA1052
+public partial class ChatResponseFormat
 {
+    private static readonly AIJsonSchemaCreateOptions _inferenceOptions = new()
+    {
+        IncludeSchemaKeyword = true,
+        TransformOptions = new AIJsonSchemaTransformOptions
+        {
+            DisallowAdditionalProperties = true,
```
eiriktsarpalis: I realize that this has been moved from the structured output extensions, but are these hardcoded transformations still necessary after the introduction of chat client level transforms?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
stephentoub: VB
https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/modifiers/notinheritable

I didn't write the comment; that's just what VS writes as part of suppressing the warning in source, copying I believe the title from the diagnostic.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
stephentoub: We could do both? Or you think it should just be with a Type?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
stephentoub: You tell me :)

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
eiriktsarpalis: We could do both.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseFormat.cs
eiriktsarpalis: It would probably require exposing transformer capabilities on the `IChatClient` interface for the extension method to be able to access, so clearly not in scope for this change.

stephentoub: > It would probably require exposing transformer capabilities on the IChatClient interface for the extension method to be able to access

Why is that? Wouldn't the intent be that whatever this produced, the leaf IChatClient would transform it to what it needs?

eiriktsarpalis: The extension method has a mode that dumps the JSON schema into a chat message. No transforms would be applicable there.

stephentoub: > The extension method has a mode that dumps the JSON schema into a chat message. No transforms would be applicable there.

Sure, but in that mode it's also not using actual structured output or subject to any constraints an LLM might impose on a schema used... it's just part of a prompt.

eiriktsarpalis: Good point. So it seems we could use the default schema generation settings in that case.

stephentoub: Should I delete the whole _inferenceOptions ? Keep it with just IncludeSchemaKeyword = true?

eiriktsarpalis: Maybe we should only include the schema keyword when we're putting in the chat message, to provide a better hint for the LLM.

eiriktsarpalis: Seems we already hardcode a schema string in that mode though, so I think we should just delete that.

stephentoub: > Seems we already hardcode a schema string in that mode though

Where? I see that being included in the object wrapper that's applied to primitives/arrays... is there another?

eiriktsarpalis: My bad, I thought it was being applied to every schema.

PR #6809, opened 2025-09-15
Add otel middleware for IImageGenerator

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6809)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
@@ -1,0 +1,313 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Metrics;
+using System.Drawing;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Shared.Diagnostics;
+
+#pragma warning disable S3358 // Ternary operators should not be nested
+#pragma warning disable SA1111 // Closing parenthesis should be on line of last parameter
+#pragma warning disable SA1113 // Comma should be on the same line as previous parameter
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents a delegating image generator that implements the OpenTelemetry Semantic Conventions for Generative AI systems.</summary>
+/// <remarks>
+/// This class provides an implementation of the Semantic Conventions for Generative AI systems v1.37, defined at <see href="https://opentelemetry.io/docs/specs/semconv/gen-ai/" />.
+/// The specification is still experimental and subject to change; as such, the telemetry output by this client is also subject to change.
+/// </remarks>
+[Experimental("MEAI001")]
+public sealed class OpenTelemetryImageGenerator : DelegatingImageGenerator
+{
+    private readonly ActivitySource _activitySource;
+    private readonly Meter _meter;
+
+    private readonly Histogram<int> _tokenUsageHistogram;
+    private readonly Histogram<double> _operationDurationHistogram;
+
+    private readonly string? _defaultModelId;
+    private readonly string? _providerName;
+    private readonly string? _serverAddress;
+    private readonly int _serverPort;
+
+    /// <summary>Initializes a new instance of the <see cref="OpenTelemetryImageGenerator"/> class.</summary>
+    /// <param name="innerGenerator">The underlying <see cref="IImageGenerator"/>.</param>
+    /// <param name="logger">The <see cref="ILogger"/> to use for emitting any logging data from the client.</param>
+    /// <param name="sourceName">An optional source name that will be used on the telemetry data.</param>
+#pragma warning disable IDE0060 // Remove unused parameter; it exists for consistency with IChatClient and future use
+    public OpenTelemetryImageGenerator(IImageGenerator innerGenerator, ILogger? logger = null, string? sourceName = null)
+#pragma warning restore IDE0060
+        : base(innerGenerator)
+    {
+        Debug.Assert(innerGenerator is not null, "Should have been validated by the base ctor");
+
+        if (innerGenerator!.GetService<ImageGeneratorMetadata>() is ImageGeneratorMetadata metadata)
+        {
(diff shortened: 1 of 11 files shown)


test/Libraries/Microsoft.Extensions.AI.Tests/Image/OpenTelemetryImageGeneratorTests.cs
```diff
@@ -0,0 +1,163 @@
...
+                    "role": "assistant",
+                    "parts": [
+                      {
+                        "type": "image",
+                        "content": "http://example/output.png"
+                      },
+                      {
+                        "type": "image",
+                        "content": "data:image/png;base64,AQIDBA=="
+                      }
```
JamesNK, line 149: Is the image part from a standard, or is it currently under discussion? Could you point me towards where it is being discussed? I don't see it in the current schema: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-output-messages.json

Aspire GenAI viewer could be improved to display the image. Either have a src with the URL, or use the base64 data.

I don't see any other reaction required in GenAI viewer, but let me know if I've missed something.

stephentoub: > Is the image part from a standard

Nope. It has "generate_content" and "image" defined in https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/, and it has the "GenericPart" which this is using and which is documented as "Represents an arbitrary message part with any type and properties.\nThis allows for extensibility with custom message part types.". Basically I'm just trying to piggyback on what's already there for chat and embeddings.

JamesNK: Ok.

At a minimum I'll improve the GenAI viewer so it displays the additional properties for generic parts. I don't think it displays anything at the moment.

And maybe then add I'll some special case code in case the part has `type="image"`. If a standard springs up in the future then we can both move towards it. Could you create an issue on the OTEL semantic conventions about image parts if it does already exist? It would be nice if it was standardized.

JamesNK: I see there are quite a few new generic parts in this PR.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
```diff
@@ -0,0 +1,313 @@
...
+        base.GetService(serviceType, serviceKey);
+
+    /// <inheritdoc/>
+    public async override Task<ImageGenerationResponse> GenerateAsync(
+        ImageGenerationRequest request, ImageGenerationOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(request);
+
+        using Activity? activity = CreateAndConfigureActivity(request, options);
+        Stopwatch? stopwatch = _operationDurationHistogram.Enabled ? Stopwatch.StartNew() : null;
```
eiriktsarpalis, line 124: Is this the standard way to measure activity duration in otel? I wonder if there should be an API that pools stopwatch objects.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
```diff
@@ -0,0 +1,313 @@
...
+
+    /// <summary>Adds image generation response information to the activity.</summary>
+    private void TraceResponse(
+        Activity? activity,
+        string? requestModelId,
+        ImageGenerationResponse? response,
+        Exception? error,
+        Stopwatch? stopwatch)
+    {
+        if (_operationDurationHistogram.Enabled && stopwatch is not null)
```
eiriktsarpalis, line 228: Aren't the two conditions equivalent?

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
JamesNK: In other telemetry code I've been using `Stopwatch.GetTimestamp` and `Stopwatch.GetElapsedTime` instead of creating a stopwatch.

test/Libraries/Microsoft.Extensions.AI.Tests/Image/OpenTelemetryImageGeneratorTests.cs
stephentoub: > I see there are quite a few new generic parts in this PR.

The default case was already handling them, but in a less clear way; this just tried to clean up the output a bit.

stephentoub: > Could you create an issue on the OTEL semantic conventions about image parts if it does already exist?

https://github.com/open-telemetry/semantic-conventions/issues/1556

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
stephentoub: While it's a bit antithetical for me in particular, I've not been terribly concerned about such allocation when it makes the code more convoluted; the amount of other overhead in one of these operations dwarfs a 24 byte temporary object allocation.

Once this repo updates to C# 14 and we can polyfill Stopwatch.GetElapsedTime for netstandard/netfx, changing this to use GetTimestamp and GetElapsedTime will be a fine thing to do.

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryImageGenerator.cs
stephentoub: They are, right now. I kept both conditions because the decision about whether to create the stopwatch is in a separate routine, and it felt more fragile relying on that here.

test/Libraries/Microsoft.Extensions.AI.Tests/Image/OpenTelemetryImageGeneratorTests.cs
stephentoub: > add I'll some special case code in case the part has type="image"

Yeah, I'd been thinking it'd be cool if the dashboard just rendered the url for type="image". Not sure we want to do that until it's part of the convention, though?

eiriktsarpalis (approved):

PR #6812, opened 2025-09-16
Fix mapping of MinLength/MaxLength/Length attribute mapping in nullable string properties.

### LINKED ISSUE #6676: AI: If string is nullable, data annotations maxLength turns to maxItems
### Description

`[MaxLength(n)]` is turned into `maxLength` propertly if the property is also required, but if it's nullable, it's turned into `maxItems` 

### Reproduction Steps

Given the following record and function:

```csharp
record Payload([property: MaxLength(24), Required] string Value);
static void SetPayload(Payload payload) { }
```

The schema becomes: 

```json
 {
  "type": "object",
  "properties": {
    "payload": {
      "type": "object",
      "properties": {
        "value": {
          "type": "string",
          "maxLength": 24
        }
      },
      "required": [
        "value"
      ]
    }
  },
  "required": [
    "payload"
  ]
}
```

But if the payload value is made nullable: `record Payload([property: MaxLength(24)] string? Value);`, the schema now becomes: 

```json
 {
  "type": "object",
  "properties": {
    "payload": {
      "type": "object",
      "properties": {
        "value": {
          "type": [
            "string",
            "null"
          ],
          "default": null,
          "maxItems": 24
        }
      }
    }
  },
  "required": [
    "payload"
  ]
}
```

### Expected behavior

`maxLength` is consistently emitted in both cases. AFAIK, using maxItems on a non-array type is invalid (and string isn't, although from .NET's point of view it is a char[].... so perhaps that's the source of the bug?)

### Actual behavior

Emits (invalid?) json schema with maxItems for a [string, null] type.

### Regression?

_No response_

### Known Workarounds

Use `[StringLength(..)]` instead.

### Configuration

net10 preview 6, MEAI 9.7.0

### Other information

_No response_

eiriktsarpalis:
Fix #6676
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6812)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
@@ -670,22 +670,22 @@ public static partial class AIJsonUtilitiesTests
                             "string",
                             "null"
                         ],
-                        "minItems": 5
+                        "minLength": 5
                     },
                     "MaxLengthProp": {
                         "type": [
                             "string",
                             "null"
                         ],
-                        "maxItems": 50
+                        "maxLength": 50
                     },
                     "LengthProp": {
                         "type": [
                             "string",
                             "null"
                         ],
-                        "minItems": 3,
-                        "maxItems": 10
+                        "minLength": 3,
+                        "maxLength": 10
                     },
                     "MinLengthArrayProp": {
                         "type": [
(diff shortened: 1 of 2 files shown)


test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Utilities/AIJsonUtilitiesTests.cs
```diff
@@ -670,22 +670,22 @@ public static void CreateJsonSchema_IncorporatesTypesAndAnnotations_Net()
                             "string",
                             "null"
                         ],
-                        "minItems": 5
+                        "minLength": 5
```
eiriktsarpalis, line 673: @stephentoub I notice that the reported behavior already has test coverage. Was this intentional, or is it just that the output of the schema generator was simply copied over?

stephentoub: Not intentional. Just copied over.

stephentoub:
Thanks

Issue #6811, opened 2025-09-11
MEAI chat web template with --aspire flag: Class name mismatch in AppHost Program.cs causes build error


ViveliDuCh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When using `dotnet new aichatweb --provider [azureopenai,openai,ollama,githubmodels] --aspire -o my.aichatweb.project`, if the project name or a related template class contains a dot (e.g. `some.name`), the generator creates a normalized class name with underscores (`some_name`) in the `Projects` namespace. However, in the generated `Program.cs` within the AppHost project, the original (non-normalized) name (`some.name`) is used in:

```csharp
var webApp = builder.AddProject<Projects.some.name>("aichatweb-app");
```

This causes a build error because `Projects.some.name` does not exist, only `Projects.some_name` does. The issue appears with any of the providers and requires manual editing of `Program.cs` to use the normalized name. This should be handled automatically by the template or generator logic.

### Expected Behavior

The generated `Program.cs` should use the normalized class name, e.g. `Projects.some_name`, to match the generated class and avoid build errors.

### Steps To Reproduce

1. Run `dotnet new aichatweb --provider openai --aspire -o my.aichatweb.project`
2. Use a project name or template parameter that contains a dot (e.g. `some.name`)
3. Configure the AI Model Provider
4. Try to build the solution
5. Observe the CS0234 build error referencing a missing type or namespace name `some.name` in the namespace 'Projects'
6. Manually change to `Projects.some_name` in `some.name.AppHost/Program.cs` to fix the build

### Exceptions (if any)

Build error: `error CS0234: The type or namespace name 'some' does not exist in the namespace 'Projects' (are you missing an assembly reference?)`

### .NET Version info

```
dotnet --info
.NET SDK:
 Version:           9.0.304
 Commit:            f12f5f689e
 Workload version:  9.0.300-manifests.75243b89
 MSBuild version:   17.14.16+5d8159c5f

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26100
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.304\
```

### Anything else?

- ASP.NET Core version: 
  Microsoft.AspNetCore.App 8.0.19 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #6818, opened 2025-09-17
Add project name normalization to match aspire's code generator logic

ViveliDuCh:
Fixes https://github.com/dotnet/extensions/issues/6811

Apply project name transformation in the templates' generated code using the same normalization logic as in `Aspire.Hosting.AppHost.targets`. This ensures that when the `--aspire` flag is used, the generated `AppHost/Program.cs` references normalized class names (e.g., `some_name` instead of `some.name`), resolving build errors caused by mismatched type or namespace references. 

Add execution tests to verify the fix for templates with dots or other non-identifier characters in their names.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6818)

### DIFF
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
@@ -499,11 +499,11 @@
       "valueTransform": "vectorStoreIndexNameTransform",
       "replaces": "data-ChatWithCustomData-CSharp.Web-"
     },
-    "projectNameReplacer": {
+    "aspireClassNameReplacer": {
       "type": "derived",
       "valueSource": "name",
-      "valueTransform": "projectNameTransform_Formatting",
-      "replaces": "ChatWithCustomData_CSharp_Web"
+      "valueTransform": "aspireClassName_Formatting",
+      "replaces": "ChatWithCustomData_CSharp_Web_AspireClassName"
     },
     "webProjectNamespaceAdjuster": {
       "type": "generated",
@@ -530,23 +530,30 @@
     }
   },
   "forms": {
-    "projectNameTransform_Normalize": {
+    "aspireClassName_ReplaceInvalidChars_DigitDots": {
       "identifier": "replace",
-      "pattern": "[^a-zA-Z0-9_]",
+      "pattern": "((?<=\\.)|^)(?=\\d)",
       "replacement": "_",
-      "description": "Replace non-word chars, start before digit, after period with underscore"
+      "description": "Insert underscore before digits at start or after a dot"
     },
-    "projectNameTransform_AppendWeb": {
+    "aspireClassName_ReplaceInvalidChars_NonWords": {
+      "identifier": "replace",
+      "pattern": "\\W",
+      "replacement": "_",
+      "description": "Replace non-word characters with underscore"
+    },
+    "aspireClassName_AppendWeb": {
       "identifier": "replace",
       "pattern": "^(.*)$",
       "replacement": "$1_Web",
       "description": "Append _Web to the project name"
     },
-    "projectNameTransform_Formatting": {
+    "aspireClassName_Formatting": {
       "identifier": "chain",
       "steps": [
-        "projectNameTransform_Normalize",
-        "projectNameTransform_AppendWeb"
+        "aspireClassName_ReplaceInvalidChars_DigitDots",
+        "aspireClassName_ReplaceInvalidChars_NonWords",
+        "aspireClassName_AppendWeb"
       ],
       "description": "Normalize and add _Web suffix"
     },
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
@@ -502,7 +502,7 @@
     "aspireClassNameReplacer": {
       "type": "derived",
       "valueSource": "name",
-      "valueTransform": "aspireClassName_Formatting",
+      "valueTransform": "aspireClassName_ReplaceInvalidChars",
       "replaces": "ChatWithCustomData_CSharp_Web_AspireClassName"
     },
     "webProjectNamespaceAdjuster": {
@@ -530,32 +530,11 @@
     }
   },
   "forms": {
-    "aspireClassName_ReplaceInvalidChars_DigitDots": {
+    "aspireClassName_ReplaceInvalidChars": {
       "identifier": "replace",
-      "pattern": "((?<=\\.)|^)(?=\\d)",
+      "pattern": "(((?<=\\.)|^)(?=\\d)|\\W)",
       "replacement": "_",
-      "description": "Insert underscore before digits at start or after a dot"
-    },
-    "aspireClassName_ReplaceInvalidChars_NonWords": {
-      "identifier": "replace",
-      "pattern": "\\W",
-      "replacement": "_",
-      "description": "Replace non-word characters with underscore"
-    },
-    "aspireClassName_AppendWeb": {
-      "identifier": "replace",
-      "pattern": "^(.*)$",
-      "replacement": "$1_Web",
-      "description": "Append _Web to the project name"
-    },
-    "aspireClassName_Formatting": {
-      "identifier": "chain",
-      "steps": [
-        "aspireClassName_ReplaceInvalidChars_DigitDots",
-        "aspireClassName_ReplaceInvalidChars_NonWords",
-        "aspireClassName_AppendWeb"
-      ],
-      "description": "Normalize and add _Web suffix"
+      "description": "Insert underscore before digits at start, or after a dot, or to replace non-word characters"
     },
     "vectorStoreIndexNameTransform": {
       "identifier": "chain",
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
@@ -499,6 +499,12 @@
       "valueTransform": "vectorStoreIndexNameTransform",
       "replaces": "data-ChatWithCustomData-CSharp.Web-"
     },
+    "projectNameReplacer": {
+      "type": "derived",
+      "valueSource": "name",
+      "valueTransform": "projectNameTransform_Formatting",
+      "replaces": "ChatWithCustomData_CSharp_Web"
+    },
     "webProjectNamespaceAdjuster": {
       "type": "generated",
       "generator": "switch",
@@ -524,6 +530,26 @@
     }
   },
   "forms": {
+    "projectNameTransform_Normalize": {
+      "identifier": "replace",
+      "pattern": "[^a-zA-Z0-9_]",
+      "replacement": "_",
+      "description": "Replace non-word chars, start before digit, after period with underscore"
+    },
+    "projectNameTransform_AppendWeb": {
+      "identifier": "replace",
+      "pattern": "^(.*)$",
+      "replacement": "$1_Web",
+      "description": "Append _Web to the project name"
+    },
+    "projectNameTransform_Formatting": {
+      "identifier": "chain",
+      "steps": [
+        "projectNameTransform_Normalize",
+        "projectNameTransform_AppendWeb"
+      ],
+      "description": "Normalize and add _Web suffix"
+    },
     "vectorStoreIndexNameTransform": {
       "identifier": "chain",
       "steps": [
--- a/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
+++ b/test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
@@ -71,6 +71,29 @@ public class AIChatWebExecutionTests : TemplateExecutionTestBase<AIChatWebExecut
         await Fixture.BuildProjectAsync(project);
     }
 
+    /// <summary>
+    /// Tests build for a project name containing a dot ('.'), which triggers the class name normalization bug described in issue #6811.
+    /// This runs for all option combinations with --aspire true.
+    /// </summary>
+    [Theory]
+    [MemberData(nameof(GetAspireTemplateOptions))]
+    public async Task CreateRestoreAndBuild_ProjectNameWithDot(params string[] args)
+    {
+        // Use a project name with a dot to trigger the bug.
+        const string ProjectNameWithDot = "dot.name";
+        var project = await Fixture.CreateProjectAsync(
+            templateName: "aichatweb",
+            projectName: ProjectNameWithDot,
+            args);
+
+        project.StartupProjectRelativePath = $"{ProjectNameWithDot}.AppHost";
+
+        await Fixture.RestoreProjectAsync(project);
+
+        // Just attempt to build; if the bug is present, this will fail with CS0234.
+        await Fixture.BuildProjectAsync(project);
+    }
+
     private static readonly (string name, string[] values)[] _templateOptions = [
         ("--provider",          ["azureopenai", "githubmodels", "ollama", "openai"]),
         ("--vector-store",      ["azureaisearch", "local", "qdrant"]),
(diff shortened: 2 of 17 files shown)


test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -71,6 +71,29 @@ public async Task CreateRestoreAndBuild_AspireTemplate(params string[] args)
         await Fixture.BuildProjectAsync(project);
     }
 
+    /// <summary>
+    /// Tests build for a project name containing a dot ('.'), which triggers the class name normalization bug described in issue #6811.
+    /// This runs for all option combinations with --aspire true.
```
jeffhandley:
```suggestion
    /// Tests build for a project name containing a dot ('.'), which triggers the class name normalization bug described in
    /// https://github.com/dotnet/extensions/issues/6811. This runs for all option combinations with --aspire true.
```

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -71,6 +71,29 @@ public async Task CreateRestoreAndBuild_AspireTemplate(params string[] args)
...
+    /// </summary>
+    [Theory]
+    [MemberData(nameof(GetAspireTemplateOptions))]
+    public async Task CreateRestoreAndBuild_ProjectNameWithDot(params string[] args)
```
jeffhandley: A single `[Fact]` with one of the Aspire scenarios should be sufficient, instead of running for every scenario.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -524,6 +530,26 @@
     }
   },
   "forms": {
+    "projectNameTransform_Normalize": {
+      "identifier": "replace",
+      "pattern": "[^a-zA-Z0-9_]",
+      "replacement": "_",
+      "description": "Replace non-word chars, start before digit, after period with underscore"
```
jozkee: > start before digit, after period

Can you please elaborate on what this means?

ViveliDuCh: Thanks for catching that! I got it wrong. The description is supposed to reflect the regex logic, which is actually replacing non-letter, non-digit, non-underscore chars with underscore. I’ll update it right away!

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -524,6 +530,26 @@
     }
   },
   "forms": {
+    "projectNameTransform_Normalize": {
+      "identifier": "replace",
+      "pattern": "[^a-zA-Z0-9_]",
```
jeffhandley: Why is this pattern different from the pattern that the Aspire code generator uses in their `_GeneratedClassNameFixupRegex`? If our pattern matches differently from theirs in any scenario, different flavors of this bug will remain.

https://github.com/dotnet/aspire/blob/d651991486f3c2fb81ea6ea7dead2fbe98260de1/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.in.targets#L8

If this pattern is a simplified version of their pattern that works completely, then the description below should be expanded to explain why our pattern is different from theirs. Imagine their pattern being updated in the future for some reason; we'd need to know how to reapply our fix here to match their updated logic.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -499,6 +499,12 @@
       "valueTransform": "vectorStoreIndexNameTransform",
       "replaces": "data-ChatWithCustomData-CSharp.Web-"
     },
+    "projectNameReplacer": {
+      "type": "derived",
+      "valueSource": "name",
+      "valueTransform": "projectNameTransform_Formatting",
+      "replaces": "ChatWithCustomData_CSharp_Web"
```
jeffhandley: I suggest updating the template to use a different (Aspire-specific) token here, similar to how we have a `data-ChatWithCustomData-CSharp.Web-` token above, so that it's less likely we'll introduce a false-positive match on this token in the future where this replacer shouldn't apply.

```suggestion
      "replaces": "ChatWithCustomData_CSharp_Web_AspireClassName"
```

(And then use that token in the template source file as well).

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -524,6 +530,26 @@
...
+      "description": "Append _Web to the project name"
+    },
+    "projectNameTransform_Formatting": {
+      "identifier": "chain",
+      "steps": [
+        "projectNameTransform_Normalize",
+        "projectNameTransform_AppendWeb"
+      ],
+      "description": "Normalize and add _Web suffix"
+    },
```
jeffhandley: I recommend naming these to call out they are specific to the Aspire scenario.

* `aspireClassName_ReplaceInvalidChars`
* `aspireClassName_AppendWeb`
* `aspireClassName_Formatting`

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
ViveliDuCh: Great catch! Using the same regex in one step wasn't working for getting the same transformation but, after checking it again, I was indeed leaving some cases out with the reduction.  I now split aspires regex logic in 2 steps instead and it works completely. Thanks!

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
ViveliDuCh: Thanks for the suggestion! Would it make sense to test it for at least the 2 template variations of `AppHost/pogram.cs` depending on the provider chosen (`Ollama` vs `OpenAI`/`AzureOpenAi`/`GithubModels`)? Or do you think that might be redundant in this context? Happy to add them if it helps clarify coverage.

jeffhandley: Good point; you're right that testing for each scenario where the class name appears in the template would be best. That would leave it as a `[Theory]` but you could consider just using `[InlineData]` for the scenarios that apply or define a new `MemberData` member if preferred.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
jeffhandley: Terrific. The test should cover each element of logic from the regex by using a name that matches each class of character the regex does a replace on while also verifying things that should not be matched/replaced. So any of these cases you found during re-test can be added into that name used in the unit test.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -71,6 +71,44 @@ public async Task CreateRestoreAndBuild_AspireTemplate(params string[] args)
...
+    {
+        string[] projectNames = new[]
+        {
+            "dot.name",
+            "space name",
+            "mix.ed-dash_name 123",
+            ".1My.Projec-",
+            "1Project123",
+            "project.123"
+        };
```
jozkee: You should create a new MemberData method that combines `projectNames` with providers.
```cs
public static IEnumerable<object[]> GetProjectNameVariants()
{
    foreach (string provider in new[] { "Ollama", "OpenAI", "AzureOpenAI", "GitHubModels" })
    {
        foreach (string projectName in new[]
        {
            "dot.name",
            "space name",
            "mix.ed-dash_name 123",
            ".1My.Projec-",
            "1Project123",
            "project.123"
        })
        {
            yield return new object[] { provider, projectName };
        }
    }
}

[Theory]
[MemberData(nameof(GetProjectNameVariants))]
public async Task CreateRestoreAndBuild_ProjectNameVariants(string provider, string projectName)
{
    // ...
}
```

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -71,6 +71,44 @@ public async Task CreateRestoreAndBuild_AspireTemplate(params string[] args)
...
+    public async Task CreateRestoreAndBuild_ProjectNameVariants(string provider)
+    {
+        string[] projectNames = new[]
+        {
+            "dot.name",
+            "space name",
+            "mix.ed-dash_name 123",
+            ".1My.Projec-",
+            "1Project123",
+            "project.123"
```
jozkee: Can you add a more comprehensive theory, e.g.
```
// leading
.1abc // dot number is replaced with two underscores + number
.abc
#abc // alt character to dot
..abc // double non-word char
##abc
.1.2abc

// mid
abc.1abc
abc.abc
abc#abc

// trailing
// trailing dots in windows are non-standard, so we test with an alt character
abc#
abc##
abc.1
abc.1.2

// single non-word character
#

// no match
abc
_a_b_c
abc1
Abc1abc
```

ViveliDuCh: Thanks for the suggestion! To keep tests efficient, I’ve combined multiple patterns (leading, mid, trailing) into single project names—this way we maintain solid coverage without adding extra restore/build overhead. Let me know if you think anything’s still missing!

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -158,4 +180,26 @@ private static IEnumerable<string[]> GetAllPossibleOptions(ReadOnlyMemory<(strin
...
+            {
+            "dot.name",
+            "project.123",
+            "space name",
+            "mix.ed-dash_name 123",
+            ".1My.Projec-",
+            "1Project123",
+            "11double",
+            "1",
+            "nomatch"
```
jozkee: Don't forget to file the issue for the cases that fail for both aspire-starter template and aichatweb. e.g. `#`, `na#me`.

jozkee (approved):

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
jozkee: Also, if these tests are adding a significant amount of runtime, please consider filing an issue for further discussion.  cc @MackinnonBuck.

jeffhandley: Good point on this, @jozkee. I had recommended we add coverage for all the scenarios, but we should dial that back given the time added to every test run. I talked with @ViveliDuCh and we agreed the risk and impact of a regression is low, and even if this bug did escape again we'd be able to resolve it without backwards-compat concerns.

So knowing we got one-off test coverage while making the fix and we're happy with the corrected behavior is most important. We can keep a single `[Fact]` that gives comprehensive coverage of the pattern for a single provider, and then we could consider keeping a `[ConditionalTheory]` that could be used for manual testing against all of the scenarios when needed.

@ViveliDuCh I'm going to send a PR to your fork/branch showing what I suggest for the conditional theory.

jeffhandley: Here's a PR into your fork/branch that introduces the ability to have an environment variable control the conditional theory, for running the test locally while having it skipped by default and in CI.

[Add an [EnvironmentVariableSkipCondition] attribute for conditional tests. Make the Aspire project name theory conditional. (ViveliDuCh/extensions#1)](https://github.com/ViveliDuCh/extensions/pull/1)

I suggest getting @MackinnonBuck's review of that PR and if you're good with it, Mackinnon, we can merge that PR which will update this PR.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -524,6 +530,33 @@
...
+    "aspireClassName_Formatting": {
+      "identifier": "chain",
+      "steps": [
+        "aspireClassName_ReplaceInvalidChars_DigitDots",
+        "aspireClassName_ReplaceInvalidChars_NonWords",
+        "aspireClassName_AppendWeb"
+      ],
+      "description": "Normalize and add _Web suffix"
+    },
```
MackinnonBuck: Just curious - have we considered using the same logic that the official [Aspire templates use](https://github.com/dotnet/aspire/blob/e038112c318e0c50c4d2a54fd6842bd00ca59989/src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json#L526-L540)? I only ask because that template logic is already battle-tested and appears to be a bit simpler.

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
```diff
@@ -71,6 +72,44 @@ public async Task CreateRestoreAndBuild_AspireTemplate(params string[] args)
...
+        var project = await Fixture.CreateProjectAsync(
+            templateName: "aichatweb",
+            projectName: projectName,
+            args: new[] { "--aspire", $"--provider={provider}" });
+
+        project.StartupProjectRelativePath = $"{projectName}.AppHost";
+
+        await Fixture.RestoreProjectAsync(project);
+        await Fixture.BuildProjectAsync(project);
+    }
```
MackinnonBuck, line 111: I like the approach here. We have a test that always runs to catch obvious breaks, and we have a more elaborate test that doesn't run in CI but can be used to debug future breaks if they're reported. Nice!

test/ProjectTemplates/Microsoft.Extensions.AI.Templates.IntegrationTests/AIChatWebExecutionTests.cs
MackinnonBuck: The new approach looks great to me 🙂

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
jeffhandley: Ah cool; I hadn't seen that Aspire had to deal with this in other templates already. It looks like we could collapse `aspireClassName_ReplaceInvalidChars_DigitDots` and `aspireClassName_ReplaceInvalidChars_NonWords` into a single `aspireClassName_ReplaceInvalidChars` pattern of `(((?<=\\.)|^)(?=\\d)|\\W)`. We'd still need to append `_Web` as an additional step, which probably keeps ours as a `"chain"` vs. the `{ "generator": "regex", ... }`.

MackinnonBuck: > We'd still need to append _Web as an additional step...

In Aspire they just add [`_Web` directly in `AppHost.cs`](https://github.com/dotnet/aspire/blob/e038112c318e0c50c4d2a54fd6842bd00ca59989/src/Aspire.ProjectTemplates/templates/aspire-starter/9.5/Aspire-StarterApplication.1.AppHost/AppHost.cs#L7). If we did the same thing, that should enable us to avoid using `"chain"`, I believe.

ViveliDuCh: > we could collapse `aspireClassName_ReplaceInvalidChars_DigitDots` and `aspireClassName_ReplaceInvalidChars_NonWords` into a single `aspireClassName_ReplaceInvalidChars` pattern of `(((?<=\\.)|^)(?=\\d)|\\W)`.

I tried to do it in just one line like here, but it wouldn't follow the order of first doing the dot-digit replacements and then the non-word character replacements, so I tried to force the order by splitting it. 

Although, for some reason, now that I tried again, it's working. I must have missed something before. I'll update it to follow the pattern suggested. 
Appending directly in AppHost/Program.cs works too. 

Thank y'all!

PR #6819, opened 2025-09-18
Use RequestOptions from chat/responses clients

stephentoub:


rogerbarreto:
Would be nice just ensuring it works with Unit Tests checking the `User-Agent` from an injected httpHandler.

PR #6821, opened 2025-09-19
Fix KeyNotFoundException on HttpRequestLatencyListener.OnEventWritten for uknown event sources

ssmelov:
## Description
The current implementation of HttpRequestLatencyListener.OnEventWritten always expects that eventSourceName should be available in the _eventToTokenMap dictionary.
Some events (e.g., EventCounters) bypasses enabled events check causing OnEventWritten to throw KeyNotFoundException.

https://github.com/dotnet/runtime/blob/ac8354e3e94f04fae2d6d942427d025395197f76/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L2187


ssmelov:

PR #6827, opened 2025-09-19
Fix a couple of issues in M.E.AI.OpenAI clients

stephentoub:
1. Setting AllowMultipleToolCalls if there aren't any tools results in failure. Fix it to only set the property if tools have been added.
2. The chat completion service fails if a participant name containing anything other than a constrained set of characters are included. Fix it with a sanitized value.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6827)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
@@ -10,6 +10,7 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
@@ -19,14 +20,16 @@ using OpenAI.Chat;
 #pragma warning disable CA1308 // Normalize strings to uppercase
 #pragma warning disable EA0011 // Consider removing unnecessary conditional access operator (?)
 #pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S2333 // Unnecessary partial
 #pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
 #pragma warning disable SA1202 // Elements should be ordered by access
+#pragma warning disable SA1203 // Constants should appear before fields
 #pragma warning disable SA1204 // Static elements should appear before instance elements
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>Represents an <see cref="IChatClient"/> for an OpenAI <see cref="OpenAIClient"/> or <see cref="ChatClient"/>.</summary>
-internal sealed class OpenAIChatClient : IChatClient
+internal sealed partial class OpenAIChatClient : IChatClient
 {
     // These delegate instances are used to call the internal overloads of CompleteChatAsync and CompleteChatStreamingAsync that accept
     // a RequestOptions. These should be replaced once a better way to pass RequestOptions is available.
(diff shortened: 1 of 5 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -19,14 +20,16 @@
 #pragma warning disable CA1308 // Normalize strings to uppercase
 #pragma warning disable EA0011 // Consider removing unnecessary conditional access operator (?)
 #pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S2333 // Unnecessary partial
```
eiriktsarpalis, line 23: Is this a false positive triggering because it can't see source gen output?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -19,14 +20,16 @@
...
 #pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
 #pragma warning disable SA1202 // Elements should be ordered by access
+#pragma warning disable SA1203 // Constants should appear before fields
```
eiriktsarpalis, line 26: Is this triggering because of the const appearing in the new method?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
stephentoub: The partial is only needed when using the source generator, and it's only used on core, so this is coming from the netstd/netfx build. My other PR about noisy analyzers downgrades this one to suggestion.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
```diff
@@ -749,11 +757,41 @@ internal static void ConvertContentParts(ChatMessageContent content, IList<AICon
...
         };
 
+    /// <summary>Sanitizes the author name to be appropriate for including as an OpenAI participant name.</summary>
+    private static string? SanitizeAuthorName(string? name)
+    {
+        if (name is not null)
+        {
+            const int MaxLength = 64;
+
+            name = InvalidAuthorNameRegex().Replace(name, string.Empty);
```
eiriktsarpalis, line 767: Maybe it's just me, but I would rather have this fail with a clear error message as opposed to surreptitiously collapsing the value to something acceptable.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
stephentoub: It's because of the const used for the regex pattern. A good example of where the analyzer leads you to make the code worse, either duplicating the pattern so that the const isn't needed, or moving the pattern far away from the code it's logically a part of.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIChatClient.cs
eiriktsarpalis: Presumably any non-ascii identifiers would collapse to an empty string here. Assuming the author string is important, I'm not sure how I feel about that.

stephentoub: That's generally not the philosophy we've taken with meai. Different providers support different things. Some only accept text, some have very strong rules about what the text is (eg anthropic doesn't like it when assistant text ends in whitespace), some don't support author names at all, etc. If we threw exceptions in all such cases, devs would need to know all the rules for all possible implementations, and nothing would work. This is even more true when orchestrating between different providers, where the output from one is piped into the other.

PR #6830, opened 2025-09-19
Add AITool.GetService

stephentoub:
Following the same pattern as elsewhere in M.E.AI, this enables a consumer to reach through layers of delegating tools to grab information from inner ones, such as whether they were marked as requiring approval.

cc: @westey-m, @rogerbarreto 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6830)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/DelegatingAIFunction.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/DelegatingAIFunction.cs
@@ -58,4 +58,14 @@ public class DelegatingAIFunction : AIFunction
     /// <inheritdoc />
     protected override ValueTask<object?> InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken) =>
         InnerFunction.InvokeAsync(arguments, cancellationToken);
+
+    /// <inheritdoc />
+    public override object? GetService(Type serviceType, object? serviceKey = null)
+    {
+        _ = Throw.IfNull(serviceType);
+
+        return
+            serviceKey is null && serviceType.IsInstanceOfType(this) ? this :
+            InnerFunction.GetService(serviceType, serviceKey);
+    }
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/AITool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/AITool.cs
@@ -1,10 +1,12 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
 using Microsoft.Shared.Collections;
+using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
 
@@ -30,6 +32,35 @@ public abstract class AITool
 
     /// <inheritdoc/>
     public override string ToString() => Name;
+
+    /// <summary>Asks the <see cref="AITool"/> for an object of the specified type <paramref name="serviceType"/>.</summary>
+    /// <param name="serviceType">The type of object being requested.</param>
+    /// <param name="serviceKey">An optional key that can be used to help identify the target service.</param>
+    /// <returns>The found object, otherwise <see langword="null"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="serviceType"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// The purpose of this method is to allow for the retrieval of strongly-typed services that might be provided by the <see cref="AITool"/>,
+    /// including itself or any services it might be wrapping.
+    /// </remarks>
+    public virtual object? GetService(Type serviceType, object? serviceKey = null)
+    {
+        _ = Throw.IfNull(serviceType);
+
+        return
+            serviceKey is null && serviceType.IsInstanceOfType(this) ? this :
+            null;
+    }
+
+    /// <summary>Asks the <see cref="AITool"/> for an object of type <typeparamref name="TService"/>.</summary>
+    /// <typeparam name="TService">The type of the object to be retrieved.</typeparam>
+    /// <param name="serviceKey">An optional key that can be used to help identify the target service.</param>
+    /// <returns>The found object, otherwise <see langword="null"/>.</returns>
+    /// <remarks>
+    /// The purpose of this method is to allow for the retrieval of strongly typed services that may be provided by the <see cref="AITool"/>,
+    /// including itself or any services it might be wrapping.
+    /// </remarks>
+    public TService? GetService<TService>(object? serviceKey = null) =>
+        GetService(typeof(TService), serviceKey) is TService service ? service : default;
 
     /// <summary>Gets the string to display in the debugger for this instance.</summary>
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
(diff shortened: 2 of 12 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/AITool.cs
```diff
@@ -31,6 +33,35 @@ protected AITool()
...
+    /// <summary>Asks the <see cref="AITool"/> for an object of type <typeparamref name="TService"/>.</summary>
+    /// <typeparam name="TService">The type of the object to be retrieved.</typeparam>
+    /// <param name="serviceKey">An optional key that can be used to help identify the target service.</param>
+    /// <returns>The found object, otherwise <see langword="null"/>.</returns>
+    /// <remarks>
+    /// The purpose of this method is to allow for the retrieval of strongly typed services that may be provided by the <see cref="AITool"/>,
+    /// including itself or any services it might be wrapping.
+    /// </remarks>
+    public TService? GetService<TService>(object? serviceKey = null) =>
+        GetService(typeof(TService), serviceKey) is TService service ? service : default;
```
rogerbarreto, line 63: Small observation.

In `IChatClient` this is exposed as an extension and not a concrete implementation, should follow the same here for consistency?

See:

https://github.com/dotnet/extensions/blob/53ef1158f9f42632e111d6873a8cd72b803b4ae6/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs#L25

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Functions/DelegatingAIFunctionTests.cs
```diff
@@ -32,6 +32,7 @@ public void DefaultOverrides_DelegateToInnerFunction()
         Assert.Same(expected.UnderlyingMethod, actual.UnderlyingMethod);
         Assert.Same(expected.AdditionalProperties, actual.AdditionalProperties);
         Assert.Equal(expected.ToString(), actual.ToString());
+        Assert.Same(expected, actual.GetService<ApprovalRequiredAIFunction>());
```
rogerbarreto, line 35: Another observation, what should be the expected behavior when the current function is a `ApprovalRequiredAIFunction` wrapping another `ApprovalRequiredFunction`, for those scenarios we might not be returning the Inner, but the actual instance.

Might worth having a dedicated `GetService` signature that prioritizes Inner matches instead of itself? Or just let it be as we won't be very supportive of returning `Up` in the stack in case multiple wraps happen with the same type.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Tools/AIToolTests.cs
```diff
@@ -17,5 +18,25 @@ public void Constructor_Roundtrips()
...
+    [Fact]
+    public void GetService_ReturnsExpectedObject()
+    {
+        DerivedAITool tool = new();
+
+        Assert.Throws<ArgumentNullException>("serviceType", () => tool.GetService(null!));
+
+        Assert.Same(tool, tool.GetService(typeof(object)));
+        Assert.Same(tool, tool.GetService(typeof(AITool)));
+        Assert.Same(tool, tool.GetService(typeof(DerivedAITool)));
```
westey-m, line 30: Consider adding the null case where GetService does not match anything.

westey-m (approved):

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/AITool.cs
stephentoub: We have to do it that way for an interface; otherwise, on netfx/netstd where DIMs aren't supported, all interface implementations would need to implement it. With classes, we don't have that issue. And from a consumption perspective, there's little difference when both are in the same assembly. Better to keep them as real overloads when possible.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Functions/DelegatingAIFunctionTests.cs
stephentoub: > what should be the expected behavior when the current function is a ApprovalRequiredAIFunction wrapping another ApprovalRequiredFunction

Once GetService exists, when would that realistically happen? My assumption is when it does after that, the dev actually wants their wrapper recognized and not skipped, though in practice even if they didn't, I'm not sure what it would impact.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/DelegatingAIFunction.cs
```diff
@@ -58,4 +58,14 @@ protected DelegatingAIFunction(AIFunction innerFunction)
     /// <inheritdoc />
     protected override ValueTask<object?> InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken) =>
         InnerFunction.InvokeAsync(arguments, cancellationToken);
+
+    /// <inheritdoc />
+    public override object? GetService(Type serviceType, object? serviceKey = null)
```
eiriktsarpalis, line 63: My only misgiving about this is that `AITool` tends to be a service-agnostic abstraction which is supposed to be interoperable with any `IChatClient` implementation it is being fed to. Adding such a method here could give the false impression that its output might be equivalent with whatever service it being used with.

stephentoub: Thanks. Could you give an example? I'm not totally following.

eiriktsarpalis: A chat client middleware that handles a `ChatOptions`instance might reasonably assume that its tools are tied to the underlying chat client implementation, and therefore calling `GetService` on a tool would provide correponding resources. However, given the way that, say, `AIFunctionFactory` works this is almost never the case.

stephentoub: Maybe I haven't had enough coffee / it's still too early, but I'm not getting it. Can you write out the pseudo-code to help me?

eiriktsarpalis: It's a high-level concern really. AITool instances are created independently of IChatClient instances. Therefore, in most cases their implementation of `GetService` will not align with what `IChatClient.GetService` returns.

stephentoub: Do you mean in terms of mapping the abstraction to the concrete implementation for a given IChatClient? Like in an implementation of IChatClient, it calling AITool.GetService and somehow expecting to get its own representation of that tool out?

rogerbarreto: > It's a high-level concern really. AITool instances are created independently of IChatClient instances. Therefore, in most cases their implementation of GetService will not align with what IChatClient.GetService returns.

I think that's fine, if we have a CustomAITool that has internal components, this method can become handy and return one of it's internal components as well as any wrapped function (if that applies). 

The only concerning issue would be if `ApprovalRequiredFunction` was not sealed and someone would `override GetService()` behavior potentially breaking the FICC, as long our internals rely in a `GetService<>` behavior that we know and cannot change, I would think this is not concerning.

eiriktsarpalis (approved):

PR #6837, opened 2025-09-23
Disable really noisy analyzers, part 1

stephentoub:
These are super annoying and add basically zero value.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6837)

### DIFF
--- a/src/Libraries/.editorconfig
+++ b/src/Libraries/.editorconfig
@@ -5648,7 +5648,7 @@ dotnet_diagnostic.SA1202.severity = suggestion
 # Title    : Constants should appear before fields
 # Category : StyleCop.CSharp.OrderingRules
 # Help Link: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1203.md
-dotnet_diagnostic.SA1203.severity = warning
+dotnet_diagnostic.SA1203.severity = suggestion
 
 # Title    : Static elements should appear before instance elements
 # Category : StyleCop.CSharp.OrderingRules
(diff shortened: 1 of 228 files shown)


src/Libraries/.editorconfig
```diff
@@ -5201,7 +5196,7 @@ dotnet_diagnostic.S881.severity = suggestion
 # Title    : "goto" statement should not be used
 # Category : Major Code Smell
 # Help Link: https://rules.sonarsource.com/csharp/RSPEC-907
-dotnet_diagnostic.S907.severity = warning
+dotnet_diagnostic.S907.severity = suggestion
```
evgenyfedorov2, line 5199: Why is this one annoying?

src/Libraries/.editorconfig
```diff
@@ -1783,7 +1778,7 @@ dotnet_diagnostic.EA0001.severity = warning
 # Title    : Use 'System.TimeProvider' to make the code easier to test
 # Category : Reliability
 # Help Link: https://aka.ms/dotnet-extensions-warnings/EA0002
-dotnet_diagnostic.EA0002.severity = warning
+dotnet_diagnostic.EA0002.severity = suggestion
```
evgenyfedorov2, line 1781: This is one of useful warnings nudging us to always prefer TimeProvider. Don't you think so?

src/Libraries/.editorconfig
stephentoub: Because gotos are common in switch statements as a way to continue from one case to another and it flags those.

src/Libraries/.editorconfig
stephentoub: No, I don't. The rule assumes you're writing a certain kind of code and that it's always important to be able to inject a time. That's fine for a certain class of library, but it's definitely not a warning in many scenarios.

evgenyfedorov2 (approved):

peterwald (approved):

Issue #6331, opened 2025-04-24
AI Chat Template - Ollama resiliency modification should be in web app, not service defaults


jongalloway:
### Description

The latest AI Chat Template includes a modification for the resiliency when Ollama is selected, since Ollama calls can take a while (especially without good GPU support) and resiliency default is 10 seconds. This modification shouldn't be in the *Extensions.cs* in *ServiceDefaults*, since this will modify the *HttpClient* for all projects in the solution. It should just be in the Web project *Program.cs*.

### Reproduction Steps

Create a new project with Ollama selected as the LLM.

### Expected behavior

- *ServiceDefaults* *Extensions.cs* is the exact same as a new .NET Aspire project.
- *Web* project *Program.cs* includes the call to *RemoveAllResilienceHandlers* and add in standard resilience handler with extended timeouts. 

### Actual behavior

ServiceDefaults Extensions.cs is modified to affect all HttpClients in the entire solution.

### Regression?

No, this was a new addition.

### Known Workarounds

Delete code from *ServiceDefaults* and add it to *Web* project.

PR #6850, opened 2025-09-25
Scope Ollama resilience settings to Web/Program.cs and restore ServiceDefaults

ViveliDuCh:
Fixes https://github.com/dotnet/extensions/issues/6331

Restore standard global HTTP client resilience configuration in `ServiceDefaults/Program.cs`, removing Ollama-specific timeouts and settings. Apply extended timeouts and custom resilience handlers only to `chat_httpClient` and `embeddings_httpClient` in `Web/Program.cs` using `RemoveAllResilienceHandlers` and `AddStandardResilienceHandler`. Scope Ollama configuration to the web project to prevent unintended global changes and align with .NET Aspire template conventions.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6850)

### DIFF
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
@@ -60,8 +60,15 @@
           "ChatWithCustomData-CSharp.AppHost/**",
           "ChatWithCustomData-CSharp.ServiceDefaults/**",
           "ChatWithCustomData-CSharp.Web/Program.Aspire.cs",
+          "ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs",
           "README.Aspire.md",
           "*.sln"
+        ]
+      },
+      {
+        "condition": "(IsAspire && !IsOllama)",
+        "exclude": [
+          "ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs"
         ]
       },
       {
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
@@ -60,15 +60,14 @@
           "ChatWithCustomData-CSharp.AppHost/**",
           "ChatWithCustomData-CSharp.ServiceDefaults/**",
           "ChatWithCustomData-CSharp.Web/Program.Aspire.cs",
-          "ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs",
           "README.Aspire.md",
           "*.sln"
         ]
       },
       {
-        "condition": "(IsAspire && !IsOllama)",
+        "condition": "(!IsAspire || !IsOllama)",
         "exclude": [
-          "ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs"
+          "ChatWithCustomData-CSharp.Web/OllamaResilienceHandlerExtensions.cs"
         ]
       },
       {
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace ChatWithCustomData_CSharp.Web.Services;
+
+public static class AspireOllamaResilienceHandlerExtensions
+{
+    public static IServiceCollection AddOllamaHttpClientResilience(this IServiceCollection services)
+    {
+        services.ConfigureHttpClientDefaults(http =>
+        {
+#pragma warning disable EXTEXP0001 // RemoveAllResilienceHandlers is experimental
+            http.RemoveAllResilienceHandlers();
+#pragma warning restore EXTEXP0001
+
+            // Turn on resilience by default
+            http.AddStandardResilienceHandler(config =>
+            {
+                // Extend the HTTP Client timeout for Ollama
+                config.AttemptTimeout.Timeout = TimeSpan.FromMinutes(3);
+
+                // Must be at least double the AttemptTimeout to pass options validation
+                config.CircuitBreaker.SamplingDuration = TimeSpan.FromMinutes(10);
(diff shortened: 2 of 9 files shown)


src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/Program.Aspire.cs
```diff
@@ -50,6 +50,38 @@
 #endif
 builder.Services.AddScoped<DataIngestor>();
 builder.Services.AddSingleton<SemanticSearch>();
+#if (IsOllama)
+// Get the IHttpClientBuilder for chat_httpClient
+var chatClientBuilder = builder.Services.AddHttpClient("chat_httpClient");
```
jozkee: how are these named http clients used by ollama?

jozkee: As we discussed offline, this is not ideal as `chat_httpClient` is an [implementation detail](https://github.com/CommunityToolkit/Aspire/blob/cf7da1aa8a380cfcd337d2dc56b13f750cfbf856/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs#L168), we shouldn't depend on it.

MackinnonBuck: Agreed, it would be nice to not rely on the internal behavior of another library like this.

That said, I don't currently see another option. We need a way to get the `IHttpClientBuilder`. This might be a feature request for https://github.com/CommunityToolkit/Aspire.

jozkee: > I don't currently see another option

I think the intent of https://github.com/dotnet/extensions/issues/6331 was to just move the following to Web/Program.cs https://github.com/dotnet/extensions/blob/c4e57fb1e6b8403a527ea3cd737f1146dcbc1f31/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.ServiceDefaults/Extensions.cs#L33-L42

And remove that `#if (IsOllama)` block, @MackinnonBuck, wouldn't that suffice?  As per @jongalloway, that would narrow it down to configure only the clients in Web.csproj.

We would have two `ConfigureHttpClientDefaults` calls, though, not sure if that's OK.

jozkee: Additionally, we tried to narrow it down further checking the IHttpClientBuilder names:
```cs
builder.Services.ConfigureHttpClientDefaults(http =>
{
    if (http.Name.Contains("chat") || http.Name.Contains("embeddings"))
    {
#pragma warning disable EXTEXP0001 // RemoveAllResilienceHandlers is experimental
        http.RemoveAllResilienceHandlers();
#pragma warning restore EXTEXP0001

        // Turn on resilience by default
        http.AddStandardResilienceHandler(config =>
        {
            // Extend the HTTP Client timeout for Ollama
            config.AttemptTimeout.Timeout = TimeSpan.FromMinutes(3);

            // Must be at least double the AttemptTimeout to pass options validation
            config.CircuitBreaker.SamplingDuration = TimeSpan.FromMinutes(10);
            config.TotalRequestTimeout.Timeout = TimeSpan.FromMinutes(10);
        });

        // Turn on service discovery by default
        http.AddServiceDiscovery();
    }
});
```
But this didn't work, we only observed builders with null names.  We would need to dig deeper to understand what's happening, just letting you know in case it rings a bell.

jongalloway: Agree, my main request is that we scope to the web project as opposed to the entire Aspire solution. Could also add a comment explaining the behavior and recommending removing the block of code before deploying to production (since Ollama won't be used in production).

MackinnonBuck: Ah, got it. Thanks for the clarification.

My main suggestion would be to consider putting this logic somewhere other than `Program.cs` - maybe another file defining an additional extension method that gets called from `Program.cs`.

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/.template.config/template.json
```diff
@@ -60,10 +60,17 @@
...
           "README.Aspire.md",
           "*.sln"
         ]
       },
+      {
+        "condition": "(IsAspire && !IsOllama)",
+        "exclude": [
+          "ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs"
+        ]
+      },
```
jeffhandley: Rather than having this file excluded from 2 different blocks, I recommend removing it from the first list and changing the second condition to be `(!IsAspire || !IsOllama)`.


```suggestion
          "README.Aspire.md",
          "*.sln"
]
},
      {
        "condition": "(!IsAspire || !IsOllama)",
        "exclude": [
          "ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs"
        ]
      },
```

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/OllamaResilienceHandlerExtensions.cs
jeffhandley, line 1: I recommend a file name of `OllamaResilienceHandlerExtensions.cs` (removing Aspire).

src/ProjectTemplates/Microsoft.Extensions.AI.Templates/src/ChatWithCustomData/ChatWithCustomData-CSharp.Web/AspireOllamaResilienceHandlerExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace ChatWithCustomData_CSharp.Web.Services;
+
+public static class AspireOllamaResilienceHandlerExtensions
+{
+    public static IServiceCollection AddOllamaHttpClientResilience(this IServiceCollection services)
```
jeffhandley: I recommend a method name of `AddOllamaResilienceHandler` for alignment with `AddStandardResilienceHandler`. "Http" is immaterial.

PR #6853, opened 2025-09-25
Fix chat template `Microsoft.Extensions.Http.Resilience` dependency

MackinnonBuck:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6853)

### DIFF
--- a/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/Microsoft.Extensions.AI.Templates.csproj
+++ b/src/ProjectTemplates/Microsoft.Extensions.AI.Templates/Microsoft.Extensions.AI.Templates.csproj
@@ -32,15 +32,6 @@
       Include="..\GenerateTemplateContent\GenerateTemplateContent.csproj"
       ReferenceOutputAssembly="false"
       PrivateAssets="all" />
-
-    <!--
-      Internal packages referenced by the templates that aren't in the AI workstream.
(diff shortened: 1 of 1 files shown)


MackinnonBuck:
> Assuming this is the only package reference to other non-MEAI packages from extensions right?

Yep - this was the only non-MEAI package whose version was inferred automatically.

src/ProjectTemplates/GeneratedContent.targets
```diff
@@ -20,11 +20,14 @@
       "TemplatePackageVersion_{PackageName}"
       where {PackageName} is the package ID with '.' characters removed.
       The value of each property will be the computed package version.
+
+      IMPORTANT: Internal packages that ship at a different cadence than these project templates should
+                 be referenced using explicit package versions instead (see the "ComputeGeneratedContentProperties"
+                 target below).
```
jeffhandley, line 26: :+1:

jeffhandley (approved):

MackinnonBuck:
The project templates also had an explicit project reference to `Microsoft.Extensions.Http.Resilience` as a way to ensure that the internal dependency was built before template tests were run. I've removed this project reference now that we're depending on a version available in nuget.org.

PR #6854, opened 2025-09-25
add support for background responses

SergeyMenshykh:
This PR adds a model for supporting background responses (long-running operations), updates the chat completion model to use it, and integrates this functionality into `OpenAIResponsesChatClient`.

Background responses use a continuation token returned by the `GetResponseAsync` and `GetStreamingResponseAsync` methods in the `ChatResponse` and `ChatResponseUpdate` classes, respectively, when background responses are enabled and supported by the chat client.

The continuation token contains all necessary details to enable polling for a background response using the non-streaming `GetResponseAsync` method. It also allows resuming a streamed background response with the `GetStreamingResponseAsync` method if the stream is interrupted. In both cases, the continuation token obtained from the initial chat result or the streamed update received before the interruption should be supplied as input for follow-up calls.

When a background response has completed, failed, or cannot proceed further (for example, when user input is required), the continuation token returned by either method will be null, signaling to consumers that processing is complete and there is nothing to poll or resume. The result returned by either method can then be used for further processing.

#### Non-streaming API:
````csharp
var chatOptions = new ChatOptions
{
    BackgroundResponsesOptions = new() { Allow = true }
};

// Get initial response with continuation token
var response = await chatClient.GetResponseAsync("What's the biggest animal?", chatOptions);

// Continue to poll until the final response is received
while (response.ContinuationToken is not null)
{
    response = await chatClient.GetResponseAsync(response.ContinuationToken, chatOptions);
}

Console.WriteLine(response);
````

#### Streaming API:
````csharp
ChatOptions chatOptions = new()
{
    BackgroundResponsesOptions = new BackgroundResponsesOptions { Allow = true },
};

string response = "";
ResumptionToken? continuationToken = null;

await foreach (var update in chatClient.GetStreamingResponseAsync("What is the capital of France?", chatOptions))
{
    response += update;

    // Simulate an interruption
    continuationToken = update.ContinuationToken;
    break;
}

// Resume streaming from the point captured by the continuation token
await foreach (var update in chatClient.GetStreamingResponseAsync(continuationToken!, chatOptions))
{
    response += update;
}

Console.WriteLine(response);
````

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
@@ -1,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Options for configuring background responses.</summary>
+[Experimental("MEAI001")]
+public sealed class BackgroundResponsesOptions
+{
+    /// <summary>Initializes a new instance of the <see cref="BackgroundResponsesOptions"/> class.</summary>
+    public BackgroundResponsesOptions()
+    {
+    }
+
+    /// <summary>Initializes a new instance of the <see cref="BackgroundResponsesOptions"/> class.</summary>
+    /// <param name="options">The options to initialize from.</param>
+    public BackgroundResponsesOptions(BackgroundResponsesOptions options)
+    {
+        _ = Throw.IfNull(options);
+
+        Allow = options.Allow;
+    }
+
+    /// <summary>Gets or sets a value indicating whether the background responses are allowed.</summary>
+    /// <remarks>
+    /// This property only takes effect if the API it's used with supports background responses.
+    /// If the API does not support background responses, this property will be ignored.
+    /// </remarks>
+    public bool? Allow { get; set; }
+}
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
@@ -6,6 +6,11 @@
 - Added new `AITool.GetService` virtual method.
 - Updated `TextReasoningContent` to include `ProtectedData` for representing encrypted/redacted content.
 - Fixed `MinLength`/`MaxLength`/`Length` attribute mapping in nullable string properties during schema export.
+- Added `[Experimental]` `ResumptionToken` to represent a base class for continuation tokens.
+- Added `[Experimental]` `BackgroundResponsesOptions` to represent options for background responses.
+- Added `[Experimental]` `ChatOptions.BackgroundResponsesOptions` and `ChatOptions.ContinuationToken` properties specify continuation token and options for background responses.
+- Added `[Experimental]` `ChatResponse.ContinuationToken` and `ChatResponseUpdate.ContinuationToken` to return continuation token for background responses.
+- Added `[Experimental]` `GetResponseAsync` and `GetStreamingResponseAsync` extension methods for `IChatClient` to continue background responses.
 
 ## 9.9.0
 
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Shared.Diagnostics;
@@ -120,6 +121,30 @@ public static class ChatClientExtensions
         return client.GetResponseAsync([chatMessage], options, cancellationToken);
     }
 
+    /// <summary>Gets a background response identified by the specified continuation token.</summary>
+    /// <param name="client">The chat client.</param>
+    /// <param name="continuationToken">The continuation token.</param>
+    /// <param name="options">The chat options to configure the request.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The background response.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="continuationToken"/> is <see langword="null"/>.</exception>
+    [Experimental("MEAI001")]
+    public static Task<ChatResponse> GetResponseAsync(
+        this IChatClient client,
+        ResumptionToken continuationToken,
+        ChatOptions? options = null,
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(client);
+        _ = Throw.IfNull(continuationToken);
+
+        ChatOptions chatOptions = options ?? new();
+        chatOptions.ContinuationToken = continuationToken;
+
+        return client.GetResponseAsync([], chatOptions, cancellationToken);
+    }
+
     /// <summary>Sends a user chat text message and streams the response messages.</summary>
     /// <param name="client">The chat client.</param>
     /// <param name="chatMessage">The text content for the chat message to send.</param>
@@ -158,5 +183,29 @@ public static class ChatClientExtensions
         _ = Throw.IfNull(chatMessage);
 
         return client.GetStreamingResponseAsync([chatMessage], options, cancellationToken);
+    }
+
+    /// <summary>Gets a background streamed response identified by the specified continuation token and streams its messages.</summary>
+    /// <param name="client">The chat client.</param>
+    /// <param name="continuationToken">The continuation token.</param>
+    /// <param name="options">The chat options to configure the request.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The background response messages.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="continuationToken"/> is <see langword="null"/>.</exception>
+    [Experimental("MEAI001")]
+    public static IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
+        this IChatClient client,
+        ResumptionToken continuationToken,
+        ChatOptions? options = null,
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(client);
+        _ = Throw.IfNull(continuationToken);
+
+        ChatOptions chatOptions = options ?? new();
+        chatOptions.ContinuationToken = continuationToken;
+
+        return client.GetStreamingResponseAsync([], chatOptions, cancellationToken);
     }
 }
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json.Serialization;
 
 namespace Microsoft.Extensions.AI;
@@ -115,6 +116,24 @@ public class ChatOptions
     [JsonIgnore]
     public IList<AITool>? Tools { get; set; }
 
+    /// <summary>Gets or sets the options for configuring background responses.</summary>
+    [Experimental("MEAI001")]
+    public BackgroundResponsesOptions? BackgroundResponsesOptions { get; set; }
+
+    /// <summary>Gets or sets the continuation token for resuming and getting result of the chat response identified by this token.</summary>
+    /// <remarks>
+    /// This property is used for background responses that can be activated via the <see cref="BackgroundResponsesOptions.Allow"/>
+    /// property if the <see cref="IChatClient"/> implementation supports them.
+    /// Streamed background responses, such as those returned by default by <see cref="IChatClient.GetStreamingResponseAsync"/>,
+    /// can be resumed if interrupted. This means that a continuation token obtained from the <see cref="ChatResponseUpdate.ContinuationToken"/> 
+    /// of an update just before the interruption occurred can be passed to this property to resume the stream from the point of interruption.
+    /// Non-streamed background responses, such as those returned by <see cref="IChatClient.GetResponseAsync"/>,
+    /// can be polled for completion by obtaining the token from the <see cref="ChatResponse.ContinuationToken"/> property 
+    /// and passing it to this property on subsequent calls to <see cref="IChatClient.GetResponseAsync"/>.
+    /// </remarks>
+    [Experimental("MEAI001")]
+    public ResumptionToken? ContinuationToken { get; set; }
+
     /// <summary>
     /// Gets or sets a callback responsible for creating the raw representation of the chat options from an underlying implementation.
     /// </summary>
@@ -152,6 +171,7 @@ public class ChatOptions
             AdditionalProperties = AdditionalProperties?.Clone(),
             AllowMultipleToolCalls = AllowMultipleToolCalls,
             ConversationId = ConversationId,
+            ContinuationToken = ContinuationToken,
             FrequencyPenalty = FrequencyPenalty,
             Instructions = Instructions,
             MaxOutputTokens = MaxOutputTokens,
@@ -165,6 +185,11 @@ public class ChatOptions
             TopK = TopK,
             TopP = TopP,
         };
+
+        if (BackgroundResponsesOptions is { } bros)
+        {
+            options.BackgroundResponsesOptions = new BackgroundResponsesOptions(bros);
+        }
 
         if (StopSequences is not null)
         {
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
@@ -88,6 +88,20 @@ public class ChatResponse
     /// <summary>Gets or sets usage details for the chat response.</summary>
     public UsageDetails? Usage { get; set; }
 
+    /// <summary>Gets or sets the continuation token for getting result of the background chat response.</summary>
+    /// <remarks>
+    /// <see cref="IChatClient"/> implementations that support background responses will return
+    /// a continuation token if background responses are enabled in <see cref="ChatOptions.BackgroundResponsesOptions"/>
+    /// and the result of the response has not been obtained yet. Otherwise, the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used in conjunction with <see cref="ChatOptions.ContinuationToken"/> to 
+    /// continue or polling for the completion the response. Pass this token to
+    /// <see cref="ChatOptions.ContinuationToken"/> on subsequent calls to <see cref="IChatClient.GetResponseAsync"/>
+    /// to poll for completion.
+    /// </para>
+    /// </remarks>
+    public ResumptionToken? ContinuationToken { get; set; }
+
     /// <summary>Gets or sets the raw representation of the chat response from an underlying implementation.</summary>
     /// <remarks>
     /// If a <see cref="ChatResponse"/> is created to represent some underlying object from another object
@@ -143,6 +157,7 @@ public class ChatResponse
                 ResponseId = ResponseId,
 
                 CreatedAt = message.CreatedAt ?? CreatedAt,
+                ContinuationToken = ContinuationToken,
             };
         }
 
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
@@ -141,6 +141,19 @@ public class ChatResponseUpdate
     /// <inheritdoc/>
     public override string ToString() => Text;
 
+    /// <summary>Gets or sets the continuation token for resuming the streamed chat response of which this update is a part.</summary>
+    /// <remarks>
+    /// <see cref="IChatClient"/> implementations that support background responses will return
+    /// a continuation token on each update if background responses are enabled in <see cref="ChatOptions.BackgroundResponsesOptions"/>
+    /// and not all updates for the response have been streamed yet. Otherwise, the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used for stream resumption, where the continuation token of the latest received update should be
+    /// passed to <see cref="ChatOptions.ContinuationToken"/> on subsequent calls to <see cref="IChatClient.GetStreamingResponseAsync"/>
+    /// to resume streaming from the point of interruption.
+    /// </para>
+    /// </remarks>
+    public ResumptionToken? ContinuationToken { get; set; }
+
     /// <summary>Gets a <see cref="AIContent"/> object to display in the debugger display.</summary>
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     private AIContent? ContentForDebuggerDisplay
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
@@ -1,5 +1,5 @@
 {
-  "Name": "Microsoft.Extensions.AI.Abstractions, Version=9.6.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
+  "Name": "Microsoft.Extensions.AI.Abstractions, Version=9.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
   "Types": [
     {
       "Type": "sealed class Microsoft.Extensions.AI.AdditionalPropertiesDictionary : Microsoft.Extensions.AI.AdditionalPropertiesDictionary<object?>",
@@ -134,11 +134,11 @@
       ],
       "Properties": [
         {
-          "Member": "System.Collections.Generic.IList<Microsoft.Extensions.AI.AnnotatedRegion>? Microsoft.Extensions.AI.AIAnnotation.AnnotatedRegions { get; set; }",
+          "Member": "Microsoft.Extensions.AI.AdditionalPropertiesDictionary? Microsoft.Extensions.AI.AIAnnotation.AdditionalProperties { get; set; }",
           "Stage": "Stable"
         },
         {
-          "Member": "Microsoft.Extensions.AI.AdditionalPropertiesDictionary? Microsoft.Extensions.AI.AIAnnotation.AdditionalProperties { get; set; }",
+          "Member": "System.Collections.Generic.IList<Microsoft.Extensions.AI.AnnotatedRegion>? Microsoft.Extensions.AI.AIAnnotation.AnnotatedRegions { get; set; }",
           "Stage": "Stable"
         },
         {
@@ -180,15 +180,15 @@
           "Stage": "Stable"
         },
         {
+          "Member": "Microsoft.Extensions.AI.AIFunctionDeclaration Microsoft.Extensions.AI.AIFunction.AsDeclarationOnly();",
+          "Stage": "Stable"
+        },
+        {
           "Member": "System.Threading.Tasks.ValueTask<object?> Microsoft.Extensions.AI.AIFunction.InvokeAsync(Microsoft.Extensions.AI.AIFunctionArguments? arguments = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));",
           "Stage": "Stable"
         },
         {
           "Member": "abstract System.Threading.Tasks.ValueTask<object?> Microsoft.Extensions.AI.AIFunction.InvokeCoreAsync(Microsoft.Extensions.AI.AIFunctionArguments arguments, System.Threading.CancellationToken cancellationToken);",
-          "Stage": "Stable"
-        },
-        {
-          "Member": "Microsoft.Extensions.AI.AIFunctionDeclaration Microsoft.Extensions.AI.AIFunction.AsDeclarationOnly();",
           "Stage": "Stable"
         }
       ],
@@ -199,26 +199,6 @@
         },
         {
           "Member": "virtual System.Reflection.MethodInfo? Microsoft.Extensions.AI.AIFunction.UnderlyingMethod { get; }",
-          "Stage": "Stable"
-        }
-      ]
-    },
-    {
-      "Type": "abstract class Microsoft.Extensions.AI.AIFunctionDeclaration : Microsoft.Extensions.AI.AITool",
-      "Stage": "Stable",
-      "Methods": [
-        {
-          "Member": "Microsoft.Extensions.AI.AIFunctionDeclaration.AIFunctionDeclaration();",
-          "Stage": "Stable"
-        }
-      ],
-      "Properties": [
-        {
-          "Member": "virtual System.Text.Json.JsonElement Microsoft.Extensions.AI.AIFunctionDeclaration.JsonSchema { get; }",
-          "Stage": "Stable"
-        },
-        {
-          "Member": "virtual System.Text.Json.JsonElement? Microsoft.Extensions.AI.AIFunctionDeclaration.ReturnJsonSchema { get; }",
           "Stage": "Stable"
         }
       ]
@@ -295,6 +275,26 @@
         },
         {
           "Member": "System.Collections.Generic.ICollection<object?> Microsoft.Extensions.AI.AIFunctionArguments.Values { get; }",
+          "Stage": "Stable"
+        }
+      ]
+    },
+    {
+      "Type": "abstract class Microsoft.Extensions.AI.AIFunctionDeclaration : Microsoft.Extensions.AI.AITool",
+      "Stage": "Stable",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.AIFunctionDeclaration.AIFunctionDeclaration();",
+          "Stage": "Stable"
+        }
+      ],
+      "Properties": [
+        {
+          "Member": "virtual System.Text.Json.JsonElement Microsoft.Extensions.AI.AIFunctionDeclaration.JsonSchema { get; }",
+          "Stage": "Stable"
+        },
+        {
+          "Member": "virtual System.Text.Json.JsonElement? Microsoft.Extensions.AI.AIFunctionDeclaration.ReturnJsonSchema { get; }",
           "Stage": "Stable"
         }
       ]
@@ -724,6 +724,16 @@
       ]
     },
     {
+      "Type": "sealed class Microsoft.Extensions.AI.ApprovalRequiredAIFunction : Microsoft.Extensions.AI.DelegatingAIFunction",
+      "Stage": "Experimental",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.ApprovalRequiredAIFunction.ApprovalRequiredAIFunction(Microsoft.Extensions.AI.AIFunction innerFunction);",
+          "Stage": "Experimental"
+        }
+      ]
+    },
+    {
       "Type": "sealed class Microsoft.Extensions.AI.AutoChatToolMode : Microsoft.Extensions.AI.ChatToolMode",
       "Stage": "Stable",
       "Methods": [
@@ -738,6 +748,26 @@
         {
           "Member": "override int Microsoft.Extensions.AI.AutoChatToolMode.GetHashCode();",
           "Stage": "Stable"
+        }
+      ]
+    },
+    {
+      "Type": "sealed class Microsoft.Extensions.AI.BackgroundResponsesOptions",
+      "Stage": "Experimental",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.BackgroundResponsesOptions.BackgroundResponsesOptions();",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "Microsoft.Extensions.AI.BackgroundResponsesOptions.BackgroundResponsesOptions(Microsoft.Extensions.AI.BackgroundResponsesOptions options);",
+          "Stage": "Experimental"
+        }
+      ],
+      "Properties": [
+        {
+          "Member": "bool? Microsoft.Extensions.AI.BackgroundResponsesOptions.Allow { get; set; }",
+          "Stage": "Experimental"
         }
       ]
     },
@@ -800,6 +830,10 @@
           "Stage": "Stable"
         },
         {
+          "Member": "static System.Threading.Tasks.Task<Microsoft.Extensions.AI.ChatResponse> Microsoft.Extensions.AI.ChatClientExtensions.GetResponseAsync(this Microsoft.Extensions.AI.IChatClient client, Microsoft.Extensions.AI.ResumptionToken continuationToken, Microsoft.Extensions.AI.BackgroundResponsesOptions options, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));",
+          "Stage": "Experimental"
+        },
+        {
           "Member": "static TService? Microsoft.Extensions.AI.ChatClientExtensions.GetService<TService>(this Microsoft.Extensions.AI.IChatClient client, object? serviceKey = null);",
           "Stage": "Stable"
         },
@@ -810,6 +844,10 @@
         {
           "Member": "static System.Collections.Generic.IAsyncEnumerable<Microsoft.Extensions.AI.ChatResponseUpdate> Microsoft.Extensions.AI.ChatClientExtensions.GetStreamingResponseAsync(this Microsoft.Extensions.AI.IChatClient client, Microsoft.Extensions.AI.ChatMessage chatMessage, Microsoft.Extensions.AI.ChatOptions? options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));",
           "Stage": "Stable"
+        },
+        {
+          "Member": "static System.Collections.Generic.IAsyncEnumerable<Microsoft.Extensions.AI.ChatResponseUpdate> Microsoft.Extensions.AI.ChatClientExtensions.GetStreamingResponseAsync(this Microsoft.Extensions.AI.IChatClient client, Microsoft.Extensions.AI.ResumptionToken continuationToken, Microsoft.Extensions.AI.BackgroundResponsesOptions options, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));",
+          "Stage": "Experimental"
         }
       ]
     },
@@ -998,15 +1036,23 @@
           "Stage": "Stable"
         },
         {
+          "Member": "Microsoft.Extensions.AI.BackgroundResponsesOptions? Microsoft.Extensions.AI.ChatOptions.BackgroundResponsesOptions { get; set; }",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "Microsoft.Extensions.AI.ResumptionToken? Microsoft.Extensions.AI.ChatOptions.ContinuationToken { get; set; }",
+          "Stage": "Experimental"
+        },
+        {
           "Member": "string? Microsoft.Extensions.AI.ChatOptions.ConversationId { get; set; }",
           "Stage": "Stable"
         },
         {
-          "Member": "string? Microsoft.Extensions.AI.ChatOptions.Instructions { get; set; }",
+          "Member": "float? Microsoft.Extensions.AI.ChatOptions.FrequencyPenalty { get; set; }",
           "Stage": "Stable"
         },
         {
-          "Member": "float? Microsoft.Extensions.AI.ChatOptions.FrequencyPenalty { get; set; }",
+          "Member": "string? Microsoft.Extensions.AI.ChatOptions.Instructions { get; set; }",
           "Stage": "Stable"
         },
         {
@@ -1087,6 +1133,10 @@
       "Properties": [
         {
           "Member": "Microsoft.Extensions.AI.AdditionalPropertiesDictionary? Microsoft.Extensions.AI.ChatResponse.AdditionalProperties { get; set; }",
+          "Stage": "Stable"
+        },
+        {
+          "Member": "Microsoft.Extensions.AI.ResumptionToken? Microsoft.Extensions.AI.ChatResponse.ContinuationToken { get; set; }",
           "Stage": "Stable"
         },
         {
@@ -1262,6 +1312,10 @@
           "Stage": "Stable"
         },
         {
+          "Member": "Microsoft.Extensions.AI.ResumptionToken? Microsoft.Extensions.AI.ChatResponseUpdate.ContinuationToken { get; set; }",
+          "Stage": "Stable"
+        },
+        {
           "Member": "string? Microsoft.Extensions.AI.ChatResponseUpdate.ConversationId { get; set; }",
           "Stage": "Stable"
         },
@@ -1412,6 +1466,14 @@
       ],
       "Properties": [
         {
+          "Member": "string? Microsoft.Extensions.AI.CitationAnnotation.FileId { get; set; }",
+          "Stage": "Stable"
+        },
+        {
+          "Member": "string? Microsoft.Extensions.AI.CitationAnnotation.Snippet { get; set; }",
+          "Stage": "Stable"
+        },
+        {
           "Member": "string? Microsoft.Extensions.AI.CitationAnnotation.Title { get; set; }",
           "Stage": "Stable"
         },
@@ -1421,14 +1483,6 @@
         },
         {
           "Member": "System.Uri? Microsoft.Extensions.AI.CitationAnnotation.Url { get; set; }",
-          "Stage": "Stable"
-        },
-        {
-          "Member": "string? Microsoft.Extensions.AI.CitationAnnotation.FileId { get; set; }",
-          "Stage": "Stable"
-        },
-        {
-          "Member": "string? Microsoft.Extensions.AI.CitationAnnotation.Snippet { get; set; }",
           "Stage": "Stable"
         }
       ]
@@ -1464,11 +1518,11 @@
           "Stage": "Stable"
         },
         {
-          "Member": "string? Microsoft.Extensions.AI.DataContent.Name { get; set; }",
+          "Member": "string Microsoft.Extensions.AI.DataContent.MediaType { get; }",
           "Stage": "Stable"
         },
         {
-          "Member": "string Microsoft.Extensions.AI.DataContent.MediaType { get; }",
+          "Member": "string? Microsoft.Extensions.AI.DataContent.Name { get; set; }",
           "Stage": "Stable"
         },
         {
@@ -1495,20 +1549,20 @@
         },
         {
           "Member": "override string Microsoft.Extensions.AI.DelegatingAIFunction.ToString();",
-          "Stage": "Experimental"
+          "Stage": "Stable"
         }
       ],
       "Properties": [
-        {
-          "Member": "Microsoft.Extensions.AI.AIFunction Microsoft.Extensions.AI.DelegatingAIFunction.InnerFunction { get; }",
-          "Stage": "Stable"
-        },
         {
           "Member": "override System.Collections.Generic.IReadOnlyDictionary<string, object?> Microsoft.Extensions.AI.DelegatingAIFunction.AdditionalProperties { get; }",
           "Stage": "Stable"
         },
         {
           "Member": "override string Microsoft.Extensions.AI.DelegatingAIFunction.Description { get; }",
+          "Stage": "Stable"
+        },
+        {
+          "Member": "Microsoft.Extensions.AI.AIFunction Microsoft.Extensions.AI.DelegatingAIFunction.InnerFunction { get; }",
           "Stage": "Stable"
         },
         {
@@ -1598,6 +1652,38 @@
         {
           "Member": "Microsoft.Extensions.AI.IEmbeddingGenerator<TInput, TEmbedding> Microsoft.Extensions.AI.DelegatingEmbeddingGenerator<TInput, TEmbedding>.InnerGenerator { get; }",
           "Stage": "Stable"
+        }
+      ]
+    },
+    {
+      "Type": "class Microsoft.Extensions.AI.DelegatingImageGenerator : Microsoft.Extensions.AI.IImageGenerator, System.IDisposable",
+      "Stage": "Experimental",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.DelegatingImageGenerator.DelegatingImageGenerator(Microsoft.Extensions.AI.IImageGenerator innerGenerator);",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "void Microsoft.Extensions.AI.DelegatingImageGenerator.Dispose();",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "virtual void Microsoft.Extensions.AI.DelegatingImageGenerator.Dispose(bool disposing);",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "virtual System.Threading.Tasks.Task<Microsoft.Extensions.AI.ImageGenerationResponse> Microsoft.Extensions.AI.DelegatingImageGenerator.GenerateAsync(Microsoft.Extensions.AI.ImageGenerationRequest request, Microsoft.Extensions.AI.ImageGenerationOptions? options = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "virtual object? Microsoft.Extensions.AI.DelegatingImageGenerator.GetService(System.Type serviceType, object? serviceKey = null);",
+          "Stage": "Experimental"
+        }
+      ],
+      "Properties": [
+        {
+          "Member": "Microsoft.Extensions.AI.IImageGenerator Microsoft.Extensions.AI.DelegatingImageGenerator.InnerGenerator { get; }",
+          "Stage": "Experimental"
         }
       ]
     },
@@ -1800,6 +1886,46 @@
       ]
     },
     {
+      "Type": "sealed class Microsoft.Extensions.AI.FunctionApprovalRequestContent : Microsoft.Extensions.AI.UserInputRequestContent",
+      "Stage": "Experimental",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.FunctionApprovalRequestContent.FunctionApprovalRequestContent(string id, Microsoft.Extensions.AI.FunctionCallContent functionCall);",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "Microsoft.Extensions.AI.FunctionApprovalResponseContent Microsoft.Extensions.AI.FunctionApprovalRequestContent.CreateResponse(bool approved);",
+          "Stage": "Experimental"
+        }
+      ],
+      "Properties": [
+        {
+          "Member": "Microsoft.Extensions.AI.FunctionCallContent Microsoft.Extensions.AI.FunctionApprovalRequestContent.FunctionCall { get; }",
+          "Stage": "Experimental"
+        }
+      ]
+    },
+    {
+      "Type": "sealed class Microsoft.Extensions.AI.FunctionApprovalResponseContent : Microsoft.Extensions.AI.UserInputResponseContent",
+      "Stage": "Experimental",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.FunctionApprovalResponseContent.FunctionApprovalResponseContent(string id, bool approved, Microsoft.Extensions.AI.FunctionCallContent functionCall);",
+          "Stage": "Experimental"
+        }
+      ],
+      "Properties": [
+        {
+          "Member": "bool Microsoft.Extensions.AI.FunctionApprovalResponseContent.Approved { get; }",
+          "Stage": "Experimental"
+        },
+        {
+          "Member": "Microsoft.Extensions.AI.FunctionCallContent Microsoft.Extensions.AI.FunctionApprovalResponseContent.FunctionCall { get; }",
+          "Stage": "Experimental"
+        }
+      ]
+    },
+    {
       "Type": "sealed class Microsoft.Extensions.AI.FunctionCallContent : Microsoft.Extensions.AI.AIContent",
       "Stage": "Stable",
       "Methods": [
@@ -1948,6 +2074,22 @@
       ]
     },
     {
+      "Type": "sealed class Microsoft.Extensions.AI.HostedFileContent : Microsoft.Extensions.AI.AIContent",
+      "Stage": "Stable",
+      "Methods": [
+        {
+          "Member": "Microsoft.Extensions.AI.HostedFileContent.HostedFileContent(string fileId);",
+          "Stage": "Stable"
+        }
+      ],
+      "Properties": [
+        {
+          "Member": "string Microsoft.Extensions.AI.HostedFileContent.FileId { get; set; }",
+          "Stage": "Stable"
+        }
+      ]
+    },
+    {
       "Type": "class Microsoft.Extensions.AI.HostedFileSearchTool : Microsoft.Extensions.AI.AITool",
       "Stage": "Stable",
       "Methods": [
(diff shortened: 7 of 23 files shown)

### CODE REVIEW

File: src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+public sealed class BackgroundResponsesOptions
+{
+    /// <summary>Initializes a new instance of the <see cref="BackgroundResponsesOptions"/> class.</summary>
+    public BackgroundResponsesOptions()
+    {
+    }
+
+    /// <summary>Initializes a new instance of the <see cref="BackgroundResponsesOptions"/> class.</summary>
+    /// <param name="options">The options to initialize from.</param>
+    public BackgroundResponsesOptions(BackgroundResponsesOptions options)
```
stephentoub: Why is the copy ctor needed?

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        Allow = options.Allow;
+    }
+
+    /// <summary>Gets or sets a value indicating whether the background responses are allowed.</summary>
+    /// <remarks>
+    /// This property only takes effect if the API it's used with supports background responses.
+    /// If the API does not support background responses, this property will be ignored.
+    /// </remarks>
+    public bool? Allow { get; set; }
```
stephentoub: Should it be Enabled instead of Allow?

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+        _ = Throw.IfNull(options);
+
+        Allow = options.Allow;
+    }
+
+    /// <summary>Gets or sets a value indicating whether the background responses are allowed.</summary>
+    /// <remarks>
+    /// This property only takes effect if the API it's used with supports background responses.
+    /// If the API does not support background responses, this property will be ignored.
+    /// </remarks>
```
stephentoub: This should ideally describe what "background responses" means. What effect does it have on Get{Streaming}ResponseAsync?

src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 - Added new `AITool.GetService` virtual method.
 - Updated `TextReasoningContent` to include `ProtectedData` for representing encrypted/redacted content.
 - Fixed `MinLength`/`MaxLength`/`Length` attribute mapping in nullable string properties during schema export.
+- Added `[Experimental]` `ResumptionToken` to represent a base class for continuation tokens.
+- Added `[Experimental]` `BackgroundResponsesOptions` to represent options for background responses.
+- Added `[Experimental]` `ChatOptions.BackgroundResponsesOptions` and `ChatOptions.ContinuationToken` properties specify continuation token and options for background responses.
+- Added `[Experimental]` `ChatResponse.ContinuationToken` and `ChatResponseUpdate.ContinuationToken` to return continuation token for background responses.
```
stephentoub: We shouldn't have both the term "resumption" and "continuation". We should use the same term consistently throughout.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -120,6 +121,30 @@ public static Task<ChatResponse> GetResponseAsync(
...
+    /// <param name="options">The chat options to configure the request.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The background response.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="continuationToken"/> is <see langword="null"/>.</exception>
+    [Experimental("MEAI001")]
+    public static Task<ChatResponse> GetResponseAsync(
+        this IChatClient client,
+        ResumptionToken continuationToken,
+        ChatOptions? options = null,
```
stephentoub: What is the purpose of the options being accepted here?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -120,6 +121,30 @@ public static Task<ChatResponse> GetResponseAsync(
...
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(client);
+        _ = Throw.IfNull(continuationToken);
+
+        ChatOptions chatOptions = options ?? new();
+        chatOptions.ContinuationToken = continuationToken;
```
stephentoub: This is going to mutate options the caller provided. We try to avoid that.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -88,6 +88,20 @@ public IList<ChatMessage> Messages
...
+    /// a continuation token if background responses are enabled in <see cref="ChatOptions.BackgroundResponsesOptions"/>
+    /// and the result of the response has not been obtained yet. Otherwise, the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used in conjunction with <see cref="ChatOptions.ContinuationToken"/> to 
+    /// continue or polling for the completion the response. Pass this token to
+    /// <see cref="ChatOptions.ContinuationToken"/> on subsequent calls to <see cref="IChatClient.GetResponseAsync"/>
+    /// to poll for completion.
+    /// </para>
+    /// </remarks>
+    public ResumptionToken? ContinuationToken { get; set; }
```
stephentoub: This should be experimental, right?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
```diff
@@ -141,6 +141,19 @@ public IList<AIContent> Contents
...
+    /// <see cref="IChatClient"/> implementations that support background responses will return
+    /// a continuation token on each update if background responses are enabled in <see cref="ChatOptions.BackgroundResponsesOptions"/>
+    /// and not all updates for the response have been streamed yet. Otherwise, the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used for stream resumption, where the continuation token of the latest received update should be
+    /// passed to <see cref="ChatOptions.ContinuationToken"/> on subsequent calls to <see cref="IChatClient.GetStreamingResponseAsync"/>
+    /// to resume streaming from the point of interruption.
+    /// </para>
+    /// </remarks>
+    public ResumptionToken? ContinuationToken { get; set; }
```
stephentoub: Experimental?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Name": "Microsoft.Extensions.AI.Abstractions, Version=9.6.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
+  "Name": "Microsoft.Extensions.AI.Abstractions, Version=9.10.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
```
stephentoub: You shouldn't need to modify this file at all, as everything this PR adds should be [Experimental].

SergeyMenshykh:

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResumptionToken.cs
```diff
@@ -0,0 +1,67 @@
...
+/// <summary>
+/// Represents a token used to resume, continue, or rehydrate an operation across multiple scenarios/calls,
+/// such as resuming a streamed response from a specific point or retrieving the result of a background operation.
+/// Subclasses of this class encapsulate all necessary information within the token to facilitate these actions.
+/// </summary>
+[JsonConverter(typeof(Converter))]
+public class ResumptionToken
+{
+    /// <summary>Bytes representing this token.</summary>
+    private readonly byte[]? _bytes;
```
stephentoub: I suggest all of these `byte[]` should instead be `ReadOnlyMemory<byte>`.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -215,28 +232,48 @@ internal static IEnumerable<ChatMessage> ToChatMessages(IEnumerable<ResponseItem
...
+        IAsyncEnumerable<StreamingResponseUpdate> streamingResponseUpdates,
+        ResponseCreationOptions? options,
+        string? resumeResponseId = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
         DateTimeOffset? createdAt = null;
-        string? responseId = null;
-        string? conversationId = null;
+        string? responseId = resumeResponseId;
+        string? conversationId = options?.StoredOutputEnabled is false ? null : resumeResponseId;
```
stephentoub, line 257: I'm confused about this line. Can you explain the rationale?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -259,10 +301,48 @@ ChatResponseUpdate CreateUpdate(AIContent? content = null) =>
...
                     conversationId = options?.StoredOutputEnabled is false ? null : responseId;
                     modelId = createdUpdate.Response.Model;
+                    latestResponseStatus = createdUpdate.Response.Status;
+                    goto default;
+
+                case StreamingResponseQueuedUpdate queuedUpdate:
+                    createdAt = queuedUpdate.Response.CreatedAt;
+                    responseId = queuedUpdate.Response.Id;
+                    conversationId = options?.StoredOutputEnabled is false ? null : responseId;
+                    modelId = queuedUpdate.Response.Model;
```
stephentoub, line 297: Are these necessary? I was under the impression they'd all come as part of the CreatedUpdate; same question for the other cases bellow (in progress, incomplete, etc.)

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        if (SequenceNumber.HasValue)
+        {
+            writer.WriteNumber("sequenceNumber", SequenceNumber.Value);
+        }
+
+        writer.WriteEndObject();
+
+        writer.Flush();
+        stream.Position = 0;
```
stephentoub:
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,106 @@
...
+        byte[] data = token.ToBytes();
+
+        if (data.Length == 0)
+        {
+            throw new InvalidOperationException("Failed to create OpenAIResponsesResumptionToken from provided token because it does not contain any data.");
+        }
+
+        Utf8JsonReader reader = new(data);
+
+        string responseId = null!;
```
stephentoub:
```suggestion
        string? responseId = null;
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,106 @@
...
+                case "sequenceNumber":
+                    _ = reader.Read();
+                    startAfter = reader.GetInt32();
+                    break;
+                default:
+                    throw new JsonException($"Unrecognized property '{propertyName}'.");
+            }
+        }
+
+        return new(responseId)
```
stephentoub, line 112: This should fail if responseId is null

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,106 @@
...
+                break;
+            }
+
+            string propertyName = reader.GetString()!;
+
+            switch (propertyName)
+            {
+                case "responseId":
+                    _ = reader.Read();
+                    responseId = reader.GetString()!;
```
stephentoub:
```suggestion
                    responseId = reader.GetString();
```

src/Libraries/Microsoft.Extensions.AI/Microsoft.Extensions.AI.csproj
```diff
@@ -14,7 +14,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(TargetFrameworks);netstandard2.0</TargetFrameworks>
-    <NoWarn>$(NoWarn);MEAI001</NoWarn>
+    <NoWarn>$(NoWarn);MEAI001;S104</NoWarn>
```
stephentoub:
```suggestion
    <NoWarn>$(NoWarn);MEAI001</NoWarn>
```

stephentoub: Let's instead change https://github.com/dotnet/extensions/blob/c4e57fb1e6b8403a527ea3cd737f1146dcbc1f31/src/Libraries/.editorconfig#L2939

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
SergeyMenshykh: It could be, though there is an opinion that it may mislead consumers into thinking that setting that property should enable or disable background responses. In reality, it may not have any effect because the chat client implementation may not support background responses, or the underlying SDK does not allow for controlling the behavior. `Allow` is more permissive than `Enable` and is intended to communicate that it can enable or disable the feature if supported; otherwise, it will have no effect. On the other hand, the `Enable` semantic presumes that the values should be respected once set.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
SergeyMenshykh: Good question. We need to supply functions for subsequent calls if they were advertised in the initial call; otherwise, the called tools won't be found by FICC.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResumptionToken.cs
SergeyMenshykh: Yep, agree, done.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
SergeyMenshykh: I didn't notice that. Thank you for pointing it out!

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
SergeyMenshykh: It's used by the ChatOptions.Clone method to clone BackgroundResponsesOptions. Taking into account that BackgroundResponsesOptions has only one bool property, Allow, we can remove the copy constructor to decrease the public API surface for now. Later, when/if needed, we can reintroduce it or add a Clone method to it.

src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
SergeyMenshykh: We can't rename the ResumptionToken class to the ContinuationToken class without introducing confusion and ambiguity between MEAI and System.ClientModel. However, we can rename the ContinuationToken properties of ChatOptions, ChatResponse, and ChatResponseUpdate to ResumptionToken to have one term throughout consistently.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: They are necessary to get all those properties when resuming (not starting) from one of those events; otherwise, if those updates are not handled, those props won't be initialized.   
  
For example, if the initial GetStreamingResponseAsync call handled the first two updates (sequence_number 0 and 1 in the JSON below) and the connection got interrupted, the consumer code will resume the stream from the update with sequence_number 3. This means that the StreamingResponseCreatedUpdate case won't be hit, nor will the StreamingResponseQueuedUpdate one, and the first case that will be invoked is the StreamingResponseInProgressUpdate case, which will initialize those properties.

```json
{
  "data": {"type":"response.created","sequence_number":0,"response":{"id":"resp_123"}}
  "data": {"type":"response.queued","sequence_number":1,"response":{"id":"resp_123"}}
  "data": {"type":"response.in_progress","sequence_number":2,"response":{"id":"resp_123"}}
  
  "data": {"type":"response.output_item.added","sequence_number":3,"item":{"id":"msg_68d","content":[]}}
  "data": {"type":"response.content_part.added","sequence_number":4,"item_id":"msg_68d","content_index":0,"part":{"text":"","logprobs":[],"type":"output_text","annotations":[]},"output_index":0}
  
  "data": {"type":"response.output_text.delta","sequence_number":5,"delta":"Hello", "item_id":"msg_68d"}
  "data": {"type":"response.output_text.delta","sequence_number":6,"delta":"!"}
  "data": {"type":"response.output_text.delta","sequence_number":7,"delta":" How"}
  "data": {"type":"response.output_text.delta","sequence_number":8,"delta":" can"}
  "data": {"type":"response.output_text.delta","sequence_number":9,"delta":" I"}
  "data": {"type":"response.output_text.delta","sequence_number":10,"delta":" assist"}
  "data": {"type":"response.output_text.delta","sequence_number":11,"delta":" you"}
  "data": {"type":"response.output_text.delta","sequence_number":12,"delta":" today"}
  "data": {"type":"response.output_text.delta","sequence_number":13,"delta":"?"}
  
  "data": {"type":"response.output_text.done","sequence_number":14,"text":"Hello! How can I assist you today?", "item_id":"msg_68d"}
  "data": {"type":"response.content_part.done","sequence_number":15,"item_id":"msg_68d", "part":{"text":"Hello! How can I assist you today?"}
  "data": {"type":"response.output_item.done","sequence_number":16,"item":{"id":"msg_68d","content":[{"text":"Hello! How can I assist you today?"}]}}
  
  "data": {"type":"response.completed","sequence_number":17,"response":{"id":"resp_123", "status":"completed"}}
}
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: Not all stream events have a response id. So, if we resume streaming, let's say, from the response.output_text.delta event with sequence_number 8, we won't get the responseId variable initialized because the update does not have it. As a result, we won't be able to create a continuation token. However, the continuation token that GetStreamingResponseAsync is called with has the response id, and we can use it to initialize the responseId and conversationId variables and consequently use them for continuation token creation.

```json
{
  "data": {"type":"response.created","sequence_number":0,"response":{"id":"resp_123"}}
  "data": {"type":"response.queued","sequence_number":1,"response":{"id":"resp_123"}}
  "data": {"type":"response.in_progress","sequence_number":2,"response":{"id":"resp_123"}}
  
  "data": {"type":"response.output_item.added","sequence_number":3,"item":{"id":"msg_68d","content":[]}}
  "data": {"type":"response.content_part.added","sequence_number":4,"item_id":"msg_68d","content_index":0,"part":{"text":"","logprobs":[],"type":"output_text","annotations":[]},"output_index":0}
  
  "data": {"type":"response.output_text.delta","sequence_number":5,"delta":"Hello", "item_id":"msg_68d"}
  "data": {"type":"response.output_text.delta","sequence_number":6,"delta":"!"}
  "data": {"type":"response.output_text.delta","sequence_number":7,"delta":" How"}
  "data": {"type":"response.output_text.delta","sequence_number":8,"delta":" can"}
  "data": {"type":"response.output_text.delta","sequence_number":9,"delta":" I"}
  "data": {"type":"response.output_text.delta","sequence_number":10,"delta":" assist"}
  "data": {"type":"response.output_text.delta","sequence_number":11,"delta":" you"}
  "data": {"type":"response.output_text.delta","sequence_number":12,"delta":" today"}
  "data": {"type":"response.output_text.delta","sequence_number":13,"delta":"?"}
  
  "data": {"type":"response.output_text.done","sequence_number":14,"text":"Hello! How can I assist you today?", "item_id":"msg_68d"}
  "data": {"type":"response.content_part.done","sequence_number":15,"item_id":"msg_68d", "part":{"text":"Hello! How can I assist you today?"}
  "data": {"type":"response.output_item.done","sequence_number":16,"item":{"id":"msg_68d","content":[{"text":"Hello! How can I assist you today?"}]}}
  
  "data": {"type":"response.completed","sequence_number":17,"response":{"id":"resp_123", "status":"completed"}}
}
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
SergeyMenshykh: Done

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
SergeyMenshykh: removed

src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
SergeyMenshykh: The ResumptionToken type, used as a continuation token, has been renamed to ResponseContinuationToken.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
SergeyMenshykh: The only type that can't be marked experimental is ResponseContinuationToken - making it experimental leads to a few compilation errors in the source-generated code. So, we mark its constructors and public static factory method as experimental instead and have to regenerate this baseline file; otherwise, compilation fails with the error: _"ResponseContinuationToken.cs(19,14,19,39): error LA0003: Newly added symbol 'Microsoft.Extensions.AI.ResponseContinuationToken' must be marked as experimental."_

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -88,6 +88,22 @@ public IList<ChatMessage> Messages
     /// <summary>Gets or sets usage details for the chat response.</summary>
     public UsageDetails? Usage { get; set; }
 
+    /// <summary>Gets or sets the continuation token for getting result of the background chat response.</summary>
+    /// <remarks>
+    /// <see cref="IChatClient"/> implementations that support background responses will return
+    /// a continuation token if background responses are enabled in <see cref="ChatOptions.BackgroundResponsesOptions"/>
```
westey-m:
```suggestion
    /// a continuation token if background responses are allowed in <see cref="ChatOptions.BackgroundResponsesOptions"/>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -88,6 +88,22 @@ public IList<ChatMessage> Messages
...
+    /// and the result of the response has not been obtained yet. Otherwise, the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used in conjunction with <see cref="ChatOptions.ContinuationToken"/> to 
+    /// continue or polling for the completion the response. Pass this token to
```
westey-m:
```suggestion
    /// continue to poll for the completion of the response. Pass this token to
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResponseContinuationToken.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a token used to resume, continue, or rehydrate an operation across multiple scenarios/calls,
+/// such as resuming a streamed response from a specific point or retrieving the result of a background operation.
+/// Subclasses of this class encapsulate all necessary information within the token to facilitate these actions.
+/// </summary>
+[JsonConverter(typeof(Converter))]
+public class ResponseContinuationToken
```
westey-m, line 20: Are we thinking this would be potentially usable for services other than chat?
I see it's not in the `ChatCompletion` folder, which would suggest that to be the case.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -115,6 +116,26 @@ public class ChatOptions
     [JsonIgnore]
     public IList<AITool>? Tools { get; set; }
 
+    /// <summary>Gets or sets the options for configuring background responses.</summary>
+    [Experimental("MEAI001")]
+    [JsonIgnore]
+    public BackgroundResponsesOptions? BackgroundResponsesOptions { get; set; }
+
+    /// <summary>Gets or sets the continuation token for resuming and getting result of the chat response identified by this token.</summary>
```
westey-m:
```suggestion
    /// <summary>Gets or sets the continuation token for resuming and getting the result of the chat response identified by this token.</summary>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Options for configuring background responses.</summary>
+[Experimental("MEAI001")]
+public sealed class BackgroundResponsesOptions
```
westey-m: What were the other options again that we were thinking we may need to add here in future to justify a separate options object?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
westey-m: Should we add some remarks here to warn users that the options should match the original options, otherwise you may have unexpected results and any such action is not supported?

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/TestJsonSerializerContext.cs
```diff
@@ -37,4 +37,5 @@ namespace Microsoft.Extensions.AI;
 [JsonSerializable(typeof(decimal))] // Used in Content tests
 [JsonSerializable(typeof(HostedMcpServerToolApprovalMode))]
 [JsonSerializable(typeof(ChatResponseFormatTests.SomeType))]
+[JsonSerializable(typeof(ResponseContinuationToken))]
```
westey-m, line 40: Do we also need this in the AIJsonUtilities for MEAI.A?
https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs#L120

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,111 @@
...
+internal sealed class OpenAIResponsesContinuationToken : ResponseContinuationToken
+{
+    /// <summary>Initializes a new instance of the <see cref="OpenAIResponsesContinuationToken"/> class.</summary>
+    internal OpenAIResponsesContinuationToken(string responseId)
+    {
+        ResponseId = responseId;
+    }
+
+    /// <summary>Gets or sets the Id of the response.</summary>
+    internal string ResponseId { get; set; }
```
westey-m: Does this need to be settable considering that we require it on the constructor?

SergeyMenshykh: Nope, the setter isn't needed - removed.

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
SergeyMenshykh: One that I can think of at the moment would be settings for polling if we ever decide to do internal polling. However, if we agree to offload polling responsibility to consumers, and if there are no other options we can think of, we can remove the class and go with the `bool? ChatOptions.AllowBackgroundResponses` property instead. That would simplify the API a little bit from:
```csharp
var chatOptions = new ChatOptions
{
    BackgroundResponsesOptions = new() 
    { 
        Allow = true
    }
};

var response = await ChatClient.GetResponseAsync("<prompt>", chatOptions);
```
to
```csharp
var chatOptions = new ChatOptions
{
    AllowBackgroundResponses = true
};

var response = await ChatClient.GetResponseAsync("<prompt>", chatOptions);
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResponseContinuationToken.cs
SergeyMenshykh: The plan is to use it for long-running operations in AF as well; hence, it's outside of the folder. We could also broaden its usage and use it for something unrelated to background responses in the future if we find a good name for it that **a.** does not have the `Response` prefix, **b.** does not have `ContinuationToken` in the name, which adds ambiguity with System.ChatClient.ContinuationToken, and **c.** could be used for the ChatResponse and ChatResponseUpdate continuation property without introducing a second term, as pointed out here - https://github.com/dotnet/extensions/pull/6854#discussion_r2382603380.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
SergeyMenshykh: Good idea - added.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/TestJsonSerializerContext.cs
SergeyMenshykh: Yep, looks like it, thanks!

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
stephentoub: > we can remove the class and go with the bool? ChatOptions.AllowBackgroundResponses property instead

Let's do that.

src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 - Added protected copy constructors to options types (e.g. `ChatOptions`).
 - Fixed `EmbeddingGeneratorOptions`/`SpeechToTextOptions` `Clone` methods to correctly copy all properties.
 - Fixed `ToChatResponse` to not overwrite `ChatMessage/ChatResponse.CreatedAt` with older timestamps during coalescing. 
+- Added `[Experimental]` `ResponseContinuationToken` to represent a base class for continuation tokens.
+- Added `[Experimental]` `BackgroundResponsesOptions` to represent options for background responses.
+- Added `[Experimental]` `ChatOptions.BackgroundResponsesOptions` and `ChatOptions.ContinuationToken` properties to accept options for background responses and continuation token.
+- Added `[Experimental]` `ChatResponse.ContinuationToken` and `ChatResponseUpdate.ContinuationToken` to return continuation token for background responses.
```
stephentoub:
```suggestion
- Added `[Experimental]` support for background responses, such that non-streaming responses are allowed to be pollable, and such that responses and response updates can be tagged with continuation tokens to support later resumption.
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -120,6 +121,34 @@ public static Task<ChatResponse> GetResponseAsync(
         return client.GetResponseAsync([chatMessage], options, cancellationToken);
     }
 
+    /// <summary>Gets a background response identified by the specified continuation token.</summary>
```
stephentoub:
```suggestion
    /// <summary>Gets a previously-submitted background response identified by the specified continuation token.</summary>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -120,6 +121,34 @@ public static Task<ChatResponse> GetResponseAsync(
...
+    /// <param name="client">The chat client.</param>
+    /// <param name="continuationToken">The continuation token.</param>
```
stephentoub:
```suggestion
    /// <param name="continuationToken">The continuation token provided with the previous response.</param>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -120,6 +121,34 @@ public static Task<ChatResponse> GetResponseAsync(
...
+    /// <param name="options">The chat options to configure the request.</param>
```
stephentoub:
```suggestion
    /// <param name="options">The chat options with which to configure the request.</param>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
```diff
@@ -120,6 +121,34 @@ public static Task<ChatResponse> GetResponseAsync(
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>The background response.</returns>
```
stephentoub: Can you elaborate on what "background response" means?  And is it necessarily a "background" response? Does the abstraction contract say that a continuation token is only ever provided if AllowBackground is true?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
stephentoub: > We need to supply functions for subsequent calls if they were advertised in the initial call; otherwise, the called tools won't be found by FICC.

The remarks should be more explicit about this. Even after reading the remarks, I didn't understand why this was necessary.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
stephentoub: > The only type that can't be marked experimental is ResponseContinuationToken - making it experimental leads to a few compilation errors in the source-generated code.

Why is it in the source-generated code at all?

stephentoub: Also, I'm hopeful you won't need to modify this json file at all... but if you absolutely must, please do so by reverting all the changes and then just adding the few new lines that are necessary to cover the new things. The regenerated file is really hard to review.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
```diff
@@ -116,6 +116,7 @@ private static JsonSerializerOptions CreateDefaultOptions()
     [JsonSerializable(typeof(Embedding<double>))]
     [JsonSerializable(typeof(AIContent))]
     [JsonSerializable(typeof(AIFunctionArguments))]
+    [JsonSerializable(typeof(ResponseContinuationToken))]
```
stephentoub, line 138: This should be commented out for now. And once you do so, presumably you can again mark ResponseContinuationToken as [Experimental], revert the per-member annotations, and completely revert the json file?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -81,12 +81,28 @@ public OpenAIResponsesChatClient(OpenAIResponseClient responseClient)
...
     {
-        _ = Throw.IfNull(messages);
+        var inputMessages = Throw.IfNull(messages) as IReadOnlyCollection<ChatMessage> ?? [.. messages];
 
-        // Convert the inputs into what OpenAIResponseClient expects.
-        var openAIResponseItems = ToOpenAIResponseItems(messages, options);
         var openAIOptions = ToOpenAIResponseCreationOptions(options);
 
+        // Convert the inputs into what OpenAIResponseClient expects.
+        var openAIResponseItems = ToOpenAIResponseItems(inputMessages, options);
```
stephentoub: Why was it necessary to change the type of inputMessages? Couldn't L94 below fail iff openAIResponseItems.Count > 0? Or what scenario has zero input messages but > 0 converted items?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -215,28 +232,48 @@ internal static IEnumerable<ChatMessage> ToChatMessages(IEnumerable<ResponseItem
     public IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
         IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default)
     {
-        _ = Throw.IfNull(messages);
+        var inputMessages = Throw.IfNull(messages) as IReadOnlyCollection<ChatMessage> ?? [.. messages];
```
stephentoub: Same question

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
SergeyMenshykh: Added more details to the remarks to make it clearer.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Microsoft.Extensions.AI.Abstractions.json
SergeyMenshykh: > Why is it in the source-generated code at all?

Because the ResponseContinuationToken referenced by ChatResponse, which is referenced by ScenarioRunResult that is [registered for source generation](https://github.com/dotnet/extensions/blob/5858a796217aab02fe4bcabe67447987c2cd6530/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting.Azure/JsonSerialization/AzureStorageJsonUtilities.cs#L42) in the Microsoft.Extensions.AI.Evaluation.Reporting project.

Despite the ChatResponse.ContinuationToken property being annotated with [JsonIgnore], it still causes compilation errors if its type, `ResponseContinuationToken`, is marked as experimental. CC: @eiriktsarpalis

> Also, I'm hopeful you won't need to modify this json file at all... but if you absolutely must, please do...

The changes were made by the `./scripts/MakeApiBaselines.ps1` script that is recommended in the building.md guide. I rolled back changes not relevant to this PR and kept only the relevant ones.

src/Libraries/Microsoft.Extensions.AI.Abstractions/BackgroundResponsesOptions.cs
SergeyMenshykh: Done

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: It doesn't seem necessary anymore. It was added because of a warning about multiple enumerations of the `messages` parameter. I’ve rolled it back.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: It doesn't seem necessary anymore. It was added because of a warning about multiple enumerations of the `messages` parameter. I’ve rolled it back.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
SergeyMenshykh: Currently, the continuation token is only used with background responses, but the abstraction does not enforce this. As a result, it may be possible to implement this in methods with non-background responses in the future when we identify a suitable use case(s).  
  
We have a few options at the moment:  
  
1. Explicitly state in the XML comments that these methods work only with background responses. Later, if we identify another use case for the continuation token, we can update the XML comments accordingly.  
2. Use more generic language in the XML comments, indicating that the methods can return responses identified by the continuation token without specifying that they are for background responses only.  
3. Rename the extension methods to GetBackground{Streaming}ResponseAsync to clarify that they are intended for background response scenarios only. We can add new method pair(s) later when we identify additional scenarios that require the use of the continuation token.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
SergeyMenshykh: I've reverted the baseline json after changing the type of all public properties and parameters representing the continuation token to an object and marking the ResponseContinuationToken as experimental.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs
SergeyMenshykh: Agreed to drop the extension methods for now and introduce them later when there's more clarity about scenarios in which the continuation token can be used.

src/Libraries/Microsoft.Extensions.AI.Abstractions/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 - Added protected copy constructors to options types (e.g. `ChatOptions`).
 - Fixed `EmbeddingGeneratorOptions`/`SpeechToTextOptions` `Clone` methods to correctly copy all properties.
 - Fixed `ToChatResponse` to not overwrite `ChatMessage/ChatResponse.CreatedAt` with older timestamps during coalescing. 
+- Added `[Experimental]` support for background responses, such that non-streaming responses are allowed to be pollable, and such that responses and response updates can be tagged with continuation tokens to support later resumption.
+- Added `[Experimental]` `GetResponseAsync` and `GetStreamingResponseAsync` extension methods for `IChatClient` to continue background responses.
```
stephentoub:
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -155,6 +158,47 @@ protected ChatOptions(ChatOptions? other)
...
     public IList<AITool>? Tools { get; set; }
 
+    /// <summary>Gets or sets a value indicating whether the background responses are allowed.</summary>
+    /// <remarks>
+    /// <para>
+    /// Background responses allow running long-running operations or tasks asynchronously in the background that can be resumed by streaming APIs
+    /// and polled for completion by non-streaming APIs.
+    /// </para>
+    /// <para>
+    /// When this property is set to true, non-streaming APIs start a background operation and return an initial
```
stephentoub:
```suggestion
    /// When this property is set to true, non-streaming APIs may start a background operation and return an initial
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -155,6 +158,47 @@ protected ChatOptions(ChatOptions? other)
...
+    /// response with a continuation token. Subsequent calls to the same API should be made in a polling manner with
+    /// the continuation token to get the final result of the operation.
+    /// </para>
+    /// <para>
+    /// When this property is set to true, streaming APIs also start a background operation and begin streaming
```
stephentoub:
```suggestion
    /// When this property is set to true, streaming APIs may also start a background operation and begin streaming
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -155,6 +158,47 @@ protected ChatOptions(ChatOptions? other)
...
+    /// response updates until the operation is completed. If the streaming connection is interrupted, the
+    /// continuation token obtained from the last update should be supplied to a subsequent call to the same streaming API
```
stephentoub:
```suggestion
    /// continuation token obtained from the last update that has one should be supplied to a subsequent call to the same streaming API
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponseUpdate.cs
```diff
@@ -136,6 +136,21 @@ public IList<AIContent> Contents
...
+    /// a continuation token on each update if background responses are allowed in <see cref="ChatOptions.AllowBackgroundResponses"/>
+    /// and not all updates for the response have been streamed yet. Otherwise, the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used for stream resumption, where the continuation token of the latest received update should be
+    /// passed to <see cref="ChatOptions.ContinuationToken"/> on subsequent calls to <see cref="IChatClient.GetStreamingResponseAsync"/>
+    /// to resume streaming from the point of interruption.
+    /// </para>
+    /// </remarks>
+    [JsonIgnore]
+    [Experimental("MEAI001")]
```
stephentoub: Nit:
```suggestion
    [Experimental("MEAI001")]
    [JsonIgnore]
```
Just for a consistent ordering with the others

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResponseContinuationToken.cs
```diff
@@ -0,0 +1,70 @@
...
+    protected ResponseContinuationToken()
+    {
+    }
+
+    /// <summary>Initializes a new instance of the <see cref="ResponseContinuationToken"/> class.</summary>
+    /// <param name="bytes">Bytes to create the token from.</param>
+    protected ResponseContinuationToken(ReadOnlyMemory<byte> bytes)
+    {
+        _ = Throw.IfNull(bytes);
+
```
stephentoub: ReadOnlyMemory is a struct
```suggestion
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResponseContinuationToken.cs
```diff
@@ -0,0 +1,70 @@
...
+/// Represents a token used to resume, continue, or rehydrate an operation across multiple scenarios/calls,
+/// such as resuming a streamed response from a specific point or retrieving the result of a background operation.
+/// Subclasses of this class encapsulate all necessary information within the token to facilitate these actions.
+/// </summary>
+[JsonConverter(typeof(Converter))]
+[Experimental("MEAI001")]
+public class ResponseContinuationToken
+{
+    /// <summary>Bytes representing this token.</summary>
+    private readonly ReadOnlyMemory<byte>? _bytes;
```
stephentoub:
```suggestion
    private readonly ReadOnlyMemory<byte> _bytes;
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ResponseContinuationToken.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <summary>Create a new instance of <see cref="ResponseContinuationToken"/> from the provided <paramref name="bytes"/>.
+    /// </summary>
+    /// <param name="bytes">Bytes representing the <see cref="ResponseContinuationToken"/>.</param>
+    /// <returns>A <see cref="ResponseContinuationToken"/> equivalent to the one from which
+    /// the original<see cref="ResponseContinuationToken"/> bytes were obtained.</returns>
+    public static ResponseContinuationToken FromBytes(ReadOnlyMemory<byte> bytes) => new(bytes);
+
+    /// <summary>Gets the bytes representing this <see cref="ResponseContinuationToken"/>.</summary>
+    /// <returns>Bytes representing the <see cref="ResponseContinuationToken"/>.</returns>"/>
+    public virtual ReadOnlyMemory<byte> ToBytes() => _bytes ?? throw new InvalidOperationException("Unable to obtain this token's bytes.");
```
stephentoub:
```suggestion
    public virtual ReadOnlyMemory<byte> ToBytes() => _bytes;
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -83,9 +83,25 @@ public async Task<ChatResponse> GetResponseAsync(
     {
         _ = Throw.IfNull(messages);
 
+        var openAIOptions = ToOpenAIResponseCreationOptions(options);
+
```
stephentoub: Why did this move up?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -83,9 +83,25 @@ public async Task<ChatResponse> GetResponseAsync(
...
+        // Provided continuation token signals that an existing background response should be fetched.
+        if (options?.ContinuationToken is { } token)
+        {
+            if (messages.Any())
+            {
+                throw new InvalidOperationException("Messages are not allowed when continuing a background response using a continuation token.");
+            }
+
+            var continuationToken = OpenAIResponsesContinuationToken.FromToken(token);
```
stephentoub: You could dedup these lines between Get{Streaming}ResponseAsync into a `GetContinuationToken(messages, options)` method, e.g. `if (GetContinuationToken(messages, options) is { } token)`. We don't need the differentiated exception message.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -217,26 +234,46 @@ public IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
...
+        if (options?.ContinuationToken is { } token)
+        {
+            if (messages.Any())
+            {
+                throw new InvalidOperationException("Messages are not allowed when resuming streamed background response using a continuation token.");
+            }
+
+            var continuationToken = OpenAIResponsesContinuationToken.FromToken(token);
+
+            IAsyncEnumerable<StreamingResponseUpdate> updates = _responseClient.GetResponseStreamingAsync(continuationToken.ResponseId, continuationToken.SequenceNumber, cancellationToken);
```
stephentoub: We're not going to get our user agent header added on this call. Something to look at fixing in the future; not a blocker.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -892,6 +973,51 @@ private static void AddAllMcpFilters(IList<string> toolNames, McpToolFilter filt
...
+
+        // Return a continuation token for in-progress or queued responses because they are not yet complete.
+        if (responseStatus is (ResponseStatus.InProgress or ResponseStatus.Queued))
+        {
+            return new OpenAIResponsesContinuationToken(responseId)
+            {
+                SequenceNumber = updateSequenceNumber,
+            };
+        }
+        else if (responseStatus is null && updateSequenceNumber is not null)
```
stephentoub:
```suggestion

        if (responseStatus is null && updateSequenceNumber is not null)
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -892,6 +973,51 @@ private static void AddAllMcpFilters(IList<string> toolNames, McpToolFilter filt
...
+        {
+            // In some cases, streaming needs to be resumed from an event (e.g., a text delta event) that does not have a response status,
+            // response Id, and potentially other properties. In these cases, we know that the response is not yet complete
+            // because we are receiving updates for it, so we create a continuation token from the response Id obtained from the continuation token
+            // supplied to resume the streaming and the sequence number available on the event.
+            return new OpenAIResponsesContinuationToken(responseId)
+            {
+                SequenceNumber = updateSequenceNumber,
+            };
```
stephentoub: Isn't this an exact copy of the previous if block? Can they not be collapsed?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -892,6 +973,51 @@ private static void AddAllMcpFilters(IList<string> toolNames, McpToolFilter filt
...
+        }
+        else if (responseStatus is null && updateSequenceNumber is not null)
+        {
+            // In some cases, streaming needs to be resumed from an event (e.g., a text delta event) that does not have a response status,
+            // response Id, and potentially other properties. In these cases, we know that the response is not yet complete
+            // because we are receiving updates for it, so we create a continuation token from the response Id obtained from the continuation token
+            // supplied to resume the streaming and the sequence number available on the event.
```
stephentoub: I'm not following this comment.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        if (data.Length == 0)
+        {
+            Throw.InvalidOperationException("Failed to create OpenAIResponsesResumptionToken from provided token because it does not contain any data.");
+        }
+
+        Utf8JsonReader reader = new(data.Span);
+
+        string? responseId = null;
+        int? startAfter = null;
```
stephentoub: Why call this startAfter rather than sequenceNumber, which is the name of both the serialized JSON property and the name of the property it's then being assigned to?

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
```diff
@@ -88,6 +88,22 @@ public IList<ChatMessage> Messages
     /// <summary>Gets or sets usage details for the chat response.</summary>
     public UsageDetails? Usage { get; set; }
 
+    /// <summary>Gets or sets the continuation token for getting result of the background chat response.</summary>
+    /// <remarks>
+    /// <see cref="IChatClient"/> implementations that support background responses will return
+    /// a continuation token if background responses are allowed in <see cref="ChatOptions.AllowBackgroundResponses"/>
+    /// and the result of the response has not been obtained yet. Otherwise, the token will be <see langword="null"/>.
```
stephentoub: Do we need to be more explicit about this returning null once the operation completes? Consumption relies on that pretty heavily to stop looping, right?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
```diff
@@ -0,0 +1,116 @@
...
+        if (token is not ResponseContinuationToken)
+        {
+            Throw.ArgumentException(nameof(token), "Failed to create OpenAIResponsesResumptionToken from provided token because it is not of type ResponseContinuationToken.");
+        }
+
+        ReadOnlyMemory<byte> data = ((ResponseContinuationToken)token).ToBytes();
+
+        if (data.Length == 0)
+        {
+            Throw.InvalidOperationException("Failed to create OpenAIResponsesResumptionToken from provided token because it does not contain any data.");
```
stephentoub: I'm unclear on when you're throwing an ArgumentException, when you're throwing an InvalidOperationException, etc. (and presumably you might throw a JsonException if the JSON fails to parse)? We should probably ensure that an invalid input always results in the same exception type.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ResponseContinuationTokenTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using Xunit;
+
+namespace Microsoft.Extensions.AI;
+
+public class ResponseContinuationTokenTests
+{
+    [Fact]
+    public void Bytes_Roundtrip()
+    {
+        byte[] testBytes = [1, 2, 3, 4, 5];
```
stephentoub: You could turn this into a [Theory] validating multiple inputs, including an empty one.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientIntegrationTests.cs
```diff
@@ -189,4 +190,146 @@ await client.GetStreamingResponseAsync(input, chatOptions).ToChatResponseAsync()
...
+        Assert.Empty(response.Messages);
+
+        int attempts = 0;
+
+        // Continue to poll until we get the final response
+        while (response.ContinuationToken is not null && ++attempts < 10)
+        {
+            chatOptions.ContinuationToken = response.ContinuationToken;
+            response = await ChatClient.GetResponseAsync([], chatOptions);
+            await Task.Delay(500);
```
stephentoub: This integration test is going to fail then if it doesn't complete in 5 seconds? Is that sufficient?

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientIntegrationTests.cs
```diff
@@ -189,4 +190,146 @@ await client.GetStreamingResponseAsync(input, chatOptions).ToChatResponseAsync()
...
+    public async Task GetResponseAsync_BackgroundResponses_WithFunction()
+    {
+        SkipIfNotEnabled();
+
+        using var chatClient = new FunctionInvokingChatClient(ChatClient);
+
+        var chatOptions = new ChatOptions
+        {
+            AllowBackgroundResponses = true,
+            Tools = [AIFunctionFactory.Create(() => "5:43", new AIFunctionFactoryOptions { Name = "GetCurrentTime" })]
```
stephentoub: It'd be good to update this function so that it mutates a captured local which can then be checked to validate the function is invoked once and, importantly, only once.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
```diff
@@ -1525,4 +1986,101 @@ private static IChatClient CreateResponseClient(HttpClient httpClient, string mo
...
+    {
+        if (status == ResponseStatus.InProgress)
+        {
+            return "in_progress";
+        }
+
+        return status.ToString().ToLowerInvariant();
+    }
+
+    private sealed class TestOpenAIResponsesContinuationToken : ResponseContinuationToken
```
stephentoub, line 2000: Why is this needed?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: It's the other line that went down because it was originally planned to be used only for creating new responses. Then it was moved up because I thought we could allow the "modification" of a response identified by a continuation token, which we agreed we should not support. Now, I have moved it down again because it's only relevant to creating new responses.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: Simplified it a little bit; hopefully, it's more clear now.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
SergeyMenshykh: It was the original name, which I've changed in other places but did not change here.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
SergeyMenshykh: Good point, captured.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatResponse.cs
SergeyMenshykh: Done

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesContinuationToken.cs
SergeyMenshykh: Aligned the exception types.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientIntegrationTests.cs
SergeyMenshykh: Increased delay to 1s to align with the other test. So far, none of the new tests have failed when run locally. We can increase the delay or the number of attempts if it appears to be insufficient.

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientIntegrationTests.cs
SergeyMenshykh: Done

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
SergeyMenshykh: Since the OpenAIResponsesContinuationToken is internal to the Microsoft.Extensions.AI.OpenAI project, I can't use it. However, I need a way to verify that OpenAIResponsesChatClient initializes tokens with the correct responseId and sequenceNumber, and that the responseId and sequenceNumber from the token are used by OpenAIResponsesChatClient in the corresponding call of the underlying API.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ResponseContinuationTokenTests.cs
SergeyMenshykh: Done

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
stephentoub: > However, I need a way to verify that OpenAIResponsesChatClient initializes tokens with the correct responseId and sequenceNumber, and that the responseId and sequenceNumber from the token are used by OpenAIResponsesChatClient in the corresponding call of the underlying API.

Can't that be determined by looking at what it in turn sends out to the http handler?

The token is supposed to be opaque; the test is relying on specific internal implementation details.

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -155,6 +158,47 @@ protected ChatOptions(ChatOptions? other)
...
+    /// </para>
+    /// <para>
+    /// When this property is set to true, streaming APIs may also start a background operation and begin streaming
+    /// response updates until the operation is completed. If the streaming connection is interrupted, the
+    /// continuation token obtained from the last update that has one should be supplied to a subsequent call to the same streaming API
+    /// to resume the stream from the point of interruption and continue receiving updates until the operation is completed.
+    /// </para>
+    /// <para>
+    /// This property only takes effect if the API it's used with supports background responses.
+    /// If the API does not support background responses, this property will be ignored.
```
stephentoub:
```suggestion
    /// This property only takes effect if the implementation it's used with supports background responses.
    /// If the implementation does not support background responses, this property will be ignored.
```

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -892,6 +959,59 @@ private static void AddAllMcpFilters(IList<string> toolNames, McpToolFilter filt
...
+        {
+            return null;
+        }
+
+        // Returns a continuation token for in-progress or queued responses as they are not yet complete.
+        // Also returns a continuation token if there is no status but there is a sequence number,
+        // which can occur for certain streaming updates related to response content part updates: response.content_part.*,
+        // response.output_text.*
+        if ((responseStatus == ResponseStatus.InProgress || responseStatus == ResponseStatus.Queued)
+            || (responseStatus == null && updateSequenceNumber != null))
```
stephentoub:
```suggestion
        if ((responseStatus is ResponseStatus.InProgress or ResponseStatus.Queued) ||
            (responseStatus is null && updateSequenceNumber is not null))
```

test/Libraries/Microsoft.Extensions.AI.OpenAI.Tests/OpenAIResponseClientTests.cs
stephentoub: We can merge the PR and address this subsequently.

stephentoub:
@SergeyMenshykh, why do you keep merging in main? It keeps resetting CI.

PR #6863, opened 2025-09-29
Update to .NET 10 SDK

eerhardt:
Bringing in the .NET 10 SDK and dotnet/arcade. No changes to which TFMs we build for in this PR. That will come in a later change.

There are a few new analyzer violations that I needed to work around.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6863)

### DIFF
--- a/src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj
+++ b/src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj
@@ -10,6 +10,7 @@
     <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
     <!-- using the ConfigurationBinder source generator is blocked by https://github.com/dotnet/runtime/issues/94547, which was only fixed in 9.0+ -->
     <EnableConfigurationBindingGenerator Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">true</EnableConfigurationBindingGenerator>
+    <!-- This is needed until we get a build with https://github.com/dotnet/sdk/pull/50662. .NET 10 RC2 should have the fix. -->
     <InterceptorsNamespaces Condition="'$(EnableConfigurationBindingGenerator)' == 'true'">$(InterceptorsNamespaces);Microsoft.Extensions.Configuration.Binder.SourceGeneration</InterceptorsNamespaces>
     <UseLoggingGenerator>true</UseLoggingGenerator>
     <UseMetricsGenerator>true</UseMetricsGenerator>
--- a/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
+++ b/src/Libraries/Microsoft.AspNetCore.Diagnostics.Middleware/Microsoft.AspNetCore.Diagnostics.Middleware.csproj
@@ -11,6 +11,7 @@
   <PropertyGroup>
     <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
     <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
+    <!-- This is needed until we get a build with https://github.com/dotnet/sdk/pull/50662. .NET 10 RC2 should have the fix. -->
     <InterceptorsNamespaces>$(InterceptorsNamespaces);Microsoft.Extensions.Configuration.Binder.SourceGeneration</InterceptorsNamespaces>
     <UseLoggingGenerator>true</UseLoggingGenerator>
     <InjectSharedDebugger>true</InjectSharedDebugger>


src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj
```diff
@@ -10,6 +10,7 @@
     <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
     <!-- using the ConfigurationBinder source generator is blocked by https://github.com/dotnet/runtime/issues/94547, which was only fixed in 9.0+ -->
     <EnableConfigurationBindingGenerator Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">true</EnableConfigurationBindingGenerator>
+    <InterceptorsNamespaces Condition="'$(EnableConfigurationBindingGenerator)' == 'true'">$(InterceptorsNamespaces);Microsoft.Extensions.Configuration.Binder.SourceGeneration</InterceptorsNamespaces>
```
eerhardt, line 14: Need to figure out if this is the correct fix here. I'm not sure why this isn't enabled automatically when setting `EnableConfigurationBindingGenerator=true`

src/Libraries/.editorconfig
```diff
@@ -2012,7 +2012,7 @@ dotnet_style_null_propagation = true
 # Title    : Use auto property
 # Category : Style
 # Help Link: https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0032
-dotnet_diagnostic.IDE0032.severity = warning
+dotnet_diagnostic.IDE0032.severity = silent
```
stephentoub: Why?

eerhardt: I'm unsure why these 2 warnings are now popping up with the new SDK. Need to figure out why.

test/Generators/Microsoft.Gen.Logging/Unit/LogParserUtilitiesTests.cs
```diff
@@ -88,7 +88,7 @@ public void RecordHasSensitivePublicMembers_ShouldNotThrow_WhenNoMembersOnType(b
 
         symbolMock
             .Setup(x => x.GetAttributes())
-            .Returns(Array.Empty<AttributeData>().ToImmutableArray());
+            .Returns(ImmutableArray.ToImmutableArray(Array.Empty<AttributeData>()));
```
stephentoub: Why did these need to switch to use static method instead of extension method syntax?

src/Libraries/.editorconfig
stephentoub: Did bringing in the new SDK update LangVersion to C# 14? If so, it's most likely because `field` now exists.

joperezr: Probably not updating the SDK, but likely updating arcade would do this, yes.

joperezr (approved):
Assuming CI is green, this LGTM. Thanks so much for taking care of this @eerhardt!

shyamnamboodiripad (approved):

src/Libraries/.editorconfig
eerhardt: > If so, it's most likely because field now exists.

yep it is this. Should I update these in this PR or log an issue and fix it separately?

eerhardt: https://github.com/dotnet/extensions/blob/6786d6d900ca18fb869baca49ea9d34972ff9f22/src/Libraries/Microsoft.Extensions.AI.Evaluation.Reporting/CSharp/JsonSerialization/JsonUtilities.cs#L17-L18

is an example of a warning.

eerhardt: See conversation in https://github.com/dotnet/extensions/pull/6863#pullrequestreview-3281793712.

stephentoub: > yep it is this. Should I update these in this PR or log an issue and fix it separately?

Either is fine, just don't want to lose track of it.

test/Generators/Microsoft.Gen.Logging/Unit/LogParserUtilitiesTests.cs
eerhardt: Because there are 2 `ToImmutableArray()` extension methods in scope:

```
'System.Collections.Immutable.ImmutableArray.ToImmutableArray<T>(System.ReadOnlySpan<T>)' 
'Microsoft.Gen.Shared.RoslynExtensions.ToImmutableArray<T>(System.ReadOnlySpan<T>)'
```

stephentoub: Can we delete https://github.com/dotnet/extensions/blob/6786d6d900ca18fb869baca49ea9d34972ff9f22/src/Generators/Shared/RoslynExtensions.cs#L106 instead?

eerhardt: Fixed.

test/Libraries/Microsoft.Extensions.Compliance.Abstractions.Tests/Redaction/RedactionAbstractionsExtensionsTest.cs
```diff
@@ -24,12 +24,7 @@ public static void When_Passed_Null_Value_String_Builder_Extensions_Does_Not_App
         var sb = new StringBuilder();
         var redactor = NullRedactor.Instance;
 
-        sb.AppendRedacted(NullRedactor.Instance,
-#if NETCOREAPP3_1_OR_GREATER
-            null);
-#else
-    (string?)null);
-#endif
+        sb.AppendRedacted(NullRedactor.Instance, null);
```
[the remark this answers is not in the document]
eerhardt: This is fine.

src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj
eerhardt: I've figured this out. Previously in .NET 9 SDK, we would use the NuGet package to reference/build, which has these settings inside. But in .NET 10 the NuGet package no longer gets used, so no one is setting this property.

https://github.com/dotnet/sdk/pull/50662 should fix this issue, but it wasn't in RC1. It should be in RC2.

I added a comment to the code.

src/Libraries/.editorconfig
eerhardt: I fixed the `IDE0040` warnings, since they were straight forward. But I logged a follow up issue https://github.com/dotnet/extensions/issues/6864 for `IDE0032` since it is not as straight-forward.

PR #6865, opened 2025-09-29
Tool grouping

MackinnonBuck:
## Summary
This PR demonstrates a prototype middleware that allows grouping related tools. Groups are hidden from the inner chat client by default, but a well-known "expand" function can be called to expand one group at a time. By reducing the number of tools available to the model at a time, its ability to select the correct tool improves.

Following is an overview of the changes in this PR:
- Add `ToolGroupingChatClient`, which provides tool grouping and expansion functionality.
- Introduce `AIToolGroup` to model stable, named batches of tools.
- Add integration tests validating the feasibility of approach.

## Usage example

```csharp

var travelUtilities = AIToolGroup.Create(
    "TravelUtilities",
    "Travel planning helpers that generate itinerary tokens.",
    [generateItinerary, summarizePacking]);

var financeUtilities = AIToolGroup.Create(
    name: "FinanceUtilities",
    description: "Budgeting helpers.",
    tools: [budgetTool]));

var options = new ChatOptions
{
    Tools = [travelUtilities, financeUtilities]
};

IChatClient client = baseClient
    .AsBuilder()
    .UseToolGrouping(/* ... */)
    .UseToolReduction(/* ... */) // Optionally further reduce the set of available tools
    .UseFunctionInvocation()
    .Build();

var response = await client.GetResponseAsync("Plan a weekend getaway to Lisbon and include a budget tip.");
// ...
```

## Implementation notes
This middleware contains a lot of the same complexities present in `FunctionInvokingChatClient`: cloning options, honoring `ConversationId`, aggregating `Usage` across inner calls, and replaying tool-result messages into the history. The expansion loop is deliberately limited to one active group per top-level request to prevent degradation of tool selection as the number of expanded groups increases.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6865)


src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
```diff
@@ -0,0 +1,423 @@
...
+        _toolGroupsByName = options.Groups.ToDictionary(g => g.Name, StringComparer.Ordinal);
+#if NET
+        _allGroupedTools = options.Groups.SelectMany(g => g.Tools).ToHashSet();
+#else
+        _allGroupedTools = [];
+        foreach (var group in options.Groups)
+        {
+            foreach (var tool in group.Tools)
+            {
+                _ = _allGroupedTools.Add(tool);
```
ericstj: So we only group tools known at initialization?  What about those added per call to GetResponse?  I could imagine some sort of callback to assign a group.  Or we could let folks annotate a tool with a group and store that in AdditionalProperties.

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
```diff
@@ -0,0 +1,423 @@
...
+#endif
+
+        var description = _options.ExpansionFunctionDescription ?? GetDefaultDescription(_options);
+
+        // We create a delegate with signature (string groupName) => string to leverage simple argument binding.
+        _expansionFunction = AIFunctionFactory.Create(
+            method: _expansionFunctionDelegate,
+            name: _options.ExpansionFunctionName,
+            description: description)
+            .AsDeclarationOnly();
```
ericstj: Interesting, so we create a signature that the model can call, but it does nothing.  We just observe that in the results when handling them from the inner client and use it to modify the tools we then send back to the client.

When an inner client gets a function invocation like this, does it know to halt communcation with the model and yield the result?

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
```diff
@@ -0,0 +1,423 @@
...
+                groupNameArg is null)
+            {
+                contents.Add(new FunctionResultContent(
+                    callId: expansionRequest.CallId,
+                    result: "No group name was specified; ignoring expansion request."));
+                continue;
+            }
+
+            var groupName = groupNameArg.ToString();
+            if (groupName is null || !_toolGroupsByName.TryGetValue(groupName, out var group))
```
ericstj: So we always return all Tools in a group, even if those tools were not part of the options for this request?   I can see how that might make sense - we have to choose a behavior - are the groupings additive to call, or does the call filter the groupings.  Given groupings are new and the call behavior exists, I had assumed the call would filter the groupings (but I wouldn't trust my perspective here 😄).

ericstj (commented):
This is very clever, I had a couple small questions on behavior.

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
ericstj: Another thought is auto-grouping based on similarity of descriptions / embedding.

MackinnonBuck: That's definitely something to consider. We could have a way to annotate a tool with a "category", similar to how tool descriptions are defined. Then chat clients like `ToolGroupingChatClient` can utilize the category to perform grouping on a per-response basis. This keeps the tool configuration completely in `ChatOptions` rather than having it be split between `ChatOptions` and other middleware.

Or maybe we could change the `AIToolGroup` type to extend `AITool`. That way you could put tool groups directly in the `ChatOptions`. You could even make `AIToolGroup` unsealed and have a virtual `GetToolsAsync()` method that can be overridden if you want to dynamically expose different sets of tools based on some other logic (this could be used to implement the "get more tools" function you mentioned in your [other comment](https://github.com/dotnet/extensions/pull/6781#discussion_r2396311340)). I'll think about this a bit more.

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
MackinnonBuck: Yep - `FuncitonInvokingChatClient` has some logic to detect if an `AITool` is not an `AIFunction`. Such tools are considered "non-invokable" and terminate the function calling loop to allow the tool to be handled by the caller:

https://github.com/dotnet/extensions/blob/d5cb8af941a5d7521d45b77a8ac626e33dfa66a8/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L828-L832

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
MackinnonBuck: Yeah, we could definitely treat the group as a "filter" more than an extra set of available tools if we wanted. Maybe this is another reason to define groups directly on `ChatOptions` rather than configuring them on the middleware.

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
ericstj: Interesting.  This is more complicated than the mental model I had.  I need to study FunctionInvokingClient.

MackinnonBuck:
Some notes:
* We probably need to update `FunctionInvokingChatClient` to handle tool groups (likely just expanding all the tools in all groups).
* Instead of `ToolGroupingChatClient`, we could consider calling it `ToolCollapsingChatClient` or `ToolExpandingChatClient`.
* We could also consider having some kind of automatic grouping mechanism so that the developer gets the benefits of tool groups without having to define them manually. This could be done as a follow-up.

src/Libraries/Microsoft.Extensions.AI/ToolGrouping/ToolGroupingChatClient.cs
```diff
@@ -7,24 +7,27 @@
...
+/// A chat client that enables tool groups (see <see cref="AIToolGroup"/>) to be dynamically expanded.
 /// </summary>
 /// <remarks>
 /// <para>
-/// On each request, the client initially presents a minimal tool surface consisting of: (a) tools in
-/// <see cref="ChatOptions.Tools"/> that are not members of any configured <see cref="AIToolGroup"/>
-/// plus (b) a synthetic expansion function. If the model calls the expansion function with a valid group name, the
+/// On each request, this chat client initially presents a minimal tool surface consisting of: (a) a function
+/// returning the current list of available groups plus (b) a synthetic expansion function plus (c) tools in
+/// <see cref="ChatOptions.Tools"/> that are not <see cref="AIToolGroup"/> instances.
```
ericstj, line 27: We might also want to note that the `FunctionInvokingChatClient` will add additional tools through it's `AdditionalTools` property.

We may want t support a similar property on this client as well as any associated API around it if folks want grouped `AdditionalTools`

stephentoub: FunctionInvokingChatClient.AdditionalTools doesn't contribute tools to ChatOptions.Tools. It's used by FICC only for the case where the inner client sends back a FunctionCallContent for a tool that _wasn't_ in ChatOptions.Tools, e.g. where a service was pre-configured to know about a particular function tool that the client isn't sending but that the model can still request.

MackinnonBuck: My understanding of `AdditionalTools` is that it actually won't change which options are sent through to the inner chat client via `ChatOptions.Tools`. It's present so that if a service has a set of built-in tools that don't need to be sent on every request, those can be handled by `FunctionInvokingChatClient`.

With that in mind, I'm not totally sure how helpful it would be to support selecting tool groups in `AdditionalProperties` for expansion. The groups would be invisible to the AI service, so it wouldn't know how to expand them, right?

I did just push an update that changes `FunctionInvokingChatClient` to expand any tool groups that it receives, including any present in `AdditionalProperties`. This allows tool groups to be used as a way to dynamically select which tools get exposed, even if `ToolGroupingChatClient` isn't used.

Let me know if I'm not understanding the suggestion! 🙂

ericstj: I see, I hadn't noticed that those wouldn't be sent to the chat client, however you're right.  I guess I don't fully understand the purpose of AdditionalTools if hte model isn't aware of them.  I had thought it was just a scoping thing - where tools could be specified on the instance or on the call, but that doesn't seem to be the case.  I agree you don't need to be worried about them if they never are exposed to the underlying client.

stephentoub: >  I guess I don't fully understand the purpose of AdditionalTools if hte model isn't aware of them.

It's for cases where the service is already aware of them, eg when you create an OpenAI Assistant and configure it with various tools. The service is aware of those tools and the model can request those tools, even though you haven't sent them as part of the request.

PederHP:
Being able to configure the Description of the tool passed to the model for the expansion / collapse of Tool groups is a must for this to be able to used in a lot of systems. It's a great pattern but having the Description hardcoded will be problematic in many cases. Different models need different Descriptions, and one can't run eval-optimize loops if it's hardcoded. Same with the name of the tool. 

A configuration delegate where can provide the Description and Name would be great. Otherwise, I think it loses some of the un-opinionated generic-ness which is one of the great things about MEAI. (I know one can just build a custom tool for this, but it's a small-ish change and makes this much more generic and customizable).

A default tool name/description isn't a problem, of course, Not saying one should be forced to provide these, just that it should be possible.

MackinnonBuck:
I appreciate all the feedback on this PR. We're going to close this out for now. The tool management space has seen rapid evolution, with fundamentally different approaches emerging (code-execution-based, sub-agent delegation, etc.). Many of the emerging approaches operate at the agent level rather than the chat client level, so it's unclear that `IChatClient` middleware is even the right layer for this. Either way, I don't think it makes sense to commit to a first-class MEAI abstraction in this PR. This can still be implemented as custom middleware by anyone who needs it today 🙂

stephentoub:
@MackinnonBuck does that mean you'll also be submitting a PR to remove IToolReductionStrategy, UseToolReduction, etc.?

PR #6866, opened 2025-09-30
Re-enable IDE0032

### LINKED ISSUE #6864: Reenable IDE0032 Use auto property
In the [upgrade to .NET 10 SDK](https://github.com/dotnet/extensions/pull/6863/), `IDE0032` is now firing to use the new `field` keyword and remove the backing field member.

Since this is a large change, I didn't include it in https://github.com/dotnet/extensions/pull/6863, but instead disabled the analyzer for now.

Once https://github.com/dotnet/extensions/pull/6863 is merged, we should re-enable this analyzer and fix the violations.

stephentoub:
Fixes #6864 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6866)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
@@ -20,15 +20,10 @@ internal sealed partial class ContentSafetyService(ContentSafetyServiceConfigura
     private const string APIVersionForServiceDiscoveryInHubBasedProjects = "?api-version=2023-08-01-preview";
     private const string APIVersionForNonHubBasedProjects = "?api-version=2025-05-15-preview";
 
-    private static HttpClient? _sharedHttpClient;
-    private static HttpClient SharedHttpClient
-    {
-        get
-        {
-            _sharedHttpClient ??= new HttpClient();
-            return _sharedHttpClient;
-        }
-    }
+    private static HttpClient SharedHttpClient =>
+        field ??
+        Interlocked.CompareExchange(ref field, new(), null) ??
+        field;
 
     private static readonly ConcurrentDictionary<UrlCacheKey, string> _serviceUrlCache =
         new ConcurrentDictionary<UrlCacheKey, string>();
--- a/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
@@ -44,9 +44,6 @@ public class DistributedCachingChatClient : CachingChatClient
     /// <summary>Additional values used to inform the cache key employed for storing state.</summary>
     private object[]? _cacheKeyAdditionalValues;
 
-    /// <summary>The <see cref="JsonSerializerOptions"/> to use when serializing cache data.</summary>
-    private JsonSerializerOptions _jsonSerializerOptions = AIJsonUtilities.DefaultOptions;
-
     /// <summary>Initializes a new instance of the <see cref="DistributedCachingChatClient"/> class.</summary>
     /// <param name="innerClient">The underlying <see cref="IChatClient"/>.</param>
     /// <param name="storage">An <see cref="IDistributedCache"/> instance that will be used as the backing store for the cache.</param>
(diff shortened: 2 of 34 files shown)


src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
```diff
@@ -20,15 +20,10 @@ internal sealed partial class ContentSafetyService(ContentSafetyServiceConfigura
...
-    {
-        get
-        {
-            _sharedHttpClient ??= new HttpClient();
-            return _sharedHttpClient;
-        }
-    }
+    private static HttpClient SharedHttpClient =>
+        field ??
+        Interlocked.CompareExchange(ref field, new(), null) ??
```
shyamnamboodiripad, line 25: Consider

```suggestion
        Interlocked.CompareExchange(ref field, new(), comparand: null) ??
```

or better yet (since I am always confused by which parameter means what for CompareExchange :)) -

```suggestion
        Interlocked.CompareExchange(location1: ref field, value: new(), comparand: null) ??
```

shyamnamboodiripad, line 25: Unless I'm mistaken, the reason to use `CompareExchange` here is to ensure that the property will always return the same `HttpClient` instance even if multiple threads happen to enter the getter concurrently. Is that correct?

In other words, regardless of whether or not we use `CompareExchange`, in concurrent situations, some of the constructed `HttpClient`s would end up being discarded. However, we want to ensure that we don't end up discarding any `HttpClient` that happened to be returned by this property and that ended up being used to execute one or more http requests. The `CompareExchange` call is basically there to ensure that.

Is this because once an `HttpClient` is used, it can hold on to resources that essentially end up being leaked if the object goes out of scope?

stephentoub: > the reason to use CompareExchange here is to ensure that the property will always return the same HttpClient instance even if multiple threads happen to enter the getter concurrently. Is that correct?

Correct.

> The CompareExchange call is basically there to ensure that.

Correct. Multiple instances could be created if there's a race, but only one will ever be published, and all consumers will always get the same instance.

> Is this because once an HttpClient is used, it can hold on to resources that essentially end up being leaked if the object goes out of scope?

Yes on the resources. The HttpClient wraps an HttpMessageHandler that holds onto the connection pool. If it's dropped, the resources won't be leaked per se, as finalizers will eventually kick in and tear things down, but it could take a while. And regardless, if two different consumers end up using two different instances, they won't be sharing a connection pool.

src/Libraries/Microsoft.Extensions.AI.Evaluation.Safety/ContentSafetyService.cs
```diff
@@ -20,15 +20,10 @@ internal sealed partial class ContentSafetyService(ContentSafetyServiceConfigura
...
+        field;
```
shyamnamboodiripad, line 26: Since `CompareExchange` always returns the original value (which would either be `null` or an existing valid `HttpClient`) can we simplify to use a single `??` check?

```suggestion
    private static HttpClient SharedHttpClient =>
        Interlocked.CompareExchange(ref field, new(), null) ??
        field;
```

shyamnamboodiripad: Ah never mindthat would result in wasteful  `HttpClient` instances...

shyamnamboodiripad: (The pattern with two `??` is interesting :) I don't think I have come across this before - I wish there were an easier way to express this - or if there were a `Interlocked.CompareExchangeAndReturnTheNewValueInsteadOfTheOriginalOne` 😄)

evgenyfedorov2 (approved):

src/Libraries/Microsoft.Extensions.AI/ChatCompletion/DistributedCachingChatClient.cs
```diff
@@ -59,9 +56,9 @@ public DistributedCachingChatClient(IChatClient innerClient, IDistributedCache s
     /// <summary>Gets or sets JSON serialization options to use when serializing cache data.</summary>
     public JsonSerializerOptions JsonSerializerOptions
     {
-        get => _jsonSerializerOptions;
-        set => _jsonSerializerOptions = Throw.IfNull(value);
-    }
+        get;
+        set => field = Throw.IfNull(value);
+    } = AIJsonUtilities.DefaultOptions;
```
eerhardt, line 61: Did you manually fix this one up? I was getting this on a separate line when using `dotnet format`.

eerhardt: Oh, I see it was because of `SA1500` and `SA1513` being enabled as well?

stephentoub: Yup. Those stylecop rules need to be updated to accomodate C# 14 features.

PR #6868, opened 2025-09-30
Move Microsoft.Extensions.ServiceDiscovery to dotnet/extensions

eerhardt:
I preserved git history using `git filter-repo`.

Then on top of that history, there is 1 commit for changing the libraries and tests to build cleanly. For now I disable any analyzer warnings that come up. https://github.com/dotnet/extensions/issues/6871 tracks enabling the analyzers in these libraries in the future.

Contributes to https://github.com/dotnet/aspire/issues/170.

**NOTE**: This PR shouldn't be "squashed" when merged. Instead it should be a merge commit.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6868)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Abstractions/ServiceEndpointQuery.cs
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Abstractions/ServiceEndpointQuery.cs
@@ -38,7 +38,7 @@ public sealed class ServiceEndpointQuery
         ArgumentException.ThrowIfNullOrEmpty(input);
 
         bool hasScheme;
-        if (!input.Contains("://", StringComparison.InvariantCulture)
+        if (!input.Contains("://", StringComparison.Ordinal)
             && Uri.TryCreate($"fakescheme://{input}", default, out var uri))
         {
             hasScheme = false;
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderFactory.cs
@@ -95,7 +95,7 @@ internal sealed partial class DnsSrvServiceEndpointProviderFactory(
         var lines = File.ReadAllLines(s_resolveConfPath);
         foreach (var line in lines)
         {
-            if (!line.StartsWith("search "))
+            if (!line.StartsWith("search ", StringComparison.Ordinal))
             {
                 continue;
             }
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Microsoft.Extensions.ServiceDiscovery.Dns.csproj
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Microsoft.Extensions.ServiceDiscovery.Dns.csproj
@@ -5,7 +5,7 @@
     <IsPackable>true</IsPackable>
     <Description>Provides extensions to HttpClient to resolve well-known hostnames to concrete endpoints based on DNS records. Useful for service resolution in orchestrators such as Kubernetes.</Description>
     <PackageIconFullPath>$(DefaultDotnetIconFullPath)</PackageIconFullPath>
-    <!-- TODO: log an issue to track enabling these -->
+    <!-- https://github.com/dotnet/extensions/issues/6871 tracks enabling these -->
     <NoWarn>$(NoWarn);IDE0018;IDE0025;IDE0032;IDE0040;IDE0058;IDE0250;IDE0251;IDE1006;CA1304;CA1307;CA1309;CA1310;CA1849;CA2000;CA2213;CA2217;S125;S1135;S1226;S2344;S3626;S4022;SA1108;SA1120;SA1128;SA1129;SA1204;SA1205;SA1214;SA1400;SA1405;SA1408;SA1515;SA1600;SA1629;SA1642;SA1649;EA0001;EA0009;EA0014;LA0001;LA0003;LA0008;VSTHRD200</NoWarn>
     <ImplicitUsings>enable</ImplicitUsings>
     <DisableMicrosoftExtensionsLoggingSourceGenerator>false</DisableMicrosoftExtensionsLoggingSourceGenerator>
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Abstractions/Microsoft.Extensions.ServiceDiscovery.Abstractions.csproj
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Abstractions/Microsoft.Extensions.ServiceDiscovery.Abstractions.csproj
@@ -6,7 +6,7 @@
     <Description>Provides abstractions for service discovery. Interfaces defined in this package are implemented in Microsoft.Extensions.ServiceDiscovery and other service discovery packages.</Description>
     <PackageIconFullPath>$(DefaultDotnetIconFullPath)</PackageIconFullPath>
     <RootNamespace>Microsoft.Extensions.ServiceDiscovery</RootNamespace>
-    <!-- TODO: log an issue to track enabling these -->
+    <!-- https://github.com/dotnet/extensions/issues/6871 tracks enabling these -->
     <NoWarn>$(NoWarn);S1144;CA1002;S2365;SA1642;IDE0040;CA1307;EA0009;LA0003</NoWarn>
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp/Microsoft.Extensions.ServiceDiscovery.Yarp.csproj
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp/Microsoft.Extensions.ServiceDiscovery.Yarp.csproj
@@ -7,7 +7,7 @@
     <IsPackable>true</IsPackable>
     <Description>Provides extensions for service discovery for the YARP reverse proxy.</Description>
     <PackageIconFullPath>$(DefaultDotnetIconFullPath)</PackageIconFullPath>
-    <!-- TODO: log an issue to track enabling these -->
+    <!-- https://github.com/dotnet/extensions/issues/6871 tracks enabling these -->
     <NoWarn>$(NoWarn);IDE0018;IDE0025;IDE0032;IDE0040;IDE0058;IDE0250;IDE0251;IDE1006;CA1304;CA1307;CA1309;CA1310;CA1849;CA2000;CA2213;CA2217;S125;S1135;S1226;S2344;S2692;S3626;S4022;SA1108;SA1120;SA1128;SA1129;SA1204;SA1205;SA1214;SA1400;SA1405;SA1408;SA1414;SA1515;SA1600;SA1615;SA1629;SA1642;SA1649;EA0001;EA0009;EA0014;LA0001;LA0003;LA0008;VSTHRD200</NoWarn>
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery/Microsoft.Extensions.ServiceDiscovery.csproj
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery/Microsoft.Extensions.ServiceDiscovery.csproj
@@ -5,7 +5,7 @@
     <IsPackable>true</IsPackable>
     <Description>Provides extensions to HttpClient that enable service discovery based on configuration.</Description>
     <PackageIconFullPath>$(DefaultDotnetIconFullPath)</PackageIconFullPath>
-    <!-- TODO: log an issue to track enabling these -->
+    <!-- https://github.com/dotnet/extensions/issues/6871 tracks enabling these -->
     <NoWarn>$(NoWarn);CS8600;CS8602;CS8604;IDE0040;IDE0055;IDE0058;IDE1006;CA1307;CA1310;CA1849;CA2007;CA2213;SA1204;SA1128;SA1205;SA1405;SA1612;SA1623;SA1625;SA1642;S1144;S1449;S2302;S2692;S3872;S4457;EA0000;EA0009;EA0014;LA0001;LA0003;LA0008;VSTHRD200</NoWarn>
     <ImplicitUsings>enable</ImplicitUsings>
     <DisableMicrosoftExtensionsLoggingSourceGenerator>false</DisableMicrosoftExtensionsLoggingSourceGenerator>
--- a/src/Shared/FxPolyfills/FxPolyfills.targets
+++ b/src/Shared/FxPolyfills/FxPolyfills.targets
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <SharedFxPolyfillsDir>$(MSBuildThisFileDirectory)</SharedFxPolyfillsDir>
-    <!-- TODO: log an issue to track enabling these -->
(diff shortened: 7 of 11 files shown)


src/Libraries/Microsoft.Extensions.ServiceDiscovery.Abstractions/ServiceEndpointQuery.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// </summary>
+    /// <param name="input">The value to parse.</param>
+    /// <param name="query">The resulting query.</param>
+    /// <returns><see langword="true"/> if the value was successfully parsed; otherwise <see langword="false"/>.</returns>
+    public static bool TryParse(string input, [NotNullWhen(true)] out ServiceEndpointQuery? query)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(input);
+
+        bool hasScheme;
+        if (!input.Contains("://", StringComparison.InvariantCulture)
```
stephentoub: StringComparison.Ordinal?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsServiceEndpointProviderBase.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                validityPeriod = _nextRefreshPeriod;
+            }
+            else if (validityPeriod > _nextRefreshPeriod)
+            {
+                validityPeriod = _nextRefreshPeriod;
+            }
+
+            _lastCollectionCancellation.Cancel();
+            var cancellation = _lastCollectionCancellation = new CancellationTokenSource(validityPeriod, _timeProvider);
```
stephentoub, line 132: Where does this get disposed?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderFactory.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            return default;
+        }
+
+        // See https://manpages.debian.org/bookworm/manpages/resolv.conf.5.en.html#search for the format of /etc/resolv.conf's search option.
+        // In our case, we are interested in determining the domain name.
+        var lines = File.ReadAllLines(s_resolveConfPath);
+        foreach (var line in lines)
+        {
+            if (!line.StartsWith("search "))
```
stephentoub: This is going to end up being culture-aware; OrdinalIgnoreCase?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/FallbackDnsResolver.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            addresses.Add(new AddressResult
+            {
+                Address = record.Address,
+                ExpiresAt = now.Add(TimeSpan.FromSeconds(record.TimeToLive))
+            });
+        }
+
+        var srvRecords = queryResult.Answers.OfType<SrvRecord>().ToList();
```
stephentoub, line 70: I'm curious when we're choosing to use and not use LINQ, e.g. here this is using OfType.ToList, but it could have instead been OfType.Select.ToArray to encompass the below array allocation and loop.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
```diff
@@ -0,0 +1,931 @@
...
+        var ipv4AddressesTask = ResolveIPAddressesAsync(name, AddressFamily.InterNetwork, cancellationToken);
+        var ipv6AddressesTask = ResolveIPAddressesAsync(name, AddressFamily.InterNetworkV6, cancellationToken);
+
+        AddressResult[] ipv4Addresses = await ipv4AddressesTask.ConfigureAwait(false);
+        AddressResult[] ipv6Addresses = await ipv6AddressesTask.ConfigureAwait(false);
+
+        AddressResult[] results = new AddressResult[ipv4Addresses.Length + ipv6Addresses.Length];
+        ipv6Addresses.CopyTo(results, 0);
+        ipv4Addresses.CopyTo(results, ipv6Addresses.Length);
+        return results;
```
stephentoub, line 143:
```csharp
return [
    .. await ResolveIPAddressesAsync(name, AddressFamily.InterNetwork, cancellationToken).ConfigureAwait(false),
    .. await ResolveIPAddressesAsync(name, AddressFamily.InterNetworkV6, cancellationToken).ConfigureAwait(false),
];
```
?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp/README.md
```diff
@@ -0,0 +1,42 @@
+# Microsoft.Extensions.ServiceDiscovery.Yarp
```
stephentoub, line 1: We're adding YARP-specific libraries to dotnet/extensions? That wouldn't be better in the YARP repo?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Microsoft.Extensions.ServiceDiscovery.Dns.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(NetCoreTargetFrameworks)</TargetFrameworks>
```
stephentoub, line 4: The .Abstractions library targets netstandard2.0, but the rest of them target core only?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Microsoft.Extensions.ServiceDiscovery.Dns.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <IsPackable>true</IsPackable>
+    <Description>Provides extensions to HttpClient to resolve well-known hostnames to concrete endpoints based on DNS records. Useful for service resolution in orchestrators such as Kubernetes.</Description>
+    <PackageIconFullPath>$(DefaultDotnetIconFullPath)</PackageIconFullPath>
+    <!-- TODO: log an issue to track enabling these -->
```
stephentoub: What's the plan for tackling that?

stephentoub:
cc: @karelz

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Microsoft.Extensions.ServiceDiscovery.Dns.csproj
eerhardt: https://github.com/dotnet/extensions/issues/6871. Will push an update to the comments pointing to this issue.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Microsoft.Extensions.ServiceDiscovery.Dns.csproj
eerhardt: - Microsoft.Extensions.ServiceDiscovery.Abstractions
- Microsoft.Extensions.ServiceDiscovery

These 2 target `netstandard2.0`. The `Dns` and `Yarp` libraries only target core.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp/README.md
eerhardt: Maybe? I'm not sure how we see the layering - if Yarp should "depend" on dotnet/extensions (for ServiceDiscovery base libraries) or dotnet/extensions should "depend" on Yarp.

cc @adityamandaleeka @danmoseley @joperezr

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderFactory.cs
eerhardt: Fixed with Ordinal. It wasn't using IgnoreCase before, and as far as I can tell from the [man page](https://manpages.debian.org/bullseye/manpages/resolv.conf.5.en.html) it is case sensitive.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Abstractions/ServiceEndpointQuery.cs
eerhardt: Fixed.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
eerhardt: I've opened [Add more tests to Microsoft.Extensions.ServiceDiscovery.Dns and clean up code (dotnet/extensions#6872)](https://github.com/dotnet/extensions/issues/6872) to make these suggestions. I don't want to change the code in this PR since this is just porting the existing logic, and it appears this method doesn't have code coverage. I changed it to unconditionally throw and no tests failed.

(Note your current suggestion changes behavior from what I can tell - they run in serial instead of concurrently and the IPv4 goes first instead of the IPv6.)

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/FallbackDnsResolver.cs
eerhardt: I've opened https://github.com/dotnet/extensions/issues/6872 to make these suggestions. I don't want to change the code in this PR since this is just porting the existing logic, and it appears this method doesn't have code coverage. I changed it to unconditionally throw and no tests failed.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsServiceEndpointProviderBase.cs
eerhardt: I've opened https://github.com/dotnet/extensions/issues/6872 to make these suggestions. I don't want to change the code in this PR since this is just porting the existing logic.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
stephentoub: > Note your current suggestion changes behavior from what I can tell - they run in serial instead of concurrently and the IPv4 goes first instead of the IPv6

That's true, though that also highlights an issue with the current code, which is that if the IPv4 call fails asynchronously, the code will never join with the IPv6 call.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp/README.md
joperezr: @eerhardt and I chatted a bit about this offline. I think both extensions and yarp are on the "same layer" of the stack, and there is value in keeping all of the ServiceDiscovery source together (similar to how we think there is value in keeping all of the MEAI code together). It will also allow things to version together, and for aspire to consume from a single place. I suggest we keep this here in extensions for now, and we can re-evaluate in the future if we want to move some of these elsewhere.

eng/packages/General-LTS.props
```diff
@@ -4,6 +4,7 @@
   of the framework, we should use the following LTS versions instead -->
   <ItemGroup Condition="'$(TargetFramework)' != 'net9.0' And '$(ForceLatestDotnetVersions)' != 'true'">
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftBclAsyncInterfacesLTSVersion)" />
+    <PackageVersion Include="Microsoft.Bcl.Memory" Version="$(MicrosoftBclMemoryVersion)" /> <!-- Note there is no 8.0 version of Microsoft.Bcl.Memory -->
```
joperezr, line 7: I'm almost wondering if we should still just create an LTS property that also points to 9.0.9 just so we don't have to special case when updating. Anyway, feel free to leave as is, just wanted to think out loud.

src/Libraries/Microsoft.Extensions.ServiceDiscovery/Configuration/ConfigurationServiceEndpointProvider.Log.cs
```diff
@@ -0,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.ServiceDiscovery.Configuration;
+
+internal sealed partial class ConfigurationServiceEndpointProvider
```
joperezr, line 9: Interesting, looks like this was added as part of the migration. Out of curiosity, how come we didn't have this over in dotnet/aspire? Where we not using the source generator?

joperezr: Oh, nvm, I'm looking at your commit wrong.

joperezr (approved):

evgenyfedorov2 (approved):
Approving on behalf of dotnet-extensions-fundamentals

eerhardt:
> What do we want to do about the versioning of these packages? Currently the version in NuGet.org is 9.5, and this will start producing 9.10. Is that fine? I'm assuming so and we want these packages to align versioning with the rest of the extensions repo but asking just in case.

This is what I was planning on doing. If Aspire can jump from `9.5` to `13.0`, I think it will be OK if the ServiceDiscovery libraries jump from `9.5` to `9.10`, and then continue using dotnet/extensions versioning going forward.

eng/packages/General-LTS.props
eerhardt: I think we should leave it as-is. Calling it the "LTS version" is misleading if there is no 8.0 version.

> that also points to 9.0.9 just so we don't have to special case when updating.

I think doing that would make it special cased. As it is right now, the `9.0.x` version will automatically be updated, and both PackageVersions will reflect the new value.

eerhardt:
@joperezr - I believe this is ready to merge. However, it shouldn't be merged with a "Squash". Instead we should "Merge commit" so the history of the libraries are maintained. Can you enable this option?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
eerhardt: > which is that if the IPv4 call fails asynchronously, the code will never join with the IPv6 call.

I'm not sure I understand the problem. Can you explain? If the IPv4 call throws in the await on line 137, the IPv6 call will run to completion or failure, and just be ignored, right?

stephentoub: It's the "just be ignored" that's the problem. If you have:
```csharp
Task a = DoWhateverA();
Task b = DoWhateverB();
await a;
await b;
```
and `a` fails, you'll never join with `b`. That means it can continue doing whatever work it was doing, accessing whatever resources it was accessing, mutating whatever state it was mutating, and so on, long after this code has moved on. In some cases, that's ignorable. In others, it can lead to really subtle bugs. And for folks that pay attention to faulted tasks that never have their exceptions observed, if `b` also faults, it'll fall into that group.

Issue #6779, opened 2025-09-08
Follow-ups for MCP tool


stephentoub:
https://github.com/dotnet/extensions/pull/6664 adds the MCP tool and supporting types as experimental. Some things to follow-up on before marking them stable:
- Do we need the base "user input" / "user output" types?
- The OpenAI Responses implementation currently looks for the tool approval as part of a Tool message. Is that right? Should it be User instead? Should it be any kind of message?
- Responses exposes a single MCP tool call message / instance, rather than a separate one for call and a separate one for result. We instead currently model it as two types, one for request and one for result. Is that the right split?
- The naming of "id" parameters is confusing; it's not clear which ID they're referring to, and more than once I passed in the wrong ID. We should revisit the naming.
- We need to ensure the hosted MCP tool works with variations, like OpenAI's connectors. This might entail replacing the Url property with something more general, possibly just a rename.

cc: @jozkee 

[David] Adding another point to above list:
- In McpServerToolCallContent, it is too harsh to require Server name since it is only useful to disambiguate in cases of tool name collisions.  We need to revisit which properties are required, at the bare minimum we should keep CallId.

jozkee:
> Do we need the base "user input" / "user output" types?

We briefly discussed having a common abstraction between `FunctionApproval*Content` and `McpServerToolApproval*Content` e.g.
```cs
public abstract class ApprovalRequestContent : UserInputRequestContent
{
    protected ApprovalRequestContent(string id) : base(id) { }

    public abstract ApprovalResponseContent CreateResponse(bool approved);
}

FunctionApprovalRequestContent : ApprovalRequestContent
McpServerToolApprovalRequestContent : ApprovalRequestContent
```
This way, you wouldn't have to write code to handle both types, just `ApprovalRequestContent`.
A problem I see with that is covariance for `CreateResponse` will not work for ns2.0 and netfx.

jozkee:
> The OpenAI Responses implementation currently looks for the tool approval as part of a Tool message. Is that right? Should it be User instead? Should it be any kind of message?

I think handling it for any kind of ChatRole would be the best.  @westey-m, why did we go with `FunctionApprovalRequestContent CreateResponse(bool)` and not `ChatMessage CreateResponse(bool)` so we could decide which role was best for an approval response?

westey-m:
[...]

PR #6881, opened 2025-10-02
Add support for Connector ID

jozkee:
Contributes to https://github.com/dotnet/extensions/issues/6779

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
@@ -37,6 +37,30 @@ public class HostedMcpServerTool : AITool
     {
         ServerName = Throw.IfNullOrWhitespace(serverName);
         Url = Throw.IfNull(url);
+        ConnectorID = null;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerTool"/> class with a connector ID.
+    /// </summary>
+    private HostedMcpServerTool(string serverName, string connectorID, Uri? url)
+    {
+        ServerName = Throw.IfNullOrWhitespace(serverName);
+        ConnectorID = Throw.IfNullOrWhitespace(connectorID);
+        Url = url; // needed for disambiguation.
+    }
+
+    /// <summary>
+    /// Creates a new instance of the <see cref="HostedMcpServerTool"/> class using a connector ID.
+    /// </summary>
+    /// <param name="serverName">The name of the remote MCP server.</param>
+    /// <param name="connectorID">The connector ID of the known MCP server supported by the provider.</param>
+    /// <returns>A new <see cref="HostedMcpServerTool"/> instance configured with the specified connector ID.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="serverName"/> or <paramref name="connectorID"/> are <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="serverName"/> or <paramref name="connectorID"/> are empty or composed entirely of whitespace.</exception>
+    public static HostedMcpServerTool CreateWithConnectorID(string serverName, string connectorID)
+    {
+        return new HostedMcpServerTool(serverName, connectorID, null);
     }
 
     /// <summary>
@@ -47,7 +71,18 @@ public class HostedMcpServerTool : AITool
     /// <summary>
     /// Gets the URL of the remote MCP server.
     /// </summary>
-    public Uri Url { get; }
+    /// <remarks>
+    /// This property is <see langword="null"/> when the tool is configured with a <see cref="ConnectorID" /> instead of a URL.
+    /// </remarks>
+    public Uri? Url { get; }
+
+    /// <summary>
+    /// Gets the connector ID of the known MCP server supported by the provider.
+    /// </summary>
+    /// <remarks>
+    /// This property is <see langword="null"/> when the tool is configured with a <see cref="Url" /> instead of a connector ID.
+    /// </remarks>
+    public string? ConnectorID { get; }
 
     /// <summary>
     /// Gets or sets the description of the remote MCP server, used to provide more context to the AI service.
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
@@ -5,6 +5,7 @@ using System;
 using System.ClientModel;
 using System.ClientModel.Primitives;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
@@ -467,11 +468,8 @@ internal sealed class OpenAIResponsesChatClient : IChatClient
                         break;
 
                     case HostedMcpServerTool mcpTool:
-                        McpTool responsesMcpTool = ResponseTool.CreateMcpTool(
-                            mcpTool.ServerName,
-                            mcpTool.Url,
-                            serverDescription: mcpTool.ServerDescription,
-                            headers: mcpTool.Headers);
+
+                        McpTool responsesMcpTool = ToOpenAIMcpTool(mcpTool);
 
                         if (mcpTool.AllowedTools is not null)
                         {
@@ -860,6 +858,28 @@ internal sealed class OpenAIResponsesChatClient : IChatClient
         }
 
         return parts;
+    }
+
+    private static McpTool ToOpenAIMcpTool(HostedMcpServerTool meaiMcpTool)
+    {
+        Debug.Assert(meaiMcpTool.Url is null != meaiMcpTool.ConnectorID is null, "Either Url or ConnectorID must be present but never both.");
+
+        if (meaiMcpTool.Url != null)
+        {
+            return ResponseTool.CreateMcpTool(
+                meaiMcpTool.ServerName,
+                meaiMcpTool.Url,
+                serverDescription: meaiMcpTool.ServerDescription,
+                headers: meaiMcpTool.Headers);
+        }
+        else
+        {
+            return ResponseTool.CreateMcpTool(
+                meaiMcpTool.ServerName,
+                new McpToolConnectorId(meaiMcpTool.ConnectorID),
+                serverDescription: meaiMcpTool.ServerDescription,
+                headers: meaiMcpTool.Headers);
+        }
     }
 
     /// <summary>Adds new <see cref="AIContent"/> for the specified <paramref name="mtci"/> into <paramref name="contents"/>.</summary>
--- a/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Tools/HostedMcpServerToolTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/Tools/HostedMcpServerToolTests.cs
@@ -17,7 +17,8 @@ public class HostedMcpServerToolTests
         Assert.Empty(tool.AdditionalProperties);
 
         Assert.Equal("serverName", tool.ServerName);
-        Assert.Equal("https://localhost/", tool.Url.ToString());
+        Assert.Equal("https://localhost/", tool.Url!.ToString());
+        Assert.Null(tool.ConnectorID);
 
         Assert.Empty(tool.Description);
         Assert.Null(tool.AllowedTools);
@@ -34,7 +35,8 @@ public class HostedMcpServerToolTests
         Assert.Equal(nameof(HostedMcpServerTool), tool.Name);
 
         Assert.Equal("serverName", tool.ServerName);
-        Assert.Equal("https://localhost/", tool.Url.ToString());
+        Assert.Equal("https://localhost/", tool.Url!.ToString());
+        Assert.Null(tool.ConnectorID);
         Assert.Empty(tool.Description);
 
         Assert.Null(tool.ServerDescription);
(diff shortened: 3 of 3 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -37,6 +37,30 @@ public HostedMcpServerTool(string serverName, Uri url)
...
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HostedMcpServerTool"/> class with a connector ID.
+    /// </summary>
+    private HostedMcpServerTool(string serverName, string connectorID, Uri? url)
+    {
+        ServerName = Throw.IfNullOrWhitespace(serverName);
+        ConnectorID = Throw.IfNullOrWhitespace(connectorID);
+        Url = url; // needed for disambiguation.
```
stephentoub: Connector ID seems very specific to OpenAI?

With their connectors, do you specify both a name and a URL? I thought they'd be mutually exclusive, in which case I was expecting we'd just have a single "location" (pick a better word) that's either a url or an identifier. That also means it doesn't make it specific to OpenAI.

jozkee: >  I thought they'd be mutually exclusive,

They are mutually exclusive.  This argument is just here for disambiguation because there's already a `.ctor(string, string)`.


> we'd just have a single "location" (pick a better word) that's either a url or an identifier

Do you mean having a type like `struct ServerLocation` with mutually exclusive properties Url and ConnectorID(for the lack of a better name)?  If so, that's sounds fine too, just wanted to make sure that we don't want to follow the simple route of having a factory method with a string connectorId

stephentoub: I meant just having it be a string, and having the implementation decide whether it's a name or a url, e.g. does it start with a schema. Would that work? Is it problematic?

> They are mutually exclusive. This argument is just here for disambiguation because there's already a .ctor(string, string).

We should not do that. Having a constructor that accepts both makes it look like both can be provided and utilized.

jozkee: > I meant just having it be a string, and having the implementation decide whether it's a name or a url

Ah so you mean updating the existing .ctor(string, string url) and rename the param to urlOrConnectorId and then have something like
```cs
if (Uri.TryCreate(urlOrConnectorId, UriKind.Absolute, out url))
    Url = url
else
    ConnectorID = urlOrConnectorId
```

My only concern with that is that we will be overloading the argument.

> We should not do that. Having a constructor that accepts both makes it look like both can be provided and utilized.

I thought it was fine since it was private, but OK, will take note.

stephentoub: > Ah so you mean updating the existing .ctor(string, string url) and rename the param

Yes, with some name that conveys "this is the thing to connect to, which could be a url or could be a descriptor that the backend service understands"

> I thought it was fine since it was private

Ah, I didn't see that it was private.

stephentoub: Do other services that support remote MCP, like Anthropic, have such a concept yet?

jozkee: > like Anthropic, have such a concept yet?

No, I know anthropic doesn't from looking at their docs. I'm not aware of other providers that support remote MCP.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -47,7 +71,18 @@ public HostedMcpServerTool(string serverName, Uri url)
     /// <summary>
     /// Gets the URL of the remote MCP server.
     /// </summary>
-    public Uri Url { get; }
+    /// <remarks>
+    /// This property is <see langword="null"/> when the tool is configured with a <see cref="ConnectorID" /> instead of a URL.
+    /// </remarks>
+    public Uri? Url { get; }
```
jozkee: @stephentoub sorry to bring this back to you.

> Would that work? Is it problematic? 

Supporting Connector ID as you proposed will be problematic for this existing Uri prop.  We will need to replace it with a string (and rename it) so it can hold both url and connector ID.  We also have the `.ctor(string, Uri)`, that will need to go away too.

I think I picked `Uri` because that's what the few existing url properties used, see [query](https://github.com/search?q=repo%3Adotnet%2Fextensions+%22public+Uri%22+path%3Asrc%2FLibraries%2F*AI*%2F**%2F*&type=code).

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/McpServerToolCallContent.cs
```diff
@@ -39,14 +35,14 @@ public McpServerToolCallContent(string callId, string toolName, string serverNam
     public string CallId { get; }
 
     /// <summary>
-    /// Gets the name of the tool called.
+    /// Gets or sets the name of the tool called.
     /// </summary>
-    public string ToolName { get; }
+    public string? ToolName { get; set; }
```
stephentoub: I'm still not convinced this makes sense. A call without any information about what's being called is useless. ToolName is what's being called. In contrast, ServerName is valuable as a disambiguator, but in general it's just informative.

If you _really_ believe it's all or nothing, that both ToolName and ServerName need to be required or neither need to be required, then we should just leave it as is rather than have neither be required.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -18,25 +18,13 @@ public class HostedMcpServerTool : AITool
...
-
-    /// <summary>
-    /// Initializes a new instance of the <see cref="HostedMcpServerTool"/> class.
-    /// </summary>
-    /// <param name="serverName">The name of the remote MCP server.</param>
-    /// <param name="url">The URL of the remote MCP server.</param>
-    /// <exception cref="ArgumentNullException"><paramref name="serverName"/> or <paramref name="url"/> is <see langword="null"/>.</exception>
-    /// <exception cref="ArgumentException"><paramref name="serverName"/> is empty or composed entirely of whitespace.</exception>
-    public HostedMcpServerTool(string serverName, Uri url)
+    /// <param name="serverAddress">The address of the remote MCP server.</param>
```
stephentoub:
```suggestion
    /// <param name="serverAddress">The address of the remote MCP server. This may be a URL, or in the case of a service providing built-in MCP servers with known names, it can be such a name.</param>
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -45,9 +33,14 @@ public HostedMcpServerTool(string serverName, Uri url)
...
-    /// Gets the URL of the remote MCP server.
+    /// Gets the address of the remote MCP server.
+    /// </summary>
+    public string ServerAddress { get; }
+
+    /// <summary>
+    /// Gets or sets the OAuth authorization token that the AI service should use when calling the remote MCP server.
     /// </summary>
-    public Uri Url { get; }
+    public string? AuthorizationToken { get; set; }
```
stephentoub, line 46: What is the relationship between this and Headers?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -591,6 +603,15 @@ internal static IEnumerable<ResponseItem> ToOpenAIResponseItems(IEnumerable<Chat
...
             {
                 yield return ResponseItem.CreateUserMessageItem(ToResponseContentParts(input.Contents));
+
+                foreach (AIContent item in input.Contents)
+                {
+                    if (item is McpServerToolApprovalResponseContent mcpApprovalResponseContent)
+                    {
+                        yield return ResponseItem.CreateMcpApprovalResponseItem(mcpApprovalResponseContent.Id, mcpApprovalResponseContent.Approved);
+                    }
+                }
```
stephentoub, line 705: Were MCP approvals just broken before?

stephentoub: Oh, I see, this is being done to accomodate the approval being in a user message in addition to being in a tools message, where it's currently handled?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
```diff
@@ -591,6 +603,15 @@ internal static IEnumerable<ResponseItem> ToOpenAIResponseItems(IEnumerable<Chat
             if (input.Role == ChatRole.User)
             {
                 yield return ResponseItem.CreateUserMessageItem(ToResponseContentParts(input.Contents));
+
+                foreach (AIContent item in input.Contents)
```
stephentoub, line 668: If the user message exists just to carry MCP approval, won't the above end up producing a separate empty user message item (with a dummy content part)?

Seems like it'd be better to restructure this, where this becomes:
yield return CreateUserResponseItems()
(or just do it inline as it's done for other kinds), iterating over the contents once rather than twice, and yielding the appropriate response items.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
jozkee: For ChatRole.User, yes.  This change is in response to your point in https://github.com/dotnet/extensions/issues/6779.  

> The OpenAI Responses implementation currently looks for the tool approval as part of a Tool message. Is that right? Should it be User instead? Should it be any kind of message?

After looking into what kind of messages the other roles process, I think it doesn't make sense to handle mcp approvals for system/developer/assistant.  However, worth mentioning that I think Function Approvals are handled for any role.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
jozkee: Yes, it does create that dummy message you describe, I left it like that because that's the case for any other kind of content that doesn't map to a `ResponseContentPart` but it does seem like something that should be fixed. Will do.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
jozkee: > to accomodate the approval being in a user message in addition to being in a tools message, where it's currently handled?

Yes.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
stephentoub: >  that's the case for any other kind of content that doesn't map to a ResponseContentPart

Not exactly. Someone would need to create an empty ChatMessage to hit that. With this, they've created a ChatMessage that contains just the McpApproval, which would be a very common thing to do (just as it is for creating tool messages just with function approval), yet we're then treating it as an empty message.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
stephentoub: > However, worth mentioning that I think Function Approvals are handled for any role.

Where? I see handling for it under `case ChatRole.Tool`.

jozkee: In FunctionInvokingChatClient, I don't think there's anything preventing you from passing the FunctionApprovalResponseContent in a message with any role, not just user/tool.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: I expect that this should serve the same purpose as `Headers.Add("Authorization", "<token>")`.  However, I tested with the connector and endpoint returns `Cannot specify 'headers' parameter with 'connector_id'.`.  A bit unfortunate that OpenAI's `ResponseTool.CreateMcpTool` accepts them.

stephentoub: We need to choose the right shared representation, and then each implementation maps that apppropriately. I  don't understand the argument for having both, if auth is representable as a header. You couldn't extract the auth token from the headers collection of just that collection existed, removing it from the collection passed to the underlying impl? 

Is the idea that you always use this property for an auth token and the headers collection for everything else?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
stephentoub: Ah, I see. @westey-m, is that desirable?

westey-m: I typically prefer to be stricter, since it helps to catch bugs.  E.g. passing FARC in anything but a user message sounds like a bug to me.  It is easy to relax the restriction later if we find a real scenario for it.
So I think tightening this in FICC as well makes sense.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: > Is the idea that you always use this property for an auth token and the headers collection for everything else?

Yes.  OpenAI started with only Headers and now it promoted the auth header to be its own field in the payload.

I think 99% use cases will be for the authorization header.  Having only Headers is more future proof but is less straightforward.  Anthropic also only accepts the auth_token.   It sucks to not have a crystal ball.

We can switch to only have AuthorizationToken.  If someone wants to pass headers in OpenAI they can pass the raw tool.   And if Headers becomes a thing in the future because more providers start adding them, we can bring it back.  Wdyt?

stephentoub: > We can switch to only have AuthorizationToken. [...] And if Headers becomes a thing in the future because more providers start adding them, we can bring it back. Wdyt?

Do we have any examples where headers are required? @halter73, any opinions?

Otherwise, sounds good.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/McpServerToolCallContent.cs
jozkee: I'm OK with sacrificing some consistency for usability.  Although, this "requirement" won't impede anyone from just passing a dummy string to bypass it as you've already done so before.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
jozkee: I don't see mcp server code handling headers in https://github.com/modelcontextprotocol/csharp-sdk/tree/main/samples.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/McpServerToolCallContent.cs
```diff
@@ -23,14 +23,12 @@ public sealed class McpServerToolCallContent : AIContent
     /// </summary>
     /// <param name="callId">The tool call ID.</param>
     /// <param name="toolName">The tool name.</param>
-    /// <param name="serverName">The MCP server name.</param>
-    /// <exception cref="ArgumentNullException"><paramref name="callId"/>, <paramref name="toolName"/>, or <paramref name="serverName"/> is <see langword="null"/>.</exception>
-    /// <exception cref="ArgumentException"><paramref name="callId"/>, <paramref name="toolName"/>, or <paramref name="serverName"/> are empty or composed entirely of whitespace.</exception>
-    public McpServerToolCallContent(string callId, string toolName, string serverName)
+    /// <exception cref="ArgumentNullException"><paramref name="callId"/> or <paramref name="toolName"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException"><paramref name="callId"/> or <paramref name="toolName"/> is empty or composed entirely of whitespace.</exception>
+    public McpServerToolCallContent(string callId, string toolName)
```
stephentoub: Another middleground option would be to keep what you had, where this accepts the serverName and the property is just a getter, but make it nullable and don't validate that it's not null or whitespace. That would also help to minimize changes for anyone currently depending on the constructor.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Tools/HostedMcpServerTool.cs
```diff
@@ -48,9 +36,14 @@ public HostedMcpServerTool(string serverName, Uri url)
     public string ServerName { get; }
 
     /// <summary>
-    /// Gets the URL of the remote MCP server.
+    /// Gets the address of the remote MCP server.
```
stephentoub: The comment should elaborate on what this means (in a service-agnostic way).

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs
stephentoub: @jozkee, can you make sure we have any follow-ups here tracked in an issue?

PR #6891, opened 2025-10-05
Fix Assistants IChatClient handling of unrelated tool calls in history

stephentoub:
The implementation is assuming that any tool results in the incoming message history mean a thread ID is required, which would be necessary if those tool results were for an active run. But that's not the case if the history just contains historical function calls/responses, as is the case in agent scenarios where the message history from one agent is passed to another.  The fix is to just stop throwing based on this incorrect assumption.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6891)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
@@ -80,14 +80,10 @@ internal sealed class OpenAIAssistantsChatClient : IChatClient
 
         // Get the thread ID.
         string? threadId = options?.ConversationId ?? _defaultThreadId;
-        if (threadId is null && toolResults is not null)
-        {
-            Throw.ArgumentException(nameof(messages), "No thread ID was provided, but chat messages includes tool results.");
-        }
 
         // Get any active run ID for this thread. This is necessary in case a thread has been left with an
-        // active run, in which all attempts other than submitting tools will fail. We thus need to cancel
-        // any active run on the thread.
+        // active run, in which case all attempts other than submitting tools will fail. We thus need to cancel
+        // any active run on the thread if we're not submitting tool results to it.
         ThreadRun? threadRun = null;
         if (threadId is not null)
         {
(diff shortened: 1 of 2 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
```diff
@@ -80,14 +80,10 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
 
         // Get the thread ID.
         string? threadId = options?.ConversationId ?? _defaultThreadId;
-        if (threadId is null && toolResults is not null)
```
eiriktsarpalis, line 83: Is this change related to the fix?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
```diff
@@ -487,7 +483,7 @@ void AppendSystemInstructions(string? toAppend)
                         messageContents.Add(MessageContent.FromImageUri(image.Uri));
                         break;
 
-                    case FunctionResultContent result:
+                    case FunctionResultContent result when chatMessage.Role == ChatRole.Tool:
```
eiriktsarpalis, line 486: Should we add a test validating the fix?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
stephentoub: This change is the fix. The other line is housekeeping.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
stephentoub: We don't have good tests for assistants because of the multi-request nature making it a lot harder to mock well.

eiriktsarpalis (approved):

Issue #6743, opened 2025-08-25
Adding Tools to the Assistant ChatClient OpenAI Assistants level


rogerbarreto:
### Description

OpenAI Assistants has a peculiar behavior that it doesn't allow Tools override when tools are provided at the Agent/Assistant creation. 

i.e:

The code below attempts to create a OpenAI Assistant with Function Tool Definition.

```csharp
var assistantOptions = new AssistantCreationOptions();

// Converts and AIFunction and add as a Tool Definition
assistantOptions.Tools.Add(ToOpenAIAssistantsFunctionToolDefinition(aiFunction));

// Create Hosted Assistant Agent
var assistantCreateResult = await client.CreateAssistantAsync(model, assistantOptions).ConfigureAwait(false);
```
After creating an agent like above, attempting to use chatClient passing any of the `AIFunction` tools provided above in the `chatOptions.Tools` will throw the redundancy error below:

```
Unhandled exception. System.ClientModel.ClientResultException: HTTP 400 (invalid_request_error: )
Parameter: tools[3]

Function tools must have unique names.
```

### Reproduction Steps

Create an Assistant Agent with tools and attempt to pass the same tools in a `chatClient.GetResponseAsync` call.

### Expected behavior

Ideally `AsIChatClient()` should provide an extra set of Agent/Assistant level `tools: []` parameters, so those tools can be used in `GetResponseAsync` or `GetStreamingResponseAsync` to check and remove redundancy. 

### Actual behavior

Creating an Assistant Agent with tools and attempt to pass the same tools in a `chatClient.GetResponseAsync` call gives back a 400 Bad Request with tool redudancy error

### Known Workarounds

1. Don't use hosted function calling.
2. The caller must have to track on his side whenever he assigned tools to the agent creation, capture manually the `FunctionInvokingChatClient` and add those tools as `AdditionalTools` (for hosted function calling to work) and avoid in any further calls to the chat client to provide those tools in the `chatOptions`.

PR #6896, opened 2025-10-06
Fix duplication between OpenAI Assistants pre-configured tools

stephentoub:
Assistants lets you configure an agent with function signatures it's implicitly aware of, rather than them needing to be provided per request. That, however, creates complication for callers, as if they provide that function in ChatTools.Options, it leads to the function's signature being sent as part of the request, and the duplication of it with the pre-configured function signature results in an error. It's possible to work around this, by simply not including that function in the request, but it's a natural thing for a developer to do, especially if they want the function to be automatically invoked when the model requests it. You can achieve that by putting the function into the FunctionInvocationChatClient's AdditionalTools, which exists for this kind of purpose, but that's harder to discover.

Rather than try something more complicated, this commit simply deduplicates all tools by putting them into a set, deduplicating any duplicates provided.

Fixes https://github.com/dotnet/extensions/issues/6743
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6896)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
@@ -306,6 +306,8 @@ internal sealed class OpenAIAssistantsChatClient : IChatClient
 
             if (options.Tools is { Count: > 0 } tools)
             {
+                HashSet<ToolDefinition> toolsOverride = new(ToolDefinitionNameEqualityComparer.Instance);
+
                 // If the caller has provided any tool overrides, we'll assume they don't want to use the assistant's tools.
                 // But if they haven't, the only way we can provide our tools is via an override, whereas we'd really like to
                 // just add them. To handle that, we'll get all of the assistant's tools and add them to the override list
@@ -318,10 +320,7 @@ internal sealed class OpenAIAssistantsChatClient : IChatClient
                         _assistantTools = assistant.Value.Tools;
                     }
 
-                    foreach (var tool in _assistantTools)
-                    {
-                        runOptions.ToolsOverride.Add(tool);
-                    }
+                    toolsOverride.UnionWith(_assistantTools);
                 }
 
                 // The caller can provide tools in the supplied ThreadAndRunOptions. Augment it with any supplied via ChatOptions.Tools.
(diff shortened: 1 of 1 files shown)


src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
```diff
@@ -543,4 +547,22 @@ void AppendSystemInstructions(string? toAppend)
...
+    /// for <see cref="FunctionToolDefinition"/> it compares names so that two function tool definitions with the
+    /// same name compare equally.
+    /// </summary>
+    private sealed class ToolDefinitionNameEqualityComparer : IEqualityComparer<ToolDefinition>
+    {
+        public static ToolDefinitionNameEqualityComparer Instance { get; } = new();
+
+        public bool Equals(ToolDefinition? x, ToolDefinition? y) =>
+            x is FunctionToolDefinition xFtd && y is FunctionToolDefinition yFtd ? xFtd.FunctionName.Equals(yFtd.FunctionName, StringComparison.Ordinal) :
+            EqualityComparer<ToolDefinition?>.Default.Equals(x, y);
```
eiriktsarpalis, line 562: Is this falling back to reference equality presumably?

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
```diff
@@ -543,4 +547,22 @@ void AppendSystemInstructions(string? toAppend)
...
+    /// Provides the same behavior as <see cref="EqualityComparer{ToolDefinition}.Default"/>, except
+    /// for <see cref="FunctionToolDefinition"/> it compares names so that two function tool definitions with the
+    /// same name compare equally.
+    /// </summary>
+    private sealed class ToolDefinitionNameEqualityComparer : IEqualityComparer<ToolDefinition>
+    {
+        public static ToolDefinitionNameEqualityComparer Instance { get; } = new();
+
+        public bool Equals(ToolDefinition? x, ToolDefinition? y) =>
+            x is FunctionToolDefinition xFtd && y is FunctionToolDefinition yFtd ? xFtd.FunctionName.Equals(yFtd.FunctionName, StringComparison.Ordinal) :
```
eiriktsarpalis, line 561: Not sure how `FunctionToolDefinition` is meant to work, but if its equality is anything like .NET methods (aka overloading is permissible) then shouldn't it also factor in its signature? In this case, this would correspond to the JSON schema.

eiriktsarpalis (approved):

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
stephentoub: It's actually customized equality, but it's currently broken:
https://github.com/Azure/azure-sdk-for-net/issues/53043

EDIT: Actually, that's for the corresponding code in Azure.AI.Agents.Persistent, not OpenAI.Assistants. For OpenAI.Assistants, it does appear to just be normal reference equality.

src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantsChatClient.cs
stephentoub: > then shouldn't it also factor in its signature?

In general with AI / LLMs / function calling, the only thing that matters is the name. The LLM requests a function by name.

PR #6900, opened 2025-10-08
Fix serialization of [Experimental] AIContent-derived types

stephentoub:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6900)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
@@ -48,6 +48,16 @@ public static partial class AIJsonUtilities
             Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
         };
 
+        // Temporary workaround: these types are [Experimental] and can't be added as [JsonDerivedType] on AIContent yet,
+        // or else consuming assemblies that used source generation with AIContent would implicitly reference them.
+        // Once they're no longer [Experimental] and added as [JsonDerivedType] on AIContent, these lines should be removed.
+        AddAIContentType(options, typeof(FunctionApprovalRequestContent), typeDiscriminatorId: "functionApprovalRequest", checkBuiltIn: false);
+        AddAIContentType(options, typeof(FunctionApprovalResponseContent), typeDiscriminatorId: "functionApprovalResponse", checkBuiltIn: false);
+        AddAIContentType(options, typeof(McpServerToolCallContent), typeDiscriminatorId: "mcpServerToolCall", checkBuiltIn: false);
+        AddAIContentType(options, typeof(McpServerToolResultContent), typeDiscriminatorId: "mcpServerToolResult", checkBuiltIn: false);
+        AddAIContentType(options, typeof(McpServerToolApprovalRequestContent), typeDiscriminatorId: "mcpServerToolApprovalRequest", checkBuiltIn: false);
+        AddAIContentType(options, typeof(McpServerToolApprovalResponseContent), typeDiscriminatorId: "mcpServerToolApprovalResponse", checkBuiltIn: false);
+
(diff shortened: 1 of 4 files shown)


stephentoub:
cc: @westey-m

stephentoub:
> Is there some context of what motivated this change?

Serializing a chat message that contains these types otherwise blows up.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.cs
```diff
@@ -186,11 +186,11 @@ static void NormalizeJsonNode(JsonNode? node)
         }
     }
 
-    private static void AddAIContentTypeCore(JsonSerializerOptions options, Type contentType, string typeDiscriminatorId)
+    private static void AddAIContentType(JsonSerializerOptions options, Type contentType, string typeDiscriminatorId, bool checkBuiltIn)
```
eiriktsarpalis, line 189: Is the idea of `checkBuiltIn` to avoid calling into `contentType.Assembly` when inputs are trusted?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
```diff
@@ -48,6 +48,16 @@ private static JsonSerializerOptions CreateDefaultOptions()
             Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
         };
 
+        // Temporary workaround: these types are [Experimental] and can't be added as [JsonDerivedType] on AIContent yet,
+        // or else consuming assemblies that used source generation with AIContent would implicitly reference them.
+        // Once they're no longer [Experimental] and added as [JsonDerivedType] on AIContent, these lines should be removed.
+        AddAIContentType(options, typeof(FunctionApprovalRequestContent), typeDiscriminatorId: "functionApprovalRequest", checkBuiltIn: false);
```
eiriktsarpalis, line 54: The issue I find with this approach is that this only really works when folks use `AIJsonUtilities.DefaultOptions`. For any other JSO (e.g. because they need to source generate a bunch of other types) users have to repeat the same sequence of incantations for the same contract to work. It's easy to forget, it can go out of sync if the list of derived types changes, and can create serious problems in production if incoherent configuration is shipped.

How about exposing an `AIJsonUtilities.ConfigureExperimentalTypes(JsonSerializerOptions options)` helper that  does all that on behalf of the user?

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.cs
stephentoub: It's to avoid the exception this would otherwise produce when calling the method with the six built-in types.

src/Libraries/Microsoft.Extensions.AI.Abstractions/Utilities/AIJsonUtilities.Defaults.cs
stephentoub: That's fair. However:
- We only expect these types to be [Experimental] temporarily. I'm not super excited at the idea of adding additional compensating experimental surface area.
- If a developer would benefit from that, they'd still need to perform a gesture to use it. If they're serializing without a custom JSO, they could instead just add AIJsonUtilities.DefaultOptions. And if they're serializing with a custom JSO, that JSO could (and arguably should) chain in AIJsonUtilities.DefaultOptions, in which case this is also addressed.

Let's go ahead with fixing the default case and then see if we need to do more.

abhishek4official:
Is it released in nuget. If so which version

stephentoub:
It'll be in a new version on nuget later today.

stephentoub:
https://www.nuget.org/packages/Microsoft.Extensions.AI.Abstractions/9.10.0

Issue #6288, opened 2025-04-12
ChatOptions.ToolMode doesn't enable require non-function tools


stephentoub:
With ToolMode, you can specify that no tools should be used, any tools should be used, at least one tool should be used, or a specific tool must be used. But for the last one, it's limited only to function tools. With other kinds of tools becoming available, like web search, systems are starting to let you require that these tools be used, but there's no way to do that today via ToolMode.

stephentoub:
One way I think might address this well would be to introduce a ToolMode-derived type that just wraps an AITool instance (or to augment the existing required mode type with the ability to be configured with an AITool. Then you could do things like:
```C#
options.Tools = [myAwesomeTool];
options.ToolMode = ChatToolMode.RequireSpecific(myAwesomeTool);
```

PR #6901, opened 2025-10-09
Add ChatToolMode.RequireSpecific(AITool)

stephentoub:
In the olden days (i.e. a few months ago), you could require the model/service to request invocation of any tool or of a specific function by name. Now, you can request that it invoke other tools as well. This adds another RequireSpecific overload that takes an AITool instead of a string function name, so that you can do things like RequireSpecific(webSearchTool).

Fixes https://github.com/dotnet/extensions/issues/6288
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6901)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
@@ -3,34 +3,59 @@
 
 using System;
 using System.Diagnostics;
+using System.Text.Json.Serialization;
 using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Extensions.AI;
 
 /// <summary>
-/// Represents a mode where a chat tool must be called. This class can optionally nominate a specific function
-/// or indicate that any of the functions can be selected.
+/// Represents a mode where a chat tool must be called. This class can optionally nominate a specific tool
+/// or indicate that any of the tools can be selected.
 /// </summary>
 [DebuggerDisplay("{DebuggerDisplay,nq}")]
 public sealed class RequiredChatToolMode : ChatToolMode
 {
     /// <summary>
-    /// Gets the name of a specific tool that must be called.
+    /// Gets the name of a specific function tool that must be called.
     /// </summary>
     /// <remarks>
-    /// If the value is <see langword="null"/>, any available tool can be selected (but at least one must be).
+    /// If both <see cref="RequiredFunctionName"/> and <see cref="RequiredTool"/> are <see langword="null"/>,
+    /// any available tool can be selected (but at least one must be).
     /// </remarks>
     public string? RequiredFunctionName { get; }
+
+    /// <summary>Gets the specific tool that must be called.</summary>
+    /// <remarks>
+    /// <para>
+    /// If both <see cref="RequiredFunctionName"/> and <see cref="RequiredTool"/> are <see langword="null"/>,
+    /// any available tool can be selected (but at least one must be).
+    /// </para>
+    /// <para>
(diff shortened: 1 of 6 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/RequiredChatToolMode.cs
```diff
@@ -3,34 +3,59 @@
...
+    /// </para>
+    /// <para>
+    /// Note that <see cref="RequiredTool"/> will not serialize to JSON as part of serializing
+    /// the <see cref="RequiredChatToolMode"/> instance, just as <see cref="ChatOptions.Tools"/> doesn't serialize. As such, attempting to
+    /// roundtrip a <see cref="RequiredChatToolMode"/> through JSON serialization may lead to the deserialized instance having <see cref="RequiredTool"/>
+    /// set to <see langword="null"/>.
+    /// </para>
+    /// </remarks>
+    [JsonIgnore]
+    public AITool? RequiredTool { get; }
```
eiriktsarpalis, line 41: If we end up roundtripping this value the resultant object is going to have a subtly different equality semantic, and will have impact in how the resultant object gets processed in the OpenAI provider. Is there a way we could potentially preserve its intended representation in such cases?

stephentoub: > If we end up roundtripping this value the resultant object is going to have a subtly different equality semantic

Will it? Do you just mean it'll no longer be equal to another instance that contains the tool since this one is null? I mean, yeah :) That's not just equality, it's no longer the same requirement. If you set it to an AIFunctionDeclaration, though, it should behave the same, as the name gets extracted.

> Is there a way we could potentially preserve its intended representation in such cases?

What did you have in mind?

eiriktsarpalis: > What did you have in mind?

Nothing specific, just asking if serializing the value could inherently change how it gets handled by the leaf clients.

stephentoub: > just asking if serializing the value could inherently change how it gets handled by the leaf clients.

It could, yes, as the AITool will be null after deserialization. That's already the case for ChatOptions.Tools, too. So if you serialize and deserialize a ChatOptions, you're going to lose both on the resulting instance.

We could revisit that. We could allow AITool to be serialized, and in most cases we can persist most data. The main (and important) thing we can't is the actual function implementation for AIFunction.

stephentoub: @eiriktsarpalis, alternative suggestions or approaches? What do you think about enabling AITools to be serialized, except an AIFunction would roundtrip as an AIFunctionDeclaration (i.e. it would lose its invocation ability)?

eiriktsarpalis: The only alternative I can think of is requiring tools by name only.

> What do you think about enabling AITools to be serialized, except an AIFunction would roundtrip as an AIFunctionDeclaration (i.e. it would lose its invocation ability)?

Seems reasonable, although that scheme would only work in the context of a `JsonConverter<AITool>`  but not for a `JsonConverter<AIFunction>` due to the lack of type relationship with `AIFunctionDefinition`.

stephentoub:
Will revisit combined with a RequireAny for selecting between multiple tools

Issue #6890, opened 2025-10-04
SummarizingChatReducer removes all FunctionCallContents and FunctionResultContents


jooooel:
### Description
Not sure if you would classify this as a bug, but I would be surprised if it was by design 😊 

A call to `SummarizingChatReducer.ReduceAsync` will remove any messages with `FunctionResultContent` or `FunctionCallContent` in their content list. This is done even is the message list isn't summarized at all. The returned message list never contain those messages.

This makes it very hard to store the correct history after a a call to `ReduceAsync`:
1. We don't know if summarization happened or not. We just get a list of fewer messages back.
2. Even if older history was summarized, we have lost more recent tool call/response messages.

### Reproduction Steps

```
            var chatClient = new AzureOpenAIClient(new Uri("https://example.com"), new ApiKeyCredential("123")).GetChatClient("model-name").AsIChatClient();
            var summarizingChatReducer = new SummarizingChatReducer(chatClient, targetCount: 99, threshold: 99);
            IEnumerable<ChatMessage> messages =
            [
                new()
                {
                    Role = ChatRole.User,
                    Contents = new List<AIContent>
                    {
                        new TextContent("Who is the current user?")
                    }
                },
                new()
                {
                    Role = ChatRole.Assistant,
                    Contents = new List<AIContent>
                    {
                        new FunctionCallContent("123", "GetCurrentUser"),
                    }
                },
                new()
                {
                    Role = ChatRole.Tool,
                    Contents = new List<AIContent>
                    {
                        new FunctionResultContent("123", "User is Bob")
                    }
                },
                new()
                {
                    Role = ChatRole.Assistant,
                    Contents = new List<AIContent>
                    {
                        new TextContent("The current user is Bob")
                    }
                }
            ];
            
            var summarizedMessages = await summarizingChatReducer.ReduceAsync(messages, cancellationToken);
            var messageCount = summarizedMessages.Count();
```

### Expected behavior

I would expect it to behave like the Semantic Kernels `ChatHistorySummarizationReducer` (https://github.com/microsoft/semantic-kernel/blob/90d158cbf8bd4598159a6fe64df745e56d9cbdf4/dotnet/src/SemanticKernel.Core/AI/ChatCompletion/ChatHistorySummarizationReducer.cs#L20) where only messages older than the threshold are summarized.

In this case all four messages should be returned (`messageCount` should be 4 but is 2).

[...]

PR #6908, opened 2025-10-09
Preserve function content in `SummarizingChatReducer`

MackinnonBuck:
## Summary

Fixes an issue where `SummarizingChatReducer` always removes function call content, even if it didn't get summarized.

## Description

This PR makes the following changes to `SummarizingChatReducer`:
1. `FunctionCallContent` and `FunctionResultContent` now get preserved in the returned message list.
2. To avoid orphaning assistant responses from their user messages, the reducer searches backward within the threshold window for a user message boundary, keeping up to `targetCount + threshold` messages.

Fixes https://github.com/dotnet/extensions/issues/6890
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6908)

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,208 +1,208 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Bcl.AsyncInterfaces" Version="9.0.9" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
+    <Dependency Name="Microsoft.Bcl.AsyncInterfaces" Version="9.0.10" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
       <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-runtime</Uri>
-      <Sha>893c2ebbd49952ca49e93298148af2d95a61a0a4</Sha>
+      <Sha>e1f19886fe3354963a4a790c896b3f99689fd7a5</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.Bcl.Memory" Version="9.0.9" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
+    <Dependency Name="Microsoft.Bcl.Memory" Version="9.0.10">
       <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-runtime</Uri>
-      <Sha>893c2ebbd49952ca49e93298148af2d95a61a0a4</Sha>
+      <Sha>e1f19886fe3354963a4a790c896b3f99689fd7a5</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.Bcl.TimeProvider" Version="9.0.9" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
+    <Dependency Name="Microsoft.Bcl.TimeProvider" Version="9.0.10" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
       <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-runtime</Uri>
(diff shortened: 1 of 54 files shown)


src/Libraries/Microsoft.Extensions.AI/ChatReduction/SummarizingChatReducer.cs
```diff
@@ -111,31 +118,68 @@ public static SummarizedConversation FromChatMessages(IEnumerable<ChatMessage> m
...
+            }
+
+            // Start at the ideal cut point (keeping exactly targetCount messages)
+            var indexOfFirstMessageToKeep = unsummarizedMessages.Count - targetCount;
+
+            // Move backward to skip over function call/result content at the boundary
+            // We want to keep complete function call sequences together with their responses
+            while (indexOfFirstMessageToKeep > 0)
+            {
+                if (!unsummarizedMessages[indexOfFirstMessageToKeep - 1].Contents.Any(c => c is FunctionCallContent or FunctionResultContent))
```
stephentoub: There's also UserInputRequestContent / UserInputResponseContent. Does this need to pay attention to those as well?

src/Libraries/Microsoft.Extensions.AI/ChatReduction/SummarizingChatReducer.cs
```diff
@@ -111,31 +118,68 @@ public static SummarizedConversation FromChatMessages(IEnumerable<ChatMessage> m
...
+            if (earliestAllowedIndex <= 0)
+            {
+                // Not enough messages to warrant summarization
+                return 0;
+            }
+
+            // Start at the ideal cut point (keeping exactly targetCount messages)
+            var indexOfFirstMessageToKeep = unsummarizedMessages.Count - targetCount;
+
+            // Move backward to skip over function call/result content at the boundary
```
stephentoub, line 156: So is this never getting rid of any FunctionCall/ResultContent? Does that mean in a really long conversation we could end up with lots and lots of these, and eventually still overflow context windows?

src/Libraries/Microsoft.Extensions.AI/ChatReduction/SummarizingChatReducer.cs
MackinnonBuck: Yes, thanks. Just updated the PR to check for these. If there's a chain of messages at the end of the chat history with `UserInputRequestContent` and/or `UserInputResponseContent`, then those will be preserved in the "unsummarized" portion of the chat history (same as function call/result content). Those messages will then be subject to summarization/removal after additional, non-tool-related messages get added to the chat history.

src/Libraries/Microsoft.Extensions.AI/ChatReduction/SummarizingChatReducer.cs
MackinnonBuck: This will still get rid of function call/result content. The logic here handles the case where you have a chain of function call/result content that crosses the boundary between the summarized and unsummarized chat history. 

It prevents scenarios like this:
- {summary}
- `FunctionResultContent`
- {assistant response}

...instead preferring this:
- {summary}
- `FunctionCallContent`
- `FunctionResultContent`
- {assistant response}

Function-related content can still get removed, so long as the call/result chain falls completely outside the `targetCount` window.

stephentoub: Ok, thanks. Do we have tests for that?

MackinnonBuck: > Do we have tests for that?

Yes, here:
https://github.com/dotnet/extensions/blob/afa8821d72ac98ea8af79f7057d4b1791e7100ff/test/Libraries/Microsoft.Extensions.AI.Tests/ChatReduction/SummarizingChatReducerTests.cs#L177

stephentoub (approved):

PR #6910, opened 2025-10-10
Update ToChatResponse{Async} to also factor in AuthorName

stephentoub:
Originally we were only factoring in Role. Then we also augmented it to factor in MessageId, since updates from different messages are by definition not part of the same message. Now it also makes sense to factor in AuthorName, as different text from different speakers / agents are also by definition not part of the same message.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6910)

### DIFF
--- a/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatResponseUpdateExtensionsTests.cs
+++ b/test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatResponseUpdateExtensionsTests.cs
@@ -36,8 +36,8 @@ public class ChatResponseUpdateExtensionsTests
         ];
 
         ChatResponse response = useAsync ?
-            updates.ToChatResponse() :
-            await YieldAsync(updates).ToChatResponseAsync();
+            await YieldAsync(updates).ToChatResponseAsync() :
+            updates.ToChatResponse();
         Assert.NotNull(response);
 
         Assert.NotNull(response.Usage);
@@ -88,8 +88,8 @@ public class ChatResponseUpdateExtensionsTests
         ];
 
         ChatResponse response = useAsync ?
-            updates.ToChatResponse() :
-            await YieldAsync(updates).ToChatResponseAsync();
+            await YieldAsync(updates).ToChatResponseAsync() :
+            updates.ToChatResponse();
         Assert.Equal(9, response.Messages.Count);
 
         Assert.Equal("!a", response.Messages[0].Text);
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @westey-m

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatResponseUpdateExtensionsTests.cs
```diff
@@ -120,6 +120,329 @@ public async Task ToChatResponse_RoleOrIdChangeDictatesMessageChange(bool useAsy
...
     }
 
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task ToChatResponse_AuthorNameChangeDictatesMessageBoundary(bool useAsync)
+    {
+        ChatResponseUpdate[] updates =
+        [
+
```
jozkee, line 130:
```suggestion
```
nit: there's an empty newline in all the arrays.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatResponseUpdateExtensionsTests.cs
```diff
@@ -120,6 +120,329 @@ public async Task ToChatResponse_RoleOrIdChangeDictatesMessageChange(bool useAsy
...
+            // Third message - AuthorName changes to "Charlie"
+            new(ChatRole.Assistant, "Greetings ") { AuthorName = "Charlie" },
+            new(null, "from Charlie!") { AuthorName = "Charlie" },
+
+            // Fourth message - AuthorName changes back to "Alice"
+            new(null, "Alice again!") { AuthorName = "Alice" },
+
+            // Fifth message - empty/null AuthorName should continue with last message
+            new(null, " Still Alice.") { AuthorName = "" },
+            new(null, " And more."),
```
jozkee, line 150: The updates from empty to non-empty will also continue in the same message, that's expected right?  for example, if you add `new(null, " Still Alice.") { AuthorName = "" },` at the begining (line 130) it will be included in Alice's message.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatResponseUpdateExtensionsTests.cs
stephentoub: There's an analyzer that annoyingly complains with an error diagnostic if that blank line is removed. We should subsequently turn off that analyzer.

test/Libraries/Microsoft.Extensions.AI.Abstractions.Tests/ChatCompletion/ChatResponseUpdateExtensionsTests.cs
stephentoub: > The updates from empty to non-empty will also continue in the same message, that's expected right?

Yup.

PR #6914, opened 2025-10-13
Extend service discovery to support Consul-based DNS lookups:

bart-vmware:
This PR provides the building blocks for https://github.com/dotnet/aspire/issues/11740.

Extend service discovery to support HashiCorp Consul-based DNS lookups:
- Enable specifying how to construct the DNS SRV query
- Enable adding the Consul DNS server host/port

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6914)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
@@ -48,16 +48,12 @@ internal sealed partial class DnsResolver : IDnsResolver, IDisposable
         }
     }
 
-    internal DnsResolver(DnsResolverOptions options) : this(TimeProvider.System, NullLogger<DnsResolver>.Instance, new OptionsWrapper<DnsResolverOptions>(options))
+    // This constructor is for unit testing only. Does not auto-add system DNS servers.
+    internal DnsResolver(DnsResolverOptions options)
     {
-    }
-
-    internal DnsResolver(IEnumerable<IPEndPoint> servers) : this(new DnsResolverOptions { Servers = servers.ToArray() })
-    {
-    }
-
-    internal DnsResolver(IPEndPoint server) : this(new DnsResolverOptions { Servers = [server] })
-    {
+        _timeProvider = TimeProvider.System;
+        _logger = NullLogger<DnsResolver>.Instance;
+        _options = options;
     }
 
     public ValueTask<ServiceResult[]> ResolveServiceAsync(string name, CancellationToken cancellationToken = default)
--- a/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/ServiceDiscoveryDnsServiceCollectionExtensions.cs
+++ b/src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/ServiceDiscoveryDnsServiceCollectionExtensions.cs
@@ -60,7 +60,7 @@ public static class ServiceDiscoveryDnsServiceCollectionExtensions
 
         services.AddSingleton<IServiceEndpointProviderFactory, DnsSrvServiceEndpointProviderFactory>();
         var options = services.AddOptions<DnsSrvServiceEndpointProviderOptions>();
-        options.Configure(o => configureOptions.Invoke(o));
+        options.Configure(configureOptions);
 
         return services;
 
@@ -109,7 +109,7 @@ public static class ServiceDiscoveryDnsServiceCollectionExtensions
 
         services.AddSingleton<IServiceEndpointProviderFactory, DnsServiceEndpointProviderFactory>();
         var options = services.AddOptions<DnsServiceEndpointProviderOptions>();
-        options.Configure(o => configureOptions.Invoke(o));
+        options.Configure(configureOptions);
 
         return services;
     }
--- a/test/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns.Tests/ServiceDiscoveryDnsServiceCollectionExtensionsTests.cs
+++ b/test/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns.Tests/ServiceDiscoveryDnsServiceCollectionExtensionsTests.cs
@@ -21,7 +21,9 @@ public class ServiceDiscoveryDnsServiceCollectionExtensionsTests
         services.AddDnsServiceEndpointProvider();
 
         using var serviceProvider = services.BuildServiceProvider(true);
-        _ = serviceProvider.GetServices<IServiceEndpointProviderFactory>();
+
+        var exception = Record.Exception(() => serviceProvider.GetServices<IServiceEndpointProviderFactory>());
+        Assert.Null(exception);
     }
 
     [Fact]
@@ -31,7 +33,9 @@ public class ServiceDiscoveryDnsServiceCollectionExtensionsTests
         services.AddDnsSrvServiceEndpointProvider();
 
         using var serviceProvider = services.BuildServiceProvider(true);
-        _ = serviceProvider.GetServices<IServiceEndpointProviderFactory>();
+
+        var exception = Record.Exception(() => serviceProvider.GetServices<IServiceEndpointProviderFactory>());
+        Assert.Null(exception);
     }
 
     [Fact]
--- a/test/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp.Tests/YarpServiceDiscoveryTests.cs
+++ b/test/Libraries/Microsoft.Extensions.ServiceDiscovery.Yarp.Tests/YarpServiceDiscoveryTests.cs
@@ -1,15 +1,16 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using Microsoft.Extensions.DependencyInjection;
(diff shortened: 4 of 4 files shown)


davidfowl:
@ReubenBond @rzikm can you take a look?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/ResolverOptions.cs
```diff
@@ -5,27 +5,26 @@
 
 namespace Microsoft.Extensions.ServiceDiscovery.Dns.Resolver;
 
-internal sealed class ResolverOptions
+/// <summary>
+/// Provides configuration options for DNS resolution, including server endpoints, retry attempts, and timeout settings.
+/// </summary>
+public class ResolverOptions
```
rzikm: Now that this is public API, we should review if the shape/naming makes sense.

Attempts => `MaxAttempt` is maybe a better name, and I feel the comment should mention that this number is per-server (not total)

Timeout - IIRC this applies per attempt (meaning that maximum time for resolving a query is `Attempts * Servers.Count * Timeout`). The comment could be improved to make this more clear.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderOptions.cs
```diff
@@ -36,8 +43,19 @@ public class DnsSrvServiceEndpointProviderOptions
     /// </remarks>
     public string? QuerySuffix { get; set; }
 
+    /// <summary>
+    /// Gets or sets a delegate that generates a DNS SRV query from a specified <see cref="ServiceEndpointQuery"/> instance.
+    /// </summary>
+    public Func<ServiceEndpointQuery, string>? GetQueryText { get; set; }
```
rzikm: Are we happy with the name here? I don't like the word Text in the name, and I am not sure if we have precedent for the Get* prefix for these callbacks.

Maybe something like `ServiceDomainNameCallback`?

+ same naming change to `GetDefaultQueryText`

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderOptions.cs
```diff
@@ -28,6 +30,11 @@ public class DnsSrvServiceEndpointProviderOptions
     /// </summary>
     public double RetryBackOffFactor { get; set; } = 2;
 
+    /// <summary>
+    /// Gets or sets the options used to configure the resolver's behavior.
+    /// </summary>
+    public ResolverOptions ResolverOptions { get; set; } = new();
+
```
rzikm: Nit re: naming, I expect users don't immediately get the connection between Resolver <=> DNS Client. At minimum, I would suggest adding Dns preffix.

```suggestion
    public double RetryBackOffFactor { get; set; } = 2;

    /// <summary>
    /// Gets or sets the options used to configure the underlying DNS resolution behavior.
    /// </summary>
    public DnsResolverOptions DnsResolverOptions { get; set; } = new();

```

I also wonder if this creates confusion as to `DnsResolverOptions` are separate since we have `RetryBackOffFactor` here.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderOptions.cs
```diff
@@ -36,8 +43,19 @@ public class DnsSrvServiceEndpointProviderOptions
...
+    /// Gets or sets a delegate that generates a DNS SRV query from a specified <see cref="ServiceEndpointQuery"/> instance.
+    /// </summary>
+    public Func<ServiceEndpointQuery, string>? GetQueryText { get; set; }
+
     /// <summary>
     /// Gets or sets a delegate used to determine whether to apply host name metadata to each resolved endpoint. Defaults to <c>false</c>.
     /// </summary>
     public Func<ServiceEndpoint, bool> ShouldApplyHostNameMetadata { get; set; } = _ => false;
+
+    internal string GetDefaultQueryText(ServiceEndpointQuery query)
```
rzikm: This feels like this does not belong to *Options* but rather to a place where `GetQueryText` is invoked

rzikm:
Also, do we not have to go through https://apireview.net/ since this is public API?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/ResolverOptions.cs
bart-vmware: Updated. Should this type be moved to the `Microsoft.Extensions.ServiceDiscovery.Dns` namespace or stay here? It used to be internal, but it is now the only public type in this subdirectory.

bart-vmware:
@rzikm Thanks for the review. I've pushed changes to address them. The link "Details about the pull request review procedure" at https://github.com/dotnet/extensions/blob/main/CONTRIBUTING.md#suggested-workflow is dead, so it's unclear to me what the process is. Can I assume you'll resolve each open conversation that is addressed as desired?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolver.cs
```diff
@@ -23,21 +25,31 @@ internal sealed partial class DnsResolver : IDnsResolver, IDisposable
...
+                : NetworkInfo.GetServers())
+            {
+                _options.Servers.Add(server);
+            }
+
+            if (_options.Servers.Count == 0)
+            {
+                throw new ArgumentException("At least one DNS server is required.", nameof(options));
+            }
+        }
```
rzikm: Please add validation of MaxAttempts here

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderOptions.cs
```diff
@@ -28,6 +30,11 @@ public class DnsSrvServiceEndpointProviderOptions
     /// </summary>
     public double RetryBackOffFactor { get; set; } = 2;
 
+    /// <summary>
+    /// Gets or sets the options used to configure the underlying DNS resolution behavior.
+    /// </summary>
+    public DnsResolverOptions DnsResolverOptions { get; set; } = new();
```
rzikm: I just realized, there is also DnsServiceEndpointProviderOptions (without Srv in the name). both are using the singleton DnsResolver under the hood.

This creates a gap where users can't configure DnsResolverOptions when plain A/AAAA based resolution is to be used.

It makes me believe that DnsResolverOptions should be moved out and configured separately (and apply for both A/AAAA and SRV-based discovery). Perhaps something like `services.ConfigureDnsResolver(Action<DnsResolverOptions>)`?

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/Resolver/DnsResolverOptions.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Net;
+
+namespace Microsoft.Extensions.ServiceDiscovery.Dns.Resolver;
```
rzikm: I think we should move this outside the .Resolver subnamespace. It's the only part of the namespace that would be public.

src/Libraries/Microsoft.Extensions.ServiceDiscovery.Dns/DnsSrvServiceEndpointProviderOptions.cs
bart-vmware: Yes, I wondered about that. But didn't see any code that registers `IDnsResolver` when `AddDnsServiceEndpointProvider` is called. Turns out it doesn't even work in Aspire:

{B325E27A-7E82-4FC3-A022-F7690B92548E}

I initially considered using the options pattern for `DnsResolverOptions`, but decided against it to keep it simple.

rzikm: > But didn't see any code that registers IDnsResolver when AddDnsServiceEndpointProvider is called. Turns out it doesn't even work in Aspire

That looks like a bug we should fix that either in this PR or separately if this PR takes longer.

bart-vmware: @rzikm I've made the requested changes. Can you please take another look?

bart-vmware:
> Also, do we not have to go through https://apireview.net/ since this is public API?

@davidfowl Changes were reviewed and approved. Do we need public API review first, or is this ready to be merged?
And once merged, how does it land in the Aspire codebase?

davidfowl:
As long as @rzikm and @ReubenBond approve, I'm good!

bart-vmware:
@rzikm Can you merge this?

PR #6930, opened 2025-10-17
Doc updates

gewarren:
Contributes to https://github.com/dotnet/dotnet-api-docs/issues/11392.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6930)

### DIFF
--- a/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
+++ b/src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
@@ -99,6 +99,11 @@ public static partial class AIFunctionFactory
     /// <para>
     /// By default, return values are serialized to <see cref="JsonElement"/> using <paramref name="options"/>'s
     /// <see cref="AIFunctionFactoryOptions.SerializerOptions"/> if provided, or else using <see cref="AIJsonUtilities.DefaultOptions"/>.
+    /// However, return values whose declared type is <see cref="AIContent"/>, a derived type of <see cref="AIContent"/>, or
+    /// any type assignable from <see cref="IEnumerable{AIContent}"/> (e.g. <c>AIContent[]</c>, <c>List&lt;AIContent&gt;</c>) are
+    /// special-cased and are not serialized: the created function returns the original instance(s) directly to enable
+    /// callers (such as an <c>IChatClient</c>) to perform type tests and implement specialized handling. If
+    /// <see cref="AIFunctionFactoryOptions.MarshalResult"/> is supplied, that delegate governs the behavior instead.
     /// Handling of return values can be overridden via <see cref="AIFunctionFactoryOptions.MarshalResult"/>.
     /// </para>
     /// </remarks>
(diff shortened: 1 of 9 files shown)


src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -165,12 +165,12 @@ protected ChatOptions(ChatOptions? other)
     /// and polled for completion by non-streaming APIs.
     /// </para>
     /// <para>
-    /// When this property is set to true, non-streaming APIs may start a background operation and return an initial
+    /// When this property is set to true, non-streaming APIs can start a background operation and return an initial
```
stephentoub: I'm unclear why some of these are changing from "may" to "can". Regardless of whether the ability to create a background operation (which I think of as "can", having an ability), this property controls whether it's permitted to (which I think of as "may"). Can you help me understand the reasoning?

gewarren: Our grammar check tool wants us to avoid "may" because it's ambiguous about whether it implies possibility (use "might" instead) or permission (use "can" instead). See also the [writing style guide](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/can-may?toc=%2Fwriting-style-guide-msft-internal%2Ftoc.json&bc=%2Fwriting-style-guide-msft-internal%2Fbreadcrumb%2Ftoc.json). But if you prefer to keep "may", that's fine with me.

stephentoub: This "may" here is "are permitted to"; I don't think "might" conveys that. I'd prefer if this one not change, or change to something else.

gewarren: This one changed to "can" - is that okay?

gewarren:
```suggestion
    /// When this property is set to <see langword="true" />, non-streaming APIs have permission to start a background operation and return an initial
```

src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs
```diff
@@ -165,12 +165,12 @@ protected ChatOptions(ChatOptions? other)
...
     /// response with a continuation token. Subsequent calls to the same API should be made in a polling manner with
     /// the continuation token to get the final result of the operation.
     /// </para>
     /// <para>
-    /// When this property is set to true, streaming APIs may also start a background operation and begin streaming
+    /// When this property is set to true, streaming APIs can also start a background operation and begin streaming
```
gewarren:
```suggestion
    /// When this property is set to <see langword="true" />, streaming APIs are also permitted to start a background operation and begin streaming
```

PR #6935, opened 2025-10-20
Special-case AIContent returned from AIFunctionFactory.Create AIFunctions to not be serialized

stephentoub:
They'll likely end up being serialized, anyway, by leaf IChatClients, but this gives those IChatClients the opportunity to do something different, such as by treating DataContent differently.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/extensions/pull/6935)


src/Libraries/Microsoft.Extensions.AI.Abstractions/Functions/AIFunctionFactory.cs
```diff
@@ -1015,6 +1029,19 @@ static void ThrowNullServices(string parameterName) =>
                         };
                     }
 
+                    // Special-case AIContent results to not be serialized, so that IChatClients can type test and handle them
```
eiriktsarpalis, line 1045: Should we be putting this in API documentation anywhere?

stephentoub: Yup, added

